PR #240, opened 2016-08-13
Using custom version of Python Social Auth

clintonb:
This resolves migration issues related to the changing of the app's name.

ECOM-5232


clintonb:
FYI @edx/ecommerce @edx/devops

rlucioni:
👍

PR #247, opened 2016-08-15
Improved Docker Support

clintonb:
The addition of a proper local volume mapping for Elasticsearch and MySQL ensures developers don't lose data between runs of the containers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,4 @@ private.py
 docs/_build/
 course_discovery/static/bower_components/
 node_modules/
+docker/volumes/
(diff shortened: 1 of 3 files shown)


clintonb:
FYI @edx/ecommerce

cpennington:
Note: OEP-5 covers local volume mounts: http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0005.html#requirements (point #4). I think the main takeaway is: can you make this configurable?

clintonb:
@cpennington will update later today.

clintonb:
@cpennington before I dig into environment variables+YAML+Docker Compose, do we have an example of this already?

cpennington:
No, but I'm pretty sure you can just use them: https://docs.docker.com/compose/environment-variables/

PR #248, opened 2016-08-16
Allow programs to be filtered by type

rlucioni:
ECOM-5167

@edx/ecommerce please review. FYI @mikedikan.


clintonb:
Use [`django-filter`](http://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend).

rlucioni:
Thanks for the tip @clintonb, updated.

clintonb:
👍

PR #249, opened 2016-08-16
Updated course and course run data models and data loaders

clintonb:
ECOM-5195 and ECOM-5196

### DIFF
--- a/course_discovery/apps/api/fields.py
+++ b/course_discovery/apps/api/fields.py
@@ -24,3 +24,17 @@ class StdImageSerializerField(serializers.Field):
     def to_internal_value(self, obj):
         """ We do not need to save/edit this banner image through serializer yet """
         pass
+
+
+class ImageField(serializers.Field):  # pylint:disable=abstract-method
+    """ This field mimics the format of `ImageSerializer`. It is intended to aid the transition away from the
+    `Image` model to simple URLs.
+    """
+
+    def to_representation(self, value):
+        return {
+            'src': value,
+            'description': None,
+            'height': None,
+            'width': None
+        }
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -9,7 +9,7 @@ from rest_framework import serializers
 from rest_framework.fields import DictField
 from taggit_serializer.serializers import TagListSerializerField, TaggitSerializer
 
-from course_discovery.apps.api.fields import StdImageSerializerField
+from course_discovery.apps.api.fields import StdImageSerializerField, ImageField
 from course_discovery.apps.catalogs.models import Catalog
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType,
@@ -210,10 +210,10 @@ class CourseRunSerializer(TimestampModelSerializer):
         help_text=_('Language in which the course is administered')
     )
     transcript_languages = serializers.SlugRelatedField(many=True, read_only=True, slug_field='code')
-    image = ImageSerializer()
+    image = ImageField(read_only=True, source='card_image_url')
     video = VideoSerializer()
     seats = SeatSerializer(many=True)
-    instructors = PersonSerializer(many=True)
+    instructors = serializers.SerializerMethodField(help_text='This field is deprecated. Use staff.')
     staff = PersonSerializer(many=True)
     marketing_url = serializers.SerializerMethodField()
     level_type = serializers.SlugRelatedField(read_only=True, slug_field='name')
@@ -229,6 +229,9 @@ class CourseRunSerializer(TimestampModelSerializer):
 
     def get_marketing_url(self, obj):
         return get_marketing_url_for_user(self.context['request'].user, obj.marketing_url)
+
+    def get_instructors(self, obj):  # pylint: disable=unused-argument
+        return []
 
 
 class CourseRunWithProgramsSerializer(CourseRunSerializer):
@@ -254,10 +257,10 @@ class CourseSerializer(TimestampModelSerializer):
     subjects = SubjectSerializer(many=True)
     prerequisites = PrerequisiteSerializer(many=True)
     expected_learning_items = serializers.SlugRelatedField(many=True, read_only=True, slug_field='value')
-    image = ImageSerializer()
+    image = ImageField(read_only=True, source='card_image_url')
     video = VideoSerializer()
-    owners = OrganizationSerializer(many=True)
-    sponsors = OrganizationSerializer(many=True)
+    owners = OrganizationSerializer(many=True, source='authoring_organizations')
+    sponsors = OrganizationSerializer(many=True, source='sponsoring_organizations')
     course_runs = CourseRunSerializer(many=True)
     marketing_url = serializers.SerializerMethodField()
 
@@ -344,7 +347,7 @@ class AffiliateWindowSerializer(serializers.ModelSerializer):
     name = serializers.CharField(source='course_run.title')
     desc = serializers.CharField(source='course_run.short_description')
     purl = serializers.CharField(source='course_run.marketing_url')
-    imgurl = serializers.CharField(source='course_run.image')
+    imgurl = serializers.CharField(source='course_run.card_image_url')
     category = serializers.SerializerMethodField()
     price = serializers.SerializerMethodField()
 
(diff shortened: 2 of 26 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/apps/api/serializers.py
```diff
@@ -343,7 +346,7 @@ class AffiliateWindowSerializer(serializers.ModelSerializer):
     name = serializers.CharField(source='course_run.title')
     desc = serializers.CharField(source='course_run.short_description')
     purl = serializers.CharField(source='course_run.marketing_url')
-    imgurl = serializers.CharField(source='course_run.image')
+    imgurl = serializers.CharField(source='course_run.card_image_url')
```
rlucioni, line 350: Would naming this field `card_image_url` would require significant changes elsewhere?

course_discovery/apps/api/fields.py
```diff
@@ -24,3 +24,17 @@ def to_representation(self, obj):
     def to_internal_value(self, obj):
         """ We do not need to save/edit this banner image through serializer yet """
         pass
+
+
+class ImageField(serializers.Field):  # pylint:disable=abstract-method
+    """ This field mimics the format of `ImageSerializer`. It is intended to aid the transition away from the
+    `Image` model to simple URLs.
```
rlucioni, line 31: For those of us with less context, what was the Image model used for and what's the motivation for moving away from it?

course_discovery/apps/api/serializers.py
```diff
@@ -254,10 +257,10 @@ class CourseSerializer(TimestampModelSerializer):
     subjects = SubjectSerializer(many=True)
     prerequisites = PrerequisiteSerializer(many=True)
     expected_learning_items = serializers.SlugRelatedField(many=True, read_only=True, slug_field='value')
-    image = ImageSerializer()
+    image = ImageField(read_only=True, source='card_image_url')
     video = VideoSerializer()
-    owners = OrganizationSerializer(many=True)
-    sponsors = OrganizationSerializer(many=True)
+    owners = OrganizationSerializer(many=True, source='authoring_organizations')
+    sponsors = OrganizationSerializer(many=True, source='sponsoring_organizations')
```
rlucioni, line 263: I see methods for setting these below. Do these need to be `SerializerMethodField`s?

course_discovery/apps/api/serializers.py
```diff
@@ -427,10 +431,10 @@ def get_seats(self, obj):
         return seats
 
     def get_owners(self, obj):
-        return ','.join([owner.key for owner in obj.course.owners.all()])
+        return ','.join([owner.key for owner in obj.course.authoring_organizations.all()])
 
     def get_sponsors(self, obj):
-        return ','.join([sponsor.key for sponsor in obj.course.sponsors.all()])
+        return ','.join([sponsor.key for sponsor in obj.course.sponsoring_organizations.all()])
```
rlucioni, line 438: Should these methods return arrays and leave conversion to a string to the serializer?

course_discovery/apps/api/serializers.py
clintonb, line 350: This schema is defined by Affiliate Window.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -94,52 +96,49 @@ def ingest(self):
...
     def update_course(self, body):
-        # NOTE (CCB): Use the data from the CourseKey since the Course API exposes display names for org and number,
-        # which may not be unique for an organization.
-        course_run_key_str = body['id']
-        course_run_key = CourseKey.from_string(course_run_key_str)
-        organization, __ = Organization.objects.get_or_create(key=course_run_key.org,
-                                                              defaults={'partner': self.partner})
-        course_key = self.convert_course_run_key(course_run_key_str)
+        course_run_key = CourseKey.from_string(body['id'])
+        course_key = self.convert_course_run_key(body['id'])
```
rlucioni: You can pass `course_run_key` as the argument here instead of looking up `id` again.

course_discovery/apps/api/fields.py
clintonb, line 31: The `Image` model was intended to store metadata about images—dimensions, alt text—in addition to the URL. We don't actually have access to this information, so we shouldn't keep the model around pretending that we ever will. Additionally, the merging of programs along with the issues we've seen with missing URLs on stage made it obvious that we should control this data if we are going to serve it to users.

course_discovery/apps/course_metadata/data_loaders/api.py
rlucioni: Or maybe this method can't deal with course keys.

course_discovery/apps/api/serializers.py
clintonb, line 263: There are no methods.

course_discovery/apps/api/serializers.py
clintonb, line 438: This is the serializer. `FlattenedCourseRunWithCourseSerializer` is used to produce a (somewhat dirty) CSV.

course_discovery/apps/api/serializers.py
rlucioni, line 263: Sorry, they were on [another serializer](https://github.com/edx/course-discovery/pull/249/files#diff-651fda19c2550b81fbdc3c7de19a0dbbR433). Darn GitHub code collapsing.

course_discovery/apps/course_metadata/migrations/0020_auto_20160817_2011.py
```diff
@@ -0,0 +1,215 @@
...
+def delete_partnerless_courses(apps, schema_editor):
+    Course = apps.get_model('course_metadata', 'Course')
+    Course.objects.filter(partner__isnull=True).delete()
+
+
+def add_uuid_to_courses_and_course_runs(apps, schema_editor):
+    Course = apps.get_model('course_metadata', 'Course')
+    CourseRun = apps.get_model('course_metadata', 'CourseRun')
+
+    for objects in (Course.objects.filter(uuid__isnull=True),  CourseRun.objects.filter(uuid__isnull=True)):
```
rlucioni: Can you avoid the need for this by setting `unique=True` on the UUID field so that Django uses the callable provided as a default?

course_discovery/templates/search/indexes/course_metadata/courserun_text.txt
```diff
@@ -1,7 +1,12 @@
 {% include 'search/indexes/course_metadata/basecourse_text.txt' %}
```
rlucioni: What's this template used for?

rlucioni:
@clintonb looks nice, the only outstanding thing for me is my question about adding a unique constraint to the new UUID fields.

course_discovery/apps/course_metadata/migrations/0020_auto_20160817_2011.py
rlucioni: I forgot that you have a compound index defined using `uuid`.

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -501,3 +312,132 @@ def test_ingest(self):
...
+
+    def assert_course_run_loaded(self, data):
+        course = self._get_course(data)
+        course_run = course.course_runs.get(uuid=data['uuid'])
+        language_names = [language['name'] for language in data['field_course_languages']]
+        language = self.loader.get_language_tags_from_names(language_names).first()
+
+        expected_values = {
+            'key': data['field_course_id'],
+            'uuid': data['uuid'],
```
rlucioni: Your test failure is because this is a string, but you're comparing it against a UUID object. You can get around this by passing the string value here to uuid.UUID:

``` py
...
'uuid': uuid.UUID(data['uuid']),
...
```

rlucioni:
👍

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -94,52 +96,49 @@ def ingest(self):
...
-        course_run_key = CourseKey.from_string(course_run_key_str)
-        organization, __ = Organization.objects.get_or_create(key=course_run_key.org,
-                                                              defaults={'partner': self.partner})
-        course_key = self.convert_course_run_key(course_run_key_str)
+        course_run_key = CourseKey.from_string(body['id'])
+        course_key = self.get_course_key_from_course_run_key(course_run_key)
+
         defaults = {
+            'key': course_key,
             'title': body['name'],
```
schenedx, line 104: @clintonb With course title retrieved from LMS here, it would override the course title retrieved from Drupal, right? Because the order of the data_loader you have is Drupal first, then this data_loader. With the order of operation, we won't satisfy the requirements on ticket ECOM-4524 and ECOM-4981, correct?

clintonb, line 104: That's correct. Fixing.

clintonb, line 104: Fixed.

clintonb:
@edx/ecommerce comments addressed. I welcome more reviewers as this is one of the largest, and most-important, data model updates.

schenedx:
@clintonb Can you iterate through the set of manual test cases you ran to validate these changes?

clintonb:
@schenedx I ran the refresh_course_metadata command against production, and took a look at the data and any logged errors. There are some errors related to invalid data for old course runs; but, I don't think they are worth fixing.

schenedx:
👍 Please squash commits

clintonb:
Thanks @schenedx. These commits are intentionally separated.

PR #250, opened 2016-08-16
Allow programs to be filtered by comma-separated UUIDs

rlucioni:
ECOM-5261

@edx/ecommerce please review. @AlasdairSwan FYI.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -8,6 +8,8 @@ from course_discovery.apps.course_metadata.tests.factories import ProgramFactory
 
 
 class ProgramViewSetTests(APITestCase):
+    list_path = reverse('api:v1:program-list')
+
     def setUp(self):
         super(ProgramViewSetTests, self).setUp()
         self.user = UserFactory(is_staff=True, is_superuser=True)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -54,3 +54,16 @@ def test_filter_by_type(self):
 
         response = self.client.get(url + 'bar')
         self.assertEqual(response.data['results'], [])
+
+    def test_filter_by_uuids(self):
+        """ Verify that the endpoint filters programs to those matching the provided UUIDs. """
+        url = reverse('api:v1:program-list') + '?uuids='
```
clintonb: Minor: Move `reverse('api:v1:program-list')` to a class variable (e.g. `list_path`) to DRY up this class.

clintonb:
@rlucioni any idea if casing of the UUID matters?

👍

course_discovery/apps/api/v1/tests/test_views/test_programs.py
rlucioni: Good call, will update.

rlucioni:
@clintonb I suspect it may, will experiment and see what I can do to make it case-insensitive. Good catch.

rlucioni:
@clintonb lookups on UUIDFields appear to be case-insensitive:

``` py
>>> from course_discovery.apps.course_metadata.models import Program
>>> Program.objects.all()
[<Program: foo>]
>>> Program.objects.get().uuid
UUID('a4dc1540-f75e-4637-ae53-be4850593510')
>>> Program.objects.filter(uuid='a4dc1540-f75e-4637-ae53-be4850593510')
[<Program: foo>]
>>> Program.objects.filter(uuid='a4dc1540-f75e-4637-ae53-be4850593510'.upper())
[<Program: foo>]
```

PR #251, opened 2016-08-17
Miscellaneous updates

clintonb:
These updates will simplify the forthcoming work for updating the Course and CourseRun models/data loaders.

ECOM-5094


clintonb:
FYI @edx/ecommerce

rlucioni:
👍

PR #253, opened 2016-08-17
ECOM-5248

awais786:
Implement [autocomplete](http://django-autocomplete-light.readthedocs.io/en/master/index.html)
Add lookups for courses , videos and orgs.
Add test cases.

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -71,18 +71,16 @@ class ProgramAdmin(admin.ModelAdmin):
 
     search_fields = ('uuid', 'title', 'marketing_slug')
 
-    filter_horizontal = (
-        'job_outlook_items', 'expected_learning_items',
-        'credit_backing_organizations', 'authoring_organizations',
-    )
+    filter_horizontal = ('job_outlook_items', 'expected_learning_items',)
 
     # ordering the field display on admin page.
     fields = (
-        'title', 'status', 'type', 'banner_image',
+        'title', 'status', 'type', 'partner', 'banner_image',
         'banner_image_url', 'card_image_url', 'overview', 'video',
     )
     fields += (
-        'courses', 'course_runs', 'excluded_course_runs'
+        'courses', 'course_runs', 'excluded_course_runs', 'authoring_organizations',
+        'credit_backing_organizations'
     )
     fields += filter_horizontal
 
--- a/course_discovery/apps/course_metadata/lookups.py
+++ b/course_discovery/apps/course_metadata/lookups.py
@@ -1,0 +1,41 @@
+from django.db.models import Q
+
+from dal import autocomplete
+from .models import Course, Organization, Video
+
+
+class CourseAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
+            qs = Course.objects.all()
+            if self.q:
+                qs = qs.filter(Q(key__icontains=self.q) | Q(title__icontains=self.q))
+
+            return qs
+
+        return []
+
+
+class OrganizationAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
+            qs = Organization.objects.all()
+
+            if self.q:
+                qs = qs.filter(Q(key__icontains=self.q) | Q(name__icontains=self.q))
+
+            return qs
+
+        return []
+
+
+class VideoAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
+            qs = Video.objects.all()
+            if self.q:
+                qs = qs.filter(Q(description__icontains=self.q) | Q(src__icontains=self.q))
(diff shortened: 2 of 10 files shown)


course_discovery/apps/course_metadata/admin.py
```diff
@@ -19,6 +19,31 @@ class PositionInline(admin.TabularInline):
     extra = 0
 
 
+class CorporateEndorsementInline(admin.TabularInline):
+    model = Program.corporate_endorsements.through
```
clintonb: This should be `CorporateEndorsement`. The same for the others below.

awais786:
@bderusha I have added the validation. 
screen shot 2016-08-19 at 12 01 44 am

awais786:
@bderusha Please review this PR. I will complete the tests cases tomorrow.

bderusha:
@awais786 any updates on how feasible it would be to only populate the `excluded_course_runs` field with course_runs with those of currently saved courses?

awais786:
@bderusha unfortunately its not working at the moment. I will try again tomorrow and let you know.

course_discovery/apps/course_metadata/admin.py
awais786: @clintonb sorry for the late reply.  For many to many relation `through` is used for accessing the table.  See [Manytomany Relation](https://docs.djangoproject.com/en/dev/ref/contrib/admin/#working-with-many-to-many-models).

awais786:
New PR created using default admin https://github.com/edx/course-discovery/pull/265

awais786:
@waheedahmed @bderusha Please review this PR.

bderusha:
Can we make the video dropdown show the description rather than the url?  Or both?

bderusha:
I'm not sure that courses should be a required field to save.  I'd imagine that it is possible we would want to create the program shell before the courses exist?  @clintonb thoughts?

clintonb:
@bderusha courses should be required to publish. I am fine with loosening restrictions for saves.

awais786:
@bderusha in that case same restriction like banners can be added "Active status should have some courses and course runs". And allow the empty courses at the time of program creation.

bderusha:
@awais786 Yes, I think that makes sense

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -0,0 +1,40 @@
...
+
+from .models import Course, Organization, Video
+
+
+class CourseAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
+            qs = Course.objects.all()
+            if self.q:
+                qs = qs.filter(key__istartswith=self.q)
```
bderusha: can we filter on key OR title using `icontains`

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -0,0 +1,40 @@
...
+        return []
+
+
+class OrganizationAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
+            qs = Organization.objects.all()
+
+            if self.q:
+                qs = qs.filter(key__istartswith=self.q)
```
bderusha: same here

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -0,0 +1,40 @@
...
+
+        return []
+
+
+class VideoAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
+            qs = Video.objects.all()
+            if self.q:
+                qs = qs.filter(description__istartswith=self.q)
```
bderusha: let's filter this on src and description using `icontains`

course_discovery/apps/course_metadata/lookups.py
bderusha: Can we also update this to make the video results populate with both src and description

awais786:
@bderusha In this PR I have removed constraint of courses for a program. I have created a followup story. [ECOM-5385](https://openedx.atlassian.net/browse/ECOM-5385)

bderusha:
Can we make the course_runs render the same as excluded_course_runs

screen shot 2016-08-26 at 11 29 41 am

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -153,3 +152,14 @@ def valid_post_form(self, data, file_data):
...
         self.assertEqual(response.status_code, 200)
+
+    def test_new_program_without_courses(self):
+        """ Verify that new program can be added without `courses`."""
+        data = self._post_data(Program.ProgramStatus.Unpublished)
+        data['courses'] = []
+        form = ProgramAdminForm(data)
+        self.assertTrue(form.is_valid())
+        program = form.save()
+        self.assertEqual(program.courses.all(), [])
```
bderusha: `self.assertEqual(len(program.courses.all()), 0)`

awais786:
@bderusha I tried that earlier. But it need some more time. I will try to find a solution again.

awais786:
PR is ready for final review.

bderusha:
👍

PR #255, opened 2016-08-18
Add authoring organizations to programs in search results

rlucioni:
@clintonb how off-base am I here?

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -1,3 +1,5 @@
+import json
+
 from haystack import indexes
 from opaque_keys.edx.keys import CourseKey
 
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -117,6 +123,7 @@ class ProgramIndex(BaseIndex, indexes.Indexable, OrganizationsMixin):
     organizations = indexes.MultiValueField(faceted=True)
     authoring_organizations = indexes.MultiValueField(faceted=True)
     credit_backing_organizations = indexes.MultiValueField(faceted=True)
+    organization_bodies = indexes.CharField()
```
rlucioni: Should this be an unfaceted `MultiValueField`? It's a list of serialized organizations.

clintonb: Yes.

clintonb:
@rlucioni this looks good to me.

rlucioni:
@clintonb cool, thanks for taking a look. I'll look into adding some tests later this evening.

rlucioni:
@clintonb please review. @AlasdairSwan FYI.

pylintrc
```diff
@@ -80,7 +80,7 @@ disable =
 	too-many-arguments,
 	too-many-locals,
 	unused-wildcard-import,
-	duplicate-code,invalid-name,missing-docstring
+	duplicate-code,invalid-name,missing-docstring,cyclic-import
```
clintonb, line 83: We don't want to disable this globally. Disable it for the one line that matters. There may be future cases where a circular import is not intended.

clintonb:
👍 (once pylint comment addressed)

pylintrc
rlucioni, line 83: Unfortunately, that doesn't appear to currently be possible. See https://github.com/PyCQA/pylint/issues/850. Pylint won't accept a module-scoped disable directive for this.

clintonb, line 83: Please add a note, and link to that issue in this file.

rlucioni, line 83: I've done that in `pylintrc_tweaks`. I was under the impression we should try to avoid editing this file.

clintonb, line 83: I missed that. Carry on.

PR #256, opened 2016-08-18
Allow ProgramSearchSerializer to deal with missing organization bodies

rlucioni:
@clintonb please review. The scenario this fix addresses only arises if code is deployed before `update_index` is run.


clintonb:
👍

PR #260, opened 2016-08-18
Restrict serialized program organizations to authoring ones only

rlucioni:
@clintonb please review. @edx/ecommerce FYI.


clintonb:
👍

PR #261, opened 2016-08-19
Code Cleanup

clintonb:
This is in preparation for the forthcoming course/course run changes.

ECOM-5094


clintonb:
FYI @edx/ecommerce

rlucioni:
👍

PR #263, opened 2016-08-19
Update the Swagger doc label for the catalog API

pdesjardins:
Update the label that's shown at the top of the swagger doc. It is currently "course discovery API," but I understand that the preferred terminology is "catalog API."

@clintonb Based on our conversation, should this label change to "catalog API"?


clintonb:
👍

PR #264, opened 2016-08-19
WIP elasticsearch boosting

bderusha:
[ECOM-5314](https://openedx.atlassian.net/browse/ECOM-5314) Boosting

FYI @clintonb @rlucioni

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/backends.py
+++ b/course_discovery/apps/edx_haystack_extensions/backends.py
@@ -1,5 +1,7 @@
 from haystack.backends.elasticsearch_backend import ElasticsearchSearchBackend, ElasticsearchSearchEngine
 
+from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
+
 
 class SimpleQuerySearchBackendMixin(object):
     """
@@ -34,10 +36,21 @@ class SimpleQuerySearchBackendMixin(object):
             'auto_generate_phrase_queries': True,
         }
 
-        if search_kwargs['query'].get('filtered', {}).get('query', {}).get('query_string'):
-            search_kwargs['query']['filtered']['query']['query_string'] = simple_query
+        # https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-function-score-query.html
+        function_score_config = ElasticsearchBoostConfig.get_solo().function_score
+
+        function_score_config['query'] = {
+            'query_string': simple_query
+        }
+
+        function_score = {
+            'function_score': function_score_config
+        }
+
+        if search_kwargs['query'].get('filtered', {}).get('query'):
+            search_kwargs['query']['filtered']['query'] = function_score
         elif search_kwargs['query'].get('query_string'):
-            search_kwargs['query']['query_string'] = simple_query
+            search_kwargs['query'] = function_score
 
         return search_kwargs
 
--- a/course_discovery/apps/edx_haystack_extensions/models.py
+++ b/course_discovery/apps/edx_haystack_extensions/models.py
@@ -1,0 +1,23 @@
+from django.utils.translation import ugettext_lazy as _
+from jsonfield.fields import JSONField
+from solo.models import SingletonModel
+
+
+class ElasticsearchBoostConfig(SingletonModel):
+    """
+    Model used to store the elasticsearch boost configuration.
+    This includes a default JSON config for the function_score.
+    """
+
+    function_score = JSONField(
+        verbose_name=_('Function Score'),
+        help_text=_('JSON string containing an elasticsearch function score config.'),
+        null=False,
+        blank=False,
+        default={
+            'functions': [],
+            'boost': 1.0,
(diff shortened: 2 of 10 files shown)


course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -28,16 +28,119 @@ def build_search_kwargs(self, *args, **kwargs):
...
+            # Date boost upcoming courses
+            {
+                'filter': {
+                    'exists': {
+                        'field': 'start'
+                    }
+                },
+                'linear': {
+                    'start': {
+                        'origin': '2016-08-19',
```
clintonb: 'now' is a valid value here.

clintonb:
I played around with this branch locally for a few minutes. This is neat!!!

bderusha:
@clintonb @rlucioni Please take a look

course_discovery/apps/api/models.py
```diff
@@ -0,0 +1,24 @@
+from django.utils.translation import ugettext_lazy as _
+from jsonfield.fields import JSONField
+from solo.models import SingletonModel
+
+
+class ElasticsearchBoostConfig(SingletonModel):
+    """
+    Model used to store default program configuration.
```
clintonb: This docstring is invalid.

course_discovery/apps/api/models.py
```diff
@@ -0,0 +1,24 @@
+from django.utils.translation import ugettext_lazy as _
+from jsonfield.fields import JSONField
+from solo.models import SingletonModel
+
+
+class ElasticsearchBoostConfig(SingletonModel):
```
clintonb: This model fits better in `edx_haystack_extensions`.

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -34,10 +36,21 @@ def build_search_kwargs(self, *args, **kwargs):
             'auto_generate_phrase_queries': True,
         }
 
-        if search_kwargs['query'].get('filtered', {}).get('query', {}).get('query_string'):
-            search_kwargs['query']['filtered']['query']['query_string'] = simple_query
+        # https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-function-score-query.html
+        function_score_config = ElasticsearchBoostConfig.get_solo().function_score
```
clintonb, line 40: What happens if this doesn't exist?

course_discovery/apps/api/models.py
clintonb: Additionally, you can create the first instance by adding a migration to `edx_catalog_extensions`.

clintonb:
Tests?

course_discovery/apps/edx_haystack_extensions/backends.py
bderusha: `get_solo` creates a default if one doesn't exist yet

course_discovery/apps/edx_haystack_extensions/models.py
```diff
@@ -0,0 +1,47 @@
...
+    # }
+
+    function_score = JSONField(
+        verbose_name=_('Function Score'),
+        help_text=_('JSON string containing an elasticsearch function score config.'),
+        null=False,
+        blank=False,
+        default={
+            "functions": [],
+            "boost": 5.0,
```
clintonb: Please add a comment explaining why we are boosting by default. Will this even have an effect if _everything_ is boosted?

course_discovery/apps/edx_haystack_extensions/models.py
```diff
@@ -0,0 +1,47 @@
...
+    #     "weight": 100.0,
+    # }
+
+    function_score = JSONField(
+        verbose_name=_('Function Score'),
+        help_text=_('JSON string containing an elasticsearch function score config.'),
+        null=False,
+        blank=False,
+        default={
+            "functions": [],
```
clintonb: Nit: use single quotes.

course_discovery/apps/edx_haystack_extensions/models.py
```diff
@@ -0,0 +1,47 @@
...
+from solo.models import SingletonModel
+
+
+class ElasticsearchBoostConfig(SingletonModel):
+    """
+    Model used to store the elasticsearch boost configuration.
+    This includes a default JSON config for the function_score.
+    """
+
+    # EXAMPLE FUNCTIONS
```
clintonb: This, along with an explanation of the effects, would be better housed in `docs/`.

clintonb:
@bderusha what I am seeing looks good. I would prefer to see tests before merging. Docs should also be updated. That's not a hard requirement; but, it would be ideal if that happened sooner (while the information is fresh on your mind).

bderusha:
@clintonb updated based on your comments.  Added tests, docs, and syntax tweaks

bderusha:
The only thing I'm not sure of is how to make the docs test that they render properly

clintonb:
👍

PR #266, opened 2016-08-22
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,7 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: ctlg
+oeps: {}
+owner: edx/ecommerce
+track-pulls: true


cpennington:
@edx/ecommerce: I'm marking this as yours.

openedx.yaml
```diff
@@ -0,0 +1,7 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: xxx
```
rlucioni: catalog

cpennington: See the field definition here: https://github.com/edx/open-edx-proposals/pull/18/files#diff-13c9b7b8133d4bcf6ab907c7f0065187R53

rlucioni:
I don't understand the need for the abbreviated nickname, but it seems reasonable. 👍

rlucioni:
Please squash your commits before merging.

openedx.yaml
```diff
@@ -0,0 +1,7 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: ctlg
```
clintonb, line 4: Why not catalog?

rlucioni: Similar question at https://github.com/edx/programs/pull/162#discussion_r76629920.

PR #268, opened 2016-08-22
 Post-merge fixes for Course model and data loader

clintonb:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml

### DIFF
--- a/course_discovery/apps/course_metadata/migrations/0022_remove_duplicate_courses.py
+++ b/course_discovery/apps/course_metadata/migrations/0022_remove_duplicate_courses.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def remove_duplicate_courses(apps, schema_editor):
+    Course = apps.get_model('course_metadata', 'Course')
+    duplicates = Course.objects.raw('SELECT'
+                                    '   c.* '
+                                    'FROM '
+                                    '  course_metadata_course AS c'
+                                    '  JOIN ('
+                                    '    SELECT'
+                                    '      LOWER(key) AS key,'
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/ecommerce

course_discovery/apps/course_metadata/migrations/0022_remove_duplicate_courses.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def remove_duplicate_courses(apps, schema_editor):
+    Course = apps.get_model('course_metadata', 'Course')
+    duplicates = Course.objects.raw('SELECT'
```
rlucioni, line 9: Why the raw SQL?

clintonb, line 9: `GROUP BY` is not well-support by Django's ORM.

rlucioni, line 9: Doesn't raw SQL seem like it should be a last resort? You could probably achieve the same effect using a combination of `values` and `annotate`.

clintonb, line 9: I've investigated. This is the best option.

rlucioni:
👍

cpennington:
@clintonb: Apologies. I had a rogue piece of automation that squashed your PR title and message.

clintonb:
No worries @cpennington

PR #269, opened 2016-08-22
Update Program Service to provide all metadata

mikedikan:
@edx/ecommerce for review

JIRA Subtask: https://openedx.atlassian.net/browse/ECOM-5096

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -12,7 +12,8 @@ from taggit_serializer.serializers import TagListSerializerField, TaggitSerializ
 from course_discovery.apps.api.fields import StdImageSerializerField, ImageField
 from course_discovery.apps.catalogs.models import Catalog
 from course_discovery.apps.course_metadata.models import (
-    Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType,
+    Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType, FAQ,
+    CorporateEndorsement, Endorsement
 )
 from course_discovery.apps.course_metadata.search_indexes import CourseIndex, CourseRunIndex, ProgramIndex
 
@@ -92,6 +93,14 @@ class NamedModelSerializer(serializers.ModelSerializer):
         fields = ('name',)
 
 
+class FAQSerializer(serializers.ModelSerializer):
+    """Serializer for the ``FAQ`` model."""
+
+    class Meta(object):
+        model = FAQ
+        fields = ('question', 'answer',)
+
+
 class SubjectSerializer(NamedModelSerializer):
     """Serializer for the ``Subject`` model."""
 
@@ -131,6 +140,33 @@ class VideoSerializer(MediaSerializer):
         fields = ('src', 'description', 'image',)
 
 
+class PersonSerializer(serializers.ModelSerializer):
+    """Serializer for the ``Person`` model."""
+
+    class Meta(object):
+        model = Person
+        fields = ('uuid', 'given_name', 'family_name', 'bio', 'profile_image_url', 'slug',)
+
+
+class EndorsementSerializer(serializers.ModelSerializer):
+    """Serializer for the ``Endorsement`` model."""
+    endorser = PersonSerializer()
+
+    class Meta(object):
+        model = Endorsement
+        fields = ('endorser', 'quote',)
+
+
+class CorporateEndorsementSerializer(serializers.ModelSerializer):
+    """Serializer for the ``CorporateEndorsement`` model."""
+    image = ImageSerializer()
+    individual_endorsements = EndorsementSerializer(many=True)
+
+    class Meta(object):
+        model = CorporateEndorsement
+        fields = ('corporation_name', 'statement', 'image', 'individual_endorsements',)
+
+
 class SeatSerializer(serializers.ModelSerializer):
     """Serializer for the ``Seat`` model."""
     type = serializers.ChoiceField(
@@ -148,14 +184,6 @@ class SeatSerializer(serializers.ModelSerializer):
     class Meta(object):
         model = Seat
         fields = ('type', 'price', 'currency', 'upgrade_deadline', 'credit_provider', 'credit_hours',)
-
-
-class PersonSerializer(serializers.ModelSerializer):
-    """Serializer for the ``Person`` model."""
-
-    class Meta(object):
-        model = Person
-        fields = ('uuid', 'given_name', 'family_name', 'bio', 'profile_image_url', 'slug',)
 
 
 class OrganizationSerializer(TaggitSerializer, serializers.ModelSerializer):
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -244,8 +278,30 @@ def test_data(self):
...
+            'min_hours_effort_per_week': None,
+            'overview': None,
+            'price_ranges': [],
+            'start': serialize_datetime_to_millis(program.start),
+            'status': program.status,
+            'subjects': [],
+            'transcript_languages': [],
         }
 
+        self.maxDiff = None
```
clintonb: Remove. This is only needed for debugging.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -278,15 +334,38 @@ def test_with_exclusions(self):
...
```
clintonb: Remove.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -58,11 +68,14 @@ def test_filter_by_uuids(self):
         """ Verify that the endpoint filters programs to those matching the provided UUIDs. """
         url = self.list_path + '?uuids='
 
-        programs = [ProgramFactory(), self.program]
```
clintonb, line 61: Why is this change necessary given that we are simply testing filtering?

clintonb, line 61: Actually, why is this file being modified at all?

clintonb:
👍 (once tests pass and comments addressed)

course_discovery/apps/api/v1/tests/test_views/test_programs.py
mikedikan: The tests in test_programs were failing because the additional metadata fields in the program serializer (specifically, the 'start' property addition) required access to the context to grab the user from the request within the CourseRunSerializer (specifically, the 'get_marketing_url' MethodField) which the 'start' property reads from.

clintonb: Got it. The additional context is fine. I don't see a need to add associated courses to these tests.

mikedikan:
@clintonb comments addressed - you were right about courses - I think that was a remnant from an earlier debug theory

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -39,6 +41,22 @@ def make_request():
     return request
 
 
+def serialize_datetime(d):
+    return d.strftime('%Y-%m-%dT%H:%M:%S') if d else None
+
+
+def serialize_datetime_to_millis(d):
```
schenedx: I am surprised we need the datetime serialized to milli-seconds. I think we should serialize the same datetime format for `program.start` vs `courserun.start`. The cause for the new function is that in the `Programs` serializer, the `start` is serialized using `serializers.DateTimeField()`, while in courseRunSerializer, the `start` is serialized by default.

clintonb: Nice catch. I agree with @schenedx. If you remove the `DateTimeField` for the start field in `ProgramSerializer` you won't need this function.

schenedx:
@mikedikan Just left 1 comment, and rest LGTM.
Another PR suggestion I want to make is, please attach the ticket number to the PR. I think the team like the ticket number in the description of the PR. I personally put ticket number in the branch name, PR title and sometimes, the description of the PR.

clintonb:
Having the ticket in the Git commit's description will ensure the commit is properly linked to the ticket in JIRA as well.

course_discovery/apps/api/serializers.py
```diff
@@ -131,6 +140,24 @@ class Meta(object):
         fields = ('src', 'description', 'image',)
 
 
+class EndorsementSerializer(serializers.ModelSerializer):
+    """Serializer for the ``Endorsement`` model."""
+
+    class Meta(object):
+        model = Endorsement
+        fields = ('endorser', 'quote',)
```
clintonb, line 157: endorser should use the `PersonSerializer`.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -1,5 +1,6 @@
 from django.core.urlresolvers import reverse
 from rest_framework.test import APITestCase
+from rest_framework.test import APIRequestFactory
```
clintonb: This should be combined with the import above.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -200,6 +244,7 @@ class Meta(object):
     card_image_url = FuzzyText(prefix='https://example.com/program/card')
     partner = factory.SubFactory(PartnerFactory)
     credit_redemption_overview = FuzzyText()
+    category = FuzzyText()
```
clintonb: Remove. This field is going away. Use the `type` field.

course_discovery/apps/api/serializers.py
```diff
@@ -331,9 +376,14 @@ def get_courses(self, program):
 
     class Meta:
         model = Program
-        fields = ('uuid', 'title', 'subtitle', 'type', 'marketing_slug', 'marketing_url', 'card_image_url',
-                  'banner_image', 'banner_image_url', 'authoring_organizations', 'credit_redemption_overview',
-                  'courses',)
+        fields = (
+            'uuid', 'title', 'subtitle', 'category', 'type', 'status', 'marketing_slug', 'marketing_url', 'courses',
```
clintonb: Remove the `category` field. It's going away.

course_discovery/apps/api/tests/test_serializers.py
mikedikan: When i attempted leaving it using default - it serialized to datetime(x, y, z) which did not match the output - default did not work here - i'm open to other suggestions, but didn't feel like burning up time in a snipe hunt.

clintonb: That's because it's a property rather than a model field. DRF cannot determine the type. This property actually doesn't need to be exposed. It is not publicly displayed anywhere. I don't recall why I added it.

mikedikan:
Latest comments addressed and added JIRA link - please review @clintonb

course_discovery/apps/api/serializers.py
```diff
@@ -317,6 +345,24 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+        many=True, read_only=True, slug_field='code',
+        help_text=_('Languages that course runs in this program are offered in.'),
+    )
+    transcript_languages = serializers.SlugRelatedField(
+        many=True, read_only=True, slug_field='code',
+        help_text=_('Languages that course runs in this program have available transcripts in.'),
+    )
+    subjects = SubjectSerializer(many=True)
+    staff = PersonSerializer(many=True)
+    start = serializers.DateTimeField()
```
clintonb: Remove this field. As mentioned in another comment, it does not need to be exposed.

clintonb:
👍 once start field removed from serializer and tests.

course_discovery/apps/api/serializers.py
mikedikan: Just saw your other comment - i'll remove the 'start' field.

mikedikan:
all outstanding comments addressed - waiting for tests to pass then merging. 🎉

PR #271, opened 2016-08-23
Adding constraint program status cannot be "Active" without a banner image.

awais786:
Add form level validation. Now without image status cannot be change to active.
Added test cases.

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -71,11 +71,15 @@ class ProgramAdmin(admin.ModelAdmin):
 
     search_fields = ('uuid', 'title', 'marketing_slug')
 
-    filter_horizontal = ('job_outlook_items', 'expected_learning_items', 'credit_backing_organizations',)
+    filter_horizontal = (
+        'job_outlook_items', 'expected_learning_items',
+        'credit_backing_organizations', 'authoring_organizations',
+    )
 
     # ordering the field display on admin page.
     fields = (
-        'title', 'status', 'type', 'banner_image_url', 'card_image_url', 'overview', 'video',
+        'title', 'status', 'type', 'banner_image',
+        'banner_image_url', 'card_image_url', 'overview', 'video',
     )
     fields += (
         'courses', 'course_runs', 'excluded_course_runs'
--- a/course_discovery/apps/course_metadata/forms.py
+++ b/course_discovery/apps/course_metadata/forms.py
@@ -1,5 +1,7 @@
 from django import forms
+from django.core.exceptions import ValidationError
 from django.forms.util import ErrorList
+from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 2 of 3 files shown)


awais786:
@bderusha just want to confirm that I have added the constrain in model clean method. But if restriction just needed for admin then I can move this condition into admin form.

awais786:
Alternate idea is following.
Add a form with adminview.

```
class ProgramAdminForm(forms.ModelForm):

    class Meta:
        model = Program
        fields = '__all__'

    def clean(self):
        cleaned_data = super(ProgramAdminForm, self).clean()
        status = cleaned_data.get("status")
        banner_image = cleaned_data.get("banner_image")

        if status == Program.ProgramStatus.Active and not banner_image:
            raise ValidationError(_('Status cannot be change to active without banner image.'))

```

clintonb:
It would be preferable to have this constraint on the model so that validation is run regardless of whether the model is modified by a form. This would be the `full_clean()` method needs to be called by the model's `save()` method. See https://docs.djangoproject.com/en/1.10/ref/models/instances/#validating-objects.

awais786:
@clintonb full_clean is raising error for category [Programs.category](https://github.com/edx/course-discovery/blob/69e1d444093dd5f876f119686bacb1caccf356c9/course_discovery/apps/course_metadata/models.py#L558)

Can i update the category as `blank=True` ?

In [api](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/data_loaders/api.py#L293) also.

clintonb:
That field should be removed.

course_discovery/apps/course_metadata/models.py
```diff
@@ -3,6 +3,7 @@
 import logging
 from urllib.parse import urljoin
 from uuid import uuid4
+from django.core.exceptions import ValidationError
```
bderusha: Please move above `django.db` import

awais786:
@bderusha, @clintonb  can we add a separate story for removing this field. This field is in use see [edx_catalog_extensions](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/edx_catalog_extensions/migrations/0002_convert_program_category_to_type.py#L12).

clintonb:
https://openedx.atlassian.net/browse/ECOM-5340

bderusha:
👍

awais786:
@bderusha Just want to confirm that is it possible that program appear as active and without image.
https://github.com/edx/course-discovery/blob/d24a13490659f5ab3dfd09be1ae6d551e3590422/course_discovery/apps/course_metadata/data_loaders/api.py#L289

waheedahmed:
👍

awais786:
@bderusha can you please review this PR.  I have created a separate ticket for model level validation.
[ECOM-5376](https://openedx.atlassian.net/browse/ECOM-5376)

bderusha:
👍

PR #272, opened 2016-08-23
Updated Program.start for search

clintonb:
- Fixed bug that occurs if no associated course runs have a start datetime
- Updated tests
- Exposed value in search index so that Programs are returned when faceting for availability

ECOM-5316


clintonb:
FYI @edx/ecommerce

PR #274, opened 2016-08-23
Adding data to program endpoint for program about page

mikedikan:
https://openedx.atlassian.net/browse/ECOM-5334

@clintonb @MatthewPiatetsky @edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -13,7 +13,7 @@ from course_discovery.apps.api.fields import StdImageSerializerField, ImageField
 from course_discovery.apps.catalogs.models import Catalog
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType, FAQ,
-    CorporateEndorsement, Endorsement
+    CorporateEndorsement, Endorsement, Position
 )
 from course_discovery.apps.course_metadata.search_indexes import CourseIndex, CourseRunIndex, ProgramIndex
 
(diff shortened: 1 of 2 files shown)


MatthewPiatetsky:
Need to test this, but from the code it looks like this has the fields I need 👍

course_discovery/apps/api/serializers.py
```diff
@@ -109,6 +109,9 @@ class SubjectSerializer(NamedModelSerializer):
 
     class Meta(NamedModelSerializer.Meta):
```
clintonb: There is no longer any value to using the `NamedModelSerializer` as a base class in this case. Update this to use `serializers.ModelSerializer` to avoid future confusion.

course_discovery/apps/api/serializers.py
```diff
@@ -195,7 +207,7 @@ class OrganizationSerializer(TaggitSerializer, serializers.ModelSerializer):
 
     class Meta(object):
         model = Organization
-        fields = ('key', 'name', 'description', 'homepage_url', 'tags',)
+        fields = ('key', 'name', 'description', 'homepage_url', 'tags', 'logo_image_url')
```
clintonb, line 208: Don't expose this yet. We will be changing how we store images soon.

course_discovery/apps/api/serializers.py
clintonb: Actually, don't modify this serializer, yet. We will soon be changing how we store images.

clintonb:
@mikedikan Ideally, we should be serving images from our own S3 bucket. Not all image fields have been updated. In order to move forward, you will need to replicate the following behaviors from the `Program.banner_image` field:
1. Create a new field with the appropriate type.
2. Update the serializer.
3. Update the data loader to download the image.

course_discovery/apps/api/serializers.py
clintonb: Only expose the field(s) you need now. It is far simpler to add these fields later than it is to remove them if we decide we don't need them (e.g. make a new API version).

mikedikan: Let’s discuss this and some of your other comments.

clintonb:
:+1: 

We will deal with image storage and field renaming in API v2.

clintonb:
https://openedx.atlassian.net/browse/ECOM-5341

mikedikan:
Cool, sounds good to me.

rlucioni:
👍

schenedx:
👍

PR #275, opened 2016-08-24
ECOM-5344

awais786:
Type is required field for the admin. Otherwise it throws errors while adding the new program using django-admin.

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -75,7 +75,7 @@ class ProgramAdmin(admin.ModelAdmin):
 
     # ordering the field display on admin page.
     fields = (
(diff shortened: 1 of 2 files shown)


awais786:
@bderusha please review this PR. I just found this by chance that while adding new program type needs to be given otherwise it throws error.

awais786:
Following is the error
`
Exception Value:    
The model '<Program: asasda>' does not have a model_attr 'seat_types'.
`

bderusha:
👍 code looks good

It's not clear to me what your last comment means.  What is the relation between the error you posted and the code implemented?

awais786:
@bderusha this is the reason https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L645

awais786:
@bderusha I have make the type as required in form.

PR #277, opened 2016-08-25
Add marketing url path

MatthewPiatetsky:
For rendering organization urls in the sidebar

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -205,7 +205,7 @@ class OrganizationSerializer(TaggitSerializer, serializers.ModelSerializer):
 
     class Meta(object):
         model = Organization
-        fields = ('key', 'name', 'description', 'homepage_url', 'tags', 'logo_image_url')
+        fields = ('key', 'name', 'description', 'homepage_url', 'tags', 'logo_image_url', 'marketing_url')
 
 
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -205,7 +205,7 @@ class OrganizationSerializer(TaggitSerializer, serializers.ModelSerializer):
 
     class Meta(object):
         model = Organization
-        fields = ('key', 'name', 'description', 'homepage_url', 'tags', 'logo_image_url')
+        fields = ('key', 'name', 'description', 'homepage_url', 'tags', 'logo_image_url', 'marketing_url_path')
```
clintonb: Return the complete URL rather than just the path. This can be determined by using the related partner. See how we do this for course runs.

course_discovery/apps/course_metadata/models.py
```diff
@@ -167,6 +167,12 @@ class Meta:
     def __str__(self):
         return '{key}: {name}'.format(key=self.key, name=self.name)
 
+    @property
+    def marketing_url(self):
```
clintonb, line 171: This needs a test.

clintonb:
👍  (once tests updated)

mikedikan:
👍  when tests pass

PR #279, opened 2016-08-26
Loading course run card image from marketing site

clintonb:
ECOM-5384

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -396,6 +396,7 @@ class CourseMarketingSiteDataLoader(AbstractMarketingSiteDataLoader):
             'uuid': uuid,
             'language': language,
             'slug': slug,
+            'card_image_url': self._get_nested_url(data.get('field_course_image_promoted')),
         }
 
         try:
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/ecommerce @mjfrey

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -396,6 +396,7 @@ def create_course_run(self, course, data):
             'uuid': uuid,
             'language': language,
             'slug': slug,
+            'card_image_url': self._get_nested_url(data.get('field_course_image_promoted')),
```
mjfrey, line 399: is this the same name it was before?  If not, this will still break the offer landing pages.

clintonb: I don't understand your question. There was nothing before.

mjfrey: I currently expect an attribute:
course_run['image']['src']
that comes back from the course_run API.  Just wanted to make sure that will still be there and populated.

clintonb: That's a different API. edx.org only uses dates from that API.

PR #281, opened 2016-08-26
CourseRun slug and filtering fixes

clintonb:


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -400,18 +400,12 @@ def create_course_run(self, course, data):
         }
 
         try:
-            course_run, created = CourseRun.objects.update_or_create(key__iexact=key, defaults=defaults)
+            course_run, __ = CourseRun.objects.update_or_create(key__iexact=key, defaults=defaults)
```
mikedikan, line 403: why is this changed to __ ?

clintonb: That is the Pythonic way of saying the value is not used.

mikedikan:
👍

bderusha:
belated 👍

PR #282, opened 2016-08-27
Resolved RemovedInDjango19Warning

clintonb:
This warning has been littering our logs for a while. Removing the noise.


clintonb:
FYI @bderusha @awais786

PR #284, opened 2016-08-29
Drupal data loader only loading published nodes

bderusha:
Please review: @clintonb 

https://www.drupal.org/node/4739


bderusha:
I don't think this is the right way to solve this.  Closing until I come up with a better solution

PR #287, opened 2016-08-30
Added active and marketable filters to CourseRun and Program models

clintonb:
API consumers can now request CourseRuns that are active/marketable and Programs that are marketable.

ECOM-5416

### DIFF
--- a/course_discovery/apps/api/v1/views.py
+++ b/course_discovery/apps/api/v1/views.py
@@ -157,9 +157,7 @@ class CatalogViewSet(viewsets.ModelViewSet):
         course_runs = []
 
         for course in courses:
-            active_course_runs = course.active_course_runs
-            for acr in active_course_runs:
-                course_runs.append(acr)
+            course_runs += list(course.course_runs.active().marketable())
 
         serializer = serializers.FlattenedCourseRunWithCourseSerializer(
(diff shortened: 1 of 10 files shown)


clintonb:
@bderusha @mjfrey please review

FYI @edx/ecommerce

course_discovery/apps/api/v1/views.py
```diff
@@ -274,6 +272,18 @@ def list(self, request, *args, **kwargs):
...
               paramType: query
               multiple: false
+            - name: active
+              description: Retrieve active course runs
+              required: false
+              type: integer
+              paramType: query
+              multiple: false
+            - name: marketable
+              description: Retrieve marketable course runs
```
bderusha: Can we include the definition of "marketable" here and below?

course_discovery/apps/api/v1/views.py
```diff
@@ -274,6 +272,18 @@ def list(self, request, *args, **kwargs):
               type: string
               paramType: query
               multiple: false
+            - name: active
+              description: Retrieve active course runs
```
bderusha: can we include our definition of "active" course run here?

bderusha:
One small thing otherwise 👍

schenedx:
👍

PR #288, opened 2016-08-31
Add User groups to top level objects

waheedahmed:
[ECOM-5389](https://openedx.atlassian.net/browse/ECOM-5389)

As a user I want all publisher objects to be owned by a group so that I know which ones to scope to which users.

**Acceptance Criteria:**
1. Validate top level publisher objects created by a user are by default assigned to the Group the user belongs to.

### DIFF
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -7,6 +7,7 @@ from django.dispatch import receiver
 from django.utils.translation import ugettext_lazy as _
 from django_extensions.db.models import TimeStampedModel
 from django_fsm import FSMField, transition
+from guardian.shortcuts import assign_perm
 from simple_history.models import HistoricalRecords
 from sortedm2m.fields import SortedManyToManyField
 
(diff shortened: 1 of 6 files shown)


course_discovery/apps/publisher/migrations/0005_auto_20160831_1124.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('publisher', '0004_auto_20160810_0854'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='course',
+            options={'get_latest_by': 'modified', 'permissions': (('view_course', 'Can view course'),), 'ordering': ('-modified', '-created')},
```
awais786: @waheedahmed whats the plans for all permissions ?

waheedahmed: @awais786 I have no idea. @bderusha can we get the list of all permission so we can add them in one migration?

waheedahmed:
@bderusha Please review this.

awais786:
I have added the comment for all permissions.  Apart from that PR looks fine. 👍

waheedahmed:
@bderusha Please take a look.

course_discovery/apps/publisher/models.py
```diff
@@ -110,6 +112,15 @@ def __str__(self):
     def post_back_url(self):
         return reverse('publisher:publisher_courses_edit', kwargs={'pk': self.id})
 
+    class Meta(TimeStampedModel.Meta):
+        permissions = (
+            ('view_course', 'Can view course'),
```
bderusha, line 117: @waheedahmed view is the only permission this needs to control

bderusha:
👍

PR #290, opened 2016-09-01
Updated search API endpoint to return only published items

clintonb:
ECOM-5412 and ECOM-5414

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -308,21 +308,23 @@ class Course(TimeStampedModel):
 
 class CourseRun(TimeStampedModel):
     """ CourseRun model. """
-    SELF_PACED = 'self_paced'
-    INSTRUCTOR_PACED = 'instructor_paced'
 
-    PACING_CHOICES = (
-        # Translators: Self-paced refers to course runs that operate on the student's schedule.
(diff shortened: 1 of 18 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/apps/course_metadata/models.py
```diff
@@ -565,7 +568,7 @@ def __str__(self):
 
 
 class Program(TimeStampedModel):
-    class ProgramStatus(DjangoChoices):
+    class Status(DjangoChoices):
```
schenedx, line 571: Don't you find this rename confusing? The `Program` model already have a field called `status`. I was confused by other area of the PR here because of this name change. What is the motivation behind this name change anyways?

schenedx:
Just one minor question, rest LGTM 👍

bderusha:
👍

course_discovery/apps/course_metadata/models.py
clintonb: I see your point regarding capitalization. `Program.Status` seems less redundant to me than `Program.ProgramStatus`.

schenedx: Got it. Either way is fine with me on the PR. Thanks for the explanation.

rlucioni:
👍

PR #292, opened 2016-09-01
Query Optimizations

clintonb:
ECOM-5437

### DIFF
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -348,6 +348,9 @@ SWAGGER_SETTINGS = {
     'permission_denied_handler': 'course_discovery.apps.api.views.api_docs_permission_denied_handler'
 }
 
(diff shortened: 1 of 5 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/settings/base.py
```diff
@@ -366,3 +366,6 @@
 COMMENTS_APP = 'course_discovery.apps.publisher_comments'
 
 TAGGIT_CASE_INSENSITIVE = True
+
+# django-solo configuration (https://github.com/lazybird/django-solo#settings)
+SOLO_CACHE = 'default'
```
rlucioni, line 374: django-solo cache timeout defaults to 5 minutes. Since we use it for configuration which doesn't change often, that TTL could be larger. Consider setting [SOLO_CACHE_TIMEOUT](https://github.com/lazybird/django-solo#cache-timeout) to 1 hour.

rlucioni:
👍

schenedx:
👍

PR #294, opened 2016-09-02
Added Top level Object Access.

waheedahmed:
[ECOM-4978](https://openedx.atlassian.net/browse/ECOM-4978)

As a user, I need access to all courses, course runs, people and organizations within my Group and should not be able to see objects owned by Groups that I'm not part of.

**Acceptance criteria:**
1. Validate staff user can have access to all Objects.
2. Validate that users have access to Objects within their parent Group.
3. Validate a user does NOT have access to objects outside of his/her Group.

### DIFF
--- a/course_discovery/apps/publisher/mixins.py
+++ b/course_discovery/apps/publisher/mixins.py
@@ -1,0 +1,34 @@
+from django.http import HttpResponseForbidden
+
+from course_discovery.apps.publisher.models import Course, Seat
+
+
+class ViewPermissionMixin(object):
+
+    def get_course(self):
+        publisher_object = self.get_object()
+        if isinstance(publisher_object, Course):
+            return publisher_object
+        if isinstance(publisher_object, Seat):
+            return publisher_object.course_run.course
+        if hasattr(publisher_object, 'course'):
+            return publisher_object.course
+
+        return None
+
+    def check_user(self, user):
+        course = self.get_course()
+        return check_view_permission(user, course)
+
+    def permission_failed(self):
+        return HttpResponseForbidden()
+
+    def dispatch(self, request, *args, **kwargs):
+        if not self.check_user(request.user):
+            return self.permission_failed()
+
+        return super(ViewPermissionMixin, self).dispatch(request, *args, **kwargs)
+
+
+def check_view_permission(user, course):
+    return user.is_staff or user.has_perm(Course.VIEW_PERMISSION, course)
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -1,13 +1,17 @@
 import ddt
+from mock import patch
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.urlresolvers import reverse
 from django.forms import model_to_dict
 from django.test import TestCase
+from guardian.shortcuts import assign_perm
 
 from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD
 from course_discovery.apps.publisher.models import Course, CourseRun, Seat, State
 from course_discovery.apps.publisher.tests import factories
+from course_discovery.apps.publisher.tests.utils import create_non_staff_user_and_login
+from course_discovery.apps.publisher.views import CourseRunDetailView
 from course_discovery.apps.publisher.wrappers import CourseRunWrapper
 from course_discovery.apps.publisher_comments.tests.factories import CommentFactory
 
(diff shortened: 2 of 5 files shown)


waheedahmed:
@awais786 @bderusha Please review.

course_discovery/apps/publisher/mixins.py
```diff
@@ -0,0 +1,32 @@
+from django.http import HttpResponseForbidden
+
+from course_discovery.apps.publisher.models import Course, CourseRun, Seat
+
+
+class ViewPermissionMixin(object):
+
+    def get_course(self):
+        course_object = self.get_object()
```
bderusha: NIT: rename this variable.  It looks like we are checking to see if a course is actually a seat.  This is confusing.  Please use something more generic

course_discovery/apps/publisher/mixins.py
```diff
@@ -0,0 +1,32 @@
...
+        if isinstance(course_object, CourseRun):
+            return course_object.course
+        if isinstance(course_object, Seat):
+            return course_object.course_run.course
+
+        return course_object
```
bderusha: I'd rather be more explicit here.  If I use this mixin with an organization endpoint `get_course` will return the organization.  Let's do some type checking for all expected returns and return null otherwise.

bderusha: A good catchall might be to try to call `.course` on whatever object is passed in if it isn't an expected type.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -46,8 +47,8 @@ def test_create_course(self):
         self.assertNotContains(response, 'Add new comment')
         self.assertNotContains(response, 'Total Comments')
 
-    def test_update_course(self):
-        """ Verify that we can update an existing course. """
+    def test_update_course_with_staff(self):
+        """ Verify that staff user can update an existing course. """
```
bderusha, line 54: Wondering about this change despite no changes to the test.  Is it true that only a staff user can hit this page?  Are we already accessing this page with a staff user?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -143,6 +219,81 @@ def test_update_course_run(self):
         self.assertContains(response, 'Add new comment')
         self.assertContains(response, comment.comment)
 
+    def test_edit_course_run_page_with_non_staff(self):
+        """ Verify that non staff user can't access course run edit page without permission. """
+        non_staff_user = UserFactory()
+        group = factories.GroupFactory()
+
+        self.client.logout()
+        self.client.login(username=non_staff_user.username, password=USER_PASSWORD)
```
bderusha: May be worth pulling this and the 5 lines above into its own method given the number of times it is used to setup for a non-staff user

course_discovery/apps/publisher/views.py
```diff
@@ -25,10 +27,11 @@ class CourseRunListView(ListView):
     def get_queryset(self):
         return [
             CourseRunWrapper(course_run) for course_run in CourseRun.objects.select_related('course').all()
+            if check_view_permission(self.request.user, course_run.course)
```
bderusha: is this a database query each time?  What is the performance of running this check against every course   run record in the db?  Can this be attached to the query at all?

bderusha:
A couple questions, but overall this looks awesome. Nice work!

course_discovery/apps/publisher/tests/test_views.py
waheedahmed: @bderusha In setup of this test case staff user is logged in, so staff user can access any course according to 1st AC of this ticket.

waheedahmed:
@bderusha does it looks good now?

course_discovery/apps/publisher/mixins.py
```diff
@@ -0,0 +1,34 @@
...
+    def get_course(self):
+        publisher_object = self.get_object()
+        if isinstance(publisher_object, Course):
+            return publisher_object
+        if isinstance(publisher_object, CourseRun):
+            return publisher_object.course
+        if isinstance(publisher_object, Seat):
+            return publisher_object.course_run.course
+
+        return None
```
bderusha, line 17: Please add a test for `None`

course_discovery/apps/publisher/mixins.py
```diff
@@ -0,0 +1,34 @@
...
+from course_discovery.apps.publisher.models import Course, CourseRun, Seat
+
+
+class ViewPermissionMixin(object):
+
+    def get_course(self):
+        publisher_object = self.get_object()
+        if isinstance(publisher_object, Course):
+            return publisher_object
+        if isinstance(publisher_object, CourseRun):
```
bderusha: Instead of checking for type here.  Let's make this the last check and change it to 

```
if hasattr(publisher_object, 'course'):
    return publisher_object.course
```

This is more robust and will do the right thing for future objects if that method is available without needing to add checks to this each time

course_discovery/apps/publisher/views.py
```diff
@@ -23,12 +26,16 @@ class CourseRunListView(ListView):
     template_name = 'publisher/course_runs_list.html'
 
     def get_queryset(self):
-        return [
-            CourseRunWrapper(course_run) for course_run in CourseRun.objects.select_related('course').all()
-        ]
+        if self.request.user.is_staff:
+            course_runs = CourseRun.objects.select_related('course').all()
+        else:
+            courses = get_objects_for_user(self.request.user, Course.VIEW_PERMISSION, Course)
```
bderusha, line 32: Does this return the actual objects or a lazy eval'd queryset?  If it returns something we can chain on it is probably worth doing so.  Otherwise what's here is fine.  Much improved!

course_discovery/apps/publisher/views.py
```diff
@@ -164,11 +174,16 @@ def post(self, request, course_run_id):
         state = request.POST.get('state')
         try:
             course_run = CourseRun.objects.get(id=course_run_id)
+
+            if not check_view_permission(request.user, course_run.course):
+                return HttpResponseForbidden()
+
             course_run.change_state(target=state)
+            # pylint: disable=no-member
             messages.success(
```
bderusha, line 183: Are these messages user-facing?

bderusha:
A few more minor comments.  👍 once addressed

course_discovery/apps/publisher/views.py
waheedahmed: @bderusha Yes.

course_discovery/apps/publisher/views.py
waheedahmed: @bderusha I think it return queryset of actual objects, see [guardian.shortcuts:get_objects_for_user](http://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#std:shortcut-get_objects_for_user) for more details.

awais786:
👍

PR #296, opened 2016-09-02
ECOM-5392 publish program data to marketing site to create or edit the program about page

schenedx:
@bderusha @rlucioni @clintonb 
Please review this PR
I believe this is everything needed to get the program publish to marketing site working. I tested this work on my local machine between Drupal and catalog.

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -1,10 +1,11 @@
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
 
 from simple_history.admin import SimpleHistoryAdmin
 from course_discovery.apps.course_metadata.forms import ProgramAdminForm
 from course_discovery.apps.course_metadata.models import *  # pylint: disable=wildcard-import
+from course_discovery.apps.course_metadata.publishers import ProgramPublisherException
 
 
 class SeatInline(admin.TabularInline):
@@ -84,6 +85,7 @@ class ProgramAdmin(admin.ModelAdmin):
         'credit_backing_organizations'
     )
     fields += filter_horizontal
+    save_error = None
 
     def custom_course_runs_display(self, obj):
         return ", ".join([str(run) for run in obj.course_runs])
@@ -91,13 +93,27 @@ class ProgramAdmin(admin.ModelAdmin):
     custom_course_runs_display.short_description = "Included course runs"
 
     def response_add(self, request, obj, post_url_continue=None):
-        return HttpResponseRedirect(reverse('admin_metadata:update_course_runs', kwargs={'pk': obj.pk}))
+        if self.save_error:
+            return self.response_post_save_add(request, obj)
+        else:
+            return HttpResponseRedirect(reverse('admin_metadata:update_course_runs', kwargs={'pk': obj.pk}))
 
     def response_change(self, request, obj):
-        if '_continue'in request.POST or '_save' in request.POST:
-            return HttpResponseRedirect(reverse('admin_metadata:update_course_runs', kwargs={'pk': obj.pk}))
+        if self.save_error:
+            return self.response_post_save_change(request, obj)
         else:
-            return HttpResponseRedirect(reverse('admin:course_metadata_program_add'))
+            if any(status in request.POST for status in ['_continue', '_save']):
+                return HttpResponseRedirect(reverse('admin_metadata:update_course_runs', kwargs={'pk': obj.pk}))
+            else:
+                return HttpResponseRedirect(reverse('admin:course_metadata_program_add'))
+
+    def save_model(self, request, obj, form, change):
+        try:
+            obj.save()
+            self.save_error = False
+        except ProgramPublisherException as ex:
+            messages.add_message(request, messages.ERROR, ex.message)
+            self.save_error = True
 
 
 @admin.register(ProgramType)
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -5,7 +5,7 @@ from urllib.parse import urljoin
 from uuid import uuid4
 
 import pytz
-from django.db import models
+from django.db import models, transaction
 from django.db.models.query_utils import Q
 from django.utils.translation import ugettext_lazy as _
 from django_extensions.db.fields import AutoSlugField
@@ -16,8 +16,10 @@ from simple_history.models import HistoricalRecords
 from sortedm2m.fields import SortedManyToManyField
 from stdimage.models import StdImageField
 from taggit.managers import TaggableManager
+import waffle
 
 from course_discovery.apps.core.models import Currency, Partner
+from course_discovery.apps.course_metadata.publishers import MarketingSitePublisher
 from course_discovery.apps.course_metadata.query import CourseQuerySet, CourseRunQuerySet, ProgramQuerySet
 from course_discovery.apps.course_metadata.utils import UploadToFieldNamePath
 from course_discovery.apps.course_metadata.utils import clean_query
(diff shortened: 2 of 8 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -16,12 +16,14 @@
...
 from stdimage.models import StdImageField
 from taggit.managers import TaggableManager
+import waffle
 
 from course_discovery.apps.core.models import Currency, Partner
 from course_discovery.apps.course_metadata.query import CourseQuerySet, CourseRunQuerySet, ProgramQuerySet
 from course_discovery.apps.course_metadata.utils import UploadToFieldNamePath
 from course_discovery.apps.course_metadata.utils import clean_query
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
+from course_discovery.apps.course_metadata.publishers import MarketingSitePublisher
```
clintonb: Alphabetize.

course_discovery/apps/course_metadata/models.py
```diff
@@ -701,6 +703,25 @@ def staff(self):
         staff = itertools.chain.from_iterable(staff)
         return set(staff)
 
+    @property
+    def is_active(self):
+        return self.status == self.Status.Active
+
+    def save(self, *args, **kwargs):
+        # Before save, get from database the existing data if exists
+        existing_program = None
```
clintonb: Only do this if the switch is active.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -0,0 +1,155 @@
...
+        }
+        response = session.post(login_url, data=login_data)
+        expected_url = '{root}/users/{username}'.format(root=self.api_url, username=self.username)
+        if not (response.status_code == 200 and response.url == expected_url):
+            raise Exception('Login failed!')
+        return session
+
+    @cached_property
+    def api_session(self):
+        self.init_session.headers.update(self.headers)
```
clintonb, line 48: This seems like it should be part of the initialization.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -0,0 +1,155 @@
...
+        return self.init_session
+
+    @cached_property
+    def csrf_token(self):
+        # Get a CSRF token
```
clintonb: This comment is redundant with the property name.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -0,0 +1,155 @@
...
+            'status': 1 if program.is_active else 0
+        }
+
+    def _get_node_id(self, api_client, uuid):
+        node_url = '{root}/node.json?field_uuid={uuid}'.format(root=api_client.api_url, uuid=uuid)
+        response = api_client.api_session.get(node_url)
+        if response.status_code == 200:
+            found = response.json()
+            if found.get('list'):
+                return found.get('list')[0]['nid']
```
clintonb: This will fail if `found.get('list')` returns `None`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -701,6 +703,25 @@ def staff(self):
...
+        if self.id:
+            existing_program = Program.objects.get(id=self.id)
+        # Pass existing data to the publisher so it can decide whether we should publish
+        publisher = MarketingSitePublisher(existing_program)
+
+        with transaction.atomic():
+            super(Program, self).save(*args, **kwargs)
+            # Once save complete, we need to update the marketing site
+            # So the marketing page for this program is automatically updated
+            if waffle.switch_is_active('publish_program_to_marketing_site'):
```
clintonb: Add `and self.partner.has_marketing_site`; otherwise, we'll end up doing a lot of work for nothing.

course_discovery/apps/course_metadata/publishers.py
schenedx: Unfortunately, it cannot because, in order to generate the header, we will need to make a call to the `session` property. If we update the header within the definition of the session property, it's literally stack overflow.

course_discovery/apps/course_metadata/publishers.py
schenedx: Why? if `found.get('list')` evaluates to `None`, the statement underneath the `if` won't get executed, right?

clintonb: Ah I missed that. Use a variable:

``` python
results = found.get('list')

if results:
  return results[0]['nid']
```

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -0,0 +1,156 @@
...
+    def _get_node_id(self, api_client, uuid):
+        node_url = '{root}/node.json?field_uuid={uuid}'.format(root=api_client.api_url, uuid=uuid)
+        response = api_client.api_session.get(node_url)
+        if response.status_code == 200:
+            found = response.json()
+            if found:
+                list_item = found.get('list')
+                if list_item:
+                    return list_item[0]['nid']
+        return None
```
rlucioni: You can exclude this. The method will implicitly return None if the status code is not 200.

course_discovery/apps/course_metadata/models.py
```diff
@@ -701,6 +703,28 @@ def staff(self):
...
+            if self.id:
+                existing_program = Program.objects.get(id=self.id)
+            # Pass existing data to the publisher so it can decide whether we should publish
+            publisher = MarketingSitePublisher(existing_program)
+
+            with transaction.atomic():
+                super(Program, self).save(*args, **kwargs)
+                # Once save complete, we need to update the marketing site
+                # So the marketing page for this program is automatically updated
+                publisher.publish_program(self)
```
rlucioni, line 724: If this fails, what kind of messaging gets bubbled up to the user? Is it made clear that nothing has been saved or published?

schenedx: Good feedback, let me try to implement this error message on django admin

schenedx:
@rlucioni I added the exception handle on django admin for program. Please review

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -4,6 +4,12 @@
 from django.utils.functional import cached_property
 
 
+class ProgramPublisherException(Exception):
+    def __init__(self, message):
+        post_fix = 'The program data has not been saved. Please check your marketing site configuration'
```
rlucioni: "Publication to the marketing site failed. To avoid inconsistencies, no data has been saved."

rlucioni: Nitpick: I think the variable name you were looking for is `suffix`. 😄

rlucioni: Ignore my message suggestion, I see what you're doing here now.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -84,20 +86,35 @@ class ProgramAdmin(admin.ModelAdmin):
...
+            return HttpResponseRedirect(reverse('admin_metadata:update_course_runs', kwargs={'pk': obj.pk}))
 
     def response_change(self, request, obj):
-        if '_continue'in request.POST or '_save' in request.POST:
-            return HttpResponseRedirect(reverse('admin_metadata:update_course_runs', kwargs={'pk': obj.pk}))
+        if self.save_error:
+            return self.response_post_save_change(request, obj)
         else:
-            return HttpResponseRedirect(reverse('admin:course_metadata_program_add'))
+            if '_continue'in request.POST or '_save' in request.POST:
```
rlucioni:
``` py
if any(status in request.POST for status in ['_continue', '_save'])
```

rlucioni:
👍 once commits squashed and tests pass.

schenedx:
@bderusha and @clintonb Do you guys want to comment on this PR? Or things looks good to you?

bderusha:
LGTM 👍

PR #297, opened 2016-09-02
Updated search endpoint to filter by default partner

clintonb:
If no partner is specified on the request, all results are filtered so that only items associated with the default partner are returned.

ECOM-5459

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -3,12 +3,13 @@ import json
 import urllib.parse
 
 import ddt
+from django.conf import settings
 from django.core.urlresolvers import reverse
 from haystack.query import SearchQuerySet
 from rest_framework.test import APITestCase
 
 from course_discovery.apps.api.serializers import CourseRunSearchSerializer, ProgramSearchSerializer
-from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD
+from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD, PartnerFactory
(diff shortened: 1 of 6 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -162,29 +171,53 @@ def assert_response_includes_availability_facets(self, response_data):
...
+        other_partner = PartnerFactory()
+        other_course_run = CourseRunFactory(course__partner=other_partner, status=CourseRun.Status.Published)
+        other_program = ProgramFactory(partner=other_partner, status=Program.Status.Active)
+        self.assertNotEqual(other_program.partner.short_code, self.partner.short_code)
+        self.assertNotEqual(other_course_run.course.partner.short_code, self.partner.short_code)
+
+        response = self.get_search_response()
+        self.assertEqual(response.status_code, 200)
+        response_data = json.loads(response.content.decode('utf-8'))
+        self.assertListEqual(response_data['objects']['results'],
```
schenedx, line 215: Is there a way to assert if `other_course_run` and `other_program` is not within the response_data? Like `self.assertFalse(response_data['objects']['results'].contains(other_program))`?

schenedx:
@clintonb One minor test comment. Rest LGTM 👍

course_discovery/apps/api/v1/tests/test_views/test_search.py
clintonb: The fact that this assertion passes asserts that the other data is not in the result set.

PR #301, opened 2016-09-06
ECOM-5402 Delete the drupal node when Course Catalog program data got hard deleted

schenedx:
@bderusha @rlucioni @clintonb Please review

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -5,7 +5,7 @@ from urllib.parse import urljoin
 from uuid import uuid4
 
 import pytz
-from django.db import models
+from django.db import models, transaction
 from django.db.models.query_utils import Q
 from django.utils.translation import ugettext_lazy as _
 from django_extensions.db.fields import AutoSlugField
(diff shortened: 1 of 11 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -725,6 +725,13 @@ def save(self, *args, **kwargs):
         else:
             super(Program, self).save(*args, **kwargs)
 
+    def delete(self, *args, **kwargs):
```
clintonb: See the note labled "Overridden model methods are not called on bulk operations" at https://docs.djangoproject.com/en/1.10/topics/db/models/#overriding-predefined-model-methods. This should be part of a `pre_delete` signal.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -89,6 +89,29 @@ def __init__(self, program_before=None):
...
+    def _get_api_client(self, program):
+        if not program.partner.has_marketing_site:
+            return
+
+        if not (program.partner.marketing_site_api_username and program.partner.marketing_site_api_password):
+            msg = 'Marketing Site API credentials are not properly configured for Partner [{partner}]!'.format(
+                partner=program.partner.short_code)
+            raise ProgramPublisherException(msg)
+
+        if self.data_before and \
```
rlucioni: Minor readability suggestion:

``` py
if self.data_before and all(self.data_before[key] == getattr(program, key) for key in ['title', 'status', 'type']):
```

Since you're building the dictionary used here, you don't need to be as defensive (i.e., no need to use `get()`).

schenedx:
@clintonb Took in your feedback. Review again?

schenedx:
Close this PR so this will be made against master branch

PR #305, opened 2016-09-07
ECOM-5402 Delete the drupal node when program is hard deleted

schenedx:
@rlucioni @clintonb @bderusha Please review.

This is the continuation of https://github.com/edx/course-discovery/pull/301

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -268,7 +268,7 @@ class AbstractValueModelTests(TestCase):
 
 
 @ddt.ddt
-class ProgramTests(MarketingSitePublisherTestMixin, TestCase):
+class ProgramTests(MarketingSitePublisherTestMixin):
     """Tests of the Program model."""
 
     def setUp(self):
@@ -395,7 +395,12 @@ class ProgramTests(MarketingSitePublisherTestMixin, TestCase):
     def test_save_without_publish(self):
         self.program.title = FuzzyText().fuzz()
         self.program.save()
-        self.assertEqual(len(responses.calls), 0)
+        self.assert_responses_call_count(0)
+
+    @responses.activate
+    def test_delete_without_publish(self):
+        self.program.delete()
+        self.assert_responses_call_count(0)
 
     @responses.activate
     def test_save_and_publish_success(self):
@@ -409,8 +414,7 @@ class ProgramTests(MarketingSitePublisherTestMixin, TestCase):
         toggle_switch('publish_program_to_marketing_site', True)
         self.program.title = FuzzyText().fuzz()
         self.program.save()
-        self.assertEqual(len(responses.calls), 6)
-        toggle_switch('publish_program_to_marketing_site', False)
+        self.assert_responses_call_count(6)
 
     @responses.activate
     def test_save_and_no_marketing_site(self):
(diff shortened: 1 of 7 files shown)


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -409,6 +414,29 @@ def test_save_and_no_marketing_site(self):
...
+        self.program.partner.marketing_site_api_username = self.username
+        self.program.partner.marketing_site_api_password = self.password
+        self.program.save()
+        self.mock_api_client(200)
+        self.mock_node_retrieval(self.program.uuid)
+        self.mock_node_delete(204)
+        toggle_switch('publish_program_to_marketing_site', True)
+        self.program.delete()
+        self.assertEqual(len(responses.calls), 5)
+        toggle_switch('publish_program_to_marketing_site', False)
```
rlucioni: Why toggle this back off? The switch is cleaned up between tests.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -385,6 +385,11 @@ def test_save_without_publish(self):
         self.assertEqual(len(responses.calls), 0)
 
     @responses.activate
+    def test_delete_without_publish(self):
+        self.program.delete()
```
rlucioni, line 402: I'd explicitly turn off the publish switch here (or otherwise verify that it's off) to make this test's purpose clear.

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -216,3 +216,30 @@ def test_publish_unmodified_program(self):
...
+
+    @responses.activate
+    def test_publish_program_no_credential(self):
+        self.program.partner.marketing_site_api_password = None
+        self.program.partner.marketing_site_api_username = None
+        self.program.save()  # pylint: disable=no-member
+        publisher = MarketingSitePublisher()
+        with self.assertRaises(ProgramPublisherException):
+            publisher.publish_program(self.program)
+            self.assertEqual(len(responses.calls), 0)
```
rlucioni: This is a really common pattern. Consider DRYing things out with a helper method.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -0,0 +1,26 @@
...
+    def test_delete_program_signal_no_publish(self, delete_program_mock):
+        self.program.delete()
+        self.assertFalse(delete_program_mock.called)
+
+    @patch('course_discovery.apps.course_metadata.publishers.MarketingSitePublisher.delete_program')
+    def test_delete_program_signal_with_publish(self, delete_program_mock):
+        toggle_switch('publish_program_to_marketing_site', True)
+        self.program.delete()
+        delete_program_mock.assert_called_once_with(self.program)
+        toggle_switch('publish_program_to_marketing_site', False)
```
rlucioni: I don't think you need to bother turning this off, it should be cleaned up after the test runs.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -0,0 +1,26 @@
...
+    def setUp(self):
+        super(SignalsTest, self).setUp()
+        self.program = factories.ProgramFactory()
+
+    @patch('course_discovery.apps.course_metadata.publishers.MarketingSitePublisher.delete_program')
+    def test_delete_program_signal_no_publish(self, delete_program_mock):
+        self.program.delete()
+        self.assertFalse(delete_program_mock.called)
+
+    @patch('course_discovery.apps.course_metadata.publishers.MarketingSitePublisher.delete_program')
```
rlucioni: Decorate the class with this once, instead of each method. For example, https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/catalog/tests/test_utils.py#L18.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -0,0 +1,26 @@
+from unittest.mock import patch
+
+from django.test import TestCase
+
+from course_discovery.apps.course_metadata.tests import factories, toggle_switch
+
+
+# pylint: disable=no-member
```
rlucioni: Nitpick: these whole-module disables tend to go on the first line of the module.

rlucioni:
@schenedx looks good, just left some comments on your tests. 👍

course_discovery/apps/course_metadata/tests/test_publishers.py
schenedx: I don't know why you believe this is a common pattern. I did a brief search within this file and found the only place I do the both lines in the `with` statement is this place.

rlucioni: @schenedx I was referring specifically to this line, where you assert the length of the calls matches an expected count. That would make a nice custom assertion you can include in a mixin. For example:

``` py
def assert_calls(self, count):
    self.assertEqual(len(responses.calls), count)
```

schenedx: Got it. Make sense. Will make the change

PR #306, opened 2016-09-08
ECOM-5026

awais786:
Adding Email functionality for course state changes.
Adding Email functionality for newly added comments for course, course run and seats.
Email on/off attribute added for users.
Emails will be send to group users who has access on that object.
Waffle switch is added to disable the email functionality on prod.

TODO. django-ses will be implemented in next story.


awais786:
@bderusha I have added the code. But many tests cases are failing due to permissions on objects and enable notifications. I am fixing them at the moment. But you can take a look on this PR.

course_discovery/apps/publisher/models.py
```diff
@@ -49,30 +52,35 @@ class State(TimeStampedModel, ChangedByMixin):
     def __str__(self):
         return self.get_name_display()
 
+    @property
+    def get_course_run(self):
+        course_run = self.publisher_course_runs_state.all()
+        return course_run[0] if course_run else None
```
waheedahmed: You can simply return this `return self.publisher_course_runs_state.first()`

course_discovery/apps/publisher/mixins.py
```diff
@@ -11,6 +11,8 @@ def get_course(self):
         publisher_object = self.get_object()
         if isinstance(publisher_object, Course):
             return publisher_object
+        if isinstance(publisher_object, CourseRun):
+            return publisher_object.course
```
waheedahmed: No need for this, last check is doing the same thing.

```
if hasattr(publisher_object, 'course'):
    return publisher_object.course
```

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -223,7 +225,6 @@ def test_update_course_run_with_staff(self):
         updated_lms_course_id = 'course-v1:testX+AS121+2018_q1'
         self.course_run_dict['lms_course_id'] = updated_lms_course_id
         self.assertNotEqual(self.course_run.lms_course_id, updated_lms_course_id)
-        self.assertNotEqual(self.course_run.changed_by, self.user)
```
waheedahmed, line 226: Why removing this?

awais786: Now they are equal because course is updated by the self.user. I have tested this with different user also in other test.

course_discovery/apps/publisher/mixins.py
```diff
@@ -62,3 +62,18 @@ def form_valid(self, form):
...
+    """ Helper method to check user in a group has permission on publisher object."""
+    users_list = []
+    user_groups = user.groups.all()
+    if not user_groups:
+        return users_list
+
+    for group in user_groups:
+        for user in group.user_set.all():
+            if check_view_permission(user, publisher_object):
+                users_list.append(user)
```
waheedahmed: You can simply use this django-gaurdian shortcut [get_objects_for_user](http://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html?highlight=get_objects_for_user#guardian.shortcuts.get_users_with_perms) to get all users which have access to this `publisher_object`.

waheedahmed:
Just one more comment, 👍 once addressed.

PR #307, opened 2016-09-08
Order courses within a serialized program

rlucioni:
Courses are sorted, ascending, by earliest run start date, with ties broken by earliest run enrollment date. ECOM-5485.

@edx/ecommerce please review. I anticipate a handful of test failures due to the new sorting logic clashing with the fuzzy dates created by factories. I'll address those tomorrow.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1,10 +1,12 @@
 # pylint: disable=abstract-method
+from datetime import datetime
 import json
 from urllib.parse import urlencode
 
 from django.contrib.auth import get_user_model
 from django.utils.translation import ugettext_lazy as _
 from drf_haystack.serializers import HaystackSerializer, HaystackFacetSerializer
+import pytz
 from rest_framework import serializers
 from rest_framework.fields import DictField
 from taggit_serializer.serializers import TagListSerializerField, TaggitSerializer
@@ -380,15 +382,55 @@ class ProgramSerializer(serializers.ModelSerializer):
     staff = PersonSerializer(many=True)
 
     def get_courses(self, program):
+        courses = self.sort_courses(program)
+
         course_serializer = ProgramCourseSerializer(
-            program.courses,
+            courses,
             many=True,
             context={
                 'request': self.context.get('request'),
                 'program': program
             }
         )
+
         return course_serializer.data
+
+    def sort_courses(self, program):
+        """
+        Sorting by enrollment start then by course start yields a list ordered by course start, with
+        ties broken by enrollment start. This works because Python sorting is stable: two objects with
+        equal keys appear in the same order in sorted output as they appear in the input.
+
+        Courses are only created if there's at least one course run belonging to that course, so
+        course_runs should never be empty. If it is, key functions in this method attempting to find the
+        min of an empty sequence will raise a ValueError.
+        """
+        def min_run_enrollment_start(course):
+            # Enrollment starts may be empty. When this is the case, we make the same assumption as
+            # the LMS: no enrollment_start is equivalent to (offset-aware) datetime.min.
+            min_datetime = datetime.min.replace(tzinfo=pytz.UTC)
+
+            # Course runs excluded from the program are excluded here, too.
+            #
+            # If this becomes a candidate for optimization in the future, be careful sorting null values
+            # in the database. PostgreSQL and MySQL sort null values as if they are higher than non-null
+            # values, while SQLite does the opposite.
+            #
+            # For more, refer to https://docs.djangoproject.com/en/1.10/ref/models/querysets/#latest.
+            run = min(program.course_runs.filter(course=course), key=lambda run: run.enrollment_start or min_datetime)
+
+            return run.enrollment_start or min_datetime
+
+        def min_run_start(course):
+            run = min(program.course_runs.filter(course=course), key=lambda run: run.start)
+
+            return run.start
+
+        courses = list(program.courses.all())
+        courses.sort(key=min_run_enrollment_start)
+        courses.sort(key=min_run_start)
+
+        return courses
 
     class Meta:
         model = Program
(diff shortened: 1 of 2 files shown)


schenedx:
@rlucioni Just a question: No enrollment date for a course run means the course run is open for enrollment now? Is this assumption widely understood? I just want to confirm.
The code LGTM.

rlucioni:
@schenedx yes: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/access.py#L346.

rlucioni:
Alright, this is ready. Test failures are resolved, and I headed off a subtle bug by ignoring excluded course runs when determining the ordering. Looking for one more review!

course_discovery/apps/api/serializers.py
```diff
@@ -381,16 +383,50 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+            run = min(program.course_runs.filter(course=course), key=lambda run: run.enrollment_start or min_datetime)
+
+            return run.enrollment_start or min_datetime
+
+        def min_run_start(course):
+            run = min(program.course_runs.filter(course=course), key=lambda run: run.start)
+
+            return run.start
+
+        courses = list(program.courses.all())
```
MatthewPiatetsky, line 429: I think you can do something like
program.courses.order_by('min_run_enrollment_start', 'min_run_start')

course_discovery/apps/api/serializers.py
```diff
@@ -381,16 +383,50 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+        equal keys appear in the same order in sorted output as they appear in the input.
+
+        Courses are only created if there's at least one course run belonging to that course, so
+        course_runs should never be empty. If it is, key functions in this method attempting to find the
+        min of an empty sequence will raise a ValueError.
+        """
+        def min_run_enrollment_start(course):
+            # Enrollment starts may be empty. When this is the case, we make the same assumption as
+            # the LMS: no enrollment_start is equivalent to (offset-aware) datetime.min.
+            min_datetime = datetime.min.replace(tzinfo=pytz.UTC)
```
MatthewPiatetsky, line 411: is this necessary? How would things be sorted if you just had None for enrollment_start?

course_discovery/apps/api/serializers.py
rlucioni: That doesn't quite work the way you'd think. Read [this](http://stackoverflow.com/a/2413908). I also don't think method names can be passed to `order_by` like that.

course_discovery/apps/api/serializers.py
rlucioni: Yes. They wouldn't be.

``` python
>>> from datetime import datetime
>>> min([datetime(2014, 1, 1), datetime(2015, 1, 1), None])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can't compare datetime.datetime to NoneType
```

course_discovery/apps/api/serializers.py
```diff
@@ -381,16 +383,50 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+
+            # Course runs excluded from the program are excluded here, too.
+            run = min(program.course_runs.filter(course=course), key=lambda run: run.enrollment_start or min_datetime)
```
bderusha, line 420: A more pure ORM/SQL solution using [annotations](https://docs.djangoproject.com/en/dev/topics/db/aggregation/#order-of-annotate-and-filter-clauses) could be:
  `program.course_runs.filter(course=course).annotate(null_enrollment_start=Count('enrollment_start')).order_by('-null_enrollment_start', 'enrollment_start').first()`

course_discovery/apps/api/serializers.py
MatthewPiatetsky: Seems it is lexicographic ordering (that is what I would expect), I'm now realizing you're sorting in the opposite order, wouldn't the equivalent be program.courses.order_by('min_run_start', 'min_run_enrollment_start', )

bderusha: Similarly to my above comment I wonder if this can be solved in SQL using [annotations](https://docs.djangoproject.com/en/dev/topics/db/aggregation/#order-of-annotate-and-filter-clauses) `program.courses.annotate(min_course_run=Min('enrollment_start')).order_by('-min_course_run')`

rlucioni: @MatthewPiatetsky I'm intentionally sorting by enrollment start, then by start. See the docstring on `sort_courses()`. It works because Python sorting is stable. Sorting the other way would yield a list of courses ordered by enrollment start, with ties broken by course start. That would be the opposite of what we want here.

@bderusha courses don't have an enrollment start, so we can't annotate them as you've proposed here. I think your suggestion below might work; I'll follow up inline there. Generally, I opted to go with the present implementation because of readability, though I'm open to any changes that will yield a performance improvement.

course_discovery/apps/api/serializers.py
rlucioni: @bderusha I did a little digging, and it looks like this can be solved more elegantly without any need for aggregation:

``` python
run = program.course_runs.filter(course=course).order_by('enrollment_start').first()
```

Records with an empty enrollment start will appear first.

rlucioni: Never mind! According to the [Django docs](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#latest), PostgreSQL and MySQL sort null values as if they are higher than non-null values, while SQLite does the opposite. @bderusha is that what your use of `Count()` is meant to resolve here?

rlucioni: @bderusha I played around with this locally and it doesn't appear to be what we want. The `null_enrollment_start` annotation is 0 for every record, which leaves the `order_by` vulnerable to the same database-specific ordering differences described above.

Even if we do get this SQL-based solution working, I'd prefer to stick with the pure-Python approach. Optimizing now is premature, and I don't want to sacrifice readability unnecessarily.

rlucioni: I will add a note about the ordering of null values across different databases.

MatthewPiatetsky:
is part of this ticket to verify that the courses are ordered correctly on the marketing site?

rlucioni:
@MatthewPiatetsky yes, see the ACs on the ticket.

MatthewPiatetsky:
👍

PR #308, opened 2016-09-09
ECOM-5392 Only publish to drupal for MicroMasters Program

schenedx:
@rlucioni @bderusha @clintonb Please review
This is the bug in the course catalog because we don't want the publish logic to be associated any non-micromaster programs.

### DIFF
--- a/course_discovery/apps/course_metadata/publishers.py
+++ b/course_discovery/apps/course_metadata/publishers.py
@@ -98,6 +98,10 @@ class MarketingSitePublisher(object):
                 partner=program.partner.short_code)
             raise ProgramPublisherException(msg)
 
+        if program.type.name != 'MicroMasters':
+            # Currently, we should not publish any programs that are not MicroMasters types to Marketing Site
+            return
+
         if self.data_before and \
                 all(self.data_before[key] == getattr(program, key) for key in ['title', 'status', 'type']):
             # We don't need to publish to marketing site because
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -17,7 +17,7 @@ from course_discovery.apps.core.utils import SearchQuerySetWrapper
 from course_discovery.apps.course_metadata.models import (
     AbstractMediaModel, AbstractNamedModel, AbstractValueModel,
     CorporateEndorsement, Program, Course, CourseRun, Endorsement,
-    FAQ, SeatType
+    FAQ, SeatType, ProgramType,
 )
 from course_discovery.apps.course_metadata.tests import factories, toggle_switch
 from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, ImageFactory
@@ -407,7 +407,7 @@ class ProgramTests(MarketingSitePublisherTestMixin):
         self.program.partner.marketing_site_url_root = self.api_root
         self.program.partner.marketing_site_api_username = self.username
         self.program.partner.marketing_site_api_password = self.password
-        self.program.save()
+        self.program.type = ProgramType.objects.get(name='MicroMasters')
         self.mock_api_client(200)
         self.mock_node_retrieval(self.program.uuid)
         self.mock_node_edit(200)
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/publishers.py
```diff
@@ -98,6 +98,10 @@ def _get_api_client(self, program):
                 partner=program.partner.short_code)
             raise ProgramPublisherException(msg)
 
+        if str(program.type) != 'MicroMasters':
```
clintonb: Why cast to a string? Use `program.type.name`.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -14,6 +15,7 @@ class SignalsTest(TestCase):
     def setUp(self):
         super(SignalsTest, self).setUp()
         self.program = factories.ProgramFactory()
+        self.program.type = ProgramType.objects.get(name='MicroMasters')
```
clintonb: Merge with the line above: `ProgramFactory(type=...)`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -417,6 +418,18 @@ def test_save_and_publish_success(self):
...
     @responses.activate
+    def test_xseries_program_save(self):
+        self.program.partner.marketing_site_url_root = self.api_root
+        self.program.partner.marketing_site_api_username = self.username
+        self.program.partner.marketing_site_api_password = self.password
+        self.program.type = self._get_program_type_by_string('XSeries')
+        self.program.save()
+        toggle_switch('publish_program_to_marketing_site', True)
+        self.program.title = FuzzyText().fuzz()
+        self.program.save()
```
clintonb, line 430: Move these two lines up, and call `save()` once.

clintonb, line 430: Actually, why does the title matter?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -417,6 +418,18 @@ def test_save_and_publish_success(self):
         self.assert_responses_call_count(6)
 
     @responses.activate
+    def test_xseries_program_save(self):
```
clintonb, line 420: Add a docstring explaining what's being verified.

course_discovery/apps/course_metadata/tests/mixins.py
```diff
@@ -134,3 +135,6 @@ def mock_node_delete(self, status):
             content_type='text/html',
             status=status
         )
+
+    def _get_program_type_by_string(self, type_string):
```
clintonb: This seems out-of-place, and somewhat unnecessary. Why not simply use `ProgramType.objects.get(name=type_string)` where necessary?

course_discovery/apps/course_metadata/tests/test_models.py
schenedx: Title matters because we don't publish to drupal unless `title` or `status` changes.
Will take your feedback to only save once

clintonb:
👍 (once comments addressed)

course_discovery/apps/course_metadata/publishers.py
bderusha: Or can we use `ProgramType.objects.get(name='MicroMasters')` to ensure that we are matching against the proper type.

I'd hate to silently miss this because we decide to rename this `micromaster` all lowercase in 6 months

course_discovery/apps/course_metadata/tests/mixins.py
bderusha: I also don't see the benefit here.  The way I could see this abstraction being useful is to mitigate against name changes needing to be applied across all tests.

I'd prefer to see this as a constant `MICROMASTERS_PROGRAM_TYPE = ProgramType.objects.get(name='MicroMasters')`

bderusha:
👍 once comments resolved

course_discovery/apps/course_metadata/publishers.py
schenedx: @bderusha 
1. I don't see why changing this to line to `ProgramType.objects.get(name='MicroMasters')` would allow us to avoid the problem when `MicroMasters` changes to 'micromasters`
2. I actually can't import `ProgramType` into this module, because this class is imported into the `course_discovery.apps.course_metadata.modules.py`. If I include `ProgramType` into this module, it's circular dependency.

course_discovery/apps/course_metadata/tests/mixins.py
schenedx: @bderusha In many places, I actually need to test with `XSeries` type as well. To keep the code simple, I'll stick with @clintonb's suggestion

PR #309, opened 2016-09-09
Removed the Program.category field

clintonb:


clintonb:
@edx/ecommerce please review

schenedx:
@clintonb Do we also need to remove `catalog` from https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L52? If not, how does that field gets populated in search results?

clintonb:
Updated. Thanks, @schenedx.

schenedx:
👍

rlucioni:
👍

PR #310, opened 2016-09-09
Expose uuids in catalog

MatthewPiatetsky:
For ECOM-5498, ECOM-5499, ECOM-5501

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -42,6 +42,9 @@ class BaseIndex(indexes.SearchIndex):
     def index_queryset(self, using=None):
         return self.model.objects.all()
 
+    def prepare_authoring_organization_uuids(self, obj):
+        return [str(organization.uuid) for organization in obj.authoring_organizations.all()]
+
 
 class BaseCourseIndex(OrganizationsMixin, BaseIndex):
     key = indexes.CharField(model_attr='key', stored=True)
@@ -114,6 +117,9 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     partner = indexes.CharField(null=True, faceted=True)
     program_types = indexes.MultiValueField()
     published = indexes.BooleanField(null=False, faceted=True)
+    authoring_organization_uuids = indexes.MultiValueField()
+    staff_uuids = indexes.MultiValueField()
+    subject_uuids = indexes.MultiValueField()
 
     def prepare_partner(self, obj):
         return obj.course.partner.short_code
@@ -151,6 +157,12 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     def prepare_program_types(self, obj):
         return obj.program_types
 
+    def prepare_staff_uuids(self, obj):
+        return [str(staff.uuid) for staff in obj.staff.all()]
+
+    def prepare_subject_uuids(self, obj):
+        return [str(subject.uuid) for subject in obj.subjects.all()]
+
 
 class ProgramIndex(BaseIndex, indexes.Indexable, OrganizationsMixin):
     model = Program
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -76,8 +76,10 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
 
     course_runs = indexes.MultiValueField()
     expected_learning_items = indexes.MultiValueField()
-
     prerequisites = indexes.MultiValueField(faceted=True)
+    authoring_organization_uuids = indexes.MultiValueField()
```
clintonb: Given that we don't actually use courses on the marketing site, this is a bit premature. Only course runs and programs matter.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -34,6 +34,9 @@ class CourseTests(TestCase):
     def setUp(self):
         super(CourseTests, self).setUp()
         self.course = factories.CourseFactory()
+        course_run = factories.CourseRunFactory()
```
clintonb: Pass the course to the factory.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -53,6 +56,13 @@ def test_course_run_update_caught_exception(self):
         with mock.patch.object(Course, 'reindex_course_runs', side_effect=Exception):
             self.course.save()
 
+    def test_staff(self):
+        """ Test that the staff property retrieves course staff correctly from course runs """
+        staff = factories.PersonFactory.create_batch(2)
+        self.course.course_runs.all()[0].staff.add(staff[0])
+        self.course.course_runs.all()[0].staff.add(staff[1])
+        self.assertEqual(self.course.staff, set(staff))
+
```
schenedx: please add a new test case for staff coming from different course runs

schenedx:
@MatthewPiatetsky just one comment, rest LGTM 👍

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -163,6 +173,16 @@ def test_program_types(self):
         other_program = factories.ProgramFactory(courses=courses)
         self.assertCountEqual(self.course_run.program_types, [program.type.name, other_program.type.name])
 
+    def test_program_types_unique(self):
```
clintonb: Why is this here?

MatthewPiatetsky: While I was working on this I noticed that if I had more than one program of the same type, I would have "program_types": ["XSeries", "XSeries"], which seems unhelpful/confusing

clintonb: Please move this to a separate commit to avoid confusion and maintain a separation of concerns.

MatthewPiatetsky: Does that mean it needs a ticket?

course_discovery/apps/course_metadata/search_indexes.py
MatthewPiatetsky: I see, I get it now, but was confused initially because all the stories refer to courses and programs not course runs and programs

course_discovery/apps/course_metadata/tests/test_models.py
clintonb: Nope. Just a commit.

MatthewPiatetsky:
@clintonb are your comments addressed?

clintonb:
DRY - clean up the duplicated methods. After that, 👍

MatthewPiatetsky:
Ok I put prepare_authoring_organization_uuids in a base class, the other methods were different, so I left them as is.

PR #311, opened 2016-09-12
Use threads to overlap I/O during data loading.

rlucioni:
ECOM-5520. @edx/ecommerce please review.

Using my default worker thread count (7), this cuts the time spent waiting on network I/O when loading course data from Drupal from 4 minutes to 50 seconds, an 80% improvement. When loading course data from the LMS, wait time is reduced from 1 minute 15 seconds to 20 seconds, a 73% improvement.

I've tuned the default number of worker threads so that we get significant time savings without taking down the stage marketing site. We can tune it further when running in production, hence the new management command argument.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -1,10 +1,12 @@
-import logging
+import concurrent.futures
 from decimal import Decimal
 from io import BytesIO
+import logging
+import math
 
-import requests
 from django.core.files import File
 from opaque_keys.edx.keys import CourseKey
+import requests
 
 from course_discovery.apps.core.models import Currency
 from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoader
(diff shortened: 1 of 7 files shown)


clintonb:
A refresh of production data on my local machine takes about 18.5 minutes with the new code. The old code took about 36 minutes.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -60,40 +62,46 @@ class CoursesApiDataLoader(AbstractDataLoader):
     """ Loads course runs from the Courses API. """
 
     def ingest(self):
-        api_url = self.partner.courses_api_url
-        count = None
-        page = 1
+        logger.info('Refreshing Courses and CourseRuns from %s...', self.partner.courses_api_url)
 
-        logger.info('Refreshing Courses and CourseRuns from %s...', api_url)
+        response = self._request(1)
```
schenedx: Let's define what this `1` means. Maybe call it `self.initial_page_number` as a class variable? I noticed in the `MarketingSiteLoader`, this variable has value `0`

schenedx:
@rlucioni Just one minor comment from me. Rest LGTM. This is awesome work and thank you for taking the initiative to get this done

course_discovery/apps/course_metadata/data_loaders/api.py
rlucioni: Updated.

schenedx:
👍

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -60,40 +62,47 @@ class CoursesApiDataLoader(AbstractDataLoader):
...
-            results = response['results']
-            logger.info('Retrieved %d course runs...', len(results))
+        with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:
+            futures = [executor.submit(self._request, page) for page in range(initial_page + 1, pages + 1)]
 
-            if response['pagination']['next']:
-                page += 1
-            else:
-                page = None
+        for future in futures:  # pragma: no cover
```
clintonb, line 76: Why is this exempted from coverage? It should be covered by tests.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -60,40 +62,47 @@ class CoursesApiDataLoader(AbstractDataLoader):
...
-                    msg = 'An error occurred while updating {course_run} from {api_url}'.format(
-                        course_run=course_run_id,
-                        api_url=api_url
-                    )
-                    logger.exception(msg)
+        self.delete_orphans()
 
-        logger.info('Retrieved %d course runs from %s.', count, api_url)
+    def _request(self, page):
+        """Make a request."""
```
clintonb, line 85: Either be more verbose, or remove this.

clintonb, line 85: Alternatively, you could make the method name more explicit about the behavior of this method.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -60,40 +62,47 @@ class CoursesApiDataLoader(AbstractDataLoader):
...
+        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE)
 
-        self.delete_orphans()
+    def _process_response(self, response):
+        """Process a response."""
```
clintonb, line 89: Same comment as above.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -170,31 +179,37 @@ class EcommerceApiDataLoader(AbstractDataLoader):
...
+            response = future.result()
+            self._process_response(response)
 
-        logger.info('Retrieved %d course seats from %s.', count, api_url)
+        logger.info('Retrieved %d course seats from %s.', count, self.partner.ecommerce_api_url)
 
         self.delete_orphans()
 
+    def _request(self, page):
+        """Make a request."""
```
clintonb, line 201: Same comment as above.

rlucioni:
@clintonb seems I merged prematurely. I'll open a second PR to address your comments.

PR #312, opened 2016-09-12
Remove unnecessary comments and make method names more verbose

rlucioni:
Addresses comments left over from https://github.com/edx/course-discovery/pull/311. @clintonb FYI.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -65,13 +65,13 @@ class CoursesApiDataLoader(AbstractDataLoader):
         logger.info('Refreshing Courses and CourseRuns from %s...', self.partner.courses_api_url)
 
         initial_page = 1
-        response = self._request(initial_page)
+        response = self._make_request(initial_page)
         count = response['pagination']['count']
         pages = response['pagination']['num_pages']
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -65,13 +65,13 @@ def ingest(self):
...
+        response = self._make_request(initial_page)
         count = response['pagination']['count']
         pages = response['pagination']['num_pages']
         self._process_response(response)
 
         with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:
-            futures = [executor.submit(self._request, page) for page in range(initial_page + 1, pages + 1)]
+            futures = [executor.submit(self._make_request, page) for page in range(initial_page + 1, pages + 1)]
 
         for future in futures:  # pragma: no cover
```
rlucioni, line 76: @clintonb this is being executed by test code, but coverage doesn't seem to realize that. The same code was excluded from coverage [earlier](https://github.com/edx/course-discovery/pull/311/files#diff-9d98cbe1dd216e0e1b6373fdb3fdbeebL64), probably for the same reason. Without this, patch coverage fails to meet the 95% mark.

clintonb:
👍

PR #314, opened 2016-09-12
Add uuids to serializers

MatthewPiatetsky:
@edx/ecommerce 

Adding uuids to serializers, so querystring will be able to filter on them

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -57,6 +57,10 @@ def serialize_language_to_code(language):
     return language.code
 
 
+def get_uuids(items):
+    return [str(item.uuid) for item in items]
+
+
 class CatalogSerializerTests(TestCase):
     def test_data(self):
         user = UserFactory()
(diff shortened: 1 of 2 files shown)


rlucioni:
@MatthewPiatetsky as we discussed earlier today, does this data actually need to be present in the response for filtering to work?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -714,6 +714,10 @@ def test_data(self):
             'published': course_run.status == CourseRun.Status.Published,
             'partner': course_run.course.partner.short_code,
             'program_types': course_run.program_types,
+            'authoring_organization_uuids': [str(organization.uuid)
```
clintonb: Minor: make a helper function to avoid the duplication:

``` python
def get_uuids(items):
  return [str(item.uuid) for item in items]
```

clintonb:
@rlucioni it does. I missed an important part of the docs.

👍  (once comment addressed)

MatthewPiatetsky:
bump

rlucioni:
👍

PR #316, opened 2016-09-12
Query Performance Improvements

clintonb:
ECOM-5440

Metrics below are on my edX MBP with the service running on Docker. Results with SQLite are similar.


clintonb:
`CourseRunViewSet` performance numbers (for list view)

|  | # Queries | Query Time (ms) | Response Time (ms) |
| --- | --- | --- | --- |
| Before | 234 | 132 | 2840 |
| After | 16 | 16 | 260 |

clintonb:
`CourseViewSet` performance numbers (for list view)

|  | # Queries | Query Time (ms) | Response Time (ms) |
| --- | --- | --- | --- |
| Before | 578 | 341 | 6950 |
| After | 21 | 31 | 400 |

clintonb:
`CatalogViewSet` performance numbers (for courses view)

|  | # Queries | Query Time (ms) | Response Time (ms) |
| --- | --- | --- | --- |
| Before | 386 | 255 | 6310 |
| After | 231 | 165 | 4730 |

clintonb:
@edx/ecommerce please review

course_discovery/settings/base.py
```diff
@@ -373,3 +373,34 @@
 # django-solo configuration (https://github.com/lazybird/django-solo#settings)
 SOLO_CACHE = 'default'
 SOLO_CACHE_TIMEOUT = 3600
+
+
+# Django Debug Toolbar settings
```
schenedx, line 378: What are these debug tools used for?

schenedx, line 378: Now I see, this is part of a move.

course_discovery/apps/course_metadata/models.py
```diff
@@ -346,9 +346,9 @@ class Pacing(DjangoChoices):
         help_text=_(
             "Title specific for this run of a course. Leave this value blank to default to the parent course's title."))
     start = models.DateTimeField(null=True, blank=True)
-    end = models.DateTimeField(null=True, blank=True)
+    end = models.DateTimeField(null=True, blank=True, db_index=True)
```
schenedx, line 349: Why do we need the `db_index` on this field? How is this being used?

clintonb: We do filtering on `end` and `enrollment_end`: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/query.py#L33-L39.

schenedx: Got it. Make sense

schenedx:
👍

course_discovery/apps/api/v1/views.py
```diff
@@ -36,6 +36,39 @@
...
+    Args:
+        queryset (QuerySet): original query
+
+    Returns:
+        QuerySet
+    """
+    # Prefetch the data for the related course runs
+    course_run_prefetch_fields = PREFETCH_FIELDS['course_run']
+    course_run_prefetch_fields = ['course_runs__' + field for field in course_run_prefetch_fields]
+    queryset = queryset.prefetch_related(*course_run_prefetch_fields)
```
mikedikan, line 72: I read a little bit about prefetch and it mentioned that it does the "join" in python by fetching the joined data in advance.  Is there no way to make Django do a DB-join?  Seems like it would be more efficient...

clintonb: No...

cpennington: `select_related` does a join in the database... And you're using it directly below. So, why `prefetch_related` rather than `select_related`, for these fields?

clintonb: `select_related` only works on foreign keys, not many-to-many fields.

cpennington: Also, good practice for these kinds of optimizations is to add some tests that use `assertNumQueries` to characterize how many queries have to happen when the function is called as the number of related objects changes.

clintonb: - https://docs.djangoproject.com/en/1.10/ref/models/querysets/#select-related
- https://docs.djangoproject.com/en/1.10/ref/models/querysets/#prefetch-related

cpennington: Ah, gotcha. I hadn't noticed that it was many-to-many. Worth a comment?

cpennington: (In particular, the comment at the top of the function has a lot of "what" in it, but not a lot of "why")

clintonb: Docstring updated.

rlucioni:
This is awesome work @clintonb, thanks. 👍

PR #317, opened 2016-09-12
Marketable programs must be active

rlucioni:
Programs have a status field that must be respected when filtering marketable programs. The LMS consumes the set of marketable MicroMasters. We don't want to leak unpublished programs.

@edx/ecommerce please review.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -1,12 +1,15 @@
+import ddt
 from django.core.urlresolvers import reverse
 from rest_framework.test import APITestCase, APIRequestFactory
 
 from course_discovery.apps.api.serializers import ProgramSerializer
 from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
+from course_discovery.apps.course_metadata.choices import ProgramStatus
 from course_discovery.apps.course_metadata.models import Program
 from course_discovery.apps.course_metadata.tests.factories import ProgramFactory
 
 
+@ddt.ddt
 class ProgramViewSetTests(APITestCase):
     list_path = reverse('api:v1:program-list')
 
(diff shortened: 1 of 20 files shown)


schenedx:
@rlucioni With your change here, which api endpoints will only return marketable programs?
Does this change affect the Django admin view as well? My concern is in the django admin view, user won't be able to edit unmarketable programs with this change

rlucioni:
@schenedx the admin will not be affected. The admin doesn't use the REST API, which is where this filtering can be applied.

schenedx:
👍

course_discovery/apps/course_metadata/query.py
```diff
@@ -4,6 +4,11 @@
 from django.db import models
 from django.db.models.query_utils import Q
 
+# It's not clear why, but importing the Program model as
+# from course_discovery.apps.course_metadata.models import Program
+# causes a ImportError when testing this code with nose.
+from course_discovery.apps.course_metadata import models as catalog_models
```
clintonb: Circular import. The models import this file.

clintonb: @schenedx and I deferred moving the status classes out of models. It may be time to do so.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -72,11 +72,17 @@ def test_marketable_exclusions(self, slug):
...
-        self.assertEqual(list(Program.objects.marketable()), [program])
+    @ddt.data(
+        (Program.Status.Unpublished, False),
+        (Program.Status.Active, True),
+    )
+    @ddt.unpack
+    def test_marketable(self, status, is_marketable):
+        """ Verify the method filters Programs to those which are active and have marketing slugs. """
+        program = ProgramFactory(status=status)
+        self.assertEqual(list(Program.objects.marketable()), [program] if is_marketable else [])
```
clintonb: This would be a bit easier to read if you added a variable: `expected = [program] if is_marketable else []`.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -83,11 +85,16 @@ def test_filter_by_uuids(self):
...
         """ Verify the endpoint filters programs to those that are marketable. """
         url = self.list_path + '?marketable=1'
         ProgramFactory(marketing_slug='')
-        expected = ProgramFactory.create_batch(3)
-        expected.reverse()
-        self.assertEqual(list(Program.objects.marketable()), expected)
-        self.assert_list_results(url, expected)
+        programs = ProgramFactory.create_batch(3, status=status)
+        programs.reverse()
+        self.assertEqual(list(Program.objects.marketable()), programs if is_marketable else [])
```
clintonb: Move `programs if is_marketable else []` to a variable.

clintonb:
👍 (once comments addressed)

course_discovery/apps/course_metadata/query.py
rlucioni: Oh, of course. Thanks for catching that. I'll take a look at moving the status classes to their own module.

course_discovery/apps/course_metadata/tests/test_query.py
rlucioni: Updated.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
rlucioni: Updated.

course_discovery/apps/course_metadata/query.py
rlucioni: Moved DjangoChoices classes to a dedicated module, `choices.py`.

PR #321, opened 2016-09-13
Optimized performance of the Catalog CSV endpoint

clintonb:
ECOM-5534


clintonb:
@edx/ecommerce please review

clintonb:
The database query count doesn't change much (24 to 21); but, the response time drops from timing out after 60s locally to returning in 4s. In production, this endpoint currently returns in about 23s.

rlucioni:
@clintonb nice, 👍 .

schenedx:
👍

PR #322, opened 2016-09-13
Optimized performance of the Affiliate Window endpoint

clintonb:
These changes cut response time in half locally.

ECOM-5534


clintonb:
@edx/ecommerce please review

rlucioni:
👍

PR #325, opened 2016-09-14
Added Docker volumes for media and static files directories

clintonb:


clintonb:
FYI @edx/ecommerce this should resolve folks' issues with static files.

rlucioni:
👍

PR #326, opened 2016-09-14
Fixed data loader bug

clintonb:
ECOM-5577


clintonb:
@rlucioni please review
FYI @edx/ecommerce

rlucioni:
👍

PR #328, opened 2016-09-15
Storing ingested data as HTML instead of Markdown

clintonb:
This removes the burden of conversion from our API clients.

ECOM-5623

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/__init__.py
+++ b/course_discovery/apps/course_metadata/data_loaders/__init__.py
@@ -1,6 +1,8 @@
 import abc
+import re
 
 import html2text
+import markdown
 from dateutil.parser import parse
 from django.utils.functional import cached_property
 from edx_rest_api_client.client import EdxRestApiClient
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/ecommerce please review

rlucioni:
@clintonb do we already have test coverage for this? If not, could you please add a test for this?

course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -82,12 +85,19 @@ def clean_strings(cls, data):
...
+        attributes), and converts the Markdown back to HTML.
+        """
+        cleaned = content.replace('&nbsp;', '')
         html_converter = html2text.HTML2Text()
         html_converter.wrap_links = False
         html_converter.body_width = None
-        return html_converter.handle(stripped).strip()
+        cleaned = html_converter.handle(cleaned).strip()
+        cleaned = markdown.markdown(cleaned)
+        cleaned = cls.MARKDOWN_CLEANUP_REGEX.sub(r'\1', cleaned)
```
rlucioni, line 99: This works. However, in the interest of less surprising, more readable code, what about doing this without the regex? You can define a small helper method:

``` py
def strip_p_tags(html):
    return html.lstrip('<p>').rstrip('</p>')
```

clintonb: The strip methods on strings will treats the arguments as a list of characters to remove, not a word. In some cases, like `<em>TOEFL</em>`, the opening angle bracket is removed.

schenedx:
👍

mikedikan:
👍

PR #330, opened 2016-09-15
ECOM-5660 Add logo image url override for organizations

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -50,11 +50,15 @@ class OrganizationsApiDataLoader(AbstractDataLoader):
         key = body['short_name']
         defaults = {
             'key': key,
-            'name': body['name'],
-            'description': body['description'],
-            'logo_image_url': body['logo'],
-            'partner': self.partner,
+            'partner': self.partner
         }
+        if not self.partner.has_marketing_site:
+            defaults.update({
+                'name': body['name'],
+                'description': body['description'],
+                'logo_image_url': body['logo']
+            })
+
         Organization.objects.update_or_create(key__iexact=key, defaults=defaults)
         logger.info('Processed organization "%s"', key)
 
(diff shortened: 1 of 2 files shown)


clintonb:
Why is this necessary? Organization data currently comes from Drupal.

course_discovery/apps/course_metadata/models.py
```diff
@@ -159,6 +159,14 @@ class Organization(TimeStampedModel):
     description = models.TextField(null=True, blank=True)
     homepage_url = models.URLField(max_length=255, null=True, blank=True)
     logo_image_url = models.URLField(null=True, blank=True)
+    logo_image_url_override = StdImageField(
+        upload_to=UploadToFieldNamePath(populate_from='uuid', path='media/organizations/logo_images'),
+        blank=True,
+        null=True,
+        variations={
+            'small': (200, 101),
```
schenedx: Are you sure this is the only variable we need? What about also store the original?

schenedx (commented):
Is there a plan to also update the dataloader so we store the original logo image for the organization from the source?

clintonb (changes requested):
The issue here is that we are pulling organization data from Drupal and LMS (Organizations API).

LMS: https://courses.edx.org/api/organizations/v0/organizations/MITx/
Drupal: https://prod-edx-mktg-edit.edx.org/sites/default/files/school/image/banner/mit_logo_200x101_0.png

The change you want is to the data loaders, and determining precedence of the data. Adding an override is not correct.

MatthewPiatetsky:
@edx/ecommerce I've updated the PR per @clintonb's comment. 
New github is confusing because you still see old comments like Simon's comment even though they are not relevant anymore.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -50,11 +50,13 @@ def update_organization(self, body):
...
         defaults = {
             'key': key,
-            'name': body['name'],
-            'description': body['description'],
-            'logo_image_url': body['logo'],
-            'partner': self.partner,
+            'partner': self.partner
         }
+        if not self.partner.has_marketing_site:
+            defaults['name'] = body['name']
```
clintonb:
``` python
defaults.update({
  'name': body['name'],
  ...
})
```

clintonb (approved):

schenedx (approved):
👍 
I agree. My comment clearly is a relic in the past!

PR #331, opened 2016-09-15
ECOM-5635 allow filter of unpublished course_runs

schenedx:
@edx/ecommerce Please review

With this change, if the course_run is unpublished, we can filter it out in the program details API endpoint

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -13,6 +13,7 @@ from taggit_serializer.serializers import TagListSerializerField, TaggitSerializ
 
 from course_discovery.apps.api.fields import StdImageSerializerField, ImageField
 from course_discovery.apps.catalogs.models import Catalog
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType, FAQ,
     CorporateEndorsement, Endorsement, Position
@@ -353,6 +354,8 @@ class ProgramCourseSerializer(CourseSerializer):
     def get_course_runs(self, course):
         program = self.context['program']
         course_runs = program.course_runs.filter(course=course)
+        if self.context.get('published_course_runs_only'):
+            course_runs = course_runs.filter(status=CourseRunStatus.Published)
         return CourseRunSerializer(
             course_runs,
             many=True,
@@ -391,7 +394,8 @@ class ProgramSerializer(serializers.ModelSerializer):
             many=True,
             context={
                 'request': self.context.get('request'),
-                'program': program
+                'program': program,
+                'published_course_runs_only': self.context.get('published_course_runs_only'),
             }
         )
 
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -176,6 +176,7 @@ class CourseRunSerializerTests(TestCase):
         self.assertDictEqual(serializer.data, expected)
 
 
+@ddt.ddt
 class ProgramCourseSerializerTests(TestCase):
     def setUp(self):
         super(ProgramCourseSerializerTests, self).setUp()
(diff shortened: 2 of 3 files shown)


schenedx:
The place in drupal that uses this querystring filter: https://github.com/edx/edx-mktg/pull/927

course_discovery/apps/api/v1/views.py
```diff
@@ -414,6 +419,12 @@ def list(self, request, *args, **kwargs):
               type: integer
               paramType: query
               multiple: false
+            - name: published_course_runs_only
```
mikedikan, line 422: what is the - before name?

course_discovery/apps/api/serializers.py
```diff
@@ -391,7 +394,8 @@ def get_courses(self, program):
             many=True,
             context={
                 'request': self.context.get('request'),
-                'program': program
+                'program': program,
+                'published_course_runs_only': self.context.get('published_course_runs_only', False)
```
mikedikan: can we put a trailing comma here?

mikedikan (changes requested):
Comment and a question (minor)

course_discovery/apps/api/serializers.py
```diff
@@ -12,6 +12,7 @@
 from taggit_serializer.serializers import TagListSerializerField, TaggitSerializer
 
 from course_discovery.apps.api.fields import StdImageSerializerField, ImageField
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
```
clintonb: Alphabetize.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -228,6 +229,34 @@ def test_with_exclusions(self):
...
         self.assertDictEqual(serializer.data, expected)
 
+    @ddt.data(
+        [CourseRunStatus.Unpublished, 1],
+        [CourseRunStatus.Unpublished, 0],
+        [CourseRunStatus.Published, 1],
+        [CourseRunStatus.Published, 0]
+    )
+    @ddt.unpack
+    def test_with_published_only_querystring(self, course_run_status, published_course_runs_only):
```
clintonb, line 239: Add a docstring explaining what you're testing.

course_discovery/apps/api/v1/views.py
clintonb: See http://django-rest-swagger.readthedocs.io/en/stable-0.3.x/yaml.html.

clintonb (approved):

course_discovery/apps/api/v1/views.py
```diff
@@ -397,6 +397,11 @@ class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
     filter_backends = (DjangoFilterBackend,)
     filter_class = filters.ProgramFilter
 
+    def get_serializer_context(self, *args, **kwargs):
+        context = super().get_serializer_context(*args, **kwargs)
+        context['published_course_runs_only'] = bool(int(self.request.GET.get('published_course_runs_only', 0)))
```
rlucioni: @schenedx you don't need to cast this to a boolean. Casting to an int is sufficient for your needs here.

course_discovery/apps/api/serializers.py
rlucioni: You also don't need to provide a default here. The default default (None) is sufficient.

course_discovery/apps/api/serializers.py
```diff
@@ -353,6 +354,8 @@ class ProgramCourseSerializer(CourseSerializer):
     def get_course_runs(self, course):
         program = self.context['program']
         course_runs = program.course_runs.filter(course=course)
+        if self.context.get('published_course_runs_only', False):
```
rlucioni: Again, no explicit default is necessary.

rlucioni (commented):
I left a little feedback for simplifying your code.

rlucioni:
@schenedx while working in this, did you consider extending https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/filters.py#L86? We use this filter mixin for our other querystring-based filtering, and it would make sense to keep it consistent, if possible.

rlucioni:
@schenedx I spoke with @clintonb briefly about the above; ignore me. The filters I was referring to are only useful for filtering the base queryset.

course_discovery/apps/api/serializers.py
rock345: Clinton please indicate if this is a nit or not!

schenedx:
@rlucioni @mikedikan Address all your comments. Please take another look.

rlucioni (approved):
LGTM

PR #332, opened 2016-09-15
Remove HistoricalRecords from course metadata models

rlucioni:
django-simple-history's HistoricalRecords conflict with django-extensions' AutoSlugField. Briefly: AutoSlugField tries to automatically generate a unique slug any time a new model instance is saved. This includes the addition of rows to historical tables. AutoSlugField attempts to generate a unique slug 100 times by appending '-i' where i is between 2 and 99, inclusive, then raises an exception, preventing a historical row from being created.

This happened very quickly for models whose slug field source(s) slugify to the empty string, but would have eventually happened for all models using HistoricalRecords. This behavior renders the historical tables (which we weren't using in the first place) useless until we can exclude specific fields from historical tracking: https://github.com/treyhunner/django-simple-history/issues/165. Note that the migration included here will drop the historical tables currently in place. This is fine - the data is not useful - and should not incur downtime.

An added benefit of fixing this issue by removing HistoricalRecords from course metadata models: a refresh with production data takes 11 minutes on my local machine.

ECOM-5579. @edx/ecommerce @fredsmith @jibsheet.

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -2,7 +2,6 @@ from django.contrib import admin, messages
 from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
 
-from simple_history.admin import SimpleHistoryAdmin
 from course_discovery.apps.course_metadata.forms import ProgramAdminForm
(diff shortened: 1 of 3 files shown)


clintonb:
👍

schenedx:
👍 Very nice

PR #337, opened 2016-09-19
Ignoring empty query string parameters sent to search endpoint

clintonb:
These empty parameters cause errors, resulting in failed responses.

ECOM-5685

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -4,6 +4,7 @@ import django_filters
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db.models import QuerySet
+from django.http import QueryDict
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/apps/api/filters.py
```diff
@@ -59,14 +60,29 @@ def build_query(self, **query_filters):
         return query
 
 
-class HaystackFacetFilterWithQueries(HaystackFacetFilter):
+class HaystackRequestFilterMixin:
+    @staticmethod
```
schenedx, line 64: #Nitpick Comments for the whole mixin class would help the readability of the code a lot.

schenedx (approved):
👍

course_discovery/apps/api/filters.py
clintonb: Other than something along the lines of "Mixin for Haystack requests," which is obvious from the name, I'm not sure what would make this more readable. The only method is well-commented.

PR #338, opened 2016-09-19
Add hidden field to CourseRun model and index

rlucioni:
Also add a filter to the list view of CourseRun admin so that users can determine which course runs are hidden. ECOM-5687.

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -359,6 +359,7 @@ class CourseRun(TimeStampedModel):
     card_image_url = models.URLField(null=True, blank=True)
     video = models.ForeignKey(Video, default=None, null=True, blank=True)
     slug = models.CharField(max_length=255, blank=True, null=True, db_index=True)
+    hidden = models.BooleanField(default=False)
 
     objects = CourseRunQuerySet.as_manager()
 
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -359,6 +359,7 @@ class CourseRun(TimeStampedModel):
     card_image_url = models.URLField(null=True, blank=True)
     video = models.ForeignKey(Video, default=None, null=True, blank=True)
     slug = models.CharField(max_length=255, blank=True, null=True, db_index=True)
+    hidden = models.BooleanField(default=False)
```
clintonb, line 362: Add `db_index=True`, and re-create the migration.

clintonb (approved):

schenedx (approved):
👍

course_discovery/apps/course_metadata/models.py
cpennington: An index on a boolean field may not actually be worth it... It's not going to have a whole lot of filtering value (unless we're frequently planning to query for the smaller of the subsets).

cpennington: (See http://stackoverflow.com/questions/6084460/mysql-not-using-index-when-checking-1-but-using-it-with-0 for a fairly good summary of why you might or might not want to index this).

cpennington: (If you've already discussed the tradeoffs, then a comment to that effect is probably in order)

rlucioni: TIL, thanks @cpennington. I will revert my most recent change and remove the index.

PR #339, opened 2016-09-19
Ingest the hidden field from Drupal if the partner has a marketing site.

rlucioni:
ECOM-5688.

@edx/ecommerce


clintonb (approved):
Update `assert_course_run_loaded`

PR #340, opened 2016-09-19
Add title boosting for search queries

bderusha:
Added title boosting the haystack way http://django-haystack.readthedocs.io/en/v2.5.0/boost.html?highlight=boosting#field-boost

@MatthewPiatetsky please have a look

### DIFF
--- a/course_discovery/apps/api/v1/views.py
+++ b/course_discovery/apps/api/v1/views.py
@@ -14,6 +14,8 @@ from drf_haystack.mixins import FacetMixin
 from drf_haystack.viewsets import HaystackViewSet
 from dry_rest_permissions.generics import DRYPermissions
 from edx_rest_framework_extensions.permissions import IsSuperuser
+from haystack.inputs import AutoQuery
+from haystack.query import SQ
 from rest_framework import status, viewsets
 from rest_framework.decorators import detail_route, list_route
 from rest_framework.exceptions import PermissionDenied, ParseError
@@ -601,6 +603,10 @@ class BaseHaystackViewSet(FacetMixin, HaystackViewSet):
 
     def filter_facet_queryset(self, queryset):
         queryset = super().filter_facet_queryset(queryset)
+
+        q = self.request.query_params.get('q')
+        if q:
+            queryset = queryset.filter(SQ(text=AutoQuery(q)) | SQ(title=AutoQuery(q)))
 
         facet_serializer_cls = self.get_facet_serializer_class()
         field_queries = getattr(facet_serializer_cls.Meta, 'field_queries', {})
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -6,6 +6,15 @@ from opaque_keys.edx.keys import CourseKey
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
 
+# http://django-haystack.readthedocs.io/en/v2.5.0/boost.html#field-boost
+# Boost title over all other parameters (multiplicative)
+# The max boost received from our boosting functions is ~6.
+# Having a boost of 25 for title gives most relevant titles a score higher
+# than our other boosting (which is what we want). But it's all relative.
+# If we altered our boosting functions to have a max score of 10
+# we would probably want to bump this number.
+TITLE_FIELD_BOOST = 25.0
+
 
 class OrganizationsMixin:
     def format_organization(self, organization):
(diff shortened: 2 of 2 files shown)


course_discovery/apps/api/v1/views.py
```diff
@@ -610,3 +612,11 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
     """ Search all content types. """
     facet_serializer_class = serializers.AggregateFacetSearchSerializer
     serializer_class = serializers.AggregateSearchSerializer
+
+    def filter_queryset(self, queryset):
```
clintonb: Move to `filter_facet_queryset` above so that all children of `BaseHaystackViewSet` can take advantage of this functionality.

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -6,6 +6,8 @@
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
 
+TITLE_FIELD_BOOST = 25.0
```
clintonb, line 16: Add a comment and link to docs explaining what this does, units, etc. For example, is this additive or multiplicative?

course_discovery/apps/api/serializers.py
```diff
@@ -700,10 +700,14 @@ class Meta:
 
 class CourseRunSearchSerializer(HaystackSerializer):
     availability = serializers.SerializerMethodField()
+    score = serializers.SerializerMethodField()
```
clintonb: Why is this necessary? Does DRF not simply render `result.score` as a float?

clintonb (approved):

course_discovery/apps/api/serializers.py
MatthewPiatetsky: talked to @bderusha, he said to remove score

course_discovery/apps/course_metadata/search_indexes.py
MatthewPiatetsky: added comment

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: is the change I made what you meant?

MatthewPiatetsky:
@clintonb are your comments addressed?  
@edx/ecommerce bump

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -6,6 +6,10 @@
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
 
+# http://django-haystack.readthedocs.io/en/v2.4.1/boost.html#field-boost
```
clintonb: Nit: We use v2.5.0

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -6,6 +6,10 @@
...
+# Boost title over all other parameters (multiplicative)
+TITLE_FIELD_BOOST = 25.0
```
clintonb, line 16: What's the justification for this multiplier?

clintonb (approved):

course_discovery/apps/api/v1/views.py
```diff
@@ -556,6 +558,10 @@ def facets(self, request):
     def filter_facet_queryset(self, queryset):
         queryset = super().filter_facet_queryset(queryset)
 
+        q = self.request.GET.get('q')
```
rlucioni: Preferred DRF style is to use `query_params` over `GET` (e.g., `q = self.request.query_params.get('q')`). http://www.django-rest-framework.org/api-guide/requests/#query_params

course_discovery/apps/course_metadata/search_indexes.py
rlucioni, line 16: Has thought been given to making this configurable via the Django admin? Is this something we might want to tune without re-deploying the application?

rlucioni (changes requested):

course_discovery/apps/course_metadata/search_indexes.py
bderusha: It requires a server restart to take affect since it is loaded when the class is loaded, so I didn't think a django-admin config was appropriate.  Do we have a way to kick the service like that in production if we did make it configurable?

rlucioni: @bderusha I see what you mean. I don't think so. Making this dynamic looks like it could be more involved than I thought.

bderusha: For our purposes the actual numerical value here doesn't matter so much as it's relation to the total amount of functional boosting that can be applied.

I chose this number after playing around with a dataset.  The max boost received from our boosting functions is ~6, having a boost of 25 for title gives most relevant titles a score higher than our other boosting (which is what we want).  But it's all relative.  If we altered our boosting functions to have a max score of 10 we would probably want to bump this number.  if we added a new function we can adjust the weights of all of them to keep the total at 6 and not require a deploy.

clintonb: @bderusha cool. Is this information documented someplace? If not, would you document it someplace, please?

rlucioni: A comment would be nice!

MatthewPiatetsky: added more to the comment

rlucioni (approved):

PR #341, opened 2016-09-19
Fixed program course/course run sorting issue

clintonb:
The sorting code now exits with the default value if a course has no course runs.

ECOM-5696


clintonb:
@edx/ecommerce please review

bderusha:
👍

schenedx (approved):
👍

rlucioni:
👍

PR #343, opened 2016-09-19
Caching the program endpoints

clintonb:
These endpoints are quite slow. Until they have been optimized, the responses must be cached.

ECOM-5709

### DIFF
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -331,6 +331,11 @@ REST_FRAMEWORK = {
     },
 }
 
+REST_FRAMEWORK_EXTENSIONS = {
+    'DEFAULT_CACHE_ERRORS': False,
+    'DEFAULT_CACHE_RESPONSE_TIMEOUT': 1800,
+}
+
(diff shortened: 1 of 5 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/settings/base.py
```diff
@@ -331,6 +331,11 @@
     },
 }
 
+REST_FRAMEWORK_EXTENSIONS = {
+    'DEFAULT_CACHE_ERRORS': False,
+    'DEFAULT_CACHE_RESPONSE_TIMEOUT': 1800,
```
MatthewPiatetsky, line 336: why 1800?

clintonb: 30 minutes. Enough time to ensure we don't kill the service, but not so much as to upset our users wanting to make frequent changes.

rlucioni:
👍 , though it's worth keeping in mind that with this caching in place, anyone editing a program will have to wait 30 minutes at a time to see their changes applied.

schenedx:
@clintonb Does the cache TTL have to be 30min? What's the reason behind this number?

PR #345, opened 2016-09-20
Programs endpoint fixes

clintonb:
- Fixed tests
- Restored usage of `ProgramSerializer` for all calls to `/api/v1/programs/`

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -1,4 +1,3 @@
-import unittest
 from datetime import datetime
 from urllib.parse import urlencode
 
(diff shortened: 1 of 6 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -94,56 +94,95 @@ def test_invalid_data_user_create(self):
...
 
 
-class CourseSerializerTests(TestCase):
-    def test_data(self):
-        course = CourseFactory()
-        video = course.video
-
-        request = make_request()
+class MinimalCourseSerializerTests(TestCase):
+    maxDiff = None
```
MatthewPiatetsky: are all the maxDiff's necessary?

clintonb: Nope. Removing.

MatthewPiatetsky:
Could you describe what fixes this PR is making?

rlucioni:
@MatthewPiatetsky this PR uses full serializers when serializing programs, restoring all fields removed yesterday. It also fixes the tests which broke yesterday (the tests were using the wrong serializers).

👍

MatthewPiatetsky:
@rlucioni what is it changing in addition to reverting yesterday's changes? if there is nothing else why not revert commit

clintonb:
@MatthewPiatetsky the commit has value in the sense that, ultimately, we will be using the minimal serializers.

PR #347, opened 2016-09-20
Query optimization

rlucioni:
@edx/ecommerce here's a WIP of today's query optimization work with @cpennington. We still need to fix any broken tests and mop up a few straggler queries, but this has resulted in a 96% reduction in queries made by program endpoints (~800 to 34) _while retaining all fields_.

These changes will allow us to keep Python-based intra-program course sorting, and should also allow us to re-enable the LMS integration.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -5,6 +5,7 @@ from urllib.parse import urlencode
 
 import pytz
 from django.contrib.auth import get_user_model
+from django.db.models.query import Prefetch
 from django.utils.translation import ugettext_lazy as _
 from drf_haystack.serializers import HaystackSerializer, HaystackFacetSerializer
 from rest_framework import serializers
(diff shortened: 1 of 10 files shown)


clintonb:
@rlucioni tests fixed, and branch rebased. The branch is yours once more.

rlucioni:
@clintonb thank you very much!

@clintonb @cpennington @edx/ecommerce any questions or concerns about this PR? The prefetched querysets might be better placed in the serializers, since those know what fields they're ultimately going to need available. However, that seems to me like an improvement we can make later. I'd like to merge and load test this in the morning so we can make a call on deployment.

clintonb (approved):

mikedikan:
@rlucioni & @cpennington I'm glad that you guys made your breakthrough!  Question: have we considered throttling on either client or server side for endpoints that we know are expensive?  We could have the throttling thresholds be environment-specific and configuration-based and utilize an in-memory or in-db queue to prevent thrashing.

cpennington:
No, I think it makes sense to go ahead and merge/release this as soon as you feel comfortable, and then take another pass to knock out the last lingering queries and improve the code layout.

cpennington:
A couple things that concern me, though:
1. The views covered by `assertNumQueries`: are any of those multi-object views, that might be scaling badly because we broke their prefetching?
2. Do we have an `assertNumQueries` test that covers the Programs list view?

rlucioni:
@mikedikan we actually already have server-side [throttling](https://github.com/edx/course-discovery/blob/master/course_discovery/settings/base.py#L326-L331) in place for all endpoints. We could tune this by viewset, if necessary.

@cpennington we do have `assertNumQueries` tests for both the [program list](https://github.com/edx/course-discovery/pull/347/files#diff-7742e0b7ac2c6a75f30dbbe9f600d9c2R75) and detail views. A look at the query count changes in our other tests suggests that our modifications to the prefetch dict in `serializers.py` did impact course and course run endpoints. We could address this by adding back the duplicated fields we removed from the queryset dict, or try moving that dict into the serializers now. The former would add some queries back to program endpoints. Thoughts?

course_discovery/apps/api/serializers.py
```diff
@@ -457,14 +424,88 @@ class ProgramSerializer(MinimalProgramSerializer):
...
+                'request': self.context.get('request'),
+                'program': program,
+                'published_course_runs_only': self.context.get('published_course_runs_only'),
+                'course_runs': course_runs,
+            }
+        )
+
+        return course_serializer.data
+
+    def sort_courses(self, program):
```
schenedx, line 511: Glad this is back

course_discovery/apps/course_metadata/models.py
```diff
@@ -635,57 +636,67 @@ def marketing_url(self):
...
     @property
     def course_runs(self):
+        """
+        Warning! You must call prefetch_related_objects_for_programs before calling this method.
+        Otherwise, this method will incur many, many queries when fetching related course runs.
+        """
         excluded_course_run_ids = [course_run.id for course_run in self.excluded_course_runs.all()]
-        return CourseRun.objects.filter(course__programs=self).exclude(id__in=excluded_course_run_ids)
+
+        for course in self.courses.all():
```
schenedx, line 645: So the risk here is, if there is another program related view that did not call the `prefetched_related_objects_for_programs`, the view will encounter significant performance issues, correct? I couldn't think of a better way to improve the situation. But I do think this is a potential trouble area

cpennington: Yeah. @rlucioni and I were just discussing how we could make it more obvious what/how to do prefetching whenever using a serializer.

rlucioni:
I took a look at the course and course_run endpoints locally and am seeing some regressions due to these changes (i.e., query counts scaling linearly). For example, course run seats are no longer being prefetched because of our changes. I don't think we want to merge this without addressing those regressions.

rlucioni:
To keep everyone on the same page: I spoke with @cpennington, and my path here is going to be distributing the prefetching currently done in [this function](https://github.com/edx/course-discovery/pull/347/files#diff-e5fc259d08849a0b07075ed748fc2890R77) to each of the serializers used by the program serializer. That will allow us to compose a prefetched queryset composed of the prefetched querysets defined by each of the child serializers, and should also not interfere with the prefetching for courses and course runs that was already in place. We can convert other top-level serializers (e.g., `CourseSerializer`) to this model and remove the static lists of prefetch fields at a later date.

mikedikan:
Thanks for the latest info @rlucioni - this approach makes sense to me.  Please continue to keep us updated.

rlucioni:
@edx/ecommerce I've just pushed up a third commit with fixes for the course and course run query regressions. Once I fix query count tests and squash, this will be ready to go. Let me know if you have questions.

course_discovery/apps/course_metadata/models.py
rlucioni: @schenedx following up on @cpennington's note: I was able to make this a bit clearer by moving the prefetched queryset onto the `ProgramSerializer`.

PR #349, opened 2016-09-21
ECOM-5436 Order courses in a program

MatthewPiatetsky:
@edx/ecommerce 
This PR makes the select2 widget used for courses reorderable.
I think this is far more user friendly than the existing many to many widget. 
Right now the code is somewhat hacky. 
I'm trying to figure out how to clean up different parts of the code and would appreciate any feedback - someone may know a much simpler way to do some of the things I am trying to do and achieve the same thing.
(If we find a way to implement this that we are satisfied with, we could easily make any elements we want orderable in this fashion)

select_widget

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -493,7 +493,10 @@ class ProgramSerializer(serializers.ModelSerializer):
         )
 
     def get_courses(self, program):
-        courses, course_runs = self.sort_courses(program)
+        if program.order_courses_by_start_date:
+            courses, course_runs = self.sort_courses(program)
+        else:
+            courses, course_runs = program.courses.all(), program.course_runs
 
         course_serializer = ProgramCourseSerializer(
             courses,
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -67,6 +67,21 @@ class ProgramViewSetTests(APITestCase):
         with self.assertNumQueries(89):
             self.assert_retrieve_success(program)
 
+    @ddt.data(
+        (True),
+        (False),
+    )
+    def test_retrieve_with_sorting_flag(self, order_courses_by_start_date=True):
+        """ Verify the number of queries is the same with sorting flag set to true. """
+        course_list = CourseFactory.create_batch(3)
+        for course in course_list:
+            CourseRunFactory(course=course)
+        program = ProgramFactory(courses=course_list, order_courses_by_start_date=order_courses_by_start_date)
+        num_queries = 132 if order_courses_by_start_date else 114
+        with self.assertNumQueries(num_queries):
+            self.assert_retrieve_success(program)
+        self.assertEqual(course_list, list(program.courses.all()))  # pylint: disable=no-member
+
     def test_retrieve_without_course_runs(self):
         """ Verify the endpoint returns data for a program even if the program's courses have no course runs. """
         course = CourseFactory()
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -81,8 +81,8 @@ class ProgramAdmin(admin.ModelAdmin):
         'min_hours_effort_per_week', 'max_hours_effort_per_week',
     )
     fields += (
-        'courses', 'custom_course_runs_display', 'excluded_course_runs', 'authoring_organizations',
-        'credit_backing_organizations'
+        'courses', 'order_courses_by_start_date', 'custom_course_runs_display', 'excluded_course_runs',
+        'authoring_organizations', 'credit_backing_organizations'
     )
     fields += filter_horizontal
     save_error = None
@@ -109,11 +109,17 @@ class ProgramAdmin(admin.ModelAdmin):
 
     def save_model(self, request, obj, form, change):
         try:
+            # courses are ordered by django id, but form.cleaned_data is ordered correctly
+            obj.courses = form.cleaned_data.get('courses')
             obj.save()
             self.save_error = False
         except ProgramPublisherException as ex:
             messages.add_message(request, messages.ERROR, ex.message)
             self.save_error = True
+
+    class Media:
+        js = ('bower_components/jquery-ui/ui/minified/jquery-ui.min.js',
+              'js/sortable_select.js')
 
 
 @admin.register(ProgramType)
--- a/course_discovery/apps/course_metadata/forms.py
+++ b/course_discovery/apps/course_metadata/forms.py
@@ -1,4 +1,4 @@
-from dal import autocomplete
+from dal import widgets
 from django import forms
 from django.core.exceptions import ValidationError
 from django.forms.utils import ErrorList
@@ -8,10 +8,32 @@ from course_discovery.apps.course_metadata.choices import ProgramStatus
 from course_discovery.apps.course_metadata.models import Program, CourseRun
 
 
-class ProgramAdminForm(forms.ModelForm):
+class HackDjangoAutocompleteMixin(object):
+    # It seems to me there is an issue with the select 2 widget in django autocomplete.
+    # When the widget loads selected choices it loads them in order of django id, not the order
+    # they are stored in in the database. This workaround works, but not sure what approach
+    # would be less hacky. Perhaps opening a PR to the django autocomplete repo if this is
+    # fact an issue?
+
+    class QuerySetSelectMixin2(widgets.WidgetMixin):
+
+        def filter_choices_to_render(self, selected_choices):
+            # preserve ordering of selected_choices in queryset
+            # https://codybonney.com/creating-a-queryset-from-a-list-while-preserving-order-using-django/
+            clauses = ' '.join(['WHEN id={} THEN {}'.format(pk, i) for i, pk in enumerate(selected_choices)])
+            ordering = 'CASE {} END'.format(clauses)
+            self.choices.queryset = self.choices.queryset.filter(
+                pk__in=[c for c in selected_choices if c]
+            ).extra(select={'ordering': ordering}, order_by=('ordering',))
+
+    widgets.QuerySetSelectMixin = QuerySetSelectMixin2
+
+
+class ProgramAdminForm(HackDjangoAutocompleteMixin, forms.ModelForm):
     class Meta:
         model = Program
         fields = '__all__'
+        from dal import autocomplete
         widgets = {
             'courses': autocomplete.ModelSelect2Multiple(
                 url='admin_metadata:course-autocomplete',
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -583,7 +583,11 @@ class Program(TimeStampedModel):
     )
     marketing_slug = models.CharField(
         help_text=_('Slug used to generate links to the marketing site'), blank=True, max_length=255, db_index=True)
-    courses = models.ManyToManyField(Course, related_name='programs')
+    courses = SortedManyToManyField(Course, related_name='programs')
+    order_courses_by_start_date = models.BooleanField(
+        default=True, verbose_name='Order Courses By Start Date',
+        help_text=_('If this box is not checked, courses will be ordered as in the courses select box above.')
+    )
     # NOTE (CCB): Editors of this field should validate the values to ensure only CourseRuns associated
     # with related Courses are stored.
     excluded_course_runs = models.ManyToManyField(CourseRun, blank=True)
--- a/course_discovery/static/js/sortable_select.js
+++ b/course_discovery/static/js/sortable_select.js
@@ -1,0 +1,44 @@
+function updateSelect2Data(visibleCourseTitles){
+    var i, j,
+        visibleCourseTitlesLength,
+        selectOptionsLength,
+        visibleCourseTitles = [],
+        selectOptions = [],
+        items = [],
+        selectOptionsSelector = '.field-courses .select2-hidden-accessible';
+
+    $('.field-courses .select2-selection__choice').each(function(index, value){
+        if (value.title){
+            visibleCourseTitles.push(value.title);
+        }
+    });
+
+    $('.field-courses .select2-hidden-accessible option').each(function(index, value){
+        selectOptions.push({id: value.value, text: value.text});
+    });
+
+    // Update select2 options with new data
+    visibleCourseTitlesLength = visibleCourseTitles.length;
+    selectOptionsLength = selectOptions.length;
+    for (i = 0; i < visibleCourseTitlesLength; i++) {
+        for (j = 0; j < selectOptionsLength; j++) {
+            if (selectOptions[j].text === visibleCourseTitles[i]){
+                items.push('<option selected="selected" value="' + selectOptions[j].id + '">' +
+                           selectOptions[j].text + '</option>'
+                );
+            }
+        }
+    }
+    if (items){
+        $(selectOptionsSelector).html(items.join('\n'));
+    }
+}
+$(window).load(function(){
+    $(function() {
+        var domSelector = '.field-courses .select2-selection--multiple';
+        $('.field-courses ul.select2-selection__rendered').sortable({
+            containment: 'parent',
+            update: updateSelect2Data
+        })
+    })
+});
--- a/bower.json
+++ b/bower.json
@@ -14,6 +14,7 @@
     "underscore": "~1.8.3",
     "moment": "~2.13.0",
     "pikaday": "https://github.com/owenmead/Pikaday.git#1.4.0",
-    "clipboard": "1.5.12"
+    "clipboard": "1.5.12",
+    "jquery-ui": "1.10.3"
   }
 }
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -434,6 +434,15 @@ class ProgramSerializerTests(TestCase):
         expected = self.get_expected_data(program, request)
         self.assertDictEqual(dict(serializer.data), expected)
 
+    def test_data_without_course_sorting(self):
+        request = make_request()
+        program = self.create_program()
+        program.order_courses_by_start_date = False
+        program.save()
+        serializer = ProgramSerializer(program, context={'request': request})
+        expected = self.get_expected_data(program, request)
+        self.assertDictEqual(dict(serializer.data), expected)
+
     def test_data_with_exclusions(self):
         """
         Verify we can specify program excluded_course_runs and the serializers will
--- a/course_discovery/apps/course_metadata/migrations/0029_auto_20160923_1306.py
+++ b/course_discovery/apps/course_metadata/migrations/0029_auto_20160923_1306.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from sortedm2m import fields, operations
+
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0028_courserun_hidden'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='program',
+            name='order_courses_by_start_date',
+            field=models.BooleanField(default=True, help_text='If this box is not checked, courses will be ordered as in the courses select box above.', verbose_name='Order Courses By Start Date'),
+        ),
+        operations.AlterSortedManyToManyField(
+            model_name='program',
+            name='courses',
+            field=fields.SortedManyToManyField(help_text=None, related_name='programs', to='course_metadata.Course'),
+        ),
+    ]
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -251,6 +251,7 @@ class ProgramFactory(factory.django.DjangoModelFactory):
     min_hours_effort_per_week = FuzzyInteger(2)
     max_hours_effort_per_week = FuzzyInteger(4)
     credit_redemption_overview = FuzzyText()
+    order_courses_by_start_date = True
 
     @factory.post_generation
     def courses(self, create, extracted, **kwargs):


course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,51 @@
...
+        $(".field-courses ul.select2-selection__rendered").sortable({
+            containment: 'parent'
+        })
+        var previousTitles = [];
+
+        // Every second, check if the dom select box has changed
+        // If it has, update the data.
+        // I was trying to trigger the updateSelect2Data off of events
+        // but it wasn't working well
+        setInterval(function(){
```
AlasdairSwan: This is not a good approach. If you are having issues with listening for moving events another approach would be to add a button alongside the input which would submit the order.

AlasdairSwan: Could you not listen to the select2 `change` event (https://select2.github.io/options.html#events)?

MatthewPiatetsky: You mean like an additional button?
I think we should try to find an approach that doesn't sacrifice user experience, since we could reuse this elsewhere.
I will try the select2 change event, though I don't know if the reordering triggers it

AlasdairSwan: I agree that the button is not ideal UX, but it would be preferable to having setInterval continuously running on the page and we could be more confident that all changes have been applied. The event listener approach is the preferred option.

AlasdairSwan: If the select2 change event doesn't work, the jQuery UI sortable one should http://api.jqueryui.com/sortable/#event-change

MatthewPiatetsky: yup we both found the jquery UI event, that was the way to go!

clintonb:
The `SortedManyToMany` field has a default widget. While not great, it is usable. Did you consider it?

course_discovery/apps/course_metadata/models.py
```diff
@@ -583,7 +583,7 @@ class Program(TimeStampedModel):
     )
     marketing_slug = models.CharField(
         help_text=_('Slug used to generate links to the marketing site'), blank=True, max_length=255, db_index=True)
-    courses = models.ManyToManyField(Course, related_name='programs')
+    courses = SortedManyToManyField(Course, related_name='programs')
```
clintonb, line 586: Unless you modify the sorting in the serializer, this won't have any effect.

MatthewPiatetsky:
@clintonb the default widget allows you to filter, but you can't sort while filtered because the other options don't show up.  
When not filtered you have a huge dropdown of every course, which makes it very difficult to use

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: good point, thanks!
I will talk with product to find out how the sorting should behave with the default

course_discovery/apps/course_metadata/admin.py
```diff
@@ -109,12 +109,18 @@ def response_change(self, request, obj):
...
+            # courses are ordered by django id, but form.cleaned_data is ordered correctly
+            obj.courses = form.cleaned_data.get('courses')
             obj.save()
             self.save_error = False
         except ProgramPublisherException as ex:
             messages.add_message(request, messages.ERROR, ex.message)
             self.save_error = True
 
+    class Media:
+        js = ('//code.jquery.com/ui/1.10.3/jquery-ui.js',
```
rlucioni: What about installing this with npm or bower? I don't think we want to load this directly from jquery.com like this.

course_discovery/apps/course_metadata/forms.py
```diff
@@ -8,10 +7,32 @@
...
+                pk__in=[c for c in selected_choices if c]
+            ).extra(select={'ordering': ordering}, order_by=('ordering',))
+    widgets.QuerySetSelectMixin = QuerySetSelectMixin2
+
+
+class ProgramAdminForm(HackDjangoAutocompleteMixin, forms.ModelForm):
     class Meta:
         model = Program
         fields = '__all__'
+        from dal import autocomplete
```
rlucioni, line 36: What's the motivation behind nesting this import and the one above?

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: Talked with Lauren and Jasper, we went with a checkbox which determines whether ordering by date is used

clintonb:
The default widget allows you to filter and sort (drag-and-drop). The only downside to the default widget is the fact that when you filter, selected items appear in their original position. It might be simpler to make selected items appear at the beginning/end of the list than to write an entirely new widget.

Regardless of the approach, we should consider contributing this back to https://github.com/gregmuellegger/django-sortedm2m.

course_discovery/apps/course_metadata/models.py
clintonb: (facepalm) Who else was involved in this decision? Has it been documented? This will lead to issues with folks wondering why pages have different sorting orders. We should avoid introducing alternate behaviors in this manner.

MatthewPiatetsky: Lauren and Jasper, if you disagree you can talk to them

MatthewPiatetsky:
@clintonb true, if the default widget was modified so that the selected options appeared at the top that would be usable, but imo the select2 option is more user friendly and consistent with our current experience.
This is what the default widget would behave like by default. 
We could certainly contribute back, though that would introduce a dependency on the django autocomplete light library which they might not want

(This is what the widget looks like by default for reference)
old_widget3

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,39 @@
...
+        }
+    })
+
+    $('.field-courses .select2-hidden-accessible option').each(function(index, value){
+        selectOptions.push({id: value.value, text: value.text});
+    })
+
+    // Update select2 options with new data
+    for (var i = 0; i < visibleCourseTitles.length; i++) {
+        for (var j = 0; j < selectOptions.length; j++) {
```
AlasdairSwan: All variables should be declared at the top of the function

AlasdairSwan: It is more performant to assign the length value to a variable instead of calculating it on each iteration.

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,39 @@
...
+            if (selectOptions[j].text === visibleCourseTitles[i]){
+                items.push("<option selected=\"selected\" value=\"" + selectOptions[j].id + "\">" +
+                           selectOptions[j].text + "</option>\n"
```
AlasdairSwan: For consistency please continue using single quotes here, as used earlier in the file.

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,39 @@
...
+                );
+            }
+        }
+    }
+    if (items){
+        $(selectOptionsSelector).html(items.join(""));
```
AlasdairSwan: For consistency please continue using single quotes here, as used earlier in the file.

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+$(window).load(function(){
+    $(function() {
+        var domSelector = ".field-courses .select2-selection--multiple";
+        $(".field-courses ul.select2-selection__rendered").sortable({
```
AlasdairSwan: For consistency please continue using single quotes here, as used earlier in the file.

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+    }
+    if (items){
+        $(selectOptionsSelector).html(items.join(""));
+    }
+}
+$(window).load(function(){
+    $(function() {
+        var domSelector = ".field-courses .select2-selection--multiple";
```
AlasdairSwan: For consistency please continue using single quotes here, as used earlier in the file.

course_discovery/apps/course_metadata/admin.py
MatthewPiatetsky: @rlucioni  is the way I am importing it now correct? (note the diff on the comment is outdated)

AlasdairSwan (changes requested):

course_discovery/apps/course_metadata/forms.py
MatthewPiatetsky: one above doesn't need to be nested, this one does because the Mixin is overriding functionality in the library, like you can see in the comment.  
Not sure what the right approach is here, perhaps to also open a PR to that library?

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,44 @@
...
+        visibleCourseTitles = [],
+        selectOptions = [],
+        items = [],
+        selectOptionsSelector = '.field-courses .select2-hidden-accessible';
+
+    $('.field-courses .select2-selection__choice').each(function(index, value){
+        if (value.title){
+            visibleCourseTitles.push(value.title);
+        }
+    })
```
AlasdairSwan: Missing `;`

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,44 @@
...
+
+    $('.field-courses .select2-hidden-accessible option').each(function(index, value){
+        selectOptions.push({id: value.value, text: value.text});
+    })
```
AlasdairSwan: Missing `;`

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // Update select2 options with new data
+    visibleCourseTitlesLength = visibleCourseTitles.length;
+    selectOptionsLength = selectOptions.length;
+    for (i = 0; i < visibleCourseTitlesLength; i++) {
+        for (j = 0; j < selectOptionsLength; j++) {
+            if (selectOptions[j].text === visibleCourseTitles[i]){
+                items.push('<option selected=\"selected\" value=\"' + selectOptions[j].id + '\">' +
```
AlasdairSwan: Because you are using single quotes you don't need to escape the double quotes, i.e.  you can just do this;
`items.push('<option selected="selected" value="' + selectOptions[j].id + '">'`

course_discovery/static/js/sortable_select.js
```diff
@@ -0,0 +1,44 @@
...
+                           selectOptions[j].text + '</option>\n'
```
AlasdairSwan: Why the `\n`?

AlasdairSwan (changes requested):
Just some minor updates I missed on my first pass then should be good to merge.

course_discovery/static/js/sortable_select.js
MatthewPiatetsky: This is generating html and in the existing html the options are on a new line, though I realized it would make more sense for this to be in the join.  
It would still work without the \n but the HTML would be different and less readable

AlasdairSwan (approved):
:shipit: 👍

course_discovery/apps/api/serializers.py
```diff
@@ -493,7 +493,10 @@ def prefetch_queryset(cls):
         )
 
     def get_courses(self, program):
-        courses, course_runs = self.sort_courses(program)
+        if program.order_courses_by_start_date:
+            courses, course_runs = self.sort_courses(program)
+        else:
+            courses, course_runs = list(program.courses.all()), list(program.course_runs)
```
rlucioni: Do you need to force queryset evaluation here?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -109,12 +109,18 @@ def response_change(self, request, obj):
...
+            # courses are ordered by django id, but form.cleaned_data is ordered correctly
+            obj.courses = form.cleaned_data.get('courses')
             obj.save()
             self.save_error = False
         except ProgramPublisherException as ex:
             messages.add_message(request, messages.ERROR, ex.message)
             self.save_error = True
 
+    class Media:
+        js = ('bower_components/jquery-ui/ui/minified/jquery-ui.min.js',
```
rlucioni, line 121: @clintonb is this the right way to include these assets, or do they get copied somewhere else during static asset collection?

course_discovery/apps/course_metadata/forms.py
```diff
@@ -8,10 +8,31 @@
...
+
+        def filter_choices_to_render(self, selected_choices):
+            # preserve ordering of selected_choices in queryset
+            # https://codybonney.com/creating-a-queryset-from-a-list-while-preserving-order-using-django/
+            clauses = ' '.join(['WHEN id=%s THEN %s' % (pk, i) for i, pk in enumerate(selected_choices)])
+            ordering = 'CASE %s END' % clauses
+            self.choices.queryset = self.choices.queryset.filter(
+                pk__in=[c for c in selected_choices if c]
+            ).extra(select={'ordering': ordering}, order_by=('ordering',))
+    widgets.QuerySetSelectMixin = QuerySetSelectMixin2
```
rlucioni, line 29: Nitpick: add an empty line above this to make it a little more readable.

course_discovery/apps/course_metadata/models.py
rlucioni: Does this change require a migration?

Edit: This is Matthew replying because I don't have a reply button for some reason, there is a migration in this PR

course_discovery/apps/api/serializers.py
MatthewPiatetsky: it looks like they are being passed as context to a serializer which is expecting the type to be list

rlucioni:
@MatthewPiatetsky how does this affect query counts for the program list endpoint? Is there a performance cost for this new sorting?

course_discovery/apps/course_metadata/admin.py
clintonb: This is fine for now.

MatthewPiatetsky:
@rlucioni I added a test which seems to indicate that the number of queries is the same

course_discovery/apps/api/serializers.py
rlucioni: Have you tried passing the querysets through to see what happens? Just curious.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -67,6 +67,14 @@ def test_retrieve(self):
         with self.assertNumQueries(89):
             self.assert_retrieve_success(program)
 
+    def test_retrieve_without_sorting(self):
+        """ Verify the number of queries is the same without sorting. """
+        course = CourseFactory()
+        program = ProgramFactory(courses=[course], order_courses_by_start_date=False)
+        program = self.create_program()
```
rlucioni: I don't think this test is doing what you want, since you've redefined `program`. The one being retrieved isn't the one with start date sorting disabled.

course_discovery/apps/course_metadata/models.py
rlucioni: I see them. They don't need to be two separate migrations. If you remove them both then recreate migrations, you'll only have to add one.

MatthewPiatetsky: merged into one

course_discovery/apps/api/serializers.py
MatthewPiatetsky: seems like it works, nvm, they just needed to be iterable

course_discovery/apps/course_metadata/forms.py
MatthewPiatetsky: done

course_discovery/apps/api/v1/tests/test_views/test_programs.py
MatthewPiatetsky: Good point, now that I've fixed the test, it is actually 55 queries rather than 89 (which seems good?).
Or should I be concerned?

course_discovery/apps/course_metadata/forms.py
```diff
@@ -8,10 +8,32 @@
...
+    # would be less hacky. Perhaps opening a PR to the django autocomplete repo if this is
+    # fact an issue?
+
+    class QuerySetSelectMixin2(widgets.WidgetMixin):
+
+        def filter_choices_to_render(self, selected_choices):
+            # preserve ordering of selected_choices in queryset
+            # https://codybonney.com/creating-a-queryset-from-a-list-while-preserving-order-using-django/
+            clauses = ' '.join(['WHEN id=%s THEN %s' % (pk, i) for i, pk in enumerate(selected_choices)])
+            ordering = 'CASE %s END' % clauses
```
rlucioni: Nitpick: this is old-style string formatting. Please stick to new-style (i.e., `.format()`).

course_discovery/apps/course_metadata/forms.py
```diff
@@ -8,10 +8,32 @@
...
+            self.choices.queryset = self.choices.queryset.filter(
+                pk__in=[c for c in selected_choices if c]
+            ).extra(select={'ordering': ordering}, order_by=('ordering',))
```
rlucioni, line 27: order_by shouldn't require a tuple for a single arg.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -67,6 +67,13 @@ def test_retrieve(self):
         with self.assertNumQueries(89):
             self.assert_retrieve_success(program)
 
+    def test_retrieve_without_sorting(self):
+        """ Verify the number of queries is the same without sorting. """
+        course = CourseFactory()
+        program = ProgramFactory(courses=[course], order_courses_by_start_date=False)
+        with self.assertNumQueries(55):
```
rlucioni: Is this unit test really reflective of the behavior we'd expect in production? You're using a program with one course and no ordering overrides. If the program instead had, say, three courses and an explicitly-defined ordering, how do things change?

Relatedly, have you had a chance to use the debug toolbar to understand what this does to query counts when using the program list endpoint? I don't think the unit test is necessarily sufficient. We don't want to introduce queries that scale linearly in the size of the response. @bderusha or I can help you if you have questions.

course_discovery/apps/course_metadata/forms.py
MatthewPiatetsky: changed to format

course_discovery/apps/course_metadata/forms.py
MatthewPiatetsky: I think it does here, if I make it a string I get an error loading the page

course_discovery/apps/api/v1/tests/test_views/test_programs.py
MatthewPiatetsky: I added two tests with the sorting toggled on and off and 3 courses, is that better?
For the django toolbar, I loaded this page 
http://localhost:8008/api/v1/programs/03a198ae-98c2-4a10-ab99-2c5ec423abb5/
that I was using for testing on master, and then on my branch with the sorting button checked and unchecked. I got 24 SQL queries each time. Is that the correct thing to check?

rlucioni: @MatthewPiatetsky did that course you were using to test with have courses and an explicit ordering defined? You should also be checking the list view (/api/v1/programs/) for several programs. I don't know if the unit test is sufficient because I don't know how the sorted m2m handles sorting. As it is, you don't seem to be defining any sort order in the tests.

MatthewPiatetsky: @rlucioni With this field type, I believe the courses are always stored in a sorted order (the one defined in the widget before you click save). I was testing the number of queries here rather than the sorting, but I added an assert to verify the order as well. It will be the order of the courses in the list (whether the flag is set or not). The stored order is overriden by the start date order if the checkbox is checked with the existing sorting that happens in the serializer.  
However, the endpoint I was testing with the django toolbar is the api endpoint, so the number of queries includes the serializer ordering

MatthewPiatetsky: I fixed the tests on this PR and the number of queries did not change in my local testing, so I think I've addressed your comments

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -67,6 +67,26 @@ def test_retrieve(self):
...
+        with self.assertNumQueries(132):
+            self.assert_retrieve_success(program)
+        self.assertEqual(course_list, list(program.courses.all()))  # pylint: disable=no-member
+
+    def test_retrieve_with_sorting_flag_false(self):
+        """ Verify the number of queries is the same with sorting flag set to false. """
+        course_list = CourseFactory.create_batch(3)
+        for course in course_list:
+            CourseRunFactory(course=course)
+        program = ProgramFactory(courses=course_list, order_courses_by_start_date=False)
```
rlucioni: Instead of duplicating this test code, use ddt to provide both Boolean arguments to one test method.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -67,6 +67,26 @@ def test_retrieve(self):
...
+        with self.assertNumQueries(114):
+            self.assert_retrieve_success(program)
+        self.assertEqual(course_list, list(program.courses.all()))  # pylint: disable=no-member
```
rlucioni, line 83: Nitpick: As before, converting to a list like this shouldn't be necessary. Querysets are iterable.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
rlucioni: @MatthewPiatetsky let's chat to close the loop on this one. I may not have been clear enough earlier.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
MatthewPiatetsky: good point, thanks!

course_discovery/apps/api/v1/tests/test_views/test_programs.py
MatthewPiatetsky: I believe assertEqual does a direct equality comparison and a list is a different type from a queryset. assertEqual doesn't return true if I don't evaluate the queryset to a list here. (i tried it)

rlucioni: Good to know, thanks.

rlucioni (approved):

PR #350, opened 2016-09-21
Fix CorporateEndorsement prefetching

rlucioni:
endorser is a field on Endorsement, not CorporateEndorsement. We should be prefetching individual_endorsements.

@edx/ecommerce


MatthewPiatetsky:
👍

clintonb (approved):
Do query numbers in tests need to be updated?

rlucioni:
@clintonb I'll add a test that includes use of this field. When I missed this earlier, it wasn't caught by tests.

clintonb:
@rlucioni It might be useful to copy the program creation code from the serializer tests to the view tests. This would result in a program with almost all data (except course subjects, I believe).

rlucioni:
@clintonb thanks for the tip. Tests updated.

PR #352, opened 2016-09-22
Ignoring empty query string parameters sent to search endpoint

clintonb:
These empty parameters cause errors, resulting in failed responses.

ECOM-5685

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -59,14 +59,34 @@ class FacetQueryBuilderWithQueries(FacetQueryBuilder):
         return query
 
 
-class HaystackFacetFilterWithQueries(HaystackFacetFilter):
-    query_builder_class = FacetQueryBuilderWithQueries
-
-
-class HaystackFilter(DefaultHaystackFilter):
+class HaystackRequestFilterMixin:
     @staticmethod
     def get_request_filters(request):
         filters = HaystackFacetFilter.get_request_filters(request)
+
+        # Remove items with empty values.
+        #
+        # NOTE 1: We copy the keys list to avoid a RuntimeError that occurs when modifying a dict while iterating.
+        #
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/ecommerce please review

clintonb:
Fixes issues related to #351 and #337.

course_discovery/apps/api/filters.py
```diff
@@ -59,14 +59,29 @@ def build_query(self, **query_filters):
...
+class HaystackRequestFilterMixin:
+    @staticmethod
+    def get_request_filters(request):
+        filters = HaystackFacetFilter.get_request_filters(request)
+
+        # Remove items with empty values.
+        # NOTE: We copy the keys list to avoid a RuntimeError that occurs when modifying a dict
+        # while iterating.
+        for key in list(filters.keys()):
+            if not filters[key]:
```
MatthewPiatetsky, line 76: What about falsey non-empty values like False?

clintonb: `False` would be the string `'False'`. `0` however, is something for which we need to account.

clintonb: Test added. Query strings are parsed as...strings. Numbers, including 0, are represented as strings. `'0'` is a non-false value.

schenedx (approved):
👍

rlucioni:
👍

course_discovery/apps/api/filters.py
MatthewPiatetsky: if we're making a copy of the keys in memory anyway, another option would be
`filters = [key for key,value in filters.items() if value]`
(or use urllib)
but doesn't matter too much

MatthewPiatetsky:
👍

course_discovery/apps/api/filters.py
clintonb: :) That's what broke before. I will add a comment in the code explaining why that doesn't work for `QueryDict` objects.

PR #354, opened 2016-09-27
[ECOM-5701] Otto performance improvement

bderusha:
Reduce the number of threads used when calling the ecommerce system to limit concurrent load.

@rlucioni please review
FYI @edx/ecommerce


clintonb (approved):

rlucioni:
@bderusha looks good, but are we confident that this is enough to address the issue? Were you able to try it out?

bderusha:
@rlucioni I was planning on testing this in stage

::EDIT::
Thinking about it now, however I realize that I can test it locally by pointing my local at stage.  I will do that first

rlucioni:
@bderusha 👍, pointing my local at it is what I would have done.

PR #355, opened 2016-09-27
ECOM-5740 Only include programs for which the course runs are not excluded

MatthewPiatetsky:
@edx/ecommerce 

This change does add an additional query, but I've verified that with the prefetch it does not scale with the size of the input, in that adding more programs does not add more queries (though it did before I added the prefetch).
Initially I tried using the django queryset exclude but that wasn't working with this prefetch.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -75,6 +75,7 @@ PREFETCH_FIELDS = {
         'course__programs',
         'course__programs__partner',
         'course__programs__type',
+        'course__programs__excluded_course_runs',
         'language',
         'seats',
         'seats__currency',
@@ -350,7 +351,13 @@ class CourseRunSerializer(TimestampModelSerializer):
 
 class CourseRunWithProgramsSerializer(CourseRunSerializer):
     """A ``CourseRunSerializer`` which includes programs derived from parent course."""
-    programs = NestedProgramSerializer(many=True)
+    programs = serializers.SerializerMethodField()
+
+    def get_programs(self, obj):
+        programs = [program for program in obj.programs.all()
+                    if obj.id not in (run.id for run in program.excluded_course_runs.all())]
+
+        return NestedProgramSerializer(programs, many=True).data
 
     class Meta(CourseRunSerializer.Meta):
         model = CourseRun
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -190,6 +190,23 @@ class CourseRunWithProgramsSerializerTests(TestCase):
 
         self.assertDictEqual(serializer.data, expected)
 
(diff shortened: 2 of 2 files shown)


mikedikan (changes requested):
Seems like a good candidate for a unit test - especially since this was a bug

course_discovery/apps/api/serializers.py
```diff
@@ -350,7 +351,13 @@ def get_instructors(self, obj):  # pylint: disable=unused-argument
 
 class CourseRunWithProgramsSerializer(CourseRunSerializer):
     """A ``CourseRunSerializer`` which includes programs derived from parent course."""
-    programs = NestedProgramSerializer(many=True)
+    programs = serializers.SerializerMethodField()
+
+    def get_programs(self, obj):
+        programs = [program for program in obj.programs.all()
```
clintonb, line 357: The filtering can be done in the DB: `obj.programs.exclude(excluded_course_runs=obj)`

clintonb (changes requested):
This needs a test.

course_discovery/apps/api/serializers.py
MatthewPiatetsky: that's what I tried initially, but then I get queries that scale linearly with the number of programs.  
Do you know what is different about this that makes the prefetching not work?

clintonb: Ah. Filtering doesn't work on pre-fetched data. The current approach is correct.

MatthewPiatetsky:
@mikedikan @clintonb added a test.  
Is this the test you were expecting?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -190,6 +190,22 @@ def test_data(self):
 
         self.assertDictEqual(serializer.data, expected)
 
+    def test_data_excluded_course_run(self):
+        request = make_request()
+        course_run = CourseRunFactory()
+        serializer_context = {'request': request}
+        serializer = CourseRunWithProgramsSerializer(course_run, context=serializer_context)
+        program = ProgramFactory(courses=[course_run.course])
```
clintonb: `program = ProgramFactory(courses=[course_run.course], excluded_course_runs=[course_run])`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -190,6 +190,22 @@ def test_data(self):
 
         self.assertDictEqual(serializer.data, expected)
 
+    def test_data_excluded_course_run(self):
```
clintonb, line 193: Add a docstring explaining what you're testing. It is not immediately obvious from the name.

clintonb (approved):
Merge after the remaining comments are addressed.

mikedikan (approved):

PR #357, opened 2016-09-28
Add exclude_utm querystring parameter

rlucioni:
Calls to course, course run, and program APIs can use this parameter to prevent UTM parameters from being appended to marketing URLs. ECOM-5782.

@edx/ecommerce please review.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -8,7 +8,7 @@ from django.db.models.functions import Lower
 from rest_framework.reverse import reverse
 from rest_framework.test import APITestCase, APIRequestFactory
 
-from course_discovery.apps.api.serializers import CourseRunWithProgramsSerializer
+from course_discovery.apps.api.v1.tests.test_views.mixins import SerializationMixin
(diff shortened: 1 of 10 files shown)


course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -28,8 +28,8 @@ def setUp(self):
         self.request = APIRequestFactory().get('/')
         self.request.user = self.user
 
-    def serialize_course_run(self, course_run, **kwargs):
-        return CourseRunWithProgramsSerializer(course_run, context={'request': self.request}, **kwargs).data
+    # def serialize_course_run(self, course_run, **kwargs):
```
schenedx: Maybe we should just remove this function rather than comment out?

schenedx (changes requested):
Just a minor comment on commented out code, rest LGTM

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
rlucioni: Whoops, missed that. Removed.

clintonb (approved):

schenedx (approved):

PR #361, opened 2016-09-29
Fix malformed table error

marjev:
@clintonb This will fix the malformed table error. Github didn't show any errors on the preview and I noticed that table is missing when I went to the documentation page. I used an online validator and noticed a space character is missing. Can you take a look?

Thanks!


clintonb (approved):

clintonb:
You can build the docs locally: `cd docs && make html`

PR #363, opened 2016-09-30
Further accelerate metadata refresh

rlucioni:
A Waffle switch can now be used to toggle parallelized execution of the data loading pipeline. The pipeline is structured so that stages that can run independently of others are interleaved using separate processes.

A separate Waffle switch can be used to toggle threaded writes to the database. Instead of using threads to request all data, then writing it all serially, use this switch to spawn separate threads responsible for reading and writing each page of data.

ECOM-5871

This was getting stale, so I decided to wrap it up while it was still fresh. @edx/ecommerce please review.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/__init__.py
+++ b/course_discovery/apps/course_metadata/data_loaders/__init__.py
@@ -26,13 +26,15 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
     SUPPORTED_TOKEN_TYPES = ('bearer', 'jwt',)
     MARKDOWN_CLEANUP_REGEX = re.compile(r'^<p>(.*)</p>$')
 
-    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None):
+    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None, is_threadsafe=False):
         """
         Arguments:
             partner (Partner): Partner which owns the APIs and data being loaded
             api_url (str): URL of the API from which data is loaded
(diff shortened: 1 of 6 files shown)


course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -26,13 +26,15 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
...
-    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None):
+    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None, is_threadsafe=False):
         """
         Arguments:
             partner (Partner): Partner which owns the APIs and data being loaded
             api_url (str): URL of the API from which data is loaded
             access_token (str): OAuth2 access token
             token_type (str): The type of access token passed in (e.g. Bearer, JWT)
+            max_workers (int): Number of worker threads to use when traversing paginated responses.
+            is_threadsafe (bool): True if threads can be used to write data.
```
clintonb: Should this be "multiple threads"?

clintonb (approved):

course_discovery/apps/course_metadata/data_loaders/__init__.py
rlucioni: I felt that was implied, but I've updated to make it explicit.

clintonb:
I just realized we forgot to add a migration creating the switch.

rlucioni:
@clintonb addressed (amongst other things) by https://github.com/edx/course-discovery/pull/364.

PR #367, opened 2016-10-04
Added header/footer.

waheedahmed:
[ECOM-5284](https://openedx.atlassian.net/browse/ECOM-5284)

As a Publisher user, I would like to see a consistent edX Header/Footer as I navigate throughout the Publisher experience. 

**Design:** http://cim31f.axshare.com/course_runs.html

**Acceptance Criteria:**
1. Validate the header/footer matches the wireframe.
2. Validate the links on the header/footer work.
3. Validate the header/footer is persistent throughout every page as navigating through the Publisher.

screencapture-localhost-8004-publisher-course_runs-1475567322324

### DIFF
--- a/course_discovery/static/sass/publisher/course_detail.scss
+++ b/course_discovery/static/sass/publisher/course_detail.scss
@@ -13,14 +13,153 @@ $light-gray: rgba(204, 204, 204, 1);
 }
 
 .course-detail {
-    @include margin(0);
+    @include margin(0px, 0px, 0px, 0px);
 }
 
 .container {
     .layout-header {
         img, h1 {
-            padding: 20px;
             @include float(left);
+            @include padding(20px, 20px, 20px, 20px);
+        }
+
+        .layout-col {
+
+            .nav-account {
+                @include padding(35px, 0px, 35px, 0px);
+                @include margin-right(20px);
+                @include float(right);
+                width: auto;
+                display: inline-block;
+                position: relative;
+                vertical-align: middle;
+
+                ol {
+                    margin-bottom: 0;
+                }
+
+                .wrapper-nav-sub {
+                    @include right(0);
+                    display: none;
+                    opacity: 1.0;
+                    pointer-events: auto;
+                    overflow: visible;
+                    height: auto;
+                    position: absolute;
+
+                    &.is-shown {
+                        display: block;
+                    }
+
+                    .nav-sub {
+                        border-radius: 4px;
+                        box-sizing: border-box;
+                        box-shadow: 0 1px 1px rgba(0,0,0,0.1);
+                        position: relative;
+                        width: 100%;
+                        border: 1px solid #ccc;
+                        background: #fff;
+
+                        &:before {
+                            @include right(20px);
+                            @include margin-right(-11px);
+                            border-color: rgba(178,178,178,0);
+                            border-bottom-color: #ccc;
+                            border-width: 11px;
+                            bottom: 100%;
+                            content: " ";
+                            height: 0;
+                            width: 0;
+                            position: absolute;
+                            pointer-events: none;
+                        }
+
+                        &:after {
+                            @include right(20px);
+                            border-color: rgba(255,255,255,0);
+                            border-bottom-color: #fff;
+                            border-width: 10px;
+                            bottom: 100%;
+                            content: " ";
+                            height: 0;
+                            width: 0;
+                            position: absolute;
+                            pointer-events: none;
+                            @include margin-right(-10px);
+                        }
+
+                        .nav-menu-item {
+                            @include padding(20px, 20px, 20px, 20px);
+                        }
+
+                        .nav-account-signout {
+                            border-top: 1px solid #ccc;
+
+                            a {
+                                @include text-align(center);
+                                text-decoration: none;
+                                width: 100%;
+                            }
+                        }
+                    }
+                }
+
+                .nav-account-user {
+
+                    .title.is-selected {
+                        box-shadow: inset 0 1px 2px 1px rgba(0,0,0,0.1);
+                        border-color: #ccc;
+                    }
+                }
+
+                .nav-item {
+                    display: inline-block;
+                    vertical-align: middle;
+                    margin-bottom: 0;
+
+                    .title {
+                        @include padding(5px, 10px, 5px, 10px);
+                        color: #666;
+                        margin-bottom: 0;
+                        border: 1px solid transparent;
+                        border-radius: 4px;
+                        max-width: 150px;
+
+                        .fa-caret-down {
+                            opacity: 0.25;
+                        }
+
+                        &:hover {
+                            cursor: pointer;
+                            color: #065683;
+                            border-color: #ccc;
+
+                            .label {
+
+                                a {
+                                    color: #065683;
+                                }
+                            }
+
+                            .fa-caret-down {
+                                opacity: 1.0;
+                                color: #065683;
+                            }
+                        }
+
+                        .label {
+
+                            a {
+                                color: #666;
+                                font-weight: 800;
+                                font-size: 18px;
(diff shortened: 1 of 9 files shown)


waheedahmed:
@awais786 @bderusha Please review.

course_discovery/templates/footer.html
```diff
@@ -1,6 +1,42 @@
...
+        <div class="footer-content-primary">
+          <div class="colophon">
+            {% now "Y" as current_year %}
+            <p>&copy; {{ current_year }} <a data-rel="edx.org" href="https://wwww.edx.org" rel="external">{{ platform_name }}</a>.</p>
+          </div>
+
+            <nav class="nav-peripheral">
+              <ol>
+              <li class="nav-item nav-peripheral-tos">
+                <a data-rel="edx.org" href="https://www.edx.org/edx-terms-service">{% trans "Terms of Service" %}</a>
```
bderusha: we can't hardcode edx links.  These need to be dynamic

course_discovery/templates/footer.html
```diff
@@ -1,6 +1,42 @@
...
+              </li>
+              <li class="nav-item nav-peripheral-pp">
+                <a data-rel="edx.org" href="https://www.edx.org/edx-privacy-policy">{% trans "Privacy Policy" %}</a>
```
bderusha: same here

course_discovery/templates/footer.html
```diff
@@ -1,6 +1,42 @@
...
+        </div>
+
+        <div class="footer-content-secondary" aria-label="{% trans 'Legal' %}">
+          <div class="footer-about-copyright">
+            <!--Site operators: Please do not remove this paragraph! This attributes back to edX and makes your acknowledgement of edX's trademarks clear.-->
+            <p>
+                <!--Translators: 'EdX', 'edX', 'Studio', and 'Open edX' are trademarks of 'edX Inc.'. Please do not translate any of these trademarks and company names.-->
+                {% with link_start='<a data-rel="edx.org" href="https://www.edx.org/">' link_end='</a>' %}
+                    {% blocktrans %}
+                        EdX, Open edX, Studio, and the edX and Open edX logos are registered trademarks or trademarks of {{link_start}}edX Inc.{{link_end}}
```
bderusha: What about white label?  How does this work?  @mattdrayer ?

awais786:
👍

course_discovery/static/sass/publisher/course_detail.scss
```diff
@@ -65,6 +202,78 @@ $light-gray: rgba(204, 204, 204, 1);
...
+                .footer-about-copyright {
+                    width: 31.91489%;
+                    float: left;
+                    margin-right: 14.28571%;
+                    display: inline-block;
+                    vertical-align: middle;
+                }
+
+                .footer-about-openedx {
+                    float: right;
```
ahsan-ul-haq: use float(right)

course_discovery/static/sass/publisher/course_detail.scss
```diff
@@ -65,6 +202,78 @@ $light-gray: rgba(204, 204, 204, 1);
...
+            .footer-content-secondary {
+                margin-top: 20px;
+                display: table;
+                width: 100%;
+                clear: both;
+
+                .footer-about-copyright {
+                    width: 31.91489%;
+                    float: left;
+                    margin-right: 14.28571%;
```
ahsan-ul-haq: use `@include margin-right(14.28571%);`

course_discovery/static/sass/publisher/course_detail.scss
```diff
@@ -65,6 +202,78 @@ $light-gray: rgba(204, 204, 204, 1);
...
+                }
+
+                .nav-peripheral {
+                    width: 48.93617%;
+                    float: right;
+                    text-align: right;
+
+                    .nav-item {
+                        display: inline-block;
+                        margin-right: 20px;
```
ahsan-ul-haq: use `@include margin-right(20px);`

course_discovery/static/sass/publisher/course_detail.scss
```diff
@@ -65,6 +202,78 @@ $light-gray: rgba(204, 204, 204, 1);
...
+
+            .footer-content-secondary {
+                margin-top: 20px;
+                display: table;
+                width: 100%;
+                clear: both;
+
+                .footer-about-copyright {
+                    width: 31.91489%;
+                    float: left;
```
ahsan-ul-haq: use `@include float(left)`

course_discovery/templates/footer.html
```diff
@@ -1,6 +1,42 @@
...
-        Copyright &copy; {{ current_year }} {{ platform_name }}.
-    </p>
-</footer>
+{% load i18n %}
+<div class="wrapper-footer wrapper">
+    <footer class="primary">
+        <div class="footer-content-primary">
+          <div class="colophon">
+            {% now "Y" as current_year %}
+            <p>&copy; {{ current_year }} <a data-rel="edx.org" href="https://wwww.edx.org" rel="external">{{ platform_name }}</a>.</p>
```
ahsan-ul-haq: I don't think href should be hardcoded here.

bderusha: Can't hardcode `www.edx.org` here.  What if the PLATFORM is not edx?

bderusha (changes requested):
Code looks fine.  I have questions about the content.  How is this supposed to work for openedx? white label?

course_discovery/templates/footer.html
mattdrayer: @douglashall should take a look as well -- in the LMS code we have a setting called PLATFORM_NAME which we use to cover certain presentation scenarios.

douglashall: In LMS we have a settings override mechanism which makes use of a `SiteConfiguration` model which allows us to override settings like `PLATFORM_NAME`. We would like to get to a point where there is a python package that contains multitenant related utilities like this. We are not there yet. For the time being you could implement a similar settings override mechanism in this service or you could start the multitenant python package.

course_discovery/static/sass/publisher/course_detail.scss
clintonb: Why the use of `@include` for basic CSS properties?

course_discovery/templates/footer.html
mattdrayer: Does it make sense to leverage the Partner model here?  That's essentially the multi-tenant concept in this service:  https://github.com/edx/course-discovery/blob/master/course_discovery/apps/core/models.py#L59

douglashall: I'd prefer to stick with the Site/SiteConfiguration paradigm to solve the specific problem that is being raised here.

course_discovery/static/sass/publisher/course_detail.scss
ahsan-ul-haq: I think we need to provide rtl support if we don't need that now then its fine with me.

course_discovery/templates/footer.html
waheedahmed: @bderusha I thought it is an in house app only for edX.

course_discovery/static/sass/publisher/course_detail.scss
clintonb: Got it. I haven't dealt with RTL, so I'll defer to you.

waheedahmed:
@bderusha Removed the hardcoded edX links and also addressed other comments, please take another look.
Created a new story [ECOM-5959](https://openedx.atlassian.net/browse/ECOM-5959) to generate dynamic links using Site/SiteConfiguration or Partner model.

bderusha (approved):
Provided that we do the additional work to fix the links in https://openedx.atlassian.net/browse/ECOM-5959 this looks good to me

PR #371, opened 2016-10-06
ECOM-5961 Fix course ordering bug

MatthewPiatetsky:
@edx/ecommerce @rlucioni

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -370,7 +370,7 @@ class MinimalProgramCourseSerializerTests(TestCase):
             context={
                 'request': request,
                 'program': program,
-                'course_runs': program.course_runs
+                'course_runs': list(program.course_runs)
             }
         )
         expected = MinimalCourseSerializer(program.courses, many=True, context={'request': request}).data
@@ -399,7 +399,7 @@ class MinimalProgramCourseSerializerTests(TestCase):
         excluded_runs.append(course_runs[0])
         program = ProgramFactory(courses=[course], excluded_course_runs=excluded_runs)
 
-        serializer_context = {'request': request, 'program': program, 'course_runs': program.course_runs}
+        serializer_context = {'request': request, 'program': program, 'course_runs': list(program.course_runs)}
         serializer = MinimalProgramCourseSerializer(course, context=serializer_context)
 
         expected = MinimalCourseSerializer(course, context=serializer_context).data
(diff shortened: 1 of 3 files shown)


rlucioni:
@MatthewPiatetsky relatively minor, but please stick to the commit message convention we've been discussing:

```
Summary

Description

Ticket number
```

We shouldn't be relying on GitHub for storing this information. The context about the generator is also good to include in the commit message.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -519,6 +519,14 @@ def get_expected_data(self, program, request):
             'price_ranges': program.price_ranges,
             'subjects': SubjectSerializer(program.subjects, many=True).data,
             'transcript_languages': [serialize_language_to_code(l) for l in program.transcript_languages],
+            'courses': MinimalProgramCourseSerializer(
```
rlucioni: This doesn't need to be here, it's already included in the expected data by the superclass.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -670,9 +678,11 @@ def test_course_ordering_with_no_start(self):
 
     def test_data_without_course_sorting(self):
```
rlucioni, line 671: This test case is really "test_data_with_custom_course_sorting."

rlucioni (changes requested):
@MatthewPiatetsky I see you converting the generator to list repeatedly, but I don't see any new test cases that would have failed without your change. Am I missing something?

MatthewPiatetsky:
@rlucioni I had a bunch of test failures once I changed the serializer, the tests and the serializer were both using program.course_runs, so they both had the incorrect data, also didn't want others to accidentally get the wrong data because they were unintentionally using a generator

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: what is the difference?  
From the serializer point of view the regular data is just stored in sorted order

rlucioni: That's true. It's not a big deal, I just thought it was a little misleading; the data _is_ sorted, just not by the serializer.

rlucioni:
@MatthewPiatetsky that makes sense, thanks. So the serializer started working right, but the tests were still doing the wrong thing with the generator and failed.

rlucioni (commented):
Good with me once you remove the duplicated test data.

rlucioni (approved):
I meant to approve :|

PR #372, opened 2016-10-06
Exposing limit and offset query string parameters in API gateway

clintonb:
This will allow users to better filter the catalog data they receive.

ECOM-5749


clintonb:
FYI @efagin @schenedx

schenedx (approved):
👍

efagin:
👍 please link this to the api-manager PR once you've opened that.

PR #373, opened 2016-10-06
Removed user choice field from browseable catalog API pages

clintonb:
The viewers field attempts to render all users in the system. This does not scale and may bring down the system. This fix renders the field as a text input.

ECOM-5964


clintonb:
@edx/ecommerce please review

bderusha:
👍

rlucioni:
👍

rlucioni (approved):

PR #374, opened 2016-10-07
Updated i18n-tool version, .travis.yml and Makefile files

iivic:
@clintonb Review changes.


iivic:
Please review latest changes. @clintonb

Makefile
```diff
@@ -46,7 +48,7 @@ requirements.js:
 	npm install
 	$(NODE_BIN)/bower install
 
-requirements:
+requirements: requirements.js
```
clintonb: Revert. There may be workflows that depend on this.

.travis.yml
```diff
@@ -24,7 +24,6 @@ before_install:
 install:
     - pip install -U pip wheel codecov
     - make requirements
-    - make requirements.js
```
clintonb, line 27: Revert.

.travis.yml
```diff
@@ -35,9 +34,11 @@ script:
     - cd docs && make html
     - cd ..
 
+    # Check if translation files are up-to-date
+    - make detect_changed_source_translations
```
clintonb: `make validate_translations`

clintonb (changes requested):

iivic:
@clintonb I can't figure out why `make validate_translations` detect changes. Locally the source files are current. Can you take a look?

.travis.yml
```diff
@@ -35,9 +35,11 @@ script:
     - cd docs && make html
     - cd ..
 
+    # Check if translation files are up-to-date
+    - make validate_translations
+
     # Compile static assets and validate the code
     - make static -e DJANGO_SETTINGS_MODULE="course_discovery.settings.test"
     - make validate
-
```
clintonb, line 41: Revert.

clintonb (approved):
My fixup should resolve the build issue. Remember to squash before merging.

iivic:
Thank you for the help. So the problem were the files that shouldn't be extracted?

clintonb:
@iivic yes. We don't normally translate our documentation; although, we probably should consider doing so at some point.

PR #376, opened 2016-10-07
Removing ampersand HTML entities from ingested data

clintonb:
Data containing &amp; will now have that entity converted to &.

ECOM-5997


clintonb:
FYI @edx/ecommerce

rlucioni (approved):

PR #377, opened 2016-10-08
Implementing emails sending functionality.

awais786:
Adding Email functionality for newly added comments for course, course run and seats.
Each course has associated group. Email will be send to all groups members except those who disabled email notifications in attribute table. 
Waffle switch is added to disable the email functionality.

https://openedx.atlassian.net/browse/ECOM-5929


awais786:
@tasawernawaz @ahsan-ul-haq @waheedahmed @jareerahsan Please review.

course_discovery/apps/publisher/models.py
```diff
@@ -156,6 +156,24 @@ def get_group_institution(self):
         available_groups = get_groups_with_perms(self)
         return available_groups[0] if available_groups else None
 
+    @property
+    def get_group_users(self):
+        """ Returns queryset of all User objects with any object permissions for the given course."""
```
ahsan-ul-haq: doc string should be less than 80 characters per line

course_discovery/apps/publisher/models.py
```diff
@@ -156,6 +156,24 @@ def get_group_institution(self):
...
+        """ Returns the list of users emails with enable email notifications against
+        a course group. By default if attribute value does not exists then user will be eligible
+        for emails.
+        """
+        users_list = self.get_group_users
+        emails = [
+            user.email for user in users_list
+            if hasattr(user, 'attributes') and user.attributes.enable_email_notification or
+            not hasattr(user, 'attributes')
+        ]
```
ahsan-ul-haq, line 171: you could do that like this as well

```
if not hasattr(user, 'attributes') or user.attributes.enable_email_notification
```

It will check if there is no user attribute than add in emails otherwise if there is user attribute then check if email notification is enabled

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,59 @@
+import logging
+from django.conf import settings
+from django.core.mail import get_connection
```
ahsan-ul-haq: django imports should be in 2nd block

course_discovery/templates/publisher/email/comment.html
```diff
@@ -0,0 +1,18 @@
+{% extends "publisher/email/email_base.html" %}
+{% load i18n %}
+{% block body %}
+<!-- Message Body -->
+<tr>
+    <td class="container-padding" bgcolor="#ffffff" style="background-color: #ffffff; padding-left: 10px; padding-right: 30px; font-size: 13px; line-height: 20px; font-family: Open Sans, sans-serif; color: #333; border-radius: 5px;" align="left">
+        <!--/message HTML content -->
+        <p>{% trans "Hi," %}</p>
+        <p>{{ comment }}</p>
```
ahsan-ul-haq: Don't we need to translate comment as well?

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,59 @@
...
+            'comment': comment.comment
+        }
+        txt_template = 'publisher/email/comment.txt'
+        html_template = 'publisher/email/comment.html'
+
+        template = get_template(txt_template)
+        plain_content = template.render(context)
+        template = get_template(html_template)
+        html_content = template.render(context)
+        subject = _("New comment added in {title}'s {model_name}.").format(
```
ahsan-ul-haq: I don't think it would be right to use model name here

ahsan-ul-haq (changes requested):

course_discovery/apps/publisher_comments/emails.py
awais786: New comment added in Computer Science's Seat. So it looks clear comment is added in course's seat.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -62,6 +63,14 @@ def setUp(self):
         self.course = factories.CourseFactory()
         self.course2 = factories.CourseFactory()
 
+        self.user1 = UserFactory(email='test@example.com')
+        self.user2 = UserFactory(email='test1@example.com')
+        self.user3 = UserFactory(email='test3@example.com')
+        self.group_a = factories.GroupFactory(name="Test Group A")
```
ahsan-ul-haq: either use `"` or `'`

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,121 @@
...
+        assign_perm(Course.VIEW_PERMISSION, self.group, self.course)
+
+        UserAttributeFactory(user=self.user, enable_email_notification=True)
+        UserAttributeFactory(user=self.user_3, enable_email_notification=False)
+        # For user_2 user-attribute is not created. So he is eligible for mails.
+
+        toggle_switch('enable_emails', True)
+
+    def test_course_comment_email(self):
+        """ Verify that after adding a comment against a course emails send to multiple users
```
ahsan-ul-haq: limit doc string 80 characters per line

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,121 @@
...
+                title=title,
+                model_name=content_object.__class__.__name__
+            )
+        )
+        self.assertIn(test_comment, mail.outbox[0].body.strip())
+        self.assertIn('Hi', mail.outbox[0].body.strip())
+        self.assertIn('Thanks', mail.outbox[0].body.strip())
+        self.assertIn('The edX team', mail.outbox[0].body.strip())
+
+    def test_email_without_different_group(self):
```
ahsan-ul-haq, line 90: add tests before methods

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,121 @@
...
+        setup_email_for_comment.assert_called_once_with(comment)
+
+    @mock.patch('course_discovery.apps.publisher_comments.models.setup_email_for_comment')
+    def test_email_without_waffle_switch(self, setup_email_for_comment):
+        """ Verify that setup_email_for_comment not called with disable waffle switch.. """
+        toggle_switch('enable_emails', False)
+        self._generate_comment(self.course, self.user, 'test')
+        setup_email_for_comment.assert_not_called()
+
+    def _generate_comment(self, content_object, user, comment):
```
ahsan-ul-haq: Why not using factory directly instead of creating this method

course_discovery/settings/base.py
```diff
@@ -375,6 +375,7 @@
 SOLO_CACHE = 'default'
 SOLO_CACHE_TIMEOUT = 3600
 
+PUBLISHER_FROM_EMAIL = 'test@example.com'
```
ahsan-ul-haq: This should be added in test settings.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,121 @@
+# pylint: disable=no-member
+import mock
+import ddt
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.test import TestCase
+from django.core import mail
+from guardian.shortcuts import assign_perm
```
ahsan-ul-haq, line 8: Alphabetize

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -7,6 +7,7 @@
 from course_discovery.apps.core.tests.factories import UserFactory
 from course_discovery.apps.publisher.models import State, Course
 from course_discovery.apps.publisher.tests import factories
+from course_discovery.apps.publisher.tests.factories import UserAttributeFactory
```
ahsan-ul-haq: If `factories` is imported a line above then why importing `UserAttributeFactory` here?

ahsan-ul-haq (changes requested):

course_discovery/apps/publisher/tests/test_model.py
awais786: UserFactory is coming from core but UserAttributeFactory belongs to publisher.

course_discovery/settings/base.py
awais786: This file is importing in test.py. For production I will create the configuration PR for secure vars.

course_discovery/apps/publisher/tests/test_model.py
ahsan-ul-haq: I am talking about `from course_discovery.apps.publisher.tests import factories` one line above this

awais786:
@rlucioni @bderusha  Please review this PR. If you guys are busy please let me know I will tag other people in cambridge.

rlucioni:
@awais786 I'm not able to look at this right now. Please tag @edx/ecommerce when looking for a review from the Cambridge team.

awais786:

ahsan-ul-haq (approved):

awais786:

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def setup_email_for_comment(comment):
+    """ Send the emails for a comment.
+
+        Arguments:
+            comment(Generic Content Type): Comment Object
+    """
+    try:
+        connection = get_connection()
```
clintonb: This is not necessary. Calling `email_msg.send()` will use the default backend.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+def setup_email_for_comment(comment):
+    """ Send the emails for a comment.
+
+        Arguments:
+            comment(Generic Content Type): Comment Object
```
clintonb: `comment (Comment): Comment object`

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    try:
+        connection = get_connection()
+        connection.open()
+
+        object_pk = comment.object_pk
+        publisher_obj = comment.content_type.get_object_for_this_type(pk=object_pk)
+        if comment.content_type.model_class() == CourseRun:
```
clintonb: Nit: Add a newline above this to clean this up a bit.

clintonb: Also, create a variable for `comment.content_type.model_class()` instead of looking it up twice.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,60 @@
...
+            title=course.title,
+            model_name=comment.content_type.model_class().__name__
+        )
+        email_msg = EmailMultiAlternatives(
+            subject, plain_content, from_address, to_addresses, connection=connection
+        )
+        email_msg.attach_alternative(html_content, "text/html")
+        email_msg.send()
+    except Exception as e:  # pylint: disable=broad-except
+        log.error(str(e))
```
clintonb: Use `log.exception`.

course_discovery/apps/publisher_comments/models.py
```diff
@@ -1,8 +1,20 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
 from django.utils.translation import ugettext_lazy as _
-
 from django_comments.models import CommentAbstractModel
 from django_extensions.db.fields import ModificationDateTimeField
```
clintonb: Remove this newline.

course_discovery/apps/publisher_comments/models.py
```diff
@@ -1,8 +1,20 @@
...
 
+import waffle
+
+from course_discovery.apps.publisher_comments.emails import setup_email_for_comment
+
 
 class Comments(CommentAbstractModel):
```
clintonb, line 11: `Comment` (singular)

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+from course_discovery.apps.course_metadata.tests import toggle_switch
+from course_discovery.apps.publisher.models import Course
+from course_discovery.apps.publisher.tests import factories
+from course_discovery.apps.publisher.tests.factories import UserAttributeFactory
+from course_discovery.apps.publisher_comments.tests.factories import CommentFactory
+
+
+@ddt.ddt
+class CommentsEmailTests(TestCase):
+    """ Tests for the Email functionality for course, course-run
```
clintonb: "e-mail". Move this to one line.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+    and seats.
+    """
+
+    def setUp(self):
+        super(CommentsEmailTests, self).setUp()
+
+        self.user = UserFactory(email='test@test-edx.org')
```
clintonb: If you aren't using the custom email address, don't bother adding it to the factory. Doing so is confusing.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self.site = Site.objects.get(pk=settings.SITE_ID)
+
+        self.group = factories.GroupFactory(name='abc')
+
+        self.user.groups.add(self.group)
+        self.user_2.groups.add(self.group)
+        self.user_3.groups.add(self.group)
+
+        self.course = factories.CourseFactory()
```
clintonb:
``` python
self.seat = factories.SeatFactory()
self.course_run = self.seat.course_run
self.course = self.course_run.course
```

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+        self.course_run = factories.CourseRunFactory(course=self.course)
+        self.seat = factories.SeatFactory(course_run=self.course_run)
+
+        assign_perm(Course.VIEW_PERMISSION, self.group, self.course)
+
+        UserAttributeFactory(user=self.user, enable_email_notification=True)
+        UserAttributeFactory(user=self.user_3, enable_email_notification=False)
+        # For user_2 user-attribute is not created. So he is eligible for mails.
```
clintonb: Move this note above the two factory calls.

``` python
# NOTE: We intentionally do NOT create an attribute for user_2. By default this user WILL receive email notifications.
```

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+            content_object=self.course,
+            user=self.user,
+            site=self.site
+        )
+        self.assertEqual(len(mail.outbox), 0)
+
+    def _assert_data(self, content_object, test_comment, title):
+        """ DRY method to assert send email data"""
+        self.assertEqual([self.user.email, self.user_2.email], mail.outbox[0].to)
+        self.assertEqual(len(mail.outbox[0].to), 2)
```
clintonb: This is not necessary given the assertion above.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+        test_comment = 'This comment is for course.'
+        CommentFactory(
+            comment=test_comment,
+            content_object=self.course,
+            user=self.user,
+            site=self.site
+        )
+        self.assertEqual(len(mail.outbox), 0)
+
+    def _assert_data(self, content_object, test_comment, title):
```
clintonb: `assert_comment_email_sent`

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+        self.assertEqual([self.user.email, self.user_2.email], mail.outbox[0].to)
+        self.assertEqual(len(mail.outbox[0].to), 2)
+        self.assertEqual(
+            str(mail.outbox[0].subject),
+            "New comment added in {title}'s {model_name}.".format(
+                title=title,
+                model_name=content_object.__class__.__name__
+            )
+        )
+        self.assertIn(test_comment, mail.outbox[0].body.strip())
```
clintonb: Make a variable for the body.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        comment = 'This comment is for course run.'
+        CommentFactory(
+            comment=comment,
+            content_object=self.course_run,
+            user=self.user,
+            site=self.site
+        )
+        self._assert_data(
+            self.course_run, comment, self.course.title
```
clintonb: One line.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+            site=self.site
+        )
+        setup_email_for_comment.assert_called_once_with(comment)
+
+    @mock.patch('course_discovery.apps.publisher_comments.models.setup_email_for_comment')
+    def test_email_with_disable_waffle_switch(self, setup_email_for_comment):
+        """ Verify that setup_email_for_comment not called with disable waffle switch.. """
+        toggle_switch('enable_emails', False)
+        CommentFactory(
+            comment='test',
```
clintonb: Remove. This value is never used.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+
+        toggle_switch('enable_emails', True)
+
+    def test_course_comment_email(self):
+        """ Verify that after adding a comment against a course emails send
+        to multiple users depending upon the course related group.
+        """
+        test_comment = 'This comment is for course.'
+        CommentFactory(
+            comment=test_comment,
```
clintonb: Remove. Use the factory-generated text.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        self._assert_data(self.course, test_comment, self.course.title)
+
+    def test_course_run_comment_email(self):
+        """ Verify that after adding a comment against a course-run emails send to multiple users
+        depending upon the parent course related group.
+        """
+        comment = 'This comment is for course run.'
+        CommentFactory(
+            comment=comment,
```
clintonb: Remove. Use the factory-generated text.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+            self.course_run, comment, self.course.title
+        )
+
+    def test_seat_comment_email(self):
+        """ Verify that after adding a comment against a seat emails send to multiple users
+        depending upon the parent course related group.
+        """
+        comment = 'This comment is for seat.'
+        CommentFactory(
+            comment=comment,
```
clintonb: Remove. Use the factory-generated text.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+            user=self.user,
+            site=self.site
+        )
+        self._assert_data(self.seat, comment, self.course.title)
+
+    @mock.patch('course_discovery.apps.publisher_comments.models.setup_email_for_comment')
+    def test_email_with_enable_waffle_switch(self, setup_email_for_comment):
+        """ Verify that setup_email_for_comment called with enable waffle switch.. """
+        comment = CommentFactory(
+            comment='test',
```
clintonb: Remove. This value is never used.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def test_email_without_different_group(self):
+        """ Verify the emails behaviour if course group has no users. """
+        self.user.groups.remove(self.group)
+        self.user_2.groups.remove(self.group)
+        self.user_3.groups.remove(self.group)
+
+        test_comment = 'This comment is for course.'
+        CommentFactory(
+            comment=test_comment,
```
clintonb: Remove. Use the factory-generated text.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,144 @@
...
+        # For user_2 user-attribute is not created. So he is eligible for mails.
+
+        toggle_switch('enable_emails', True)
+
+    def test_course_comment_email(self):
+        """ Verify that after adding a comment against a course emails send
+        to multiple users depending upon the course related group.
+        """
+        test_comment = 'This comment is for course.'
+        CommentFactory(
```
clintonb: This call is replicated a lot. Make a helper method:

``` python
def create_comment(self, content_object):
  return CommentFactory(content_object=content_object, user=self.user, site=self.site)
```

course_discovery/apps/publisher_comments/tests/test_views.py
```diff
@@ -15,16 +17,22 @@ class CommentsTests(TestCase):
...
         self.site = Site.objects.get(pk=settings.SITE_ID)
         self.course_edit_page = 'publisher:publisher_courses_edit'
         self.course_run_edit_page = 'publisher:publisher_course_runs_edit'
         self.seat_edit_page = 'publisher:publisher_seats_edit'
         self.edit_comment_page = 'publisher_comments:comment_edit'
         self.course = factories.CourseFactory()
-        self.course_run = factories.CourseRunFactory()
-        self.seat = factories.SeatFactory(type=Seat.PROFESSIONAL, credit_hours=0)
+        self.course_run = factories.CourseRunFactory(course=self.course)
+        self.seat = factories.SeatFactory(course_run=self.course_run, type=Seat.PROFESSIONAL, credit_hours=0)
```
clintonb: Start with the `Seat`, and the factory will make the `Course` and `CourseRun`:

``` python
self.seat = factories.SeatFactory(type=Seat.PROFESSIONAL, credit_hours=0)
self.course_run = self.seat.course_run
self.course = self.course_run.course
```

course_discovery/apps/publisher_comments/tests/test_views.py
```diff
@@ -129,6 +123,9 @@ def _edit_comment_page(self, content_object, expected_url):
         response = self.client.get(comment_url)
         self.assertContains(response, new_comment)
 
+        # mail has 2 emails one due to newly added comment and other is due to editing.
+        self.assertEqual(len(mail.outbox), 2)
```
clintonb: This is a random assertion. Make a separate test for it.

course_discovery/apps/publisher_comments/tests/test_views.py
```diff
@@ -150,3 +147,21 @@ def _generate_comments_for_all_content_types(self):
...
 
         return data
+
+    def _add_assert_multiple_comments(self, content_object, page_path):
+        """ DRY method to add comments on edit page for specific object. """
+        response = self.client.get(reverse(page_path, kwargs={'pk': content_object.id}))
+        self.assertContains(response, 'Total Comments 0')
+
+        comments = []
+        for num in range(1, 10):    # pylint: disable=unused-variable
```
clintonb: Why 10? Why not 2?

course_discovery/settings/base.py
```diff
@@ -375,6 +375,7 @@
 SOLO_CACHE = 'default'
 SOLO_CACHE_TIMEOUT = 3600
 
+PUBLISHER_FROM_EMAIL = 'SET-ME-PLEASE'
```
clintonb: If someone were to enable emails with this value set, what would happen? If the value is set to `None` will it fail catastrophically (e.g. raised exception)? It would be great if we failed in a way that raised an error, so that issues are not squashed.

course_discovery/templates/publisher/email/comment.html
clintonb: Nope.

course_discovery/templates/publisher/email/comment.html
```diff
@@ -0,0 +1,18 @@
+{% extends "publisher/email/email_base.html" %}
+{% load i18n %}
+{% block body %}
+<!-- Message Body -->
+<tr>
+    <td class="container-padding" bgcolor="#ffffff" style="background-color: #ffffff; padding-left: 10px; padding-right: 30px; font-size: 13px; line-height: 20px; font-family: Open Sans, sans-serif; color: #333; border-radius: 5px;" align="left">
+        <!--/message HTML content -->
+        <p>{% trans "Hi," %}</p>
```
clintonb: Let's make this more useful:

``` html
<p>{{ comment.user.first_name }} commented on a {{ object_type }} belonging to the course {{ course.title }} ({{ course.key }}).
</p>

<p>
{{ comment.comment }}
</p>

<p>
<a href="">View comment</a>
</p>
```

course_discovery/templates/publisher/email/comment.txt
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
```
clintonb: See comment about the HTML version. Make this message actionable.

course_discovery/templates/publisher/email/email_base.html
```diff
@@ -0,0 +1,24 @@
...
+    <meta name="viewport" content="initial-scale=1.0">    <!-- So that mobile webkit will display zoomed in -->
+    <meta name="format-detection" content="telephone=no"> <!-- disable auto telephone linking in iOS -->
+</head>
+<body style="font-family:Arial,'Helvetica Neue',Helvetica,sans-serif;font-size:14px;line-height:150%;margin:auto">
+
+<br>
+<table border="0" width="100%" height="100%" cellpadding="0" cellspacing="0" style="padding: 5px;">
+    <tr>
+        <td align="" valign="top">
+            <table>
```
clintonb: Why are you nesting tables?

clintonb (changes requested):

course_discovery/settings/base.py
awais786: With None it will raise following error.
`Message: 'SESAddressNotVerifiedError: 400 Email address is not verified.\n<ErrorResponse xmlns="http://ses.amazonaws.com/doc/2010-12-01/">\n  <Error>\n    <Type>Sender</Type>`

awais786: I have set it to None in base.py.

awais786: In-fact without proper verified email address it will raise the error.

course_discovery/templates/publisher/email/comment.txt
```diff
@@ -0,0 +1,15 @@
...
+{{ first_name }} commented on a {{ object_type }} belonging to the course {{ title }} ({{ number }}).
+{% endblocktrans %}
+
+{{ comment.comment }}
+
+{% trans "To view the detail page, visit the following link." %}
+{{ page_url }}
+
+
+{% trans "Thanks," %}
```
clintonb: Drop this. There is nothing to thank anyone for.

course_discovery/templates/publisher/email/comment.txt
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+
+{% blocktrans with first_name=comment.user.first_name object_type=object_type|lower title=course.title number=course.number %}
+{{ first_name }} commented on a {{ object_type }} belonging to the course {{ title }} ({{ number }}).
+{% endblocktrans %}
+
+{{ comment.comment }}
+
+{% trans "To view the detail page, visit the following link." %}
```
clintonb:
```
{% trans "View comment:" %} {{ page_url }}
```

course_discovery/templates/publisher/email/comment.html
```diff
@@ -0,0 +1,21 @@
...
+            {% endblocktrans %}
+        </p>
+        <p>
+            {{ comment.comment }}
+        </p>
+
+        <p>
+            <a href="{{ page_url }}">{% trans "View comment" %}</a>
+        </p>
+        <p>{% trans "Thanks," %}</p>
```
clintonb: Drop.

course_discovery/settings/test.py
```diff
@@ -23,3 +23,6 @@
 JWT_AUTH['JWT_SECRET_KEY'] = 'course-discovery-jwt-secret-key'
 
 LOGGING['handlers']['local'] = {'class': 'logging.NullHandler'}
+
+PUBLISHER_FROM_EMAIL = 'test@example.com'
+DISCOVERY_URL_ROOT = 'http://test/'
```
clintonb: This is not multi-tenant-friendly. The host should come from the `Site` model.

course_discovery/settings/base.py
```diff
@@ -375,6 +375,7 @@
 SOLO_CACHE = 'default'
 SOLO_CACHE_TIMEOUT = 3600
 
+PUBLISHER_FROM_EMAIL = None
```
clintonb, line 382: I noticed there are no docs for the Publisher app in the `docs/` directory. This is a good time to initiate that documentation and document this setting, along with the general email functionality.

course_discovery/apps/publisher_comments/tests/test_views.py
```diff
@@ -150,3 +148,21 @@ def _generate_comments_for_all_content_types(self):
...
 
         return data
+
+    def _add_assert_multiple_comments(self, content_object, page_path):
+        """ DRY method to add comments on edit page for specific object. """
+        response = self.client.get(reverse(page_path, kwargs={'pk': content_object.id}))
+        self.assertContains(response, 'Total Comments 0')
+
+        comments = []
+        for num in range(1, 2):    # pylint: disable=unused-variable
```
clintonb: If you are going to use the variable use `__`.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,132 @@
+# pylint: disable=no-member
+from urllib.parse import urljoin
+
+import ddt
+import mock
+
```
clintonb: Third-party imports should be grouped together.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,132 @@
+# pylint: disable=no-member
```
clintonb: What is this for? Disabling Pylint for an entire file is rarely needed.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,64 @@
...
+            title=course.title,
+            model_name=comment_class.__name__
+        )
+        email_msg = EmailMultiAlternatives(
+            subject, plain_content, from_address, to_addresses
+        )
+        email_msg.attach_alternative(html_content, "text/html")
+        email_msg.send()
+    except Exception as e:  # pylint: disable=broad-except
+        log.exception(str(e))
```
clintonb: See the docs at https://docs.python.org/3/library/logging.html#logging.Logger.exception. You don't need to touch the exception. You need to log a useful message.

``` python
try:
 ...
except Exception:  # pylint: disable=broad-except
  log.exception('Failed to send email notifications for comment %s', comment.id)
```

course_discovery/apps/publisher_comments/emails.py
clintonb: 1. Single quotes.
2. Remove the period at the end of the subject line.

Check with the PMs to determine if having the course title is more useful than the course ID, or both.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,64 @@
...
+            comment (Comment): Comment object
+    """
+    try:
+        object_pk = comment.object_pk
+        publisher_obj = comment.content_type.get_object_for_this_type(pk=object_pk)
+        comment_class = comment.content_type.model_class()
+
+        if comment_class == CourseRun:
+            course = publisher_obj.course
+            page_url = reverse('publisher:publisher_course_run_detail', args=[publisher_obj.id])
```
clintonb: `object_path` is a more accurate variable name.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,64 @@
...
+        elif comment_class == Seat:
+            course = publisher_obj.course_run.course
+            page_url = reverse('publisher:publisher_seats_edit', args=[publisher_obj.id])
```
clintonb: Why does this point to the edit page instead of the detail page?

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,64 @@
...
+            page_url = reverse('publisher:publisher_courses_edit', args=[publisher_obj.id])
+
+        to_addresses = course.get_group_users_emails()
+        from_address = settings.PUBLISHER_FROM_EMAIL
+
+        context = {
+            'comment': comment,
+            'course': course,
+            'object_type': comment_class.__name__,
+            'page_url': urljoin(settings.DISCOVERY_URL_ROOT, page_url)
```
clintonb:
``` python
'https://{host}{path}'.format(host=comment.site.domain.strip('/'), path=object_path)
```

course_discovery/apps/publisher/urls.py
```diff
@@ -12,7 +12,6 @@
     url(r'^course_runs/(?P<pk>\d+)/$', views.CourseRunDetailView.as_view(), name='publisher_course_run_detail'),
     url(r'^course_runs/$', views.CourseRunListView.as_view(), name='publisher_course_runs'),
     url(r'^course_runs/new$', views.CreateCourseRunView.as_view(), name='publisher_course_runs_new'),
-    url(r'^course_runs/(?P<pk>\d+)/$', views.CourseRunDetailView.as_view(), name='publisher_course_run_detail'),
```
clintonb, line 15: Now would be a great time to group/namespace these URLs. Waiting until later will be painful. See https://github.com/edx/edx-analytics-dashboard/blob/master/analytics_dashboard/courses/urls.py for an example of grouping. This doesn't have to be done in this PR, but it should happen soon.

course_discovery/apps/publisher/urls.py
```diff
@@ -12,7 +12,6 @@
     url(r'^course_runs/(?P<pk>\d+)/$', views.CourseRunDetailView.as_view(), name='publisher_course_run_detail'),
     url(r'^course_runs/$', views.CourseRunListView.as_view(), name='publisher_course_runs'),
     url(r'^course_runs/new$', views.CreateCourseRunView.as_view(), name='publisher_course_runs_new'),
```
clintonb, line 14: Use a trailing slash for all URLs. (Not a prerequisite for merging this PR.)

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -73,24 +81,62 @@ def test_post_back_url(self):
...
+        self._assert_false_permissions(self.user2, self.course2)
+        self._assign_assert()
+
+    def test_assign_permission_by_group(self):
+        """ Verify that permission can be assigned using the group. """
+        self._assert_false_permissions(self.user1, self.course)
+        self._assert_false_permissions(self.user2, self.course2)
+        self._assign_assert()
+
+    def _assert_false_permissions(self, user, course):
```
clintonb:
``` python
def assert_user_cannot_view_course(self, user, course):
  """ Asserts the user can NOT view the course. """
  ...
```

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -73,24 +81,62 @@ def test_post_back_url(self):
...
+        # DRY method to assert user has no permission on course
+        self.assertFalse(user.has_perm(Course.VIEW_PERMISSION, course))
+
+    def _assert_true_permissions(self, user, course):
```
clintonb:
``` python
def assert_user_can_view_course(self, user, course):
  """ Asserts the user can view the course. """
  ...
```

course_discovery/apps/publisher/models.py
```diff
@@ -156,6 +156,25 @@ def get_group_institution(self):
         available_groups = get_groups_with_perms(self)
         return available_groups[0] if available_groups else None
```
clintonb, line 160: Why only one group? It's not clear what this method is doing. Also, a property name should not be prefixed with `get_` or `set_`.

clintonb: Additionally, `assign_user_groups` above is only used in tests. It should be removed.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -73,24 +81,62 @@ def test_post_back_url(self):
...
+        # DRY method to assert user has permission on course
+        self.assertTrue(user.has_perm(Course.VIEW_PERMISSION, course))
+
+    def test_get_group_institution(self):
+        """ Verify that method returns the institution against the course."""
```
clintonb: Please clarify this docstring, and the one for the method being tested. It is not clear what the method is returning.

clintonb (changes requested):

course_discovery/apps/publisher_comments/emails.py
awais786: I did that to see the exact message. I have updated this now.

course_discovery/apps/publisher/urls.py
awais786: I have created the ticket for this https://openedx.atlassian.net/browse/ECOM-6109.

course_discovery/apps/publisher_comments/models.py
awais786: I will do this change in other story [ECOM-6110](https://openedx.atlassian.net/browse/ECOM-6110). Otherwise this PR will become too large.

course_discovery/apps/publisher_comments/emails.py
awais786: At the moment all pages are in re-designing phase. Once we implement all pages then we will update these links accordingly.

course_discovery/apps/publisher/mixins.py
```diff
@@ -41,17 +41,13 @@ def dispatch(self, request, *args, **kwargs):
...
-    assign_user_groups = False
 
     def form_valid(self, form):
         user = self.request.user
         publisher_object = form.save(commit=False)
         publisher_object.changed_by = user
         publisher_object.save()
 
-        if self.assign_user_groups:
-            publisher_object.assign_user_groups(user)
```
clintonb, line 53: If this is the same method as below, I was wrong about it not being used, and it should remain.

course_discovery/apps/publisher/models.py
clintonb: This docstring is much better, but now I'm confused as to why you're only returning the first group. Clearly an assumption is being made about group ordering. What is that assumption? How can we go from assumption to certainty?

course_discovery/apps/publisher/mixins.py
awais786: I have remove this method. Because previously we were assigning permissions according to the request.user's group. But now implementation has been changed and author will choose the group/institution from the dropdown.

course_discovery/apps/publisher/models.py
awais786: At the moment course will belong to the one institution only. 
And course creation page has the the dropdown to select the group. According to the group/institution only that group users can view that object.  So its not possible that course will belong with multiple groups. 

@bderusha FYI

awais786:
@clintonb please review this PR again.

course_discovery/apps/publisher/models.py
clintonb: I courses can only be associated with one institution that should be fully enforced, or you should fully support multiple institutions in the backend and enforce the single institution at the front-end level. Besides being confusing to later developers, I can see this leading to confusion and errors in the future due to the single group assumption only being enforced here (and not informing the user/developer of the error when they attempt to add another institution).

course_discovery/apps/publisher/models.py
```diff
@@ -143,19 +143,37 @@ class Meta(TimeStampedModel.Meta):
             ('view_course', 'Can view course'),
         )
 
-    def assign_user_groups(self, user):
-        for group in user.groups.all():
-            assign_perm(self.VIEW_PERMISSION, group, self)
-
     def assign_permission_by_group(self, institution):
```
clintonb, line 150: Unrelated: why is this a public method? I would expect this to be a private method called only only when modifying institutions. See https://docs.djangoproject.com/en/1.10/ref/signals/#django.db.models.signals.m2m_changed.

course_discovery/apps/publisher/models.py
```diff
@@ -143,19 +143,37 @@ class Meta(TimeStampedModel.Meta):
...
     def get_group_institution(self):
-        """ Returns the Group object with for the given course object. """
+        """ Returns the group object having permissions on the given course.
+        Course will be associated with one group only.
+        """
         available_groups = get_groups_with_perms(self)
         return available_groups[0] if available_groups else None
 
+    @property
+    def get_group_users(self):
```
clintonb: This is only used in the method below. This doesn't need exist, and it's misnamed—properties should not be prefixed with `get_` or `set_`.

clintonb (changes requested):
@awais786 you've asked for another review; but, you haven't addressed all of the comments from the earlier review. Please address all comments—if you aren't going to accept the recommended change, just provide a reason why not—before requesting additional reviews.

course_discovery/apps/publisher/urls.py
awais786: I have created the seperate ticket for this change.
https://openedx.atlassian.net/browse/ECOM-6109

course_discovery/apps/publisher/models.py
awais786: I did not understand the purpose of m2m_changed here. 
[institution](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/forms.py#L49) is form field.

clintonb: Institution is a form field that ultimately sets a value on the model. If that is the field that truly determines which institutions can access the course, you should be triggering permissions based on changes to that field. For example, say you need to modify the institution via Django admin. Your current implementation will not support changing permissions in that case.

If you implement permissions at the model level you don't have to worry about any of these issues, because you've implemented at the lowest possible level.

awais786: And institution value is not saving in db its just use to assign the permission on the course.

clintonb: That does not seem ideal, and opens up interesting permissions issues. For example, say you have a course owned by MIT. If permissions are not directly linked to the value of the institution field, you could end up in a situation where the MIT-owned course is _not_ visible to MIT users or made visible to non-MIT institution users.

awais786: For this scenario [django-guardian->get-groups-with-perms](http://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-groups-with-perms)  provides us the group against the course. But I agree with you this value should be save in db. This is what we did earlier but later dropped the field after some discussion. I am adding @bderusha here for his thoughts and if all agrees here I will create the separate ticket to store  the group against a course and it should be OneToOneField because each course can associated with one group only Or It can be manytomany as well if in future multiple groups can be assigned to one course.

course_discovery/settings/base.py
awais786: Ticket for documentation https://openedx.atlassian.net/browse/ECOM-6139

clintonb: Umm...this hasn't yet merged. The documentation should be part of this commit. We've gotten quite lax with our documentation practices over the last few months. A ticket is nice, but I can easily see that ticket being ignored for "higher priority" work in the future.

course_discovery/templates/publisher/email/comment.txt
awais786: Done according to the HTML.

course_discovery/settings/base.py
awais786: Fine I will add the documentation in this commit.

awais786: I have added the documentation part here. I just add one setting variable here because we have already defined all variables in private.py.

course_discovery/apps/publisher/models.py
awais786: I have created a separate ticket for this purpose.
https://openedx.atlassian.net/browse/ECOM-6142

course_discovery/apps/publisher/models.py
awais786: I have created a separate ticket for this purpose.
https://openedx.atlassian.net/browse/ECOM-6142

awais786:
@clintonb I have double checked to make sure all requested changes are fixed. I have created few tickets to fix all things separately.  I have added translations files and initial documentation file also.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -62,6 +62,14 @@ def setUp(self):
         self.course = factories.CourseFactory()
         self.course2 = factories.CourseFactory()
 
+        self.user1 = UserFactory(email='test@example.com')
```
clintonb: Given that none of these names matter to your tests, they should be removed. It's extraneous information.

awais786:
@clintonb I have double checked to make sure all requested changes are fixed. I have created few tickets to fix all things separately. I have added translations files and initial documentation file also.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -72,25 +80,50 @@ def test_post_back_url(self):
...
-        self.assertTrue(user1.has_perm(Course.VIEW_PERMISSION, self.course))
-        self.assertTrue(user2.has_perm(Course.VIEW_PERMISSION, self.course2))
-
-        self.assertFalse(user1.has_perm(Course.VIEW_PERMISSION, self.course2))
-        self.assertFalse(user2.has_perm(Course.VIEW_PERMISSION, self.course))
+    def test_assign_permission_by_group(self):
+        """ Verify that permission can be assigned using the group. """
+        self.assert_user_cannot_view_course(self.user1, self.course)
+        self.assert_user_cannot_view_course(self.user2, self.course2)
+        self._assign_assert()
```
clintonb: This is called once. It doesn't need to be a separate method.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    try:
+        object_pk = comment.object_pk
+        publisher_obj = comment.content_type.get_object_for_this_type(pk=object_pk)
+        comment_class = comment.content_type.model_class()
+
+        if comment_class == CourseRun:
+            course = publisher_obj.course
+            object_path = reverse('publisher:publisher_course_run_detail', args=[publisher_obj.id])
+            subject = 'New comment added in Course-run : {title}-{pacing_type}-{start}'.format(
```
clintonb: Mark for translation.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,70 @@
...
+                title=course.title,
+                pacing_type=publisher_obj.get_pacing_type_display(),
+                start=publisher_obj.start.strftime("%B %d, %Y")
+            )
+        elif comment_class == Seat:
+            course = publisher_obj.course_run.course
+            object_path = reverse('publisher:publisher_seats_edit', args=[publisher_obj.id])
+            subject = 'New comment added in Seat : {title}-{pacing_type}-{start}-{type}'.format(
```
clintonb: Mark for translation.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,70 @@
...
+                title=course.title,
+                pacing_type=publisher_obj.course_run.get_pacing_type_display(),
+                start=publisher_obj.course_run.start.strftime("%B %d, %Y"),
+                type=publisher_obj.type
+            )
+        else:
+            course = publisher_obj
+            object_path = reverse('publisher:publisher_courses_edit', args=[publisher_obj.id])
+            subject = 'New comment added in Course : {title}'.format(title=course.title)
```
clintonb: Mark for translation.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,70 @@
...
+        html_template = 'publisher/email/comment.html'
+
+        template = get_template(txt_template)
+        plain_content = template.render(context)
+        template = get_template(html_template)
+        html_content = template.render(context)
+        email_msg = EmailMultiAlternatives(
+            subject, plain_content, from_address, to_addresses
+        )
+        email_msg.attach_alternative(html_content, "text/html")
```
clintonb: Single quotes.

clintonb (commented):
I am headed out on vacation. Someone else from @edx/ecommerce needs to finish reviewing.

awais786:
@edx/ecommerce Please review this PR. All requested changes are fixed.

awais786:
@edx/ecommerce Please review this PR. @clintonb already reviewed this PR but it needs final review.

awais786:

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if comment_class == CourseRun:
+            course = publisher_obj.course
+            object_path = reverse('publisher:publisher_course_run_detail', args=[publisher_obj.id])
+            subject = _('New comment added in Course-run : {title}-{pacing_type}-{start}').format(
+                title=course.title,
+                pacing_type=publisher_obj.get_pacing_type_display(),
+                start=publisher_obj.start.strftime('%B %d, %Y') if publisher_obj.start else ''
+            )
+        elif comment_class == Seat:
```
bderusha: Seats are not a top level object that we have pages for.  We should not be emailing people links to seat pages.  A seat change should result in an email about the course_run

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,72 @@
...
+
+        Arguments:
+            comment (Comment): Comment object
+    """
+    try:
+        object_pk = comment.object_pk
+        publisher_obj = comment.content_type.get_object_for_this_type(pk=object_pk)
+        comment_class = comment.content_type.model_class()
+
+        if comment_class == CourseRun:
```
bderusha, line 26: Rather than checking for the class, the duck-typing approach would be to give each class a method which can be called that returns the appropriate data.

e.g. `.url_path` or `.email_subject`

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,72 @@
...
+from django.template.loader import get_template
+from django.utils.translation import ugettext_lazy as _
+
+from course_discovery.apps.publisher.models import CourseRun, Seat
+
+
+log = logging.getLogger(__name__)
+
+
+def setup_email_for_comment(comment):
```
bderusha: Method name could be better.  This method does more than just setup the email.  It sends it too.  That is not clear

bderusha (changes requested):

course_discovery/apps/publisher_comments/emails.py
awais786: Do you mean comment cannot be made against a seat object ?

bderusha: Correct, based on the new designs I don't see how this is possible

course_discovery/apps/publisher_comments/emails.py
awais786: @bderusha each comment hold the content type of different models. That's why I am checking the class. And there is no data which we need here.

course_discovery/apps/publisher_comments/emails.py
awais786: Removed

awais786:

awais786:
@clintonb Please review this PR.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -62,6 +62,14 @@ def setUp(self):
         self.course = factories.CourseFactory()
         self.course2 = factories.CourseFactory()
 
+        self.user1 = UserFactory()
+        self.user2 = UserFactory()
+        self.user3 = UserFactory()
+        self.group_a = factories.GroupFactory(name='Test Group A')
```
clintonb: Unless these names are relevant for testing, they should be removed. They are extraneous noise.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -72,25 +80,47 @@ def test_post_back_url(self):
...
+    def test_group_institution(self):
+        """ Verify that method returns the institution against the course."""
+        self.assertEqual(self.course.group_institution, None)
+        self.course.assign_permission_by_group(self.group_a)
+        self.assertEqual(self.course.group_institution, self.group_a)
 
-        self.assertTrue(user1.has_perm(Course.VIEW_PERMISSION, self.course))
-        self.assertTrue(user2.has_perm(Course.VIEW_PERMISSION, self.course2))
+    def test_get_group_users_emails(self):
+        """ Verify that method returns the emails against the course group."""
```
clintonb: Verify the method returns the email addresses of users who are permitted to access the course AND have not disabled email notifications.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -72,25 +80,47 @@ def test_post_back_url(self):
...
-        self.assertFalse(user1.has_perm(Course.VIEW_PERMISSION, self.course))
-        self.assertFalse(user2.has_perm(Course.VIEW_PERMISSION, self.course2))
+    def assert_user_can_view_course(self, user, course):
+        """ Asserts the user can view the course. """
+        self.assertTrue(user.has_perm(Course.VIEW_PERMISSION, course))
 
-        self.course.assign_user_groups(user1)
-        self.course2.assign_user_groups(user2)
+    def test_group_institution(self):
+        """ Verify that method returns the institution against the course."""
```
clintonb: Verify the method returns groups permitted to access the course.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -72,25 +80,47 @@ def test_post_back_url(self):
...
-        self.assertTrue(user2.has_perm(Course.VIEW_PERMISSION, self.course2))
+    def test_get_group_users_emails(self):
+        """ Verify that method returns the emails against the course group."""
+        self.user3.groups.add(self.group_a)
+        self.course.assign_permission_by_group(self.group_a)
+        self.assertListEqual(self.course.get_group_users_emails(), [self.user1.email, self.user3.email])
 
-        self.assertFalse(user1.has_perm(Course.VIEW_PERMISSION, self.course2))
-        self.assertFalse(user2.has_perm(Course.VIEW_PERMISSION, self.course))
+        # After making user1 enable_email_notification as false it should not appear in
```
clintonb: The email addresses of users who have disabled email notifications should NOT be returned.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@ddt.ddt
+class CommentsEmailTests(TestCase):
+    """ Tests for the e-mail functionality for course, course-run and seats. """
+
+    def setUp(self):
+        super(CommentsEmailTests, self).setUp()
+
+        self.user = UserFactory(first_name='test-user1')
```
clintonb: The factory will generate a first name for you. See https://github.com/edx/course-discovery/pull/396.

course_discovery/apps/publisher_comments/tests/test_views.py
```diff
@@ -15,16 +17,23 @@ class CommentsTests(TestCase):
     """ Tests for the Comment functionality on `Courser`, `CourseRun` And `Seat` edit pages. """
     def setUp(self):
         super(CommentsTests, self).setUp()
-        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.user = UserFactory(is_staff=True, is_superuser=True, email='test@example.com')
```
clintonb: The factory will generate a email for you. See https://github.com/edx/course-discovery/pull/396.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    try:
+        object_pk = comment.object_pk
+        publisher_obj = comment.content_type.get_object_for_this_type(pk=object_pk)
+        comment_class = comment.content_type.model_class()
+
+        if comment_class == CourseRun:
+            course = publisher_obj.course
+            object_path = reverse('publisher:publisher_course_run_detail', args=[publisher_obj.id])
+            subject = _('New comment added in Course-run : {title}-{pacing_type}-{start}').format(
```
clintonb: Remove the space before the colon.

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -0,0 +1,62 @@
...
+                title=course.title,
+                pacing_type=publisher_obj.get_pacing_type_display(),
+                start=publisher_obj.start.strftime('%B %d, %Y') if publisher_obj.start else ''
+            )
+        else:
+            course = publisher_obj
+            object_path = reverse('publisher:publisher_courses_edit', args=[publisher_obj.id])
+            subject = _('New comment added in Course : {title}').format(title=course.title)
```
clintonb: Remove the space before the colon.

course_discovery/apps/publisher_comments/emails.py
clintonb: "course run" (no hyphen)

course_discovery/templates/publisher/email/comment.txt
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+
+{% blocktrans with first_name=comment.user.first_name object_type=object_type|lower title=course.title number=course.number %}
```
clintonb: Use the `trimmed` option to eliminate whitespace. See https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#blocktrans-template-tag for more info. Remember to update the translations, and ensure there are no newline characters in the source files, once this change is made.

course_discovery/templates/publisher/email/email_base.html
```diff
@@ -0,0 +1,21 @@
+{% load i18n %}
+<!DOCTYPE html>
+<html lang="en">
```
clintonb: This content will be translated. The language should be set to the current/request language. See https://docs.djangoproject.com/en/dev/topics/i18n/translation/#get-current-language.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
+Publisher
+===============
```
clintonb: The underlining elements should not extend beyond the width of the text being underlined.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
...
+
+The ``publisher`` app at :file:`course_discovery/apps/publisher` is for A single point information management
+system that supports the authoring, review and approval workflow for [Course About] Section between all the
+stakeholders (Ed Services, Marketing & Partners) and subsequently publish the finalized content onto the live page.
+
+
+Configure Emails
+-----------------------------------
```
clintonb: The underlining elements should not extend beyond the width of the text being underlined.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
...
+The table below describes the relevant settings.
+
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+| Setting                            | Description                                                                | Value                                                                    |
++====================================+============================================================================+==========================================================================+
+| PUBLISHER_FROM_EMAIL               | Official email address for sending emails.                                 | Email address                                                            |
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+
+You can use different email backend for sending emails as defined in django configuration.
```
clintonb: Since the publisher tool uses the built-in Django email functionality, any Django email backend can be used. For info on configuring email backends see https://docs.djangoproject.com/en/1.10/topics/email/.

We at edX.org use Amazon's Simple Email Service (SES) (https://aws.amazon.com/ses/). If you'd also like to use SES, the django-ses (https://github.com/django-ses/django-ses) is installed as a base requirement. Simply define values for the settings below to configure the backend.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
+Publisher
+===============
+
+The ``publisher`` app at :file:`course_discovery/apps/publisher` is for A single point information management
```
clintonb: The `publisher` tool is an information management system that supports the course authoring, review, and approval workflow. The tool manages courses as they transition through the lifecycle of creation (across various subsystems), release/publication to the edX.org marketing site, and eventual retirement/archival.

Note: This app is primarily built for the edX.org use case.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
...
+system that supports the authoring, review and approval workflow for [Course About] Section between all the
+stakeholders (Ed Services, Marketing & Partners) and subsequently publish the finalized content onto the live page.
+
+
+Configure Emails
+-----------------------------------
+For emails functionality you can update your local settings (ideally in :file:`course_discovery/settings/private.py`).
```
clintonb: The publisher tool supports notifying responsible parties when users comment on courses and course runs. Notifications are sent via Django's email backend. In order to send emails the `enable_publisher_comment_email_notifications` switch must be activated, and the following settings must be defined:

course_discovery/apps/publisher_comments/models.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 class Comments(CommentAbstractModel):
     modified = ModificationDateTimeField(_('modified'))
+
+
+@receiver(post_save, sender=Comments)
+def send_email(sender, instance, **kwargs):    # pylint: disable=unused-argument
+    """ Send email on new comment. """
+    if waffle.switch_is_active('enable_emails'):
```
clintonb: 1. Rename this to `'enable_publisher_comment_email_notifications'` so that is perfectly clear what this switch affects.
2. Add a migration creating this switch, and defaulting it to the inactive state.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
...
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+| AWS_SES_REGION_NAME                | Region your SES service is using.                                          | (This should be set to the value from the AWS account.)                  |
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+| AWS_SES_REGION_ENDPOINT            | Region your SES service is using.                                          | (This should be set to the value from the AWS account.)                  |
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+
+For default email backend SMTP, following settings are require
```
clintonb: Remove this section and simply link to the Django docs.

course_discovery/templates/publisher/email/comment.html
```diff
@@ -0,0 +1,20 @@
...
+        <p>
+            {{ comment.comment }}
+        </p>
+
+        <p>
+            <a href="{{ page_url }}">{% trans "View comment" %}</a>
+        </p>
+        <p>{% trans "The edX team" %}</p>
+<!-- End Message Body -->
+<br>
```
clintonb: This doesn't seem necessary.

clintonb (changes requested):

course_discovery/apps/publisher_comments/models.py
awais786: I renamed this to `enable_publisher_email_notifications` so that it can be use for all kinds of emails.

awais786:
@clintonb PR is ready for final review.

course_discovery/apps/publisher/migrations/0013_create_enable_email_notifications_switch.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('publisher', '0012_auto_20161020_0718'),
+        ('waffle', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_switch, reverse_code=delete_switch),
```
clintonb: Nit: No need for the kwarg: `migrations.RunPython(create_switch, delete_switch),`

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
...
++====================================+============================================================================+==========================================================================+
+| PUBLISHER_FROM_EMAIL               | Official email address for sending emails.                                 | Email address                                                            |
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+
+Since the publisher tool uses the built-in Django email functionality, any Django email backend can be used. For info on configuring email backends see https://docs.djangoproject.com/en/1.10/topics/email/.
+
+We at edX.org use Amazon's Simple Email Service (SES) (https://aws.amazon.com/ses/). If you'd also like to use SES, the django-ses (https://github.com/django-ses/django-ses) is installed as a base requirement. Simply define values for the settings below to configure the backend.
```
clintonb: The URLs in parentheses should be setup as proper RST links.

docs/publisher.rst
```diff
@@ -0,0 +1,53 @@
...
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+| AWS_SES_REGION_NAME                | Region your SES service is using.                                          | (This should be set to the value from the AWS account.)                  |
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+| AWS_SES_REGION_ENDPOINT            | Region your SES service is using.                                          | (This should be set to the value from the AWS account.)                  |
++------------------------------------+----------------------------------------------------------------------------+--------------------------------------------------------------------------+
+
+https://docs.djangoproject.com/en/1.10/topics/email/#smtp-backend
```
clintonb: You don't really need this section. We've already linked to the Django docs. We only provide SES info because we include that functionality.

clintonb (approved):

PR #380, opened 2016-10-12
ECOM-5680 Sortable organizations in admin program form

MatthewPiatetsky:
Context: https://openedx.atlassian.net/browse/ECOM-5680
Reusing work from this PR: https://github.com/edx/course-discovery/pull/349
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_admin.py
+++ b/course_discovery/apps/course_metadata/tests/test_admin.py
@@ -2,7 +2,9 @@ import itertools
 
 import ddt
 from django.core.urlresolvers import reverse
-from django.test import TestCase
+from django.test import TestCase, LiveServerTestCase
+from selenium import webdriver
+from selenium.webdriver.common.action_chains import ActionChains
 
 from course_discovery.apps.course_metadata.choices import ProgramStatus
 from course_discovery.apps.course_metadata.forms import ProgramAdminForm
@@ -179,3 +181,75 @@ class AdminTests(TestCase):
         self.assertEqual(0, program.courses.all().count())
         response = self.client.get(reverse('admin:course_metadata_program_change', args=(program.id,)))
         self.assertEqual(response.status_code, 200)
+
+
+class ProgramAdminFunctionalTests(LiveServerTestCase):
+    """ Functional Tests for Admin page."""
+    def setUp(self):
+        super(ProgramAdminFunctionalTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+        self.course_runs = factories.CourseRunFactory.create_batch(2)
+        self.courses = [course_run.course for course_run in self.course_runs]
+
+        self.excluded_course_run = factories.CourseRunFactory(course=self.courses[0])
+        self.program = factories.ProgramFactory(
+            courses=self.courses, excluded_course_runs=[self.excluded_course_run]
+        )
+        self.browser = webdriver.Firefox()
+
+        # Get Page
+        domain = self.live_server_url
+        url = reverse('admin:course_metadata_program_change', args=(self.program.id,))
+        self.browser.get(domain + url)
+
+        # Login
+        username = self.browser.find_element_by_id('id_username')
+        password = self.browser.find_element_by_id('id_password')
+        username.send_keys(self.user.username)
+        password.send_keys(USER_PASSWORD)
+        self.browser.find_element_by_css_selector('input[type=submit]').click()
+
+        # This window size is close to the window size when running on travis
+        self.browser.set_window_size(548, 768)
+
+    def tearDown(self):
+        super(ProgramAdminFunctionalTests, self).tearDown()
+        self.browser.quit()
+
+    def test_all_fields(self):
+        # Make sure that all expected fields are present
+        classes = [css_class for field in self.browser.find_elements_by_class_name('form-row')
+                   for css_class in field.get_attribute('class').split(' ')
+                   if css_class.startswith('field-') or css_class.startswith('dynamic-')]
+        expected_classes = ['field-title', 'field-subtitle', 'field-status', 'field-type',
+                            'field-partner', 'field-banner_image', 'field-banner_image_url',
+                            'field-card_image_url', 'field-marketing_slug', 'field-overview',
+                            'field-credit_redemption_overview', 'field-video',
+                            'field-weeks_to_complete', 'field-min_hours_effort_per_week',
+                            'field-max_hours_effort_per_week', 'field-courses',
+                            'field-order_courses_by_start_date', 'field-custom_course_runs_display',
+                            'field-excluded_course_runs', 'field-authoring_organizations',
+                            'field-credit_backing_organizations', 'field-job_outlook_items',
+                            'field-expected_learning_items', 'dynamic-Program_faq',
+                            'dynamic-Program_individual_endorsements',
+                            'dynamic-Program_corporate_endorsements']
+        self.assertEqual(classes, expected_classes)
+
+    def test_sortable_select_drag_and_drop(self):
+        # Get order of select elements
+        hidden_options_text = [el.text for el in
+                               self.browser.find_elements_by_css_selector('.field-courses option')]
+        first_select_element = self.browser.find_element_by_css_selector('.field-courses .select2-selection__choice')
+
+        # Drag and drop
+        first_select_element.click()
+        ActionChains(self.browser).drag_and_drop_by_offset(first_select_element, 500, 0).perform()
+
+        # Simulate expected drag and drop
+        hidden_options_text = [hidden_options_text[1], hidden_options_text[0]]
+
+        # Get actual results of drag and drop
+        new_hidden_options_text = [el.text for el in
+                                   self.browser.find_elements_by_css_selector('.field-courses option')]
+        self.assertEqual(hidden_options_text, new_hidden_options_text)
--- a/course_discovery/static/js/sortable_select.js
+++ b/course_discovery/static/js/sortable_select.js
@@ -1,44 +1,49 @@
-function updateSelect2Data(visibleCourseTitles){
+function updateSelect2Data(el){
     var i, j,
-        visibleCourseTitlesLength,
+        visibleTitlesLength,
         selectOptionsLength,
-        visibleCourseTitles = [],
+        visibleTitles = [],
         selectOptions = [],
         items = [],
-        selectOptionsSelector = '.field-courses .select2-hidden-accessible';
+        selectOptionsElement = $(el).find('.select2-hidden-accessible'),
+        selectChoicesElement = $(el).find('.select2-selection__choice'),
+        selectOptionElement = $(selectOptionsElement).find('option');
 
-    $('.field-courses .select2-selection__choice').each(function(index, value){
+    selectChoicesElement.each(function(index, value){
         if (value.title){
-            visibleCourseTitles.push(value.title);
+            visibleTitles.push(value.title);
         }
     });
 
-    $('.field-courses .select2-hidden-accessible option').each(function(index, value){
+    selectOptionElement.each(function(index, value){
         selectOptions.push({id: value.value, text: value.text});
     });
 
     // Update select2 options with new data
-    visibleCourseTitlesLength = visibleCourseTitles.length;
+    visibleTitlesLength = visibleTitles.length;
     selectOptionsLength = selectOptions.length;
-    for (i = 0; i < visibleCourseTitlesLength; i++) {
+    for (i = 0; i < visibleTitlesLength; i++) {
         for (j = 0; j < selectOptionsLength; j++) {
-            if (selectOptions[j].text === visibleCourseTitles[i]){
+            if (selectOptions[j].text === visibleTitles[i]){
                 items.push('<option selected="selected" value="' + selectOptions[j].id + '">' +
                            selectOptions[j].text + '</option>'
                 );
             }
         }
     }
+
     if (items){
-        $(selectOptionsSelector).html(items.join('\n'));
+        selectOptionsElement.html(items.join('\n'));
     }
 }
+
 $(window).load(function(){
     $(function() {
-        var domSelector = '.field-courses .select2-selection--multiple';
-        $('.field-courses ul.select2-selection__rendered').sortable({
-            containment: 'parent',
-            update: updateSelect2Data
+        $('.sortable-select').parents('.form-row').each(function(index, el){
+            $(el).find('ul.select2-selection__rendered').sortable({
+                containment: 'parent',
+                update: function(){updateSelect2Data(el);}
+            })
         })
     })
 });
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -111,6 +111,8 @@ class ProgramAdmin(admin.ModelAdmin):
         try:
             # courses are ordered by django id, but form.cleaned_data is ordered correctly
             obj.courses = form.cleaned_data.get('courses')
+            obj.authoring_organizations = form.cleaned_data.get('authoring_organizations')
+            obj.credit_backing_organizations = form.cleaned_data.get('credit_backing_organizations')
             obj.save()
             self.save_error = False
         except ProgramPublisherException as ex:
--- a/course_discovery/apps/course_metadata/forms.py
+++ b/course_discovery/apps/course_metadata/forms.py
@@ -39,18 +39,21 @@ class ProgramAdminForm(HackDjangoAutocompleteMixin, forms.ModelForm):
                 url='admin_metadata:course-autocomplete',
                 attrs={
                     'data-minimum-input-length': 3,
+                    'class': 'sortable-select',
                 },
             ),
             'authoring_organizations': autocomplete.ModelSelect2Multiple(
                 url='admin_metadata:organisation-autocomplete',
                 attrs={
                     'data-minimum-input-length': 3,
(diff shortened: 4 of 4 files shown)


course_discovery/static/js/sortable_select.js
```diff
@@ -33,12 +35,19 @@ function updateSelect2Data(visibleCourseTitles){
...
 $(window).load(function(){
     $(function() {
-        var domSelector = '.field-courses .select2-selection--multiple';
-        $('.field-courses ul.select2-selection__rendered').sortable({
-            containment: 'parent',
-            update: updateSelect2Data
-        })
+        sortableSelect('.field-courses');
+        sortableSelect('.field-authoring_organizations');
+        sortableSelect('.field-credit_backing_organizations');
```
rlucioni: I wonder if these fields should be configurable or otherwise more visible. Just a thought.

rlucioni (approved):

course_discovery/static/js/sortable_select.js
MatthewPiatetsky: Do you have thoughts on how that could work?
Can I use django to add a specific class to these fields?
Currently I'm just including sortable_select.js in the Meta class of ProgramAdmin

course_discovery/static/js/sortable_select.js
```diff
@@ -1,19 +1,21 @@
-function updateSelect2Data(visibleCourseTitles){
+function updateSelect2Data(field){
     var i, j,
         visibleCourseTitlesLength,
```
clintonb: This seems very field-specific. This file needs to be cleaned to make it generic to any field.

course_discovery/static/js/sortable_select.js
clintonb: You should be making a custom widget. See https://docs.djangoproject.com/en/1.10/ref/forms/widgets/. It may be easier to use an existing widget.

clintonb (changes requested):

course_discovery/static/js/sortable_select.js
MatthewPiatetsky: good point, I just updated it to be more generic, and rather than specifying specific fields in the file, I have it looking for a class set on widget initialization.
Let me know your thoughts.

course_discovery/static/js/sortable_select.js
MatthewPiatetsky: we were already using the django autocomplete light widget and this was just extending some functionality they did not have.
We could try contributing back, though just because this functionality could be useful for us, doesn't necessarily mean they'll want to accept/maintain it

clintonb: A proper widget should not have hard-coded class names. Those should be derived from the field passed to the template. We don't have to contribute this back to the main repo; but, we should not be hardcoding field names in our own code.

MatthewPiatetsky: Do you mean the classes I hardcode above? That diff is outdated.  
If you mean my current approach, I think I could use some more details on what you're suggesting

clintonb (commented):
I approve of the implementation; but, I would like to see tests before approving this to be merged. This form is pretty important, and we aren't testing it at all. See https://docs.djangoproject.com/en/1.8/topics/testing/tools/#liveservertestcase.

MatthewPiatetsky:
@clintonb I don't object to writing functional tests, but maybe that should be a separate ticket we pull into the sprint (we could mark it as blocking to this ticket/PR)?
It seems this repo doesn't have much in way of these tests, and before we pick a path to go down, I think it would be useful to consult with the team, especially @AlasdairSwan.
For example, maybe we would want to go with the karma setup we have on the marketing site rather than selenium (and having a separate suite of front end tests could be useful).
This seems like the type of thing we should have some discussion around.

clintonb:
This repo has a lot of tests actually. I wrote many of them. 😉 We've been discussing testing for quite some time now. It's time we actually wrote some tests. Selenium tests are quick and easy to write, and meet our needs of testing the UI as the user would. I see no reason to introduce another testing tool unnecessarily.

MatthewPiatetsky:
I was discussing functional tests, not unit tests. (If you use functional to mean something different, I use it to refer to browser based tests).
The only file I see with selenium tests is affiliate_cookie_tests, if I'm missing other selenium tests let me know.

Not objecting to selenium, (in fact I think it has functionality that would be better suited to this case than JS focused test runners), but nevertheless, that seems to me like something we should discuss. 
If we decided we want to standardize on a specific framework for functional tests and it wasn't selenium, it would be easier to do that now, than when we already had a bunch of selenium tests written and then rewriting those. I'm fine with not merging this PR until we do that.

I know we have tickets focused around front end technologies, https://openedx.atlassian.net/browse/ECOM-5776, and we may want to do similar thinking about front end test frameworks.

clintonb:
Ignore the `acceptance_tests` directory for now. You can use `DjangoLiveServerTestCase` to create functional tests alongside the unit tests in in `course_metadata/tests/test_admin.py`. The advantage of this is that you get Django setup for free. Additionally, these Selenium tests will easily transfer to `acceptance_tests` if/when we decide to make that change.

MatthewPiatetsky:
ok, I have some thoughts, but I will save them until we start work on acceptance tests in catalog and I will write selenium tests for this particular page

MatthewPiatetsky:
@clintonb I added a starting point, but what kind of stuff do we want to test in this test?

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -1,19 +1,21 @@
...
 from course_discovery.apps.course_metadata.tests import factories
 from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD
 from course_discovery.apps.core.tests.helpers import make_image_file
 
+from selenium import webdriver
 
 # pylint: disable=no-member
 @ddt.ddt
-class AdminTests(TestCase):
+class AdminTests(TestCase, LiveServerTestCase):
```
clintonb: There should be two classes. Leave the existing tests. Add a new test class, `ProgramAdminFunctionalTests`, for your new tests.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -44,6 +46,41 @@ def assert_form_valid(self, data, files):
...
+        self.browser.get(domain + url)
+
+        # Login
+        username = self.browser.find_element_by_id("id_username")
+        password = self.browser.find_element_by_id("id_password")
+        username.send_keys(self.user.username)
+        password.send_keys(USER_PASSWORD)
+        self.browser.find_elements_by_tag_name('input')[-1].click()
+
+        # Make sure that all expected fields are present
```
clintonb: All you need to test, for this PR, are the fields you modified.

MatthewPiatetsky:
@clintonb I've updated the test, and it works locally (consistently), but not on Travis (consistently), so don't think the issue is a flappy test. Any ideas?

clintonb:
@MatthewPiatetsky it seems you have a sorting error. What have you done to investigate/fix?

package.json
```diff
@@ -7,7 +7,8 @@
   },
   "dependencies": {
     "bower": "^1.7.9",
-    "requirejs": "^2.2.0"
```
clintonb, line 10: Why are these changes necessary?

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,70 @@ def test_new_program_without_courses(self):
...
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+        self.course_runs = factories.CourseRunFactory.create_batch(3)
+        self.courses = [course_run.course for course_run in self.course_runs]
+
+        self.excluded_course_run = factories.CourseRunFactory(course=self.courses[0])
+        self.program = factories.ProgramFactory(
+            courses=self.courses, excluded_course_runs=[self.excluded_course_run]
+        )
+        self.browser = webdriver.Chrome('node_modules/chromedriver/bin/chromedriver')
```
clintonb: Use Firefox. This might be the source of your trouble with Travis.

package.json
MatthewPiatetsky: I added chromedriver

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: will try Firefox

MatthewPiatetsky: @clintonb don't need to change package.json now that I am using Firefox, though error is still present.
I may try some other implementations, but could be slow if I can't reproduce issue locally.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,70 @@ def test_new_program_without_courses(self):
         self.assertEqual(0, program.courses.all().count())
         response = self.client.get(reverse('admin:course_metadata_program_change', args=(program.id,)))
         self.assertEqual(response.status_code, 200)
+
+
+class ProgramAdminFunctionalTests(TestCase, LiveServerTestCase):
```
clintonb: You only need `LiveServerTestCase` here.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+
+    def test_sortable_select_drag_and_drop(self):
+        # Get order of select elements
+        hidden_options_text = [el.text for el in
+                               self.browser.find_elements_by_css_selector('.field-courses option')]
+        first_select_element = self.browser.find_element_by_css_selector('.field-courses .select2-selection__choice')
+        # Drag and drop
+        first_select_element.click()
+        # set window size
+        self.browser.set_window_size(548, 768)
```
clintonb: Why such an arbitrary size? Why not something more common (e.g. 1024x768)? Can we open the browser at this size?

Do this as part of `setUp()` in case future tests run into this issue. Also, add a comment explaining why this is necessary.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+                            'row2  empty-form', 'row1  dynamic-Program_corporate_endorsements',
+                            'row2  empty-form']
+        self.assertEqual(classes, expected_classes)
+
+    def test_sortable_select_drag_and_drop(self):
+        # Get order of select elements
+        hidden_options_text = [el.text for el in
+                               self.browser.find_elements_by_css_selector('.field-courses option')]
+        first_select_element = self.browser.find_element_by_css_selector('.field-courses .select2-selection__choice')
+        # Drag and drop
```
clintonb, line 245: This all reads like a wall of text. Add a new line before each comment to make this method a bit more legible.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+                            'field-partner', 'field-banner_image', 'field-banner_image_url',
+                            'field-card_image_url', 'field-marketing_slug', 'field-overview',
+                            'field-credit_redemption_overview', 'field-video',
+                            'field-weeks_to_complete', 'field-min_hours_effort_per_week',
+                            'field-max_hours_effort_per_week', 'field-courses',
+                            'field-order_courses_by_start_date', 'field-custom_course_runs_display',
+                            'field-excluded_course_runs', 'field-authoring_organizations',
+                            'field-credit_backing_organizations', 'field-job_outlook_items',
+                            'field-expected_learning_items', 'row1  dynamic-Program_faq',
+                            'row2  empty-form', 'row1  dynamic-Program_individual_endorsements',
```
clintonb: `'row2  empty-form'` is duplicated numerous times, and it is not clear if there is any value to validating its presence.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+        domain = 'http://localhost:8081'
+        url = reverse('admin:course_metadata_program_change', args=(self.program.id,))
+        self.browser.get(domain + url)
+
+        # Login
+        username = self.browser.find_element_by_id("id_username")
+        password = self.browser.find_element_by_id("id_password")
+        username.send_keys(self.user.username)
+        password.send_keys(USER_PASSWORD)
+        self.browser.find_elements_by_tag_name('input')[-1].click()
```
clintonb: This is brittle. You can get the submit button by finding the element matching `input[type=submit]`.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+        self.browser = webdriver.Firefox()
+        self.browser.implicitly_wait(3)
+
+        # Get Page
+        domain = 'http://localhost:8081'
+        url = reverse('admin:course_metadata_program_change', args=(self.program.id,))
+        self.browser.get(domain + url)
+
+        # Login
+        username = self.browser.find_element_by_id("id_username")
```
clintonb: Single quotes here and below.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+
+        self.excluded_course_run = factories.CourseRunFactory(course=self.courses[0])
+        self.program = factories.ProgramFactory(
+            courses=self.courses, excluded_course_runs=[self.excluded_course_run]
+        )
+        self.browser = webdriver.Firefox()
+        self.browser.implicitly_wait(3)
+
+        # Get Page
+        domain = 'http://localhost:8081'
```
clintonb: This is brittle and will break if we ever change the arguments. Use `self.live_server_url`.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -179,3 +182,72 @@ def test_new_program_without_courses(self):
...
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+        self.course_runs = factories.CourseRunFactory.create_batch(2)
+        self.courses = [course_run.course for course_run in self.course_runs]
+
+        self.excluded_course_run = factories.CourseRunFactory(course=self.courses[0])
+        self.program = factories.ProgramFactory(
+            courses=self.courses, excluded_course_runs=[self.excluded_course_run]
+        )
+        self.browser = webdriver.Firefox()
+        self.browser.implicitly_wait(3)
```
clintonb: Add a comment explaining the need for the wait.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -2,7 +2,10 @@
 
 import ddt
 from django.core.urlresolvers import reverse
-from django.test import TestCase
+from django.test import TestCase, LiveServerTestCase
+
```
clintonb: Remove this newline.

clintonb (changes requested):

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: removed

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: think in the end it ended up being not necessary

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: This was the list of classes that the top level field elements have on the form.
I could do some sort of contains rather than an assertEqual, though it could be useful to know if something little changed in the form?

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: added new lines

course_discovery/apps/course_metadata/tests/test_admin.py
MatthewPiatetsky: If we will be doing computations that could have to do with window size, would be good to have window size closer to what we're getting on travis.
I got this window size by printing the window size during a travis test run.  
(moved and added comment)

course_discovery/apps/course_metadata/tests/test_admin.py
clintonb: I don't think there is much value in knowing if order shifted. We really only care that the fields are present on the page.

MatthewPiatetsky: ok will adjust

MatthewPiatetsky: adjusted to only check classes that contain the field name

clintonb (approved):

PR #381, opened 2016-10-12
Exclude deleted programs from course and course_run api

schenedx:
Please review @edx/ecommerce 
Originally, the associated programs array in the course and course_run api object would include all programs regardless of status. With this change, the deleted programs are filtered. Only the query string parameter include_deleted_programs will list out all associated programs
ECOM-5729

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -15,7 +15,7 @@ from taggit_serializer.serializers import TagListSerializerField, TaggitSerializ
 
 from course_discovery.apps.api.fields import StdImageSerializerField, ImageField
 from course_discovery.apps.catalogs.models import Catalog
-from course_discovery.apps.course_metadata.choices import CourseRunStatus
+from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType, FAQ,
     CorporateEndorsement, Endorsement, Position
@@ -384,8 +384,11 @@ class CourseRunWithProgramsSerializer(CourseRunSerializer):
     programs = serializers.SerializerMethodField()
 
     def get_programs(self, obj):
+        # Filter out non-deleted programs which this course_run is part of the program course_run exclusion
         programs = [program for program in obj.programs.all()
-                    if obj.id not in (run.id for run in program.excluded_course_runs.all())]
+                    if (self.context.get('include_deleted_programs') or
+                        program.status != ProgramStatus.Deleted) and
+                    obj.id not in (run.id for run in program.excluded_course_runs.all())]
 
         return NestedProgramSerializer(programs, many=True).data
 
@@ -459,7 +462,15 @@ class CourseSerializer(MinimalCourseSerializer):
 
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
-    programs = NestedProgramSerializer(many=True)
+    programs = serializers.SerializerMethodField()
+
+    def get_programs(self, obj):
+        if self.context.get('include_deleted_programs'):
+            eligible_programs = obj.programs.all()
+        else:
+            eligible_programs = obj.programs.exclude(status=ProgramStatus.Deleted)
+
+        return NestedProgramSerializer(eligible_programs, many=True).data
 
     class Meta(CourseSerializer.Meta):
         model = Course
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -161,10 +161,42 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
     def get_expected_data(self, course, request):
         expected = super().get_expected_data(course, request)
         expected.update({
-            'programs': NestedProgramSerializer(course.programs, many=True, context={'request': request}).data,
+            'programs': NestedProgramSerializer(
+                course.programs,
+                many=True,
+                context={'request': request}
+            ).data,
         })
 
         return expected
+
+    def setUp(self):
+        super().setUp()
+        self.request = make_request()
+        self.course = CourseFactory()
+        self.deleted_program = ProgramFactory(
+            courses=[self.course],
+            status=ProgramStatus.Deleted
+        )
+
+    def test_exclude_deleted_programs(self):
+        """
+        If the associated program is deleted,
+        CourseWithProgramsSerializer should not return any serialized programs
+        """
+        serializer = self.serializer_class(self.course, context={'request': self.request})
+        self.assertEqual(serializer.data['programs'], [])
+
+    def test_include_deleted_programs(self):
+        """
+        If the associated program is deleted, but we are sending in the 'include_deleted_programs' flag
+        CourseWithProgramsSerializer should return deleted programs
+        """
+        serializer = self.serializer_class(
+            self.course,
+            context={'request': self.request, 'include_deleted_programs': 1}
+        )
+        self.assertEqual(serializer.data, self.get_expected_data(self.course, self.request))
 
 
 class MinimalCourseRunSerializerTests(TestCase):
@@ -234,16 +266,23 @@ class CourseRunSerializerTests(MinimalCourseRunSerializerTests):
 
 
 class CourseRunWithProgramsSerializerTests(TestCase):
-    def test_data(self):
-        request = make_request()
-        course_run = CourseRunFactory()
-        serializer_context = {'request': request}
-        serializer = CourseRunWithProgramsSerializer(course_run, context=serializer_context)
-        ProgramFactory(courses=[course_run.course])
+    def setUp(self):
+        super().setUp()
+        self.request = make_request()
+        self.course_run = CourseRunFactory()
+        self.serializer_context = {'request': self.request}
 
-        expected = CourseRunSerializer(course_run, context=serializer_context).data
+    def test_data(self):
+        serializer = CourseRunWithProgramsSerializer(self.course_run, context=self.serializer_context)
+        ProgramFactory(courses=[self.course_run.course])
+
+        expected = CourseRunSerializer(self.course_run, context=self.serializer_context).data
         expected.update({
-            'programs': NestedProgramSerializer(course_run.course.programs, many=True, context=serializer_context).data,
+            'programs': NestedProgramSerializer(
+                self.course_run.course.programs,
+                many=True,
+                context=self.serializer_context
+            ).data,
         })
 
         self.assertDictEqual(serializer.data, expected)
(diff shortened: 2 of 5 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -383,7 +383,11 @@ class CourseRunWithProgramsSerializer(CourseRunSerializer):
     programs = serializers.SerializerMethodField()
 
     def get_programs(self, obj):
-        programs = [program for program in obj.programs.all()
+        eligible_programs = obj.programs.exclude(status=ProgramStatus.Deleted)
+        if self.context.get('include_deleted_programs'):
+            eligible_programs = obj.programs.all()
```
rlucioni: An if/else seems appropriate here, to help with readability.

course_discovery/apps/api/serializers.py
```diff
@@ -458,7 +462,13 @@ def get_marketing_url(self, obj):
 
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
-    programs = NestedProgramSerializer(many=True)
+    programs = serializers.SerializerMethodField()
+
+    def get_programs(self, obj):
+        eligible_programs = obj.programs.exclude(status=ProgramStatus.Deleted)
+        if self.context.get('include_deleted_programs'):
+            eligible_programs = obj.programs.all()
```
rlucioni, line 469: If/else might make this a little easier to read, too.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -161,11 +161,43 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
...
+    def setUp(self):
+        super(CourseWithProgramsSerializerTests, self).setUp()
+        self.request = make_request()
+        self.course = CourseFactory()
+        self.deleted_program = ProgramFactory(
+            courses=[self.course],
+            status=ProgramStatus.Deleted
+        )
+
+    def test_exlude_deleted_programs(self):
```
rlucioni: s/exlude/exclude

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -161,11 +161,43 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
...
+                course.programs,
+                many=True,
+                context={'request': request}
+            ).data,
         })
 
         return expected
 
+    def setUp(self):
+        super(CourseWithProgramsSerializerTests, self).setUp()
```
rlucioni: `super()` is smarter in Python 3. You can just do `super().setUp()`, both here an anywhere else you're adding a call to super.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -253,18 +292,37 @@ def test_data_excluded_course_run(self):
...
+        serializer = CourseRunWithProgramsSerializer(self.course_run, context=self.serializer_context)
+        ProgramFactory(courses=[self.course_run.course], excluded_course_runs=[self.course_run])
+        expected = CourseRunSerializer(self.course_run, context=self.serializer_context).data
         expected.update({
             'programs': [],
         })
 
         self.assertDictEqual(serializer.data, expected)
 
+    def test_exlude_deleted_programs(self):
```
rlucioni: s/exlude/exclude

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -253,18 +292,37 @@ def test_data_excluded_course_run(self):
...
+        serializer = CourseRunWithProgramsSerializer(self.course_run, context=self.serializer_context)
+        self.assertEqual(serializer.data['programs'], [])
+
+    def test_include_deleted_programs(self):
+        """
+        If the associated program is deleted, but we are sending in the 'include_deleted_programs' flag
+        CourseRunWithProgramsSerializer should return deleted programs
+        """
+        deleted_program = ProgramFactory(courses=[self.course_run.course], status=ProgramStatus.Deleted)
+        self.serializer_context.update({'include_deleted_programs': 1})
```
rlucioni: Minor, but there's no need to call `update()` if you're adding a single key. You can do this more concisely as `self.serializer_context['include_deleted_programs'] = 1`.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -32,17 +33,46 @@ def test_get(self):
...
+                response.data,
+                self.serialize_course_run(self.course_run, extra_context={'include_deleted_programs': True})
+            )
+
     def test_list(self):
         """ Verify the endpoint returns a list of all catalogs. """
         url = reverse('api:v1:course_run-list')
 
-        with self.assertNumQueries(11):
+        with self.assertNumQueries(13):
```
rlucioni: Do we know what these extra queries are? Are any duplicated unnecessarily?

course_discovery/apps/api/v1/views.py
```diff
@@ -262,6 +263,12 @@ def list(self, request, *args, **kwargs):
               type: integer
               paramType: query
               multiple: false
+            - name: include_deleted_program
```
rlucioni: Make this plural to match the expected querystring param.

course_discovery/apps/api/v1/views.py
```diff
@@ -358,6 +366,12 @@ def list(self, request, *args, **kwargs):
...
```
rlucioni: Make this plural to match the expected querystring param.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -21,16 +22,41 @@ def test_get(self):
...
             self.assertEqual(response.data, self.serialize_course(self.course))
 
+    def test_get_exclude_deleted_programs(self):
+        """ Verify the endpoint returns no deleted associated programs """
+        ProgramFactory(courses=[self.course], status=ProgramStatus.Deleted)
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
+        with self.assertNumQueries(12):
+            response = self.client.get(url)
+            self.assertEqual(response.status_code, 200)
+            self.assertEqual(response.data.get('programs'), [])
```
rlucioni, line 37: Both here and below, you can try folding these assertions together into one using Django's [`assertContains()`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.SimpleTestCase.assertContains). It's useful any time you need to assert on response data while also checking status code.

rlucioni (changes requested):

course_discovery/apps/api/serializers.py
clintonb: Either an if-else, or start with all programs, and filter from there.

course_discovery/apps/api/tests/test_serializers.py
clintonb: Use a mixin to de-dupe these methods.

course_discovery/apps/api/serializers.py
```diff
@@ -383,7 +383,11 @@ class CourseRunWithProgramsSerializer(CourseRunSerializer):
     programs = serializers.SerializerMethodField()
 
     def get_programs(self, obj):
-        programs = [program for program in obj.programs.all()
+        eligible_programs = obj.programs.exclude(status=ProgramStatus.Deleted)
```
clintonb: Similar to the line above, filter in Python to avoid an extra database call.

clintonb (changes requested):
We should figure out how to avoid the extra database calls. Filtering in Python should fix this.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
schenedx: I don't understand why you think `assertContains()` would cover both asserts here. The second assert I am doing make sure the `programs` array is empty. I don't know if we can use `assertContains()' for that, can we?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
schenedx: I digged into it and found I did make extra query unnecessarily. I have updated the serializer to avoid the extra query

course_discovery/apps/api/tests/test_serializers.py
schenedx: @clintonb Which two method you think we can dedupe?

schenedx:
@clintonb I made changes according to review suggestions. Please take another review pass.

course_discovery/apps/api/serializers.py
```diff
@@ -458,7 +461,15 @@ def get_marketing_url(self, obj):
 
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
-    programs = NestedProgramSerializer(many=True)
+    programs = serializers.SerializerMethodField()
+
+    def get_programs(self, obj):
+        if self.context.get('include_deleted_programs'):
```
clintonb, line 468: Nit

``` python
eligible_programs = obj.programs.all()

if self.context.get('include_deleted_programs'):
  eligible_programs = eligible_programs.exclude(status=ProgramStatus.Deleted)
```

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -21,16 +22,41 @@ def test_get(self):
         """ Verify the endpoint returns the details for a single course. """
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
 
-        with self.assertNumQueries(18):
+        with self.assertNumQueries(19):
```
clintonb, line 25: Why do we still have an increase?

course_discovery/apps/api/serializers.py
clintonb: Actually...this might be the source of the query count increase. Try filtering via Python, as you do above, to see if that resolves the issue.

clintonb (approved):
With the exception of the increase in query count, everything looks good.

course_discovery/apps/api/serializers.py
schenedx: @clintonb Unfortunately, that's the incorrect logic. Should really be either:

``` python
eligible_programs = obj.programs.all()
if not self.context.get('include_deleted_programs'):
    eligible_programs = eligible_programs.exclude(status=ProgramStatus.Deleted)
```

or

``` python
eligible_programs = obj.programs.exclude(status=ProgramStatus.Deleted)
if self.context.get('include_deleted_programs'):
    eligible_programs = obj.programs.all()
```

For the first variation, @rlucioni gave me the feedback on difficulty to read.
For the second variation that actually introduces 3 extra unnecessary DB queries if we do want all programs.
That's why the `if` and `else` statement.

clintonb: @schenedx try filtering via Python to see if that removes the extra database call.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
schenedx: It's related to the `exclude()` call. What happens is, if the course has empty Programs reference in DB, the `exclude()` call for the querySet will make a call anyways to make sure exclusion logic has been applied, even on an empty set. I don't know how I can get around this.

clintonb: Filter in Python.

schenedx: @clintonb I tried that. Filter in Python won't save queries here. What happens is, when filter in python, you have to first get the set of Programs associated with the course, even if that set is empty. That get call is already a extra query down in DB. It has essentially the same effect of the exclude call. What's worse is that, if we only filter in python, if there are programs associated with the course, the query count will increase by additional 2 for getting Partners and getting ProgramType from DB

PR #382, opened 2016-10-13
Added new course run form.

waheedahmed:
[ECOM-5175](https://openedx.atlassian.net/browse/ECOM-5175)

As a course team member, I would like to have help text and examples on the Course run data entry form so that it's easy for me to complete the form.

**Design:**
http://o7fq9h.axshare.com/new_run_1.html

**Acceptance Criteria:**
1. Validate that the data entry form contains help text and examples as shown in the wireframe above.

**Screenshot:**
screencapture-localhost-8004-publisher-course_runs-new-1477393650563


waheedahmed:
@awais786 @jareerahsan @ahsan-ul-haq @tasawernawaz Please review.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,64 @@ def get_context_data(self, **kwargs):
...
+        seat_form = self.seat_form(request.POST)
+        course = ctx.get('parent_course')
+        course_number = course_form.data['number']
+        if course and course_number and run_form.is_valid() and seat_form.is_valid():
+            with transaction.atomic():
+                seat = seat_form.save(commit=False)
+                course_run = run_form.save(commit=False)
+                user = self.request.user
+
+                team_admin = get_object_or_404(User, pk=course_form.data['team_admin'])
```
tasawernawaz: We can check this at start of `if` condition.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,64 @@ def get_context_data(self, **kwargs):
...
+        return context
+
+    def post(self, request, *args, **kwargs):
+        ctx = self.get_context_data()
+        course_form = self.course_form(request.POST)
+        run_form = self.run_form(request.POST)
+        seat_form = self.seat_form(request.POST)
+        course = ctx.get('parent_course')
+        course_number = course_form.data['number']
+        if course and course_number and run_form.is_valid() and seat_form.is_valid():
```
tasawernawaz: We can also check here `course_form` validation.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -309,38 +311,56 @@ def _pop_valuse_from_dict(self, data_dict, key_list):
...
+        response = self.client.post(
+            reverse('publisher:publisher_course_runs_new', kwargs={'parent_course_id': self.course.id}),
+            self.course_run_dict
+        )
+        self.assertEqual(response.status_code, 400)
+
+    def test_create_course_run_and_seat(self):
+        """ Verify that we can create a new course run with seat. """
+        post_data = model_to_dict(self.course)
         self.course_run_dict['start'] = '2050-07-08 05:59:53'
```
tasawernawaz: Avoid using hardcoded dates.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,199 @@
...
+        </p>
+    </div>
+
+<div class="alert-messages">
+{% if messages %}
+{% for message in messages %}
+<div class="alert alert-{{ message.tags }}" role="alert" aria-labelledby="alert-title-{{ message.tags }}" tabindex="-1">
+    <div class="alert-message-with-action">
+        <p class="alert-copy">
+            {{ message }}
```
tasawernawaz: Do we need to translate `message` ?

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,199 @@
...
+</div>
+
+</form>
+</div>
+</div>
+
+{% endblock content %}
+
+{% block extra_js %}
+    <link rel="stylesheet" href="//code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
```
tasawernawaz: Why we are adding `.css` file under js block ?

tasawernawaz (changes requested):

course_discovery/templates/publisher/add_courserun_form.html
waheedahmed: @tasawernawaz No, we are already translating them at backend.

course_discovery/apps/publisher/views.py
waheedahmed: @tasawernawaz No need to check `course_form` validation, because we are only creating course run and seat here against course object.

course_discovery/templates/publisher/add_courserun_form.html
waheedahmed: @tasawernawaz Actually we don't need `jquery-ui.css` now, removing. Good catch.

course_discovery/apps/publisher/views.py
waheedahmed: No need to do that, we will always get a value for this from parent course.

waheedahmed:
@tasawernawaz Please take another look.

tasawernawaz (approved):

waheedahmed:
@edx/ecommerce Please Review.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -289,16 +291,19 @@ class CreateUpdateCourseRunViewTests(TestCase):
 
     def setUp(self):
         super(CreateUpdateCourseRunViewTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.course = factories.CourseFactory(team_admin=self.user)
         self.course_run = factories.CourseRunFactory()
```
clintonb, line 307: It's not clear why you are keeping this around as an instance variable. If you need a dictionary, see http://factoryboy.readthedocs.io/en/latest/recipes.html#converting-a-factory-s-output-to-a-dict.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -289,16 +291,19 @@ class CreateUpdateCourseRunViewTests(TestCase):
...
+        self.course_run_dict.update({'number': self.course.number, 'team_admin': self.course.team_admin.id})
         self._pop_valuse_from_dict(
             self.course_run_dict,
             [
                 'end', 'enrollment_start', 'enrollment_end',
                 'priority', 'certificate_generation', 'video_language'
             ]
         )
-        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.course_run_dict['start'] = str(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
```
clintonb: 1. There is no need to cast a string to a string.
2. Use single quotes.

course_discovery/apps/publisher/urls.py
```diff
@@ -9,9 +9,13 @@
     url(r'^courses/new$', views.CreateCourseView.as_view(), name='publisher_courses_new'),
     url(r'^courses/(?P<pk>\d+)/view/$', views.ReadOnlyView.as_view(), name='publisher_courses_readonly'),
     url(r'^courses/(?P<pk>\d+)/edit/$', views.UpdateCourseView.as_view(), name='publisher_courses_edit'),
+    url(
+        r'^courses/(?P<parent_course_id>\d+)/course_runs/new$',
```
clintonb: Include a trailing slash.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,68 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class CreateCourseRunView(mixins.LoginRequiredMixin, mixins.FormValidMixin, CreateView):
+class CreateCourseRunView(CreateCourseView):
```
clintonb: You're definitely going to need to update the docstring explaining why you're inheriting from the seemingly unrelated `CreateCourseView`.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,68 @@ def get_context_data(self, **kwargs):
...
+        context['course_form'] = self.course_form(instance=parent_course)
+        return context
+
+    def post(self, request, *args, **kwargs):
+        ctx = self.get_context_data()
+        course_form = self.course_form(request.POST)
+        run_form = self.run_form(request.POST)
+        seat_form = self.seat_form(request.POST)
+        course = ctx.get('parent_course')
+        course_number = course_form.data['number']
```
clintonb: Shouldn't course number come from the course? Why is it coming from the form?

course_discovery/apps/publisher/views.py
clintonb: If there is no need to validate the form, why is it being used? It seems you should be using data from the `Course` that is specified on the URL. If that is the case, do that instead of dealing with an unnecessary form.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,68 @@ def get_context_data(self, **kwargs):
...
+            try:
+                with transaction.atomic():
+                    seat = seat_form.save(commit=False)
+                    course_run = run_form.save(commit=False)
+                    user = self.request.user
+
+                    team_admin = get_object_or_404(User, pk=course_form.data['team_admin'])
+
+                    course.team_admin = team_admin
+                    course.number = course_number
```
clintonb: This doesn't seem that it should ever be changing once created.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,68 @@ def get_context_data(self, **kwargs):
...
+                    seat.save()
+
+                    # pylint: disable=no-member
+                    success_msg = _('Course run created successfully for course "{course_title}".').format(
+                        course_title=course.title
+                    )
+                    messages.success(request, success_msg)
+                    return HttpResponseRedirect(reverse(self.success_url, kwargs={'pk': course_run.id}))
+            except Exception as e:  # pylint: disable=broad-except
+                messages.error(request, str(e))
```
clintonb: Avoid displaying exception messages to users. It's just noise. You should do two things:
1. Display user-friendly error message. This may include the basic exception message, but not the full stack trace.
2. ALWAYS use [`logging.exception()`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) to log the error and stack trace. The value passed to this method should be a developer-friendly message (e.g. "Unable to save course run for course %s."). The method will handle logging the exception and stack trace.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +161,68 @@ def get_context_data(self, **kwargs):
...
+        ctx = self.get_context_data()
+        course_form = self.course_form(request.POST)
+        run_form = self.run_form(request.POST)
+        seat_form = self.seat_form(request.POST)
+        course = ctx.get('parent_course')
+        course_number = course_form.data['number']
+        if course and course_number and run_form.is_valid() and seat_form.is_valid():
+            try:
+                with transaction.atomic():
+                    seat = seat_form.save(commit=False)
```
clintonb: What's the point of these two calls with `commit=False`? If there is extra processing that needs to be done, do so on the forms, not here. That way your forms are truly re-usable and independent of the parent view.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,197 @@
+{% extends 'base.html' %}
+{% load i18n %}
+{% load staticfiles %}
+{% block title %}
```
clintonb, line 5: Add a newline to keep this legible.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,197 @@
...
+            {% url 'publisher:publisher_courses_readonly' parent_course.id as course_url %}
+            {% with link_start='<a href="' link_middle='">' link_end='</a>' %}
+            {% blocktrans %}
+                The fields below will only affect this new course run. If you feel the need edit information from the parent course then you may do so on the {{ link_start }}{{ course_url }}{{ link_middle }}parent course for this run{{ link_end }}
+            {% endblocktrans %}
+            {% endwith %}
+        </p>
+    </div>
+
+<div class="alert-messages">
```
clintonb: This should be inside the if-clause. This entire block should be moved to a partial that can be shared by multiple pages.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,197 @@
...
+    <div class="alert-message-with-action">
+        <p class="alert-copy">
+            {{ message }}
+        </p>
+    </div>
+</div>
+{% endfor %}
+{% endif %}
+
+<form class="form" method="post" action="" enctype="multipart/form-data">
```
clintonb: Either include the action, or remove the attribute.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,197 @@
...
+        </div>
+
+    </fieldset>
+</div>
+</div>
+</div>
+
+<div class="layout-full layout publisher-layout">
+<div class="card course-form">
+<div class="course-information">
```
clintonb: Add some tabs. The lack of tabs makes this a bit confusing to read.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,197 @@
...
+</form>
+</div>
+</div>
+
+{% endblock content %}
+
+{% block extra_js %}
+    <script src="{% static 'js/publisher/publisher.js' %}"></script>
+    <script type="text/javascript">
+        $(document).ready(function () {
```
clintonb: Why is this not in the above file, or its own file?

clintonb (changes requested):

course_discovery/apps/publisher/tests/test_views.py
waheedahmed: @clintonb Yes, we need a dictionary but also need the object in tests to create child objects like seats and need to verify the values.

course_discovery/apps/publisher/views.py
waheedahmed: @clintonb We are using two fields `course_form.number` and `course_form.team_admin` from`course_form` and auto populating their values and user can change values of those fields according to wireframes, we can't call validate on that form because it will always going to fail due to only two fields.

course_discovery/apps/publisher/views.py
waheedahmed: @clintonb It can be changed according to wireframes.

course_discovery/apps/publisher/views.py
waheedahmed: @clintonb It can be changed according to wireframes, adding if-clause to check if changed than assign.

course_discovery/apps/publisher/views.py
waheedahmed: @clintonb Need to assign `ForeignKey` objects before committing them to DB.

course_discovery/apps/publisher/views.py
waheedahmed: @clintonb Inherited it from `CreateCourseView` to reuse the code but I end up in a situation where I can't use the code from it and need to override the functionality, removing it.

course_discovery/apps/publisher/views.py
waheedahmed: Refactored this view with custom save method for forms.

waheedahmed:
@clintonb Addressed your comments, please take another look.

course_discovery/apps/publisher/forms.py
```diff
@@ -47,8 +48,8 @@ class CustomCourseForm(CourseForm):
     """ Course Form. """
 
     institution = forms.ModelChoiceField(queryset=Group.objects.all(), required=True)
-    title = forms.CharField(label='Course Title', required=True, max_length=255)
-    number = forms.CharField(label='Course Number', required=True, max_length=255)
+    title = forms.CharField(label=_('Course Title'), required=True, max_length=255)
+    number = forms.CharField(label=_('Course Number'), required=True, max_length=255)
```
clintonb: This maximum length differs from that on the model (50). Remove the `max_length` kwargs. The model validation will handle that.

course_discovery/apps/publisher/forms.py
```diff
@@ -76,23 +77,44 @@ class CustomCourseRunForm(CourseRunForm):
...
             'keywords', 'start', 'end', 'length',
             'transcript_languages', 'language', 'min_effort', 'max_effort', 'keywords',
             'contacted_partner_manager', 'target_content', 'pacing_type', 'is_seo_review',
-            'video_language', 'staff',
+            'video_language', 'staff', 'is_self_paced',
         )
 
+    def clean(self):
+        super(CustomCourseRunForm, self).clean()
+        self.cleaned_data['pacing_type'] = CourseRunPacing.Instructor
```
clintonb: Minor: `self.cleaned_data['pacing_type'] = CourseRunPacing.Self if self.cleaned_data['is_self_paced'] else CourseRunPacing.Instructor`

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +166,72 @@ def get_context_data(self, **kwargs):
...
 
 
-class CreateCourseRunView(mixins.LoginRequiredMixin, mixins.FormValidMixin, CreateView):
+class CreateCourseRunView(mixins.LoginRequiredMixin, CreateView):
     """ Create Course Run View."""
     model = CourseRun
-    form_class = CourseRunForm
-    template_name = 'publisher/course_run_form.html'
-    success_url = 'publisher:publisher_course_runs_edit'
+    course_form = CustomCourseForm
```
clintonb: I'm still failing to see why this is here. You are not modifying the course. You don't need a form for it.

course_discovery/apps/publisher/views.py
```diff
@@ -160,15 +166,72 @@ def get_context_data(self, **kwargs):
...
+        parent_course = get_object_or_404(Course, pk=self.kwargs.get('parent_course_id'))
+        context = {
+            'parent_course': parent_course,
+            'course_form': self.course_form(instance=parent_course),
+            'run_form': self.run_form,
+            'seat_form': self.seat_form
+        }
+        return context
+
+    def get(self, request, *args, **kwargs):
```
clintonb: I'm fairly certain you get this for free from `CreateView`.

clintonb (commented):
I am going on vacation until Monday. I don't want to hold up progress; but, the form structure is concerning, along with the usage of custom save methods. You might investigate FormSets to determine if they may be helpful to simplify the logic here.

waheedahmed:
@clintonb We don't need FormSets here because we are not adding multiple objects of a single item, we have three different forms.

course_discovery/apps/publisher/views.py
waheedahmed: @clintonb We are modifying two fields of course.

waheedahmed:
@edx/ecommerce Can anyone else review this PR? Clinton is out until Monday.

waheedahmed:
@bderusha Please take a look on this PR.

course_discovery/apps/publisher/forms.py
```diff
@@ -76,23 +77,43 @@ class CustomCourseRunForm(CourseRunForm):
     contacted_partner_manager = forms.BooleanField(
         widget=forms.RadioSelect(choices=((1, _("Yes")), (0, _("No")))), initial=0, required=False
     )
-    start = forms.DateTimeField(required=True)
+    start = forms.DateTimeField(required=True, label=_('Course start date'))
```
bderusha: NIT: consistency.  Labels first, then required.

waheedahmed:
@clintonb I just realized that this story was only to implement help text and examples for new course run form according to wireframes, so created a new ticket [ECOM-6208](https://openedx.atlassian.net/browse/ECOM-6208) to implement the backend for course run form.

Please take another look on this PR.

course_discovery/apps/publisher/forms.py
```diff
@@ -78,20 +78,22 @@ class CustomCourseRunForm(CourseRunForm):
...
+    end = forms.DateTimeField(required=False, label=_('Course end date'))
     staff = forms.ModelMultipleChoiceField(
         queryset=Person.objects.all(), widget=forms.SelectMultiple, required=False
     )
     target_content = forms.BooleanField(
         widget=forms.RadioSelect(
             choices=((1, _("Yes")), (0, _("No")))), initial=0, required=False
     )
+    is_self_paced = forms.BooleanField(label=_('Yes, course will be Self-Paced'), required=False)
```
clintonb, line 113: Update: "Indicate if this course will be self-paced or instructor-paced".

course_discovery/apps/publisher/views.py
```diff
@@ -164,15 +161,27 @@ def get_context_data(self, **kwargs):
...
-        return reverse(self.success_url, kwargs={'pk': self.object.id})
+    def get_context_data(self, **kwargs):
+        context = {
+            'course_form': self.course_form,
+            'run_form': self.run_form,
+            'seat_form': self.seat_form
+        }
+        return context
+
+    def post(self, request, *args, **kwargs):
```
clintonb, line 216: Remove this until you get to ECOM-6208.

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,222 @@
...
+                        <div class="copy-base">
+                            <p>
+                                {% blocktrans %}
+                                    The information in this section is required to create a course studio instance. You must fill all required information but are welcome to come back and enter the rest of the information when you are ready to announce the course, or you can do it all at once if you are ready to do so.
+                                {% endblocktrans%}
+                            </p>
+                        </div>
+                        <fieldset class="form-group grid-container grid-manual">
+
+                            <div class="field-title">{% trans "COURSE TITLE" %}</div>
```
clintonb, line 40: Why is this in all-caps?

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,222 @@
...
+                                        <p>{% trans 'If the course falls in a sequence, our titling convention is: "Course Title: Subtitle"' %}</p>
+                                    </div>
+                                    <div id="tab-example" class="content">
+                                        <p>"{% trans "English Grammar and Essay Writing" %}"</p>
+                                        <p>{% trans "Sequence Courses:" %}</p>
+                                        <p>"{% trans "Introduction to Statistics" %}"</p>
+                                        <p>"{% trans "Statistics: Inference" %}"</p>
+                                        <p>"{% trans "Statistics: Probability" %}"</p>
+                                        <div class="margin-top20">
+                                            <a href="#">{% trans "View our documentation for more help" %}</a>
```
clintonb: Where is this supposed to go?

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,222 @@
...
+                                    <div class="field-admin margin-top20">
+                                        <label class="field-label">
+                                            {% trans "Institution Course Admin" %}
+                                        </label>
+                                        {{ course_form.team_admin }}
+                                    </div>
+                                </div>
+                            </div>
+
+                            <div class="field-title">{% trans "COURSE START DATE" %}</div>
```
clintonb, line 102: Why is this in all-caps?

course_discovery/templates/publisher/add_courserun_form.html
```diff
@@ -0,0 +1,222 @@
...
+                                        <p>"{% trans "Statistics: Probability" %}"</p>
+                                        <div class="margin-top20">
+                                            <a href="#">{% trans "View our documentation for more help" %}</a>
+                                        </div>
+                                    </div>
+                                </div>
+                                <div class="col col-6">
+
+                                    <label class="field-label ">{{ course_form.title.label }}
+                                        <span class="required">* {% trans "required" %}</span>
```
clintonb: Put this inside a conditional:

``` Django
{% if course_form.title.field.required %}<span class="required">* {% trans "required" %}</span>{% endif %}
```

clintonb (changes requested):
It is strange to me that the UI is being implemented without the actual form backend (now moved to ECOM-6208). That order of operations removes the context necessary to properly review this request. I recommend getting the backend in order, and returning to this once that work is done. 

Without a backend, this work is incomplete; and, I would prefer not to merge incomplete code to the master branch. A feature branch should be used in such cases.

course_discovery/apps/publisher/forms.py
waheedahmed: @clintonb Its in the provided wireframes.

course_discovery/templates/publisher/add_courserun_form.html
waheedahmed: @clintonb Its according to provided wireframes.

course_discovery/templates/publisher/add_courserun_form.html
waheedahmed: @clintonb Its according to provided wireframes.

waheedahmed:
@clintonb Please take final look, created a PR https://github.com/edx/course-discovery/pull/401 for backend against this branch.

bderusha (approved):

bderusha:
@waheedahmed Please resolve the conflicts and squash the commits

PR #383, opened 2016-10-13
Awais786/ecom 6003 course cleanup

awais786:
Cleanup story for create course form.
- Hide time from the datetime widget.
- Move is_seo_review and keywords ( db fields ) from course-run to course table.
- Add JS against Seat Type drop and disable price box for honor and audit seat.
- Change Brief description from textbox to text area widget.


awais786:
@waheedahmed @ahsan-ul-haq Please review.

waheedahmed (approved):
Changes looks good to me except one thing, don't hide the price field just enable/disable it on seat type selection.

awais786:

course_discovery/apps/publisher/models.py
```diff
@@ -129,6 +129,12 @@ class Course(TimeStampedModel, ChangedByMixin):
         }
     )
 
+    is_seo_review = models.BooleanField(default=False)
+
+    keywords = models.TextField(
```
clintonb: This might be a good place to use django-taggit, which is already installed.

course_discovery/apps/publisher/models.py
```diff
@@ -129,6 +129,12 @@ class Course(TimeStampedModel, ChangedByMixin):
...
+        null=True, blank=True, help_text=_("Please add top 10 comma separated keywords for your course content")
```
clintonb: Add the top 10 keywords for your course content. The keywords should be separated by a comma.

clintonb: If you don't use taggit, it would be ideal if you used a widget to enforce the 10 entry limit and standardize comma separation.

course_discovery/apps/publisher/migrations/0012_auto_20161013_0950.py
```diff
@@ -0,0 +1,66 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('publisher', '0011_userattributes'),
+    ]
+
+    operations = [
+
+        # This is custom alteration to make migration reversible
```
clintonb: This is confusing. Identify what, if anything, you changed. Django nearly always generates reversible migrations unless the developer generating them makes an override. If the previous migration for this field is irreversible, that migration should be updated.

awais786: @clintonb at the moment model field is like this.

```
keywords = models.TextField(
        default=None, blank=True, help_text=_("Please add top 10 comma separated keywords for your course content")
)
```

At the moment null=False. And If the table contains some data then after removing the field it becomes the non reversible migration. Because when migration tries to add back the column in db with existing rows it needs some default value otherwise it throws error.

`django.db.utils.IntegrityError: NOT NULL constraint failed: publisher_courserun__new.keywords
`

awais786: You can check this link also [revert-django-1-7-removefield-migration](http://stackoverflow.com/questions/28878428/revert-django-1-7-removefield-migration).

awais786: [Alters a field’s with preserve_default.](https://docs.djangoproject.com/en/1.10/ref/migration-operations/#alterfield)

awais786:

bderusha (approved):
👍 once tests are passing

PR #384, opened 2016-10-14
Increase the default facet organizations field limit from 100 to 10000

schenedx:
Fix the bug where we only load 100 organization in our `Schools and Partners` dropdown on the search results page on the right panel. We have 215 organizations now. We need to load all of them
BM-6105

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -4,6 +4,7 @@ import json
 from urllib.parse import urlencode
 
 import pytz
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db.models.query import Prefetch
 from django.utils.translation import ugettext_lazy as _
@@ -29,7 +30,7 @@ COMMON_SEARCH_FIELD_ALIASES = {
 }
 COURSE_RUN_FACET_FIELD_OPTIONS = {
     'level_type': {},
-    'organizations': {},
+    'organizations': {'size': settings.SEARCH_FACET_LIMIT},
     'prerequisites': {},
     'subjects': {},
     'language': {},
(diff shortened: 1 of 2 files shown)


schenedx:
@edx/ecommerce Please review

course_discovery/apps/api/serializers.py
```diff
@@ -27,9 +27,12 @@
 COMMON_SEARCH_FIELD_ALIASES = {
     'q': 'text',
 }
+
+FACET_LIMIT = 10000
```
clintonb: 1. Move this to a setting.
2. Can we use -1, or some other value, to retrieve all items?

schenedx: @clintonb Setting sounds fine. I'll do that. We cannot use `-1`. I tested it locally. `-1` means `0` for elasticsearch. I am hesitant to use `sys.maxsize` because I don't know if that will introduce a performance issue.

course_discovery/settings/base.py
```diff
@@ -362,6 +362,8 @@
 
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
 
+# Elasticsearch search query facet "size" option to increase from the default value of "100"
```
clintonb, line 365: Include a link to the appropriate docs as well: https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-facets-terms-facet.html#_accuracy_control.

course_discovery/apps/api/serializers.py
```diff
@@ -27,9 +28,11 @@
 COMMON_SEARCH_FIELD_ALIASES = {
     'q': 'text',
 }
+
```
clintonb: Revert.

clintonb (approved):

PR #386, opened 2016-10-17
Exclude unpublished programs from the course run endpoint

mikedikan:
ECOM-6013

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -389,6 +389,9 @@ class CourseRunWithProgramsSerializer(CourseRunSerializer):
                     if (self.context.get('include_deleted_programs') or
                         program.status != ProgramStatus.Deleted) and
                     obj.id not in (run.id for run in program.excluded_course_runs.all())]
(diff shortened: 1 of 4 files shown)


clintonb (approved):

MatthewPiatetsky:
👍

eagleofpc:
@mikedikan I am sorry I got to this PR after it's merged.
However, I think you should consider adding test cases to https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py#L20 to ensure:

1. From the view perspective, the query string flag is working as it should be
2. Assert the number of queries to ensure we do not incur extra query counts

schenedx:
@mikedikan I am sorry I got to this PR after it's merged.
However, I think you should consider adding test cases to https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py#L20 to ensure:

From the view perspective, the query string flag is working as it should be
Assert the number of queries to ensure we do not incur extra query counts

PR #387, opened 2016-10-18
ECOM-5946 Allow course length in program to be represented as a range of weeks

MatthewPiatetsky:
@edx/ecommerce 
https://openedx.atlassian.net/browse/ECOM-5946

Current Issue: This causes the number of queries to change in the tests. 
I'm not sure why this is because I see the same number in the django toolbar and the course runs are already being prefetched


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -449,6 +450,9 @@ def create_course_run(self, course, data):
         language = language_tags[0] if language_tags else None
         start = data.get('field_course_start_date')
         start = datetime.datetime.fromtimestamp(int(start), tz=pytz.UTC) if start else None
+        end = data.get('field_course_end_date')
+        end = datetime.datetime.fromtimestamp(int(end), tz=pytz.UTC) if end else None
+        weeks_to_complete = data.get('field_course_required_weeks')
```
clintonb, line 455: This should be cast to an integer.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -464,6 +468,12 @@ def create_course_run(self, course, data):
             'hidden': self.get_hidden(data),
         }
 
+        if weeks_to_complete:
+            defaults['weeks_to_complete'] = weeks_to_complete
+        elif start and end:
```
clintonb, line 473: How likely is this clause to be invoked? In other words, how many courses, if any, are missing data for `field_course_required_weeks`? Given that the marketing site is not our source of truth for the end date, I can see this becoming problematic.

course_discovery/apps/course_metadata/models.py
```diff
@@ -352,6 +352,9 @@ class CourseRun(TimeStampedModel):
     max_effort = models.PositiveSmallIntegerField(
         null=True, blank=True,
         help_text=_('Estimated maximum number of hours per week needed to complete a course run.'))
+    weeks_to_complete = models.PositiveSmallIntegerField(
+        null=True, blank=True,
+        help_text=_('Estimated number of weeks needed to complete a course run.'))
```
clintonb: `'Estimated number of weeks needed to complete this course run'`

clintonb: Also, there is no migration for this field.

course_discovery/apps/course_metadata/models.py
```diff
@@ -593,7 +596,11 @@ class Program(TimeStampedModel):
     excluded_course_runs = models.ManyToManyField(CourseRun, blank=True)
     partner = models.ForeignKey(Partner, null=True, blank=False)
     overview = models.TextField(null=True, blank=True)
-    weeks_to_complete = models.PositiveSmallIntegerField(null=True, blank=True)
+    # The weeks_to_complete field is now deprecated
+    weeks_to_complete = models.PositiveSmallIntegerField(
```
clintonb, line 607: Is there a task to remove this field? It should be removed fairly quickly after this is released. I see no reason to keep this around for any amount of time beyond the initial release.

course_discovery/apps/course_metadata/models.py
```diff
@@ -630,6 +637,20 @@ class Program(TimeStampedModel):
     def __str__(self):
         return self.title
 
+    def _course_run_weeks_to_complete(self):
```
clintonb, line 648: `@cached_property`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -523,7 +524,8 @@ def create_program(self):
 
         courses = CourseFactory.create_batch(3)
         for course in courses:
-            CourseRunFactory.create_batch(2, course=course, staff=[person], start=datetime.now())
+            CourseRunFactory.create_batch(2, course=course, staff=[person], start=datetime.now(),
+                                          weeks_to_complete=randrange(1, 10))
```
clintonb: This is not necessary given your factory changes.

clintonb (changes requested):

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
MatthewPiatetsky: I believe that the majority of courses do not have field_course_required_weeks

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: what about 0031_courserun_weeks_to_complete.py?
will change

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: There was no task currently to remove it.
Should it be removed as part of this PR?
Or should I create a ticket to remove this field?

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
MatthewPiatetsky: updated

course_discovery/apps/course_metadata/models.py
```diff
@@ -628,7 +636,22 @@ class Program(TimeStampedModel):
     objects = ProgramQuerySet.as_manager()
 
     def __str__(self):
-        return self.title
+        return self.title_override
+
+    @cached_property
+    def _course_run_weeks_to_complete(self):
```
MatthewPiatetsky, line 648: @clintonb 
I think this property is the thing generating extra queries.
Is that ok if the caching will cause it to not scale with the size of the input?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -592,6 +592,8 @@ def get_expected_data(self, program, request):
             'job_outlook_items': [item.value for item in program.job_outlook_items.all()],
             'languages': [serialize_language_to_code(l) for l in program.languages],
             'weeks_to_complete': program.weeks_to_complete,
+            'weeks_to_complete_min': program.weeks_to_complete_min,
```
bderusha, line 595: didn't we talk about making this more accurately named to something that makes it clear that it is talking about an individual course completion?

course_discovery/apps/course_metadata/models.py
schenedx: @MatthewPiatetsky Create a new devtask story to remove. We should not remove db columns as part of the existing deployment. Let's only remove db columns when the new code is in production and stable.

bderusha: Create a ticket please

course_discovery/apps/course_metadata/models.py
```diff
@@ -630,6 +638,21 @@ class Program(TimeStampedModel):
...
         return self.title
 
+    @cached_property
+    def _course_run_weeks_to_complete(self):
+        return [course_run.weeks_to_complete for course_run in self.course_runs
+                if course_run.weeks_to_complete is not None]
+
+    @property
+    def weeks_to_complete_min(self):
+        course_run_weeks_to_complete = self._course_run_weeks_to_complete
```
schenedx: #nitpick Why the temp local variable while you can just reference `self._course-run_weeks_to_complete` directly?

bderusha (approved):
A few minor comments, but otherwise looks good.  👍

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -76,15 +76,15 @@ def test_retrieve_with_sorting_flag(self, order_courses_by_start_date):
         for course in course_list:
             CourseRunFactory(course=course)
         program = ProgramFactory(courses=course_list, order_courses_by_start_date=order_courses_by_start_date)
-        with self.assertNumQueries(82):
+        with self.assertNumQueries(87):
```
schenedx, line 79: Could you explain why the number of query increase here? If not, can you dig in and see if we can avoid it?

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: I didn't want to incur extra queries, but now that I'm using cached_property this is fine.
addressed.

MatthewPiatetsky: addressed.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
MatthewPiatetsky: This bit of code here is generating extra queries.

```
    @cached_property
    def _course_run_weeks_to_complete(self):
        return [course_run.weeks_to_complete for course_run in self.course_runs
                if course_run.weeks_to_complete is not None]

```

I talked to @clintonb about it, and given that it is a limited number of additional queries and does not affect the Program List endpoint he was ok with it.
We can't actually prefetch properties that are calculated at runtime (though the caching helps a bit). If you have other ideas let me know.

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: I forget, but do you have thoughts on a better name?
We could do course_run_min_weeks_to_complete though that might be a bit too verbose and still confusing

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: ticket https://openedx.atlassian.net/browse/ECOM-6170

schenedx (approved):
👍

MatthewPiatetsky:
Thank you @bderusha and @schenedx for the reviews, I think I addressed @clintonb's comments, so I will merge, but happy to address further comments when you're back.  
will do further testing first and make sure I'm clear on the requirements

PR #389, opened 2016-10-18
Update Search Indexer to exclude unpublished programs

mikedikan:
Exclude when nested unpublished programs are under course runs on course run search endpoint.

ECOM-6147

@schenedx & @clintonb for review

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -366,7 +366,9 @@ class CourseRun(TimeStampedModel):
 
     @property
     def program_types(self):
-        return [program.type.name for program in self.programs.all()]
+        # Exclude unpublished programs from list so we don't identify that program type if not available
+        all_programs = [program for program in self.programs.all() if program.status != ProgramStatus.Unpublished]
+        return [program.type.name for program in all_programs]
 
     @property
     def marketing_url(self):
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -366,8 +366,12 @@ class CourseRun(TimeStampedModel):
 
     @property
     def program_types(self):
-        # Exclude unpublished programs from list so we don't identify that program type if not available
-        all_programs = [program for program in self.programs.all() if program.status != ProgramStatus.Unpublished]
+        """
+        Exclude unpublished and deleted programs from list
+        so we don't identify that program type if not available
+        """
+        program_statuses_to_exclude = (ProgramStatus.Unpublished, ProgramStatus.Deleted)
+        all_programs = [program for program in self.programs.exclude(status__in=program_statuses_to_exclude)]
         return [program.type.name for program in all_programs]
 
     @property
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -163,6 +163,13 @@ def test_program_types(self):
         other_program = factories.ProgramFactory(courses=courses)
         self.assertCountEqual(self.course_run.program_types, [program.type.name, other_program.type.name])
 
+    def test_unpublished_program_types(self):
+        """ Verify the property exludes program types that are unpublished. """
+        courses = [self.course_run.course]
+        program = factories.ProgramFactory(courses=courses)
```
schenedx, line 169: I think it would be easier we combine line 168 and 169 with just `factories.ProgramFactory(courses=[self.course_run.course])`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -163,6 +163,13 @@ def test_program_types(self):
...
+        other_program = factories.ProgramFactory(courses=courses, status=ProgramStatus.Unpublished)
```
schenedx: You can just do `factories.ProgramFactory(courses=[self.course_run.course], status=ProgramStatus.Unpublished)`

schenedx (changes requested):
I think it's worth adding another view level test to ensure the correctness as well as the query count

course_discovery/apps/course_metadata/models.py
```diff
@@ -366,7 +366,9 @@ class CourseRun(TimeStampedModel):
 
     @property
     def program_types(self):
-        return [program.type.name for program in self.programs.all()]
+        # Exclude unpublished programs from list so we don't identify that program type if not available
+        all_programs = [program for program in self.programs.all() if program.status != ProgramStatus.Unpublished]
```
clintonb: Combine into a single loop.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -163,6 +163,13 @@ def test_program_types(self):
...
+        self.assertCountEqual(self.course_run.program_types, [program.type.name])
```
clintonb: What does count matter? Check the actual values.

clintonb (changes requested):

course_discovery/apps/course_metadata/models.py
clintonb: Also, move the comment to a docstring for the property.

mikedikan: Why combine into a single loop? This seems easier to read...

course_discovery/apps/course_metadata/tests/test_models.py
mikedikan: I was mimicking the approach in the method above - should we change both places so it's consistent?

course_discovery/apps/course_metadata/tests/test_models.py
mikedikan: See comment above

mikedikan:
@clintonb @schenedx please rereview

schenedx (approved):
👍

PR #393, opened 2016-10-21
Fix migrations issue

MatthewPiatetsky:
@edx/ecommerce 

I should have made the changes below in a new migration. I forgot they already automatically ran against staging.
https://github.com/edx/course-discovery/pull/392/commits/cab5ff0a9afaab43551ca30bde2e32cf323441ef
I think this should fix the issue

### DIFF
--- a/course_discovery/apps/course_metadata/migrations/0031_auto_20161020_1814.py
+++ b/course_discovery/apps/course_metadata/migrations/0031_auto_20161020_1814.py
@@ -14,11 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='courserun',
             name='weeks_to_complete',
-            field=models.PositiveSmallIntegerField(blank=True, help_text='Estimated number of weeks needed to complete this course run.', null=True),
-        ),
-        migrations.AlterField(
-            model_name='program',
-            name='weeks_to_complete',
-            field=models.PositiveSmallIntegerField(blank=True, help_text='This field is now deprecated (ECOM-6021).Estimated number of weeks needed to complete a course run belonging to this program.', null=True),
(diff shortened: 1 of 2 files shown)


mikedikan (approved):

course_discovery/apps/course_metadata/migrations/0031_auto_20161020_1814.py
```diff
@@ -14,11 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='courserun',
             name='weeks_to_complete',
-            field=models.PositiveSmallIntegerField(blank=True, help_text='Estimated number of weeks needed to complete this course run.', null=True),
```
schenedx, line 17: Has this migration been applied? If so, why the change?

schenedx: Updating an applied migration may not take any effect, or fail the migration. Are you sure this does not fail the `/.manage.py migrate` command on machines where `0031_auto_20161020_1814.py` is already applied?

MatthewPiatetsky: If you look at the commit I mentioned in the description, I previously tried to edit a migration that had already been applied on staging, so I am changing this back to what it was previously, and putting the changes in a new migration

MatthewPiatetsky: (I am doing this because the commit I mentioned is already breaking migrations on staging in that the previous version of 0031 was already applied)

schenedx: I see. Thanks for the explanation.

schenedx (approved):
👍

PR #394, opened 2016-10-21
ECOM-5040 Search in Mobile

MatthewPiatetsky:
@edx/ecommerce


mikedikan (approved):

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -157,6 +157,7 @@ def update_course_run(self, course, body):
                 'video': self.get_courserun_video(body),
                 'status': CourseRunStatus.Published,
                 'pacing_type': self.get_pacing_type(body),
+                'mobile_available': body['mobile_available'],
```
schenedx: Would this throw an exception if the key `mobile_available` is not there? Maybe use `.get()` is better?

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -485,6 +485,7 @@ def assert_course_run_loaded(self, data):
             'start': start,
             'pacing_type': self.loader.get_pacing_type(data),
             'hidden': self.loader.get_hidden(data),
+            'mobile_available': data['field_course_enrollment_mobile'] or False,
```
schenedx, line 488: do `data.get('field_course_enrollment_mobile', False)`

schenedx (approved):
👍

course_discovery/apps/course_metadata/data_loaders/api.py
MatthewPiatetsky: good idea

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
MatthewPiatetsky: I will do `data.get('field_course_enrollment_mobile') or False`
`data.get('field_course_enrollment_mobile', False)` would set the field to null if the value was present and null, right now the field defaults to False and is not nullable..though I'm not sure what would be better, this, or allowing nulls in the boolean field.  
In my personal opinion if we don't know if a field is mobile_available, then for our purposes, it is not mobile_available

schenedx: I don't understand what the difference is. If the `data.get()` returns null, then, the code will assign `False` anyways.

schenedx: I don't have a strong opinion either case.

MatthewPiatetsky: I don't know about the model, was referring to this difference  

```
>>> a = {'b': None}
>>> print(a.get('b', False))
None
>>> print(a.get('b') or False)
False
```

course_discovery/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
```
clintonb, line 7: You don't need to commit translations if you didn't change anything.

clintonb (approved):

course_discovery/conf/locale/en/LC_MESSAGES/django.po
MatthewPiatetsky: I think an earlier travis build on this PR failed on the validate_translations step which is why I ended up doing this

clintonb: What was the error?

MatthewPiatetsky:
```
Source translations are out-of-date! Please update them.
make: *** [detect_changed_source_translations] Error 1
```

That's what I just got locally when I ran 

```
git checkout master course_discovery/conf/locale/
git commit -am "undo translations commit"
make validate_translations
```

PR #395, opened 2016-10-21
Try again to fix migrations

MatthewPiatetsky:
@edx/ecommerce 
Is there a way I can deploy a branch to stage catalog to test if my migration fix worked?
I think the issue with my previous PR was that I reverted my migration changed but didn't rename the file back.


schenedx:
@MatthewPiatetsky I think you should investigate whether we can setup discovery service in a sandbox. See https://openedx.atlassian.net/wiki/display/EdxOps/Sandboxes.

schenedx (approved):
👍  But what's with travis build on this PR?

MatthewPiatetsky:
@schenedx all my travis builds were failing due to timeouts with github
This time around the log message is unclear, but think it is related to that.
I had to restart the build like 5 times on my previous PR to get it to pass.

cpennington:
@MatthewPiatetsky @schenedx: What was it about your dev environments that made it so you couldn't reproduce this issue locally (and had to rely on stage/a sandbox)?

PR #396, opened 2016-10-24
Added email and name attributes to UserFactory

clintonb:
The UserFactory now generates an email address and name—full, first, and last—for every user.

ECOM-5929


clintonb:
FYI @edx/ecommerce

PR #397, opened 2016-10-25
Publisher app global admin flag

awais786:
Create the is_global_admin flag in UserAttributes table. 
User having this value will be a super-user for publisher app.
Update the context processor for this new field.

https://openedx.atlassian.net/browse/ECOM-5950

### DIFF
--- a/course_discovery/apps/publisher/migrations/0013_create_admin_group.py
+++ b/course_discovery/apps/publisher/migrations/0013_create_admin_group.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+ADMIN_GROUP_NAME = 'Publisher Admins'
+
+
+def create_admin_group(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    Group.objects.get_or_create(name=ADMIN_GROUP_NAME)
+
+
+def remove_admin_group(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    Group.objects.filter(name=ADMIN_GROUP_NAME).delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('publisher', '0012_auto_20161020_0718'),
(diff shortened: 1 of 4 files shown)


awais786:
@ahsan-ul-haq @tasawernawaz @waheedahmed Please review.

clintonb (changes requested):
Why the user attribute instead of a Django group?

awais786:

awais786:
@clintonb This user will be the super user for the publisher app only. That's why add the field in UserAttributes table. This table will specifically hold the value for publisher app users.

@bderusha FYI

clintonb:
@awais786 I get that; but, on previous pull requests you're using Django Guardian and groups. Now you've added a custom attribute. Using a custom attribute will require you to make custom handlers to process that attribute. If you use a Django group, you can take advantage of all of the functionality that already exists in Django with minimal custom work.

clintonb:
@awais786 I wrote up a quick example of using a group: https://github.com/edx/course-discovery/commit/6ef5118c5faba4e2b9dc3c7d693e8915c705bb71

awais786:
@clintonb commits looks good. I will update my PR that way.

course_discovery/apps/publisher/migrations/0013_create_admin_group.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+ADMIN_GROUP_NAME = 'Publisher Admins'
```
bderusha, line 6: Can we import this from `constants`?

course_discovery/apps/publisher/utils.py
```diff
@@ -14,3 +15,15 @@ def is_email_notification_enabled(user):
         return user.attributes.enable_email_notification
 
     return True
+
+
+def is_global_admin(user):
```
bderusha, line 20: can we call this `is_publisher_admin`

bderusha: and rename the appropriate tests

PR #398, opened 2016-10-25
i18n Cleanup

clintonb:
Line numbers in PO files are noisy in pull requests. This commit updates the Makefile to use edx-i18n-tools to extract translations and remove line numbers.

Additionally, a few uses of `blocktrans` have been updated to use the `trimmed` argument to remove surrounding whitespace.

ECOM-5175


clintonb:
FYI @edx/ecommerce

clintonb:
@waheedahmed this will help reduce the line count in #382.

MatthewPiatetsky:
@clintonb forget if I asked this before, is there documentation somewhere on how the whole translations system works in course_discovery
There is this https://open-edx-course-catalog.readthedocs.io/en/latest/internationalization.html, but it doesn't seem to have many details on course_discovery specifically rather than transifex

clintonb:
@MatthewPiatetsky all of our IDAs behave the same, so the documentation is pretty standard. The changes regarding validation are pretty new, and documentation has not been updated.

PR #402, opened 2016-10-26
Roles with course runs.

awais786:
WIP: Initial db structure for roles

Created table for UserRoles.
Created relationship between Organizations and groups with 3 default roles. 

ECOM-6216


course_discovery/apps/publisher/models.py
```diff
@@ -78,6 +79,37 @@ def publish(self):
...
+        )
+
+    class Meta(object):
+
+        unique_together = (
+            ('user', 'role'),
+        )
+
+
 class Course(TimeStampedModel, ChangedByMixin):
```
bderusha, line 113: We also need this user_role for a `Course`

course_discovery/apps/publisher/models.py
```diff
@@ -354,3 +387,21 @@ def __str__(self):
 
     class Meta:
         verbose_name_plural = 'UserAttributes'
+
+
+class OrganizationGroup(TimeStampedModel):
```
bderusha: Do we need to do this this way as opposed to letting Django handle this for us?

course_discovery/apps/publisher/models.py
awais786: Do we need user_role for both course-run and course ?

course_discovery/apps/publisher/models.py
awais786: How django can handle this ? For org and group we need some kind of relationship.

awais786:
@bderusha  @jareerahsan @waheedahmed I have created db structure. Please review these db changes.

course_discovery/apps/publisher/models.py
awais786: Because stage changes will be occur for course runs only.

clintonb:
Are there design documents stating how roles/permissions are supposed to work? If not, can we get those before we continue with coding? I've now seen 2-3 pull requests dealing with permissions, but no documentation. This documentation will (a) make it easier to understand/implement this functionality, and (b) make it easier for reviewers to validate it.

bderusha:
@awais786 Can we close this PR until we have a data model design document clearly stating our approach?

bderusha:
@awais786 can we close this?

bderusha:
@awais786 I am closing this PR.  Please reopen if you think it should remain open.

PR #403, opened 2016-10-26
ECOM-5777 Squash migrations

MatthewPiatetsky:
Command:

```
time coverage run ./manage.py test course_discovery.apps.api.tests.test_serializers.CourseRunSerializerTests --settings=course_discovery.settings.test
```

Master:

```
Run 1:
real    0m44.579s
user    0m38.363s
sys 0m0.817s

Run 2:
real    0m48.977s
user    0m43.111s
sys 0m1.123s
```

Updated Squashed Migrations:

```
Run 1:
real    0m18.008s
user    0m12.994s
sys 0m0.325s

Run 2:
real    0m18.505s
user    0m15.610s
sys 0m0.377s
```


course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+    SeatType = apps.get_model('course_metadata', 'SeatType')
+
+    for name in SEAT_TYPES:
+        SeatType.objects.update_or_create(name=name)
+
+def drop_seat_types(apps, schema_editor):
+    SeatType = apps.get_model('course_metadata', 'SeatType')
+    SeatType.objects.filter(name__in=SEAT_TYPES).delete()
+
+def update_subjects(apps, schema_editor):
```
clintonb: This was a data fix. This is not needed for new installations; therefore, it does not need to be present in squashed migrations.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+        Partner = apps.get_model('core', 'Partner')
+        partner = Partner.objects.get(id=settings.DEFAULT_PARTNER_ID)
+
+        # We iterate over all subjects, instead of calling .update(), to trigger slug generation
+        for subject in subjects:
+            subject.partner = partner
+            subject.uuid = uuid.uuid4()
+            subject.save()
+
+def update_organizations(apps, schema_editor):
```
clintonb: Same as above. Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+                ('name', models.CharField(max_length=64, unique=True)),
+                ('slug', django_extensions.db.fields.AutoSlugField(blank=True, editable=False, populate_from='name')),
+            ],
+            options={
+                'get_latest_by': 'modified',
+                'abstract': False,
+                'ordering': ('-modified', '-created'),
+            },
+        ),
+        migrations.AddField(
```
clintonb: There should be few, if any, calls to `AddField()`. Django is being a bit conservative with many of these. http://www.rkblog.rk.edu.pl/w/p/squashing-and-optimizing-migrations-django/ has some info; but, ultimately, you will need to modify some of these by hand.

When I last squashed migrations I removed all migrations and generated initial migrations. This provides a baseline from which you can pull operations into your squashed migration. Ideally, you should end up with the following operations:
1. Create models
2. Create circular foreign keys (if any)
3. Perform data operations (if any)

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+        migrations.AlterField(
+            model_name='subject',
+            name='name',
+            field=models.CharField(max_length=255),
+        ),
+        migrations.AlterUniqueTogether(
+            name='subject',
+            unique_together=set([('partner', 'name'), ('partner', 'slug'), ('partner', 'uuid')]),
+        ),
+        migrations.RunPython(
```
clintonb: Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+        ),
+        migrations.RemoveField(
+            model_name='organization',
+            name='banner_image',
+        ),
+        migrations.RemoveField(
+            model_name='organization',
+            name='logo_image',
+        ),
+        migrations.RunPython(
```
clintonb: Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+            model_name='program',
+            name='excluded_course_runs',
+            field=models.ManyToManyField(blank=True, to='course_metadata.CourseRun'),
+        ),
+        migrations.AddField(
+            model_name='historicalorganization',
+            name='marketing_url_path',
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+        migrations.RunPython(
```
clintonb: Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+        ),
+        migrations.AddField(
+            model_name='historicalperson',
+            name='uuid',
+            field=models.UUIDField(default=uuid.uuid4, editable=False, verbose_name='UUID'),
+        ),
+        migrations.AddField(
+            model_name='person',
+            name='family_name',
+            field=mod
```
clintonb: Remove.

clintonb: Remove.

clintonb: Remove.

clintonb: Remove.

clintonb: This should remain.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,1654 @@
...
+            model_name='program',
+            name='job_outlook_items',
+            field=sortedm2m.fields.SortedManyToManyField(blank=True, help_text=None, to='course_metadata.JobOutlookItem'),
+        ),
+        migrations.AddField(
+            model_name='program',
+            name='type',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='course_metadata.ProgramType'),
+        ),
+        migrations.RunPython(
```
clintonb: This should remain.

clintonb (changes requested):

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
MatthewPiatetsky: I see, sounds like I'll need to do a bit more reading into this.
The ticket I looked at https://openedx.atlassian.net/browse/ECOM-5777 provided this link as the instructions https://docs.djangoproject.com/en/1.10/topics/migrations/#migration-squashing
This suggested squashing migrations by using the squashmigrations command.
Are you saying I should instead do this by removing the migrations and generating brand new ones?
Or should I go from the output of the squashmigrations command and tweak it.

clintonb: I'm saying the `squashmigrations` management command is not perfect. In order to get to an ideal case, and minimize database calls, you will need to handcraft some of the migrations.

MatthewPiatetsky: You mentioned that there should be few calls to AddField.
In this migration I see 112 calls to AddField.
Is the issue that some of these are added and removed?
There are 55 calls to RemoveField and not all match an add field, but even then, there would still be many AddField left over. What is the issue with the rest?

course_discovery/apps/edx_catalog_extensions/migrations/0001_create_program_types.py
```diff
@@ -27,7 +27,7 @@ def drop_program_types(apps, schema_editor):
 
 class Migration(migrations.Migration):
     dependencies = [
-        ('course_metadata', '0012_create_seat_types'),
+        ('course_metadata', '0001_squashed_0033_courserun_mobile_available'),
```
MatthewPiatetsky, line 30: Is this what I want to do with these dependencies?

MatthewPiatetsky:
@clintonb What I currently have I got by removing all the course_metadata migrations, creating initial migrations, then adding back the data migrations at the end except the ones you said to remove.
What is the best way to test this locally?
I'm not sure how to get my database in the same state as staging/prod and then run the squashed migrations the same way they will be run on deployment

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,609 @@
...
+                ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified')),
+                ('src', models.URLField(max_length=255, unique=True)),
+                ('description', models.CharField(blank=True, max_length=255, null=True)),
+                ('image', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='course_metadata.Image')),
+            ],
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
```
MatthewPiatetsky: @clintonb the initial migrations when I ran them still generated 18 AddField's, is there a reason these fields need to be in AddField rather than CreateModel?  
Is this due to the order the models are created and the fields depending on a relationship to models that have not be created yet (so it adds the field separately)

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,609 @@
...
+            name='course',
+            unique_together=set([('partner', 'key'), ('partner', 'uuid')]),
+        ),
+        # Data Migrations
+        migrations.RunPython(
+            code=create_switches,
+            reverse_code=delete_switches,
+        ),
+        migrations.RunPython(
+            code=clear_slug_values,
```
clintonb: Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,609 @@
...
+            reverse_code=django.db.migrations.operations.special.RunPython.noop,
+        ),
+        migrations.RunPython(
+            code=remove_duplicate_courses,
```
clintonb: Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,609 @@
...
+            reverse_code=django.db.migrations.operations.special.RunPython.noop,
+        ),
+        migrations.RunPython(
+            code=add_uuid_to_courses_and_course_runs,
```
clintonb: Remove.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
```diff
@@ -0,0 +1,609 @@
...
+            reverse_code=django.db.migrations.operations.special.RunPython.noop,
+        ),
+        migrations.RunPython(
+            code=delete_partnerless_courses,
```
clintonb: Remove.

course_discovery/apps/edx_catalog_extensions/migrations/0001_create_program_types.py
clintonb: Yes.

course_discovery/apps/course_metadata/migrations/0001_squashed_0033_courserun_mobile_available.py
clintonb: A cursory glance points to that being the most likely reason. You can try manually reordering.

MatthewPiatetsky:
@clintonb I manually reordered the models and put the remaining add field operations in.  
(though that particular change doesn't seem to have had any effect on the time)

clintonb (approved):
This looks good. Now we need to do the same for the other apps; however, I would do those after this makes it to stage/production.

mikedikan (approved):
Approving after face to face discussion

clintonb:
@MatthewPiatetsky please remember to squash your commits in the future, unless the commits represent distinct changes. This will help keep the commit history clean and making it easier to use `git bisect` for debugging (if necessary). [Git's fixup/autosquash functionality](https://robots.thoughtbot.com/autosquashing-git-commits) makes squashing pretty easy.

PR #404, opened 2016-10-26
Ingest LMS hidden courses into catalog

MatthewPiatetsky:
https://openedx.atlassian.net/browse/ECOM-5769

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -96,7 +96,7 @@ class CoursesApiDataLoader(AbstractDataLoader):
         self._process_response(response)
 
     def _make_request(self, page):
-        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE)
+        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE, username='course_discovery_worker')
 
     def _process_response(self, response):
         results = response['results']
@@ -140,11 +140,13 @@ class CoursesApiDataLoader(AbstractDataLoader):
 
     def update_course_run(self, course, body):
         key = body['id']
+
         defaults = {
             'key': key,
             'end': self.parse_date(body['end']),
             'enrollment_start': self.parse_date(body['enrollment_start']),
             'enrollment_end': self.parse_date(body['enrollment_end']),
+            'hidden': body.get('hidden', False),
         }
 
         # When using a marketing site, only dates (excluding start) should come from the Course API.
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
@@ -39,6 +39,7 @@ COURSES_API_BODIES = [
         'start': '2015-06-15T13:00:00Z',
         'pacing': 'self',
         'mobile_available': True,
+        'hidden': False,
     },
     {
         'effort': None,
@@ -61,6 +62,7 @@ COURSES_API_BODIES = [
         'start': '2015-10-29T09:00:00Z',
         'pacing': 'instructor,',
         'mobile_available': False,
+        'hidden': False,
     },
     {
         # Add a second run of KyotoUx+000x (3T2016) to test merging data across
@@ -84,6 +86,7 @@ COURSES_API_BODIES = [
         'short_description': '',
         'start': None,
         'mobile_available': None,
+        'hidden': True,
     },
 ]
 
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -372,7 +372,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
         fields = MinimalCourseRunSerializer.Meta.fields + (
             'course', 'full_description', 'start', 'end', 'enrollment_start', 'enrollment_end', 'announcement',
             'video', 'seats', 'content_language', 'transcript_languages', 'instructors', 'staff', 'pacing_type',
-            'min_effort', 'max_effort', 'modified', 'level_type', 'availability', 'mobile_available',
+            'min_effort', 'max_effort', 'modified', 'level_type', 'availability', 'mobile_available', 'hidden',
         )
 
     def get_instructors(self, obj):  # pylint: disable=unused-argument
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -244,6 +244,7 @@ class CourseRunSerializerTests(MinimalCourseRunSerializerTests):
             'video': VideoSerializer(course_run.video).data,
             'pacing_type': course_run.pacing_type,
             'mobile_available': course_run.mobile_available,
+            'hidden': course_run.hidden,
             'content_language': course_run.language.code,
             'transcript_languages': [],
             'min_effort': course_run.min_effort,
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
@@ -167,6 +167,7 @@ class CoursesApiDataLoaderTests(ApiClientTestMixin, DataLoaderTestMixin, TestCas
             'title_override': None,
             'short_description_override': None,
             'video': None,
+            'hidden': body.get('hidden', False),
         }
 
         if not partner_has_marketing_site:


course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -84,6 +86,7 @@
         'short_description': '',
         'start': None,
         'mobile_available': False,
+        'hidden': False,
```
mikedikan: Can we add a test where hidden is true?

MatthewPiatetsky: updated

schenedx (approved):
👍

mikedikan:
👍

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -96,7 +96,7 @@ def _load_data(self, page):  # pragma: no cover
         self._process_response(response)
 
     def _make_request(self, page):
-        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE)
+        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE, username='course_discovery_worker')
```
MatthewPiatetsky, line 99: @edx/ecommerce this user does not exist locally, are we fine with people just creating it locally if they want to run against local LMS?

rlucioni: @MatthewPiatetsky please don't hardcode this kind of information. If it's necessary, this should be a setting. Also consider that it may be possible to extract this username from the token being used for authentication. If a JWT is in use, this can be done directly. If you're using an OAuth access token, you could get this from the LMS.

MatthewPiatetsky: Ok i'Il make a pr to make this a setting

rlucioni: @clintonb thoughts on this? Would it make more sense to pull this username from the token to avoid the need for another setting?

clintonb: Remove this. This is a bad idea. Update the Course API to use the authorized user.

MatthewPiatetsky: @clintonb I can remove this, but I don't know what else uses the course API.
Currently if you don't pass the target_username you get authenticated as an AnonymousUser. Given that the course_discovery auth was not actually working, we probably want to be careful about not breaking the auth of other services with the CourseApi? 
What @rlucioni said could work, (though I can't seem to decode the token I'm getting)

clintonb: @MatthewPiatetsky when you're uncertain the best way to gain certainty is to ask those who have touched the code. Look at the Guy history. Ask in the dev HipChat room.

Almost everything we do is ope source. We cannot have hard-coded values like this in our code base as they may not work for open source users.

MatthewPiatetsky: by I don't know I meant that we can't really know (for instance since the platform is open source). 
There is a whole set of logic around authenticating as a target user and getting an AnonymousUser otherwise, and I don't think we would want to change that behavior. I suppose we could add a parameter like 'authenticate_as_requesting_user' so the platform would still be backwards compatible, but at that point I think renzo's suggestion to get the user from the token would be better (because we wouldn't need to change LMS)
That would look something like this https://github.com/edx/course-discovery/compare/ECOM-6198
Not sure what other possibilities we have

rlucioni: @MatthewPiatetsky I don't think any of us understands why this API handles authentication the way it does. I second @clintonb's suggestion to ask around a bit. I would rather we not simply accept this and attempt to work around it without trying to understand first.

MatthewPiatetsky: @rlucioni I already asked these questions earlier. The idea behind the current implementation is to allow users to impersonate other users. 
You can see the logic here: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/course_api/api.py#L16
But my earlier point was that if we wanted to make the change fully backwards compatible we would need to add another parameter that would tell this code to ignore the target_username (which would be fairly straightforward if we wanted to do that)

MatthewPiatetsky: just talked to @rlucioni about this, will update discussion shortly

MatthewPiatetsky: Looks like here https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions?focusedCommentId=130580574#comment-130580574 is a document with the relevant description (and seems there is a relevant lengthy comment thread that I link to there)  
From my understanding, the recommendation on that page/comment thread is different from what @clintonb and @rlucioni are suggesting. Do we feel like we want to make the case for a different approach with the stakeholders on that page? What should my next steps be?

rlucioni: Do we have examples of other edX APIs that work this way? Is this a pattern we still want to encourage? I still think it's odd (and hard to work with). If we want to continue with this approach, I noticed this [comment](https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions?focusedCommentId=61800874#comment-61800874) from March. Does `/api/user/v1/me` exist?

MatthewPiatetsky: Well this is on the edX REST API Conventions page..so I would imagine there may be other APIs using this convention. I'm not sure if that endpoint exists (but I think the suggestion to create it wouldn't mean you wouldn't have to pass the username)

clintonb: @nasthagiri @mattdrayer @efagin thoughts on this? We are finding it weird to need to pass in the username. In some cases, we may not even have the username to pass in. How strongly do you feel about this sort of filtering? If your opinions have changed, how do we get away from it and simply rely on API authorization which we already have?

efagin: The `/api/user/v1/me` endpoint has been proposed in an edx-platform PR (https://github.com/edx/edx-platform/pull/13423) and was something we had already decided to do a while back. So that'll be available soon. Passing in the username makes sense for APIs that expose user-specific data. (Note that `/me` is not necessary when working with JWTs, which already have the username in them.)

I admit, I'm having trouble following this particular thread and therefore I don't fully understand why the best practice as defined is not sufficient for your needs / otherwise making things unnecessarily difficult.

clintonb: @efagin today, regardless of who you are authenticated as, calls to the LMS Courses API (https://courses.edx.org/api/courses/v1/courses/) will return data for an anonymous user. You can only get user-specific data (e.g. hidden courses only visible to staff) if you pass a username query string parameter.

I understand the motivation but find it a bit strange given that the authenticated request already has a user. I am wondering if we can/should revisit this decision.

efagin: We can certainly bring it up. I think the original goal of that best practice was to cover user _specific_ data, not necessarily user authorizations. It's subtle, but really pertains to the case where a user "owns" the data in question. (Users are generally authorized to see data they own, but can also be authorized to see other data that doesn't have specific user ownership.)

This is confusing, I know. I'm trying to figure out the best way to explain it, because it makes a ton of sense in my head...

Examples of user ownership: a profile, course progress, enrollments, a catalog.
Examples of user authorization beyond user-owned data:  course authoring (including this visibility concept), reporting, administration.

So I think it's reasonable to consider rewriting the best practice to focus on "requesting user data" specifically, which would enable you to build an endpoint that shows a different subset of a catalog depending on the user asking (since courses are not user-owned entities). I believe that's what you want, and the direction you'd like to go - with the added benefit of not changing the best practice for user data.

nasthagiri: FWIW, I am very much open to changing the current implementation of this auth logic in the Courses API.  The current design has been questioned by multiple developers by now - who find it confusing.  And here is yet again the latest thread on this.  Possibly because this auth logic not consistent with the rest of our APIs, not an expected behavior, and/or not intuitive.

Since this API was implemented after the "new" best practice was suggested by @mattdrayer, we implemented it with this recommended auth logic.  Even the first time the "new" practice was suggested, there was a lot of contention and questioning of this approach.  We implemented it in the Courses API with the mentality of trying it out - but very open to revisiting if it's causing confusion/pain.

mattdrayer: @nasthagiri needs to stop pointing her finger at me every time this issue comes around :)  All I've ever said is that we have real-world integration use-cases in which the authenticated user is not the same as the user being requested from the system.  So user resources (just like all resources) should be made available via  explicit, individual addresses.  

If we would like to roll out a set of custom endpoints/views which make assumptions based on information like who the authenticated user is, that is fine, but those endpoints will not support the use cases and stakeholders for whom I am apparently waving the flag.

nasthagiri: Lol.  The reason I always bring you in @mattdrayer, is because you seem to have the best grasp on the rationale for this and can explain it every time.

nasthagiri: @mattdrayer For the Courses API, we can still allow the requesting user to specify another user with the optional `username` parameter.  It's a question of what the default behavior should be if the username is not provided.  Currently, it defaults to assuming an `AnonymousUser` rather than the requesting user.

mattdrayer: Haha, no worries -- although I'm not very familiar with the actual implementation for this Courses API endpoint.  To be honest, it seems odd to me that we would want to access resources via the Courses API and invisibly filter the results by the authenticated user.  If we want to view courses pertaining to a specific user, why not access that information via a Users API endpoint?

clintonb: The use case here is to retrieve all courses visible on a white label site. Normal users cannot see hidden courses. Staff can see all courses.

I vaguely remember you asking "why not access that information via a Users API endpoint" a long time ago. If we follow that practice to its natural conclusion, we end up with a monolithic User API and tiny non-user-specific APIs. That balance strikes me as odd.

I see the Courses API as so when I access /api/courses/v1/courses/:
- Not authenticated: Give me a list of all courses that any non-authenticated user can view
- Authenticated: Give me a list of all courses that I (the authenticated user) can view

While I was somewhat in favor of ensuring that a call to the endpoint always returns the same data, regardless of user, my opinion has changed now that I'm actually using such an API. It's different from many APIs (edX and non-edX), and requires clients to store extra data (username) they might not otherwise need to store. This especially applies to OAuth clients using bearer tokens.

mattdrayer: I'm certainly not going to fall on a sword for an API, so whatever the mob wants, it can have -- we'll simply add endpoints as needed to support the use cases we encounter if they do not exist.

MatthewPiatetsky: Reading this conversation it seems like @rlucioni and @clintonb are in favor of a behavior where passing a username is not required to get your authenticated user's courses, @nasthagiri is neutral, @mattdrayer is theoretically holding a different stance, but ok with us changing this api, not entirely clear on stance of @efagin 
@efagin How would you feel about a change to the endpoint where users that successfully authenticate will see the courses they can see rather than those of an anonymous user (if they do not pass a username)
This will slightly change existing behavior (for instance, staff users who were not passing username, would now use the staff user rather than anonymous user and would see more courses)

efagin: Oh, my comment above was me agreeing with "the mob" as @mattdrayer put it, because I believe catalogs are not _user-owned_ data.

My only request is that you make this toggleable - as in, a user should be able to move between the anonymous and staff views freely. This can be done in two ways:
1. Anonymous by default (as in, user needs to pass `&include_hidden_courses` or something to get the extra courses, and only staff users are allowed to use that flag)
2. Anonymous on demand (as in, user needs to pass `&exclude_hidden_courses` or something, and only staff users are allowed to use that flag)

My preference is option 1 as that preserves the current system behavior by default, but gives staff users flexibility to use their staff access as needed (following the principle of "least privilege").

MatthewPiatetsky: @efagin 
I think we are on the same page about having a flag.
Though, this approach worries me a little bit. 
Suppose you have user groups A,B,C,D,Staff, E,F and course groups 1,2,3,4, Hidden, 5
Lets say Staff users have access to course groups 3,5,Hidden and D users have Access to 1,4,5 groups. 
I'm not too familiar with our permission structure, but If we had a growth in the types of users and permissions, I don't know that we would want clients to need to understand how the permissions of user types work in order to query the service. 
We wouldn't want to say include_hidden_courses, include_4_courses, include_2_courses
I suppose in some ways this makes the username thing make sense to me
However, another potential option is you could have a flag that says ?authorize_with_request_username
Then you wouldn't need to have the username on the client side if you know you don't want to impersonate anyone
Though this may not work for some of the simplicity @rlucioni and @clintonb were looking for
Thoughts?

nasthagiri: I'm in favor of making the change without introducing a new flag for this.
We already have a parameter for specifying `username`, which we will continue to support.  So if a staff user wants to see which courses a student can see, they can pass in the username of a student.

I'm not in favor of passing in permission names as filters/parameters to our APIs until we have clearer sense of the LMS permissioning model - which we don't today.

For rollout of this, I suggest we email the edx-code mailing list so they know this change is coming and whether the change would be an issue for anybody.  Internally, I believe all callers (mobile, insights, etc) always pass in `username`, so it shouldn't be a problem.  I can easily verify this with a Splunk query.

If we really must introduce a staff-access flag that bypasses permission checking altogether, we can be consistent with how we do this in the Course Blocks API.  We have a parameter called `all_blocks` that bypasses permission/group-membership/etc checking and doesn't require the `username` parameter.

nasthagiri: Docstring for `all_blocks`: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/course_api/blocks/views.py#L40

efagin: @nasthagiri I'm all for not adding new things if we can thoughtfully reuse existing concepts - I only threw in flag names as illustrations, not implementation details! In your proposal, what's the behavior if no `username` is specified? I guess there are the following scenarios, let me know if I'm reading your suggestion correctly:
1. No `username`. I assume that this is the current "anonymous" view, preserving existing behavior.
2. `username`, where the value matches the caller's user. Show user-specific view.
3. `username`, where the value does NOT match the caller's user.
3a. If caller is not staff, return "not found".
3b. If caller is staff, and is allowed to access the user's course views, show them.
3c. Otherwise, return "forbidden".

Edit: in scenario 1, no username is needed which means there's no added burden on the API client that "just wants to keep doing what it's already doing today", to @clintonb's earlier concern.

@MatthewPiatetsky, the "is allowed to access" in 3b touches on your complex future-scenario. I'm not sure what use cases you had in mind there, but I think this API structure would account for future user/course groupings without any change required. 

@MatthewPiatetsky, while this is a fun thought exercise on really complex and nuanced API behavior, if you don't have sufficient confidence to move forward by end of day tomorrow, I suggest scheduling an in-person meeting next week with the folks on this thread to try to land somewhere reasonable.

efagin: Oh, and, one thing to keep in mind is that users that aren't staff may be confused by scenario 2 ("why can't I see all my courses? oh, I need to also pass in my username to do that?") which I know goes back to the original concerns that spawned this thread. I don't have a great answer there, unless we want to physically separate "unauthenticated" from "authenticated" endpoints - I could see two separate APIs, one for anonymous course listings that can just be cached outright and can scale nicely, and another for authenticated listings that is far more flexible but is less scaleable.

MatthewPiatetsky: We just met with @clintonb @nasthagiri @mattdrayer and I think we came to a solution that was satisfactory for everyone.
Add a new flag called all_courses. 
If you pass it, you get all the courses your authenticated user can see.
This way, we do not have to worry about backwards compatibility, we do not have to pass the username on the client side and the flag can be more general/reusable than to only cover the single case of hidden courses. 
Unless anyone has objections, I will say this discussion is resolved. I will create a ticket in the ECOM backlog to do this.

efagin: SGTM, thanks for driving this one to resolution @MatthewPiatetsky!

PR #405, opened 2016-10-27
Sort CourseRunViewSet queryset by start date [SOL-1975]

iivic:
Offers on ecommerce offer landing page should be sorted by course start date. Ecommerce is getting the results with a limit of 50 which means that the sorting needs to be done every time new results come in. This will result in constant changes on the page (bad user experience). Because of that I did the sorting in this repo.
https://openedx.atlassian.net/browse/SOL-1975
Please review @bderusha  @mjfrey

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -110,6 +110,19 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
             self.serialize_course_run(CourseRun.objects.all().order_by(Lower('key')), many=True)
         )
 
+    def test_list_sorted_by_course_start_date(self):
+        """ Verify the endpoint returns a list of all catalogs sorted by course start date. """
+        url = '{root}?ordering=start'.format(root=reverse('api:v1:course_run-list'))
+
+        with self.assertNumQueries(11):
+            response = self.client.get(url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertListEqual(
+            response.data['results'],
+            self.serialize_course_run(CourseRun.objects.all().order_by('start'), many=True)
+        )
+
     def test_list_query(self):
         """ Verify the endpoint returns a filtered list of courses """
         course_runs = CourseRunFactory.create_batch(3, title='Some random title', course__partner=self.partner)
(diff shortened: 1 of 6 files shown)


clintonb (changes requested):
We sort by key by default to make the responses predictable. If you want other sort possibilities use [`OrderingFilter`](http://www.django-rest-framework.org/api-guide/filtering/#orderingfilter) and ensure that all ordering fields are indexed to avoid performance issues.

iivic:
@clintonb When I add OrderingFilter filter_backends and make a GET request with the ordering parameter this happens AttributeError: 'SearchQuerySetWrapper' object has no attribute 'order_by'.
I'm not sure how to approach this issue. Any advice?

clintonb:
@iivic update `SearchQuerySetWrapper`. It's essentially a proxy to `SearchQuerySet` that aids in making it look more like `QuerySet` and better handle iteration/pagination. To make it a proper proxy, replace the `count()` method with the following:

``` python
def __getattr__(self, item):
    try:
        super().__getattr__(item)
    except AttributeError:
        # If the attribute is not found on this class, 
        # proxy the request to the SearchQuerySet.
        return getattr(self.qs, item)
```

iivic:
Comments addressed. Please review changes.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -98,7 +98,7 @@ def test_get_include_unpublished_programs(self):
             )
 
     def test_list(self):
-        """ Verify the endpoint returns a list of all catalogs. """
+        """ Verify the endpoint returns a list of all catalogs sorted by key (default sorting). """
```
clintonb: Revert.

course_discovery/apps/course_metadata/models.py
```diff
@@ -331,7 +331,7 @@ class CourseRun(TimeStampedModel):
         max_length=255, default=None, null=True, blank=True,
         help_text=_(
             "Title specific for this run of a course. Leave this value blank to default to the parent course's title."))
-    start = models.DateTimeField(null=True, blank=True)
+    start = models.DateTimeField(null=True, blank=True, db_index=True)
```
clintonb, line 334: This needs a migration.

clintonb (approved):
1. Add a migration.
2. Squash your commits.

mjfrey (approved):

PR #407, opened 2016-10-28
Fixed program admin bug

clintonb:
The previous override of save_model() was largely unnecessary and wrong. The override now relies on the super method, and simply does error handling.

ECOM-6238


clintonb:
@edx/ecommerce please review. This solves the immediate problem. Long-term, we should investigate options around using a wizard. At the very least, on publisher errors, we should direct the user back to the form rather than show the error page.

course_discovery/apps/course_metadata/forms.py
```diff
@@ -16,7 +16,6 @@ class HackDjangoAutocompleteMixin(object):
     # fact an issue?
 
     class QuerySetSelectMixin2(widgets.WidgetMixin):
-
```
mikedikan, line 19: curious why this wasn't a quality error before?

course_discovery/conf/locale/eo/LC_MESSAGES/django.po
```diff
@@ -237,6 +237,24 @@ msgstr "Pärtnér Ⱡ'σяєм ιρѕυм #"
...
+#: apps/course_metadata/admin.py
+msgid "Included course runs"
+msgstr "Ìnçlüdéd çöürsé rüns Ⱡ'σяєм ιρѕυм ∂σłσя ѕιт αмєт, #"
+
+#: apps/course_metadata/admin.py
+msgid ""
+"An error occurred while publishing the program to the marketing site. Please"
+" try again. If the error perisists, please contact the Engineering Team."
+msgstr ""
+"Àn érrör öççürréd whïlé püßlïshïng thé prögräm tö thé märkétïng sïté. Pléäsé"
```
mikedikan, line 249: what is this?

awais786:
There are some test cases for this admin page. I am surprised why they did not fail due to the bug.

MatthewPiatetsky:
@clintonb have you tested that the drag and drop widget functionality still works in the sense that after you drag and drop and save when you reload the new order is preserved?

course_discovery/apps/course_metadata/forms.py
clintonb: It's not a quality error. Having this newline here is more of a personal choice. In my case, I rely on PyCharm's code cleanup tools.

course_discovery/conf/locale/eo/LC_MESSAGES/django.po
clintonb: Dummy strings generated to validate that all strings are marked for translation. See http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html.

schenedx (approved):
👍

MatthewPiatetsky:
@clintonb I just tried this locally, and I think the drag and drop no longer works due to the save_model change (the purpose of using the cleaned data was to get the data in the new order post drag and drop).  
You'll notice if you drag and drop, save, once the page reloads, you're back to the original order before the drag and drop.

clintonb:
Drag and drop works, and maintains the order. The issue is the JS widget.

MatthewPiatetsky:
yes you can physically drag and drop on the screen, but that new order does not get saved to the model. 
the purpose of that save_model change was to save the new order.
If that change prevents creating new programs we should fix that, but we shouldn't break the ability to order courses and organizations. (This will be a regression for several completed tickets)

clintonb:
As I mentioned, order is maintained. Django forms behave correctly. The issue is that your JS widget displays the items alphabetically. I am investigating.

clintonb:
Replaced by #409.

PR #409, opened 2016-10-31
Fixed bug on program admin page

clintonb:
The previous override of save_model() was largely unnecessary and broke because it attempts to set m2m fields on unsaved models. The override now relies on the super method (which performs the correct behavior) and simply does error handling.

The usage of Select2 has been removed from the courses field due to the fact that the widget no longer renders values in the correct order.

ECOM-6238


clintonb:
@edx/ecommerce please review. I have removed the select2 widget for the courses field. I have spent way too many hours looking for a better solution, and come up empty. The widget no longer works for data that needs to be presented in a set order. This has nothing to do with the backend changes, as the data is stored in the correct order.

course_discovery/apps/course_metadata/forms.py
```diff
@@ -33,15 +32,8 @@ class ProgramAdminForm(HackDjangoAutocompleteMixin, forms.ModelForm):
     class Meta:
         model = Program
         fields = '__all__'
-        from dal import autocomplete
```
MatthewPiatetsky, line 36: Will the override for the mixin still work if it is declared after the import? You can try checking out forms.py from master to see if that fixes your issue

clintonb: What override? Did someone implement a hack and not have the professional courtesy to at least leave a comment!?

MatthewPiatetsky: There are multiple comments in the relevant class above.
Agreed that this import should also have a comment

schenedx (approved):
👍

course_discovery/apps/course_metadata/forms.py
```diff
@@ -8,32 +8,30 @@
...
-
-class ProgramAdminForm(HackDjangoAutocompleteMixin, forms.ModelForm):
+class ProgramAdminForm(forms.ModelForm):
     class Meta:
         model = Program
         fields = '__all__'
-        from dal import autocomplete
+
+        # Monkey patch filter_choices_to_render with our own definition which preserves ordering.
+        autocomplete.ModelSelect2Multiple.filter_choices_to_render = filter_choices_to_render_with_order_preserved
```
MatthewPiatetsky, line 33: 👍

MatthewPiatetsky (approved):

PR #411, opened 2016-11-01
ECOM-6252 Ensure data loader does not violate null constraint for mobile available flag

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
@@ -38,7 +38,7 @@ COURSES_API_BODIES = [
         'short_description': '',
         'start': '2015-06-15T13:00:00Z',
         'pacing': 'self',
-        'mobile_available': False,
+        'mobile_available': True,
     },
     {
         'effort': None,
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -38,7 +38,7 @@
         'short_description': '',
         'start': '2015-06-15T13:00:00Z',
         'pacing': 'self',
-        'mobile_available': False,
+        'mobile_available': True,
```
mikedikan, line 41: It looks like there are 3 conditions, null, true, and false, and we only have testing for true and null - should we add a test for false also?

MatthewPiatetsky: we are testing for False, previously all three were false, and I changed them to test all three conditions

mikedikan: ahh, sorry, I see now, thanks.

mikedikan (approved):

rlucioni (approved):

PR #413, opened 2016-11-01
Updated Program.banner_image_url help text

clintonb:
This field is now identified as deprecated to avoid user confusion.

ECOM-6238


clintonb:
@edx/ecommerce please review

course_discovery/apps/course_metadata/models.py
```diff
@@ -623,7 +623,7 @@ class Program(TimeStampedModel):
             'x-small': (348, 116),
         }
     )
-    banner_image_url = models.URLField(null=True, blank=True, help_text=_('Image used atop detail pages'))
+    banner_image_url = models.URLField(null=True, blank=True, help_text='DEPRECATED: Use the banner image field.')
```
rlucioni, line 626: Should this be marked for translation?

course_discovery/apps/core/migrations/0011_auto_20161101_2207.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('core', '0010_auto_20160731_0023'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='username',
+            field=models.CharField(error_messages={'unique': 'A user with that username already exists.'}, help_text='Required. 30 characters or fewer. Letters, digits and @/./+/-/_ only.', max_length=30, unique=True, validators=[django.core.validators.RegexValidator('^[\\w.@+-]+$', 'Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.')], verbose_name='username'),
```
rlucioni, line 19: How'd this migration get in here? Should it be removed?

course_discovery/conf/locale/eo/LC_MESSAGES/django.po
```diff
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-10-31 19:34-0400\n"
+"POT-Creation-Date: 2016-11-01 18:08-0400\n"
```
rlucioni: Aside: I noticed that .po and .mo files are being updated in PRs more frequently than before. What changed? What Make targets are you running to achieve this?

course_discovery/apps/course_metadata/models.py
clintonb: I thought about it. It doesn't make sense to waste energy translating text for a field we are going to remove.

course_discovery/conf/locale/eo/LC_MESSAGES/django.po
clintonb: `make fake_translations`

This is necessary to ensure Transifex always pulls the latest source translations.

course_discovery/apps/core/migrations/0011_auto_20161101_2207.py
clintonb: I suspect this came from Django 1.9; but, I'm not certain. Either way, it needs to be committed.

course_discovery/conf/locale/eo/LC_MESSAGES/django.po
rlucioni: Got it. I know this is only tangentially related, but: I just noticed that Travis verifies that translations are up to date. Had you not run `make fake_translations`, would the build log make it clear what the next steps were? If not, I wonder if a comment in the Makefile or Travis YAML about what to do would be helpful.

rlucioni (approved):

course_discovery/conf/locale/eo/LC_MESSAGES/django.po
clintonb: Travis is simply calling `make validate_translations`, which calls this: https://github.com/edx/i18n-tools/blob/master/i18n/changed.py#L48. On failure, a message will be displayed noting that translations need to be updated.

rlucioni: I see. My point is that I wouldn't have known to run `make fake_translations` if I saw that message. Maybe we should rename the target `make update_translations` to make it a little more clear?

PR #417, opened 2016-11-02
Add 'canonical' field

bderusha:
Add field to indicate when course_run data should be used to update courses
[ECOM-6226](https://openedx.atlassian.net/browse/ECOM-6226)

@edx/ecommerce please review

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -4,7 +4,7 @@ from django.http import HttpResponseRedirect
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 
-from course_discovery.apps.course_metadata.forms import ProgramAdminForm
+from course_discovery.apps.course_metadata.forms import ProgramAdminForm, CourseAdminForm
 from course_discovery.apps.course_metadata.models import *  # pylint: disable=wildcard-import
 from course_discovery.apps.course_metadata.publishers import ProgramPublisherException
 
@@ -42,6 +42,7 @@ class CorporateEndorsementsInline(admin.TabularInline):
 
 @admin.register(Course)
 class CourseAdmin(admin.ModelAdmin):
+    form = CourseAdminForm
     list_display = ('uuid', 'key', 'title',)
     list_filter = ('partner',)
     ordering = ('key', 'title',)
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -107,8 +107,20 @@ class CoursesApiDataLoader(AbstractDataLoader):
 
             try:
                 body = self.clean_strings(body)
-                course = self.update_course(body)
-                self.update_course_run(course, body)
+                course_run = self.get_course_run(body)
+
+                if course_run:
+                    self.update_course_run(course_run, body)
+                    course = getattr(course_run, 'canonical_for_course', False)
+                    if course:
+                        course = self.update_course(course, body)
+                        logger.info('Processed course with key [%s].', course.key)
+                else:
+                    course, created = self.get_or_create_course(body)
+                    course_run = self.create_course_run(course, body)
+                    if created:
+                        course.canonical_course_run = course_run
+                        course.save()
             except:  # pylint: disable=bare-except
                 msg = 'An error occurred while updating {course_run} from {api_url}'.format(
                     course_run=course_run_id,
@@ -116,14 +128,33 @@ class CoursesApiDataLoader(AbstractDataLoader):
                 )
                 logger.exception(msg)
 
-    def update_course(self, body):
+    def get_course_run(self, body):
+        course_run_key = body['id']
+        try:
+            return CourseRun.objects.get(key__iexact=course_run_key)
+        except CourseRun.DoesNotExist:
+            return None
+
+    def update_course_run(self, course_run, body):
+        validated_data = self.format_course_run_data(body)
+        self._update_instance(course_run, validated_data)
+
+        logger.info('Processed course run with UUID [%s].', course_run.uuid)
+
+    def create_course_run(self, course, body):
+        defaults = self.format_course_run_data(body, course=course)
+
+        return CourseRun.objects.create(**defaults)
+
+    def get_or_create_course(self, body):
         course_run_key = CourseKey.from_string(body['id'])
         course_key = self.get_course_key_from_course_run_key(course_run_key)
+        defaults = self.format_course_data(body)
+        # We need to add the key to the defaults because django ignores kwargs with __
+        # separators when constructing the create request
+        defaults['key'] = course_key
+        defaults['partner'] = self.partner
 
-        defaults = {
-            'key': course_key,
-            'title': body['name'],
-        }
         course, created = Course.objects.get_or_create(key__iexact=course_key, partner=self.partner, defaults=defaults)
 
         if created:
@@ -133,16 +164,27 @@ class CoursesApiDataLoader(AbstractDataLoader):
             defaults = {'key': key}
             organization, __ = Organization.objects.get_or_create(key__iexact=key, partner=self.partner,
                                                                   defaults=defaults)
+
             course.authoring_organizations.add(organization)
 
-        logger.info('Processed course with key [%s].', course_key)
+        return (course, created)
+
+    def update_course(self, course, body):
+        validated_data = self.format_course_data(body)
+        self._update_instance(course, validated_data)
+
+        logger.info('Processed course with key [%s].', course.key)
+
         return course
 
-    def update_course_run(self, course, body):
-        key = body['id']
+    def _update_instance(self, instance, validated_data):
+        for attr, value in validated_data.items():
+            setattr(instance, attr, value)
+        instance.save()
 
+    def format_course_run_data(self, body, course=None):
         defaults = {
-            'key': key,
+            'key': body['id'],
             'end': self.parse_date(body['end']),
             'enrollment_start': self.parse_date(body['enrollment_start']),
             'enrollment_end': self.parse_date(body['enrollment_end']),
@@ -162,10 +204,17 @@ class CoursesApiDataLoader(AbstractDataLoader):
                 'mobile_available': body.get('mobile_available') or False,
             })
 
-        course_run, __ = course.course_runs.update_or_create(key__iexact=key, defaults=defaults)
+        if course:
+            defaults['course'] = course
 
-        logger.info('Processed course run with key [%s].', course_run.key)
-        return course_run
+        return defaults
+
+    def format_course_data(self, body):
+        defaults = {
+            'title': body['name'],
+        }
+
+        return defaults
 
     def get_pacing_type(self, body):
         pacing = body.get('pacing')
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
@@ -133,8 +133,9 @@ class CoursesApiDataLoaderTests(ApiClientTestMixin, DataLoaderTestMixin, TestCas
     def api_url(self):
         return self.partner.courses_api_url
 
-    def mock_api(self):
-        bodies = mock_data.COURSES_API_BODIES
+    def mock_api(self, bodies=None):
+        if not bodies:
+            bodies = mock_data.COURSES_API_BODIES
         url = self.api_url + 'courses/'
         responses.add_callback(
             responses.GET,
@@ -185,6 +186,8 @@ class CoursesApiDataLoaderTests(ApiClientTestMixin, DataLoaderTestMixin, TestCas
         for field, value in expected_values.items():
             self.assertEqual(getattr(course_run, field), value, 'Field {} is invalid.'.format(field))
 
+        return course_run
+
     @responses.activate
     @ddt.data(True, False)
     def test_ingest(self, partner_has_marketing_site):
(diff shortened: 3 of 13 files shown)


bderusha:
[CR](https://trello.com/c/2I8C0GWX/268-cr)

bderusha:
[CR](https://trello.com/c/2I8C0GWX/268-cr)

bderusha:
[CR](https://trello.com/c/2I8C0GWX/268-cr)

clintonb (commented):
Who/what determines if a course run is considered canonical?

course_discovery/apps/course_metadata/models.py
```diff
@@ -324,6 +324,7 @@ class CourseRun(TimeStampedModel):
 
     uuid = models.UUIDField(default=uuid4, editable=False, verbose_name=_('UUID'))
     course = models.ForeignKey(Course, related_name='course_runs')
+    canonical = models.BooleanField(default=True)
```
clintonb: This runs the risk of having multiple canonical course runs for a single course. There are a few fixes:
1. Create a separate model, `CanonicalCourseRuns`, that maps `Course`s to `CourseRuns` and enforces uniqueness on the course column.
2. Add a unique key to this model on the course and canonical fields.
3. Reverse the linkage, and creating a foreign key from `Course` to `CourseRun`.

Personally, I'm leaning toward option 1 since it is quite explicit about its purpose, constraints can be easily enforced without try/except blocks around calls to `save()`, and we don't end up with weird FKs.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -116,7 +119,15 @@ def _process_response(self, response):
...
-    def update_course(self, body):
+    def use_course_run_for_course(self, body):
+        course_run_key = body['id']
+        try:
+            course_run = CourseRun.objects.get(key__iexact=course_run_key)
+            return course_run.canonical
+        except CourseRun.DoesNotExist:
+            return True
+
+    def update_course(self, body, ):
```
rlucioni: Nitpick: looks like you were experimenting with a new argument here but left a trailing comma.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -116,7 +119,15 @@ def _process_response(self, response):
                 )
                 logger.exception(msg)
 
-    def update_course(self, body):
+    def use_course_run_for_course(self, body):
+        course_run_key = body['id']
+        try:
+            course_run = CourseRun.objects.get(key__iexact=course_run_key)
```
rlucioni: Do you have a sense of what this extra query does to the runtime of the refresh command?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -160,8 +171,10 @@ def update_course_run(self, course, body):
                 'mobile_available': body.get('mobile_available') or False,
             })
 
-        course_run, __ = course.course_runs.update_or_create(key__iexact=key, defaults=defaults)
-
+        if course:
+            course_run, __ = course.course_runs.update_or_create(key__iexact=key, defaults=defaults)
+        else:
+            course_run, __ = CourseRun.objects.update_or_create(key__iexact=key, defaults=defaults)
```
rlucioni: For your consideration:

``` py
getattr(course.course_runs if course else CourseRun.objects, 'update_or_create')(key__iexact=key, defaults=defaults)
```

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -226,6 +226,68 @@ def test_ingest_exception_handling(self):
...
+    @responses.activate
+    def test_canonical_update(self):
+        """ Verify the method ingests data and respects canonical flag. """
+        self.mock_api()
+        self.assertEqual(Course.objects.count(), 0)
+        self.assertEqual(CourseRun.objects.count(), 0)
+        self.loader.ingest()
+
+        # Verify the API was called with the correct authorization header
+        self.assert_api_called(1)
```
rlucioni: Is this relevant to what you're trying to verify in this test?

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -226,6 +226,68 @@ def test_ingest_exception_handling(self):
...
+
+        # Verify course data is reset when canonical flag is True
+        self.loader.ingest()
+        course = Course.objects.get(key__exact=course_key)
+        self.assertEqual(course.title, original_title)
+
+    @responses.activate
+    @ddt.data(True, False)
+    def test_use_course_run_for_course_existing(self, is_canonical):
+        """ Verify the method returns None if the API response does not include a pacing field. """
```
rlucioni: Looks like this docstring and the one below need to be updated.

bderusha:
@clintonb Currently we default to having all course_runs be considered `canonical` since we update the course with every course_run.  The idea behind this is that we could decide that certain course_runs are no longer going update course metadata by manually marking them as not canonical sources.  Old courses for example. 

Essentially it is being explicit about something that we are implicit about today, which is using "last one wins" to determine which course_run gets to set the data on a course.

course_discovery/apps/course_metadata/models.py
bderusha: Separate model is fine if we do indeed want to enforce uniqueness.  I envisioned this as trying to converge on one canonical course_run for all courses, but allowing the option to continue working as we work today (letting multiple course runs update the course).  But thinking more about it I think we could make this work.  it would also speed up the data loaders quite a bit by not updating all of the course data for each run

bderusha:
@clintonb @rlucioni Still writing tests for this but please take a look at the new approach when you get a chance.  Thanks!

course_discovery/apps/course_metadata/models.py
```diff
@@ -494,6 +508,16 @@ def __str__(self):
         return '{key}: {title}'.format(key=self.key, title=self.title)
 
 
+class CanonicalCourseRunCourse(TimeStampedModel):
```
clintonb: Sorry for the churn here. I just remembered [`OneToOneField`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.OneToOneField) exists. Would that make this even simpler?

``` python
class Course(...):
  ...
  # TODO Handle on_delete to ensure course_run deletion does not result in course deletion
  canonical_course_run = models.OneToOneField('course_metadata.CourseRun', related_name='+')
```

course_discovery/apps/course_metadata/admin.py
```diff
@@ -187,6 +187,13 @@ class PersonAdmin(admin.ModelAdmin):
     search_fields = ('uuid', 'family_name', 'given_name', 'slug',)
 
 
+@admin.register(CanonicalCourseRunCourse)
+class CanonicalCourseRunCourseAdmin(admin.ModelAdmin):
```
clintonb: You _might_ be able to inline this on the `Course` admin page.

course_discovery/apps/course_metadata/models.py
```diff
@@ -290,6 +293,10 @@ def active_course_runs(self):
             )
         )
 
+    @property
+    def canonical_course_run(self):
+        return self._canonical_course_runs.first()
```
clintonb: You can avoid this weirdness by doing the following:

``` python
try:
  return CanonicalCourseRunCourse.objects.get(course=self).course_run
except CanonicalCourseRunCourse.DoesNotExist:
  # Alternatively, you can let the error be raised
  return None
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -477,6 +487,10 @@ def availability(self):
         else:
             return _('Upcoming')
 
+    @property
+    def canonical_course(self):
+        return self._canonical_courses.first()
```
clintonb: Same as above:

``` python
try:
  return CanonicalCourseRunCourse.objects.get(course_run=self).course
except CanonicalCourseRunCourse.DoesNotExist:
  # Alternatively, you can let the error be raised
  return None
```

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -53,6 +53,13 @@ def test_course_run_update_caught_exception(self):
         with mock.patch.object(Course, 'reindex_course_runs', side_effect=Exception):
             self.course.save()
 
+    def test_canonical_course_run(self):
```
clintonb: Remember to test the negative case, where a course does not have a canonical course run.

clintonb (approved):
I'm not a fan of the changes to the Course and CourseRun models. Otherwise, this looks good.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -5,14 +5,16 @@
 import math
 
 from django.core.files import File
+from django.db import transaction
```
rlucioni: This import is unused.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -107,42 +109,98 @@ def _process_response(self, response):
...
+        defaults['partner'] = self.partner
 
-        defaults = {
-            'key': course_key,
-            'title': body['name'],
-        }
-        course, created = Course.objects.get_or_create(key__iexact=course_key, partner=self.partner, defaults=defaults)
+        # http://stackoverflow.com/questions/2235318/how-do-i-deal-with-this-race-condition-in-django
+        # BD: This is a hack ported from the above SO article because get_or_create is
+        # not threadsafe and this method is being explicitly called from many threads
```
rlucioni: @bderusha discovery's MySQL isolation level is set to READ COMMITTED, which means the situation called out in that SO post shouldn't be possible here. In fact, switching to READ COMMITTED is one of the solutions called out there. Even if we were using REPEATABLE READ, I don't think this would work because you're not committing any transactions here.

Are you trying to preempt something here, or are you trying to fix an observed failure? [This line](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py#L147-L151) is meant to prevent what appears to be happening here.

course_discovery/apps/course_metadata/urls.py
```diff
@@ -5,12 +5,13 @@
...
 from course_discovery.apps.course_metadata.views import CourseRunSelectionAdmin
 from course_discovery.apps.course_metadata.lookups import (
-    CourseAutocomplete, OrganizationAutocomplete, VideoAutocomplete
+    CourseAutocomplete, CourseRunAutocomplete, OrganizationAutocomplete, VideoAutocomplete
 )
 
 urlpatterns = [
     url(r'^update_course_runs/(?P<pk>\d+)/$', CourseRunSelectionAdmin.as_view(), name='update_course_runs',),
     url(r'^course-autocomplete/$', CourseAutocomplete.as_view(), name='course-autocomplete',),
+    url(r'^course-run-autocomplete/$', CourseRunAutocomplete.as_view(), name='course-run-autocomplete',),
```
rlucioni, line 14: What's this for?

course_discovery/apps/course_metadata/data_loaders/api.py
bderusha: @rlucioni The tests fail because SQLite and because the tests call the ingest method directly.  I can have the tests set that waffleswitch, which may alleviate the issue

course_discovery/apps/course_metadata/urls.py
bderusha: This is for the typeahead autocomplete used in django-admin

course_discovery/apps/course_metadata/models.py
bderusha: Ooo yea, that would probably do the trick. Let me try it out

course_discovery/apps/course_metadata/data_loaders/api.py
rlucioni: @bderusha yeah, I'd try going that route. There are two switches, one to control threaded writes within the loaders and another to control use of multiple processes when running the pipeline. Either or both should help here. I'd include a comment alluding to why you're using them!

bderusha: Hmm.  I lied about the test setup.  We default to not being threadsafe so this is just a race condition within the test because of using SQLite

rlucioni: @bderusha so is the fix to make the tests run in threadsafe mode? I don't like the idea of changing this code just to placate tests. Maybe we should chat?

bderusha:
@rlucioni @clintonb ready for another (final?) round of reviews

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -107,23 +107,53 @@ def _process_response(self, response):
...
+    def create_course_run(self, course, body):
+        defaults = self.format_course_run_data(body, course=course)
+
+        return CourseRun.objects.create(**defaults)
+
+    def get_or_create_course(self, body):
         course_run_key = CourseKey.from_string(body['id'])
         course_key = self.get_course_key_from_course_run_key(course_run_key)
+        defaults = self.format_course_data(body)
+        defaults['key'] = course_key
```
rlucioni, line 155: Please add a comment about why doing this is necessary. A future maintainer might see this and not realize this is critical.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -107,23 +107,53 @@ def _process_response(self, response):
...
-                course = self.update_course(body)
-                self.update_course_run(course, body)
+                course_run = self.get_course_run(body)
+
+                if course_run:
+                    self.update_course_run(course_run, body)
+                    try:
+                        course = self.update_course(course_run.canonical_for_course, body)
+                        logger.info('Processed course with key [%s].', course.key)
+                    except AttributeError:
```
rlucioni: I have nothing against using exception handlers this way, but I think it could read more clearly. Consider something more explicit, like:

``` py
is_canonical = getattr(course_run, 'canonical_for_course', False)
if is_canonical:
    ...
else:
    pass
```

course_discovery/apps/course_metadata/urls.py
rlucioni: Why are you modifying it as part of this? I don't understand how it's related.

rlucioni (approved):
Looks good, just some minor comments inline.

course_discovery/apps/course_metadata/urls.py
bderusha: https://github.com/edx/course-discovery/pull/417/files#diff-f9fbbbbc3421066d028344acb03a0b55R110

To make the admin easier to work with for setting a canonical course run  I included the typeahead widget, but we didn't have typeahead for course-runs, so I implemented it here

rlucioni: Got it, makes sense. Thanks.

PR #418, opened 2016-11-02
Update getting_started.rst

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -48,7 +48,7 @@ provider and updating the Django settings for this project with the client crede
 A new OAuth 2.0 client can be created at ``http://127.0.0.1:8000/admin/oauth2/client/``.
 
     1. Click the :guilabel:`Add client` button.
-    2. Leave the user field blank.
+    2. Select a user.
     3. Specify the name of this service, ``Course Discovery Service``, as the client name.
     4. Set the :guilabel:`URL` to the root path of this service: ``http://localhost:18381/``.
     5. Set the :guilabel:`Redirect URL` to the OIDC client endpoint: ``http://localhost:18381/complete/edx-oidc/``.


docs/getting_started.rst
```diff
@@ -48,10 +48,10 @@ provider and updating the Django settings for this project with the client crede
 A new OAuth 2.0 client can be created at ``http://127.0.0.1:8000/admin/oauth2/client/``.
 
     1. Click the :guilabel:`Add client` button.
-    2. Leave the user field blank.
+    2. Pick a user.
     3. Specify the name of this service, ``Course Discovery Service``, as the client name.
-    4. Set the :guilabel:`URL` to the root path of this service: ``http://localhost:18381/``.
```
clintonb, line 53: Revert. These ports are relevant for Docker, which we should all be using.

docs/getting_started.rst
```diff
@@ -48,10 +48,10 @@ provider and updating the Django settings for this project with the client crede
 A new OAuth 2.0 client can be created at ``http://127.0.0.1:8000/admin/oauth2/client/``.
 
     1. Click the :guilabel:`Add client` button.
-    2. Leave the user field blank.
+    2. Pick a user.
```
clintonb: Nit: Select a user.

schenedx (commented):
I believe the instruction is based on the docker setup. I believe you are using the local machine setup. 
So, I don't know if we should make the change here if we haven't made the decision on which setup we should officially support

schenedx:
@MatthewPiatetsky But I agree, we should change the setup to `select a user`

clintonb (approved):

clintonb:
Closed due to inactivity.

PR #419, opened 2016-11-03
Revert "Revert "Merge pull request #403 from edx/ECOM-6122""

MatthewPiatetsky:
This reverts commit 8fbdb2869dbdf90f39d52456c19a699e78c22f4f.

@edx/ecommerce


clintonb (approved):
Check to ensure no other migrations have been committed since this was first reverted.

MatthewPiatetsky:
actually there has been (but tests will catch that) will update  
EDIT: actually I think we probably want to leave it as is.
The squashed migration squashes 1-33, 34 was added, but if the squashed migration does not get applied, and 34 has not been applied yet, then we probably want to have 34 be applied, and then when we delete 1-33 update the dependency of 34?

mikedikan (approved):

MatthewPiatetsky:
ok lets try this again

mikedikan:
@MatthewPiatetsky please delete branch

PR #423, opened 2016-11-07
Updated ALLOWED_HOSTS for local development

clintonb:
The Django 1.9.11 security update changed the default value of ALLOWED_HOSTS. This commit restores the default as we will never use local.py in production.

ECOM-6274


clintonb:
FYI @edx/ecommerce

bderusha (approved):

rlucioni (approved):

PR #424, opened 2016-11-07
Get username from JWT token and remove tokens from command line args

MatthewPiatetsky:
Also only allow JWT tokens
@edx/ecommerce @rlucioni

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/__init__.py
+++ b/course_discovery/apps/course_metadata/data_loaders/__init__.py
@@ -23,10 +23,10 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
     """
 
     PAGE_SIZE = 50
-    SUPPORTED_TOKEN_TYPES = ('bearer', 'jwt',)
     MARKDOWN_CLEANUP_REGEX = re.compile(r'^<p>(.*)</p>$')
 
-    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None, is_threadsafe=False):
+    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None,
+                 is_threadsafe=False, **kwargs):
         """
         Arguments:
             partner (Partner): Partner which owns the APIs and data being loaded
@@ -39,9 +39,6 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
         if token_type:
             token_type = token_type.lower()
 
-            if token_type not in self.SUPPORTED_TOKEN_TYPES:
-                raise ValueError('The token type {token_type} is invalid!'.format(token_type=token_type))
-
         self.access_token = access_token
         self.token_type = token_type
         self.partner = partner
@@ -49,6 +46,7 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
 
         self.max_workers = max_workers
         self.is_threadsafe = is_threadsafe
+        self.username = kwargs.get('username')
 
     @cached_property
     def api_client(self):
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -96,7 +96,7 @@ class CoursesApiDataLoader(AbstractDataLoader):
         self._process_response(response)
 
     def _make_request(self, page):
-        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE, username='course_discovery_worker')
+        return self.api_client.courses().get(page=page, page_size=self.PAGE_SIZE, username=self.username)
 
     def _process_response(self, response):
         results = response['results']
@@ -294,9 +294,10 @@ class ProgramsApiDataLoader(AbstractDataLoader):
     image_height = 480
     XSERIES = None
 
-    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None, is_threadsafe=False):
+    def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None,
+                 is_threadsafe=False, **kwargs):
         super(ProgramsApiDataLoader, self).__init__(
-            partner, api_url, access_token, token_type, max_workers, is_threadsafe
+            partner, api_url, access_token, token_type, max_workers, is_threadsafe, **kwargs
         )
         self.XSERIES = ProgramType.objects.get(name='XSeries')
 
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
@@ -1,30 +1,23 @@
-import ddt
 import responses
-from edx_rest_api_client.auth import BearerAuth, SuppliedJwtAuth
+from edx_rest_api_client.auth import SuppliedJwtAuth
 from edx_rest_api_client.client import EdxRestApiClient
 
 from course_discovery.apps.course_metadata.tests.factories import PartnerFactory
 
 ACCESS_TOKEN = 'secret'
-ACCESS_TOKEN_TYPE = 'Bearer'
+ACCESS_TOKEN_TYPE = 'JWT'
 
 
-@ddt.ddt
 class ApiClientTestMixin(object):
-    @ddt.unpack
-    @ddt.data(
-        ('Bearer', BearerAuth),
-        ('JWT', SuppliedJwtAuth),
-    )
-    def test_api_client(self, token_type, expected_auth_class):
+    def test_api_client(self):
         """ Verify the property returns an API client with the correct authentication. """
-        loader = self.loader_class(self.partner, self.api_url, ACCESS_TOKEN, token_type)
+        loader = self.loader_class(self.partner, self.api_url, ACCESS_TOKEN, ACCESS_TOKEN_TYPE)
         client = loader.api_client
         self.assertIsInstance(client, EdxRestApiClient)
         # NOTE (CCB): My initial preference was to mock the constructor and ensure the correct auth arguments
         # were passed. However, that seems nearly impossible. This is the next best alternative. It is brittle, and
         # may break if we ever change the underlying request class of EdxRestApiClient.
-        self.assertIsInstance(client._store['session'].auth, expected_auth_class)  # pylint: disable=protected-access
+        self.assertIsInstance(client._store['session'].auth, SuppliedJwtAuth)  # pylint: disable=protected-access
 
 
 # pylint: disable=not-callable
@@ -45,15 +38,10 @@ class DataLoaderTestMixin(object):
         """ Asserts the API was called with the correct number of calls, and the appropriate Authorization header. """
         self.assertEqual(len(responses.calls), expected_num_calls)
         if check_auth:
-            self.assertEqual(responses.calls[0].request.headers['Authorization'], 'Bearer {}'.format(ACCESS_TOKEN))
+            self.assertEqual(responses.calls[0].request.headers['Authorization'], 'JWT {}'.format(ACCESS_TOKEN))
 
     def test_init(self):
         """ Verify the constructor sets the appropriate attributes. """
         self.assertEqual(self.loader.partner.short_code, self.partner.short_code)
         self.assertEqual(self.loader.access_token, ACCESS_TOKEN)
         self.assertEqual(self.loader.token_type, ACCESS_TOKEN_TYPE.lower())
-
-    def test_init_with_unsupported_token_type(self):
-        """ Verify the constructor raises an error if an unsupported token type is passed in. """
-        with self.assertRaises(ValueError):
-            self.loader_class(self.partner, self.api_url, ACCESS_TOKEN, 'not-supported')
--- a/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
@@ -5,6 +5,7 @@ import logging
 from django.core.management import BaseCommand, CommandError
 from django.db import connection
 from edx_rest_api_client.client import EdxRestApiClient
+import jwt
 import waffle
 
 from course_discovery.apps.core.models import Partner
@@ -21,9 +22,9 @@ from course_discovery.apps.course_metadata.models import Course
 logger = logging.getLogger(__name__)
 
 
-def execute_loader(loader_class, *loader_args):
+def execute_loader(loader_class, *loader_args, **loader_kwargs):
     try:
-        loader_class(*loader_args).ingest()
+        loader_class(*loader_args, **loader_kwargs).ingest()
     except Exception:  # pylint: disable=broad-except
         logger.exception('%s failed!', loader_class.__name__)
 
@@ -51,22 +52,6 @@ class Command(BaseCommand):
     help = 'Refresh course metadata from external sources.'
 
     def add_arguments(self, parser):
-        parser.add_argument(
-            '--access_token',
-            action='store',
-            dest='access_token',
-            default=None,
-            help='OAuth2 access token used to authenticate API calls.'
-        )
-
-        parser.add_argument(
-            '--token_type',
-            action='store',
-            dest='token_type',
-            default=None,
-            help='The type of access token being passed  (e.g. Bearer, JWT).'
-        )
-
         parser.add_argument(
             '--partner_code',
             action='store',
@@ -98,27 +83,22 @@ class Command(BaseCommand):
         if not partners:
             raise CommandError('No partners available!')
 
+        token_type = 'JWT'
         for partner in partners:
-            access_token = options.get('access_token')
-            token_type = options.get('token_type')
+            logger.info('Retrieving access token for partner [{}]'.format(partner_code))
 
-            if access_token and not token_type:
-                raise CommandError('The token_type must be specified when passing in an access token!')
-
-            if not access_token:
-                logger.info('No access token provided. Retrieving access token using client_credential flow...')
-                token_type = 'JWT'
-
-                try:
-                    access_token, __ = EdxRestApiClient.get_oauth_access_token(
-                        '{root}/access_token'.format(root=partner.oidc_url_root.strip('/')),
-                        partner.oidc_key,
-                        partner.oidc_secret,
-                        token_type=token_type
-                    )
-                except Exception:
-                    logger.exception('No access token provided or acquired through client_credential flow.')
-                    raise
+            try:
+                access_token, __ = EdxRestApiClient.get_oauth_access_token(
+                    '{root}/access_token'.format(root=partner.oidc_url_root.strip('/')),
+                    partner.oidc_key,
+                    partner.oidc_secret,
+                    token_type=token_type
+                )
+            except Exception:
+                logger.exception('No access token acquired through client_credential flow.')
+                raise
+            username = jwt.decode(access_token, verify=False)['preferred_username']
+            kwargs = {'username': username} if username else {}
 
             # The Linux kernel implements copy-on-write when fork() is called to create a new
             # process. Pages that the parent and child processes share, such as the database
@@ -192,6 +172,7 @@ class Command(BaseCommand):
                                     token_type,
                                     (max_workers_override or max_workers),
                                     is_threadsafe,
+                                    **kwargs,
                                 )
             else:
                 # Flatten pipeline and run serially.
@@ -205,6 +186,7 @@ class Command(BaseCommand):
                             token_type,
                             (max_workers_override or max_workers),
                             is_threadsafe,
+                            **kwargs,
                         )
 
             # TODO Cleanup CourseRun overrides equivalent to the Course values.
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
@@ -1,6 +1,7 @@
 import json
 
 import ddt
+import jwt
 import mock
 import responses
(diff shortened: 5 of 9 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -70,7 +71,11 @@ def ingest(self):
         logger.info('Refreshing Courses and CourseRuns from %s...', self.partner.courses_api_url)
 
         initial_page = 1
-        response = self._make_request(initial_page)
+        try:
+            username = jwt.decode(self.access_token, verify=False).get('preferred_username', '')
+        except jwt.exceptions.DecodeError:
+            username = ''
+        response = self._make_request(initial_page, username)
```
MatthewPiatetsky: If the username is in fact only necessary for the CoursesApi it seems like we probably don't want to put the username variable in the abstract data loader, using self outside the init is a pylint error, which is I'm passing it here (though we could disable it)

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -70,7 +71,11 @@ def ingest(self):
         logger.info('Refreshing Courses and CourseRuns from %s...', self.partner.courses_api_url)
 
         initial_page = 1
-        response = self._make_request(initial_page)
+        try:
+            username = jwt.decode(self.access_token, verify=False).get('preferred_username', '')
```
MatthewPiatetsky: verify=False just pulls out the information without verifying the signature

course_discovery/apps/course_metadata/data_loaders/api.py
clintonb: Update `AbstractDataLoader` to accept `kwargs`. Pass the username as a `kwarg` and store it for this particular data loader. Move all of the decoding logic up to the management command level.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -70,7 +71,11 @@ def ingest(self):
...
+        except jwt.exceptions.DecodeError:
+            username = ''
```
MatthewPiatetsky: If we can't get the username from the token just fall back to what the result of not passing a username to the course api would be. (If this is in fact the behavior that is happening being explicit about it could at least alleviate some of the confusion)

course_discovery/apps/course_metadata/data_loaders/api.py
clintonb: Move this to the management command. Keep verification since we should always have a valid token.

clintonb (changes requested):

course_discovery/apps/course_metadata/data_loaders/api.py
MatthewPiatetsky: isn't the secret for encoding the JWT token set in LMS or the extension that LMS uses?
If catalog does not have that wouldn't I fail on signature verification?

MatthewPiatetsky:
did we want to just remove the ability to pass token_type or remove the ability to pass a token command line altogether?  
Also did I understand correctly that we don't want the course discovery service to use bearer tokens anymore?

course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -23,10 +23,11 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
     """
 
     PAGE_SIZE = 50
-    SUPPORTED_TOKEN_TYPES = ('bearer', 'jwt',)
+    SUPPORTED_TOKEN_TYPES = ('jwt',)
```
MatthewPiatetsky: Should we remove this? Or keep if in the future we might use other types of tokens?

clintonb: We should remove the ability to pass in a token altogether. We've never actually used it.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -70,6 +70,7 @@ def ingest(self):
         logger.info('Refreshing Courses and CourseRuns from %s...', self.partner.courses_api_url)
 
         initial_page = 1
+
```
clintonb: Revert.

course_discovery/apps/course_metadata/data_loaders/__init__.py
MatthewPiatetsky: Ok. We actually say to use it here https://open-edx-course-catalog.readthedocs.io/en/latest/course_metadata.html#retrieving-course-metadata but I can update that

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -116,6 +105,11 @@ def handle(self, *args, **options):
                         partner.oidc_secret,
                         token_type=token_type
                     )
+                    try:
+                        username = jwt.decode(access_token, verify=False).get('preferred_username', '')
+                    except jwt.exceptions.DecodeError:
```
clintonb: If decode fails, the entire command should fail. There is clearly something wrong with the token.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -116,6 +105,11 @@ def handle(self, *args, **options):
                         partner.oidc_secret,
                         token_type=token_type
                     )
+                    try:
+                        username = jwt.decode(access_token, verify=False).get('preferred_username', '')
```
clintonb: If the token doesn't contain the `preferred_username` claim, the command should fail.

course_discovery/apps/course_metadata/data_loaders/__init__.py
clintonb: Correction: it is used by `ManagementViewSet.refresh_course_metadata()` which should also be removed.

clintonb (changes requested):

course_discovery/apps/course_metadata/data_loaders/__init__.py
clintonb: Update the docs, too.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -98,27 +83,22 @@ def handle(self, *args, **options):
...
+            try:
+                access_token, __ = EdxRestApiClient.get_oauth_access_token(
+                    '{root}/access_token'.format(root=partner.oidc_url_root.strip('/')),
+                    partner.oidc_key,
+                    partner.oidc_secret,
+                    token_type=token_type
+                )
+                username = jwt.decode(access_token, verify=False)['preferred_username']
+                kwargs = {'username': username} if username else {}
+            except Exception:
```
rlucioni, line 97: Keep this exception handler confined to the token request. The decode should occur outside of it, in case, say, the preferred_username key is missing. If that were to occur, the message logged below would be confusing.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -19,7 +19,7 @@
 from course_discovery.apps.course_metadata.tests import toggle_switch
 from course_discovery.apps.course_metadata.tests.factories import CourseFactory
 
-ACCESS_TOKEN = 'secret'
+ACCESS_TOKEN = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsInByZWZlcnJlZF91c2VybmFtZSI6ImpvaG4ifQ.SFJ4FVliPTP27EhxLHeTHDJJ6bTanoHFq-DWtMgUgPA'
```
rlucioni: Generate this in test code instead of hardcoding it.

rlucioni (changes requested):
@MatthewPiatetsky left some comments inline. Let's keep an eye on https://github.com/edx/course-discovery/pull/404#pullrequestreview-7261966. I still think fixing that API is preferable to making changes here.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: why, isn't the actual token generation something that happens in the extension in LMS? There isn't actually code to test in this management command that generates a token

rlucioni: In production-like deployments, JWTs come from our auth provider, which happens to be from the LMS. However, that doesn't preclude you from generating your own tokens for testing, which you can do with the `jwt` module. Doing that allows you to more clearly show what's being encoded in the token, and makes it easier to change in the future.

MatthewPiatetsky:
@clintonb @rlucioni I think I've addressed most of your feedback. 
As to the discussion on the other PR, I don't have a strong opinion on any approach, but it sounds like there are some differences of opinion. How do we think we will come to a decision?
(Note that most of the changes in this PR do not relate to pulling the username from the token, so if we decide not to do that, I can just remove those sections)

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -1,19 +1,18 @@
...
 
 ACCESS_TOKEN = 'secret'
-ACCESS_TOKEN_TYPE = 'Bearer'
+ACCESS_TOKEN_TYPE = 'JWT'
 
 
 @ddt.ddt
 class ApiClientTestMixin(object):
     @ddt.unpack
     @ddt.data(
```
clintonb: `ddt` is no longer needed here.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -98,27 +83,22 @@ def handle(self, *args, **options):
         if not partners:
             raise CommandError('No partners available!')
 
+        token_type = 'JWT'
```
clintonb, line 86: This variable is not necessary.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -98,27 +83,22 @@ def handle(self, *args, **options):
...
-                    access_token, __ = EdxRestApiClient.get_oauth_access_token(
-                        '{root}/access_token'.format(root=partner.oidc_url_root.strip('/')),
-                        partner.oidc_key,
-                        partner.oidc_secret,
-                        token_type=token_type
-                    )
-                except Exception:
-                    logger.exception('No access token provided or acquired through client_credential flow.')
-                    raise
+            logger.info('No access token provided. Retrieving access token using client_credential flow...')
```
clintonb: Update this: `Retrieving access token for partner [{partner.short_code}]...`

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -38,9 +38,9 @@ def mock_apis(self):
         self.mock_marketing_courses_api()
         self.mock_programs_api()
 
-    def mock_access_token_api(self, requests_mock=None):
+    def mock_access_token_api(self, requests_mock=None, callback=None):
         body = {
-            'access_token': ACCESS_TOKEN,
+            'access_token': str(jwt.encode({'preferred_username': 'bob'}, 'secret', algorithm='HS256'), 'utf-8'),
```
clintonb: Store this in the constant.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -136,9 +136,16 @@ def test_refresh_course_metadata_with_invalid_partner_code(self):
 
     def test_refresh_course_metadata_with_no_token_type(self):
         """ Verify an error is raised if an access token is passed in without a token type. """
-        with self.assertRaises(CommandError):
-            command_args = ['--access_token=test-access-token']
-            call_command('refresh_course_metadata', *command_args)
+        with responses.RequestsMock() as rsps:
+            def raise_exception():
```
clintonb: This seems dirty. Mock `EdxRestApiClient.get_oauth_access_token()` so that it raises an exception.

clintonb (changes requested):

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: isn't that what I'm doing in mock_access_token_api?

clintonb: The idea is the same. However, `mock_access_token_api` mocks the HTTP response, not `EdxRestApiClient.get_oauth_access_token()`. 

I find the implementation below to be a bit cleaner than making fake functions.

``` python
with mock.patch('edx_rest_api_client.clientEdxRestApiClient.get_oauth_access_token', side_effect=Exception):
  ...
```

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
MatthewPiatetsky: It is used in a few places in this file. I could remove the need for specifying token type altogether, but imo it is helpful to explicitly clarify what the token_type is so people have an easier time understanding the auth.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: got it, thanks!

MatthewPiatetsky:
@edx/ecommerce talked with @rlucioni @mikedikan during standup I believe the approach we settled on was finish this PR to get things working and simultaneously settle on an approach that everyone is happy with on the other discussion, and then create a dev story and pick it up (potentially next sprint)

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -98,27 +83,22 @@ def handle(self, *args, **options):
...
+                access_token, __ = EdxRestApiClient.get_oauth_access_token(
+                    '{root}/access_token'.format(root=partner.oidc_url_root.strip('/')),
+                    partner.oidc_key,
+                    partner.oidc_secret,
+                    token_type=token_type
+                )
+            except Exception:
+                logger.exception('No access token acquired through client_credential flow.')
+                raise
+            username = jwt.decode(access_token, verify=False)['preferred_username']
```
rlucioni, line 100: Why are you passing `verify=False` here? I understand that all we need to do here is read the contents of the token, but is it wise to do away with token verification in this context?

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -19,8 +20,8 @@
 from course_discovery.apps.course_metadata.tests import toggle_switch
 from course_discovery.apps.course_metadata.tests.factories import CourseFactory
 
-ACCESS_TOKEN = 'secret'
 JSON = 'application/json'
+ACCESS_TOKEN = str(jwt.encode({'preferred_username': 'bob'}, 'secret', algorithm='HS256'), 'utf-8')
```
rlucioni: The algorithm should already be specified in settings. Do you really need to provide it again here?

docs/course_metadata.rst
```diff
@@ -26,7 +26,7 @@ For example, if you use the JWT access token "secret-ecommerce-key" to authentic
 
 .. code-block:: bash
 
-    $ ./manage.py refresh_course_metadata --access_token secret-ecommerce-key --token_type jwt
+    $ ./manage.py refresh_course_metadata
```
rlucioni, line 28: Also change the line above this that refers to the JWT. Maybe make a note that the command retrieves it's own access token via the client credentials grant.

rlucioni (changes requested):

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
MatthewPiatetsky: I was getting a signature verification error presumably because I don't know what the secret is if the secret is stored in LMS, so I can't verify the signature if I'm not in the LMS

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: changed

docs/course_metadata.rst
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
clintonb: We do have the secret. Look at `settings.JWT_AUTH['JWT_SECRET']`.

MatthewPiatetsky: But the token gets generated in the LMS and verified by the LMS.
We do not pass the JWT_SECRET to LMS when it generates its token, only the OpenID client key/secret pair.
So the JWT_SECRET in course_discovery could be different from the one in LMS, (and probably is) right?

rlucioni: We currently use symmetric keys. The secrets are the same.

MatthewPiatetsky: Locally, JWT_SECRET_KEY is not set for either, so it should fall back to SECRET_KEY
I checked the SECRET_KEY in my local catalog and local platform and they are not the same

rlucioni: They are the same in working, production-like environments. We define values for JWT_SECRET_KEY in private settings specific to each developer.

MatthewPiatetsky: so what should we do to make it work locally?
And what if I want to point my local catalog to stage or prod LMS, if I don't want to run my local LMS
Then I don't know what those secrets are, or if I am able to go to the environment and find out I will need to update my local settings everytime the secret changes
This seems like a major complication to use this command locally

rlucioni: That's a fair point, forcing verification here would require access to the signing token from the LMS you're pointing at. What's at stake if we opt not to verify the token here? We might end up trying to use a bad username on some requests? That doesn't seem like a big deal. Thoughts?

MatthewPiatetsky: Since LMS is the one actually responsible for generating and verifying the token, it doesn't seem necessary to me to verify it on the catalog side. I would hope that if LMS cannot create a token, it would fail on token creation, rather than giving back some sort of broken token and the catalog not knowing it failed

rlucioni: @MatthewPiatetsky anyone using/trusting the token is responsible for verifying its validity, not just the LMS. That said, in this context, I don't see any harm in extracting the username without checking the signature. I'm in favor of what you have here, unless @clintonb is opposed.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -38,7 +39,7 @@ def mock_apis(self):
         self.mock_marketing_courses_api()
         self.mock_programs_api()
 
-    def mock_access_token_api(self, requests_mock=None):
+    def mock_access_token_api(self, requests_mock=None, callback=None):
```
clintonb: The changes here are not necessary.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
clintonb: This is fine as-is.

docs/course_metadata.rst
```diff
@@ -21,12 +21,11 @@ Retrieving Course Metadata
 The ``refresh_course_metadata`` command in :file:`course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py` is used to retrieve metadata. This is run daily in production through a Jenkins job, and can be manually run to set up your local environment. The data loaders are each run in series by the command. The data
 loaders should be idempotent -- that is, running this command once will populate the database, and if nothing has
 changed upstream then running it again should not change the database.
-
-For example, if you use the JWT access token "secret-ecommerce-key" to authenticate API calls, run the following:
+This command retrieves a JWT access token from the LMS instance that your OpenID Connect URL points to.
```
clintonb: The command retrieves a JWT access token from the configured OpenID Connect provider.

clintonb (approved):
Remember to squash your commits, and delete your branch after merging.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: thanks, changed

docs/course_metadata.rst
MatthewPiatetsky: changed

rlucioni (approved):

PR #425, opened 2016-11-07
Update New Relic Agent to avoid bug

mikedikan:
ECOM-6312

@rlucioni for review
FYI @edx/ecommerce

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -1,1 +1,1 @@
-newrelic==2.58.1.44
+newrelic==2.72.1.53


clintonb (approved):
Correct your commit message: "Updated New Relic dependency". This has nothing to do with bugs.

rlucioni (approved):

mikedikan:
@clintonb the jira story mentioned this was as a result of a bug: https://openedx.atlassian.net/browse/ECOM-6297 - is that inaccurate?

clintonb:
Got it. Carry on.

PR #426, opened 2016-11-07
Remove hardcoded username

MatthewPiatetsky:
@edx/ecommerce


rlucioni:
@MatthewPiatetsky why not just do this as part of #424?

rlucioni:
Let me rephrase: it seems like #424 obviates the need for this.

MatthewPiatetsky:
@rlucioni I put this up in case we decided we did not want to do that, or want to merge this earlier. Seems like merging anything is a moot point for now until we figure out what is happening with the stage catalog

PR #440, opened 2016-11-15
ECOM-6376 Add kwargs to parallel loader

MatthewPiatetsky:
@edx/ecommerce 
Thanks @schenedx for finding this issue.  
Didn't notice this locally before because I didn't have the waffle switch on.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
@@ -17,6 +17,7 @@ from course_discovery.apps.course_metadata.data_loaders.marketing_site import (
     SponsorMarketingSiteDataLoader, PersonMarketingSiteDataLoader, CourseMarketingSiteDataLoader
 )
 from course_discovery.apps.course_metadata.data_loaders.tests import mock_data
+from course_discovery.apps.course_metadata.management.commands.refresh_course_metadata import execute_parallel_loader
 from course_discovery.apps.course_metadata.tests import toggle_switch
 from course_discovery.apps.course_metadata.tests.factories import CourseFactory
 
@@ -29,7 +30,18 @@ class RefreshCourseMetadataCommandTests(TransactionTestCase):
     def setUp(self):
         super(RefreshCourseMetadataCommandTests, self).setUp()
         self.partner = PartnerFactory()
-
+        partner = self.partner
+        self.pipeline = [(SubjectMarketingSiteDataLoader, partner.marketing_site_url_root, None),
+                         (SchoolMarketingSiteDataLoader, partner.marketing_site_url_root, None),
+                         (SponsorMarketingSiteDataLoader, partner.marketing_site_url_root, None),
+                         (PersonMarketingSiteDataLoader, partner.marketing_site_url_root, None),
+                         (CourseMarketingSiteDataLoader, partner.marketing_site_url_root, None),
+                         (OrganizationsApiDataLoader, partner.organizations_api_url, None),
+                         (CoursesApiDataLoader, partner.courses_api_url, None),
+                         (EcommerceApiDataLoader, partner.ecommerce_api_url, 1),
+                         (ProgramsApiDataLoader, partner.programs_api_url, None),
+                         (XSeriesMarketingSiteDataLoader, partner.marketing_site_url_root, None)]
+        self.kwargs = {'username': 'bob'}
         self.mock_access_token_api()
 
     def mock_apis(self):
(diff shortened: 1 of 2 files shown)


schenedx (approved):
👍

rlucioni (approved):

clintonb (changes requested):
This needs tests. Clearly, the lack of tests allowed this bug to go unnoticed.

MatthewPiatetsky:
@clintonb looks like there actually is a test with the flag on and in fact the job does not run properly when the flag is on and we are not asserting that it worked successfully
Looking into it

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -29,7 +30,18 @@ class RefreshCourseMetadataCommandTests(TransactionTestCase):
     def setUp(self):
         super(RefreshCourseMetadataCommandTests, self).setUp()
         self.partner = PartnerFactory()
-
+        partner = self.partner
```
schenedx, line 33: Why do you need this line? Just to save `self.` in the statements below?

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -126,8 +151,14 @@ def test_refresh_course_metadata(self, is_parallel):
             # courses, the command won't risk race conditions between threads trying to
             # create the same course.
             CourseFactory(partner=self.partner)
+            with mock.patch('concurrent.futures.ProcessPoolExecutor.submit') as mock_executor:
```
schenedx, line 154: Instead of mock `concurrent.futures.ProcessPoolExecutor.submit`, don't you want to mock the `execute_parallel_loader` function instead? And check the call arguments to ensure `**loader_kwargs` is passed in?

schenedx (changes requested):
I think the test is checking the wrong function for parallel execution.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: yeah

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
MatthewPiatetsky: hmm..well if I don't mock the original call then I might as well mock execute_loader, right?

schenedx: Yeah, mock `execute_loader` is fine. Make sure you check the arguments passed in

MatthewPiatetsky: actually I just tried both and seems both do not work with this approach (for the same reasons that other approaches were not working, where the mocks don't work on the subprocesses) thoughts?

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -112,11 +124,24 @@ def mock_programs_api(self):
         )
         return bodies
 
-    @ddt.data(True, False)
-    def test_refresh_course_metadata(self, is_parallel):
-        if is_parallel:
-            for name in ['threaded_metadata_write', 'parallel_refresh_pipeline']:
-                toggle_switch(name)
+    def test_refresh_course_metadata(self):
```
rlucioni: If you're going to break things up like this, make it clear in the name of this test that the pipeline is being executed serially.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
rlucioni: @MatthewPiatetsky what you have here is what I was suggesting earlier. I think this is the closest we can get without figuring out how to set up mocks in separate processes.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -112,11 +124,24 @@ def mock_programs_api(self):
...
+
+            with mock.patch('course_discovery.apps.course_metadata.management.commands.'
+                            'refresh_course_metadata.execute_loader') as mock_executor:
+                call_command('refresh_course_metadata')
+
+                # Set up expected calls
+                expected_calls = [mock.call(loader_class, self.partner, api_url,
+                                            ACCESS_TOKEN, 'JWT', max_workers_override or 7, False, **self.kwargs)
+                                  for loader_class, api_url, max_workers_override in self.pipeline]
+                mock_executor.assert_has_calls(expected_calls, any_order=True)
```
rlucioni: In this serial case, the order of the calls is deterministic. The calls should be executed in the order defined by the pipeline.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
```diff
@@ -126,8 +151,14 @@ def test_refresh_course_metadata(self, is_parallel):
             # courses, the command won't risk race conditions between threads trying to
             # create the same course.
             CourseFactory(partner=self.partner)
+            with mock.patch('concurrent.futures.ProcessPoolExecutor.submit') as mock_executor:
+                call_command('refresh_course_metadata')
 
-            call_command('refresh_course_metadata')
+                # Set up expected calls
+                expected_calls = [mock.call(execute_parallel_loader, loader_class, self.partner, api_url,
```
rlucioni: I would spread this across several lines to make it easier to read.

course_discovery/apps/course_metadata/management/commands/tests/test_refresh_course_metadata.py
schenedx: @rlucioni @MatthewPiatetsky I think I get it now. This works for now.

schenedx (approved):
👍

clintonb (approved):
Squash your commits, and add the ticket number to the commit description.

MatthewPiatetsky:
@clintonb added the ticket number, but I had intentionally made two commits, because it seemed like the tests were not directly related to the other change. However, I guess we generally don't make separate commits for tests, so will squash

clintonb:
@MatthewPiatetsky the tests are definitely related. We prefer to fix bugs once. The key to ensuring that is a good test suite. A bug fix is incomplete without tests.

PR #442, opened 2016-11-16
Pin version of requests at 2.9.1

bderusha:
ECOM-6380

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -40,3 +40,4 @@ pillow==3.3.0
 pycountry==1.20
 python-dateutil==2.5.3
 pytz==2016.6.1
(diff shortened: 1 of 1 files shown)


bderusha:
FYI @edx/ecommerce

requirements/base.txt
```diff
@@ -40,3 +40,4 @@ pillow==3.3.0
 pycountry==1.20
 python-dateutil==2.5.3
 pytz==2016.6.1
+requests==2.9.1
```
clintonb, line 43: Please add a comment explaining why this is necessary. Otherwise, I will probably wipe it out the next time I update requirements.

PR #445, opened 2016-11-17
Updated cryptography and pillow libraries

clintonb:
These update include security fixes.


clintonb:
FYI @edx/ecommerce 

@edx/devops should I be concerned that I learned of these issues from Gemnasium rather than VersionEye?

rlucioni (approved):

PR #446, opened 2016-11-17
ECOM-6377 Add fields to MinimalCourseRunSerializer

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -331,11 +331,15 @@ class MinimalCourseRunSerializer(TimestampModelSerializer):
 
     @classmethod
     def prefetch_queryset(cls):
-        return CourseRun.objects.all().select_related('course').prefetch_related('course__partner')
+        return CourseRun.objects.all().select_related('course').prefetch_related(
+            'course__partner',
+            Prefetch('seats', queryset=SeatSerializer.prefetch_queryset()),
+        )
 
     class Meta:
         model = CourseRun
-        fields = ('key', 'uuid', 'title', 'image', 'short_description', 'marketing_url',)
+        fields = ('key', 'uuid', 'title', 'image', 'short_description', 'marketing_url',
+                  'start', 'end', 'enrollment_start', 'enrollment_end', 'pacing_type', 'type',)
 
     def get_marketing_url(self, obj):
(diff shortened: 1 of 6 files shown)


schenedx:
@MatthewPiatetsky Should we be concerned about the performance impact this change makes? Like would the connection between LMS and Discovery go down because of this change? 
I guess, a related question is, does this change increase the database count for related unit tests?

MatthewPiatetsky:
@schenedx yeah I noticed that those tests are failing, specifically the reason is that if we want to retrieve type it is a property. We can't prefetch the property, though we could prefetch seats and potentially use cached_property instead of property. I will do a bit more looking, but not sure exactly what we want to do yet

course_discovery/apps/api/serializers.py
```diff
@@ -335,7 +335,8 @@ def prefetch_queryset(cls):
 
     class Meta:
         model = CourseRun
-        fields = ('key', 'uuid', 'title', 'image', 'short_description', 'marketing_url',)
+        fields = ('key', 'uuid', 'title', 'image', 'short_description', 'marketing_url',
+                  'start', 'end', 'enrollment_start', 'enrollment_end', 'pacing_type', 'type',)
```
rlucioni, line 342: You'll want to move the seat prefetching to account for moving type.

``` py
return CourseRun.objects.all().select_related('course').prefetch_related(
    'course__partner',
    Prefetch('seats', queryset=SeatSerializer.prefetch_queryset()),
)
```

rlucioni:
@MatthewPiatetsky after you move the prefetching, can you take a look at the django debug toolbar before and after moving those fields to see what the difference in query counts is?

MatthewPiatetsky:
@rlucioni I queried http://localhost:8008/api/v1/programs/ on master and got 11 queries, with this branch I get 29 (and without the cached_property additions I get 47). Am I missing something here with the prefetching? Or are the properties not taking advantage of it?

MatthewPiatetsky:
@rlucioni Now the query I mentioned above only adds 2 queries. I think the issue was with value_list, using a list comprehension seems to resolve the issue.

rlucioni (approved):
Very nice @MatthewPiatetsky!

schenedx (approved):
👍

PR #447, opened 2016-11-21
ECOM-4738 Add Typeahead endpoint to course discovery

MatthewPiatetsky:
@edx/ecommerce 
Query: `http://localhost:8008/api/v1/search/typeahead/?q=test`
Response: 
```
{
    "course_runs": [
        {
            "title": "Introduction to Solar Systems Astronomy",
            "key": "course-v1:edx-testing+edx-testing+cp-testing",
            "additional_details": "edx-testing"
        },...
    ],
    "programs": [
        {
            "uuid": "66f95b94-5a6c-44db-bf8f-957483912200",
            "title": "Andy's Test MicroMasters",
            "type": "XSeries",
            "additional_details": "edX"
        },...
    ]
}
```

Will return 0-3 course_runs and 0-3 programs depending on total available results


course_discovery/apps/api/serializers.py
```diff
@@ -937,6 +950,20 @@ class Meta:
         index_classes = [ProgramIndex]
 
 
+class TypeaheadProgramSearchSerializer(HaystackSerializer):
+    alias = serializers.SerializerMethodField()
+
+    def get_alias(self, result):
+        authoring_organizations = [json.loads(org) for org in result.authoring_organization_bodies]
```
MatthewPiatetsky, line 960: The org key is stored as its own field in the index for course_runs, but not programs

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,24 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+
+class TypeaheadSearchViewSet(BaseHaystackViewSet):
+    """
+    Typeahead for courses and programs.
+    """
+
+    serializer_class = serializers.TypeaheadSearchSerializer
+    index_models = (CourseRun, Program,)
+
+    def list(self, request, *args, **kwargs):
```
MatthewPiatetsky, line 691: There may be a cleaner way to do this using DRF/Haystack built ins that I am not familiar with

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,24 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+        response = super(TypeaheadSearchViewSet, self).list(request, *args, **kwargs)
+        results = response.data['results']
+
+        course_runs = [course_run for course_run in results if course_run.get('content_type') == 'courserun'][:3]
```
rlucioni: What up with the array slicing here and below? Is there a more readable way to do this?

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,24 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+        programs = [program for program in results if program.get('content_type') == 'program'][:3]
+        for item in course_runs + programs:
+            item.pop('content_type', None)
```
rlucioni: You're using three loops here to do what you could do with one. You should only have to iterate across results once, decide if something is a course run or a program, and pop the content type.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -279,3 +295,24 @@ def test_empty_query(self):
         response_data = json.loads(response.content.decode('utf-8'))
         self.assertListEqual(response_data['objects']['results'],
                              [self.serialize_course_run(course_run), self.serialize_program(program)])
+
+
+class TypeaheadSearchViewSet(TypeaheadSerializationMixin, LoginMixin, APITestCase):
+    path = reverse('api:v1:search-typeahead-list')
+
+    def get_typeahead_response(self, querystring=None):
+        querystring = querystring or {}
```
rlucioni: This keyword argument doesn't appear to be used. Remove it?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1098,3 +1098,50 @@ def test_data_without_organizations(self):
...
+        self.assertDictEqual(serialized_course.data, expected)
+
+    def serialize_course_run(self, course_run):
+        """ Serializes the given `CourseRun` as a typeahead result. """
+        result = SearchQuerySet().models(CourseRun).filter(key=course_run.key)[0]
+        serializer = TypeaheadCourseRunSearchSerializer(result)
+        return serializer
+
+
+class TypeaheadProgramSearchSerializerTests(TestCase):
```
rlucioni, line 1125: This test class is very similar to the one above. Have you considered implementing a single shared `TypeaheadSerializerTestBase` which you can then subclass for testing course run and program serialization?

course_discovery/apps/api/serializers.py
```diff
@@ -912,6 +912,19 @@ class Meta:
         index_classes = [CourseRunIndex]
 
 
+class TypeaheadCourseRunSearchSerializer(HaystackSerializer):
+    alias = serializers.SerializerMethodField()
+
+    def get_alias(self, result):
```
rlucioni: I would either leave a comment about what an "alias" is, or I would choose a better name for it to begin with.

rlucioni (changes requested):

course_discovery/apps/api/serializers.py
MatthewPiatetsky: This was the preexisting name in the current typeahead, but I think renaming would be a good idea. Any ideas?

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: Could you clarify how this would work, given that all the methods are slightly different I'm not sure what would be in the base class

course_discovery/apps/api/serializers.py
MatthewPiatetsky: Nvm Jasper said it is "literally an alias" so I added a comment

rlucioni: So is "alias" a standard term for that? I was thinking something along the lines of "caption" or "subtitle" would be more descriptive.

course_discovery/apps/api/tests/test_serializers.py
rlucioni: I guess you'd just end up with a single method in the base class which serialized some object and compared it to an expected structure. Maybe not the most useful thing, in retrospect. Feel free to ignore.

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,25 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+        response = super(TypeaheadSearchViewSet, self).list(request, *args, **kwargs)
+        results = response.data['results']
+        course_runs, programs = [], []
+        for item in results:
+            item_type = item.pop('content_type', None)
+            if item_type == 'courserun':
+                course_runs.append(item)
+            elif item_type == 'program':
+                programs.append(item)
+        response.data = {'course_runs': course_runs[:3], 'programs': programs[:3]}
```
rlucioni: Please include a comment explaining that the slicing is meant to limit the size of each category to 3 results.

Also, thoughts about exiting the loop above when the `course_runs` and `programs` contain 3 items? I can see that loop doing a bunch of wasted work.

MatthewPiatetsky: hmm..adding the if statement would also be some overhead, but I suppose it limits the worst case (in the average case it probably depends on the query, some would have few results and some would have many), makes sense to me

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -26,6 +28,20 @@ def serialize_program(self, program):
         return ProgramSearchSerializer(result).data
 
 
+class TypeaheadSerializationMixin:
+    def serialize_course_run(self, course_run):
+        result = SearchQuerySet().models(CourseRun).filter(key=course_run.key)[0]
+        data = TypeaheadCourseRunSearchSerializer(result).data
+        data.pop('content_type')
```
mikedikan, line 37: Can you add a comment about why you remove content_type from the response?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -26,6 +28,20 @@ def serialize_program(self, program):
...
+        return data
+
+    def serialize_program(self, program):
+        result = SearchQuerySet().models(Program).filter(uuid=program.uuid)[0]
+        data = TypeaheadProgramSearchSerializer(result).data
+        data.pop('content_type')
```
mikedikan, line 44: see above

course_discovery/apps/api/serializers.py
schenedx: @MatthewPiatetsky I think this is confusing to me as well. Why do you think this `alias` name here makes more sense? Even with the comment, I am confused.

schenedx (approved):
Just a question on `Alias` field. Rest LGTM 👍

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: See the sample response in my description, we know the type based on which list each item is a part of, so we don't need the content_type

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: let me know if this is what you were thinking

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,29 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+    index_models = (CourseRun, Program,)
+
+    def list(self, request, *args, **kwargs):
+        response = super(TypeaheadSearchViewSet, self).list(request, *args, **kwargs)
+        results = response.data['results']
+        course_runs, programs = [], []
+        # Retrieve first 3 results for course runs and programs
+        for item in results:
+            # Items are grouped by content type so we don't need it in the response
+            item_type = item.pop('content_type', None)
```
rlucioni, line 697: You appear to be doing this in the serializer. Is it necessary here as well?

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,29 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+            if item_type == 'courserun' and len(course_runs) < 3:
+                course_runs.append(item)
+            elif item_type == 'program' and len(programs) < 3:
+                programs.append(item)
+            elif len(programs) == 3 and len(course_runs) == 3:
```
rlucioni: Let's try to avoid hardcoding magic numbers like this. The result count (3, in this case) should be a constant defined at the top of the file, or in settings.

course_discovery/apps/api/v1/views.py
rlucioni: I'm sorry, I was looking at test code. However, could you do this in the serializer by just excluding the `content_type` field?

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +676,29 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+                break
```
rlucioni, line 705: I also don't see a test case covering this branch.

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: I was including the content_type field there so I would know if something is a courserun or program. Otherwise I wouldn't know

rlucioni: Understood, ignore me.

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: fixed

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: good point, didn't think of that when I addressed your other comment. added.

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +678,29 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+    """
+
+    serializer_class = serializers.TypeaheadSearchSerializer
+    index_models = (CourseRun, Program,)
+
+    def list(self, request, *args, **kwargs):
+        response = super(TypeaheadSearchViewSet, self).list(request, *args, **kwargs)
+        results = response.data['results']
+        course_runs, programs = [], []
+        # Retrieve first 3 results for course runs and programs
```
rlucioni: This comment is going to get stale if the result count value changes. Your code is now clear enough that I don't think this comment is necessary, so I'd remove it.

MatthewPiatetsky: ok removed

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -279,3 +297,31 @@ def test_empty_query(self):
...
+        program = ProgramFactory()
+        response = self.get_typeahead_response()
+        self.assertEqual(response.status_code, 200)
+        response_data = json.loads(response.content.decode('utf-8'))
+        self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
+                                             'programs': [self.serialize_program(program)]})
+
+    def test_typeahead_multiple_results(self):
+        """ Test typeahead response with 3 course_runs and programs. """
+        course_runs = CourseRunFactory.create_batch(3)
```
rlucioni: Don't assume 3, import the constant you defined in the views.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -279,3 +297,31 @@ def test_empty_query(self):
...
+        """ Test typeahead response. """
+        course_run = CourseRunFactory()
+        program = ProgramFactory()
+        response = self.get_typeahead_response()
+        self.assertEqual(response.status_code, 200)
+        response_data = json.loads(response.content.decode('utf-8'))
+        self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
+                                             'programs': [self.serialize_program(program)]})
+
+    def test_typeahead_multiple_results(self):
```
rlucioni, line 319: Since this test case only creates a `RESULT_COUNT` number of course and programs, you're not going to hit the break you've added. You could do `RESULT_COUNT + 1` to exercise all branches.

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: good idea, changed

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: actually good point, this makes me realize that the condition is currently incorrect, it should be >= RESULT_COUNT-1

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -8,12 +8,15 @@
...
-from course_discovery.apps.api.serializers import CourseRunSearchSerializer, ProgramSearchSerializer
+from course_discovery.apps.api.serializers import (CourseRunSearchSerializer, ProgramSearchSerializer,
+                                                   TypeaheadCourseRunSearchSerializer, TypeaheadProgramSearchSerializer)
+
 from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD, PartnerFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import CourseRun, Program
 from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, ProgramFactory
+from course_discovery.apps.api.v1.views import RESULT_COUNT
```
rlucioni: Nitpick: keep your imports in alphabetical order, please.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -279,3 +298,31 @@ def test_empty_query(self):
...
+    def get_typeahead_response(self):
+        return self.client.get(self.path)
+
+    def test_typeahead(self):
+        """ Test typeahead response. """
+        course_run = CourseRunFactory()
+        program = ProgramFactory()
+        response = self.get_typeahead_response()
+        self.assertEqual(response.status_code, 200)
+        response_data = json.loads(response.content.decode('utf-8'))
```
rlucioni: Can you use https://docs.djangoproject.com/en/1.9/topics/testing/tools/#django.test.Response.json?

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/tests/test_views/test_search.py

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -279,3 +298,30 @@ def test_empty_query(self):
...
+                                             'programs': [self.serialize_program(program)]})
+
+    def test_typeahead_multiple_results(self):
+        """ Test typeahead response with max number of course_runs and programs. """
+        course_runs = CourseRunFactory.create_batch(RESULT_COUNT + 1)
+        programs = ProgramFactory.create_batch(RESULT_COUNT + 1)
+        response = self.get_typeahead_response()
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+        self.assertEqual(len(response_data['course_runs']), 3)
```
rlucioni: Again, don't assume 3. You want to assert that the result count equals `RESULT_COUNT`.

course_discovery/apps/api/v1/views.py
```diff
@@ -676,3 +678,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+    index_models = (CourseRun, Program,)
+
+    def list(self, request, *args, **kwargs):
+        response = super(TypeaheadSearchViewSet, self).list(request, *args, **kwargs)
+        results = response.data['results']
+        course_runs, programs = [], []
+        for item in results:
+            # Items are grouped by content type so we don't need it in the response
+            item_type = item.pop('content_type', None)
+            if item_type == 'courserun' and len(course_runs) < RESULT_COUNT:
```
rlucioni: Minor, but it's a little cleaner to avoid recomputing these lengths. Just do each one once:
```py
course_run_count = len(course_run)
program_count = len(programs)

if item_type == 'courserun' and course_run_count < RESULT_COUNT:
    ...
```

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1098,3 +1098,50 @@ def test_data_without_organizations(self):
...
+            'uuid': str(program.uuid),
+            'title': program.title,
+            'type': program.type.name,
+            'content_type': 'program',
+            'additional_details': program.authoring_organizations.first().key
+        }
+
+    def test_data(self):
+        authoring_organization = OrganizationFactory()
+        program = ProgramFactory(authoring_organizations=[authoring_organization])
```
rlucioni, line 1137: What about the case of a program with two authoring organizations? Your code handles that gracefully, but I don't see it covered here.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -279,3 +298,30 @@ def test_empty_query(self):
...
+        response = self.get_typeahead_response()
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+        self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
+                                             'programs': [self.serialize_program(program)]})
+
+    def test_typeahead_multiple_results(self):
+        """ Test typeahead response with max number of course_runs and programs. """
+        course_runs = CourseRunFactory.create_batch(RESULT_COUNT + 1)
+        programs = ProgramFactory.create_batch(RESULT_COUNT + 1)
```
rlucioni: These variables (`course_runs`, `programs`) are unused.

course_discovery/apps/api/v1/tests/test_views/test_search.py

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: not sure if this actually creates savings, but changed

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: added test in serializer and view tests

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: removed

rlucioni (approved):

PR #448, opened 2016-11-23
Multiple front-end fixes

clintonb:


clintonb:
@edx/ecommerce please review

schenedx (approved):
👍

rlucioni (approved):
I'm not sure what changes the new pattern library makes, but it may be worth checking that the publisher continues to work as expected.

PR #450, opened 2016-11-23
Publisher front-end fixes

clintonb:
The publisher HTML and JS has been properly organized so that this functionality is not inherited when extending the base template for the entire project.

ECOM-6466


clintonb:
@edx/ecommerce please review

rlucioni (approved):

PR #454, opened 2016-11-27
course discovery with 16.04

bilalahmad99:


docker-compose.yml
```diff
@@ -11,8 +11,6 @@ services:
       MYSQL_USER: "discov001"
       MYSQL_PASSWORD: "password"
       MYSQL_DATABASE: "discovery"
-    ports:
```
clintonb, line 14: Shouldn't tests be using a Travis-specific `docker-compose.yml`?

feanil: We can do that as well. I was hoping to avoid it because then you wouldn't have to make similar changes in two different compose files.  So far the ports are the first issue that makes using a different compose file necessary.  This change got me past that error so I'll make a separate test compose file with all the ports removed.

feanil:
@clintonb I believe this is now ready for final review along with https://github.com/edx/configuration/pull/3493

feanil:
@edx/devops one of you should also review this.

.travis/docker-compose-travis.yml
```diff
@@ -0,0 +1,39 @@
...
+      MYSQL_DATABASE: "discovery"
+  es:
+    image: elasticsearch:1.5.2
+    container_name: es
+  memcache:
+    image: memcached:1.4.24
+    container_name: memcache
+  course-discovery:
+    # Uncomment this line to use the official course-discovery base image
+    image: edxops/course-discovery:bilal-test
```
fredsmith: do we want to give this a more completed sounding tag name before we merge?

fredsmith (approved):

.travis/docker-compose-travis.yml
feanil: yes, before I merge I'll push up a new latest course-discovery image and change this back to latest

clintonb (approved):

PR #459, opened 2016-11-29
ECOM-4738 Add partial term matching to typeahead 

MatthewPiatetsky:
Tweaking my previous version of this endpoint because I realized I hadn't tested that partial terms returned results (currently our regular search results do not do this).
Talked with Jasper and we discussed searching just on title (and preserving search result boosting).
I'm not sure about whether this is the best way to do this, but it works.
Thoughts?
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -913,20 +913,6 @@ class CourseRunSearchSerializer(HaystackSerializer):
         index_classes = [CourseRunIndex]
 
 
-class TypeaheadCourseRunSearchSerializer(HaystackSerializer):
-    additional_details = serializers.SerializerMethodField()
-
-    def get_additional_details(self, result):
-        """ Value of the grey text next to the typeahead result title. """
-        return result.org
-
-    class Meta:
-        field_aliases = COMMON_SEARCH_FIELD_ALIASES
-        fields = ['key', 'title', 'content_type']
-        ignore_fields = COMMON_IGNORED_FIELDS
-        index_classes = [CourseRunIndex]
-
-
 class CourseRunFacetSerializer(BaseHaystackFacetSerializer):
     serialize_objects = True
 
@@ -948,21 +934,6 @@ class ProgramSearchSerializer(HaystackSerializer):
         field_aliases = COMMON_SEARCH_FIELD_ALIASES
         field_options = PROGRAM_FACET_FIELD_OPTIONS
         fields = PROGRAM_SEARCH_FIELDS
-        ignore_fields = COMMON_IGNORED_FIELDS
-        index_classes = [ProgramIndex]
-
-
-class TypeaheadProgramSearchSerializer(HaystackSerializer):
-    additional_details = serializers.SerializerMethodField()
-
-    def get_additional_details(self, result):
-        """ Value of the grey text next to the typeahead result title. """
-        authoring_organizations = [json.loads(org) for org in result.authoring_organization_bodies]
-        return ', '.join([org['key'] for org in authoring_organizations])
-
-    class Meta:
-        field_aliases = COMMON_SEARCH_FIELD_ALIASES
-        fields = ['uuid', 'title', 'content_type', 'type']
         ignore_fields = COMMON_IGNORED_FIELDS
         index_classes = [ProgramIndex]
 
@@ -990,15 +961,32 @@ class AggregateSearchSerializer(HaystackSerializer):
         }
 
 
-class TypeaheadSearchSerializer(HaystackSerializer):
+class TypeaheadCourseRunSearchSerializer(serializers.Serializer):
+    org = serializers.CharField()
+    title = serializers.CharField()
+    key = serializers.CharField()
+
     class Meta:
-        field_aliases = COMMON_SEARCH_FIELD_ALIASES
-        fields = COURSE_RUN_SEARCH_FIELDS + PROGRAM_SEARCH_FIELDS
-        ignore_fields = COMMON_IGNORED_FIELDS
-        serializers = {
-            ProgramIndex: TypeaheadProgramSearchSerializer,
-            CourseRunIndex: TypeaheadCourseRunSearchSerializer,
-        }
+        fields = ['key', 'title']
+
+
+class TypeaheadProgramSearchSerializer(serializers.Serializer):
+    orgs = serializers.SerializerMethodField()
+    uuid = serializers.CharField()
+    title = serializers.CharField()
+    type = serializers.CharField()
+
+    def get_orgs(self, result):
+        authoring_organizations = [json.loads(org) for org in result.authoring_organization_bodies]
+        return [org['key'] for org in authoring_organizations]
+
+    class Meta:
+        fields = ['uuid', 'title', 'type']
+
+
+class TypeaheadSearchSerializer(serializers.Serializer):
+    course_runs = TypeaheadCourseRunSearchSerializer(many=True)
+    programs = TypeaheadProgramSearchSerializer(many=True)
 
 
 class AggregateFacetSearchSerializer(BaseHaystackFacetSerializer):
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -1110,8 +1110,7 @@ class TypeaheadCourseRunSearchSerializerTests(TestCase):
         expected = {
             'key': course_run.key,
             'title': course_run.title,
-            'content_type': 'courserun',
-            'additional_details': course_run_key.org
+            'org': course_run_key.org
         }
         self.assertDictEqual(serialized_course.data, expected)
 
@@ -1128,16 +1127,13 @@ class TypeaheadProgramSearchSerializerTests(TestCase):
             'uuid': str(program.uuid),
             'title': program.title,
             'type': program.type.name,
-            'content_type': 'program',
-            'additional_details': program.authoring_organizations.first().key
+            'orgs': list(program.authoring_organizations.all().values_list('key', flat=True))
         }
 
     def test_data(self):
         authoring_organization = OrganizationFactory()
         program = ProgramFactory(authoring_organizations=[authoring_organization])
-
         serialized_program = self.serialize_program(program)
-
         expected = self._create_expected_data(program)
         self.assertDictEqual(serialized_program.data, expected)
 
@@ -1145,8 +1141,8 @@ class TypeaheadProgramSearchSerializerTests(TestCase):
         authoring_organizations = OrganizationFactory.create_batch(3)
         program = ProgramFactory(authoring_organizations=authoring_organizations)
         serialized_program = self.serialize_program(program)
-        expected = ', '.join([org.key for org in authoring_organizations])
-        self.assertEqual(serialized_program.data['additional_details'], expected)
+        expected = [org.key for org in authoring_organizations]
+        self.assertEqual(serialized_program.data['orgs'], expected)
 
     def serialize_program(self, program):
         """ Serializes the given `Program` as a typeahead result. """
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -10,8 +10,7 @@ from rest_framework.test import APITestCase
 
 from course_discovery.apps.api.serializers import (CourseRunSearchSerializer, ProgramSearchSerializer,
                                                    TypeaheadCourseRunSearchSerializer, TypeaheadProgramSearchSerializer)
-from course_discovery.apps.api.v1.views import RESULT_COUNT
-
+from course_discovery.apps.api.v1.views import TypeaheadSearchView
 from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD, PartnerFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
@@ -33,15 +32,11 @@ class TypeaheadSerializationMixin:
     def serialize_course_run(self, course_run):
         result = SearchQuerySet().models(CourseRun).filter(key=course_run.key)[0]
         data = TypeaheadCourseRunSearchSerializer(result).data
-        # Items are grouped by content type so we don't need it in the response
-        data.pop('content_type')
         return data
 
     def serialize_program(self, program):
         result = SearchQuerySet().models(Program).filter(uuid=program.uuid)[0]
         data = TypeaheadProgramSearchSerializer(result).data
-        # Items are grouped by content type so we don't need it in the response
-        data.pop('content_type')
         return data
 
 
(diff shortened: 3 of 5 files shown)


bderusha (commented):
Code looks good, but I have some performance concerns about this type of double wildcard search.  Have we/can we test this?

MatthewPiatetsky:
@bderusha this endpoint isn't yet used by anything, but yeah I was going to do load testing on the load test environment
I was concerned about performance as well. I'm not sure if there is a more efficient approach to implementing this.

bderusha:
I wonder if there is a way to create an index of typeahead searches, so that rather than scanning everything front and back for a match we are doing exact match on partial words.

In theory we want a query like "one" to match `Done and done`, `The one big thing`, `Onerous Opinions`, `One + one`

In practice however we may be safe in assuming that users searching for "one" do not NEED `Done and done` as a result.

Therefore If we created an index of partial words e.g. "One", "Oner", "Onero", "Onerou", "Onerous"
we could match documents based on that without a performance penalty.

That said. Let's load test this.  We only have a few thousand documents.  This may be a WAY premature optimization.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -303,24 +303,31 @@ def test_empty_query(self):
...
+
+        url = '{path}?{qs}'.format(path=self.path, qs=qs)
+        return self.client.get(url)
 
     def test_typeahead(self):
         """ Test typeahead response. """
-        course_run = CourseRunFactory()
-        program = ProgramFactory()
-        response = self.get_typeahead_response()
+        course_run = CourseRunFactory(title="Python")
```
clintonb: Use a constant.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -303,24 +303,31 @@ def test_empty_query(self):
...
         self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
                                              'programs': [self.serialize_program(program)]})
 
     def test_typeahead_multiple_results(self):
         """ Test typeahead response with max number of course_runs and programs. """
-        CourseRunFactory.create_batch(RESULT_COUNT + 1)
-        ProgramFactory.create_batch(RESULT_COUNT + 1)
-        response = self.get_typeahead_response()
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="Test{}".format(i))
```
clintonb: Use a constant.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -329,10 +336,27 @@ def test_typeahead_multiple_results(self):
     def test_typeahead_multiple_authoring_organizations(self):
         """ Test typeahead response with multiple authoring organizations. """
         authoring_organizations = OrganizationFactory.create_batch(3)
-        course_run = CourseRunFactory(authoring_organizations=authoring_organizations)
-        program = ProgramFactory(authoring_organizations=authoring_organizations)
-        response = self.get_typeahead_response()
+        course_run = CourseRunFactory(title="Design", authoring_organizations=authoring_organizations)
```
clintonb: Use a constant.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -329,10 +336,27 @@ def test_typeahead_multiple_results(self):
...
+        program = ProgramFactory(title="Design", authoring_organizations=authoring_organizations)
+        response = self.get_typeahead_response("Design")
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
         self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
                                              'programs': [self.serialize_program(program)]})
+
+    def test_partial_term_search(self):
+        """ Test typeahead response with partial term search. """
+        course_run = CourseRunFactory(title="Learn Data Science")
```
clintonb: Use a constant.

course_discovery/apps/api/v1/views.py
```diff
@@ -680,13 +680,23 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
 
     serializer_class = serializers.TypeaheadSearchSerializer
     index_models = (CourseRun, Program,)
+    permission_classes = (IsAuthenticated,)
+
+    def get_queryset(self):
+        query = self.request.GET.get('q')
+        if not query:
+            raise ParseError("missing required q parameter")
+        query = 'title:*{}*'.format(query.lower())
```
clintonb: Why only title?

course_discovery/apps/api/v1/views.py
```diff
@@ -680,13 +680,23 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
 
 
-class TypeaheadSearchViewSet(BaseHaystackViewSet):
+class TypeaheadSearchViewSet(viewsets.ReadOnlyModelViewSet):
     """
     Typeahead for courses and programs.
     """
 
     serializer_class = serializers.TypeaheadSearchSerializer
     index_models = (CourseRun, Program,)
```
clintonb: This is not used.

course_discovery/apps/api/v1/views.py
```diff
@@ -680,13 +680,23 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+        query = self.request.GET.get('q')
+        if not query:
+            raise ParseError("missing required q parameter")
+        query = 'title:*{}*'.format(query.lower())
+        course_run_results = SearchQuerySet().models(CourseRun).raw_search(query)[:RESULT_COUNT]
+        program_results = SearchQuerySet().models(Program).raw_search(query)[:RESULT_COUNT]
+        return course_run_results + program_results
 
     def list(self, request, *args, **kwargs):
         response = super(TypeaheadSearchViewSet, self).list(request, *args, **kwargs)
```
clintonb: This is quite odd. You're doing too much work here. Just bypass the base class altogether:

```python
# serializers.py
class TypeaheadItemSerializer(serializers.Serializer):
  class Meta:
    # TODO Add other fields that will actually be displayed
    fields = ('title',)

class TypeaheadSearchSerializer(serializers.Serializer):
  course_runs = TypeaheadItemSerializer(many=True)
  programs = TypeaheadItemSerializer(many=True)

# views.py
class TypeaheadSearchView(APIView):
  def get_results(self, query):
    # TODO Do the search
    return course_runs, programs

  def get(self, request, format=None):
    course_runs, programs = self.get_results(request.GET['q'])
    serializer = TypeaheadSearchSerializer(data={'course_runs': course_runs, 'programs': programs)
    return Response(serializer.data, status=status.HTTP_200_OK)
```

clintonb (changes requested):

MatthewPiatetsky:
@bderusha 
I will probably need to add to the load tests since I don't think they're currently testing this, but I believe for peak load should we can say 120 rpm based on New Relic data and talking with Jasper (average is around 50)
Local response time for the search/typeahead endpoint seems to be around 150-250 ms and search/all is around 250-350 ms
Since nothing is using this endpoint yet I could probably run the load tests on the load test environment once this is merged.
@clintonb have I addressed your comments?

course_discovery/apps/api/v1/views.py
MatthewPiatetsky: Thanks for the help! 
I'm still new to DRF and I figured there was probably a simpler way to do what I was trying to do.

course_discovery/apps/api/serializers.py
```diff
@@ -990,15 +961,37 @@ class Meta:
...
 
 
-class TypeaheadSearchSerializer(HaystackSerializer):
+class TypeaheadCourseRunSearchSerializer(serializers.Serializer):
+    additional_details = serializers.SerializerMethodField()
+    title = serializers.CharField()
+    key = serializers.CharField()
+
+    def get_additional_details(self, result):
+        """ Value of the grey text next to the typeahead result title. """
```
clintonb: The API doesn't care about display values. This docstring should go.

course_discovery/apps/api/serializers.py
```diff
@@ -990,15 +961,37 @@ class Meta:
...
+        return result.org
```
clintonb: Why not expose this as an actual field named `org`?

course_discovery/apps/api/serializers.py
```diff
@@ -990,15 +961,37 @@ class Meta:
...
+        fields = ['key', 'title']
+
+
+class TypeaheadProgramSearchSerializer(serializers.Serializer):
+    additional_details = serializers.SerializerMethodField()
+    uuid = serializers.CharField()
+    title = serializers.CharField()
+    type = serializers.CharField()
+
+    def get_additional_details(self, result):
```
clintonb: Why not expose this as a field named `orgs`?

course_discovery/apps/api/serializers.py
```diff
@@ -990,15 +961,37 @@ class Meta:
...
+        """ Value of the grey text next to the typeahead result title. """
+        authoring_organizations = [json.loads(org) for org in result.authoring_organization_bodies]
+        return ', '.join([org['key'] for org in authoring_organizations])
```
clintonb: The API should not be formatting for display. Return the array and do the formatting client-side.

course_discovery/apps/api/v1/views.py
```diff
@@ -680,28 +681,22 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
-                course_runs.append(item)
-            elif item_type == 'program' and programs_length < RESULT_COUNT:
-                programs.append(item)
-            elif programs_length == RESULT_COUNT and course_run_length == RESULT_COUNT:
-                break
-        response.data = {'course_runs': course_runs, 'programs': programs}
-        return response
+    def get_results(self, query):
+        if not query:
+            raise ParseError("missing required q parameter")
```
clintonb: Do this validation in `get()`. `get_results` shouldn't care about query strings.

course_discovery/apps/api/v1/views.py
```diff
@@ -680,28 +681,22 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+        query = '*{}*'.format(query.lower())
+        course_runs = SearchQuerySet().models(CourseRun).raw_search(query)[:RESULT_COUNT]
```
clintonb: Move `RESULT_COUNT` to be a constant on this view. It's not really needed outside of this view and its tests.

clintonb (changes requested):

MatthewPiatetsky:
@clintonb @bderusha have I addressed your comments?

course_discovery/apps/api/serializers.py
```diff
@@ -990,15 +961,36 @@ class Meta:
         }
 
 
-class TypeaheadSearchSerializer(HaystackSerializer):
+class TypeaheadCourseRunSearchSerializer(serializers.Serializer):
+    orgs = serializers.SerializerMethodField()
```
clintonb: `orgs` implies multiple orgs. This should be `org`, and you can drop the method. For future reference, you can alias fields by using the [`source` parameter](http://www.django-rest-framework.org/api-guide/fields/#source).

course_discovery/apps/api/serializers.py
```diff
@@ -990,15 +961,36 @@ class Meta:
...
+
+
+class TypeaheadProgramSearchSerializer(serializers.Serializer):
+    orgs = serializers.SerializerMethodField()
+    uuid = serializers.CharField()
+    title = serializers.CharField()
+    type = serializers.CharField()
+
+    def get_orgs(self, result):
+        """ Value of the grey text next to the typeahead result title. """
```
clintonb: This docstring is irrelevant.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1128,25 +1127,22 @@ def _create_expected_data(self, program):
             'uuid': str(program.uuid),
             'title': program.title,
             'type': program.type.name,
-            'content_type': 'program',
-            'additional_details': program.authoring_organizations.first().key
+            'orgs': [program.authoring_organizations.first().key]
```
clintonb: For completeness this should be the keys of all of the organizations: `list(program.authoring_organizations.all().values_list('key', flat=True))`.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -10,14 +10,15 @@
 
 from course_discovery.apps.api.serializers import (CourseRunSearchSerializer, ProgramSearchSerializer,
                                                    TypeaheadCourseRunSearchSerializer, TypeaheadProgramSearchSerializer)
-from course_discovery.apps.api.v1.views import RESULT_COUNT
 
 from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD, PartnerFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import CourseRun, Program
 from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, ProgramFactory, OrganizationFactory
```
clintonb, line 19: This newline shouldn't be here. All of the imports should be alphabetized.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -300,39 +297,68 @@ def test_empty_query(self):
...
                                              'programs': [self.serialize_program(program)]})
 
     def test_typeahead_multiple_results(self):
         """ Test typeahead response with max number of course_runs and programs. """
-        CourseRunFactory.create_batch(RESULT_COUNT + 1)
-        ProgramFactory.create_batch(RESULT_COUNT + 1)
-        response = self.get_typeahead_response()
+        RESULT_COUNT = TypeaheadSearchView.RESULT_COUNT
+        title = "Test"
+        for i in range(RESULT_COUNT):
```
clintonb: Create more items to ensure your limits are actually working: `range(RESULT_COUNT) + 1`.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -300,39 +297,68 @@ def test_empty_query(self):
...
+        course_run = CourseRunFactory(title=title)
+        program = ProgramFactory(title=title)
+        response = self.get_typeahead_response(title)
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
         self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
                                              'programs': [self.serialize_program(program)]})
 
     def test_typeahead_multiple_results(self):
         """ Test typeahead response with max number of course_runs and programs. """
```
clintonb: Verify the typeahead responses always returns a limited number of results, even if there are more hits.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -300,39 +297,68 @@ def test_empty_query(self):
...
+    def test_partial_term_search(self):
+        """ Test typeahead response with partial term search. """
+        title = "Learn Data Science"
+        course_run = CourseRunFactory(title=title)
+        program = ProgramFactory(title=title)
+        query = "Data Sci"
+        response = self.get_typeahead_response(query)
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+        self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
```
clintonb: Nit: Move the dictionary to a variable.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -300,39 +297,68 @@ def test_empty_query(self):
...
+                                             'programs': [self.serialize_program(program)]})
+
+    def test_exception(self):
+        """ Test typeahead response with missing q parameter. """
```
clintonb: Verify the view raises an error if the 'q' query string parameter is not provided.

course_discovery/apps/api/v1/views.py
```diff
@@ -680,28 +679,24 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+    def get_results(self, query):
+        query = '*{}*'.format(query.lower())
+        course_runs = SearchQuerySet().models(CourseRun).raw_search(query)[:self.RESULT_COUNT]
+        programs = SearchQuerySet().models(Program).raw_search(query)[:self.RESULT_COUNT]
+        return course_runs, programs
+
+    def get(self, request, *args, **kwargs):
+        query = request.query_params.get('q')
+        if not query:
+            raise ParseError("missing required q parameter")
```
clintonb: The 'q' querystring parameter is required for searching.

clintonb (approved):
Please address the comments before merging.

course_discovery/apps/api/serializers.py
MatthewPiatetsky: Currently implementation of the indexed field assumes that there is one org, but we know we will have course runs with multiple orgs at some point in the future. This was the reason I put orgs, but I suppose if we don't fully support multiple orgs I can put org for now and we can change it to orgs in the future

clintonb: That would be best. True multi-organization courses are a long ways out. I haven't even heard them discussed.

bderusha (approved):

PR #460, opened 2016-11-29
Awais786/ecom 6142

awais786:
Courses are now associate with the organization. While creating the organization user can choose the org from the dropdown and it will be added in course table field ( organizations ). Each course will be associated with one org for the time being.  Each Org will be joined with one group only through GroupOrganization table.

I have update the existing code and replace the institution as organization.


awais786:
@bderusha PR is in progress. Just take an overview and let me know if things looks fine.

awais786:
Instead of loading the groups in form. I am loading the GroupOrganization table there.

course_discovery/apps/publisher/views.py
```diff
@@ -142,9 +142,11 @@ def post(self, request, *args, **kwargs):
                     seat.changed_by = self.request.user
                     seat.save()
 
-                    institution = get_object_or_404(Group, pk=course_form.data['institution'])
+                    organization = get_object_or_404(GroupOrganization, pk=course_form.data['organization'])
```
bderusha: prefer `group_organization` as a variable name

course_discovery/apps/publisher/views.py
```diff
@@ -142,9 +142,11 @@ def post(self, request, *args, **kwargs):
...
+                    course.organizations.add(organization.organization)
+
                     # assign guardian permission.
-                    course.assign_permission_by_group(institution)
+                    # course.assign_permission_by_group(institution)
```
bderusha: If this is not needed we should remove it

bderusha (commented):
Looks good.  A few minor comments

awais786:
Due to the organization base permissions I have to fix all the test cases also some mixins.  But we have to update them also in future stories.

awais786:
@bderusha I don't think we need to add any permissions in this PR. 
I can close this sub-task without any implementation https://openedx.atlassian.net/browse/ECOM-6412?

awais786:
@bderusha please review this PR again.

awais786:
@waheedahmed @tasawernawaz @ahsan-ul-haq Please review this.

awais786:

course_discovery/apps/publisher/mixins.py
```diff
@@ -57,4 +57,4 @@ def form_valid(self, form):
 
 
 def check_view_permission(user, course):
-    return user.is_staff or user.has_perm(Course.VIEW_PERMISSION, course)
+    return user.is_staff or (course and course.get_group_from_organizations() in user.groups.all())
```
bderusha: This is no longer checking the view permission.  We should be leveraging django's permision checking syntax as much as possible.

```
return user.is_staff or any([user.has_perm(Course.VIEW_PERMISSION, organization) for organization in course.organizations])
```

course_discovery/apps/publisher/models.py
```diff
@@ -181,6 +172,11 @@ def keywords_data(self):
 
         return None
 
+    def get_group_from_organizations(self):
+        """ Returns the first group associated with organizations."""
```
bderusha: This docstring is confusing to me.  Do we allow multiple groups to be associated with an organization?

course_discovery/apps/publisher/models.py
```diff
@@ -151,24 +151,15 @@ class Meta(TimeStampedModel.Meta):
...
-        available_groups = get_groups_with_perms(self)
-        return available_groups[0] if available_groups else None
-
     def get_group_users_emails(self):
         """ Returns the list of users emails with enable email notifications
         against a course group. By default if attribute value does not exists
         then user will be eligible for emails.
         """
-        users_list = get_users_with_perms(self)
+        group = self.get_group_from_organizations()
```
bderusha: Again, we should be getting users with permissions on the org objects directly, not just checking for groups.

course_discovery/apps/publisher/models.py
```diff
@@ -181,6 +172,11 @@ def keywords_data(self):
 
         return None
 
+    def get_group_from_organizations(self):
+        """ Returns the first group associated with organizations."""
+        groups = [grp_org.publisher_organizations.group for grp_org in self.organizations.all()]
```
bderusha: verbose variables please. `group_organization`

course_discovery/apps/publisher/tests/test_emails.py
```diff
@@ -32,17 +31,19 @@ def setUpClass(cls):
 
         cls.site = Site.objects.get(pk=settings.SITE_ID)
 
-        cls.group = factories.GroupFactory()
-        cls.user.groups.add(cls.group)
-        cls.user_2.groups.add(cls.group)
-        cls.user_3.groups.add(cls.group)
+        cls.group_organization_1 = factories.GroupOrganizationFactory()
```
bderusha: if there is only one `group_organization` there is no need to number it.  Or we need to number `user` as `user_1`.  Either way is fine by me as long as we are consistent.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -82,43 +85,48 @@ def test_post_back_url(self):
             reverse('publisher:publisher_courses_edit', kwargs={'pk': self.course.id})
         )
 
-    def test_assign_permission_by_group(self):
-        """ Verify that permission can be assigned using the group. """
+    def test_assign_organization_to_course(self):
+        """ Verify that only group associated users can view the course. """
```
bderusha: We will need to update this test to `""" Verify that only users with permissions on the org can view the course. """`

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -280,7 +282,7 @@ def _assert_test_data(self, response, course, expected_type, expected_price):
             status_code=302,
             target_status_code=200
         )
-        self.assertEqual(course.group_institution, self.group)
+        # self.assertEqual(course.group_institution, self.group_organization_1)
```
bderusha: Delete if not needed

bderusha (changes requested):

course_discovery/apps/publisher/models.py
```diff
@@ -151,24 +151,15 @@ class Meta(TimeStampedModel.Meta):
             ('view_course', 'Can view course'),
         )
 
-    def assign_permission_by_group(self, institution):
```
awais786, line 154: @bderusha so we should assign the permission here. I had remove this

course_discovery/apps/publisher/mixins.py
awais786: It mean we have to assign permissions on org object ?

awais786: Yesterday we decided we will not assign permissions thats close the sub-task also.

awais786:

awais786:

course_discovery/apps/publisher/forms.py
```diff
@@ -49,7 +48,7 @@ class Meta:
 class CustomCourseForm(CourseForm):
     """ Course Form. """
 
-    institution = forms.ModelChoiceField(queryset=Group.objects.all(), required=True)
+    organization = forms.ModelChoiceField(queryset=GroupOrganization.objects.all(), required=True)
```
bderusha: Why are we doing it this way?  Why not just attach the actual org

course_discovery/apps/publisher/models.py
```diff
@@ -151,17 +152,18 @@ class Meta(TimeStampedModel.Meta):
...
+    def assign_permission_by_group(self, group):
```
bderusha, line 155: I still don't understand why we need to assign permissions in the code here.  Please explain?

course_discovery/apps/publisher/models.py
```diff
@@ -151,17 +152,18 @@ class Meta(TimeStampedModel.Meta):
...
         """ Assigns permission on the course against the group. """
-        assign_perm(self.VIEW_PERMISSION, institution, self)
+        assign_perm(self.VIEW_PERMISSION, group, self)
 
     @property
-    def group_institution(self):
-        """ Returns the group object having permissions on the given course.
-        Course will be associated with one group only.
+    def group_organization(self):
```
bderusha: The course should not need access to this and if it does (please provide reasoning) it needs to return ALL of them not just the first one.

course_discovery/apps/publisher/models.py
```diff
@@ -407,3 +409,14 @@ def __str__(self):
             user=self.user,
             role=self.role
         )
+
+
+class GroupOrganization(TimeStampedModel):
```
bderusha: Given that we are going to also be attaching permissions to this I think it would be more extensible if we called this `OrganizationExtension` or something that doesn't pin it to just being a join table.  I know this is a big change and I'm sorry for not bringing it up before. I am only now realizing this.  But better to catch it now than after we run the migrations.

If we could have we would have just added a group directly to `Organization` but we don't want to do that so we must extend that object with a publisher specific one. 

If we could have we would have just added the permissions directly to the Meta of the `Organization` but we don't want to do that so we must find a publisher specific place for those as well. 

Imagine tomorrow we realize that Publisher needs to associate a default seat type with an `Organization`.  If we call it `OrganizationExtension` or `PublisherOrganization` or something like that it makes sense why `default_seat_type` should live there.  If we continue to call it `GroupOrganization` it feels very strange attaching all of this stuff to it.  It's just for groups.

The functionality ends up being the same, but the intention is more clear.

course_discovery/apps/publisher/models.py
```diff
@@ -407,3 +409,14 @@ def __str__(self):
...
+    """ Group-Organization relation model. """
+    organization = models.OneToOneField(Organization, related_name='publisher_organizations')
```
bderusha: `related_name='group_organization'`  Singular.  Or `organization_extension`/`publisher_organization` if we update the name

course_discovery/apps/publisher/models.py
```diff
@@ -407,3 +409,14 @@ def __str__(self):
...
+    group = models.OneToOneField(Group, related_name='publisher_groups')
```
bderusha: Same as above

bderusha (changes requested):
Still confused about some of the permission settings.  Also I think we need to updated the name of the class.  Sorry for springing it on you kind of late in the PR process.

course_discovery/apps/publisher/forms.py
awais786: It will load only those organizations which are associated with any group. If I load from organizations table and user choose any of that organization which has no record in GroupOrganization it will create problem.

course_discovery/apps/publisher/models.py
awais786: Actually we don't need to assign permission now. I thought we will remove this code in any other story. I just rename the variable. But i think it's better to remove.

awais786: I think its better to remove the permissions code in my other PR.
Because there I am fixing how to pick the courses based on roles.  
[PR 479](https://github.com/edx/course-discovery/blob/71a2dd4f1c4f8816d1718e6cb4184a8f16d1d178/course_discovery/apps/publisher/views.py#L43)

course_discovery/apps/publisher/models.py
awais786: We need to show the associated organization with course on dashboard. We have a m2m field as organization in the table. but in publisher app only one org can be join with one course.

awais786: If return all orgs from m2m. Then do we need to show all orgs on dashboard for a course ?

course_discovery/apps/publisher/forms.py
awais786: queryset=Organization.objects.filter(id__in=OrganizationExtension.objects.all())

course_discovery/apps/publisher/models.py
awais786: I have renamed the table as per your suggestion.

course_discovery/apps/publisher/models.py
awais786: This will be remove in another PR where we are implementing the courses by roles.

course_discovery/apps/publisher/models.py
awais786: But we have to assign the permission from OrganizationExtension table. So I think it make sense to create a ticket for that purpose. Where we can remove the existing code of assignment and fix all the test cases.

awais786:
@bderusha I have fixed all of your requested changes. For permissions we have to remove the assignment of permissions from course. My suggestion is to add a separate ticket to remove this or we can implement this when implement the OrganiztionExtension permission assignment from the admin.

course_discovery/apps/publisher/models.py
```diff
@@ -407,3 +409,14 @@ def __str__(self):
             user=self.user,
             role=self.role
         )
+
+
+class OrganizationExtension(TimeStampedModel):
+    """ Organization-Extension relation model. """
+    organization = models.OneToOneField(Organization, related_name='publisher_organization')
```
bderusha: the `related_name` is `organization_extension`

course_discovery/apps/publisher/models.py
```diff
@@ -407,3 +409,14 @@ def __str__(self):
...
+    group = models.OneToOneField(Group, related_name='publisher_group')
```
bderusha: the `related_name` is `organization_extension`

course_discovery/apps/publisher/models.py
bderusha: group organization isn't a thing anymore.  This method should be renamed

course_discovery/apps/publisher/views.py
```diff
@@ -142,9 +141,10 @@ def post(self, request, *args, **kwargs):
                     seat.changed_by = self.request.user
                     seat.save()
 
-                    institution = get_object_or_404(Group, pk=course_form.data['institution'])
+                    group_organization = get_object_or_404(OrganizationExtension, pk=course_form.data['organization'])
```
bderusha: update variable name

course_discovery/templates/publisher/dashboard/_in_progress.html
```diff
@@ -31,8 +31,8 @@
                     <td>
                         <a href="{% url 'publisher:publisher_course_run_detail' course_run.id %}">{{ course_run.title }}</a>
                     </td>
-                    <td>{% if course_run.course.group_institution %}
-                            {{ course_run.course.group_institution }}
+                    <td>{% if course_run.course.group_organization %}
```
bderusha: we should be accessing the course's organizations directly.  There is no need to go through this middle-man

bderusha: Here and all other templates

course_discovery/apps/publisher/views.py
bderusha: here and throughout

course_discovery/apps/publisher/models.py
bderusha: I think that in the templates we should just get the first org for now.  However that is entirely a scope issue.  We know that we want to support multiple orgs in the future.  So the few places we have that make assumptions about there being only one org, the better

bderusha (changes requested):
Please clean up all of the references to group_organizations since that is no longer a thing.

awais786:
@bderusha I have made all changes.

course_discovery/apps/publisher/models.py
```diff
@@ -151,17 +152,18 @@ class Meta(TimeStampedModel.Meta):
...
+    def assign_permission_by_group(self, group):
         """ Assigns permission on the course against the group. """
-        assign_perm(self.VIEW_PERMISSION, institution, self)
+        assign_perm(self.VIEW_PERMISSION, group, self)
 
     @property
-    def group_institution(self):
-        """ Returns the group object having permissions on the given course.
-        Course will be associated with one group only.
+    def group_extension(self):
```
bderusha: I still don't understand why this is necessary and why we have chosen this name.  This method has nothing to do with groups

course_discovery/apps/publisher/models.py
```diff
@@ -151,17 +152,18 @@ class Meta(TimeStampedModel.Meta):
...
+        """ Returns the group-extension object associated with the course via organization.
+        It should come from the group-organization table not from M2M field.
```
bderusha: Why not?

course_discovery/apps/publisher/views.py
```diff
@@ -142,9 +141,10 @@ def post(self, request, *args, **kwargs):
                     seat.changed_by = self.request.user
                     seat.save()
 
-                    institution = get_object_or_404(Group, pk=course_form.data['institution'])
+                    group_extension = get_object_or_404(OrganizationExtension, pk=course_form.data['organization'])
```
bderusha: What is a `group_extension`?  Am I missing something?

course_discovery/apps/publisher/models.py
bderusha: Also this method is not used at all in the application outside of tests.  Let's just delete it.

bderusha (changes requested):

course_discovery/apps/publisher/views.py
awais786: This should be organization_extension. I mixed up the old and new variable name thats why it is causing confusion.

awais786:
Update the variable name and also remove the method.

bderusha (approved):
👍 once tests pass

waheedahmed (changes requested):
You need to recreate your migration, two migrations with same number.

waheedahmed (approved):

PR #462, opened 2016-11-30
Override custom method to replace img variations

tasawernawaz:
[ECOM-6394](https://openedx.atlassian.net/browse/ECOM-6394)

@schenedx Please review this.

### DIFF
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -1,3 +1,4 @@
+from stdimage.models import StdImageFieldFile
 from stdimage.utils import UploadTo
 
 RESERVED_ELASTICSEARCH_QUERY_OPERATORS = ('AND', 'OR', 'NOT', 'TO',)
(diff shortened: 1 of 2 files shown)


schenedx (approved):
👍

schenedx:
@clintonb I think you should review this PR as well. 
@tasawernawaz Is it right that we cannot really reproduce this on local machine because we use local filesystem? Can you explain to @clintonb why the `replace=True` do not need to happen?

tasawernawaz:
@schenedx I added a comment on the ticket as well that we can't reproduce it locally which is due to file system storage create unique name every time before save. Reference: https://docs.djangoproject.com/en/1.10/ref/files/storage/#django.core.files.storage.Storage.get_available_name

course_discovery/apps/course_metadata/utils.py
```diff
@@ -42,3 +43,23 @@ def __call__(self, instance, filename):
...
+    Args:
+        file_name (str): name of the image file.
+        variations (dict): dict containing variations of image
+        storage (storage class): e.g DefaultStorage()
+
+    Returns:
+        False (bool): to prevent its default behaviour
+    """
+
+    for _, variation in variations.items():
```
clintonb: `for variation in variations.values():`

course_discovery/apps/course_metadata/utils.py
```diff
@@ -42,3 +43,23 @@ def __call__(self, instance, filename):
...
+
+
+def custom_render_variations(file_name, variations, storage, replace=True):
+    """ Utility method used to override default behaviour of StdImageFieldFile by
+    passing it replace=True.
+
+    Args:
+        file_name (str): name of the image file.
+        variations (dict): dict containing variations of image
+        storage (storage class): e.g DefaultStorage()
```
clintonb: `storage (Storage): Storage class responsible for storing the image.`

clintonb (approved):
It would be great if we could find a way to unit test this. No worries if that is not immediately feasible.

schenedx:
@clintonb @tasawernawaz and I talked about unit tests but because of the filesystem library, we couldn't really run the test unless the test can run a different django library for storage instead. Don't know if we want to do the research at the moment for this bug. Considering we want to get this out this week to unblock Tatiana, I say we let this go without unit test covering this strange case.

PR #464, opened 2016-11-30
ECOM-4738 Expose marketing url for typeahead

MatthewPiatetsky:
@edx/ecommerce 

I was working on the front end portion and realized that it generates urls based on drupal nodes, so we need the marketing_url instead

or do we want to use the slug instead?


clintonb (approved):

MatthewPiatetsky:
Need to update one more thing

PR #467, opened 2016-12-01
Using django-choices for seat type choices

clintonb:


course_discovery/apps/course_metadata/choices.py
```diff
@@ -20,3 +20,11 @@ class ProgramStatus(DjangoChoices):
...
     Retired = ChoiceItem('retired', _('Retired'))
     Deleted = ChoiceItem('deleted', _('Deleted'))
+
+
+class SeatType(DjangoChoices):
+    Audit = ChoiceItem('audit', _('Audit'))
+    Credit = ChoiceItem('credit', _('Credit'))
+    Honor = ChoiceItem('honor', _('Honor'))
+    Professional = ChoiceItem('professional', _('Professional'))
+    Verified = ChoiceItem('verified', _('Verified'))
```
davec-edx, line 30: Curious why these start with an upper case letter?

clintonb: I...don't know. That's the convention we started using and kept using. It's simply become habit at this point.

PR #469, opened 2016-12-01
Added endpoints to read organizations

clintonb:
ECOM-6501


clintonb:
@edx/ecommerce please review

clintonb:
@edx/ecommerce please review

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -0,0 +1,105 @@
...
+        organizations[1].tags.add(tag)
+        self.assert_list_tag_filter(Organization.objects.all(), [tag, tag2])
+
+    def test_retrieve(self):
+        """ Verify the endpoint returns details for a single organization. """
+        organization = OrganizationFactory()
+        url = reverse('api:v1:organization-detail', kwargs={'uuid': organization.uuid})
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
```
davec-edx, line 108: Do these cover both positive and negative? I see checking for 200s for example. Is it possible to get anything else?

clintonb: If you passed an invalid UUID, you'd get a 404. Other than that the only other codes would be 401/403 for auth (which is tested) and 500 for unexpected errors.

I can add a test for the 404.

schenedx (approved):
👍

course_discovery/apps/api/v1/views/organizations.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework import viewsets
+from rest_framework.filters import DjangoFilterBackend
+from rest_framework.permissions import IsAuthenticated
+
+from course_discovery.apps.api import filters, serializers
+
+
+# pylint: disable=no-member
+class OrganizationViewSet(viewsets.ReadOnlyModelViewSet):
```
rlucioni, line 9: @clintonb what does this mean for the "organizations API" hosted by the LMS? Is it deprecated in favor of this viewset?

clintonb: Long term, it probably should be. In the near term, we will have two APIs; although, the API hosted by the Catalog Service will include info from Drupal.

PR #470, opened 2016-12-02
Using raw ID field for user field on OrganizationUserRoleInline

clintonb:


clintonb:
FYI @edx/ecommerce please be cognizant of your admin changes. Many-to-many admin fields will load *all* of the possible options. In general, at our scale, we should always use [`raw_id_fields`](https://docs.djangoproject.com/en/1.10/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields) for `User` associations or any other model that may be linked back to a user. Given that we have single sign-on across all of our IDAs, pretty much any user in LMS can end up with an entry in any of the IDAs' user tables. That's on the order of millions.

rlucioni (commented):
👍

PR #471, opened 2016-12-02
ECOM-4738 Verify typeahead result boosting and only apply partial search on title and course key

MatthewPiatetsky:
@edx/ecommerce 

After discussion with Jasper we decided to try to only search against title and course number since searching against the whole document was not giving us the results we wanted (partial searches were resulting in too many false positives that we were not interested in)
Also, I realized we had no testing of boosting, so I added unit tests of boosting

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -14,8 +14,9 @@ from course_discovery.apps.api.v1.views import TypeaheadSearchView
 from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWORD, PartnerFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
-from course_discovery.apps.course_metadata.models import CourseRun, Program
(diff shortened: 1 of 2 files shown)


clintonb (commented):
Your commit message should explain what the commit does (not mention "new ACs") and include the ticket number.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -305,6 +320,8 @@ def get_typeahead_response(self, query=None):
             qs = urllib.parse.urlencode({'q': query})
 
         url = '{path}?{qs}'.format(path=self.path, qs=qs)
+        config = ElasticsearchBoostConfig(function_score=self.function_score)
```
clintonb: You will want to call `ElasticsearchBoostConfig.get_solo()` to instantiate the model. See https://github.com/lazybird/django-solo.

bderusha (approved):

mikedikan:
LGTM, but i'd like to hear @bderusha weigh in on whether or not we are already testing boosting somewhere else.  I'd hate to have that duplicated.

MatthewPiatetsky:
@clintonb further comments?

clintonb (approved):

PR #474, opened 2016-12-06
Replaced RequireJS with Webpack

clintonb:
This set of changes removes RequireJS, and replaces it with Webpack. This results in two major benefits:

1. We no longer need `bower` (once the Publisher Tool is updated), and can rely solely on `npm` for front-end libraries.
2. Webpack is aware of the `node_modules` directory which lives outside of the static files directory, as opposed to `django-compressor` relying on the nesting of the `bower_components` in the static files directory. When we run the `collectstatic`management command we no longer need to waste time copying data that will never be accessed. Webpack's output files are consolidated, meaning we get only what we need.

Unfortunately, there are tradeoffs. Webpack is not as integrated with Django as `django-compressor`. Webpack is run in a completely separate process, hence the new `static.watch` Make target. This means that local development will require running a separate process to watch and compile CSS and JS. Re-compilation takes about 2-2.5 seconds. Given that `django-compressor` hid its compilation time from us by stalling the request, it is likely we are already experiencing this compilation time but have not been consciously aware of it.

Please share your thoughts.

@edx/ecommerce @andy-armstrong @dsjen @bjacobel @cpennington


course_discovery/webpack.config.js
```diff
@@ -0,0 +1,75 @@
...
+            name: 'common',
+            children: true
+        })
+    ],
+
+    module: {
+        loaders: [
+            {
+                test: /\.s?css$/,
+                loader: ExtractTextPlugin.extract('style-loader', 'css-loader?minimize!sass')
```
bjacobel: Do you want css-loader to minimize output even in development mode?

Also along these lines - it would be easy to enable Sass sourcemaps here. I find them quite helpful when debugging UI issues.

bjacobel (commented):
This is awesome. I noticed one opportunity to make styles in development even more developer-friendly.

We'll likely lean on this PR when/if we do the same thing to Platform.

course_discovery/webpack.config.js
clintonb: 1. Hmm...I intentionally avoided a separate configuration so that developers actually run what is run in production. For example, I removed the [PurifyCSS plugin](https://github.com/purifycss/purifycss-webpack-plugin) because it was removing the styles for DataTables (since the classes, etc. are set at page load time). If I had only enabled that plugin for production, I would not have known about the issue until the code was merged and deployed to stage. Although there is an extra cost to minify in development, I think it is lower than the cost of rework.

2. Okay. I'll look into source maps. I've heard of them, but never worked on anything complex enough to need them.

rlucioni: +1 to keeping development config the same as production config.

Makefile
```diff
@@ -34,12 +36,18 @@ help:
 	@echo ""
 
 static:
-	$(NODE_BIN)/r.js -o build.js
+	cd course_discovery && $(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress -p
 	python manage.py collectstatic --noinput
 	python manage.py compress -v3 --force
 
+static.dev:
+	cd course_discovery && $(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress
```
rlucioni: Can `webpack.config.js` live at the root of the repo, so this cd'ing isn't necessary? We tend to put config files like it at the root.

Makefile
```diff
@@ -34,12 +36,18 @@ help:
 	@echo ""
 
 static:
-	$(NODE_BIN)/r.js -o build.js
+	cd course_discovery && $(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress -p
 	python manage.py collectstatic --noinput
 	python manage.py compress -v3 --force
 
+static.dev:
```
rlucioni: A better name might be `static.build`, or something that implies this runs once and exits.

Makefile
```diff
@@ -34,12 +36,18 @@ help:
...
+	cd course_discovery && $(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress
+
+static.watch:
+	cd course_discovery && $(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress --watch
```
rlucioni: Does this run in the foreground? Put differently, would I need to open a separate tab to run this as-is if I also want to run the server?

Makefile
```diff
@@ -25,6 +25,8 @@ help:
 	@echo "  test                                 run tests and generate coverage report"
 	@echo "  validate                             run tests and quality checks"
 	@echo "  static                               gather all static assets for production"
+	@echo "  static.dev                           build unoptimized CSS and JS"
+	@echo "  static.watch                         build unoptimized CSS and JS whenever an asset is modified"
```
rlucioni: We should consider [invoke](http://www.pyinvoke.org/) for these kinds of tasks. We could replace these three targets with repeated code with a single "build" task that takes arguments (e.g., `--watch`, `--unoptimized`).

Makefile
```diff
@@ -34,12 +36,18 @@ help:
 	@echo ""
 
 static:
-	$(NODE_BIN)/r.js -o build.js
+	cd course_discovery && $(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress -p
```
rlucioni: Is there a less arcane alternative to `-p`? Looks like it's interchangeable with `--optimize-minimize`.

bjacobel: I believe `-p` also sets `NODE_ENV=production` in addition to setting `--optimize-minimize`, which may or may not be desired.

Makefile
bjacobel: IIRC, you don't need to specify `--config` if you're using the standard config file name and location.

package.json
```diff
@@ -6,12 +6,26 @@
...
+    "style-loader": "^0.13.1",
+    "url-loader": "^0.5.7",
+    "webpack": "^1.13.3",
+    "webpack-bundle-tracker": "0.0.93"
   },
   "devDependencies": {
     "gulp": "^3.9.1",
-    "gulp-jscs": "4.0.0",
-    "geckodriver": "1.2.0"
+    "gulp-jscs": "^4.0.0",
```
rlucioni: Are you sure about adding the carets? I'm not a fan of allowing our dependencies to change underneath us, potentially during deployment. This has bitten us in the past (e.g., datatables). I'd argue for removing all the carets.

package.json
```diff
@@ -6,12 +6,26 @@
...
+    "edx-pattern-library": "^0.18.0",
+    "extract-text-webpack-plugin": "^1.0.1",
+    "file-loader": "^0.9.0",
+    "imports-loader": "^0.6.5",
+    "jquery": "^3.1.1",
+    "node-sass": "^3.13.0",
+    "sass-loader": "^4.0.2",
+    "style-loader": "^0.13.1",
+    "url-loader": "^0.5.7",
+    "webpack": "^1.13.3",
```
bjacobel: FYI, Webpack 2 is due for an RC any week now. 2 is currently on a 2.1.0-beta.27 version or something, it's very stable and the breaking changes are minor. Might be worth going directly to 2, if the Django library is compatible.

clintonb: Great! I'll update once it's released.

bjacobel: 1.14 was also released yesterday.

course_discovery/webpack.config.js
andy-armstrong: Just to add my perspective: we explicitly skip optimization in development mode on edx-platform for a few reasons:
 - we do aggressive uglification which is a significant portion of already slow asset generation times
 - we support hot reloading which isn't very hot if it takes multiple minutes
 - we want to be able to use the browser's debugger on the source files which is much harder with minification (although source maps improve matters)

It may be that none of these are issues for you guys because your repo is smaller, but I just wanted to lay out our rationale. It is also possible that we could be smarter about incremental rebuilding of assets because right now we usually rebuild everything. 

Note that there is a potential downside that the minification might cause behavioral differences, but I've never seen that happen. I think the payoff of faster developer turnaround and improved debugging is well worth it.

Makefile
clintonb: Updated.

Makefile
clintonb: Yes, this will need to be run in its own terminal tab/window.

rlucioni (approved):

clintonb:
This is blocked until we upgrade to NodeJS 4.7+ on the Docker images. See https://github.com/webpack/webpack/issues/3930.

clintonb:
PR for Node.js upgrade: https://github.com/edx/configuration/pull/3610. I have built the image locally, and pushed it to Docker Hub. Rebuilding...

bjacobel:
FYI, Webpack 2.x was just released: https://medium.com/webpack/webpack-2-2-the-final-release-76c3d43bf144#.f6zb8cfqm

clintonb:
#notetoself: https://webpack.js.org/guides/migrating/

clintonb:
FYI @edx/learner I intend to merge this today. These changes will only affect the query preview page (at /), but lay the foundation for other changes (e.g. updating Publisher).

muhammad-ammar:
@clintonb It looks like that after this PR, `Publisher` styling is messed up. Can you look into this?
 [Stage Publisher](https://stage-edx-discovery.edx.org/publisher/) 
[Prod Publisher](https://prod-edx-discovery.edx.org/publisher/)

clintonb:
@muhammad-ammar, please define "messed up". I see no issues.

muhammad-ammar:
@clintonb Please see the below screenshots. This is how the publisher looks like before.
publisher_courses_list_page
publisher_dashboard_page

clintonb:
@muhammad-ammar, please highlight the _exact_ problems you are seeing along with any console errors (aside from the missing favicon file). I see no errors in the console and no difference in the styles.

muhammad-ammar:
@clintonb There are no errors on console. These are the current screens shots. For both pages header and footer are different before and after. Placement of search box is different. Column sort indicators not shown. 

publisher_courses_list_now

Dashboard page table row height is different. Even/Odd coloring missing. 
publisher_dashboard_now

PR #476, opened 2016-12-06
Added Branding Logo Image to the ProgramType

afzaledx:


course_discovery/apps/course_metadata/migrations/0039_merge.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
clintonb: This is unnecessary. If there is already a migration numbered `0038`, you need to regenerate your migration.

clintonb (changes requested):
WL-775 calls for having a logo *per program type*. You have added the logo to the `Program` model which is incorrect. You should be modifying the `ProgramType` model. Additionally, we should chat about your use cases to understand how this will be exposed via API.

afzaledx:
@clintonb, I have made the changes. Can you please review?

course_discovery/apps/course_metadata/models.py
```diff
@@ -580,6 +580,17 @@ class ProgramType(TimeStampedModel):
                               'associated courses, but enrolled in other seat types, will NOT have their completion '
                               'of the course counted toward the completion of the program.'),
     )
+    logo_image = StdImageField(
+        upload_to=UploadToFieldNamePath(populate_from='name', path='media/program_types/logo_images'),
+        blank=True,
+        null=True,
+        variations={
+            'large': (256, 256),
```
afzaledx, line 589: These dimensions might be changed in the future since we're not sure at this stage what the exact sizes are going to be.

afzaledx:
@ibrahimahmed443 @hasnain-naveed @douglashall 

The branch where the logo image has been added to the Programs dictionary
The logo image can be added to a `ProgramType` from the Django admin. The logo will not be visible when modifying a program, but the image will be returned whenever all programs, or a specific program (with a uuid) is requested.

course_discovery/apps/course_metadata/models.py
```diff
@@ -580,6 +580,17 @@ class ProgramType(TimeStampedModel):
                               'associated courses, but enrolled in other seat types, will NOT have their completion '
                               'of the course counted toward the completion of the program.'),
     )
+    logo_image = StdImageField(
+        upload_to=UploadToFieldNamePath(populate_from='name', path='media/program_types/logo_images'),
```
douglashall: Can `UploadToFieldNamePath` handle the case when the `name` field has spaces/non-alphanumeric characters?

afzaledx:
Probably not. I will fix it

course_discovery/apps/api/serializers.py
```diff
@@ -534,6 +534,7 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     banner_image = StdImageSerializerField()
     courses = serializers.SerializerMethodField()
     type = serializers.SlugRelatedField(slug_field='name', queryset=ProgramType.objects.all())
+    logo_image = StdImageSerializerField(source='type.logo_image')
```
clintonb: We have done a lot of work to tune these API endpoints. If you need to expose information about program types, that should be done via a new API endpoint (/program_types). Also, you need tests.

course_discovery/apps/course_metadata/models.py
clintonb: Nope. Use ID.

clintonb (changes requested):
Create a new API endpoint, and test it.

course_discovery/apps/api/serializers.py
douglashall: Will this be an absolute URL to the program type logo image file?

afzaledx: works like the "banner_image" dictionary from here: https://stage-edx-discovery.edx.org/api/v1/programs/0df34738-be78-4a08-8382-566cf4682691/

Running locally, I am getting something like this: http://127.0.0.1:18381/media/media/program_types/logo_images/professional_wIQA9lv.medium.png

and the banner_image is like: 
http://127.0.0.1:18381/media/media/programs/banner_images/71a70b27-b189-4e1d-8fb3-3e84ce02bf8e.medium.png

douglashall: Excellent, thanks for the confirmation.

course_discovery/apps/api/v1/views/program_types.py
```diff
@@ -0,0 +1,18 @@
...
+# pylint: disable=no-member
+from course_discovery.apps.course_metadata.models import ProgramType
+
+
+class ProgramTypeViewSet(viewsets.ReadOnlyModelViewSet):
+    """ Program resource. """
+    lookup_field = 'name'
+    serializer_class = serializers.ProgramTypeSerializer
+    permission_classes = (IsAuthenticated,)
+    filter_backends = (DjangoFilterBackend,)
```
clintonb: You're not filtering, so this isn't needed.

course_discovery/apps/api/v1/views/program_types.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework import viewsets
+from rest_framework.filters import DjangoFilterBackend
+from rest_framework.permissions import IsAuthenticated
+
+from course_discovery.apps.api import serializers
+
+
+# pylint: disable=no-member
```
clintonb: This is not needed.

course_discovery/apps/api/v1/views/program_types.py
```diff
@@ -0,0 +1,18 @@
...
+from course_discovery.apps.course_metadata.models import ProgramType
```
clintonb: Group this with the other import.

course_discovery/apps/api/v1/views/program_types.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class ProgramTypeViewSet(viewsets.ReadOnlyModelViewSet):
+    """ Program resource. """
```
clintonb: This is wrong.

course_discovery/apps/api/v1/views/program_types.py
```diff
@@ -0,0 +1,18 @@
...
+from rest_framework.permissions import IsAuthenticated
+
+from course_discovery.apps.api import serializers
+
+
+# pylint: disable=no-member
+from course_discovery.apps.course_metadata.models import ProgramType
+
+
+class ProgramTypeViewSet(viewsets.ReadOnlyModelViewSet):
```
clintonb: This can live in the programs module.

course_discovery/apps/api/v1/views/program_types.py
```diff
@@ -0,0 +1,18 @@
...
+    """ Program resource. """
+    lookup_field = 'name'
```
clintonb: This is not a good lookup field, because there may be spaces. Add a slug field, use it.

course_discovery/apps/course_metadata/models.py
```diff
@@ -574,12 +574,24 @@ def __str__(self):
 
 
 class ProgramType(TimeStampedModel):
-    name = models.CharField(max_length=32, unique=True, null=False, blank=False)
+    name = models.CharField(max_length=32, unique=True, null=False, blank=False, db_index=True)
```
clintonb: Unique fields are indexed by definition.

clintonb (changes requested):

course_discovery/apps/api/v1/views/program_types.py
clintonb: You'll want the AutoSlugfield to be unique and non-nullable.

afzaledx: This is still outstanding.

afzaledx: @clintonb, when I query the Programs list endpoint, I get the list of program dictionaries having just the ProgramType 'name' field.

https://github.com/edx/course-discovery/blob/375d616f9faf21a28d642d29e95ea959f895ffc4/course_discovery/apps/api/serializers.py#L536

After I get the list of Programs, I loop through the list each of those and hit the ProgramType endpoint that I just created to get the image_logo filed for each program.

clintonb: That is an inefficient solution. If you get 20 programs, you will poll the endpoint 20 times. Given the limited number of program types, you should get the full list of program types and cache it on the client. When you need details of a program type, refer to your local cache.

Additionally, `name` is a bad identity field because it is editable and may contain whitespace. `/api/v1/program_types/Professional Certification` is not a path we want in the API. Thus you want to use either a slug or UUID.

afzaledx: In that case, I will not need the lookup field at all since I only want the list of all the programs for my local cache.

afzaledx: If you remember, my first attempt was to add the type.logo_image in the programs endpoint. We already have the type (ProgramType) in the Programs model, but you suggested that it might cause a performance hit to include the type.logo_image. That's why I created this new endpoint.

afzaledx: The lookup_field defaults to id, and that is working just fine for individual program types.

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -71,7 +71,6 @@ def list(self, request, *args, **kwargs):
 
 class ProgramTypeViewSet(viewsets.ReadOnlyModelViewSet):
     """ ProgramType resource. """
-    lookup_field = 'name'
```
afzaledx: @clintonb, I have removed the lookup_field from here, mostly because of the "spaces" in the URL messing up the lookup. The efficiency of multiple lookups was easily handled because the catalog integration is cached by default, so they would not have been an overhead.

Since I was only getting the ProgramType's "name" field in the Programs endpoint that I need for looking up the logo_image, the only remaining solution is to fetch all the program types and do a local lookup. This is what I have done on the edx-platform side.

course_discovery/apps/api/serializers.py
```diff
@@ -689,6 +689,16 @@ class Meta(MinimalProgramSerializer.Meta):
         )
 
 
+class ProgramTypeSerializer(serializers.ModelSerializer):
+    logo_image = StdImageSerializerField()
+
+    class Meta:
+        model = ProgramType
+        fields = (
+            'id', 'name', 'logo_image',
```
clintonb: Do not expose database IDs. We don't want to repeat mistakes of the past and reference objects in this manner externally.

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -66,3 +67,10 @@ def list(self, request, *args, **kwargs):
               multiple: false
         """
         return super(ProgramViewSet, self).list(request, *args, **kwargs)
+
+
+class ProgramTypeViewSet(viewsets.ReadOnlyModelViewSet):
```
clintonb: Either add a slug field or make this a list-only view. Providing external references as database IDs is a non-starter. It opens us up to numerous problems that we can easily avoid.

course_discovery/conf/locale/en/LC_MESSAGES/djangojs.po
```diff
@@ -1,35 +0,0 @@
-# SOME DESCRIPTIVE TITLE.
```
clintonb, line 1: Why are these files changed?

clintonb (changes requested):

afzaledx:
@clintonb, I need help with the strange translations error that I am not familiar with.

clintonb (changes requested):
`ProgramTypeListViewSet` needs tests.

douglashall:
@clintonb Can you please describe for Afzal what needs to be changed in order to resolve the translation related errors in the failed checks? Thanks.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -160,3 +161,49 @@ def test_list_exclude_utm(self):
...
+            response = self.client.get(url)
+
+        self.assertEqual(
+            response.data['results'],
+            self.serialize_program_type(expected, many=True, extra_context=extra_context)
+        )
+
+    def test_list(self):
+        """ Verify the endpoint returns a list of all program_types. """
+        expected = [ProgramTypeFactory() for __ in range(3)]
```
clintonb: `ProgramTypeFactory.create_batch(3)`

course_discovery/apps/course_metadata/models.py
```diff
@@ -580,6 +580,18 @@ class ProgramType(TimeStampedModel):
...
+        upload_to=UploadToFieldNamePath(populate_from='id', path='media/program_types/logo_images'),
+        blank=True,
+        null=True,
+        variations={
+            'large': (256, 256),
+            'medium': (128, 128),
+            'small': (64, 64),
+            'x-small': (32, 32),
+        },
+        help_text=_('Please provide an image file with transparent background'),
```
clintonb, line 594: This is the source of your build failures. Run `make fake_translations`.

clintonb (approved):
See comments. Remember to squash your commits before merging. Delete your branch after merging.

afzaledx:
This test is passing locally but failing on travis. Will look into it later.

course_discovery/apps/course_metadata/models.py
```diff
@@ -580,6 +580,18 @@ class ProgramType(TimeStampedModel):
                               'associated courses, but enrolled in other seat types, will NOT have their completion '
                               'of the course counted toward the completion of the program.'),
     )
+    logo_image = StdImageField(
+        upload_to=UploadToFieldNamePath(populate_from='id', path='media/program_types/logo_images'),
```
douglashall: @afzaledx When I save a new program type with a logo_image in Django admin, the logo_image [filename](https://discovery-edx.sandbox.edx.org/api/v1/program_types/) is 'none.medium.png'. Since the `id` field isn't populated until you save the model to the database, it looks like it is not the right candidate for use as the filename. The best path forward is probably to add a UUIDField and use it as the filename.

afzaledx: @douglashall, Instead of the uuid, I have changed it to:

`upload_to=UploadToAutoSlug(populate_from='name', path='media/program_types/logo_images')`

With the ProgramType name as : `ProgramType with Caps And Spaces`, the file name becomes: `programtype-with-caps-and-spaces.png`

afzaledx: Moreover, if the file is changed, it auto-adds a uuid in the end to make it unique, e.g. `programtype-with-caps-and-spaces_FOPO7av.png`

douglashall:
👍  Latest is deployed to https://discovery-edx.sandbox.edx.org. Ready for merge.

douglashall:
@afzaledx Looks like you will need to rebase one more time before merging.

clintonb (commented):
Please add a migration to the [`edx_catalog_extensions` app](https://github.com/edx/course-discovery/tree/master/course_discovery/apps/edx_catalog_extensions) that adds the images now, or when they are available. Ideally, we should avoid manually adding these items so that it is easy to provision devstack and sandboxes.

douglashall:
@clintonb Good to know. We may or may not be able to add them as migrations depending on the contents of the name/logo and the confidentiality status of the partner.

PR #477, opened 2016-12-06
ECOM-4738 Use ngrams instead of wildcard for autocomplete and set up index settings/mappings

MatthewPiatetsky:
@edx/ecommerce 

I finally made a breakthrough last night in understanding how to set up autocomplete correctly.
Previously I had figured out a way to get things to sort of work with wildcards, but the missing pieces for things to really work the way we want were the index settings and mappings which needed to be properly set.
Ensuring that the ngram analyzer and lowercase analyzers are available in index settings and that the ngram analyzer is set as the index_analyzer and the lowercase analyzer is set as the search analyzer in the mappings allows the autocomplete to work much better.
There is quite a bit going on here, so let me know if you have questions.

On the plus side, it should be very simple to do add the synonym analyzers now!

### DIFF
--- a/course_discovery/apps/api/v1/views/search.py
+++ b/course_discovery/apps/api/v1/views/search.py
@@ -121,23 +121,36 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
 
 
 class TypeaheadSearchView(APIView):
-    """
-    Typeahead for courses and programs.
-    """
+    """ Typeahead for courses and programs. """
     RESULT_COUNT = 3
     permission_classes = (IsAuthenticated,)
 
     def get_results(self, query):
-        query = '(title:*{query}* OR course_key:*{query}*)'.format(query=query.lower())
-        course_runs = SearchQuerySet().models(CourseRun).raw_search(query)
+        course_runs = SearchQuerySet().models(CourseRun).filter(SQ(title_autocomplete=query) | SQ(course_key=query))
         course_runs = course_runs.filter(published=True).exclude(hidden=True)
         course_runs = course_runs[:self.RESULT_COUNT]
-        programs = SearchQuerySet().models(Program).raw_search(query)
+        programs = SearchQuerySet().models(Program).filter(SQ(title_autocomplete=query))
         programs = programs.filter(status=ProgramStatus.Active)
         programs = programs[:self.RESULT_COUNT]
         return course_runs, programs
 
     def get(self, request, *args, **kwargs):
+        """
+        Typeahead uses the ngram_analyzer as the index_analyzer to generate ngrams of the title during indexing.
+        i.e. Data Science -> da, dat, at, ata, data, etc...
+        Typeahead uses the lowercase analyzer as the search_analyzer.
+        The ngram_analyzer uses the lowercase filter as well, which makes typeahead case insensitive.
+        Available analyzers are defined in index _settings and field level analyzers are defined in the index _mapping.
+        NGrams are used rather than EdgeNgrams because NGrams allow partial searches across white space:
+        i.e. data sci - > data science, but not data analysis or scientific method
+        ---
+        parameters:
+            - name: q
+              description: "Search text"
+              paramType: query
+              required: true
+              type: string
+        """
         query = request.query_params.get('q')
         if not query:
             raise ParseError("The 'q' querystring parameter is required for searching.")
--- a/course_discovery/apps/core/tests/mixins.py
+++ b/course_discovery/apps/core/tests/mixins.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.conf import settings
-from elasticsearch import Elasticsearch
+from haystack import connections as haystack_connections
 
 from course_discovery.apps.core.utils import ElasticsearchUtils
 
@@ -12,12 +12,19 @@ class ElasticsearchTestMixin(object):
     @classmethod
     def setUpClass(cls):
         super(ElasticsearchTestMixin, cls).setUpClass()
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
+        # Make use of the changes in our custom ES backend
+        # This is required for typeahead autocomplete to work in the tests
+        connection = haystack_connections['default']
+        cls.backend = connection.get_backend()
+        # Without this line, haystack doesn't fully recreate the connection
+        # The first test using this backend succeeds, but the following tests
+        # do not set the Elasticsearch _mapping
 
     def setUp(self):
         super(ElasticsearchTestMixin, self).setUp()
+        self.backend.setup_complete = False
+        self.es = self.backend.conn
         self.reset_index()
         self.refresh_index()
 
--- a/course_discovery/apps/core/utils.py
+++ b/course_discovery/apps/core/utils.py
@@ -1,6 +1,8 @@
 import datetime
 import logging
 
+from django.conf import settings
+
 logger = logging.getLogger(__name__)
 
 
@@ -16,7 +18,8 @@ class ElasticsearchUtils(object):
             # Create an index with a unique (timestamped) name
             timestamp = datetime.datetime.utcnow().strftime("%Y%m%d%H%M%S")
             index = '{alias}_{timestamp}'.format(alias=alias, timestamp=timestamp)
-            es.indices.create(index=index)
+            index_settings = settings.ELASTICSEARCH_INDEX_SETTINGS
+            es.indices.create(index=index, body=index_settings)
             logger.info('...index [%s] created.', index)
 
             # Point the alias to the new index
--- a/course_discovery/apps/edx_haystack_extensions/backends.py
+++ b/course_discovery/apps/edx_haystack_extensions/backends.py
@@ -74,8 +74,37 @@ class NonClearingSearchBackendMixin(object):
 
 
 # pylint: disable=abstract-method
+class ConfigurableElasticBackend(ElasticsearchSearchBackend):
+
+    def specify_analyzers(self, mapping, field, index_analyzer, search_analyzer):
+        """ Specify separate index and search analyzers for the given field.
+          Args:
+            mapping (dict): /_mapping attribute on index (maps analyzers to fields)
+            field (str): which field to modify
+            index_analyzer (str): name of the index_analyzer (should be defined in the /_settings attribute)
+            search_analyzer (str): name of the search_analyzer (should be defined in the /_settings attribute)
+        """
+        # The generic analyzer is used for both if index_analyzer and search_analyzer are not specified
+        mapping[field].pop('analyzer')
+        mapping[field].update({
+            'index_analyzer': index_analyzer,
+            'search_analyzer': search_analyzer
+        })
+
+    def build_schema(self, fields):
+        content_field_name, mapping = super().build_schema(fields)
+        # Use the ngram analyzer as the index_analyzer and the lowercase analyzer as the search_analyzer
+        # This is necessary to support partial searches/typeahead
+        # If we used ngram analyzer for both, then 'running' would get split into ngrams like "ing"
+        # and all words containing ing would come back in typeahead.
+        self.specify_analyzers(mapping=mapping, field='title_autocomplete',
+                               index_analyzer='ngram_analyzer', search_analyzer='lowercase')
+        return (content_field_name, mapping)
+
+
+# pylint: disable=abstract-method
 class EdxElasticsearchSearchBackend(SimpleQuerySearchBackendMixin, NonClearingSearchBackendMixin,
-                                    ElasticsearchSearchBackend):
+                                    ConfigurableElasticBackend):
     pass
 
 
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -350,6 +350,67 @@ SWAGGER_SETTINGS = {
     'permission_denied_handler': 'course_discovery.apps.api.views.api_docs_permission_denied_handler'
 }
 
+# Elasticsearch uses index settings to specify available analyzers.
+# We are adding the lowercase analyzer and tweaking the ngram analyzers here,
+# so we need to use these settings rather than the index defaults.
+# We are making these changes to enable autocomplete for the typeahead endpoint.
+ELASTICSEARCH_INDEX_SETTINGS = {
+    'settings': {
+        'analysis': {
+            'tokenizer': {
+                'haystack_edgengram_tokenizer': {
+                    'type': 'edgeNGram',
+                    'side': 'front',
+                    'min_gram': 2,
+                    'max_gram': 15
+                },
+                'haystack_ngram_tokenizer': {
+                    'type': 'nGram',
+                    'min_gram': 2,
+                    'max_gram': 15
+                }
+            },
+            'analyzer': {
+                'lowercase': {
+                    'type': 'custom',
+                    'tokenizer': 'keyword',
+                    'filter': [
+                        'lowercase'
+                    ]
+                },
+                'ngram_analyzer': {
+                    'type':'custom',
+                    'filter': [
+                        'haystack_ngram',
+                        'lowercase'
+                    ],
+                    'tokenizer': 'standard'
+                },
+                'edgengram_analyzer': {
+                    'type': 'custom',
+                    'filter': [
+                        'haystack_edgengram',
+                        'lowercase'
+                    ],
+                    'tokenizer': 'standard'
+                }
+            },
+            'filter': {
+                'haystack_edgengram': {
+                    'type': 'edgeNGram',
+                    'min_gram': 2,
+                    'max_gram': 15
+                },
+                'haystack_ngram': {
+                    'type': 'nGram',
+                    'min_gram': 2,
+                    'max_gram': 15
+                }
+            }
+        }
+    }
+}
+
 # Haystack configuration (http://django-haystack.readthedocs.io/en/v2.5.0/settings.html)
 HAYSTACK_ITERATOR_LOAD_PER_QUERY = 200
 
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -298,7 +298,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
                              [self.serialize_course_run(course_run), self.serialize_program(program)])
 
 
-class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, APITestCase):
+class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase):
     path = reverse('api:v1:search-typeahead')
     function_score = {
         'functions': [
@@ -307,12 +307,6 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, APITestC
             {'linear': {'start': {'origin': 'now', 'scale': '1d', 'decay': 0.95}}, 'weight': 5.0}
         ],
         'boost': 1.0, 'score_mode': 'sum', 'boost_mode': 'sum',
-        'query': {
-            'query_string': {
-                'auto_generate_phrase_queries': True, 'analyze_wildcard': True,
-                'query': '((title:*pytho* OR course_key:*pytho*) AND status:(active))'
-            }
-        }
     }
 
     def get_typeahead_response(self, query=None):
@@ -323,7 +317,7 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, APITestC
 
         url = '{path}?{qs}'.format(path=self.path, qs=qs)
         config = ElasticsearchBoostConfig.get_solo()
-        config.function_score = self.function_score
+        config.function_score.update(self.function_score)
         config.save()
         return self.client.get(url)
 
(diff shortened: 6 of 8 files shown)


course_discovery/apps/api/v1/views/search.py
```diff
@@ -128,11 +128,11 @@ class TypeaheadSearchView(APIView):
     permission_classes = (IsAuthenticated,)
 
     def get_results(self, query):
-        query = '(title:*{query}* OR course_key:*{query}*)'.format(query=query.lower())
-        course_runs = SearchQuerySet().models(CourseRun).raw_search(query)
+        query = query.lower()
```
rlucioni: This can be removed (the lowercase analyzer handles this).

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,19 +12,22 @@ class ElasticsearchTestMixin(object):
...
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
 
     def setUp(self):
         super(ElasticsearchTestMixin, self).setUp()
+        backends = list(haystack_connections.connections_info.keys())
+        connection = haystack_connections[backends[0]]
+        backend = connection.get_backend()
+        backend.setup_complete = False
```
rlucioni: I'd include a comment about why this is necessary/what it achieves.

course_discovery/apps/core/utils.py
```diff
@@ -16,18 +43,12 @@ def create_alias_and_index(cls, es, alias):
...
-            # Point the alias to the new index
-            body = {
-                'actions': [
-                    {'remove': {'alias': alias, 'index': '*'}},
-                    {'add': {'alias': alias, 'index': index}},
-                ]
-            }
-            es.indices.update_aliases(body)
-            logger.info('...alias updated.')
+            if is_test:
```
rlucioni: Instead, just override HAYSTACK_CONNECTIONS in settings/test.py.

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -74,8 +74,23 @@ def clear(self, models=None, commit=True):  # pylint: disable=unused-argument
...
+class ConfigurableElasticBackend(ElasticsearchSearchBackend):
+
+    def build_schema(self, fields):
+        content_field_name, mapping = super(ConfigurableElasticBackend,
+                                            self).build_schema(fields)
+        title_autocomplete_mapping = mapping['title_autocomplete']
+        title_autocomplete_mapping.pop('analyzer')
+        title_autocomplete_mapping['index_analyzer'] = 'ngram_analyzer'
+        title_autocomplete_mapping['search_analyzer'] = 'lowercase'
+        mapping['title_autocomplete'] = title_autocomplete_mapping
```
rlucioni: Consider breaking this out into a separate method to make what's going on here more clear (e.g., `configure_title_autocomplete_analyzers`).

course_discovery/settings/base.py
```diff
@@ -353,6 +353,64 @@
 # Haystack configuration (http://django-haystack.readthedocs.io/en/v2.5.0/settings.html)
 HAYSTACK_ITERATOR_LOAD_PER_QUERY = 200
 
+ELASTICSEARCH_DEFAULT_ANALYZER = 'snowball'
+ELASTICSEARCH_INDEX_SETTINGS = {
+  "settings": {
```
rlucioni: Nitpick (I'm surprised pylint didn't complain): 4 spaces and single quotes, to be consistent with everything else, please.

rlucioni (changes requested):

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: fixed

course_discovery/apps/core/tests/mixins.py
MatthewPiatetsky: added

course_discovery/apps/core/utils.py
MatthewPiatetsky: fixed

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: changed

course_discovery/settings/base.py
MatthewPiatetsky: fixed

clintonb:
How does this differ from the Auto-complete functionality built into `drf-haystack`: https://drf-haystack.readthedocs.io/en/latest/02_autocomplete.html?

course_discovery/apps/api/v1/views/search.py
```diff
@@ -123,16 +123,31 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
 class TypeaheadSearchView(APIView):
     """
     Typeahead for courses and programs.
-    """
+
```
MatthewPiatetsky: @rlucioni I also added this comment below, which I think may be helpful in explaining the general idea (and is readable in swagger)

MatthewPiatetsky:
@clintonb This is using the same functionality, but has some additional changes..using different index vs search analyzers (by default autocomplete uses ngram for both) as well as actually initializing the index settings/mapping for both index creation and tests, the comment I tagged you on has some context on this.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -123,16 +123,31 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+    Typeahead uses the ngram_analyzer as the index_analyzer to generate ngrams of the title during indexing.<br>
+    i.e. Data Science -> da, dat, at, ata, data, etc... <br><br>
+    Typeahead uses the lowercase analyzer as the search_analyzer. <br>
+    The ngram_analyzer uses the lowercase filter as well, which makes typeahead case insensitive. <br><br>
+    Available analyzers are defined in index _settings and field level analyzers are defined in _mapping. <br>
+    NGrams are used rather than EdgeNgrams because NGrams allow partial searches across white space: <br>
+    i.e. data sci - > data science, but not data analysis or scientific method <br>
+    ---
+    GET:
```
MatthewPiatetsky: @clintonb I'm trying to figure this out now, but do you know if there is a way to get this docstring to work properly for both the drf swagger UI and the self describing API UI?

clintonb: http://django-rest-swagger.readthedocs.io/en/stable-0.3.x/yaml.html

MatthewPiatetsky: What I have currently seems to work fine with swagger, but get a error message rendered in the drf view (the results do come back), if I omit the GET then it works in drf view but doesn't even load in swagger

MatthewPiatetsky: thanks! I just needed to add the docstring to the method rather than the class

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,12 +12,17 @@ class ElasticsearchTestMixin(object):
...
     def setUpClass(cls):
         super(ElasticsearchTestMixin, cls).setUpClass()
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
 
     def setUp(self):
         super(ElasticsearchTestMixin, self).setUp()
+        backends = list(haystack_connections.connections_info.keys())
+        connection = haystack_connections[backends[0]]
```
rlucioni: Can you look up the connection by name? Is the backend at index 0 always going to be correct? Typically, Django would have you use a name to look up a connection in a situation like this (e.g., default).

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,12 +12,17 @@ class ElasticsearchTestMixin(object):
...
+        backend = connection.get_backend()
+        # Without this line, haystack doesn't fully recreate the connection
+        # This results in the tests after the first run missing the field mapping
+        backend.setup_complete = False
```
rlucioni: Should this be set in a teardown method?

course_discovery/settings/base.py
```diff
@@ -350,6 +350,63 @@
     'permission_denied_handler': 'course_discovery.apps.api.views.api_docs_permission_denied_handler'
 }
 
+ELASTICSEARCH_INDEX_SETTINGS = {
+    'settings':{
```
rlucioni: pep8 should be upset with you about this, too. You need a space between the colon and the brace, here and throughout.

course_discovery/settings/test.py
```diff
@@ -25,3 +25,11 @@
 LOGGING['handlers']['local'] = {'class': 'logging.NullHandler'}
 
 PUBLISHER_FROM_EMAIL = 'test@example.com'
+
+HAYSTACK_CONNECTIONS = {
+    'default': {
+        'ENGINE': 'course_discovery.apps.edx_haystack_extensions.backends.EdxElasticsearchSearchEngine',
+        'URL': 'http://127.0.0.1:9200/',
```
rlucioni: The [Travis build log](https://travis-ci.org/edx/course-discovery/builds/181793725#L7239) suggests this may not work out after all. Looks like the tests can't connect to Elasticsearch.

MatthewPiatetsky: there were some builds that passed fully earlier I'll figure out what broke tomorrow

course_discovery/apps/core/tests/mixins.py
MatthewPiatetsky: fixed

course_discovery/apps/core/tests/mixins.py
MatthewPiatetsky: idk, this flag is set in the set up process (on the previous line I think), if we don't want it set think it makes sense to set to False right after

course_discovery/settings/base.py
MatthewPiatetsky: fixed

course_discovery/apps/api/v1/views/search.py
```diff
@@ -121,23 +121,42 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
         course_runs = course_runs[:self.RESULT_COUNT]
-        programs = SearchQuerySet().models(Program).raw_search(query)
+        programs = SearchQuerySet().models(Program).filter(SQ(title_autocomplete=query))
         programs = programs.filter(status=ProgramStatus.Active)
         programs = programs[:self.RESULT_COUNT]
         return course_runs, programs
 
     def get(self, request, *args, **kwargs):
+        """
+        Typeahead uses the ngram_analyzer as the index_analyzer to generate ngrams of the title during indexing.
```
clintonb, line 139: Clean this up to be a proper paragraph. As is, it reads like a disjointed list, and is unclear.

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,12 +12,16 @@ class ElasticsearchTestMixin(object):
...
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
 
     def setUp(self):
         super(ElasticsearchTestMixin, self).setUp()
+        connection = haystack_connections['default']
+        backend = connection.get_backend()
+        # Without this line, haystack doesn't fully recreate the connection
+        # This results in the tests after the first run missing the field mapping
```
clintonb: You're missing something in this sentence.

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,12 +12,16 @@ class ElasticsearchTestMixin(object):
     @classmethod
     def setUpClass(cls):
         super(ElasticsearchTestMixin, cls).setUpClass()
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
 
     def setUp(self):
         super(ElasticsearchTestMixin, self).setUp()
+        connection = haystack_connections['default']
```
clintonb: Why are these changes necessary?

course_discovery/settings/base.py
```diff
@@ -350,6 +350,63 @@
     'permission_denied_handler': 'course_discovery.apps.api.views.api_docs_permission_denied_handler'
 }
 
+ELASTICSEARCH_INDEX_SETTINGS = {
```
clintonb, line 357: Add a comment explaining what you're doing, and why this is necessary.

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -74,8 +74,29 @@ def clear(self, models=None, commit=True):  # pylint: disable=unused-argument
 
 
 # pylint: disable=abstract-method
+class ConfigurableElasticBackend(ElasticsearchSearchBackend):
```
clintonb, line 77: This should be a mixin added to `EdxElasticsearchSearchBackend`.

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -74,8 +74,29 @@ def clear(self, models=None, commit=True):  # pylint: disable=unused-argument
...
+        # The generic analyzer is used for both if index_analyzer and search_analyzer are not specified
+        mapping[field].pop('analyzer')
+        mapping[field].update({
+            'index_analyzer': index_analyzer,
+            'search_analyzer': search_analyzer
+        })
+
+    def build_schema(self, fields):
+        content_field_name, mapping = super().build_schema(fields)
+        # We need to use a lowercase analyzer for search instead of the ngram_analyzer.
```
clintonb:
```python
# Use the ngram analyzer for the autocomplete field to support partial searches/typeahead.
```

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -74,8 +74,29 @@ def clear(self, models=None, commit=True):  # pylint: disable=unused-argument
 
 
 # pylint: disable=abstract-method
+class ConfigurableElasticBackend(ElasticsearchSearchBackend):
+
+    def specify_analyzers(self, mapping, field, index_analyzer, search_analyzer):
+        # The generic analyzer is used for both if index_analyzer and search_analyzer are not specified
```
clintonb, line 87:
```python
""" Specify separate index and search analyzers for the given field.
  Args:
    ...
"""
```

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: I did it this way to add line breaks in swagger. I can change it to be a paragraph, but then I think the swagger will be one big block with no line breaks?

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: It already is, right?

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: That's not quite what this means, since I am using the lowercase analyzer as the search_analyzer
Is this clearer?
```
# Use the ngram analyzer as the index_analyzer and the lowercase analyzer as the search_analyzer
# This is necessary to support partial searches/typeahead
# If we used ngram analyzer for both, then 'running' would get split into ngrams like "ing"
# and all words containing ing would come back in typeahead.
```

course_discovery/settings/base.py
```diff
@@ -350,6 +350,63 @@
...
+    'settings': {
+        'analysis': {
+            'tokenizer': {
+                'haystack_edgengram_tokenizer': {
+                    'type':'edgeNGram',
+                    'side':'front',
+                    'min_gram':2,
```
rlucioni: You're still missing a space between the colon and the characters that come afterwards, here and throughout this dict. I also just realized that pep8 isn't failing your build because we tell it to ignore settings files. You can try running it locally against this file to pick out all the places were you need to add a space.

course_discovery/apps/core/tests/mixins.py
MatthewPiatetsky: added comment, let me know if it is unclear

course_discovery/apps/core/tests/mixins.py
MatthewPiatetsky: not sure what you meant, but is it clearer now?

course_discovery/settings/base.py
MatthewPiatetsky: added a comment

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: added a comment

course_discovery/settings/base.py
MatthewPiatetsky: fixed

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,12 +12,19 @@ class ElasticsearchTestMixin(object):
...
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
 
     def setUp(self):
         super(ElasticsearchTestMixin, self).setUp()
+        # Make use of the changes in our custom ES backend
+        # This is required for typeahead autocomplete to work in the tests
+        connection = haystack_connections['default']
+        backend = connection.get_backend()
```
rlucioni: What happens if you move most of this to setUpClass? It seems like the only line required here should be `backend.setup_complete = False`, since that's what you appear to need run before each test.

rlucioni (approved):
@MatthewPiatetsky looks good, just one last question about moving some test setup code.

rlucioni:
@MatthewPiatetsky looks like you might have introduced a flaky test. Can you take a look at the [build log](https://travis-ci.org/edx/course-discovery/builds/182077215#L6172) and see about clearing that up?

course_discovery/apps/api/v1/views/search.py
clintonb: The Swagger readability is less important than code readability. Also, at some point when we upgrade DRF, the method by which we extract Swagger text will change.

MatthewPiatetsky:
@rlucioni thanks, looking into it

course_discovery/apps/edx_haystack_extensions/backends.py
clintonb: No. You're subclassing `ConfigurableElasticBackend`, which is not necessary.

course_discovery/settings/base.py
```diff
@@ -350,6 +350,66 @@
     'permission_denied_handler': 'course_discovery.apps.api.views.api_docs_permission_denied_handler'
 }
 
+# Elasticsearch uses index settings to specify available analyzers
+# We are adding the lowercase analyzer and tweaking the ngram analyzers here
+# so we need to use these settings rather than the index defaults
```
clintonb: Explain _why_ this is necessary. Would @MatthewPiatetsky from last week know why we need an ngram analyzer?

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -74,8 +74,37 @@ def clear(self, models=None, commit=True):  # pylint: disable=unused-argument
 
 
 # pylint: disable=abstract-method
+class ConfigurableElasticBackend(ElasticsearchSearchBackend):
+
+    def specify_analyzers(self, mapping, field, index_analyzer, search_analyzer):
+        """ Specify separate index and search analyzers for the given field.
+          Args:
+            mapping: /_mapping attribute on index (maps analyzers to fields)
```
clintonb: Add types after the argument names.

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: I need the parent to be ElasticsearchSearchBackend to get the mapping in build_schema.
Are you suggesting I put just the specify analyzers function in a mixin?

course_discovery/settings/base.py
MatthewPiatetsky: not sure exactly what you mean by why, does the line below clarify?

clintonb: Yes, that works. You just need some periods in these sentences.

clintonb (approved):

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: updated

course_discovery/apps/core/tests/mixins.py
MatthewPiatetsky: good idea, changed

course_discovery/apps/core/tests/mixins.py
```diff
@@ -12,12 +12,19 @@ class ElasticsearchTestMixin(object):
...
-        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
         cls.index = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        cls.es = Elasticsearch(host)
+        # Make use of the changes in our custom ES backend
+        # This is required for typeahead autocomplete to work in the tests
+        connection = haystack_connections['default']
+        cls.backend = connection.get_backend()
+        # Without this line, haystack doesn't fully recreate the connection
+        # The first test using this backend succeeds, but the following tests
+        # do not set the Elasticsearch _mapping
```
rlucioni, line 22: @MatthewPiatetsky you might want to open a new PR to move this comment above the line it refers to.

PR #482, opened 2016-12-08
Add logo image url to search results

MatthewPiatetsky:
@edx/ecommerce 

We want to expose the authoring organization's logo in the new list view

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -1008,11 +1008,12 @@ class AffiliateWindowSerializerTests(TestCase):
 
 class CourseRunSearchSerializerTests(TestCase):
     def test_data(self):
-        course_run = CourseRunFactory(transcript_languages=LanguageTag.objects.filter(code__in=['en-us', 'zh-cn']))
+        course_run = CourseRunFactory(transcript_languages=LanguageTag.objects.filter(code__in=['en-us', 'zh-cn']),
+                                      authoring_organizations=[OrganizationFactory()])
         ProgramFactory(courses=[course_run.course])
         serializer = self.serialize_course_run(course_run)
         course_run_key = CourseKey.from_string(course_run.key)
-
+        orgs = course_run.authoring_organizations.all()
         expected = {
             'transcript_languages': [serialize_language(l) for l in course_run.transcript_languages.all()],
             'short_description': course_run.short_description,
@@ -1038,6 +1039,7 @@ class CourseRunSearchSerializerTests(TestCase):
             'published': course_run.status == CourseRunStatus.Published,
             'partner': course_run.course.partner.short_code,
             'program_types': course_run.program_types,
+            'logo_image_urls': [org.logo_image_url for org in orgs],
             'authoring_organization_uuids': get_uuids(course_run.authoring_organizations.all()),
             'subject_uuids': get_uuids(course_run.subjects.all()),
             'staff_uuids': get_uuids(course_run.staff.all())
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -65,9 +65,14 @@ class BaseCourseIndex(OrganizationsMixin, BaseIndex):
     subjects = indexes.MultiValueField(faceted=True)
     organizations = indexes.MultiValueField(faceted=True)
     authoring_organizations = indexes.MultiValueField(faceted=True)
+    logo_image_urls = indexes.MultiValueField()
     sponsoring_organizations = indexes.MultiValueField(faceted=True)
     level_type = indexes.CharField(null=True, faceted=True)
     partner = indexes.CharField(model_attr='partner__short_code', null=True, faceted=True)
+
+    def prepare_logo_image_urls(self, obj):
+        orgs = obj.authoring_organizations.all()
+        return [org.logo_image_url for org in orgs]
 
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -200,6 +206,10 @@ class ProgramIndex(BaseIndex, indexes.Indexable, OrganizationsMixin):
     seat_types = indexes.MultiValueField(model_attr='seat_types', null=True, faceted=True)
     published = indexes.BooleanField(null=False, faceted=True)
 
+    def prepare_logo_image_url(self, obj):
+        org = obj.authoring_organizations.first()
+        return org.logo_image_url if org else ''
```
mikedikan: curious why empty string instead of null?

MatthewPiatetsky: good point fixed

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -65,10 +65,15 @@ class BaseCourseIndex(OrganizationsMixin, BaseIndex):
     subjects = indexes.MultiValueField(faceted=True)
     organizations = indexes.MultiValueField(faceted=True)
     authoring_organizations = indexes.MultiValueField(faceted=True)
+    logo_image_url = indexes.CharField()
     sponsoring_organizations = indexes.MultiValueField(faceted=True)
     level_type = indexes.CharField(null=True, faceted=True)
     partner = indexes.CharField(model_attr='partner__short_code', null=True, faceted=True)
 
+    def prepare_logo_image_url(self, obj):
+        org = obj.authoring_organizations.first()
```
clintonb: Given the our data supports multiple authoring organizations, we should index them and return them from the API. Leave it up to the client to decide which to use. We should not artificially limit ourselves, especially given that we know we eventually want to support multiple authoring organizations.

clintonb (changes requested):

mikedikan (approved):

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -200,6 +206,10 @@ class ProgramIndex(BaseIndex, indexes.Indexable, OrganizationsMixin):
     seat_types = indexes.MultiValueField(model_attr='seat_types', null=True, faceted=True)
     published = indexes.BooleanField(null=False, faceted=True)
 
+    def prepare_logo_image_url(self, obj):
+        org = obj.authoring_organizations.first()
+        return org.logo_image_url if org else None
+
```
davec-edx: This method is the same in both BaseCourseIndex and ProgramIndex, should it be moved to BaseIndex?

course_discovery/apps/course_metadata/search_indexes.py
MatthewPiatetsky: I got a different impression from Jasper, but updated

course_discovery/apps/course_metadata/search_indexes.py
MatthewPiatetsky: fixed

MatthewPiatetsky:
@clintonb @davec-edx have I addressed your comments?

davec-edx:
👍

MatthewPiatetsky:
@clintonb @davec-edx have I addressed your comments? fyi I realized that we already had access to logo_image_url for programs through the authoring_organization bodies, so I only added for course runs/courses

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -65,10 +65,15 @@ class BaseCourseIndex(OrganizationsMixin, BaseIndex):
...
     organizations = indexes.MultiValueField(faceted=True)
     authoring_organizations = indexes.MultiValueField(faceted=True)
+    logo_image_urls = indexes.MultiValueField()
     sponsoring_organizations = indexes.MultiValueField(faceted=True)
     level_type = indexes.CharField(null=True, faceted=True)
     partner = indexes.CharField(model_attr='partner__short_code', null=True, faceted=True)
 
+    def prepare_logo_image_urls(self, obj):
+        orgs = obj.authoring_organizations.all()
+        return [org.logo_image_url for org in orgs if org]
```
clintonb: `return list(obj.authoring_organizations.values_list('logo_image_url', flat=True))`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1038,6 +1039,7 @@ def test_data(self):
             'published': course_run.status == CourseRunStatus.Published,
             'partner': course_run.course.partner.short_code,
             'program_types': course_run.program_types,
+            'logo_image_urls': [org.logo_image_url for org in orgs if org],
```
clintonb: The `if` statement is not necessary.

clintonb (approved):
See my comments.

course_discovery/apps/course_metadata/search_indexes.py
MatthewPiatetsky: last time I did this I noticed that values_list ended up generating additional SQL queries, this seems simpler rather than doing that and prefetching

course_discovery/apps/api/tests/test_serializers.py

course_discovery/apps/course_metadata/search_indexes.py
clintonb: Get rid of the if statement.

PR #490, opened 2016-12-14
Boost orgs autocomplete in typeahead

MatthewPiatetsky:
@edx/ecommerce 

Jasper found yet another issue where a smithsonian course was coming up above a MIT course.
We tested this by reindexing locally and it fixed the issue.
There are no other typeahead issues that we are aware of

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -974,29 +974,23 @@ class AggregateSearchSerializer(HaystackSerializer):
         }
 
 
-class TypeaheadCourseRunSearchSerializer(serializers.Serializer):
-    org = serializers.CharField()
-    title = serializers.CharField()
-    key = serializers.CharField()
-    marketing_url = serializers.CharField()
-
-    class Meta:
-        fields = ['key', 'title']
-
-
-class TypeaheadProgramSearchSerializer(serializers.Serializer):
+class TypeaheadBaseSearchSerializer(serializers.Serializer):
     orgs = serializers.SerializerMethodField()
-    uuid = serializers.CharField()
     title = serializers.CharField()
-    type = serializers.CharField()
     marketing_url = serializers.CharField()
 
     def get_orgs(self, result):
         authoring_organizations = [json.loads(org) for org in result.authoring_organization_bodies]
         return [org['key'] for org in authoring_organizations]
 
-    class Meta:
-        fields = ['uuid', 'title', 'type']
+
+class TypeaheadCourseRunSearchSerializer(TypeaheadBaseSearchSerializer):
+    key = serializers.CharField()
+
+
+class TypeaheadProgramSearchSerializer(TypeaheadBaseSearchSerializer):
+    uuid = serializers.CharField()
+    type = serializers.CharField()
 
 
 class TypeaheadSearchSerializer(serializers.Serializer):
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -1125,14 +1125,14 @@ class ProgramSearchSerializerTests(TestCase):
 
 class TypeaheadCourseRunSearchSerializerTests(TestCase):
     def test_data(self):
-        course_run = CourseRunFactory()
+        authoring_organization = OrganizationFactory()
+        course_run = CourseRunFactory(authoring_organizations=[authoring_organization])
         serialized_course = self.serialize_course_run(course_run)
-        course_run_key = CourseKey.from_string(course_run.key)
 
         expected = {
             'key': course_run.key,
             'title': course_run.title,
-            'org': course_run_key.org,
+            'orgs': [org.key for org in course_run.authoring_organizations.all()],
             'marketing_url': course_run.marketing_url,
         }
         self.assertDictEqual(serialized_course.data, expected)
@@ -1150,7 +1150,7 @@ class TypeaheadProgramSearchSerializerTests(TestCase):
             'uuid': str(program.uuid),
             'title': program.title,
             'type': program.type.name,
-            'orgs': list(program.authoring_organizations.all().values_list('key', flat=True)),
+            'orgs': [org.key for org in program.authoring_organizations.all()],
             'marketing_url': program.marketing_url,
         }
 
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -465,3 +465,33 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, Elastics
             'programs': [self.serialize_program(program)]
         }
         self.assertDictEqual(response.data, expected)
+
+    def test_typeahead_org_course_runs_come_up_first(self):
+        """ Test typeahead response to ensure org is taken into account. """
+        MITx = OrganizationFactory(key='MITx')
+        HarvardX = OrganizationFactory(key='HarvardX')
+        mit_run = CourseRunFactory(
+            authoring_organizations=[MITx, HarvardX],
+            title='MIT Testing1'
+        )
+        harvard_run = CourseRunFactory(
+            authoring_organizations=[HarvardX],
+            title='MIT Testing2'
+        )
+        mit_program = ProgramFactory(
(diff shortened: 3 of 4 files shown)


course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -465,3 +465,24 @@ def test_typeahead_authoring_organizations_partial_search(self):
             'programs': [self.serialize_program(program)]
         }
         self.assertDictEqual(response.data, expected)
+
+    def test_typeahead_org_course_runs_come_up_first(self):
+        """ Test typeahead response to ensure an org's course runs come up first. """
+        MITx = OrganizationFactory(key="MITx")
```
rlucioni: Nitpick: keep your quote level consistent. We tend to prefer single quotes.

clintonb (approved):

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -465,3 +465,24 @@ def test_typeahead_authoring_organizations_partial_search(self):
...
+        smithsonian_run = CourseRunFactory(
+            authoring_organizations=[SmithsonianX],
+            title="Objects That Define America"
+        )
+        response = self.get_typeahead_response("mit")
+        self.assertEqual(response.status_code, 200)
+        expected = {
+            'course_runs': [self.serialize_course_run(mit_run),
+                            self.serialize_course_run(smithsonian_run)],
+            'programs': []
```
rlucioni: Since you're boosting the org field for programs, too, a complete test should also cover programs.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -465,3 +465,24 @@ def test_typeahead_authoring_organizations_partial_search(self):
...
+        """ Test typeahead response to ensure an org's course runs come up first. """
+        MITx = OrganizationFactory(key="MITx")
+        SmithsonianX = OrganizationFactory(key="SmithsonianX")
+        mit_run = CourseRunFactory(
+            authoring_organizations=[MITx],
+            title="Visualizing Japan (1850s-1930s): Westernization, Protest, Modernity"
+        )
+        smithsonian_run = CourseRunFactory(
+            authoring_organizations=[SmithsonianX],
+            title="Objects That Define America"
```
mikedikan: Shouldn't we be testing that the title contains the "mit" string?

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -14,6 +14,10 @@
 # If we altered our boosting functions to have a max score of 10
 # we would probably want to bump this number.
 TITLE_FIELD_BOOST = 25.0
+# We want to boost org the same as title
+# Primarily this is to prevent courses from other institutions
+# coming up first if there is a partial match on title
+ORG_FIELD_BOOST = 25.0
```
rlucioni: In case `TITLE_FIELD_BOOST` is changed in the future, I would make this:
```py
ORG_FIELD_BOOST = TITLE_FIELD_BOOST
```
Along with your comment, that makes the expected relationship between the two very clear.

rlucioni (changes requested):

course_discovery/apps/api/v1/tests/test_views/test_search.py

course_discovery/apps/api/v1/tests/test_views/test_search.py

course_discovery/apps/course_metadata/search_indexes.py
MatthewPiatetsky: fixed

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: good point, I've changed the test to test that the org is taken into account but with a title match in both, Jasper and I did some discovery and there are some complex things going on, so we aren't quite fixing everything yet, but we will be doing some follow up work after releasing typeahead. These changes should give us the results we want with our current data, which should be fine for unblocking release of typeahead

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1125,14 +1125,14 @@ def test_data_without_organizations(self):
...
+        authoring_organization = OrganizationFactory()
+        course_run = CourseRunFactory(authoring_organizations=[authoring_organization])
         serialized_course = self.serialize_course_run(course_run)
-        course_run_key = CourseKey.from_string(course_run.key)
 
         expected = {
             'key': course_run.key,
             'title': course_run.title,
-            'org': course_run_key.org,
+            'orgs': list(course_run.authoring_organizations.all().values_list('key', flat=True)),
```
rlucioni: Can you make this more readable with a list comprehension?
```py
'orgs': [org.key for org in course_run.authoring_organizations.all()]
```

course_discovery/apps/api/serializers.py
```diff
@@ -975,11 +975,15 @@ class Meta:
 
 
 class TypeaheadCourseRunSearchSerializer(serializers.Serializer):
-    org = serializers.CharField()
+    orgs = serializers.SerializerMethodField()
     title = serializers.CharField()
     key = serializers.CharField()
     marketing_url = serializers.CharField()
 
+    def get_orgs(self, result):
```
rlucioni, line 982: `TypeaheadCourseRunSearchSerializer` and `TypeaheadProgramSearchSerializer` are looking very similar. You may want to subclass a new `TypeaheadBaseSearchSerializer` where you can move shared fields like `orgs` and `title`, and shared methods, like `get_orgs()`.

course_discovery/apps/api/tests/test_serializers.py

course_discovery/apps/api/serializers.py
MatthewPiatetsky: good idea, changed

course_discovery/apps/api/serializers.py
```diff
@@ -974,26 +974,26 @@ class Meta:
         }
 
 
-class TypeaheadCourseRunSearchSerializer(serializers.Serializer):
-    org = serializers.CharField()
+class TypeaheadBaseSearchSerializer(serializers.Serializer):
```
rlucioni, line 977: Nice. You could take this a step further and define a metaclass here which can be extended by the child classes, like we do for the [minimal serializers](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L287-L290).

course_discovery/apps/api/serializers.py
```diff
@@ -974,26 +974,26 @@ class Meta:
...
+    orgs = serializers.SerializerMethodField()
     title = serializers.CharField()
-    key = serializers.CharField()
     marketing_url = serializers.CharField()
```
rlucioni, line 980: Asking just because I don't see it: how is the `marketing_url` field populated? I don't see it listed in `fields`.

course_discovery/apps/api/serializers.py
MatthewPiatetsky: I think in this case they only share title for the fields so using the meta class would be overly verbose

MatthewPiatetsky: good idea, fixed

course_discovery/apps/api/serializers.py
MatthewPiatetsky: I think I was a bit confused about this, http://www.django-rest-framework.org/api-guide/serializers/#specifying-fields-explicitly, seems like if you specify the fields explicitly they don't need to be in the Meta class. Anyway I have all the fields defined both ways now

mikedikan (approved):
:shipit:

course_discovery/apps/api/serializers.py
rlucioni: @MatthewPiatetsky judging from the DRF serializer [docs](http://www.django-rest-framework.org/api-guide/serializers/#serializers), fields specified in the metaclass &mdash; and the metaclass itself for that matter &mdash; only seem to be relevant for the `ModelSerializer`. You should be able to drop the metaclass altogether here.

rlucioni (approved):
@MatthewPiatetsky I left one last comment about cleaning up the metaclasses, 👍 otherwise. Sorry about the back and forth, just making sure we're keeping things clean!

PR #491, opened 2016-12-15
Added edit permission in OrganizationGroup.

waheedahmed:
[ECOM-6073](https://openedx.atlassian.net/browse/ECOM-6073)

As a publisher user, I would like to get into the edit mode of a course-run so that I can make changes.

**AC:**
1. Validate that this button is only visible to those who have edit access of this course-run.
2. Validate that when clicking on the "Edit" button, the user is directed to the edit page.

**Screenshot:**
screen shot 2016-12-15 at 12 47 53 pm


waheedahmed:
@awais786, @tasawernawaz Please review this PR.

awais786:
How this permission will be assign to the user ? 
Please add some steps here.

course_discovery/templates/publisher/course_run_detail/_approval_widget.html
```diff
@@ -1,10 +1,14 @@
 {% load i18n %}
 
 <div class="approval-widget">
-    <a href="{% url 'publisher:publisher_course_runs_edit' pk=object.id %}" class="btn btn-neutral btn-course-edit">{% trans "EDIT" %}</a>
+    {% if can_edit %}
+        <a href="{% url 'publisher:publisher_course_runs_edit' pk=object.id %}" class="btn btn-neutral btn-courserun-edit">
```
awais786, line 5: If another user try to open this link directly into the browser then how it behave ?

awais786: I think we need to add the permission check on the edit view also.

waheedahmed: @awais786 Its not part of this story.

waheedahmed:
@awais786 We are going to assign the permissions with migration maybe, but again its not part of this story.

course_discovery/templates/publisher/course_run_detail/_approval_widget.html
awais786: I think it should be part of this story because when you redirect the user only authorized users can view that page.

waheedahmed: @awais786 Why, it will be handled when we create an edit page.

waheedahmed: @awais786 We are going to check that in edit page implementation not here, this ticket is only for edit button not for edit page.

awais786 (approved):

waheedahmed:

course_discovery/apps/publisher/views.py
```diff
@@ -127,12 +127,17 @@ def get_role_widgets_data(self, course_roles):
     def get_context_data(self, **kwargs):
         context = super(CourseRunDetailView, self).get_context_data(**kwargs)
 
+        user = self.request.user
         course_run = CourseRunWrapper(self.get_object())
         context['object'] = course_run
         context['comment_object'] = course_run.course
+        context['can_edit'] = any(
```
bderusha, line 134: This looks good.  Let's make sure we think about whether we will need to extract this into some helper method when we are doing checks on other permissions.  No need to change this, just keep in mind when we add more.

course_discovery/templates/publisher/course_run_detail/_approval_widget.html
bderusha: This is fine for this story.  Let's keep the scope small to get it over the line.

bderusha (approved):

PR #494, opened 2016-12-15
Filter typeahead results by partner

MatthewPiatetsky:
@edx/ecommerce 

Jasper found one last thing 😿  , he says he can't find any other issues.
I realized I didn't notice this locally because I only had the edx partner configured, so I didn't have white label courses showing up.

Marketing site is already passing the partner param, so changing that shouldn't be necessary.
I think it works for search because the querystring was getting filtered directly with ES.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -17,7 +17,9 @@ from course_discovery.apps.core.tests.factories import UserFactory, USER_PASSWOR
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import CourseRun, Program, ProgramType
-from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, ProgramFactory, OrganizationFactory
+from course_discovery.apps.course_metadata.tests.factories import (
+    CourseFactory, CourseRunFactory, ProgramFactory, OrganizationFactory
+)
 from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
 
 
@@ -316,7 +318,8 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
         self.assertEqual(response.data['objects']['results'], expected)
 
 
-class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase):
+class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin, LoginMixin, ElasticsearchTestMixin,
+                               APITestCase):
     path = reverse('api:v1:search-typeahead')
     function_score = {
         'functions': [
@@ -327,11 +330,11 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, Elastics
         'boost': 1.0, 'score_mode': 'sum', 'boost_mode': 'sum',
     }
 
-    def get_typeahead_response(self, query=None):
+    def get_typeahead_response(self, query=None, partner=None):
         qs = ''
-
+        query_dict = {'q': query, 'partner': partner or self.partner.short_code}
         if query:
-            qs = urllib.parse.urlencode({'q': query})
+            qs = urllib.parse.urlencode(query_dict)
 
         url = '{path}?{qs}'.format(path=self.path, qs=qs)
         config = ElasticsearchBoostConfig.get_solo()
@@ -342,8 +345,8 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, Elastics
     def test_typeahead(self):
         """ Test typeahead response. """
         title = "Python"
-        course_run = CourseRunFactory(title=title)
-        program = ProgramFactory(title=title, status=ProgramStatus.Active)
+        course_run = CourseRunFactory(title=title, course__partner=self.partner)
+        program = ProgramFactory(title=title, status=ProgramStatus.Active, partner=self.partner)
         response = self.get_typeahead_response(title)
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
@@ -355,8 +358,8 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, Elastics
         RESULT_COUNT = TypeaheadSearchView.RESULT_COUNT
         title = "Test"
         for i in range(RESULT_COUNT + 1):
-            CourseRunFactory(title="{}{}".format(title, i))
-            ProgramFactory(title="{}{}".format(title, i), status=ProgramStatus.Active)
+            CourseRunFactory(title="{}{}".format(title, i), course__partner=self.partner)
+            ProgramFactory(title="{}{}".format(title, i), status=ProgramStatus.Active, partner=self.partner)
         response = self.get_typeahead_response(title)
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
@@ -367,9 +370,14 @@ class TypeaheadSearchViewTests(TypeaheadSerializationMixin, LoginMixin, Elastics
         """ Test typeahead response with multiple authoring organizations. """
         title = "Design"
         authoring_organizations = OrganizationFactory.create_batch(3)
-        course_run = CourseRunFactory(title=title, authoring_organizations=authoring_organizations)
+        course_run = CourseRunFactory(
+            title=title,
+            authoring_organizations=authoring_organizations,
+            course__partner=self.partner
+        )
         program = ProgramFactory(
-            title=title, authoring_organizations=authoring_organizations, status=ProgramStatus.Active
+            title=title, authoring_organizations=authoring_organizations,
+            status=ProgramStatus.Active, partner=self.partner
         )
         response = self.get_typeahead_response(title)
         self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/search.py
```diff
@@ -138,13 +138,18 @@ def get_results(self, query):
...
         course_runs = course_runs.filter(published=True).exclude(hidden=True)
-        course_runs = course_runs[:self.RESULT_COUNT]
 
         programs = sqs.models(Program).filter(
             SQ(title_autocomplete=clean_query) |
             SQ(authoring_organizations_autocomplete=clean_query)
         )
         programs = programs.filter(status=ProgramStatus.Active)
+
+        if partner:
```
rlucioni: This implies `partner` is optional. If that's the case, it should be a kwarg defaulted to None.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -495,3 +500,21 @@ def test_typeahead_org_course_runs_come_up_first(self):
                          self.serialize_program(harvard_program)]
         }
         self.assertDictEqual(response.data, expected)
+
+    def test_typeahead_partner_filter(self):
+        """ Ensure that a partner param limits results to that partner. """
+        title = "Partner"
+
+        edx = PartnerFactory(short_code="edx")
```
rlucioni: You could make this more concise:
```py
for index, partner in enumerate(['primary', 'secondary']):
    title = 'Belongs to partner ' + index
    CourseRunFactory(title=title, course=CourseFactory(partner=partner))
    ProgramFactory(title=title, partner=partner, status=ProgramStatus.Active)
```
Then query for the ones you need.

rlucioni (changes requested):

course_discovery/apps/api/v1/views/search.py
```diff
@@ -167,9 +172,10 @@ def get(self, request, *args, **kwargs):
               type: string
         """
         query = request.query_params.get('q')
+        partner = request.query_params.get('partner')
```
bderusha: Per @rlucioni 's comment above, should this also raise a parseError if it doesnt exist?

bderusha (changes requested):

course_discovery/apps/api/v1/views/search.py
mikedikan: I don't know if this needs to be changed if this is a private method @rlucioni - but in that case, it should start with _, right?

MatthewPiatetsky: good question, imo partner should be optional, if this will be open source maybe people will want to see multiple partners.

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: i thought partner should be optional but if you think it should be required i'll do that

mikedikan: i vote for keeping it optional and if it must be, defined as a kwarg with default value, but i don't see why it needs to be a default if get_results isn't exposed

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: good idea, fixed

course_discovery/apps/api/v1/views/search.py
```diff
@@ -167,9 +172,10 @@ def get(self, request, *args, **kwargs):
...
         if not query:
             raise ParseError("The 'q' querystring parameter is required for searching.")
-        course_runs, programs = self.get_results(query)
+        course_runs, programs = self.get_results(query, partner)
```
rlucioni, line 179: This technically works, but it's a kwarg now, so call it like one.

rlucioni (approved):

bderusha (approved):
:shipit:

course_discovery/apps/api/v1/views/search.py
clintonb: Partner should not be optional. We may have a sane default; but, assume we are only returning data for a single tenant/partner. At some point we will implement proper multi-tenancy, as we have done with our other services.

MatthewPiatetsky: so should I make partner required and raise an error if it is not passed?

clintonb: See https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views/course_runs.py#L30-L41. You don't need to require the parameter, for now. Use the default partner, and allow an override.

MatthewPiatetsky: are the changes I made what you were looking for?

course_discovery/apps/api/v1/views/__init__.py
```diff
@@ -43,3 +46,18 @@ def prefetch_related_objects_for_courses(queryset):
     queryset = queryset.select_related(*_select_related_fields['course'])
     queryset = queryset.prefetch_related(*_prefetch_fields['course'])
     return queryset
+
+
+class PartnerParamMixin:
+    def _get_partner(self, request):
+        """ Return the partner for the code passed in or the default partner """
```
clintonb: s/code/short code

course_discovery/apps/api/v1/views/__init__.py
```diff
@@ -43,3 +46,18 @@ def prefetch_related_objects_for_courses(queryset):
     queryset = queryset.select_related(*_select_related_fields['course'])
     queryset = queryset.prefetch_related(*_prefetch_fields['course'])
     return queryset
+
+
+class PartnerParamMixin:
```
clintonb: `PartnerMixin`

course_discovery/apps/api/v1/views/__init__.py
```diff
@@ -43,3 +46,18 @@ def prefetch_related_objects_for_courses(queryset):
...
+    def _get_partner(self, request):
```
clintonb: `get_partner`

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -52,7 +36,7 @@ def get_queryset(self):
               multiple: false
         """
         q = self.request.query_params.get('q', None)
-        partner = self._get_partner()
+        partner = self._get_partner(request=self.request)
```
clintonb: You can drop the kwarg here, and below.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -165,11 +166,19 @@ def get(self, request, *args, **kwargs):
               paramType: query
               required: true
               type: string
+            - name: partner
+              description: "Partner"
```
clintonb: Partner short code

course_discovery/apps/api/v1/views/search.py
```diff
@@ -165,11 +166,19 @@ def get(self, request, *args, **kwargs):
...
+              paramType: query
+              required: false
+              type: string
         """
         query = request.query_params.get('q')
+        partner = self._get_partner(request=request)
         if not query:
             raise ParseError("The 'q' querystring parameter is required for searching.")
-        course_runs, programs = self.get_results(query)
+        if not partner:
```
clintonb: This will never be true, and should be removed.

course_discovery/apps/api/v1/views/__init__.py
```diff
@@ -43,3 +46,18 @@ def prefetch_related_objects_for_courses(queryset):
...
+        """ Return the partner for the code passed in or the default partner """
+        partner_code = request.query_params.get('partner')
+        if partner_code:
+            try:
+                partner = Partner.objects.get(short_code=partner_code)
+            except Partner.DoesNotExist:
+                raise InvalidPartnerError('Unknown Partner')
```
clintonb: `'Unknown partner: {}'.format(partner_code)`

course_discovery/apps/api/v1/views/search.py
```diff
@@ -165,11 +166,19 @@ def get(self, request, *args, **kwargs):
...
+            - name: partner
+              description: "Partner"
+              paramType: query
+              required: false
+              type: string
         """
         query = request.query_params.get('q')
+        partner = self._get_partner(request=request)
         if not query:
             raise ParseError("The 'q' querystring parameter is required for searching.")
```
clintonb: This is more of a [`ValidationError`](http://www.django-rest-framework.org/api-guide/exceptions/#validationerror) than `ParseError`.

clintonb (approved):

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: ok, though DEFAULT_PARTNER_ID could theoretically not be set, right?

course_discovery/apps/api/v1/views/__init__.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/views/__init__.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/views/__init__.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/views/__init__.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/views/course_runs.py
MatthewPiatetsky: fixed

course_discovery/apps/api/v1/views/search.py
clintonb: It is _always_ set. Currently it is set to the edX partner.

PR #496, opened 2016-12-16
Added searchable courseruns filter

ahsan-ul-haq:
Added searchable filter for course runs so we can use to send data to the sailthru. Related to https://github.com/edx/ecommerce-scripts/pull/34

@clintonb @mikedikan please review

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -150,7 +150,7 @@ class CourseRunFilter(FilterSetMixin, django_filters.FilterSet):
 
     class Meta:
         model = CourseRun
-        fields = ['keys']
+        fields = ['keys', 'hidden']
 
 
 class ProgramFilter(FilterSetMixin, django_filters.FilterSet):
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -173,6 +173,16 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         url = reverse('api:v1:course_run-list') + '?marketable=1'
         self.assert_list_results(url, expected)
 
+    def test_filter_by_hidden(self):
+        """ Verify the endpoint filters course runs that are hidden. """
+        CourseRun.objects.all().delete()
+        course_runs = CourseRunFactory.create_batch(3, course__partner=self.partner)
+        hidden_course_runs = CourseRunFactory.create_batch(3, hidden=True, course__partner=self.partner)
+        url = reverse('api:v1:course_run-list')
+        self.assert_list_results(url, course_runs + hidden_course_runs)
+        url = reverse('api:v1:course_run-list') + '?hidden=False'
+        self.assert_list_results(url, course_runs)
+
     def test_filter_by_active(self):
         """ Verify the endpoint filters course runs to those that are active. """
         CourseRun.objects.all().delete()
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -81,6 +81,12 @@ class CourseRunViewSet(PartnerMixin, viewsets.ReadOnlyModelViewSet):
               type: string
               paramType: query
               multiple: false
+            - name: hidden
+              description: Filter based on wether the course run is hidden from search.
+              required: false
+              type: Boolean
+              paramType: query
+              multiple: false
             - name: active
               description: Retrieve active course runs. A course is considered active if its end date has not passed,
                 and it is open for enrollment.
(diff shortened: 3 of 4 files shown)


course_discovery/apps/course_metadata/query.py
```diff
@@ -54,6 +54,17 @@ def marketable(self):
 
         return self.exclude(slug__isnull=True).exclude(slug='').filter(status=CourseRunStatus.Published)
 
+    def searchable(self):
```
clintonb: Remove this.

course_discovery/apps/api/filters.py
```diff
@@ -137,6 +140,7 @@ class Meta:
 class CourseRunFilter(FilterSetMixin, django_filters.FilterSet):
     active = django_filters.MethodFilter()
     marketable = django_filters.MethodFilter()
+    searchable = django_filters.MethodFilter()
```
clintonb: This should be a filter on the hidden flag. See https://django-filter.readthedocs.io/en/latest/. We should prefer to use composition—`?active=1&marketable=1&hidden=0`—when possible rather than completely customized filters. The customization in the case of `active` and `marketable` is only there because those filters are complex and potentially the most-requested.

In this case, we only need to exclude hidden courses. There is no reason to add additional complexity.

clintonb (changes requested):

ahsan-ul-haq:
@clintonb please review

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -173,6 +173,14 @@ def test_filter_by_marketable(self):
         url = reverse('api:v1:course_run-list') + '?marketable=1'
         self.assert_list_results(url, expected)
 
+    def test_filter_by_hidden(self):
```
mikedikan, line 176: Please add tests for if flag isn't present in request, to make sure hidden runs are still returned - this default behavior needs to be maintained.

mikedikan (changes requested):

ahsan-ul-haq:
@mikedikan I have updated the test

ahsan-ul-haq:
@rock345

ahsan-ul-haq:
@clintonb @mikedikan please review

mikedikan (approved):
Thanks!

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -37,6 +37,7 @@ def get_queryset(self):
         """
         q = self.request.query_params.get('q', None)
         partner = self.get_partner()
+        hidden = self.request.query_params.get('hidden', None)
```
clintonb: Update CourseRunFilter. We should not be writing custom filters when we have a library that will do this for us.

clintonb (changes requested):

ahsan-ul-haq:
@clintonb made the suggested change

course_discovery/apps/api/filters.py
```diff
@@ -138,6 +138,7 @@ class CourseRunFilter(FilterSetMixin, django_filters.FilterSet):
     active = django_filters.MethodFilter()
     marketable = django_filters.MethodFilter()
     keys = CharListFilter(name='key', lookup_expr='in')
+    hidden = django_filters.BooleanFilter(name='hidden')
```
clintonb: I don't think this line is necessary.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -173,6 +173,17 @@ def test_filter_by_marketable(self):
...
+        """ Verify the endpoint filters course runs to those that are searchable. """
```
clintonb: Correct this.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -173,6 +173,17 @@ def test_filter_by_marketable(self):
...
+        CourseRun.objects.all().delete()
+        expected = CourseRunFactory.create_batch(3, course__partner=self.partner)
+        hidden_course_runs = CourseRunFactory.create_batch(3, hidden=True, course__partner=self.partner)
+        url = reverse('api:v1:course_run-list') + '?hidden=False'
```
clintonb: If you invert your test order (test the default case first) this will look less like a wall of text. A beeline between the setup (variable creation), and each condition will also help.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -81,6 +81,12 @@ def list(self, request, *args, **kwargs):
               type: string
               paramType: query
               multiple: false
+            - name: hidden
+              description: Filter by hidden
```
clintonb: Filter based on wether the course run is hidden from search.

clintonb (approved):

PR #499, opened 2016-12-19
ECOM-5562 Update program alias in drupal when saving program

MatthewPiatetsky:
@edx/ecommerce


clintonb:
Something went wrong with the language update here. You should not be changing every single language.

MatthewPiatetsky:
yeah I put it in separate commit because it seemed odd, if I remove that what are the correct commands I should run?

clintonb:
See https://openedx.atlassian.net/wiki/display/ECOM/Automated+Translation+Management.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -157,6 +160,47 @@ def _delete_node(self, api_client, nid):
         node_url = '{root}/node.json/{nid}'.format(root=api_client.api_url, nid=nid)
         api_client.api_session.delete(node_url)
 
+    def _get_form_build_id_and_form_token(self, api_client, add_aliases_url, headers):
+        form_attributes = {}
+        response = api_client.api_session.get(add_aliases_url, headers=headers)
+        if response.status_code != 200:
+            raise ProgramPublisherException("Marketing site alias form retrieval failed!")
```
clintonb: Single quotes.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -157,6 +160,47 @@ def _delete_node(self, api_client, nid):
...
+            'alias': alias,
+            'form_id': 'path_admin_form',
+            'op': 'Save'
+        }
+        headers = self._get_headers(api_client.api_session)
+        form_attributes = self._get_form_build_id_and_form_token(api_client, add_aliases_url, headers)
+        data.update(form_attributes)
+
+        response = api_client.api_session.post(add_aliases_url, headers=headers, data=data)
+        if response.status_code != 200:
```
clintonb, line 200: What happens if the alias already exists?

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -157,6 +160,47 @@ def _delete_node(self, api_client, nid):
...
+        node_url = 'node/{nid}'.format(nid=nid)
+        alias = '{program_type}/{slug}'.format(program_type=program.type.name.lower(), slug=program.marketing_slug)
+
+        data = {
+            'source': node_url,
+            'alias': alias,
+            'form_id': 'path_admin_form',
+            'op': 'Save'
+        }
+        headers = self._get_headers(api_client.api_session)
```
clintonb: This isn't necessary. The cookies are already in the session object. I suspect Drupal will interpret the content type correctly.

rlucioni:
@MatthewPiatetsky to add to what @clintonb said: it looks like you're trying to update the source translation file, but you downloaded translated files from Transifex instead. See the first section in the link @clintonb posted, "Keeping master's source translation files current."

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -157,6 +160,47 @@ def _delete_node(self, api_client, nid):
         node_url = '{root}/node.json/{nid}'.format(root=api_client.api_url, nid=nid)
         api_client.api_session.delete(node_url)
 
+    def _get_form_build_id_and_form_token(self, api_client, add_aliases_url, headers):
+        form_attributes = {}
+        response = api_client.api_session.get(add_aliases_url, headers=headers)
+        if response.status_code != 200:
+            raise ProgramPublisherException("Marketing site alias form retrieval failed!")
+        form = response.text.split('\n')
```
clintonb: You seem to be doing a lot of what Beautiful Soup is supposed to do for us—parsing HTML.

```python
from bs4 import BeautifulSoup

form = BeautifulSoup(response.text, 'html.parser')

for field in ('form_build_id', 'form_token'):
  form_attributes[field] = soup.find('input', {'name': field}).get('value')
```

requirements/base.txt
```diff
@@ -1,4 +1,5 @@
 boto==2.42.0
+bs4==0.0.1
```
clintonb: `beautifulsoup4==4.5.1`

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -157,6 +160,47 @@ def _delete_node(self, api_client, nid):
...
+        headers = {
+            'cookie': cookie_string,
+            'content-type': 'application/x-www-form-urlencoded'
+        }
+        return headers
+
+    def _add_alias(self, api_client, program, nid):
+        add_aliases_url = '{root}/admin/config/search/path/add'.format(root=api_client.api_url)
+        node_url = 'node/{nid}'.format(nid=nid)
+        alias = '{program_type}/{slug}'.format(program_type=program.type.name.lower(), slug=program.marketing_slug)
```
clintonb: You will need to add a slug field to the `ProgramType` model. Otherwise, you will have issues with "Professional Certificate".

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -168,6 +212,7 @@ def publish_program(self, program):
             else:
                 # We should create a new node
                 self._create_node(api_client, node_data)
+            self._add_alias(api_client, program, nid)
```
clintonb: Please rename `nid` to `node_id` here and elsewhere to avoid future confusion.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -1,6 +1,7 @@
 import json
 import requests
```
clintonb: This is a third-party library, and should be grouped accordingly.

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: fixed

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: thanks, changed

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: added

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: yeah not sure why I had the cookies, but I do need the content type for it to work, maybe because the headers property on the client specifies application/json

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: changed

requirements/base.txt
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: I've added a check to only add the alias if the alias changed. 
The slug field in the model being unique should mean there aren't duplicate aliases. 
This wouldn't necessarily cover the case of changing an alias and changing it back. If you're trying to create an alias for the same object it will delete the old one
screen shot 2016-12-20 at 1 46 43 pm

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: changed

MatthewPiatetsky:
@clintonb @rlucioni I removed the translations changes and ran make fake_translations and got this again. It seems like something with the date may be changing?

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -98,12 +103,13 @@ def _get_api_client(self, program):
                 partner=program.partner.short_code)
             raise ProgramPublisherException(msg)
 
-        if program.type.name != 'MicroMasters':
+        if program.type.name not in ['MicroMasters', 'Professional Certificate']:
             # Currently, we should not publish any programs that are not MicroMasters types to Marketing Site
```
mikedikan: update comment

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -133,14 +139,12 @@ def _get_node_id(self, api_client, uuid):
...
-
-    def _edit_node(self, api_client, nid, node_data):
-        if node_data.get('uuid'):
-            # Drupal do not allow us to update the UUID field on node update
-            del node_data['uuid']
-        node_url = '{root}/node.json/{nid}'.format(root=api_client.api_url, nid=nid)
+                return list_item[0]['nid']
+
+    def _edit_node(self, api_client, node_id, node_data):
+        # Drupal do not allow us to update the UUID field on node update
```
mikedikan: do->does

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -153,25 +157,63 @@ def _create_node(self, api_client, node_data):
         else:
             raise ProgramPublisherException("Marketing site page creation failed!")
 
-    def _delete_node(self, api_client, nid):
-        node_url = '{root}/node.json/{nid}'.format(root=api_client.api_url, nid=nid)
+    def _delete_node(self, api_client, node_id):
+        node_url = '{root}/node.json/{node_id}'.format(root=api_client.api_url, node_id=node_id)
```
mikedikan, line 154: 👍

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -153,25 +157,63 @@ def _create_node(self, api_client, node_data):
...
+            'alias': alias,
+            'form_id': 'path_admin_form',
+            'op': 'Save'
+        }
+        form_attributes = self._get_form_build_id_and_form_token(api_client, add_aliases_url)
+        data.update(form_attributes)
+        headers = self._get_headers()
+        response = api_client.api_session.post(add_aliases_url, headers=headers, data=data)
+        if response.status_code != 200:
+            raise ProgramPublisherException('Marketing site alias creation failed!')
```
mikedikan, line 201: did you test what happens if you change alias and then change back to original?

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -208,8 +225,23 @@ def test_publish_modified_program(self):
...
+                self.assert_responses_call_count(8)
+
+    @responses.activate
+    def test_add_alias_failed(self):
+        self.mock_api_client(200)
+        self.mock_node_retrieval(self.program.uuid)
+        self.mock_node_edit(200)
+        program_before = ProgramFactory()
+        publisher = MarketingSitePublisher(program_before)
+        self.mock_add_alias(500)
```
mikedikan, line 294: does drupal return a 500 if you try to add an existing alias?  does it update the alias?  does it leave it alone?

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -153,25 +157,63 @@ def _create_node(self, api_client, node_data):
...
+        return form_attributes
+
+    def _get_headers(self):
+        headers = {
+            'content-type': 'application/x-www-form-urlencoded'
+        }
+        return headers
+
+    def _add_alias(self, api_client, program, node_id):
+        add_aliases_url = '{root}/admin/config/search/path/add'.format(root=api_client.api_url)
```
mikedikan: if adding alias fails, should we message the user to go to the drupal site to enter alias manually?

mikedikan (changes requested):

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: I don't know, should we?
You could say the same thing for the rest of the publishing process (currently it just fails).
Is this a product question?

course_discovery/apps/course_metadata/tests/test_publishers.py
MatthewPiatetsky: I believe if you try to add an alias for the same node and it already has one it does this
https://github.com/edx/course-discovery/pull/499#discussion_r93299759

course_discovery/apps/course_metadata/publishers.py
mikedikan: i think it is ultimately a product question - do you want to ping jasper?

course_discovery/apps/course_metadata/tests/test_publishers.py
mikedikan: got it, thanks.  so under what circumstances will we see a 500?  what if someone added an alias in drupal and not in django (pointing to a different node)

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: ok I asked him, I think I will update the error messages to say that you can update drupal + let ecom know

course_discovery/apps/course_metadata/tests/test_publishers.py
MatthewPiatetsky: In this particular case I just added a test to cover the exceptions, I wasn't thinking of a particular case. 
There are a whole bunch of possible cases we could come up with, do we really want to cover all of them?
My guess would be that we should only cover the cases of what people can do through django. In this case the marketing slug will be required and unique, so duplications should not occur.

mikedikan: i guess the scenario i'm trying to illustrate is if the person attempts to enter an alias that collides with something else in drupal (maybe not even a program)  Is this not possible because all of the aliases are namespaced to start with /program_type?

course_discovery/apps/course_metadata/publishers.py
MatthewPiatetsky: actually nvm I don't think we want to do this, the whole publishing program process is atomic, we don't want to update drupal manually if django failed to update

course_discovery/apps/course_metadata/tests/test_publishers.py
MatthewPiatetsky: Yes that is correct regarding program type, so I don't think this would happen unless someone manually created an alias such as micromasters/columbiax-artificial-intelligence for something other than that (which seems unlikely)

mikedikan: got it, thanks

mikedikan (approved):
lgtm

rlucioni:
The `.mo` files on this PR are the result of running Django's [`compilemessages`](https://docs.djangoproject.com/en/1.9/ref/django-admin/#compilemessages) on new `.po` files introduced by our translation management bot. The script we use updates `.po` files, but doesn't generate or commit freshly compiled `.mo` files. (Travis compiles messages when running status checks against the "update translations" PR, but those `.mo` files aren't included in the PR.)

To clean up this PR, I would create a separate commit/PR for just the `.mo` files. I will update the script we use so that `.mo` are generated, so this doesn't happen again.

MatthewPiatetsky:
@rlucioni good idea, merged pr with fake_translations off master and updated this pr

course_discovery/apps/course_metadata/models.py
```diff
@@ -610,7 +612,7 @@ class Program(TimeStampedModel):
         choices=ProgramStatus.choices, validators=[ProgramStatus.validator]
     )
     marketing_slug = models.CharField(
-        help_text=_('Slug used to generate links to the marketing site'), blank=True, max_length=255, db_index=True)
+        help_text=_('Slug used to generate links to the marketing site'), unique=True, max_length=255, db_index=True)
```
rlucioni, line 615: Have you checked to see if this uniqueness constraint is going to clash with existing marketing slugs? In other words, are there any duplicates today?

course_discovery/apps/course_metadata/models.py
```diff
@@ -593,6 +593,8 @@ class ProgramType(TimeStampedModel):
         },
         help_text=_('Please provide an image file with transparent background'),
     )
+    slug = AutoSlugField(populate_from='name', editable=True, blank=True,
```
rlucioni, line 596: Will this populate a slug at migration-time, or do we need to go in and fill them in manually?

MatthewPiatetsky: I'm pretty sure there aren't currently any collisions, though I'm not sure how I could properly test the migration because I don't have stage or prod programs locally..maybe I can do some sort of sql dump from the read replicas and load it locally

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: I'm pretty sure there aren't currently any collisions, though I'm not sure how I could properly test the migration because I don't have stage or prod programs locally..maybe I can do some sort of sql dump from the read replicas and load it locally

rlucioni: You could start by checking to see if there are any duplicates in the read replica.

MatthewPiatetsky: there are not, (though i had to look manually for staging because the staging read replica script doesn't seem to be working)

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: We will need to fill them in manually, but there are only 3, so I can do this

rlucioni (approved):
Please update your commit message to include at least the ticket number. A small description of the changes contained within would be nice, too, as is our standard.

MatthewPiatetsky:
i think this branch is ready, I will do some more manual testing before merging, but won't merge until we're back from the holiday in case anything breaks

PR #501, opened 2016-12-20
MicroMasters logo showing on a non-MM course discovery card

ahsan-ul-haq:
ECOM-6372

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -200,6 +200,46 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
         """ Verify the unpublished programs do not show in the program_types representation. """
         self._test_exclude_program_types(ProgramStatus.Unpublished)
 
+    @ddt.data(
+        [{'title': 'Software Testing', 'excluded': True}],
+        [{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True}],
+        [{'title': 'Software Testing', 'excluded': False}, {'title': 'Software Testing 2', 'excluded': False}],
+        [{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True},
+         {'title': 'Software Testing 3', 'excluded': False}],
+    )
+    def test_excluded_course_run(self, course_runs):
+        course_list = []
+        course_run_list = []
+        excluded_course_run_list = []
+        non_excluded_course_run_list = []
+        for run in course_runs:
+            course_run = CourseRunFactory(course__partner=self.partner, course__title=run['title'],
+                                          status=CourseRunStatus.Published)
+            course_list.append(course_run.course)
+            course_run_list.append(course_run)
+            if run['excluded']:
+                excluded_course_run_list.append(course_run)
+            else:
+                non_excluded_course_run_list.append(course_run)
+
+        ProgramFactory(courses=course_list, status=ProgramStatus.Active, excluded_course_runs=excluded_course_run_list)
+
+        with self.assertNumQueries(6):
+            response = self.get_search_response('software', faceted=False)
+
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+
+        self.assertEqual(response_data['count'], len(course_run_list))
+        for result in response_data['results']:
+            for course_run in excluded_course_run_list:
+                if result.get('title') == course_run.title:
+                    self.assertEqual(result.get('program_types'), [])
+
+            for course_run in non_excluded_course_run_list:
+                if result.get('title') == course_run.title:
+                    self.assertEqual(result.get('program_types'), course_run.program_types)
+
     def _test_exclude_program_types(self, program_status):
         """ Verify that programs with the provided type do not show in the program_types representation. """
         course_run = CourseRunFactory(course__partner=self.partner, course__title='Software Testing',
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -388,8 +388,11 @@ class CourseRun(TimeStampedModel):
         so we don't identify that program type if not available
         """
         program_statuses_to_exclude = (ProgramStatus.Unpublished, ProgramStatus.Deleted)
-        all_programs = [program for program in self.programs.exclude(status__in=program_statuses_to_exclude)]
-        return [program.type.name for program in all_programs]
+        associated_programs = []
+        for program in self.programs.exclude(status__in=program_statuses_to_exclude):
+            if self not in program.excluded_course_runs.all():
+                associated_programs.append(program)
+        return [program.type.name for program in associated_programs]
 
     @property
     def marketing_url(self):


ahsan-ul-haq:
@MatthewPiatetsky please review.

course_discovery/apps/course_metadata/models.py
```diff
@@ -439,7 +439,8 @@ def prerequisites(self):
 
     @property
     def programs(self):
-        return self.course.programs  # pylint: disable=no-member
+        if not self.course.programs.filter(excluded_course_runs=self):
```
MatthewPiatetsky: you can assign self.course.programs to a variable

ahsan-ul-haq:
@MatthewPiatetsky made suggested change

ahsan-ul-haq:
@MatthewPiatetsky please review

MatthewPiatetsky:
can you use django toolbar to see if this is actually increasing the number of SQL queries? let me know if you don't know how to do this

ahsan-ul-haq:
@MatthewPiatetsky, unfortunately, I don't know how to do this.

MatthewPiatetsky:
@ahsan-ul-haq 
you run ENABLE_DJANGO_TOOLBAR=1 ./manage.py runserver localhost:8008
and look at the number of SQL queries in the sidebar
I looked at this branch briefly locally and it seemed like your change did not introduce additional queries that were not prefetched, despite the changes you needed in your test.
might be good for @rlucioni to check this when he gets back. in the meantime you can get another review.

👍

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -198,6 +198,26 @@ def test_exclude_unpublished_program_types(self):
...
+
+        # Validate the search results
+        expected = {
+            'count': 1,
+            'results': [
+                self.serialize_course_run(course_run)
+            ]
+        }
+        self.assertDictContainsSubset(expected, response_data)
+        self.assertEqual(response_data['results'][0].get('program_types'), [])
```
rlucioni: What's the purpose of this assertion?

course_discovery/apps/course_metadata/models.py
```diff
@@ -439,7 +439,9 @@ def prerequisites(self):
 
     @property
     def programs(self):
-        return self.course.programs  # pylint: disable=no-member
+        programs = self.course.programs  # pylint: disable=no-member
+        if not programs.filter(excluded_course_runs=self):
```
rlucioni: This looks wrong. What if the same course is part of two separate programs, and this course run is only excluded from one? This will filter those programs, yield the one program from which the run is excluded, and then return None, even though there is one program from which the run is not excluded (which we would expect to be returned).

rlucioni (changes requested):
Not only does this change incur many new queries on the course runs endpoint (from 22 to 61 queries, with two programs, each with two courses, each with one run), it doesn't even appear to be necessary. The filtering logic you're trying to introduce here works already, which you can verify by setting up programs as I've described above.

rlucioni:
My earlier comment was a little premature. This excluded course run filtering is implemented by the [`CourseRunWithProgramsSerializer`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L387), which is why the logic at /api/v1/course_runs is already correct. I realize now that this is meant to be a fix for search. Regardless, the change proposed here introduces way too many additional queries on that endpoint. Try reproducing the query count changes locally using the debug toolbar.

schenedx:
@ahsan-ul-haq: @rlucioni is right And you are not writing the exclusion logic at the right place.
It should be excluded at the Program_types property at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L382

Also, you should evaluate the query count increase to figure out if we can above the query count increase by using the Django_Toolbar. See instructions provided by @MatthewPiatetsky

schenedx (changes requested):
The update should be at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L382 instead

course_discovery/apps/course_metadata/models.py
```diff
@@ -380,7 +380,10 @@ def program_types(self):
         so we don't identify that program type if not available
         """
         program_statuses_to_exclude = (ProgramStatus.Unpublished, ProgramStatus.Deleted)
-        all_programs = [program for program in self.programs.exclude(status__in=program_statuses_to_exclude)]
+        all_programs = []
```
schenedx: A better name for this variable is `associated_programs`

course_discovery/apps/api/v1/tests/test_views/test_search.py
schenedx: This is to ensure the program excluded course run do not have any `program_types` associated

course_discovery/apps/course_metadata/models.py
```diff
@@ -380,7 +380,10 @@ def program_types(self):
...
+        for program in self.programs.exclude(status__in=program_statuses_to_exclude,):
+            if self not in program.excluded_course_runs:
```
schenedx: According to travis unit tests, you cannot do this line. The is was `TypeError: argument of type 'ManyRelatedManager' is not iterable`. Basically, you cannot apply keyword `in` with `excluded_course_runs` object because `excluded_course_runs` object is a `ManyrelatedManager`, not a iterable array.

schenedx (changes requested):
Please fix the error.

ahsan-ul-haq:
@schenedx @rlucioni made the suggested changes please take a look.

schenedx (approved):
👍

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -201,6 +201,26 @@ def test_exclude_unpublished_program_types(self):
...
 
+    def test_excluded_course_run(self):
+        course_run = CourseRunFactory(course__partner=self.partner, course__title='Software Testing',
+                                      status=CourseRunStatus.Published)
+        ProgramFactory(courses=[course_run.course], status=ProgramStatus.Active, excluded_course_runs=[course_run])
+
+        response = self.get_search_response('software', faceted=False)
+
+        self.assertEqual(response.status_code, 200)
+        response_data = json.loads(response.content.decode('utf-8'))
```
rlucioni: Use `response.json()`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -380,8 +380,11 @@ def program_types(self):
         so we don't identify that program type if not available
         """
         program_statuses_to_exclude = (ProgramStatus.Unpublished, ProgramStatus.Deleted)
-        all_programs = [program for program in self.programs.exclude(status__in=program_statuses_to_exclude)]
-        return [program.type.name for program in all_programs]
+        associated_programs = []
+        for program in self.programs.exclude(status__in=program_statuses_to_exclude,):
+            if self not in program.excluded_course_runs.all():
```
rlucioni, line 393: @ahsan-ul-haq have you used the debug toolbar to look at the impact this change will have on queries made by this endpoint? I don't see any prefetching being done here, so I'm concerned that this will introduce many new queries that will impact the performance of endpoints using this property.

ahsan-ul-haq:
@rlucioni please review

course_discovery/apps/course_metadata/models.py
```diff
@@ -380,8 +380,11 @@ def program_types(self):
         so we don't identify that program type if not available
         """
         program_statuses_to_exclude = (ProgramStatus.Unpublished, ProgramStatus.Deleted)
-        all_programs = [program for program in self.programs.exclude(status__in=program_statuses_to_exclude)]
-        return [program.type.name for program in all_programs]
+        associated_programs = []
+        for program in self.programs.exclude(status__in=program_statuses_to_exclude,):
```
rlucioni: Nitpick: remove the trailing comma.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -201,6 +201,27 @@ def test_exclude_unpublished_program_types(self):
         """ Verify the unpublished programs do not show in the program_types representation. """
         self._test_exclude_program_types(ProgramStatus.Unpublished)
 
+    def test_excluded_course_run(self):
+        course_run = CourseRunFactory(course__partner=self.partner, course__title='Software Testing',
+                                      status=CourseRunStatus.Published)
+        ProgramFactory(courses=[course_run.course], status=ProgramStatus.Active, excluded_course_runs=[course_run])
+
+        with self.assertNumQueries(9):
```
rlucioni: @ahsan-ul-haq including this assertion is good, but it's not sufficient on its own. What was the query count before your change? How does it scale as the number of programs and course runs increases? We have nearly a hundred programs and thousands of course runs in production. Something small like this can have a big impact in that case.

I'll repeat what I wrote below: have you used the debug toolbar to look at the impact this change will have on queries made by this endpoint? I don't see any prefetching being done here, so I'm concerned that this will introduce many new queries that will impact the performance of endpoints using this property.

ahsan-ul-haq: @rlucioni I have checked that and it seems like there is only increase of one query per record.

rlucioni: @ahsan-ul-haq what data were you testing with? Were you using the debug toolbar or relying on this assertion? The query count change may only appear to be one if you only have one program with one course and one run. How do things change if you use multiple programs, each with multiple courses, each with multiple runs?

rlucioni: @ahsan-ul-haq can you be more specific than "one query per record"? One query per program? One per course run? That aside, "one per" suggests the same query is being run repeatedly. Prefetching could limit this to a single new query.

ahsan-ul-haq: I have tested with 3 programs among which 2 having 2 courses and 1 with 3 courses. Each course having 1 course_run and I excluded 4 course runs still the increase in the number of queries was of 1 query. Could you please verify that on your end as well maybe I would be doing something wrong and it would be better with two people testing this.

schenedx: @rlucioni In the team standup, I confirmed with @ahsan-ul-haq that he is indeed using debug tool bar to test and watch for the query count changes

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -201,6 +201,27 @@ def test_exclude_unpublished_program_types(self):
         """ Verify the unpublished programs do not show in the program_types representation. """
         self._test_exclude_program_types(ProgramStatus.Unpublished)
 
+    def test_excluded_course_run(self):
+        course_run = CourseRunFactory(course__partner=self.partner, course__title='Software Testing',
+                                      status=CourseRunStatus.Published)
+        ProgramFactory(courses=[course_run.course], status=ProgramStatus.Active, excluded_course_runs=[course_run])
```
schenedx: Can you also test the scenario where a program has more than 1 (maybe 3) course runs, and only one of the course runs is excluded?
Then call the excluded course run and ensure the NumQueries stays 9?

ahsan-ul-haq:
@schenedx added test

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -201,6 +201,51 @@ def test_exclude_unpublished_program_types(self):
...
+        with self.assertNumQueries(9):
+            response = self.get_search_response('software', faceted=False)
+
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+
+        # Validate the search results
+        self.assertEqual(response_data['results'][0].get('program_types'), [])
+        self.assertEqual(response_data['results'][1].get('program_types'), [])
+        self.assertNotEqual(response_data['results'][2].get('program_types'), [])
```
schenedx: Instead of "Not Equal", is there a better assert to make sure the 'programs_types' array is not empty?

schenedx:
@rlucioni Could you please help review again?

course_discovery/apps/api/v1/tests/test_views/test_search.py
ahsan-ul-haq: I can do assertTrue as well if you suggest so.

schenedx (approved):
👍

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -201,6 +201,51 @@ def test_exclude_unpublished_program_types(self):
...
+                                        status=CourseRunStatus.Published)
+        course_run_3 = CourseRunFactory(course__partner=self.partner, course__title='Software Testing 3',
+                                        status=CourseRunStatus.Published)
+        ProgramFactory(
+            courses=[course_run.course],
+            status=ProgramStatus.Active,
+            excluded_course_runs=[course_run_2, course_run_3]
+        )
+
+        with self.assertNumQueries(9):
```
rlucioni: You should be using ddt for this. As it is, not only have you duplicated an entire test case, it's not clear that the query counts are expected to be the same across both test cases.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -201,6 +201,51 @@ def test_exclude_unpublished_program_types(self):
...
+            response = self.get_search_response('software', faceted=False)
+
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+
+        # Validate the search results
+        self.assertEqual(response_data['results'][0].get('program_types'), [])
+        self.assertEqual(response_data['results'][1].get('program_types'), [])
+        self.assertTrue(response_data['results'][2].get('program_types'))
```
rlucioni: Be explicit about what is expected. This should contain the type of the program returned by the program factory.

rlucioni (approved):
@ahsan-ul-haq your tests need some cleanup, but looks fine otherwise.

PR #502, opened 2016-12-21
Publisher app: Add/Remove Permissions.

awais786:
Removed the course view permissions.
Add the permission for OrganizationExtension.
Add migration for the new permission.
Fix all old tests cases.


course_discovery/apps/publisher/models.py
```diff
@@ -412,6 +407,7 @@ def __str__(self):
 class OrganizationExtension(TimeStampedModel):
     """ Organization-Extension relation model. """
     EDIT_COURSE_RUN = 'edit_course_run'
+    VIEW_PERMISSION = 'organization_extension_access'
```
waheedahmed: ```ACCESS_PERMISSION = 'organization_access'```

course_discovery/apps/publisher/models.py
```diff
@@ -421,6 +417,7 @@ class OrganizationExtension(TimeStampedModel):
     class Meta(TimeStampedModel.Meta):
         permissions = (
             ('edit_course_run', 'Can edit course run'),
+            ('organization_extension_access', 'Can access the organization related course'),
```
waheedahmed: ```organization_access```

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,29 @@ def form_valid(self, form):
...
+            for org in course.organizations.all()
+        ]
+    )
+
+
 def check_view_permission(user, course):
-    return user.is_staff or user.has_perm(Course.VIEW_PERMISSION, course)
+    """ Return True if user is admin/internal user or has view permission. """
+
+    if check_roles_access(user):
```
waheedahmed: Simply ```return check_roles_access(user) or check_organization_permission(user, course)```

course_discovery/apps/publisher/models.py
```diff
@@ -152,17 +150,14 @@ class Meta(TimeStampedModel.Meta):
             ('view_course', 'Can view course'),
         )
 
-    def assign_permission_by_group(self, group):
-        """ Assigns permission on the course against the group. """
-        assign_perm(self.VIEW_PERMISSION, group, self)
-
     def get_group_users_emails(self):
```
waheedahmed: Change the name of this function and doc string.

course_discovery/apps/publisher/views.py
```diff
@@ -425,6 +429,11 @@ def get_queryset(self):
         elif is_internal_user(user):
             courses = Course.objects.filter(course_user_roles__user=user).distinct()
         else:
-            courses = get_objects_for_user(user, Course.VIEW_PERMISSION, Course)
+            organization_extensions = get_objects_for_user(
```
waheedahmed:
```
organizations = get_objects_for_user(
        user,
        OrganizationExtension.VIEW_PERMISSION,
        OrganizationExtension,
        use_groups=False,
        with_superuser=False
    ).values_list('organization')
courses = Course.objects.filter(organizations__in=organizations)

**Review by waheedahmed — CHANGES_REQUESTED** [review]

**Comment by waheedahmed** [comment]
@bderusha Please review.

**Comment by awais786** [comment]
@bderusha Please review.

File: course_discovery/apps/publisher/forms.py [hunk 6]
```diff
@@ -69,7 +69,7 @@ class UpdateCourseForm(BaseCourseForm):
     """ Course form to update specific fields for already created course. """
 
     number = forms.CharField(label=_('Course Number'), required=True)
-    team_admin = forms.ModelChoiceField(queryset=User.objects.filter(is_staff=True), required=True)
+    team_admin = forms.ModelChoiceField(queryset=User.objects.filter(), required=True)
```
bderusha: Why are we filtering on nothing?  Should this be the same as above `User.objects.all()` ?

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,26 @@ def form_valid(self, form):
...
+
+
+def check_organization_permission(user, course):
+    """ Return True if user has view permission on organization. """
+    if not hasattr(course, 'organizations'):
+        return False
+
+    return any(
+        [
+            user.has_perm(OrganizationExtension.ACCESS_PERMISSION, org.organization_extension)
```
bderusha: `OrganizationExtension.ACCESS_PERMISSION` is too broad of a permission.  While the current asks are this broad, I think we should be implementing more granular permissions here to prevent future rework and be more clear about exactly what views/methods/etc require what permissions.

Rather than checking `ACCESS_PERMISSION` can we pass in the appropriate permission to the method.  These should look something like `OrganizationExtension.COURSE_EDIT_PERMISSION` and `OrganizationExtension.COURSE_RUN_DELETE_PERMISSION` and `OrganizationExtension.SEAT_VIEW_PERMISSION` as appropriate

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,26 @@ def form_valid(self, form):
         return HttpResponseRedirect(self.get_success_url())
 
 
-def check_view_permission(user, course):
-    return user.is_staff or user.has_perm(Course.VIEW_PERMISSION, course)
+def check_roles_access(user):
+    """ Return True if user is admin/internal user. """
```
bderusha: Return True if user is part of a role that gives implicit access

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,26 @@ def form_valid(self, form):
...
+    if is_publisher_admin(user) or is_internal_user(user):
+        return True
+
```
bderusha, line 64: `return False` rather than `None`

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,26 @@ def form_valid(self, form):
...
+
+def check_organization_permission(user, course):
```
bderusha: This method is checking permissions for a course object.  We should either rename this or pass the organizations as the objects to be checked.

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,26 @@ def form_valid(self, form):
...
+
+    return any(
+        [
+            user.has_perm(OrganizationExtension.ACCESS_PERMISSION, org.organization_extension)
+            for org in course.organizations.all()
+        ]
+    )
+
+
+def check_user_access(user, course):
```
bderusha: currently is only checking for course objects.  We should rename or refactor.

course_discovery/apps/publisher/models.py
```diff
@@ -416,6 +407,7 @@ def __str__(self):
 class OrganizationExtension(TimeStampedModel):
     """ Organization-Extension relation model. """
     EDIT_COURSE_RUN = 'edit_course_run'
+    ACCESS_PERMISSION = 'organization_access'
```
bderusha: Per my comment above this is too broad. We want more specific permissions like the one above `edit_course_run`

bderusha (changes requested):

course_discovery/apps/publisher/forms.py
awais786: Yes. Refactoring mistake.

course_discovery/apps/publisher/migrations/0022_auto_20161221_1521.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('publisher', '0021_auto_20161214_1356'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='organizationextension',
+            options={'get_latest_by': 'modified', 'ordering': ('-modified', '-created'), 'permissions': (('edit_course_run', 'Can edit course run'), ('organization_access', 'Can access the organization related course'))},
+        ),
```
davec-edx: reformat please

bderusha: @davec-edx This is autogenerated, so I think it's fine to leave as is

davec-edx: absolutely correct, I didn't notice it was a migration.  ignore :-)

course_discovery/apps/publisher/mixins.py
awais786: I have rename the permission. e.g
VIEW_COURSE_RUN = 'view_course_run'

But passing a permission to a method will be a big change. So I would suggest to create separate ticket for that change. Because with that we will have to update many views where we are checking permission in the view `ChangeStateView`.

awais786: Also do we need to consider multiple permissions on single view ?

awais786:
@bderusha passing a permission to the ViewPermissionMixin is time consuming task and it should be cover in separate ticket.

awais786:
@bderusha I have created a separate ticket here.
https://openedx.atlassian.net/browse/ECOM-6687

course_discovery/apps/publisher/mixins.py
```diff
@@ -56,5 +57,28 @@ def form_valid(self, form):
...
+
+
+def check_course_organization_permission(user, course):
+    """ Return True if user has view permission on organization. """
+    if not hasattr(course, 'organizations'):
+        return False
+
+    return any(
+        [
+            user.has_perm(OrganizationExtension.VIEW_COURSE_RUN, org.organization_extension)
```
bderusha: should use the `VIEW_COURSE`

course_discovery/apps/publisher/models.py
```diff
@@ -416,6 +407,7 @@ def __str__(self):
 class OrganizationExtension(TimeStampedModel):
     """ Organization-Extension relation model. """
     EDIT_COURSE_RUN = 'edit_course_run'
+    VIEW_COURSE_RUN = 'view_course_run'
```
bderusha: Based on our conversation this morning can we update this to be namespaced. E.G. `publisher_view_course_run`

this should probably be `publisher_view_course` based on usage

course_discovery/apps/publisher/models.py
```diff
@@ -416,6 +407,7 @@ def __str__(self):
 class OrganizationExtension(TimeStampedModel):
     """ Organization-Extension relation model. """
     EDIT_COURSE_RUN = 'edit_course_run'
```
bderusha, line 409: Can you namespace this one as well while you are in here?

course_discovery/apps/publisher/tests/test_utils.py
```diff
@@ -1,11 +1,15 @@
 """ Tests publisher.utils"""
 from django.contrib.auth.models import Group
 from django.test import TestCase
+from guardian.shortcuts import assign_perm
 
 from course_discovery.apps.core.tests.factories import UserFactory
 from course_discovery.apps.publisher.constants import (
     ADMIN_GROUP_NAME, INTERNAL_USER_GROUP_NAME, PARTNER_COORDINATOR_GROUP_NAME
 )
+from course_discovery.apps.publisher.mixins import check_roles_access, check_course_organization_permission, \
```
bderusha: Prefer `import ( )` with parens for multiline imports

course_discovery/apps/publisher/views.py
```diff
@@ -59,9 +61,14 @@ def get_queryset(self):
             internal_user_courses = Course.objects.filter(course_user_roles__user=user)
             course_runs = CourseRun.objects.filter(course__in=internal_user_courses).select_related('course').all()
         else:
-            # in future we will change permission from course to OrganizationExtension model
-            courses = get_objects_for_user(user, Course.VIEW_PERMISSION, Course)
-            course_runs = CourseRun.objects.filter(course__in=courses).select_related('course').all()
+            organizations = get_objects_for_user(
+                user, OrganizationExtension.VIEW_COURSE_RUN, OrganizationExtension,
```
bderusha: This is the correct permission.  So It looks like we need to add both `VIEW_COURSE_RUN` and `VIEW_COURSE` as permissions on the `OrganizationExtension` model

course_discovery/apps/publisher/views.py
```diff
@@ -425,6 +429,13 @@ def get_queryset(self):
         elif is_internal_user(user):
             courses = Course.objects.filter(course_user_roles__user=user).distinct()
         else:
-            courses = get_objects_for_user(user, Course.VIEW_PERMISSION, Course)
+            organizations = get_objects_for_user(
+                user,
+                OrganizationExtension.VIEW_COURSE_RUN,
```
bderusha: `VIEW_COURSE`

bderusha (approved):
Some small comments.  :shipit: once resolved.  No need for re-review

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -33,10 +30,8 @@ def setUp(self):
             self.other_internal_users.append(user)
             internal_user_group.user_set.add(user)
 
-        assign_perm(Course.VIEW_PERMISSION, internal_user_group, self.course)
-
-        organization = OrganizationFactory()
-        self.course.organizations.add(organization)
+        self.organization_extension = factories.OrganizationExtensionFactory()
```
tasawernawaz: If we are not using `self.organization_extension` in any test than no need to add `self` with it.

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -104,14 +98,13 @@ def setUp(self):
         self.client.login(username=user.username, password=USER_PASSWORD)
 
         # create group and add test users in the group
```
tasawernawaz: update comment

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -104,14 +98,13 @@ def setUp(self):
...
-        group = factories.GroupFactory()
+        self.organization_extension = factories.OrganizationExtensionFactory()
```
tasawernawaz: Similarly here, remove `self`

course_discovery/apps/publisher/forms.py
```diff
@@ -53,7 +53,7 @@ class CustomCourseForm(CourseForm):
     )
     title = forms.CharField(label=_('Course Title'), required=True)
     number = forms.CharField(label=_('Course Number'), required=True)
-    team_admin = forms.ModelChoiceField(queryset=User.objects.filter(is_staff=True), required=True)
+    team_admin = forms.ModelChoiceField(queryset=User.objects.all(), required=True)
```
tasawernawaz, line 56: why this change ?

course_discovery/apps/publisher/forms.py
```diff
@@ -69,7 +69,7 @@ class UpdateCourseForm(BaseCourseForm):
     """ Course form to update specific fields for already created course. """
 
     number = forms.CharField(label=_('Course Number'), required=True)
-    team_admin = forms.ModelChoiceField(queryset=User.objects.filter(is_staff=True), required=True)
+    team_admin = forms.ModelChoiceField(queryset=User.objects.all(), required=True)
```
tasawernawaz, line 72: why this change ?

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -34,23 +34,26 @@ def setUp(self):
...
 
-        # assign the role against a course
+        self.course.organizations.add(self.organization_extension.organization)
+
+        # NOTE: We intentionally do NOT create an attribute for user_2.
+        # By default this user WILL receive email notifications.
+
+        # add user in course-user-role table
+
+        # add user in course-user-role table
```
tasawernawaz: remove redundant comment.

course_discovery/apps/publisher/tests/test_utils.py
```diff
@@ -82,3 +92,51 @@ def test_is_partner_coordinator_user(self):
...
+        self.user.groups.add(self.admin_group)
+        self.assertTrue(check_user_course_access(self.user, self.course))
+        self.user.groups.remove(self.admin_group)
+        self.assertFalse(check_user_course_access(self.user, self.course))
+        self.user.groups.add(self.internal_user_group)
+        self.assertTrue(check_user_course_access(self.user, self.course))
+
+    def test_check_user_access_with_permission(self):
+        """ Verify the function returns a boolean indicating if the user
+        organization permission on given course or user is internal or admin user.
```
tasawernawaz: I think this docstring needs to be updated, we are not checking for admin or internal users access.

tasawernawaz (changes requested):

course_discovery/apps/publisher/forms.py
awais786: We are not going to use staff any where now.

course_discovery/apps/publisher/forms.py
awais786: We are not going to use staff any where now.

course_discovery/apps/publisher/models.py
awais786: We will do this in a separate ticket. For this change we have to remove the old permission from the database with data migration.

tasawernawaz (approved):

tasawernawaz:

PR #503, opened 2016-12-21
Add flag to hide unpublished course runs on the course endpoint

MatthewPiatetsky:
@edx/ecommerce  @AlasdairSwan

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -472,7 +472,23 @@ class CourseSerializer(MinimalCourseSerializer):
 
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
+    course_runs = serializers.SerializerMethodField()
     programs = serializers.SerializerMethodField()
+
+    def get_course_runs(self, course):
+        course_runs = course.course_runs.exclude(hidden=True)
+
+        if self.context.get('published_course_runs_only'):
+            course_runs = course_runs.filter(status=CourseRunStatus.Published)
+
+        return CourseRunSerializer(
+            course_runs,
+            many=True,
+            context={
+                'request': self.context.get('request'),
+                'exclude_utm': self.context.get('exclude_utm'),
+            }
+        ).data
 
     def get_programs(self, obj):
         if self.context.get('include_deleted_programs'):
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -155,6 +155,7 @@ class CourseSerializerTests(MinimalCourseSerializerTests):
         self.assertEqual(serializer.data['marketing_url'], course.marketing_url)
 
 
+@ddt.ddt
 class CourseWithProgramsSerializerTests(CourseSerializerTests):
     serializer_class = CourseWithProgramsSerializer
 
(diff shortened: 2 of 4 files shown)


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -44,19 +46,37 @@ def test_get_include_deleted_programs(self):
...
             self.assertEqual(response.status_code, 200)
             self.assertEqual(
                 response.data,
                 self.serialize_course(self.course, extra_context={'include_deleted_programs': True})
             )
 
+    @ddt.data(1, 0)
+    def test_get_include_published_course_run(self, published_course_runs_only):
+        """
+        Verify the endpoint returns associated deleted programs
```
mikedikan: change to published/unpublished (says deleted)

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -46,6 +46,7 @@ def get_serializer_context(self, *args, **kwargs):
         context.update({
             'exclude_utm': get_query_param(self.request, 'exclude_utm'),
             'include_deleted_programs': get_query_param(self.request, 'include_deleted_programs'),
+            'published_course_runs_only': get_query_param(self.request, 'published_course_runs_only'),
```
mikedikan, line 49: why 'published_course_runs_only' instead of 'exclude_unpublished_course_runs' or changing default to excluding unpublished and make the flag 'include_unpublished_course_runs

mikedikan (changes requested):

course_discovery/apps/api/v1/views/courses.py
MatthewPiatetsky: there is already an existing flag for programs called published_course_runs_only

mikedikan (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -472,8 +472,24 @@ def get_marketing_url(self, obj):
 
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
+    course_runs = serializers.SerializerMethodField()
     programs = serializers.SerializerMethodField()
 
+    def get_course_runs(self, course):
```
amangano-edx, line 478: Where does this get called?

MatthewPiatetsky: this is serializer magic. See other fields that use serializers.SerializerMethodField() in this file

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -198,6 +198,19 @@ def test_include_deleted_programs(self):
...
+    def test_published_course_runs_only(self):
+        """
+        Test that the published_course_runs_only flag hides unpublished course runs
+        """
+        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        self.course.course_runs.add(unpublished_course_run)
+        self.request = make_request()
+        serializer = self.serializer_class(
+            self.course,
+            context={'request': self.request, 'published_course_runs_only': 1}
```
amangano-edx: Why are we using 1 for published_course_runs_only instead of a bool?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -198,6 +198,19 @@ def test_include_deleted_programs(self):
...
+        )
+        self.assertEqual(serializer.data['course_runs'], [])
```
amangano-edx: Minor, but it might be nice to verify that it returns published course runs when this flag is present and only excludes those that are unpublished. As it stands, this test will pass even when all course runs are excluded.

amangano-edx:
I have limited context on this, but LGTM

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: in this case because we were sending that from the marketing site (also works out for the new test)

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: now testing both published/unpublished with/without the flag

PR #504, opened 2016-12-21
ECOM-6597/ECOM-6596 Management command for creating programs

MatthewPiatetsky:
Creates test partner, test program, and a program that looks like the wharton pc
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/create_test_program.py
+++ b/course_discovery/apps/course_metadata/management/commands/create_test_program.py
@@ -1,0 +1,93 @@
+import logging
+
+from django.core.management import BaseCommand
+
+from course_discovery.apps.core.models import Partner
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
+from course_discovery.apps.course_metadata.models import CourseRun, Program, ProgramType, Organization
+from course_discovery.apps.course_metadata.tests.factories import (
+    OrganizationFactory, CorporateEndorsementFactory, EndorsementFactory, ExpectedLearningItemFactory,
+    FAQFactory, JobOutlookItemFactory, ProgramFactory, CourseFactory, CourseRunFactory, PartnerFactory
+)
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Create program with test data.'
+
+    def handle(self, *args, **options):
+        partner = self.create_test_partner()
+        self.create_test_program(partner)
+        self.create_test_wharton_program(partner)
+
+    def create_test_partner(self):
+        partner = Partner.objects.filter(name='test-partner')
+        if partner:
+            partner.delete()
+        logger.info('Creating test-partner')
+        partner = PartnerFactory(name='test-partner', marketing_site_url_root=None)
+        return partner
+
+    def create_test_program(self, partner):
+        program = Program.objects.filter(title='test-program')
+        if program:
+            program.first().delete()
+        logger.info('Creating test-program')
+        course = CourseFactory(partner=partner)
+        excluded_course_run = CourseRunFactory(course=course)
+        ProgramFactory(
+            title='test-program',
+            partner=partner,
+            courses=[course],
+            excluded_course_runs=[excluded_course_run],
+            authoring_organizations=[OrganizationFactory(partner=partner)],
+            credit_backing_organizations=[OrganizationFactory(partner=partner)],
+            corporate_endorsements=[CorporateEndorsementFactory()],
+            individual_endorsements=[EndorsementFactory(endorser__partner=partner)],
+            expected_learning_items=[ExpectedLearningItemFactory()],
+            faq=[FAQFactory()],
+            job_outlook_items=[JobOutlookItemFactory()]
+        )
+
+    def create_test_wharton_program(self, partner):
+        program = Program.objects.filter(title='Test Digital Marketing Professional Certificate')
+        if program:
+            program.first().delete()
+        logger.info('Creating Test Wharton Program')
+        course_run_keys = [
+            'course-v1:Wharton+CustomerCentricityx+3T2016',
+            'course-v1:Wharton+DigitalMarketing1.1x+3T2016',
+            'course-v1:Wharton+MarketingAnalytics101x+2T2016',
+            'course-v1:Wharton+SellingIdeas101x+3T2016'
+        ]
+        courses = [course_run.course for course_run in CourseRun.objects.filter(key__in=course_run_keys)]
+        wharton, _ = Organization.objects.get_or_create(key="Wharton")
+        ProgramFactory(
+            title="Test Digital Marketing Professional Certificate",
+            subtitle="Gain a competitive advantage and learn essential digital marketing skills and strategies in this 4-course program from the top ranked Wharton School of the University of Pennsylvania.",  # pylint: disable=line-too-long
+            status=CourseRunStatus.Unpublished,
+            type=ProgramType.objects.get(name="Professional Certificate"),
+            partner=partner,
+            banner_image_url="https://www.edx.org/sites/default/files/prod-land-pg/banner/certificate_banner_1440x326.jpg",  # pylint: disable=line-too-long
+            card_image_url="https://www.edx.org/digital-marketing-professional",
+            marketing_slug="digital-marketing-professional",
+            overview="Make yourself more marketable and put your career in high gear with a Professional Program in Digital Marketing from Wharton. Learn the key marketing skills most in-demand today: omni-channel marketing, marketing analytics, social media strategy and analysis, and data-driven customer-centric approaches to customer retention. Designed by world-renowned marketing professors at the Wharton School, home of one of the top marketing programs in the country, the Professional Certificate in Digital Marketing from Wharton helps you develop the digital marketing skills you need to take advantage of the explosive growth in the marketing industry. Today’s marketers face a “digital skills gap,” and when you earn a Professional Certificate in Digital Marketing from the Wharton School, you’ll be able to help fill it.",  # pylint: disable=line-too-long
+            expected_learning_items=[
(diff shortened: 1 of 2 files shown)


MatthewPiatetsky:
@edx/ecommerce based on the codecov looks like I need to write tests for this command. 
To be able to do this, I need to further understand the requirements.
Do we want this command to be run once, on every gocd deploy (except prod)?
Do we want the command to overwrite the test data if it already exists or do nothing in that case?

mikedikan:
@MatthewPiatetsky I think at the outset, it won't be run by gocd.  Also, i think the overwrite idea makes the most sense to me, it's like hitting the reset button

mikedikan:
jenkins run a11y

course_discovery/apps/course_metadata/management/commands/create_test_program.py
```diff
@@ -0,0 +1,95 @@
...
+from course_discovery.apps.course_metadata.models import CourseRun, Program, ProgramType, Organization
+from course_discovery.apps.course_metadata.tests.factories import (
+    OrganizationFactory, CorporateEndorsementFactory, EndorsementFactory, ExpectedLearningItemFactory,
+    FAQFactory, JobOutlookItemFactory, ProgramFactory, CourseFactory, CourseRunFactory, PartnerFactory
+)
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
mikedikan, line 16: Command seems like too generic of a name for this class - perhaps CreateTestDataCommand?

course_discovery/apps/course_metadata/management/commands/create_test_program.py
```diff
@@ -0,0 +1,95 @@
...
+        courses = [course_run.course for course_run in CourseRun.objects.filter(key__in=course_run_keys)]
+        wharton, _ = Organization.objects.get_or_create(key="Wharton")
+        ProgramFactory(
+            title="Test Digital Marketing Professional Certificate",
+            subtitle="Gain a competitive advantage and learn essential digital marketing skills and strategies in"
+            "this 4-course program from the top ranked Wharton School of the University of Pennsylvania.",
+            status=CourseRunStatus.Unpublished,
+            type=ProgramType.objects.get(name="Professional Certificate"),
+            partner=partner,
+            banner_image_url="https://www.edx.org/sites/default/files/prod-land-pg/banner/certificate_banner_1440x326.jpg",  # pylint: disable=line-too-long
```
mikedikan, line 72: Curious why you chose the pylint hints instead of breaking these up

course_discovery/apps/course_metadata/management/commands/create_test_program.py
```diff
@@ -0,0 +1,95 @@
...
+            program.first().delete()
+        logger.info('Creating Test Wharton Program')
+        course_run_keys = [
+            'course-v1:Wharton+CustomerCentricityx+3T2016',
+            'course-v1:Wharton+DigitalMarketing1.1x+3T2016',
+            'course-v1:Wharton+MarketingAnalytics101x+2T2016',
+            'course-v1:Wharton+SellingIdeas101x+3T2016'
+        ]
+        courses = [course_run.course for course_run in CourseRun.objects.filter(key__in=course_run_keys)]
+        wharton, _ = Organization.objects.get_or_create(key="Wharton")
```
mikedikan, line 65: what is the _ for?

course_discovery/apps/course_metadata/management/commands/tests/test_create_test_program.py
```diff
@@ -0,0 +1,23 @@
...
+
+from course_discovery.apps.core.models import Partner
+from course_discovery.apps.course_metadata.models import Program
+from course_discovery.apps.course_metadata.tests.factories import PartnerFactory, ProgramFactory
+
+
+class CreateTestProgramCommandTests(TestCase):
+    def test_create_command(self):
+        call_command('create_test_program')
+        # Verify that these objects have been created
```
mikedikan, line 12: why are there no asserts in any of these tests?

course_discovery/apps/course_metadata/management/commands/create_test_program.py
MatthewPiatetsky: All our management commands have this class, think it is standard

course_discovery/apps/course_metadata/management/commands/create_test_program.py
MatthewPiatetsky: get_or_create return a tuple, the second value is a bool which indicates whether the object was created or not, which we don't care about.  
if you look at other uses of get_or_create in the catalog we use _ to assign that value that we don't care about

course_discovery/apps/course_metadata/management/commands/create_test_program.py
MatthewPiatetsky: well these aren't lines of code just really long strings of text, it seemed like it would make it less rather than more readable to split each string into many many lines

course_discovery/apps/course_metadata/management/commands/tests/test_create_test_program.py
MatthewPiatetsky: I guess I could add asserts to check the value of specific data points on the program, but it seems to me like the main thing I care about is whether the objects actually get created or not? Most of the data is just some sort of random text from the factory

davec-edx: Are we also testing creation of a duplicate, verification that getting bad names, etc. will fail appropriately?  Ie, both positive and negative paths.

MatthewPiatetsky: This will overwrite an existing partner/program of the same name so you can't get a duplicate. Not sure what you mean by bad names

course_discovery/apps/course_metadata/management/commands/create_test_program.py
mikedikan: 👌

course_discovery/apps/course_metadata/management/commands/create_test_program.py
mikedikan: 👌

course_discovery/apps/course_metadata/management/commands/create_test_program.py
mikedikan: that's fine, but on some lines you wrap the text, and on others you don't - can you make it consistent?

course_discovery/apps/course_metadata/management/commands/tests/test_create_test_program.py
mikedikan: 👌

course_discovery/apps/course_metadata/management/commands/tests/test_create_test_program.py
```diff
@@ -0,0 +1,23 @@
...
+        # If they have not they will throw a DoesNotExist error
+        Partner.objects.get(name='test-partner')
+        Program.objects.get(title='test-program')
+        Program.objects.get(title='Test Digital Marketing Professional Certificate')
+
+    def test_create_command_overwrite(self):
+        # Verify that these objects can be overwritten
+        PartnerFactory(name='test-partner')
```
mikedikan: I'm confused about this test - it's testing that by calling the command, no exception is thrown?  shouldn't we also test that you can still get the objects with the same title?

mikedikan: should we also test that it will overwrite data and reset to default (meaning, if you give title and subtitle, calling the command will reset the subtitle to the test default)?

course_discovery/apps/course_metadata/management/commands/create_test_program.py
MatthewPiatetsky: made it consistent

course_discovery/apps/course_metadata/management/commands/tests/test_create_test_program.py
MatthewPiatetsky: changed test to set a value on the existing objects and check that the value is different after running the command

course_discovery/apps/course_metadata/management/commands/tests/test_create_test_program.py
```diff
@@ -0,0 +1,29 @@
...
+        PartnerFactory(name='test-partner', short_code='test')
+        ProgramFactory(title='test-program', subtitle='test')
+        ProgramFactory(title='Test Digital Marketing Professional Certificate', subtitle='test')
+        call_command('create_test_program')
+        partner = Partner.objects.get(name='test-partner')
+        test_program = Program.objects.get(title='test-program')
+        test_wharton_program = Program.objects.get(title='Test Digital Marketing Professional Certificate')
+        self.assertNotEqual(partner.short_code, 'test')
+        self.assertNotEqual(test_program.subtitle, 'test')
+        self.assertNotEqual(test_wharton_program.subtitle, 'test')
```
mikedikan, line 29: 👌

mikedikan (approved):

bderusha (approved):
Overall I think this is fine.  My only concern is about extensibility.  Would all of the content be better contained in a YAML file that we manage and maintain?  Are there any attributes that we may want to programmatically create iterations of?

Not going to block this PR on any of that, but I would want to know what the plan is

course_discovery/apps/course_metadata/management/commands/create_test_program.py
clintonb: The base class is standard. The class name is not. Please rename the class to be more descriptive.

clintonb: Never mind. Disregard.

MatthewPiatetsky:
@bderusha agreed (though personally I prefer json to yaml)
However, I don't think we currently have a plan for how we would extend this moving forward. 
I think it makes sense to just have something basic for now and then make it configurable once we know what we'll want to configure

PR #506, opened 2016-12-22
Added studio instance required fields.

waheedahmed:
[ECOM-6093](https://openedx.atlassian.net/browse/ECOM-6093)

```Course Organization/Admin``` fields will be handled in [ECOM-6090](https://openedx.atlassian.net/browse/ECOM-6090)

screencapture-localhost-8004-publisher-courses-new-1482402415872

### CODE REVIEW

File: course_discovery/apps/publisher/forms.py [hunk 1]
```diff
@@ -49,11 +49,16 @@ class CustomCourseForm(CourseForm):
...
     organization = forms.ModelChoiceField(
         queryset=Organization.objects.filter(organization_extension__organization_id__isnull=False),
+        label=_('Institution Name'),
         required=True
     )
     title = forms.CharField(label=_('Course Title'), required=True)
     number = forms.CharField(label=_('Course Number'), required=True)
-    team_admin = forms.ModelChoiceField(queryset=User.objects.filter(is_staff=True), required=True)
+    team_admin = forms.ModelChoiceField(
+        queryset=User.objects.filter(is_staff=True),
```
awais786: User.objects.all().
We are not going to use is_staff any where.

awais786:
What about if datetime not given ?

course_discovery/apps/publisher/forms.py
```diff
@@ -100,33 +105,35 @@ class CustomCourseRunForm(CourseRunForm):
...
     target_content = forms.BooleanField(
         widget=forms.RadioSelect(
             choices=((1, _("Yes")), (0, _("No")))), initial=0, required=False
     )
-    is_self_paced = forms.BooleanField(label=_('Yes, course will be Self-Paced'), required=False)
+    pacing_type = forms.ChoiceField(
+        label=_('Pace'),
+        widget=forms.RadioSelect,
+        choices=CourseRunPacing.choices,
+        initial=CourseRunPacing.Self,
```
awais786: **Radio button - Instructor paced/Self paced - required**

I think you don't need to select using initial value.

awais786:
Few minor things otherwise looks good.

course_discovery/apps/publisher/forms.py
waheedahmed: @awais786 I think we have to set an initial value for this, maybe to ```CourseRunPacing.Instructor``` because it was the requirement for new course-run form. We will confirm that with Deen.

waheedahmed:
@awais786 Which datetime?

course_discovery/apps/publisher/forms.py
```diff
@@ -100,33 +105,35 @@ class CustomCourseRunForm(CourseRunForm):
...
     target_content = forms.BooleanField(
         widget=forms.RadioSelect(
             choices=((1, _("Yes")), (0, _("No")))), initial=0, required=False
     )
-    is_self_paced = forms.BooleanField(label=_('Yes, course will be Self-Paced'), required=False)
+    pacing_type = forms.ChoiceField(
+        label=_('Pace'),
+        widget=forms.RadioSelect,
+        choices=CourseRunPacing.choices,
+        initial=CourseRunPacing.Instructor,
```
awais786: For some reason last comment is not appearing. My question was do we need initial value here ?

awais786:
Datetime for ( start date or end-date ) if user did not give any date ?
do we need to save any thing in db ?

course_discovery/apps/publisher/forms.py
waheedahmed: @awais786 I'll remove that.

waheedahmed:
@awais786 User has to select a start/end date, we will show an error if nothing provided.

bderusha (approved):
Once the radio buttons are defaulted to nothing selected this looks good 👍

course_discovery/apps/publisher/forms.py
```diff
@@ -100,33 +105,35 @@ class CustomCourseRunForm(CourseRunForm):
     """ Course Run Form. """
 
     contacted_partner_manager = forms.BooleanField(
-        widget=forms.RadioSelect(choices=((1, _("Yes")), (0, _("No")))), initial=0, required=False
+        label=_('Contacted PM'),
+        widget=forms.RadioSelect(choices=((1, _("Yes")), (0, _("No")))),
+        initial=0,
```
bderusha: This also shouldn't have a default selected (per our conversation this morning)

course_discovery/apps/publisher/forms.py
```diff
@@ -18,10 +18,10 @@ def __init__(self, *args, **kwargs):
             field_classes = 'field-input input-text'
             if isinstance(field, forms.Textarea):
                 field_classes = 'field-textarea input-textarea'
+            if isinstance(field, (forms.BooleanField, forms.ChoiceField,)):
+                field_classes = 'field-input input-checkbox'
```
tasawernawaz, line 22: why we are not using `if-elif` instead of multiple ifs.?

course_discovery/apps/publisher/forms.py
```diff
@@ -49,10 +49,12 @@ class CustomCourseForm(CourseForm):
     """ Course Form. """
     organization = forms.ModelChoiceField(
         queryset=Organization.objects.filter(organization_extension__organization_id__isnull=False),
+        label=_('Institution Name'),
         required=True
     )
     title = forms.CharField(label=_('Course Title'), required=True)
     number = forms.CharField(label=_('Course Number'), required=True)
+    team_admin = forms.ModelChoiceField(queryset=User.objects.filter(is_staff=True), required=True)
```
tasawernawaz: remove this line,  its already there https://github.com/edx/course-discovery/pull/506/files#diff-af9668030177ee3112a8eb5f7d81a484L58

awais786:
Radio buttons are now not selected by default.

screen shot 2016-12-29 at 11 13 49 pm

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -271,6 +272,18 @@ def test_create_with_invalid_team_admin(self):
         response = self.client.post(reverse('publisher:publisher_courses_new'), course_dict, files=data['image'])
         self.assertEqual(response.status_code, 400)
 
+    @ddt.data('contacted_partner_manager', 'pacing_type')
+    def test_create_without_selecting_radio_buttons(self, button_field):
+        """
+        Verify that without selecting radio buttons fields objects cannot be
+        created.
```
tasawernawaz: Please update this docstring like `Verify that without selecting radio button fields course cannot be created.`

tasawernawaz (changes requested):

course_discovery/apps/publisher/forms.py
awais786: Existing code.

awais786: We will discuss this with the actual author @waheedahmed.

awais786:
@tasawernawaz You can review the PR again.
contacted_partner_manager has initial value as default. It's a limitation in django that if boolean field is required True then selecting No will raise the error.

course_discovery/apps/publisher/forms.py
```diff
@@ -112,33 +113,34 @@ class CustomCourseRunForm(CourseRunForm):
     """ Course Run Form. """
 
     contacted_partner_manager = forms.BooleanField(
-        widget=forms.RadioSelect(choices=((1, _("Yes")), (0, _("No")))), initial=0, required=False
+        label=_('Contacted PM'),
+        widget=forms.RadioSelect(choices=((1, _("Yes")), (0, _("No")))),
+        required=False,
```
tasawernawaz: @awais786  `contacted_partner_manager ` is required field in https://openedx.atlassian.net/browse/ECOM-6093, why you make it `False`

tasawernawaz (changes requested):

course_discovery/apps/publisher/forms.py
tasawernawaz: So are we waiting for @waheedahmed on this one ?

tasawernawaz (changes requested):

awais786:
@tasawernawaz this is not some thing new. Its already in code base.

tasawernawaz:
@awais786 Ok we can leave it for now.

course_discovery/apps/publisher/forms.py
awais786: It remains required. I just gave its initial value. Its a limitation in django forms. If boolean field is required True then selecting No will raise the missing value error.

tasawernawaz: Just curious about this ? other than looks LGTM.

course_discovery/apps/publisher/forms.py
awais786: nop. We can discuss this in separate stories. its existing code.

tasawernawaz:
@awais786 Thats what I mean, we can leave code as it is, and will discuss it later.

course_discovery/apps/publisher/forms.py
awais786: https://docs.djangoproject.com/en/1.10/ref/forms/fields/#booleanfield

tasawernawaz (approved):

awais786:
screen shot 2016-12-30 at 2 43 49 pm

PR #507, opened 2016-12-22
Removed "team admin" from course model

tasawernawaz:
Institution/Course Admin widget

Subtasks of "Institution/Course Admin widget" [ECOM-6690](https://openedx.atlassian.net/browse/ECOM-6690)
*  [ECOM-6665](https://openedx.atlassian.net/browse/ECOM-6665)
*  [ECOM-6667](https://openedx.atlassian.net/browse/ECOM-6667)

A/C:
* Remove team admin field from course.
* On saving course save the team admin user as course team role in CourseUserRole table against the course.


awais786:
Is it backward compatible ?

awais786:
team_admin is in use at several places. It will crash the build. Please remove it from every where.

bderusha:
I agree with @awais786 all dependencies on this field must also be removed

tasawernawaz:
@bderusha @awais786 Yes you were right, initially I thought that could be done in  ECOM-6667, but now I will work on both subtasks in same PR, because both are interrelated.

course_discovery/apps/publisher/models.py
```diff
@@ -403,6 +402,18 @@ def __str__(self):
...
+    def add_course_roles(cls, course, user, role):
+        """ Create course roles .
+        Args:
+            course (:obj:`Course`): Course
+            user (:obj:`User`): User
+            role (str): role description
+        Returns:
+            CourseUserRole obj
+        """
+        return cls.objects.get_or_create(course=course, user=user, role=role)
```
tasawernawaz: this method is copied from https://github.com/edx/course-discovery/pull/510/files#diff-ea7d9c84b7af3aab7f0c05a73be35752R413, will be updated with latest code there in #510

tasawernawaz:
@ziafazal Please review this.

course_discovery/apps/publisher/views.py
```diff
@@ -212,6 +211,9 @@ def post(self, request, *args, **kwargs):
                     )
                     course.organizations.add(organization_extension.organization)
 
+                    CourseUserRole.add_course_roles(course=course, user=self.request.user,
```
ziafazal: Should not here we have `team_admin_id` instead of `self.request.user`?

course_discovery/templates/publisher/add_course_form.html
```diff
@@ -39,7 +39,7 @@ <h2 class="layout-title">{% trans "Base information" %}</h2>
                     {% trans "Please choose the school that will be providing the course. Once chosen then you can select an administrator for the studio shell." %}
                   </div>
                   <div class="col col-6">
-                    {% if course_form.organization.choices|length > 1 %}
+                    {% if course_form.organization.field.queryset.all|length > 1 %}
```
ziafazal: i think we should use count instead of applying length filter i.e.`field.queryset.all.count`

ziafazal (changes requested):
@tasawernawaz a couple of suggestions.

ziafazal (approved):

awais786 (approved):

PR #509, opened 2016-12-26
ECOM-6666 front end of istitution admin

tasawernawaz:
[ECOM-6666](https://openedx.atlassian.net/browse/ECOM-6666)

* Added AJAX call to load admin users related to organization selected
* If there is only one organization it will be show as text otherwise it will be a drop down list.

### DIFF
--- a/course_discovery/apps/publisher/forms.py
+++ b/course_discovery/apps/publisher/forms.py
@@ -53,7 +53,9 @@ class CustomCourseForm(CourseForm):
     )
     title = forms.CharField(label=_('Course Title'), required=True)
     number = forms.CharField(label=_('Course Number'), required=True)
-    team_admin = forms.ModelChoiceField(queryset=User.objects.all(), required=True)
+
+    # users will be loaded through AJAX call based on organization
+    team_admin = forms.ModelChoiceField(queryset=User.objects.none(), required=True)
 
     class Meta(CourseForm.Meta):
         model = Course
@@ -63,6 +65,13 @@ class CustomCourseForm(CourseForm):
             'tertiary_subject', 'prerequisites', 'level_type', 'image', 'team_admin',
             'level_type', 'organization', 'is_seo_review', 'keywords',
         )
+
+    def __init__(self, *args, **kwargs):
+        queryset = kwargs.pop('team_admin_queryset', None)
+        if queryset:
+            self.declared_fields['team_admin'].queryset = queryset
+
+        super(CustomCourseForm, self).__init__(*args, **kwargs)
 
 
 class UpdateCourseForm(BaseCourseForm):
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -16,6 +16,7 @@ from django_fsm import TransitionNotAllowed
 from guardian.shortcuts import get_objects_for_user
 from rest_framework.generics import UpdateAPIView
 
+from course_discovery.apps.core.models import User
 from course_discovery.apps.publisher.choices import PublisherUserRole
 from course_discovery.apps.publisher.forms import (
     CourseForm, CourseRunForm, SeatForm, CustomCourseForm, CustomCourseRunForm,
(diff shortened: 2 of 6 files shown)


course_discovery/apps/publisher/forms.py
```diff
@@ -13,6 +13,9 @@ class BaseCourseForm(forms.ModelForm):
     """ Base Course Form. """
 
     def __init__(self, *args, **kwargs):
```
awais786, line 15: You don't need to do this in baseform.  Make change in courseForm only.

tasawernawaz: Moved in `CustomCourseForm`

tasawernawaz:
@ziafazal Please review this.

course_discovery/apps/publisher/forms.py
```diff
@@ -64,6 +66,13 @@ class Meta(CourseForm.Meta):
             'level_type', 'organization', 'is_seo_review', 'keywords',
         )
 
+    def __init__(self, *args, **kwargs):
+        queryset = kwargs.pop('team_admin_queryset', None)
```
ziafazal, line 70: What is the reason for this override?. I'm not sure if it is required or not.

course_discovery/apps/publisher/views.py
```diff
@@ -173,9 +174,13 @@ def get(self, request, *args, **kwargs):
 
     def post(self, request, *args, **kwargs):
         ctx = self.get_context_data()
-        course_form = self.course_form(request.POST, request.FILES)
+
+        # add selected team admin into choice of ChoiceField
+        team_admin_queryset = User.objects.filter(id=self.request.POST.get('team_admin'))
```
ziafazal, line 184: This also does not make sense to me.

course_discovery/static/js/publisher/publisher.js
```diff
@@ -23,4 +23,36 @@ $(document).ready(function(){
...
     });
+
+    var org_id = $('#organization-name').data('org_id');
+    if (org_id){
+        loadAdminUsers(org_id);
+    }
+});
+
+$(document).on('change', '#id_organization', function (e) {
+    var org_id = this.value;
```
ziafazal: Can we use more safe jqeury API `$(this).value()` here?

course_discovery/static/js/publisher/publisher.js
```diff
@@ -23,4 +23,36 @@ $(document).ready(function(){
...
+
+    // it will reset the select input
+    $("#id_team_admin").prop("selectedIndex", 0);
```
ziafazal, line 37: What if we empty team admin drop down here instead of setting its selectedIndex to zero.

course_discovery/static/js/publisher/publisher.js
```diff
@@ -23,4 +23,36 @@ $(document).ready(function(){
...
+        url: '/publisher/api/admins/organizations/'+ org_id +'/users/',
+        success: function (data) {
+            var teamAdminDropDown = $('#id_team_admin');
+            teamAdminDropDown.empty();
+
+            // it will looks same like other django model choice fields
+            teamAdminDropDown.append('<option selected="selected">---------</option>');
+
+            $.each(data.results, function () {
+                teamAdminDropDown.append($('<option> </option>').val(this.id).html(this.full_name));
```
ziafazal: nit: can we make it more readable?
```
 $.each(data.results, function (user) {
     teamAdminDropDown.append($('<option> </option>').val(user.id).html(user.full_name));
});
```

ziafazal (changes requested):
@tasawernawaz overall LGTM just couple of suggestions.

course_discovery/apps/publisher/forms.py
tasawernawaz: Problem is selected choice must be in choice set of `ModelChoiceField`. Thats why we need to add selected choice in the choices set of `team_admin`. If not then it will give error `selected choice is not valid choice` .

course_discovery/apps/publisher/views.py
tasawernawaz: `ModelChoiceField` has `queryset` parameter for choices is by default None. After selection of team admin that admin will be sent to `Form` class to add `queryset` in choices.

course_discovery/static/js/publisher/publisher.js
tasawernawaz: `0` index will contain django empty choice e.g `--------` that will be selected, no admin will be selected on change of organization

course_discovery/static/js/publisher/publisher.js
tasawernawaz: `$.each(data.results, function (index, user)` because first parameter contains index and element goes into 2nd parameter.

tasawernawaz:
@ziafazal Requested changes addressed. Please review.

ziafazal (approved):

awais786 (approved):

PR #511, opened 2016-12-28
Add mo files for new translations

MatthewPiatetsky:
@edx/ecommerce


rlucioni (approved):
For anyone wondering, the reason for this PR is elaborated on in [this comment](https://github.com/edx/course-discovery/pull/499#issuecomment-269513662). I'm updating the script we use to pull translations that should prevent the need for this in the future.

PR #521, opened 2017-01-04
Certificate type and price fields updated

tasawernawaz:
[ECOM-6087](https://openedx.atlassian.net/browse/ECOM-6087)

* Validate the drop down contains "Choose course type" (default value), "Audit Only", "Verified Certificate" and "Professional Education".
* Price will be required only for verified and professional education.


course_discovery/apps/publisher/forms.py
```diff
@@ -201,15 +201,33 @@ def clean(self):
...
+        if seat_type in [Seat.PROFESSIONAL,  Seat.VERIFIED] and not price:
+            self.add_error('price', _('Only audit seat can be without price.'))
 
         return self.cleaned_data
 
 
 class CustomSeatForm(SeatForm):
     """ Course Seat Form. """
 
+    def __init__(self, *args, **kwargs):
```
waheedahmed, line 213: No need for this ```__init__()```, already being handled in ```BaseCourseForm```'s init at L17.

course_discovery/apps/publisher/forms.py
```diff
@@ -201,15 +201,33 @@ def clean(self):
...
+        super(CustomSeatForm, self).__init__(*args, **kwargs)
+
+        field_classes = 'field-input input-select'
+
+        if 'type' in self.errors:
+            field_classes = '{} has-error'.format(field_classes)
+
+        self.fields['type'].widget.attrs = {'class': field_classes}
+
+    type_choices = [
```
waheedahmed: I think we should change the choices in model instead adding here.

waheedahmed (changes requested):

waheedahmed:
And I think you also need to update the save method [here](https://github.com/edx/course-discovery/blob/c13f0849407917112c293e70179a193c3bbfecc9/course_discovery/apps/publisher/forms.py#L174-L187).

course_discovery/apps/publisher/forms.py
tasawernawaz: We need a Choicefield because modelChoiceField requires queryset. Can we get choices of a field as queryset ?

course_discovery/apps/publisher/forms.py
tasawernawaz: I used it here because we are not adding class `input-select` for `ChoiceField` so make it similar to other dropdown I used it here. The reason why I did not change in base class is other `ChoiceFields` in form have only two values and are shown as radio buttons. But in this case we have a dropdown of `ChoiceField` so I need to override `__init__` here.

tasawernawaz: This method updated to add `empty_label` for Seat Type dropdown.

course_discovery/apps/publisher/models.py
```diff
@@ -295,20 +295,14 @@ def post_back_url(self):
...
     VERIFIED = 'verified'
     PROFESSIONAL = 'professional'
-    NO_ID_PROFESSIONAL = 'no-id-professional'
-    CREDIT = 'credit'
 
     SEAT_TYPE_CHOICES = (
-        (HONOR, _('Honor')),
-        (AUDIT, _('Audit')),
+        (AUDIT, _('Audit Only')),
         (VERIFIED, _('Verified')),
```
waheedahmed, line 308: ```Verified Certificate```

waheedahmed (changes requested):

course_discovery/apps/publisher/forms.py
waheedahmed: Looks good now 👍.

course_discovery/apps/publisher/forms.py
waheedahmed: ```TYPE_CHOICES```

waheedahmed (approved):
One minor comment, otherwise looks good 👍

tasawernawaz:
@bderusha Please review this.

bderusha (approved):

PR #522, opened 2017-01-04
WL-914 | Added staff members in Course data loader.

hasnain-naveed:
@douglashall @afzaledx

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -126,7 +126,7 @@ class CoursesApiDataLoader(AbstractDataLoader):
                         course.canonical_course_run = course_run
                         course.save()
                 if not self.partner.has_marketing_site:
-                    self._populate_course_run_staff(course_run, body.get('instructor_info'))
+                    self._create_or_update_course_run_staff(course_run, body.get('instructor_info'))
             except:  # pylint: disable=bare-except
                 msg = 'An error occurred while updating {course_run} from {api_url}'.format(
                     course_run=course_run_id,
@@ -215,7 +215,7 @@ class CoursesApiDataLoader(AbstractDataLoader):
 
         return defaults
 
-    def _populate_course_run_staff(self, course_run, staff_members):
+    def _create_or_update_course_run_staff(self, course_run, staff_members):
         staff_list = []
         for member in staff_members:
             defaults = {}
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -12,7 +12,7 @@ from course_discovery.apps.core.models import Currency
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, CourseRunPacing
 from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoader
 from course_discovery.apps.course_metadata.models import (
-    Video, Organization, Seat, CourseRun, Program, Course, ProgramType,
+    Video, Organization, Seat, CourseRun, Program, Course, ProgramType, Person, Position,
 )
 
 logger = logging.getLogger(__name__)
@@ -125,6 +125,8 @@ class CoursesApiDataLoader(AbstractDataLoader):
                     if created:
                         course.canonical_course_run = course_run
                         course.save()
+                if not self.partner.has_marketing_site:
+                    self._populate_course_run_staff(course_run, body.get('instructor_info'))
             except:  # pylint: disable=bare-except
                 msg = 'An error occurred while updating {course_run} from {api_url}'.format(
                     course_run=course_run_id,
@@ -212,6 +214,55 @@ class CoursesApiDataLoader(AbstractDataLoader):
             defaults['course'] = course
 
         return defaults
+
+    def _populate_course_run_staff(self, course_run, staff_members):
+        staff_list = []
+        for member in staff_members:
+            defaults = {}
+            name = member.get('name')
+            image = member.get('image')
+            bio = member.get('bio')
+
+            if name:
+                defaults['given_name'] = name
+            if image:
+                defaults['profile_image_url'] = image
+            if bio:
+                defaults['bio'] = self.clean_html(bio)
+
+            try:
+                person, _ = Person.objects.update_or_create(
+                    partner=self.partner,
+                    uuid=member.get('uuid'),
+                    defaults=defaults
+                )
+            except ValueError:
+                logger.error("Unable to populate instructor in %s due to insufficient information.", course_run.key)
+            else:
+                self._set_position_in_org(
+                    person,
+                    organization_key=member.get('organization'),
+                    title=member.get('title')
+                )
+                staff_list.append(person)
+        if staff_list:
+            course_run.staff = staff_list
+            course_run.save()
+
+    def _set_position_in_org(self, person, organization_key=None, title=None):
+        if not organization_key and not title:
+            logger.error('Unable to set position for instructor with uuid %s (title does not exist).', person.uuid)
+        else:
+            defaults = {}
+            if organization_key:
+                defaults['organization'], _ = Organization.objects.get_or_create(
+                    key__iexact=organization_key,
+                    partner=self.partner,
+                    defaults={'key': organization_key}
+                )
+            if title:
+                defaults['title'] = title
+            Position.objects.update_or_create(person=person, defaults=defaults)
 
     def format_course_data(self, body):
         defaults = {
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
@@ -84,6 +84,14 @@ COURSES_API_BODIES = [
         'hidden': False,
         'instructor_info': [
             {
+                'name': 'TestInstructor5',
+                'organization': 'KyotoUx',
+                'title': 'Test Title',
+                'uuid': '431d42a0-7c1f-40a1-b785-0dd11ba8cb33',
+                'image': '/TestUser/image.jpg',
+                'bio': 'TestInstructor1 BIO'
+            },
+            {
                 'name': 'TestInstructor3',
                 'organization': 'KyotoUx',
                 'title': '',
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
@@ -40,6 +40,25 @@ COURSES_API_BODIES = [
         'pacing': 'self',
         'mobile_available': True,
         'hidden': False,
+        'instructor_info': [
+            {
+                'name': 'TestInstructor1',
+                'organization': 'MITx',
+                'title': 'SE',
+                'uuid': '',
+                'image': '/TestUser/image.jpg',
+                'bio': 'TestInstructor1 BIO'
+            },
+            {
+                'name': '',
+                'organization': 'MITx',
+                'title': 'Tester',
+                'uuid': '28508d33-4638-475e-9ec9-87c543db97d2',
+                'image': '/TestInstructor/image.jpg',
+                'bio': 'TestInstructor2 BIO'
+            },
+
+        ],
     },
     {
         'effort': None,
@@ -63,6 +82,26 @@ COURSES_API_BODIES = [
         'pacing': 'instructor,',
         'mobile_available': False,
         'hidden': False,
+        'instructor_info': [
+            {
+                'name': 'TestInstructor3',
+                'organization': 'KyotoUx',
+                'title': '',
+                'uuid': 'ba52060a-8fb9-4278-8f7d-49357fac78a4',
+                'image': '/TestName1/image.jpg',
+                'bio': 'TestInstructor3 Bio'
+
+            },
+            {
+                'name': 'TestInstructor4',
+                'organization': '',
+                'title': '',
+                'uuid': '66910bb5-4b9b-4927-b8e2-75e0b8bfa4cd',
+                'image': '/TestName2/image.jpg',
+                'bio': ''
+
+            },
+        ],
     },
     {
         # Add a second run of KyotoUx+000x (3T2016) to test merging data across
@@ -87,6 +126,7 @@ COURSES_API_BODIES = [
         'start': None,
         'mobile_available': None,
         'hidden': True,
+        'instructor_info': [],
     },
 ]
 
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
@@ -15,7 +15,7 @@ from course_discovery.apps.course_metadata.data_loaders.api import (
 from course_discovery.apps.course_metadata.data_loaders.tests import JSON, JPEG, mock_data
 from course_discovery.apps.course_metadata.data_loaders.tests.mixins import ApiClientTestMixin, DataLoaderTestMixin
 from course_discovery.apps.course_metadata.models import (
-    Course, CourseRun, Organization, Seat, Program, ProgramType,
+    Course, CourseRun, Organization, Seat, Program, ProgramType, Person, Position,
 )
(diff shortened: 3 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -209,6 +210,36 @@ def format_course_run_data(self, body, course=None):
 
         return defaults
 
+    def _update_or_create_course_staff(self, staff_members):
+        staff_list = []
+        for member in staff_members:
+            person, created = Person.objects.update_or_create(
```
clintonb: This will always create a new object. The unique key is UUID. If you want to load people from LMS, you will need to add a UUID field there. You will also need to ensure any instructors defined for edx.org have the UUID back-populate with the value pulled from the marketing site.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -137,13 +137,14 @@ def get_course_run(self, body):
...
         validated_data = self.format_course_run_data(body)
+        validated_data['staff'] = self._update_or_create_course_staff(body["instructor_info"])
         self._update_instance(course_run, validated_data)
 
         logger.info('Processed course run with UUID [%s].', course_run.uuid)
 
     def create_course_run(self, course, body):
         defaults = self.format_course_run_data(body, course=course)
-
+        defaults['staff'] = self._update_or_create_course_staff(body["instructor_info"])
```
clintonb: Single quotes. Also, you don't want to do this if a marketing site is in use.

clintonb (changes requested):

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -209,6 +210,36 @@ def format_course_run_data(self, body, course=None):
...
+                    'bio': self.clean_html(member.get('bio', '')),
+                    'family_name': member.get('family_name', '')
+                }
+            )
+            self._set_position_in_org(
+                person,
+                organization_name=member.get('organization', ''),
+                title=member.get('title', '')
+            )
+            staff_list.append(person)
```
davec-edx: We're doing several things here without any exception handling at this level. Does it create a problem if we fail in the middle of the list of staff members and have created/updated some but not all?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -137,13 +137,14 @@ def get_course_run(self, body):
 
     def update_course_run(self, course_run, body):
         validated_data = self.format_course_run_data(body)
+        validated_data['staff'] = self._update_or_create_course_staff(body["instructor_info"])
```
douglashall: Can we move this [inside](https://github.com/edx/course-discovery/pull/522/files#diff-b225ce3f72b8695dadd76de7d8f77aacR206) `formate_course_run_data` under the marketing_site check?

course_discovery/apps/course_metadata/data_loaders/api.py
douglashall: Same as above, move inside `format_course_run_data`.

course_discovery/apps/course_metadata/data_loaders/api.py
douglashall: I think we may need the flexibility of maintaining separate `Person` objects for the same physical person across partners. Matching on partner/name makes sense to me here.

course_discovery/apps/course_metadata/data_loaders/api.py
douglashall: I think the only thing that could go wrong here is some kind of database connection issue which would cause more then just this function to fail. Recovering from such a failure would just be a matter of running `refresh_course_metadata` again.

course_discovery/apps/course_metadata/data_loaders/api.py
clintonb: I'm not opposed to duplicate people across partners. The issue here is that your unique key is a value that is (a) not indexed, (b) not unique, and (c) not constant. Names change. UUIDs do not. UUID is the publicly-exposed unique identifier for almost every object in this system. That field must be used to avoid issues with subsequent runs of the data loaders.

Also, empty string is not a sane default. At some point we need to enforce a minimum amount of information for each `Person`—full name, bio, image (maybe). If that information is not available, the `Person` should not be imported. I'd rather have no information than incomplete information.

douglashall: Got it. This makes sense now. Thanks for the clarification.

afzaledx:
@clintonb, can you please review this PR again? we've tried to fix the issues you raised.

course_discovery/apps/api/serializers.py
```diff
@@ -718,11 +718,12 @@ class ProgramTypeSerializer(serializers.ModelSerializer):
     Serializer for the Program Types
     """
     logo_image = StdImageSerializerField()
+    faq = FAQSerializer()
```
clintonb: Why is this here? [WL-914](https://openedx.atlassian.net/browse/WL-914) doesn't call for changes to the `ProgramType` model.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
 
         return defaults
 
+    def _populate_course_run_staff(self, course_run, staff_members):
+        if not self.partner.has_marketing_site:
```
clintonb: Move this conditional to `_process_response`.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+                        organization_name=member.get('organization'),
+                        title=member.get('title')
+                    )
+                    staff_list.append(person)
+            if staff_list:
+                course_run.staff = staff_list
+                course_run.save()
+
+    def _populate_course_run_seats(self, course_run, course_modes):
+        if not self.partner.has_marketing_site:
```
clintonb: Move this conditional to `_process_response`.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+            for mode in course_modes:
+                Seat.objects.update_or_create(
+                    course_run=course_run,
+                    type=mode.get('name', '').lower(),
```
clintonb: Empty string is not a safe default. If there is no value, the data should not be ingested. You should log an exception, and make sure you've updated Splunk to alert on it. If the data is bad, we don't want it in our system.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+                    currency=Currency.objects.get(
+                        code=mode.get('currency', '').upper()
```
clintonb: No empty strings.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+                    ),
+                    defaults={
+                        'upgrade_deadline': mode.get('expires', ''),
+                        'price': mode.get('price', '')
```
clintonb: No empty strings. Also, this needs to be a `decimal` type.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+        if not self.partner.has_marketing_site:
+            for mode in course_modes:
+                Seat.objects.update_or_create(
+                    course_run=course_run,
+                    type=mode.get('name', '').lower(),
+                    currency=Currency.objects.get(
+                        code=mode.get('currency', '').upper()
+                    ),
+                    defaults={
+                        'upgrade_deadline': mode.get('expires', ''),
```
clintonb: This must be a `datetime` type or `None`. `str` is not valid.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+            for member in staff_members:
+                if not member.get('uuid'):
+                    logger.error('Unable to import instructor %s in %s (uuid does not exist)',
+                                 member.get('name'), course_run.key)
+                else:
+                    person, _ = Person.objects.update_or_create(
+                        partner=self.partner,
+                        uuid=member.get('uuid'),
+                        defaults={
+                            'given_name': member.get('name', ''),
```
clintonb: There is no need for empty strings in any of these fields. The default of `None` is valid.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
 
         return defaults
 
+    def _populate_course_run_staff(self, course_run, staff_members):
+        if not self.partner.has_marketing_site:
+            staff_list = []
+            for member in staff_members:
+                if not member.get('uuid'):
```
clintonb: Bring this into a separate variable to avoid duplication with the line below.

clintonb: Also, considering that [WL-912](https://openedx.atlassian.net/browse/WL-912) must be merged before this PR is merged, is this check even necessary?

FYI @douglashall @hasnain-naveed the UUID back-population needs to happen for all instructors, not just those for white label.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+                        code=mode.get('currency', '').upper()
+                    ),
+                    defaults={
+                        'upgrade_deadline': mode.get('expires', ''),
+                        'price': mode.get('price', '')
+                    }
+                )
+
+    def _set_position_in_org(self, person, organization_name, title):
+        if organization_name or title:
```
clintonb: Avoid giant conditional blocks like this. Your code is much more legible if you exit early. Also, there is no need to put anything besides UUID in your error messages. That's the primary key, so it's all we care about when debugging.

```python
if not organization_name and not title:
  logger.warning('Unable to create position for Person [%s]. No organization or title provided.', person.uuid)
```

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+                'title': ''
+            }
+
+            if organization_name:
+                defaults['organization'], _ = Organization.objects.get_or_create(
+                    partner=self.partner,
+                    key=organization_name
+                )
+
+            if title:
```
clintonb: This conditional is not necessary.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,73 @@ def format_course_run_data(self, body, course=None):
...
+        if organization_name or title:
+            defaults = {
+                'organization': '',
+                'title': ''
+            }
+
+            if organization_name:
+                defaults['organization'], _ = Organization.objects.get_or_create(
+                    partner=self.partner,
+                    key=organization_name
```
clintonb: What is `organization_name`? The key should, ideally, be the short name (e.g. MITx) instead of the full name (e.g. Massachusetts Institute of Technology). See https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/data_loaders/api.py#L21. Also, the `name` field needs to be set.

clintonb: @douglashall @hasnain-naveed it would be ideal if your changes to the Course API incorporated the exact same data from the Organization API so that we can re-use that data loader code.

course_discovery/apps/course_metadata/data_loaders/api.py
clintonb: Also, if no data is provided, you should remove any existing positions for the person.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -63,6 +100,36 @@
...
+            },
+        ],
+        "course_modes": [
+            {
+                "name": "professional",
+                "currency": "usd",
+                "price": 10,
+                "sku": "",
+                "bulk_sku": "",
+                "expires": None
```
clintonb: At least one of your test objects needs to have this field set to a real date.

clintonb: Also, where is the bad data? We should verify the data loaders behave appropriately when given non-ideal data.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -189,6 +189,33 @@ def assert_course_run_loaded(self, body, partner_has_marketing_site=True):
 
         return course_run
 
+    def assert_staff_loaded(self, body, partner_has_marketing_site=True):
+        if not partner_has_marketing_site:
```
clintonb: These assertions should not have conditionals. If there is no marketing site, don't call the function.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -189,6 +189,33 @@ def assert_course_run_loaded(self, body, partner_has_marketing_site=True):
...
+                            self.assertEqual(person.position.organization.key, instructor.get('organization'))
+                    else:
+                        with self.assertRaises(Position.DoesNotExist):
+                            Position.objects.get(person=person)
+                else:
+                    with self.assertRaises(Person.DoesNotExist):
+                        Person.objects.get(given_name=instructor.get("name"))
+
+    def assert_seats_loaded(self, body, course_run, partner_has_marketing_site=True):
+        if not partner_has_marketing_site:
```
clintonb: Same as above.

course_discovery/apps/course_metadata/migrations/0044_programtype_faq.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
clintonb: Again, why is this model being touched? If it must be modified, let's discuss that in a separate PR. However, we should really have the discussion before any changes are made to this model.

clintonb (changes requested):

course_discovery/apps/course_metadata/migrations/0044_programtype_faq.py
afzaledx: @clintonb , we'll pull the unrelated changes to a new PR. I will try to explain why this field is being added. If you visit https://www.edx.org/micromasters/ritx-project-management and scroll down  a bit  and look at the right column, you'll see the "What is a MicroMasters Program Program?" and the copy below the question. This question+answer is related to the ProgramType and the ProgramType model had no place to hold this info.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -125,6 +125,8 @@ def _process_response(self, response):
                     if created:
                         course.canonical_course_run = course_run
                         course.save()
+                if not self.partner.has_marketing_site:
+                    self._populate_course_run_staff(course_run, body.get('instructor_info'))
```
douglashall: `self._create_or_update_course_run_staff(course_run, body.get('instructor_info'))`

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -213,6 +215,55 @@ def format_course_run_data(self, body, course=None):
 
         return defaults
 
+    def _populate_course_run_staff(self, course_run, staff_members):
```
douglashall: `def _create_or_update_course_run_staff(self, course_run, staff_members):`

douglashall:
No longer needed.

PR #523, opened 2017-01-04
Awais786/ecom 6687 permissions mixins

awais786:
Update the Mix-ins and pass the permission as param to the method.
Update the changestate view code.
Fix broken tests cases.
Update the get_objects_for_user method.

https://openedx.atlassian.net/browse/ECOM-6687

### DIFF
--- a/course_discovery/apps/publisher/api/permissions.py
+++ b/course_discovery/apps/publisher/api/permissions.py
@@ -1,14 +1,21 @@
 from rest_framework.permissions import BasePermission
 
-from course_discovery.apps.publisher.mixins import check_user_course_access
+from course_discovery.apps.publisher.mixins import check_roles_access, check_course_organization_permission
+from course_discovery.apps.publisher.models import OrganizationExtension
 from course_discovery.apps.publisher.utils import is_internal_user
 
 
 class CanViewAssociatedCourse(BasePermission):
-    """ Permission class to check user can view a publisher course. """
+    """ Permission class to check user can view a publisher course or also if
+    user has view permission on organization.
+    """
 
     def has_object_permission(self, request, view, obj):
-        return check_user_course_access(request.user, obj.course)
+        user = request.user
+        return (
+            check_roles_access(user) or
+            check_course_organization_permission(user, obj.course, OrganizationExtension.VIEW_COURSE)
+        )
 
 
 class InternalUserPermission(BasePermission):
--- a/course_discovery/apps/publisher/mixins.py
+++ b/course_discovery/apps/publisher/mixins.py
@@ -2,11 +2,13 @@ from django.contrib.auth.decorators import login_required
 from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.utils.decorators import method_decorator
 
-from course_discovery.apps.publisher.models import Course, Seat, OrganizationExtension
+from course_discovery.apps.publisher.models import Course, Seat
 from course_discovery.apps.publisher.utils import is_publisher_admin, is_internal_user
 
 
 class ViewPermissionMixin(object):
+
+    permission = None
 
     def get_course(self):
         publisher_object = self.get_object()
@@ -19,15 +21,18 @@ class ViewPermissionMixin(object):
 
         return None
 
-    def check_user(self, user):
+    def has_user_access(self, user):
         course = self.get_course()
-        return check_user_course_access(user, course)
+        return (
+            check_roles_access(user) or
+            check_course_organization_permission(user, course, self.permission)
+        )
 
     def permission_failed(self):
         return HttpResponseForbidden()
 
     def dispatch(self, request, *args, **kwargs):
-        if not self.check_user(request.user):
+        if not self.has_user_access(request.user):
             return self.permission_failed()
 
         return super(ViewPermissionMixin, self).dispatch(request, *args, **kwargs)
@@ -65,20 +70,14 @@ def check_roles_access(user):
     return False
 
 
-def check_course_organization_permission(user, course):
+def check_course_organization_permission(user, course, permission):
     """ Return True if user has view permission on organization. """
     if not hasattr(course, 'organizations'):
         return False
 
     return any(
         [
-            user.has_perm(OrganizationExtension.VIEW_COURSE, org.organization_extension)
+            user.has_perm(permission, org.organization_extension)
             for org in course.organizations.all()
         ]
     )
-
-
-def check_user_course_access(user, course):
-    """ Return True if user is admin/internal user or has access permission. """
-
-    return check_roles_access(user) or check_course_organization_permission(user, course)
--- a/course_discovery/apps/publisher/tests/test_utils.py
+++ b/course_discovery/apps/publisher/tests/test_utils.py
@@ -7,9 +7,7 @@ from course_discovery.apps.core.tests.factories import UserFactory
 from course_discovery.apps.publisher.constants import (
     ADMIN_GROUP_NAME, INTERNAL_USER_GROUP_NAME, PARTNER_COORDINATOR_GROUP_NAME
 )
-from course_discovery.apps.publisher.mixins import (
-    check_roles_access, check_course_organization_permission, check_user_course_access
-)
+from course_discovery.apps.publisher.mixins import check_course_organization_permission, check_roles_access
 from course_discovery.apps.publisher.models import OrganizationExtension
 from course_discovery.apps.publisher.tests import factories
 from course_discovery.apps.publisher.utils import (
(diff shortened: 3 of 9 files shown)


awais786:
@bderusha @waheedahmed Please review this PR.

course_discovery/apps/publisher/api/permissions.py
```diff
@@ -1,14 +1,14 @@
...
+from course_discovery.apps.publisher.mixins import check_roles_access
 from course_discovery.apps.publisher.utils import is_internal_user
 
 
 class CanViewAssociatedCourse(BasePermission):
     """ Permission class to check user can view a publisher course. """
 
     def has_object_permission(self, request, view, obj):
-        return check_user_course_access(request.user, obj.course)
+        return check_roles_access(request.user)
```
bderusha: This seems wrong.  Why is the `CanViewAssociatedCourse` class only checking roles?

course_discovery/apps/publisher/mixins.py
```diff
@@ -2,11 +2,12 @@
 from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.utils.decorators import method_decorator
 
-from course_discovery.apps.publisher.models import Course, Seat, OrganizationExtension
+from course_discovery.apps.publisher.models import Course, Seat
 from course_discovery.apps.publisher.utils import is_publisher_admin, is_internal_user
 
 
 class ViewPermissionMixin(object):
+    permission_required = None
```
bderusha: Not sure how I feel about this.  I think it would be better to have a specific view permission.

course_discovery/apps/publisher/mixins.py
```diff
@@ -65,20 +69,17 @@ def check_roles_access(user):
     return False
 
 
-def check_course_organization_permission(user, course):
+def check_course_organization_permission(user, course, permission_required):
```
bderusha: `permission_required` sounds like a boolean.

This should just be called `permission`

course_discovery/apps/publisher/mixins.py
```diff
@@ -19,15 +20,18 @@ def get_course(self):
 
         return None
 
-    def check_user(self, user):
+    def has_user_access(self, user):
         course = self.get_course()
-        return check_user_course_access(user, course)
+        return (
+            check_roles_access(user) or
+            check_course_organization_permission(user, course, self.permission_required)
```
bderusha: This will always return false as written.  Is that the way it is supposed to work?

course_discovery/apps/publisher/tests/test_utils.py
```diff
@@ -8,7 +8,7 @@
     ADMIN_GROUP_NAME, INTERNAL_USER_GROUP_NAME, PARTNER_COORDINATOR_GROUP_NAME
 )
 from course_discovery.apps.publisher.mixins import (
-    check_roles_access, check_course_organization_permission, check_user_course_access
+    check_course_organization_permission, check_roles_access
```
bderusha: This no longer needs to be multi-line

bderusha (changes requested):

course_discovery/apps/publisher/mixins.py
awais786: No it will not return false always. Permission will be set in the view e.g
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/views.py#L243

course_discovery/apps/publisher/api/permissions.py
awais786: This is existing method. I just updated the name. Would you please elaborate ?

course_discovery/apps/publisher/mixins.py
awais786: Do we need multiple permissions mixin e.g 
```
    VIEW_COURSE = 'publisher_view_course'
    VIEW_COURSE_RUN = 'publisher_view_course_run'
```

course_discovery/apps/publisher/api/permissions.py
awais786: Ok I got it. You mean why checking user roles only. Should check organizationExtension permissions also ?

awais786:
@bderusha Please review this again.

course_discovery/apps/publisher/api/permissions.py
bderusha: To maintain functionality it seems like it should be checking VIEW_COURSE permission in addition to checking the role.

bderusha: @waheedahmed can you weigh in on this?

course_discovery/apps/publisher/mixins.py
```diff
@@ -8,6 +8,10 @@
 
 class ViewPermissionMixin(object):
 
+    # default permission is VIEW_COURSE. Different permission can
+    # be set inside the views.
+    permission = OrganizationExtension.VIEW_COURSE
```
bderusha: I see now what you are doing.  Sorry, I think I mispoke before.  You are right that `None` is a good default.

But maybe we should add a check to make sure that `permission` is set by the view.

course_discovery/apps/publisher/mixins.py
```diff
@@ -65,20 +72,17 @@ def check_roles_access(user):
     return False
 
 
-def check_course_organization_permission(user, course):
+def check_course_organization_permission(user, course, permission):
     """ Return True if user has view permission on organization. """
+    if not permission:
+        return False
```
bderusha: Per my comment above:  I think this should raise an exception rather than just return false.  You should not be allowed to check the permission if no permission is passed

bderusha (commented):
A few small comments.  I think this will be good to go once resolved.

course_discovery/apps/publisher/mixins.py
awais786: Right.

awais786: Or we can set the default permission. e.g
`
 permission = OrganizationExtension.VIEW_COURSE
`

course_discovery/apps/publisher/api/permissions.py
waheedahmed: @bderusha You are right, we should check ```VIEW_COURSE``` permission along with role checking here.

course_discovery/apps/publisher/mixins.py
awais786: @bderusha in that case we might need to add a permission in each view.

course_discovery/apps/publisher/mixins.py
awais786: @bderusha in that case we might need to add a permission in each view.

course_discovery/apps/publisher/mixins.py
waheedahmed: If we can set different permissions in views then we should rename this mixin too, with this name it looks like this is only to check view related permission.

waheedahmed: @bderusha What do you think about ^^?

bderusha: I like the idea of making this more generic.  Going back to setting `permission = None`.  Renaming the mixin.  And having it raise an exception if you do not set the `permission` in the view

awais786:
@bderusha Please review this PR again.

bderusha (approved):

bderusha:
Squash commits before merging.  Nice work on this!

awais786:
@tasawernawaz @waheedahmed Please review this PR.

awais786:
I already got 1 approval. But need 2nd approval also.

course_discovery/apps/publisher/mixins.py
```diff
@@ -2,12 +2,14 @@
 from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.utils.decorators import method_decorator
 
-from course_discovery.apps.publisher.models import Course, Seat, OrganizationExtension
+from course_discovery.apps.publisher.models import Course, Seat
 from course_discovery.apps.publisher.utils import is_publisher_admin, is_internal_user
 
 
 class ViewPermissionMixin(object):
```
waheedahmed, line 9: I think we decided to rename this mixin to more generic name because now it can check multiple type of permissions not only the ```VIEW_COURSE``` permission.

waheedahmed (approved):
Just one comment otherwise looks good 👍

course_discovery/apps/publisher/mixins.py
awais786: @waheedahmed I will create a sub-task and do that change in separate PR. This PR already become too big so its good to merge this one and only rename in 2nd PR.

PR #527, opened 2017-01-05
Course edit button permission.

waheedahmed:
[ECOM-6045](https://openedx.atlassian.net/browse/ECOM-6045)

As a user, I would like to get into the edit mode of a course so that I can make changes.

**AC:**

1. **Visible to:**
Validate that this button is visible to all users who can edit.

2. **Actions:**
Validate that when clicking on the "Edit" button, the user is directed to the Course edit page.


waheedahmed:

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1435,3 +1435,21 @@ def test_detail_page_with_admin(self):
         self.user.groups.add(Group.objects.get(name=ADMIN_GROUP_NAME))
         response = self.client.get(self.detail_page_url)
         self.assertEqual(response.status_code, 200)
+
+    def test_details_page_with_edit_permission(self):
+        """ Test that user can see edit button on course detail page. """
+        assign_perm(OrganizationExtension.VIEW_COURSE, self.user, self.organization_extension)
+
+        # Verify that user cannot see edit button without edit permission.
+        self.assert_can_edit_permission()
```
tasawernawaz: I think it will be better if you can pass 'can_edit=False' here instead of default arguments, it will increase readability of the code.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1435,3 +1435,21 @@ def test_detail_page_with_admin(self):
...
+
+        assign_perm(OrganizationExtension.EDIT_COURSE, self.user, self.organization_extension)
```
tasawernawaz, line 1445: Should not we assign permissions to group related to `self.organization_extension` ?

tasawernawaz (approved):
A minor change and a questions otherwise 👍

course_discovery/apps/publisher/tests/test_views.py
waheedahmed: @tasawernawaz Yeah we can also use group here then we have to add user to that group also, it's a same thing we just want to test that user can edit course if he has edit permission and it can be done in both ways.

waheedahmed:
@bderusha Please review.

course_discovery/apps/publisher/models.py
```diff
@@ -428,6 +428,7 @@ def add_course_roles(cls, course, role, user):
 
 class OrganizationExtension(TimeStampedModel):
     """ Organization-Extension relation model. """
+    EDIT_COURSE = 'edit_course'
```
bderusha: `publisher_edit_course`

course_discovery/apps/publisher/models.py
```diff
@@ -428,6 +428,7 @@ def add_course_roles(cls, course, role, user):
...
     EDIT_COURSE_RUN = 'edit_course_run'
```
bderusha: We should also update this one to use the `publisher` prefix

course_discovery/apps/publisher/views.py
```diff
@@ -260,7 +260,13 @@ class CourseDetailView(mixins.LoginRequiredMixin, mixins.ViewPermissionMixin, De
 
     def get_context_data(self, **kwargs):
         context = super(CourseDetailView, self).get_context_data(**kwargs)
+
+        user = self.request.user
         context['comment_object'] = self
+        context['can_edit'] = any(
```
bderusha, line 266: this `any([user.perm(...) for org in organizations.all()])` logic is very common.  We should put this in a shared place.

bderusha (changes requested):

course_discovery/apps/publisher/views.py
waheedahmed: @bderusha Yes, we will update this with ```check_course_organization_permission()``` when its ready to except permissions.

waheedahmed:
@bderusha Addressed your feedback, please take another look.

bderusha (approved):

PR #530, opened 2017-01-05
Upgrade edx-lint to 0.5.2

rlucioni:
This allows us to upgrade from pylint 1.5.4 to 1.6.4.

@nedbat @edx/ecommerce FYI.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -806,7 +806,7 @@ class FlattenedCourseRunWithCourseSerializer(CourseRunSerializer):
                 else:
                     seats[seat.type][key] = SeatSerializer(seat).data[key]
 
-        for credit_attr in seats['credit'].keys():
+        for credit_attr in seats['credit']:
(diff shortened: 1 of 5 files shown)


clintonb:
@nedbat @rlucioni remind me why we are hardcoding to a specific version of Pylint instead of a range, please. Can we add this reasoning to the README in https://github.com/edx/edx-lint for the next time this question arises?

rlucioni:
@clintonb new versions of Pylint often introduce new errors and/or warnings that we don't want to be surprised by in the middle of a build. Pylint has also been known to release broken versions of the package that we don't want to blindly upgrade to. That said, maybe edx-lint could support a range of versions (e.g., pylint>=1.5.4,<1.6.4). @nedbat are there reasons I'm not thinking of for not doing this?

nedbat:
Nope, those are the reasons.  If 1.6.4 changes the warnings produced by 1.6.3, then auto-updating to 1.6.4 will break builds of other repos that fail builds on pylint violations.

PR #539, opened 2017-01-09
Extend CourseRunQuerySet marketable filter to exclude runs without seats

rlucioni:
Runs without seats should not be considered marketable.

ECOM-6768.

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/query.py
+++ b/course_discovery/apps/course_metadata/query.py
@@ -46,13 +46,22 @@ class CourseRunQuerySet(models.QuerySet):
     def marketable(self):
         """ Returns CourseRuns that can be marketed to learners.
 
-         A CourseRun is considered marketable if it has a defined slug and has been published.
+         A CourseRun is considered marketable if it has a defined slug, has seats, and has been published.
 
          Returns:
             QuerySet
          """
 
-        return self.exclude(slug__isnull=True).exclude(slug='').filter(status=CourseRunStatus.Published)
+        return self.exclude(
+            slug__isnull=True
+        ).exclude(
+            slug=''
+        ).exclude(
+            # This will exclude any course run without seats (e.g., CCX runs).
+            seats__isnull=True
+        ).filter(
+            status=CourseRunStatus.Published
+        )
 
 
 class ProgramQuerySet(models.QuerySet):
--- a/course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
@@ -134,14 +134,20 @@ class CatalogViewSetTests(ElasticsearchTestMixin, SerializationMixin, OAuth2Mixi
     def test_courses(self):
         """ Verify the endpoint returns the list of courses contained in the catalog. """
         url = reverse('api:v1:catalog-courses', kwargs={'id': self.catalog.id})
+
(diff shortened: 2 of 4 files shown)


MatthewPiatetsky:
fyi marketing site also needs to be updated to use marketable filter. 
So the seats change will cover all the troublesome course runs?

course_discovery/apps/course_metadata/query.py
```diff
@@ -46,13 +46,22 @@ def active(self):
...
-        return self.exclude(slug__isnull=True).exclude(slug='').filter(status=CourseRunStatus.Published)
+        return self.exclude(
+            slug__isnull=True
+        ).exclude(
+            slug=''
+        ).exclude(
+            # This will exclude any course run without seats (e.g., CCX runs).
+            seats__isnull=True
+        ).filter(
+            status=CourseRunStatus.Published
```
MatthewPiatetsky, line 63: Do we want to allow publishing courses if any of these conditions are true? Or we do for some sort of testing purposes?

rlucioni:
@MatthewPiatetsky this will allow us to exclude any course runs which aren't ready to be purchased. A complete solution also requires use of the `active` filter. Refer to https://github.com/edx/edx-mktg/pull/1080.

course_discovery/apps/course_metadata/query.py
rlucioni: Depends where run publishing is being done from right now. If that's still being done from Drupal, I'm not sure we have a good way of accomplishing this. If the Django admin were being used to control publication, then yes, we could prevent that change from saving if seats weren't present.

rlucioni: Also, not all runs need to have seats (e.g., CCX courses). If for some reason we wanted to use the catalog to designate those as published, we wouldn't want to force the author to add seats.

course_discovery/apps/course_metadata/query.py
```diff
@@ -46,13 +46,22 @@ def active(self):
...
 
-         A CourseRun is considered marketable if it has a defined slug and has been published.
+         A CourseRun is considered marketable if it has a defined slug, has seats, and has been published.
 
          Returns:
             QuerySet
          """
 
-        return self.exclude(slug__isnull=True).exclude(slug='').filter(status=CourseRunStatus.Published)
+        return self.exclude(
```
clintonb, line 55: Nit: My preference is to keep the full method call on one line:

```python
return self.exclude(slug__isnull=True) \ 
               .exclude(...)
```

clintonb (approved):

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: what are CCX courses?

course_discovery/apps/course_metadata/query.py
rlucioni: @clintonb but then you need the slash at the end of the line 🙄.

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky http://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/latest/set_up_course/custom_courses.html.

MatthewPiatetsky:
👍 (once tests fixed)

PR #540, opened 2017-01-10
ECOM-6793 Publish alias when creating new node

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/publishers.py
+++ b/course_discovery/apps/course_metadata/publishers.py
@@ -147,7 +147,8 @@ class MarketingSitePublisher(object):
         node_url = '{root}/node.json'.format(root=api_client.api_url)
         response = api_client.api_session.post(node_url, data=json.dumps(node_data))
         if response.status_code == 201:
-            return response.json()
+            response_json = response.json()
+            return response_json['id']
         else:
             raise ProgramPublisherException("Marketing site page creation failed!")
 
@@ -228,7 +229,7 @@ class MarketingSitePublisher(object):
                 self._edit_node(api_client, node_id, node_data)
             else:
                 # We should create a new node
-                self._create_node(api_client, node_data)
+                node_id = self._create_node(api_client, node_data)
             before_alias = self._make_alias(self.program_before) if self.program_before else None
             new_alias = self._make_alias(program)
             before_slug = self.program_before.marketing_slug if self.program_before else None
--- a/course_discovery/apps/course_metadata/tests/test_publishers.py
+++ b/course_discovery/apps/course_metadata/tests/test_publishers.py
@@ -115,6 +115,12 @@ class MarketingSitePublisherTests(MarketingSitePublisherTestMixin):
                     self.password,
                     self.api_root
                 )
+        self.expected_node = {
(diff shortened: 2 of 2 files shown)


course_discovery/apps/course_metadata/publishers.py
```diff
@@ -229,6 +229,9 @@ def publish_program(self, program):
             else:
                 # We should create a new node
                 self._create_node(api_client, node_data)
+                # Need to recreate session to see new data
+                api_client = self._get_api_client(program)
```
MatthewPiatetsky: I'm not entirely sure why, but I don't get a node id back unless I recreate the api client. I think it may have something to do with session caching or something like that?

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -204,7 +204,7 @@ def test_publish_program_create(self):
         with mock.patch.object(MarketingSitePublisher, '_get_headers', return_value={}):
             with mock.patch.object(MarketingSitePublisher, '_get_form_build_id_and_form_token', return_value={}):
                 publisher.publish_program(self.program)
-                self.assert_responses_call_count(8)
+                self.assert_responses_call_count(13)
```
clintonb: Increasing the number of API calls by over 50% seems like a bad idea.

course_discovery/apps/course_metadata/publishers.py
clintonb: Session caching should not change the API response. What response is returned without these changes?

clintonb (changes requested):
We should do some more investigation before adding more API calls.

course_discovery/apps/course_metadata/tests/test_publishers.py
MatthewPiatetsky: I was getting no node_id back if I didn't recreate the session. Regardless I realized we were creating the node through the node api rather than drupal form hacking so I already had access to what I need. fixed.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -228,7 +228,8 @@ def publish_program(self, program):
                 self._edit_node(api_client, node_id, node_data)
             else:
                 # We should create a new node
-                self._create_node(api_client, node_data)
+                response = self._create_node(api_client, node_data)
```
schenedx: Instead of returning the response with `self._create_node()` function, why not make the change so the function returns `node_id` and other interesting node related information being used in this code block?

MatthewPiatetsky: @schenedx 
I'm not sure I understand your comment, the response has the node id, uuid and node url.
Are you saying to name it something else or to return those things as variables?  
I don't want to return unused variables, but I'm open to only returning the node id

schenedx: Just return the node_id for now is fine with me. I don't see a need to return the whole response.

clintonb (approved):

schenedx (approved):
👍  Yes. LGTM

rlucioni:
@MatthewPiatetsky please delete your merged branches.

PR #544, opened 2017-01-10
In Refresh Course metadata handle wrong course id errors

umar-nawaz:
[ECOM-6457](https://openedx.atlassian.net/browse/ECOM-6457)
Refresh Course Metadata Issues.
@schenedx @MatthewPiatetsky  Please review PR.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -10,6 +10,7 @@ import pytz
 import requests
 from django.db.models import Q
 from django.utils.functional import cached_property
+from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, CourseRunPacing
@@ -389,8 +390,13 @@ class CourseMarketingSiteDataLoader(AbstractMarketingSiteDataLoader):
             except AttributeError:
                 pass
         else:
-            course, created = self.get_or_create_course(data)
-            course_run = self.create_course_run(course, data)
+            created = False
+            try:
+                course, created = self.get_or_create_course(data)
+                course_run = self.create_course_run(course, data)
+            except InvalidKeyError:
+                logger.error('Invalid course key [%s].', data['field_course_id'])
+
             if created:
                 course.canonical_course_run = course_run
                 course.save()
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
@@ -6,11 +6,12 @@ from uuid import UUID
 
 from dateutil import rrule
 import ddt
+from django.test import TestCase
 import mock
+from opaque_keys.edx.keys import CourseKey
 import pytz
 import responses
-from django.test import TestCase
-from opaque_keys.edx.keys import CourseKey
+from testfixtures import LogCapture
 
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, CourseRunPacing
 from course_discovery.apps.course_metadata.data_loaders.marketing_site import (
@@ -426,6 +427,23 @@ class CourseMarketingSiteDataLoaderTests(AbstractMarketingSiteDataLoaderTestMixi
     def test_get_pacing_type(self, data_value, expected_pacing_type):
         data = {'field_course_self_paced': data_value}
         self.assertEqual(self.loader.get_pacing_type(data), expected_pacing_type)
+
+    @ddt.data(
+        {'field_course_id': ''},
+        {'field_course_id': 'EPtestx'},
+        {'field_course_id': 'Paradigms-comput'},
+        {'field_course_id': 'Bio Course ID'}
+    )
+    def test_process_node(self, data):
+        with LogCapture() as l:
+            self.loader.process_node(data)
+            l.check(
+                (
+                    'course_discovery.apps.course_metadata.data_loaders.marketing_site',
+                    'ERROR',
+                    'Invalid course key [{}].'.format(data['field_course_id'])
+                )
+            )
 
     def assert_course_loaded(self, data):
         course = self._get_course(data)


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -389,7 +390,10 @@ def process_node(self, data):
             except AttributeError:
                 pass
         else:
-            course, created = self.get_or_create_course(data)
+            try:
+                course, created = self.get_or_create_course(data)
+            except InvalidKeyError:
+                logger.warning('Invalid course key [%s].', data['field_course_id'])
```
schenedx: Once you logged the warning, should the execution continue in this function? Should we stop the code to create the course_run? Have you tested this code to ensure the data are correct?

schenedx (changes requested):
Please also write a unit test for this scenario

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
umar-nawaz: @schenedx you are right. Does the latest update looks better.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -389,11 +390,15 @@ def process_node(self, data):
...
-            course, created = self.get_or_create_course(data)
-            course_run = self.create_course_run(course, data)
-            if created:
-                course.canonical_course_run = course_run
-                course.save()
+            try:
+                course, created = self.get_or_create_course(data)
+                course_run = self.create_course_run(course, data)
+
+                if created:
```
schenedx: You can wrap the try and except block just before this block, and use `created` flag to determine the validity of the database data creation.

umar-nawaz: @schenedx do you mean like this? 
```
try:
    course, created = self.get_or_create_course(data)
    course_run = self.create_course_run(course, data)
except InvalidKeyError:
    logger.warning('Invalid course key [%s].', data['field_course_id'])

if 'created' in locals() and created: 
    course.save()
```

schenedx: Yes. But instead of checking whether the variable `created` exists in the `locales`, you can just initiate the variable as `False` before the try block

clintonb (commented):
Why do we have invalid data on stage/production? What is being done to fix it? This treats the symptoms rather than the disease.

umar-nawaz:
@clintonb the invalid data came from marketing site. This PR would try to make itself more robust. We can create a ticket to remove such contents ([like this](https://www.edx.org/admin/content?title=derp&author=&status=0&field_course_availability_tid_op=or)) from Drupal and communicate with marketing team. Does it cover your concerns?

clintonb:
@umar-nawaz @schenedx https://openedx.atlassian.net/browse/ECOM-6809

Removing the fake data should be part of the original story. If it's a separate story/task, I fear it will never get done.

schenedx (changes requested):
Please add unit tests

schenedx:
@clintonb I asked @umar-nawaz to also work on the subtask you created ECOM-6809 together with this PR. Please help review this PR once unit test is in place.

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -9,7 +9,10 @@
 import mock
 import pytz
 import responses
+from testfixtures import LogCapture
+
 from django.test import TestCase
+from opaque_keys import InvalidKeyError
```
schenedx: You don't need this import

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -427,6 +430,20 @@ def test_get_pacing_type(self, data_value, expected_pacing_type):
         data = {'field_course_self_paced': data_value}
         self.assertEqual(self.loader.get_pacing_type(data), expected_pacing_type)
 
+    @ddt.data(
+        {'field_course_id': 'EPtestx'}
```
schenedx: Add more test case to use ddt.

schenedx (approved):
👍

schenedx:
@umar-nawaz Looks good to me with the new test case.

umar-nawaz:
@edx/ecommerce Anyone please help review.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -420,7 +424,12 @@ def create_course_run(self, course, data):
         return course_run
 
     def get_or_create_course(self, data):
-        course_run_key = CourseKey.from_string(data['field_course_id'])
+        try:
+            course_run_key = CourseKey.from_string(data['field_course_id'])
+        except InvalidKeyError:
+            logger.warning('Invalid course key [%s].', data['field_course_id'])
```
clintonb: `logger.error`

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -420,7 +424,12 @@ def create_course_run(self, course, data):
         return course_run
 
     def get_or_create_course(self, data):
-        course_run_key = CourseKey.from_string(data['field_course_id'])
+        try:
+            course_run_key = CourseKey.from_string(data['field_course_id'])
+        except InvalidKeyError:
```
clintonb: Do this exception handling in `process_node` and you can drop the if statement there.

clintonb (changes requested):

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
schenedx: That was our first implementation. But then we found that there are other exception cases during `get_or_create_course` function that we should not mask. We figure it is more specific to maintainers if we catch the error here.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
schenedx: @umar-nawaz Since we have removed all the courses that cause this issue, I am agreeing with @clintonb we should log error here now instead of just warning.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
clintonb: If you catch InvalidKeyErrror you won't be masking anything. Returning special values to indicate an error is an anti-pattern. We have exceptions. Let's use them.

umar-nawaz:
@clintonb I have covered your requested changes, please take a look.

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -9,6 +9,7 @@
 import mock
 import pytz
 import responses
+from testfixtures import LogCapture
```
clintonb, line 14: Alphabetize

clintonb (approved):

PR #545, opened 2017-01-10
Add better instructions for running the service on your host

rlucioni:
@clintonb @amangano-edx @davec-edx these changes are from the notes I took today. You should feel free to revise or edit these docs any time.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -1,10 +1,19 @@
 Getting Started
 ===============
 
-If you have not already done so, create/activate a `virtualenv`_ running Python 3.5. Unless otherwise stated, assume all terminal code
-below is executed within the virtualenv.
+This guide will walk you through the steps necessary to run the course-discovery service locally, on your host. In the future, we intend to switch to Docker to avoid the need for many of these manual steps.
 
-.. _virtualenv: https://virtualenvwrapper.readthedocs.org/en/latest/
+
+Install dependencies
+--------------------
+
+If you have not already done so, create and activate a `virtualenv`_ running Python 3.5. We suggest using `virtualenvwrapper`_.
+
+.. code-block:: bash
+
+    $ mkvirtualenv --python=$(which python3)
+
+.. _virtualenvwrapper: https://virtualenvwrapper.readthedocs.org/en/latest/
 
 .. note:: Installing virtualenvwrapper with pip on OS X El Capitan may result
    in a strange OSError due to `a compatibility issue with the six package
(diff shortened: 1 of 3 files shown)


docs/getting_started.rst
```diff
@@ -15,77 +24,95 @@ below is executed within the virtualenv.
...
 
-Configure Partners
+LMS integration
+---------------
+
+To integrate with the LMS, bring up the LMS and navigate to http://localhost:8000/admin/catalog/catalogintegration/. Click "Add catalog integration," and add the URL to the course-discovery service running on your host: ``http://192.168.33.1:8008/api/v1/``.
+
+.. note:: When inside the Vagrant VM, you need to use a special IP to refer to your host. You can find it by running ``ifconfig`` and looking at the IPV4 address for vboxnet0. It's usually 192.168.33.1.
+
+In order for the LMS running in the Vagrant VM to access course-discovery, you will need to run it at 0.0.0.0:8008.
```
amangano-edx, line 96: It might be clearer if we do this on line 76.

docs/getting_started.rst
```diff
@@ -15,77 +24,95 @@ below is executed within the virtualenv.
...
+
+
+Private settings
+----------------
+
+When developing locally, it may be useful to have settings overrides that you do not wish to commit to the repository.
+If you need such overrides, create a file :file:`course_discovery/settings/private.py`. This file's values are
+read by :file:`course_discovery/settings/local.py`, but ignored by Git.
+
+If you are an edX employee, see :ref:`edx-extensions`.
```
amangano-edx, line 110: I think we need more detailed instructions for doing this. Maybe an example.

It wasn't clear to me that it was necessary to import local.py from within private.py in order to modify INSTALLED_APPS.

rlucioni: @amangano-edx what about opening a PR modifying private.py.example to make that clear? That seems like a better place for it.

docs/getting_started.rst
rlucioni: I'd prefer to leave it here. It's only necessary when you're trying to access the service from within the Vagrant VM, which isn't all the time.

rlucioni:
I'm going to go ahead and merge this. Let's build on it if we want to change/add things.

docs/getting_started.rst
```diff
@@ -15,77 +24,95 @@ below is executed within the virtualenv.
 
        $ pip install virtualenvwrapper --upgrade --ignore-installed six
```
davec-edx, line 25: I would flip this block and the above block around, or make them one block.  It makes more sense to me to put them in the order needed and maybe wrap it so it will work as a gist
ie: $(type mkvirtualenv &> /dev/null) || pip install virtualenvwrapper --upgrade --ignore-installed six

rlucioni: @davec-edx addressed in https://github.com/edx/course-discovery/pull/546.

PR #547, opened 2017-01-11
Allow clients requesting courses to filter out unmarketable runs

rlucioni:
Unmarketable runs are defined as those course runs that are published, have seats, and can still be enrolled in, either now or in the future.

ECOM-6768

@edx/ecommerce please review.

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -21,16 +21,6 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = serializers.CourseWithProgramsSerializer
 
     def get_queryset(self):
-        """ List one course
-        ---
-        parameters:
-            - name: include_deleted_programs
-              description: Will include deleted programs in the associated programs array
-              required: false
-              type: integer
-              paramType: query
-              multiple: false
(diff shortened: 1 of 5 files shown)


rlucioni:
This will need to be deployed before changes in https://github.com/edx/edx-mktg/pull/1084 can go out.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -46,6 +46,7 @@ def get_serializer_context(self, *args, **kwargs):
         context.update({
             'exclude_utm': get_query_param(self.request, 'exclude_utm'),
             'include_deleted_programs': get_query_param(self.request, 'include_deleted_programs'),
+            'marketable_course_runs_only': get_query_param(self.request, 'marketable_course_runs_only'),
```
clintonb, line 39: Add this parameter to the docstring for `list()` so this value appears in Swagger (/api-docs). Also, please correct the docstring for `get_queryset()`.

clintonb (approved):

course_discovery/apps/api/v1/views/courses.py
rlucioni: @clintonb updated.

PR #549, opened 2017-01-11
Course Team groups only see the 'In Progress" Tab on the Dashboard.

awais786:
If user is publisher admin or internal user he can view all tabs. Otherwise only in-progress tab is visible only.

https://openedx.atlassian.net/browse/ECOM-6795

PR #550, opened 2017-01-11
Address Pilot Gaps: Course-run Detail Page

tasawernawaz:
[ECOM-6798](https://openedx.atlassian.net/browse/ECOM-6798)


tasawernawaz:
@bderusha Please review it.

bderusha (approved):

course_discovery/apps/publisher/views.py
```diff
@@ -160,6 +161,12 @@ def get_context_data(self, **kwargs):
                 )
             }
         ]
+
+        if mixins.check_roles_access(self.request.user):
```
waheedahmed: Instead doing this you can simply assign like this:
```
context['can_view_tab'] = mixins.check_roles_access(self.request.user)
```

course_discovery/apps/publisher/views.py
```diff
@@ -160,6 +161,12 @@ def get_context_data(self, **kwargs):
...
+            context['can_view_tab'] = True
```
waheedahmed: ```is_internal_user```

course_discovery/templates/publisher/course_run_detail.html
```diff
@@ -58,11 +63,18 @@ <h2 class="hd-2 emphasized">
                 <div id="tab-5" class="tab-content">
                     {% include 'publisher/course_run_detail/_salesforce.html' %}
                 </div>
+              {% else %}
+                <div id="tab-2" class="tab-content active">
+                    {% include 'publisher/course_run_detail/_all.html' %}
```
waheedahmed: Move this outside else above if, no need to add else part.

waheedahmed (changes requested):

course_discovery/apps/publisher/views.py
tasawernawaz: renamed it to 'can_view_all_tabs' like @awais786  is doing in #549

tasawernawaz:
@waheedahmed Feedback addressed you can review again.

waheedahmed (approved):

PR #556, opened 2017-01-12
Changed the marketing_url property to use the slug from the program_t…

davec-edx:
Changed the marketing_url property to use the slug from the program_type object instead of the name of the program_type.  Added a unittest to verify that the URL generated by the marketing_url property not only has the elements we're expecting but that it is a valid URL in general.

ECOM-6813

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -322,8 +322,12 @@ class ProgramTests(MarketingSitePublisherTestMixin):
 
     def test_marketing_url(self):
         """ Verify the property creates a complete marketing URL. """
+        # In test, the factory will likely have the ProgramType's slug and ProgramType's name be the same, we need
+        # to verify that the ProgramType's slug is used and not the ProgramType's name; therefore, set the name to
+        # something obviously different from the name in order to verify that the correct object is being used.
+        self.program.type.slug = '8675309'
         expected = '{root}/{type}/{slug}'.format(root=self.program.partner.marketing_site_url_root.strip('/'),
-                                                 type=self.program.type.name.lower(), slug=self.program.marketing_slug)
+                                                 type=self.program.type.slug, slug=self.program.marketing_slug)
         self.assertEqual(self.program.marketing_url, expected)
 
     def test_marketing_url_without_slug(self):
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -681,7 +681,7 @@ class Program(TimeStampedModel):
     @property
     def marketing_url(self):
         if self.marketing_slug:
-            path = '{type}/{slug}'.format(type=self.type.name.lower(), slug=self.marketing_slug)
+            path = '{type}/{slug}'.format(type=self.type.slug.lower(), slug=self.marketing_slug)
             return urljoin(self.partner.marketing_site_url_root, path)
 
         return None


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -545,13 +548,55 @@ def test_str(self):
         self.assertEqual(str(seat_type), seat_type.name)
 
 
-class ProgramTypeTests(TestCase):
+class ProgramTypeTests(MarketingSitePublisherTestMixin):
```
MatthewPiatetsky: I don't think we need to be testing publishing here.  
We have unit tests for that in test_publishers

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -545,13 +548,55 @@ def test_str(self):
...
+            3, transcript_languages=transcript_languages, course__subjects=subjects)
+        self.courses = [course_run.course for course_run in self.course_runs]
+        self.excluded_course_run = factories.CourseRunFactory(course=self.courses[0])
+        self.program = factories.ProgramFactory(courses=self.courses, excluded_course_runs=[self.excluded_course_run])
+
     def test_str(self):
         program_type = factories.ProgramTypeFactory()
         self.assertEqual(str(program_type), program_type.name)
 
+    def test_marketing_url(self):
```
MatthewPiatetsky: I'm not sure why we want to test this way.  
This seems like a lot of extra complexity.  
Why not just test the property directly and make sure it does the correct thing?
(We have tests that start with test_marketing_url. I would modify those)

davec-edx: This will test specifically that it can generate a valid url for each data type we have.  Testing each one by hand could be done, but when we add new ones we'll have to add them here too so this is just grabbing all of them.  It is comparing the various slugs because if, for example, a slug is missing from the program_type, a URL will be generated and miss that completely and the URL could appear to be correct, doing this I think handles missing and present, correct and incorrect.

course_discovery/apps/course_metadata/tests/test_models.py
davec-edx: I thought I needed to have it in order to generate the URL at the program level.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -545,13 +548,55 @@ def test_str(self):
...
+        """
+        Verify that the URL generated by the marketing_url property contains the expected slugs, and is
+        a valid URL
+        """
+        url_validator = URLValidator()
+
+        self.program.partner.marketing_site_url_root = self.api_root
```
mikedikan: why set up user/pass and all that jazz?  it doesn't look like it's being used below...

davec-edx: Couple reasons: A) I wanted to make sure we're testing all types against programs and generating a URL from it.  B) Thought I needed to in order to do A based on how other unit tests were working.
Very likely it's there because I'm not familiar enough with that area and it is overkill.

davec-edx: Very likely it's there because I'm not familiar enough with that area and it is overkill.

course_discovery/apps/course_metadata/tests/test_models.py
clintonb: Revert. Testing should be on the `Program` model, which already has a test for the `marketing_url` property. Also, we don't necessarily need to have test cases for every one of edX's program types, so you only need to one line of the existing test: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/tests/test_models.py#L326.

course_discovery/apps/course_metadata/tests/test_models.py
mikedikan: let's try killing it and seeing if the tests still pass

clintonb (changes requested):

course_discovery/apps/course_metadata/tests/test_models.py
davec-edx: Changed it so it tested just the property generation itself, not using the generated data, please take another look.

MatthewPiatetsky: I'm not sure what you changed, but this still seems not necessary. 
I think you just need to change the already existing test_marketing_url test inside the ProgramTests (not ProgramTypeTests)

davec-edx: Oh!  I had added it in the tests for program types, didn't catch that you were saying it should be in the programs instead.  I see.  Ok, I adjusted that test to ensure that the slug was different than the name.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -322,8 +323,9 @@ def test_weeks_to_complete_range(self):
 
     def test_marketing_url(self):
         """ Verify the property creates a complete marketing URL. """
+        self.program.type.slug = 'slug.goes.here'
```
MatthewPiatetsky: You can make sure the program type factory creates the program type slug rather than doing it here.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1,5 +1,6 @@
 import itertools
 from decimal import Decimal
+import urllib.parse
```
MatthewPiatetsky: unused import

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -190,13 +190,18 @@ class Meta(object):
         model = ProgramType
 
     name = FuzzyText()
+    slug = ''
```
clintonb: There is no need to modify the factory. This field uses `AutoSlug`, which generates a value from the name field. https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L596-L597

clintonb (changes requested):

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -322,8 +322,10 @@ def test_weeks_to_complete_range(self):
 
     def test_marketing_url(self):
         """ Verify the property creates a complete marketing URL. """
+        # Set slug to something we'll recognized so we can check that it is used and not the name
```
clintonb: `# We set the ProgramType slug to a value that is obviously different from its name so that we can ensure the slug is being used in the URL rather than the name.`

clintonb (approved):

PR #561, opened 2017-01-16
Added error message on saving course-key.

waheedahmed:
[ECOM-6748](https://openedx.atlassian.net/browse/ECOM-6748)

As a PC, when I update course-run key on publisher dashboard, show user friendly messages on the page when errors occur, so I know what corrections I need to make.

screen shot 2017-01-16 at 3 05 14 pm
screen shot 2017-01-16 at 3 04 42 pm
screen shot 2017-01-16 at 3 04 10 pm


waheedahmed:

course_discovery/static/js/publisher/views/dashboard.js
```diff
@@ -38,6 +39,16 @@ $(document).ready(function() {
                 $("#studio-count").html(data_table_studio.rows().count());
                 $studioInstanceSuccess.find(".copy-meta").html(successMessage);
                 $studioInstanceSuccess.css("display", "block");
+            },
+            error: function (response) {
+                if(response.responseJSON.lms_course_id) {
```
awais786, line 44: Instead of adding if else conditions here you can do this in serialiser.

awais786:
Please add test case which check all three responses. You can check it by assertContains in response.

course_discovery/static/js/publisher/views/dashboard.js
waheedahmed: @awais786 How, can you send me some samples?

waheedahmed:
@bderusha Please review.

bderusha (approved):
This looks good, pending @awais786 's comment about moving the `if` block out of the JS and into the serializer.

waheedahmed:
@bderusha I discussed that with @awais786 already and we cannot do that in serializers because there are different type of error's, some are related to permissions and some are validation errors raised by serializer.

awais786 (approved):

PR #562, opened 2017-01-16
Make statement field in corporate endorsement not required

ahsan-ul-haq:
ECOM-6544

@schenedx @rlucioni @umar-nawaz please review


umar-nawaz:
@ahsan-ul-haq These Travis tests are also failing on my PR: https://travis-ci.org/edx/course-discovery/builds/192293317

ahsan-ul-haq:
@umar-nawaz It seems like some sort of translation issue

rlucioni:
@umar-nawaz @ahsan-ul-haq keep an eye on https://github.com/edx/course-discovery/pull/568 for status on a fix for the build issue.

schenedx (changes requested):
Can you please add a new test case to ensure we can insert null into the `statement` field?

ahsan-ul-haq:
@schenedx added test and migration please take a look

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -596,6 +596,14 @@ def setUp(self):
     def test_str(self):
         self.assertEqual(str(self.individual_endorsements), self.corporation_name)
 
+    def test_null_statement(self):
```
clintonb: This is essentially testing Django. This may be one of the few times where I say you don't need tests.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -596,6 +596,14 @@ def setUp(self):
...
+        endorsement = CorporateEndorsement.objects.create(
```
clintonb: In the future, use a factory.

```python
endorsement = CorporateEndorsementFactory(statement=None)
self.assertIsNone(endorsement.statement)
```

schenedx (approved):
@ahsan-ul-haq The test case came out not as I expected. I was wrong to call for a test case. I agree with @clintonb. Go ahead and remove the test case and you should be 👍

PR #564, opened 2017-01-16
Success message remains on course creation.

awais786:
https://openedx.atlassian.net/browse/ECOM-6825


waheedahmed (approved):
Also remove that function```alertTimeout()```, no need of this now.

tasawernawaz (approved):

PR #566, opened 2017-01-17
Use MySQL when running tests on Travis

rlucioni:
We run Docker containers on Travis, including one for MySQL. However, we haven't actually be using it! These changes update the test settings required to make the switch from SQLite to MySQL, fix a variety of bugs in tests that were hidden by SQLite, and introduce new Make targets which make it easy to run tests locally in the same way that Travis runs them. The next time someone tells you that a test suite which formerly ran on SQLite runs on MySQL with no changes, you should be suspicious.

@edx/ecommerce please review. @edx/devops FYI. Similar changes will need to be made for all other projects which switched to Docker-based testing on Travis during the Xenial upgrade.


course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -324,7 +331,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
     function_score = {
         'functions': [
             {'filter': {'term': {'pacing_type_exact': 'self_paced'}}, 'weight': 1.0},
-            {'filter': {'term': {'type_exact': 'micromasters'}}, 'weight': 1.0},
+            {'filter': {'term': {'type_exact': 'MicroMasters'}}, 'weight': 1.0},
```
rlucioni, line 334: @MatthewPiatetsky I'm hoping you can help us understand the ramifications of this change. This test used to pass on SQLite, but fails consistently against MySQL without this change. I suspect the issue is tied to casing differences between the two databases, but haven't been able to pin it down. Why don't the lowercase analyzer and filter handle this?

Makefile
```diff
@@ -1,115 +1,99 @@
...
-	@echo ""
-
-static:
+.PHONY: accept clean clean_static compile_translations detect_changed_source_translations dummy_translations extract_translations \
+	fake_translations help html_coverage migrate open-devstack production-requirements pull_translations quality requirements.js \
+	requirements start-devstack static stop-devstack test travis_down travis_start travis_stop travis_test travis_up validate validate_translations
+
+help: ## Display this help message
+	@echo "Please use \`make <target>\` where <target> is one of"
+	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
```
rlucioni, line 13: I maintain that a package like [Invoke](https://github.com/pyinvoke/invoke) is preferable to this, but this is better than what we had, and I'm not going to switch us to Invoke right now.

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -43,7 +43,8 @@ def setUp(self):
 
         # Create three internal user course roles for internal users against a course
         # so we can test change role assignment on these roles.
-        for user, role in zip(self.other_internal_users, PublisherUserRole.choices):
+        roles = [role for role, __ in PublisherUserRole.choices]
+        for user, role in zip(self.other_internal_users, roles):
```
rlucioni, line 47: If you're confused about why I changed this: SQLite was happily accepting the string representation of the tuples contained in `PublisherUserRole.choices`. SQLite [does not enforce the length of a VARCHAR](http://www.sqlite.org/faq.html#q9). MySQL, on the other hand, was having none of it.

course_discovery/settings/test.py
```diff
@@ -1,24 +1,38 @@
...
+
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': ':memory:',
-        'USER': '',
-        'PASSWORD': '',
-        'HOST': '',
-        'PORT': '',
+        'ENGINE': os.environ.get('DB_ENGINE', 'django.db.backends.sqlite3'),
```
rlucioni, line 27: The defaults here allow the tests to continue running against SQLite. My thinking was that you might want to run a single test without starting the Docker containers. I'm on the fence about removing the defaults altogether and forcing developers to run against MySQL. Thoughts?

requirements/base.txt
```diff
@@ -37,6 +37,7 @@ elasticsearch>=1.0.0,<2.0.0
 html2text==2016.9.19
 jsonfield==1.0.3
 markdown==2.6.6
+mysqlclient==1.3.9
```
rlucioni, line 40: This is now a base requirement because it's necessary both for tests and in production.

.travis/docker-compose-travis.yml
```diff
@@ -35,5 +35,12 @@ services:
       - "es"
       - "memcache"
     environment:
-      TEST_ELASTICSEARCH_URL: "http://es:9200"
+      CONN_MAX_AGE: 60
```
rlucioni, line 38: Note the absence of `ATOMIC_REQUESTS`. We weren't testing with it before, either, even though we currently use it in production. Rather than add it, I'm proposing we stop using it altogether in https://github.com/edx/configuration/pull/3609.

rlucioni:
Failing build is due to an issue with the discovery image. A temporary fix is at https://github.com/edx/course-discovery/pull/568.

feanil:
The changes related to testing on MySQL look good to me, as do the makefile changes, I can't speak to the test changes.  I would have preferred 3 different PRs here honestly as there seem to be 3 sets of changes intermingled here.

Makefile
amangano-edx, line 13: Is this an attempt to programmatically generate the help message? Can you add a comment explaining what this is doing?

rlucioni: Yes, that's what it is. It's borrowed from https://github.com/pydanny/cookiecutter-djangopackage. I don't like how arcane it is, but it's an improvement over what we had before. I'll add a note.

Makefile
```diff
@@ -1,115 +1,100 @@
...
 
-accept:
-	nosetests --with-ignore-docstrings -v acceptance_tests
+travis_up: ## Create containers used to run tests on Travis CI
+	docker-compose -f .travis/docker-compose-travis.yml up -d
 
-make dev.test:
-	coverage run ./manage.py test course_discovery --settings=course_discovery.settings.devstack_test --keepdb
-	coverage report
+travis_start: ## Start containers stopped by `travis_stop`
```
clintonb: I'd rather keep these commands in the Travis file as they were previously, or use a separate Makefile. Aside from your initial testing, I don't see a use case for including these in the primary file.

Makefile
```diff
@@ -1,115 +1,100 @@
 .DEFAULT_GOAL := test
 NODE_BIN=./node_modules/.bin
 
-.PHONY: accept clean compile_translations dummy_translations extract_translations fake_translations help html_coverage \
```
clintonb, line 4: Do all of this cleanup in a separate commit.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -361,34 +361,31 @@ def _assert_records(self, count):
...
-        course_run = course.publisher_course_runs.all()[0]
+
         self.assertEqual(self.course_run.language, course_run.language)
         self.assertEqual(self.course_run.contacted_partner_manager, course_run.contacted_partner_manager)
         self.assertEqual(self.course_run.pacing_type, course_run.pacing_type)
         self.assertEqual(course_run.start.strftime("%Y-%m-%d %H:%M:%S"), self.start_date_time)
         self.assertEqual(course_run.end.strftime("%Y-%m-%d %H:%M:%S"), self.end_date_time)
-        seat = course_run.seats.all()[0]
+
+        seat = course_run.seats.get()
```
clintonb: `.first()` is a bit cleaner and more obvious about what is being returned.

clintonb (approved):

Makefile
rlucioni: I've split out Makefile changes to a separate commit.

clintonb (approved):

PR #567, opened 2017-01-17
Revert "Update translations"

rlucioni:
This reverts commit 82585976e3e2c524bf0e597ba2241c34f8b91476.

https://github.com/edx/course-discovery/pull/559 appears to have introduced a change breaking builds on Travis. This is a check to see if reverting the commit resolves the issue.


rlucioni:
@edx/ecommerce FYI. This issue is affecting builds project-wide, and I can't reproduce it locally. Builds are still failing on this PR, so the problem doesn't appear to be part of the translations commit merged on Sunday. It may be in a dependency.

rlucioni:
This is actually an issue with a newly published discovery image. A temporary fix is at https://github.com/edx/course-discovery/pull/568.

PR #569, opened 2017-01-17
Added the certificate_logo_image_url field to the Organization model

clintonb:
ECOM-6858


clintonb:
@edx/ecommerce please review

course_discovery/apps/course_metadata/models.py
```diff
@@ -162,6 +162,10 @@ class Organization(TimeStampedModel):
     homepage_url = models.URLField(max_length=255, null=True, blank=True)
     logo_image_url = models.URLField(null=True, blank=True)
     banner_image_url = models.URLField(null=True, blank=True)
+    certificate_logo_image_url = models.URLField(
+        null=True, blank=True, help_text=_('Logo to be displayed on certificates. If this logo is the same as '
```
rlucioni, line 166: I'm in favor of making this field required. Does it mean we don't need to perform the check we discussed at render-time?

rlucioni (commented):
Looks like you need to add a migration.

course_discovery/apps/course_metadata/models.py
clintonb: We cannot make this required. Not all organizations have certificates.

rlucioni: My bad, ignore me. I didn't see `blank=True`.

clintonb:
@edx/ecommerce please review

rlucioni (approved):

MatthewPiatetsky:
👍 after rebase

schenedx (approved):
👍

PR #572, opened 2017-01-18
Loading Organization certificate logo image from Organizations API

clintonb:
ECOM-6860

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -48,15 +48,19 @@ class OrganizationsApiDataLoader(AbstractDataLoader):
 
     def update_organization(self, body):
         key = body['short_name']
+        logo = body['logo']
+
         defaults = {
             'key': key,
-            'partner': self.partner
+            'partner': self.partner,
+            'certificate_logo_image_url': logo,
         }
+
         if not self.partner.has_marketing_site:
             defaults.update({
                 'name': body['name'],
                 'description': body['description'],
-                'logo_image_url': body['logo']
+                'logo_image_url': logo,
             })
 
         Organization.objects.update_or_create(key__iexact=key, defaults=defaults)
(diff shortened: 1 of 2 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -48,15 +48,19 @@ def ingest(self):
...
+            'partner': self.partner,
+            'certificate_logo_image_url': logo,
         }
+
         if not self.partner.has_marketing_site:
             defaults.update({
                 'name': body['name'],
                 'description': body['description'],
-                'logo_image_url': body['logo']
+                'logo_image_url': logo,
```
MatthewPiatetsky, line 63: Should this one come from both marketing site and organizations api?  
If this existing behavior is what we want, maybe we should clarify on the ticket to make sure everyone is on the same page. The story seems to refer to this logo directly as the marketing site logo, which is confusing if it can also come from the organizations API

MatthewPiatetsky:
👍

course_discovery/apps/course_metadata/data_loaders/api.py
clintonb: The certificate logo should come from the Organizations API, as we discuss multiple times over multiple conversations.

clintonb: This is also mentioned in the ticket.

MatthewPiatetsky: My comment was about the logo_image_url not the certificate_logo_image_url, I realize this was existing functionality before your PR, but the confusing part for me were ACs like this one
```
Verify the marketing site logo can be retrieved using the Catalog API.
```
The ticket refers to the logo other than the certificate one as the marketing site logo, but it seems it can also come from LMS and be overridden by marketing site? If that is the case, want to make sure Jasper and others understand

clintonb: Understand what? The existing logo already comes from the marketing site. This simply adds a logo from the Organizations API.

rlucioni:
Belated 👍

PR #573, opened 2017-01-18
Add ProxiedPagination class

rlucioni:
ProxiedPagination proxies to either DRF's PageNumberPagination or LimitOffsetPagination. If no query parameters are passed, proxies to LimitOffsetPagination by default to preserve pre-existing behavior.

ECOM-6834

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/pagination.py
+++ b/course_discovery/apps/api/pagination.py
@@ -1,5 +1,104 @@
-from rest_framework import pagination
+from rest_framework.pagination import (
+    PageNumberPagination as BasePageNumberPagination,
+    LimitOffsetPagination,
+)
 
 
-class PageNumberPagination(pagination.PageNumberPagination):
+class PageNumberPagination(BasePageNumberPagination):
     page_size_query_param = 'page_size'
+
+
+class ProxiedCall:
+    """
+    Utility class used in conjunction with ProxiedPagination to route method
+    calls between pagination classes.
+    """
+    def __init__(self, proxy, method_name):
+        self.proxy = proxy
+        self.method_name = method_name
+
+    def __call__(self, *args, **kwargs):
+        try:
+            # Currently, the only methods on DRF pagination classes which accept
+            # requests as positional arguments expect them as the second positional
+            # argument. Hence, we expect the same to be true here. If DRF's pagination
+            # classes are changed such that this is no longer true, code below will
+            # break, loudly, so that the maintainer realizes there is a problem.
+            request = args[1]
+        except IndexError:
+            request = None
+
+        paginator = self._get_paginator(request=request if request else False)
+
+        # Look up the method and call it.
+        return getattr(paginator, self.method_name)(*args, **kwargs)
+
+    def _get_paginator(self, request=False):
+        for paginator, query_param in self.proxy.paginators:
+            # DRF's ListModelMixin calls paginate_queryset() prior to get_paginated_response(),
+            # storing the original request on the paginator's `request` attribute. If the paginator
+            # has this attribute, it means we've routed a previous paginate_queryset() call
+            # to it and should continue using it.
+            is_request_stored = hasattr(paginator, 'request')
+
+            # If a request is available, look for the presence of a query parameter
+            # indicating that we should use this paginator.
+            is_query_param_present = request and request.query_params.get(query_param)  # pylint: disable=no-member
+
+            if is_request_stored or is_query_param_present:
+                return paginator
+
+        # If we don't have a stored request or query parameter to go off of,
+        # default to the last paginator in the list on the proxy. To preserve
+        # pre-existing behavior, this is currently LimitOffsetPagination.
+        return paginator  # pylint: disable=undefined-loop-variable
+
+
+class ProxiedPagination:
+    """
+    Pagination class which proxies to either DRF's PageNumberPagination or
+    LimitOffsetPagination.
+
+    The following are all valid:
+
+        http://api.example.org/accounts/?page=4
+        http://api.example.org/accounts/?page=4&page_size=100
+        http://api.example.org/accounts/?limit=100
+        http://api.example.org/accounts/?offset=400&limit=100
+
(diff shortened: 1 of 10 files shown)


course_discovery/apps/api/pagination.py
```diff
@@ -1,5 +1,101 @@
...
+            # indicating that we should use this paginator.
+            query_param = getattr(paginator, query_param_name)
+            is_query_param_present = request and request.query_params.get(query_param)  # pylint: disable=no-member
+
+            if is_request_stored or is_query_param_present:
+                return paginator
+
+        # If we don't have a stored request or query parameter to go off of,
+        # we default to pre-existing behavior (LimitOffsetPagination).
+        return self.proxy.paginators[-1][0]
```
MatthewPiatetsky: What is this doing? Is there a clearer way to write this? Using a dictionary or named tuple maybe

MatthewPiatetsky: When is the other element of the tuple used?

rlucioni: @MatthewPiatetsky we can't use a dictionary if we want a consistent ordering, and a named tuple seems like overkill. This retrieves the last paginator from the list on the proxy. The other element is used in `ProxiedCall` when looking for query parameters.

rlucioni: @MatthewPiatetsky comment updated to make this a little more clear.

rlucioni: @MatthewPiatetsky actually, I realized a cleaner way of doing this is to just use the variable I'm already working with in the for loop.

MatthewPiatetsky:
Would the caller need to specify the ProxiedPagination? Or do we need to specify it for the programs view?

rlucioni:
@MatthewPiatetsky no to both, see the settings change being made. `ProxiedPagination` is set as the value of `DEFAULT_PAGINATION_CLASS`, making it the default for all endpoints.

MatthewPiatetsky:
ah missed that thanks

course_discovery/apps/api/pagination.py
```diff
@@ -1,5 +1,102 @@
-from rest_framework import pagination
+from rest_framework.pagination import (
+    PageNumberPagination as BasePageNumberPagination,
```
MatthewPiatetsky, line 2: What does this change do?

MatthewPiatetsky:
👍

course_discovery/apps/api/pagination.py
rlucioni: @MatthewPiatetsky imports the class with a different name, so that I can use the name below for our overridden version of the class.

course_discovery/apps/api/tests/test_pagination.py
```diff
@@ -0,0 +1,93 @@
...
+        # Access an attribute common to both PageNumberPagination and LimitOffsetPagination.
+        self.assertEqual(
+            self.proxied_paginator.display_page_controls,
+            self.limit_offset_paginator.display_page_controls
+        )
+
+        # Access an attribute found on neither PageNumberPagination nor LimitOffsetPagination.
+        with self.assertRaises(AttributeError):
+            zach = self.proxied_paginator
+            zach.cool  # pylint: disable=pointless-statement
```
mikedikan, line 93: 😆

course_discovery/apps/api/pagination.py
```diff
@@ -1,5 +1,102 @@
...
+        http://api.example.org/accounts/?page=4
+        http://api.example.org/accounts/?page=4&page_size=100
+        http://api.example.org/accounts/?limit=100
+        http://api.example.org/accounts/?offset=400&limit=100
+
+    If no query parameters are passed, proxies to LimitOffsetPagination by default.
+    """
+    def __init__(self):
+        self.paginators = [
+            (PageNumberPagination(), 'page_query_param'),
```
clintonb: Nit: Instead of using `getattr(paginator, query_param_name)` above pass the attribute here.

course_discovery/settings/base.py
```diff
@@ -314,7 +314,7 @@
         'edx_rest_framework_extensions.authentication.JwtAuthentication',
         'rest_framework.authentication.BasicAuthentication',
     ),
-    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',
+    'DEFAULT_PAGINATION_CLASS': 'course_discovery.apps.api.pagination.ProxiedPagination',
```
clintonb: Explicitly set the pagination class on the `v1` views. It's tedious; but, I don't want `v2` to mistakenly support both forms of pagination when we should only support one.

clintonb (approved):
See my comment on DEFAULT_PAGINATION_CLASS.

course_discovery/apps/api/pagination.py
rlucioni: @clintonb you mean using `page` and `limit` here directly? You can change those query params by overriding these attributes on the paginators, so this route seemed more robust. Unnecessary complexity?

course_discovery/settings/base.py
rlucioni: I'm confused. I thought the goal was for both API versions to support both forms of pagination. This pagination change is being made to avoid the need for changes on the LMS.

If a new MicroMaster is published today, the LMS will try and fail to traverse a paginated v1 response (it mistakenly assumes the API is using `PageNumberPagination`). We're not touching that code on the LMS, so the same failure will occur when the LMS makes requests to v2 if `ProxiedPagination` isn't used for both versions. This change isn't backwards-incompatible; why not use it for both?

rlucioni: We caught up about this offline. There's currently only one version of this API (I was confusing it with the credentials API). The idea is to prevent `LimitOffsetPagination` from being supported by future versions of this API. We can achieve that by setting `PageNumberPagination` as the default here and setting `ProxiedPagination` explicitly on existing views.

clintonb (approved):

PR #574, opened 2017-01-19
Added new role Partner Manager (PM) and its group.

waheedahmed:
[ECOM-6545](https://openedx.atlassian.net/browse/ECOM-6545)

As a Publisher user I need to manage PM (Partner Manager) Role.

**AC:**

- Validate that we have a Partner Manager role
- Validate that we have a group for Partner Manager

screen shot 2017-01-19 at 2 51 51 pm


waheedahmed:

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -33,7 +33,7 @@ def setUp(self):
 
         self.internal_user_group.user_set.add(self.internal_user)
         self.other_internal_users = []
-        for __ in range(3):
+        for __ in range(4):
```
tasawernawaz: Can we use `UserFactory().batch()` here instead of `for loop`.

tasawernawaz (approved):
Just a minor question otherwise LGTM.

waheedahmed:
@bderusha Please review.

bderusha (approved):

PR #577, opened 2017-01-20
Using custom Python Social Auth Strategy

clintonb:
This strategy incorporates all of the settings we previously set manually (but clearly forgot to set for this project and the cookiecutter template).

ECOM-6842

### DIFF
--- a/course_discovery/settings/local.py
+++ b/course_discovery/settings/local.py
@@ -35,6 +35,8 @@ EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 INTERNAL_IPS = ('127.0.0.1',)
 
 # AUTHENTICATION
+SOCIAL_AUTH_REDIRECT_IS_HTTPS = False
+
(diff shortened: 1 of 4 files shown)


clintonb:
@edx/ecommerce please review

course_discovery/settings/local.py
```diff
@@ -35,6 +35,8 @@
 INTERNAL_IPS = ('127.0.0.1',)
 
 # AUTHENTICATION
+SOCIAL_AUTH_REDIRECT_IS_HTTPS = False
```
davec-edx, line 38: Do we need to do this in the devstack settings file as well?

rlucioni (approved):

course_discovery/settings/local.py
rlucioni: @davec-edx I think so, nice catch.

clintonb: Done.

PR #579, opened 2017-01-20
Add migration to populate ElasticsearchBoostConfig

amangano-edx:
This PR adds a migration to populate the ElasticsearchBoostConfig with it's initial config, which was taken from production on January 20, 2017. We want to move towards updating this config via migrations instead of django admin, as doing so helps us to ensure the correct config is used in all environments.

@edx/ecommerce please review.


course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
...
+    """Create or update the ElasticsearchBoostConfig instance."""
+
+    # The imported ElasticsearchBoostConfig class shouldn't be used directly to modify the config, as it might be
+    # the wrong version. Instead, we should get the model from the versioned app registry, as described in 
+    # https://docs.djangoproject.com/en/1.8/ref/migration-operations/#runpython
+    model = apps.get_model("edx_haystack_extensions", "ElasticsearchBoostConfig")
+
+    # Create or update the config using what's currently in production, as of January 20, 2017.
+    record = model(pk=ElasticsearchBoostConfig.SINGLETON_INSTANCE_PRIMARY_KEY)
+    record.function_score = {
```
davec-edx: Comment says this is creating or updating...  This is replacing function_score.  Is the comment correct?

amangano-edx: It might be a little misleading. What's happening is that it's creating or updating the ElasticsearchBoostConfig record in the DB, and replacing what's in the function_score column with the new value.

davec-edx: I see, can you update the comment then :-)

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.11 on 2017-01-18 18:26
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
+
+def forward_func(apps, schema_editor):
```
rlucioni: Give this a name indicating what it does. For example, `create_or_update_boost_config`.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
...
+    """Create or update the ElasticsearchBoostConfig instance."""
+
+    # The imported ElasticsearchBoostConfig class shouldn't be used directly to modify the config, as it might be
+    # the wrong version. Instead, we should get the model from the versioned app registry, as described in 
+    # https://docs.djangoproject.com/en/1.8/ref/migration-operations/#runpython
+    model = apps.get_model("edx_haystack_extensions", "ElasticsearchBoostConfig")
```
rlucioni: By convention, this variable is given the same name as the model. This would become:
```py
ElasticsearchBoostConfig = apps.get_model('edx_haystack_extensions', 'ElasticsearchBoostConfig')
```
Nitpick: stick with single quotes to stay consistent with the rest of the project, here and throughout this migration.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Create or update the config using what's currently in production, as of January 20, 2017.
+    record = model(pk=ElasticsearchBoostConfig.SINGLETON_INSTANCE_PRIMARY_KEY)
```
rlucioni: Instead of relying on a hardcoded primary key, take advantage of the fact that this is a singleton model and use `first()`:
```py
boost_config = ElasticsearchBoostConfig.objects.first()
```

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
...
+    record.save()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('edx_haystack_extensions', '0001_squashed_0002_auto_20160826_1750'),
+    ]
+
+    operations = [
+        migrations.RunPython(forward_func, migrations.RunPython.noop)
```
rlucioni: When possible, I suggest being explicit about keyword arguments to aid readability. This would make more sense to someone with less context using the full
```py
migrations.RunPython(create_or_update_boost_config, reverse_code=migrations.RunPython.noop)
```

rlucioni (changes requested):

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.11 on 2017-01-18 18:26
+from __future__ import unicode_literals
```
rlucioni: We're working in Python 3 here, so this can be removed.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: There is nothing in the DB enforcing that there can be just one row. I'm using the SINGLETON_INSTANCE_PRIMARY_KEY, which is used explicitly by `solo` to load the singleton model from the DB, to protect against the possibility that duplicate rows could have be inserted.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: This was auto-generated, but I will remove if we're sure that it's not needed.

amangano-edx: Actually, I might not have been using the course-discovery venv when I auto-generated this. I will re-generate it in the venv.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
rlucioni: I see that. This works, I just think using `first()` is cleaner. Ultimately your call, but just because the library is using a hardcoded primary key doesn't mean we have to. If I'd written the package I would have used `get()` instead of hardcoding anything.

amangano-edx: I'm not advocating using the key just because that's what the library does, I'm advocating it because I feel it is better to be explicit about which record we want to modify. 

In future versions of that library you can configure the primary key to be whatever you want, and I would encourage us to explicitly set the key if we ever upgrade.

rlucioni: @amangano-edx we run the latest version of the library. According to the [docs](https://github.com/lazybird/django-solo#use-cases) (scroll down a bit), setting of the `singleton_instance_id` is only suggested when converting an existing model to a singleton. Otherwise, it doesn't make a difference.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,40 @@
...
+
+from django.db import migrations
+
+from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
```
clintonb: You should not import models in migrations.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
davec-edx: With the current code creating a new instance of the ElasticsearchBoostConfig class.  When the save is performed, will all fields/columns be overwritten with default values?   ie, we're starting with a brand new row with all defaults, changing the function_score field, then performing a save.  Won't that lose any other field values in the db?  I would suggest using the .first as well, which will slice and only return the first object, or do a get(pk=1) if that's the preferred way.

course_discovery/apps/edx_haystack_extensions/models.py
```diff
@@ -9,6 +9,11 @@ class ElasticsearchBoostConfig(SingletonModel):
     This includes a default JSON config for the function_score.
     """
 
+    # The primary key used to save the solo record to the DB.
+    # This is hardcoded to 1 in https://github.com/lazybird/django-solo/blob/1.1.2/solo/models.py
+    # but may be configurable in later releases.
+    SINGLETON_INSTANCE_PRIMARY_KEY = 1
```
clintonb: This is not necessary. Regardless of what instance you retrieve from the database, the `save()` method will always set the primary key to 1: https://github.com/lazybird/django-solo/blob/1.1.2/solo/models.py#L18.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: Hmm, this auto-generated the same way even when within the venv. I will remove this include in any case.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
clintonb: `boost_config = ElasticsearchBoostConfig.get_solo()`

clintonb (changes requested):

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: @clintonb that doesn't work when looking up the model from the versioned registry. None of the model methods are available, just the properties.

course_discovery/apps/edx_haystack_extensions/models.py
amangano-edx: I prefer using the explicit primary key used by the library when looking up or creating the initial record, especially when the model methods like `get_solo` are not available, as they are not when `apps.get_model("edx_haystack_extensions", "ElasticsearchBoostConfig")` is used to lookup the model in migrations.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: Necessary to retrieve the primary key for lookup/creation of the model.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
clintonb: Ignore this line please. It's irrelevant.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: The versioned app registry returns a model of type `<class '__fake__.ElasticsearchBoostConfig'>`

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
clintonb: `boost_config, __ = ElasticsearchBoostConfig.objects.get_or_create(pk=1)`

I see no need for the constant.

amangano-edx: IMO the constant better communicates what is going on here, as in the future someone might look at these migrations and question why we hardcode the value 1. It doesn't hurt anything. With that being said, if everyone really doesn't like its use or the use of the explicit primary key in this migration file, I will change it.

amangano-edx:
@clintonb @rlucioni @davec-edx I've remove the primary key constant and made many of the other changes that were requested. Please take another look when you have a minute.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.11 on 2017-01-18 18:26
+from django.db import migrations
+
+def create_or_update_config_record(apps, schema_editor):
+    """Create or update the ElasticsearchBoostConfig record."""
+
+    # Get the model from the versioned app registry to ensure the correct version is used, as described in 
+    # https://docs.djangoproject.com/en/1.8/ref/migration-operations/#runpython
+    ElasticsearchBoostConfig = apps.get_model("edx_haystack_extensions", "ElasticsearchBoostConfig")
```
clintonb: Single quotes here, and elsewhere.

davec-edx:
Is the filename correct?  I already see a migration step 0002 -- I realize that the rest of the step is different but shouldn't this be 0003?

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,37 @@
...
+    ElasticsearchBoostConfig.objects.update_or_create(
+        # The `solo` library uses 1 for the PrimaryKey to create/lookup the singleton record
+        # See https://github.com/lazybird/django-solo/blob/1.1.2/solo/models.py
+        pk=1,
+
+        # These values were taken from production on January 20, 2017.
```
clintonb, line 16: Is this your new config, or the existing? This migration should include whichever configuration we want to be present after closing ECOM-6715. There's no need for a second migration.

rlucioni:
@davec-edx it's correct. The original migrations (1 and 2) were squashed. You can tell by looking at the `replaces` field on `0001_squashed_0002_auto_20160826_1750.py`.

clintonb:
@davec-edx the previous two migrations were squashed. `0002_` is the correct prefix.

rlucioni (approved):

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
amangano-edx: This is the existing config. A second (and third migration after Matt's changes) doesn't hurt anything. It is valuable to be able to see how the config has evolved over time in case there are any problems that need to be addressed. It is also valuable, IMO to set a common initial state before we start introducing new modifications to the config.

davec-edx:
@rlucioni @clintonb After we squash and let that go through, we don't remove the squashed commits and the 'replaces' directive?

clintonb (approved):

clintonb:
@davec-edx we will eventually. The thought was that we will give any open source users, who haven't run migrations since squashing, time to run that before we do away with them. There is no rush to get rid of them.

course_discovery/apps/edx_haystack_extensions/migrations/0002_auto_20170118_1826.py
```diff
@@ -0,0 +1,39 @@
...
+        defaults={
+            'function_score': {
+                'boost_mode': 'sum',
+                'boost': 1.0,
+                'score_mode': 'sum',
+                'functions': [
+                    {'filter': {'term': {'pacing_type_exact': 'self_paced'}}, 'weight': 1.0},
+                    {'filter': {'term': {'type_exact': 'micromasters'}}, 'weight': 1.0},
```
rlucioni: @amangano-edx are we sure lowercase `micromasters` is going to give us the boost we're looking for here? The [tests](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_search.py#L331) use camel cased `MicroMasters`.

amangano-edx: I'm not 100% sure, but this value was pulled directly from the production config in Django admin.

MatthewPiatetsky: I suspect this should actually be MicroMasters. Could be the config in production was wrong, if that is the case probably another good reason to have it in the migration.

amangano-edx: This does seem to be case sensitive. Therefore, if we know that we will be storing `"MicroMasters"` in the `type_exact` field, this will need to change.

MatthewPiatetsky: you should change it, I have Professional Certificate on my branch

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -304,7 +304,7 @@ def test_empty_query(self):
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
         self.assertListEqual(response_data['objects']['results'],
-                             [self.serialize_course_run(course_run), self.serialize_program(program)])
```
amangano-edx, line 307: Order changed here because the start date boosting config also boosts documents with a null start date.

course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
```diff
@@ -46,7 +46,7 @@ def test_without_authentication(self):
 
     def test_affiliate_with_supported_seats(self):
         """ Verify that endpoint returns course runs for verified and professional seats only. """
-        with self.assertNumQueries(11):
```
amangano-edx, line 49: All of these query counts reduced by 3 because by default, the `solo` library saves a new record when the `ElasticsearchBoostConfig.get_solo` method is called for the first time, resulting in 3 SQL statements for `SAVEPOINT`, `INSERT`, and `RELEASE SAVEPOINT`.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -434,20 +423,20 @@ def test_micromasters_boosting(self):
...
-            title=title + "1",
+            title=title + "1", status=ProgramStatus.Active,
+            type=ProgramType.objects.get(name='XSeries'), partner=self.partner
+        )
+        ProgramFactory(
+            title=title + "2",
             status=ProgramStatus.Active,
             type=ProgramType.objects.get(name='MicroMasters'),
             partner=self.partner
         )
```
amangano-edx, line 433: Changed the order here because this test was passing even though the boost wasn't being properly applied.

MatthewPiatetsky:
👍 once tests pass/conflicts resolved

PR #583, opened 2017-01-23
Fixed autocomplete queryset for LanguageTag.

waheedahmed:
[ECOM-6885](https://openedx.atlassian.net/browse/ECOM-6885)

autocomplete queryset expects ```pk``` field but ```LanguageTag``` have ```code``` as primary key instead of ```id```.

### DIFF
--- a/course_discovery/apps/publisher/mixins.py
+++ b/course_discovery/apps/publisher/mixins.py
@@ -1,4 +1,6 @@
 from functools import wraps
+
+from dal import autocomplete
 from django.contrib.auth.decorators import login_required
 from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.utils.decorators import method_decorator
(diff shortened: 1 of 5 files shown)


awais786 (approved):

waheedahmed:
@bderusha Please review.

course_discovery/apps/publisher/mixins.py
```diff
@@ -108,3 +111,13 @@ class PublisherUserRequiredMixin(object):
     @method_decorator(publisher_user_required)
     def dispatch(self, request, *args, **kwargs):
         return super(PublisherUserRequiredMixin, self).dispatch(request, *args, **kwargs)
+
+
+class LanguageModelSelect2Multiple(autocomplete.ModelSelect2Multiple):
```
bderusha, line 116: Please add a comment as to why this is necessary

bderusha (approved):
👍 once comment addressed

course_discovery/apps/publisher/mixins.py
waheedahmed: Comment added.

PR #587, opened 2017-01-23
Update the marketing api loader to login properly

schenedx:
ECOM-6880
Instead of relying on the user page after login to be 200, check whether we can now access the admin panel to determine if we logged in.

@edx/ecommerce Please review

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -51,7 +51,9 @@ class AbstractMarketingSiteDataLoader(AbstractDataLoader):
         login_url = '{root}/user'.format(root=self.api_url)
         response = session.post(login_url, data=login_data)
         expected_url = '{root}/users/{username}'.format(root=self.api_url, username=username)
-        if not (response.status_code == 200 and response.url == expected_url):
+        admin_url = '{root}/admin'.format(root=self.api_url)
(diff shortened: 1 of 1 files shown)


MatthewPiatetsky:
@schenedx Since you are using the same approach as in the publishers, is there a way we can share the code for making the marketing site client somewhere?
That way if we need to change this in the future, we can ensure all the login strategies continue to work

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -51,7 +51,9 @@ def api_client(self):
         login_url = '{root}/user'.format(root=self.api_url)
         response = session.post(login_url, data=login_data)
         expected_url = '{root}/users/{username}'.format(root=self.api_url, username=username)
-        if not (response.status_code == 200 and response.url == expected_url):
+        admin_url = '{root}/admin'.format(root=self.api_url)
```
mikedikan, line 54: I think this is fine, but can you add a comment to say this is a temporary measure - we should fix this with https://openedx.atlassian.net/browse/ECOM-6496

schenedx:
@MatthewPiatetsky I created https://github.com/edx/course-discovery/pull/589 with your call to refactor. As you can see, the effort is pretty extensive.
If you think that PR is the right way to go, please help review that PR.

schenedx:
Closed because of https://github.com/edx/course-discovery/pull/589

PR #589, opened 2017-01-23
Refactor to use MarketingSiteAPIClient on both publisher and data_loader

schenedx:


MatthewPiatetsky:
A bit hard to figure out some of the changes, but I think you were primarily just refactoring into a shared file.
👍 once commits are squashed and conflicts fixed

schenedx:
@edx/ecommerce Can I get one more reviewer on this?
We should merge this and deploy as soon as possible. I found out that this problem exists on both stage and production. This means, we haven't refreshed the course_metadata from marketing site for a while now.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1,3 +1,6 @@
+import requests
```
clintonb: This should be grouped with other third-party imports.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -63,3 +66,76 @@ def custom_render_variations(file_name, variations, storage, replace=True):
 
     # to prevent default behaviour
     return False
+
+
+class MarketingSiteAPIClientException(Exception):
```
clintonb: Exceptions should be in their own file, `exceptions.py`

clintonb (approved):

course_discovery/apps/course_metadata/admin.py
```diff
@@ -123,7 +124,7 @@ def save_model(self, request, obj, form, change):
         try:
             super().save_model(request, obj, form, change)
             self.save_error = False
-        except ProgramPublisherException:
+        except (ProgramPublisherException, MarketingSiteAPIClientException):
```
davec-edx, line 127: I realize it wasn't before, but should we be logging the exception?  Additionally, where there are now multiple types of exceptions, should we log at least the type?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -63,3 +66,76 @@ def custom_render_variations(file_name, variations, storage, replace=True):
...
+    """
+    The marketing site API client we can use to communicate with the marketing site
+    """
+    username = None
+    password = None
+    api_url = None
+
+    def __init__(self, marketing_site_api_username, marketing_site_api_password, api_url):
+        if not (marketing_site_api_username and marketing_site_api_password):
+            raise Exception('Marketing Site API credentials are not properly configured!')
```
davec-edx: Should this be a MarketingSiteAPIClientException instead of a general exception?

schenedx: Good catch. Will update

course_discovery/apps/course_metadata/admin.py
schenedx: We do this already by calling `logger.exception`. See documentation on https://docs.python.org/3/library/logging.html

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -29,3 +31,84 @@ def test_upload_to(self, path, field, ext):
...
+        self.mock_admin_response(200)
+        session = self.api_client.init_session
+        self.assert_responses_call_count(3)
+        self.assertIsNotNone(session)
+
+    @responses.activate
+    def test_init_session_failed(self):
+        self.mock_login_response(500)
+        self.mock_admin_response(500)
+        with self.assertRaises(Exception):
```
davec-edx: Are some of these tests testing for `MarketingSiteAPIClientException`?  We're catching all exceptions, would be better to watch for just the expected ones.

PR #591, opened 2017-01-24
Add course_overridden field to CourseRun

rlucioni:
This field can be used to explicitly denote whether a CourseRun has had its course relation manually overridden.

ECOM-6890

@edx/ecommerce


course_discovery/apps/course_metadata/models.py
```diff
@@ -374,6 +374,10 @@ class CourseRun(TimeStampedModel):
     slug = models.CharField(max_length=255, blank=True, null=True, db_index=True)
     hidden = models.BooleanField(default=False)
     mobile_available = models.BooleanField(default=False)
+    course_overridden = models.BooleanField(
+        default=False,
+        help_text=_('Whether the course relation has been manually overridden.')
```
clintonb: `Indicates whether...`

Also, update translations.

clintonb (approved):

course_discovery/apps/course_metadata/models.py
rlucioni: Updated both. Thanks for the reminder.

PR #594, opened 2017-01-24
ECOM-6780 Boost Professional Certificates above XSeries

MatthewPiatetsky:
@edx/ecommerce


course_discovery/apps/edx_haystack_extensions/migrations/0003_auto_20170124_1834.py
```diff
@@ -0,0 +1,30 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
+
+def add_professional_certificate_boosting(apps, schema_editor):
+    """ Add professional certificate boosting to the ElasticsearchBoostConfig."""
+
+    model = apps.get_model("edx_haystack_extensions", "ElasticsearchBoostConfig")
```
clintonb: Single quotes here and elsewhere.

course_discovery/apps/edx_haystack_extensions/migrations/0003_auto_20170124_1834.py
```diff
@@ -0,0 +1,30 @@
...
+    record = model.objects.get(pk=1)
+    professional_certificate_boost = {
+        "filter": {"term": {"type_exact": "Professional Certificate"}},
+        "weight": 1.0
+    }
+    if professional_certificate_boost not in record.function_score['functions']:
```
clintonb: I foresee issues with this if someone manually modifies the boost configuration before this migration is run. We can avoid these issues by updating the entire config, instead of trying just add this one item to a list.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -418,23 +419,24 @@ def test_exception(self):
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.data, ["The 'q' querystring parameter is required for searching."])
 
-    def test_micromasters_boosting(self):
+    @ddt.data('MicroMasters', 'Professional Certificate')
+    def test_micromasters_boosting(self, program_type):
```
rlucioni: `test_program_type_boosting`

course_discovery/apps/edx_haystack_extensions/migrations/0003_auto_20170124_1834.py
```diff
@@ -0,0 +1,42 @@
...
+                'score_mode': 'sum',
+                'functions': [
+                    {'filter': {'term': {'pacing_type_exact': 'self_paced'}}, 'weight': 1.0},
+                    {'filter': {'term': {'type_exact': 'Professional Certificate'}}, 'weight': 1.0},
+                    {'filter': {'term': {'type_exact': 'MicroMasters'}}, 'weight': 1.0},
+                    {'linear': {'start': {'origin': 'now', 'decay': 0.95, 'scale': '1d'}}, 'weight': 5.0}
+                ]
+            }
+        }
+    )
```
amangano-edx, line 30: This will overwrite the entire config. I thought we decided to just push into the list of functions since the first migration takes care of initializing the config?

course_discovery/apps/edx_haystack_extensions/migrations/0003_auto_20170124_1834.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.11 on 2017-01-24 18:34
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
```
amangano-edx: This is not needed.

course_discovery/apps/edx_haystack_extensions/migrations/0003_auto_20170124_1834.py
amangano-edx: Here is the approach that I took: https://github.com/edx/course-discovery/pull/595/files#diff-d127eaee43a2905eb5f83ebbee353a3a. I'm not sure which is better, but it might be worth having a discussion and coming to a consensus before merging anything.

clintonb: I recommend pushing the entire confit every time. This will ensure any manual changes that happen (but should not be made) do not cause issues with these migrations.

amangano-edx: One issue I could see with pushing the entire config each time is that multiple people could be working on migrations to adjust the boost at the same time (as @MatthewPiatetsky and I are now). In such a situation, it is possible that both developers could merge their changes without being aware of the other, and upon deploy only one set of changes would make it to production (the last migration would overwrite the earlier one).

MatthewPiatetsky: @amangano-edx for that particular issue, i think it isn't really a problem because one person will have to merge first and then they will have a merge conflict

amangano-edx: @MatthewPiatetsky are you sure? Wouldn't the migrations be in different files and therefore they wouldn't confict?

MatthewPiatetsky: we can test it, but i think it may be a problem if you try to run migrations and there are two different migrations in the same app depending on the same previous migration

rlucioni: @amangano-edx GitHub shouldn't allow the situation you're describing. @MatthewPiatetsky's right. If you and I both attempt to add a migration to the same app at the same time - let's call it `0002_foo.py` - and you merge first, I won't be able to merge my PR until I rebase onto master. GitHub requires that your branch be up to date before master before allowing you to merge. The conflict should become apparent at that point.

amangano-edx: @rlucioni the names of the files will different, since they contain the date that the migration was generated. Thus, I believe when you merge git might not detect the conflict since the files are actually named differently.

clintonb: I estimate the odds of our needing to make future simultaneous changes to the boosting configuration being low. My justification for replacing the entire value is below:

1. There are no surprises if someone manually modifies the configuration and removes an expected field.
2. If I need to figure out what _should_ be in production, I need to look at a single migration.
3. When we, eventually, squash these migrations I only need to look at the latest.

There are downsides:

1. Manual changes are overridden.
2. If we are modifying migrations simultaneously, we will need to be a bit more vigilant.

The advantages outweigh the disadvantages to me.

clintonb (approved):

course_discovery/apps/edx_haystack_extensions/migrations/0003_auto_20170124_1834.py
MatthewPiatetsky: @rlucioni @amangano-edx the files won't conflict but I was saying I think the process of running migrations will fail (though we'll see if i'm right). 
I think due to clinton's point 2 I agree with this approach

amangano-edx: I'll preface this by saying that I'm fine with overwriting the entire config each time if that's what everyone wants to do.

I disagree that the advantages outweigh the disadvantages here. The main advantage you've described is developer convenience. That is important, but one could argue that it's still pretty easy to just run the migrations locally and look at the table if you need to see what the production state should be.

The disadvantages have the potential to be more of a nuisance. If someone manually changes the config (which I know we discourage, but is currently allowed) and we overwrite it with a new migration, we could inconvenience a lot of stakeholders. Also, as this table evolves, each subsequent migration had the potential to be bigger and bigger, which increases the possibility for developer error (copy from wrong file, accidentally deleting something important, etc). 

Again, I'm fine with moving forward like this, but IMHO, I think it'd be better and safer to do incremental migrations.

MatthewPiatetsky: I disagree that it is just developer convenience. We want to be able to easily tell what our boosting settings are for making changes. We don't want to have to run migrations and look at the database to figure that out. Even product people might want to see a link to the configuration in the github web interface to know what kind of things we're boosting. 
A lot of our search changes with search 2.0, typeahead, etc... have been problematic due to a lack of clear understanding regarding what we were actually doing/what the business rules were. For this reason I think simpler communication would be better. The migrations here are in an edx app, so I don't think non-edx people would be running this, and if edx people mess with that migration without notifying us (which shouldn't happen) it will be at their own risk.

PR #595, opened 2017-01-24
Ecom 6715 boosting for enrollable paid seats

amangano-edx:
Adds new fields to CourseRunIndex and migration to add configuration to boost CourseRuns for which it is possible for a user to enroll and purchase an upgrade.

JIRA: https://openedx.atlassian.net/browse/ECOM-6715
@edx/ecommerce


course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -1,6 +1,8 @@
 import datetime
 import json
 import urllib.parse
+from unittest.mock import patch
```
rlucioni: We install mock from PyPI to stay ahead of the version in the standard library. Consider using that version to stay consistent with other tests.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -1,6 +1,8 @@
...
+import pytz
```
rlucioni: `pytz` is a third-party library (not part of the standard library) and should be grouped as such.

course_discovery/apps/course_metadata/models.py
```diff
@@ -381,6 +381,47 @@ class CourseRun(TimeStampedModel):
...
+        return self.seats.exclude(type__in=Seat.SEATS_WITH_PREREQUISITES).filter(price__gt=0.0)
+
+    def has_enrollable_paid_seats(self):
+        """
+        Return a boolean indicating whether or not enrollable paid Seats (Seats with price > 0 and no prerequisites)
+        are available for this CourseRun.
+        """
+        return len(self._enrollable_paid_seats()) > 0
+
+    def get_paid_seat_enrollment_end(self):
```
rlucioni, line 398: These methods can all be properties (they don't take any arguments). You might want to change the name of this method to `paid_seat_enrollment_end`.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -186,6 +186,83 @@ def test_exclude_deleted_program_types(self):
...
+        Verify that has_enrollable_paid_seats is True when CourseRun has Seats with price > 0 and no prerequisites.
+        """
+        course_run = factories.CourseRunFactory.create()
+        for seat_type, price in seat_config:
+            factories.SeatFactory.create(course_run=course_run, type=seat_type, price=price)
+        self.assertEqual(course_run.has_enrollable_paid_seats(), expected_result)
+
+    @ddt.data(
+        # Case 1: Return None when there are no enrollable paid Seats.
+        ([('audit', 0, None)], '2016-12-31 00:00:00Z', '2016-08-31 00:00:00Z', None),
```
rlucioni, line 214: Can you clean up this data by using `datetime` and `timedelta` objects? Using string representations everywhere doesn't seem as clean (and it's hard to read).

course_discovery/apps/edx_haystack_extensions/migrations/0004_auto_20170126_1510.py
```diff
@@ -0,0 +1,64 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def add_enrollable_paid_seat_boosting(apps, schema_editor):
+    """Add enrollable paid Seat boosting function to ElasticsearchBoostConfig instance."""
+
+    # Get the model from the versioned app registry to ensure the correct version is used, as described in
+    # https://docs.djangoproject.com/en/1.8/ref/migration-operations/#runpython
+    ElasticsearchBoostConfig = apps.get_model("edx_haystack_extensions", "ElasticsearchBoostConfig")
```
rlucioni: Single quotes here and everywhere, please.

rlucioni (commented):
Looks nice @amangano-edx.

course_discovery/apps/course_metadata/models.py
amangano-edx: I thought about making them properties, but I couldn't think of a good reason for doing so since:
- Only the prepare* methods in search_index.py need access to them.
- I wasn't sure if the use of `@property` properly communicates to developers that these calls involve DB calls and therefore could be expensive.
- Using `@cached_property` would break our ability to update, then reindex a record after those fields have been accessed. If we were to use cached_property, we'd probably still want to have non-cached getter methods that can be used whenever we need to re-index an instance.

rlucioni: Your call. I tend to turn methods into properties when I see my code littered with empty function calls.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -460,6 +462,49 @@ def test_self_paced_boosting(self):
...
+            with patch.object(CourseRun, 'get_paid_seat_enrollment_end', return_value=paid_seat_enrollment_end):
+                test_record = CourseRunFactory(
+                    title=title + "2", course__partner=self.partner, start=course_start, pacing_type=pacing_type
+                )
+
+        response = self.get_typeahead_response(title)
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+
+        expected_title = test_record.title if expects_boost else control_record.title
```
MatthewPiatetsky: You may want to add a comment clarifying this (that without the boost the first one will deterministically get returned first)

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: If for the time being this is used only for boosting then not having a property is probably fine

Since currently this is only being used for reindexing, we shouldn't expect load issues, but in similar cases for properties we've used prefetching (you can't prefetch the property, but you can prefetch all the things the property depends on)

course_discovery/apps/edx_haystack_extensions/migrations/0004_auto_20170126_1510.py
```diff
@@ -0,0 +1,64 @@
...
+                    #       - The course run has at least one enrollable paid Seat (has_enrollable_paid_seats is True)
+                    # AND one of the following two conditions are true
+                    #       - The paid_seat_enrollment_end is unspecified.
+                    #       - The paid_seat_enrollment_end is in the future.
+                    # We apply a weight of 1.0 to match the boost given for self paced courses.
+                    {
+                        "filter": {
+                            "bool": {
+                                "must": [
+                                    {"exists": {"field": "has_enrollable_paid_seats"}},
```
MatthewPiatetsky: not really sure about how these queries work, but is this necessary?

course_discovery/apps/course_metadata/models.py
```diff
@@ -381,6 +381,47 @@ class CourseRun(TimeStampedModel):
 
     objects = CourseRunQuerySet.as_manager()
 
+    def _enrollable_paid_seats(self):
+        """
+        Return a QuerySet that may be used to fetch the enrollable paid Seats (Seats with price > 0 and no
+        prerequisites) associated with this CourseRun.
+        """
+        return self.seats.exclude(type__in=Seat.SEATS_WITH_PREREQUISITES).filter(price__gt=0.0)
```
davec-edx, line 389: We're making calls to this for multiple reasons; 1) see if there are any seats and 2) to get the full list of seats.  Should there be a switch to determine if we slice the data or not?  There's no need to scan the whole table and get all the seats if we just care that there's one.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -460,6 +462,49 @@ def test_self_paced_boosting(self):
         response_data = response.json()
         self.assertEqual(response_data['course_runs'][0]['title'], title + "2")
 
+    @ddt.data(
+        # Case 1: Should not get boost if has_enrollable_paid_seats is False, has_enrollable_paid_seats is None or
+        #   paid_seat_enrollment_end is in the past.
+        (False, None, False),
+        (None, None, False),
+        (True, datetime.datetime.now(pytz.timezone('utc')) - datetime.timedelta(days=15), False),
+        (True, datetime.datetime.now(pytz.timezone('utc')) - datetime.timedelta(minutes=1), False),
```
MatthewPiatetsky: I don't think the two different date conditions add anything useful, they are testing the same case

course_discovery/apps/course_metadata/models.py
amangano-edx: My understanding is that calling `len(self.seats.exclude(type__in=Seat.SEATS_WITH_PREREQUISITES).filter(price__gt=0.0))` would do a `SELECT COUNT` query. We actually care about filtering out Seats with prerequisites and price <= 0 for both the count query and the query to get the list of Seats.

course_discovery/apps/api/v1/tests/test_views/test_search.py
amangano-edx: The second one is intended to test same-day behavior, but since we're not doing anything special for that case I guess it's not really needed.

course_discovery/apps/course_metadata/models.py
davec-edx: Sure, but you don't want a count, you want a limit.  ie, return after 1.

amangano-edx: It appears I was wrong about the behavior of `len` on a QuerySet (https://docs.djangoproject.com/en/1.8/ref/models/querysets/). I will change this.

course_discovery/apps/course_metadata/models.py
amangano-edx: I think I will leave them as is, absent a clear need for them to be properties. If we decide to use this logic in other places, we can change it at that time.

course_discovery/apps/course_metadata/tests/test_models.py
amangano-edx: I don't really agree about this. It's easier to see which dates are actually being used when using Strings than it is with `datetimes` and `timedeltas`. Furthermore, using `datetimes` and `timedeltas` here will lengthen all of these lines, making them more difficult to read IMHO.

MatthewPiatetsky:
reminder, could you fixup to one commit, rebase and add ticket number to 3rd line of commit message

course_discovery/apps/edx_haystack_extensions/migrations/0004_auto_20170126_1510.py
amangano-edx: Elasticsearch seems to behave strangely when filtering on `null` fields, (CourseRuns with `null` values for `start` get a full boost), so to be safe I added this as an additional check. At the very least, I don't think it harms anything.

course_discovery/apps/course_metadata/models.py
amangano-edx: Changed.

amangano-edx:
@rlucioni, @davec-edx, @MatthewPiatetsky I believe I've addressed your comments. Please take one more look when you have a minute.

amangano-edx:
@MatthewPiatetsky I will fixup to one commit just prior to merging

MatthewPiatetsky:
@amangano-edx I'm good with the code 👍 I'll leave further code comments to others if they have them. However, I don't have as good an understanding of the business requirements. Were there specific examples that Jasper wanted to see ordered a certain way? If there are, I would recommend reviewing with Jasper. IMO with the search changes a functional type review is just as important as the code review

course_discovery/apps/edx_haystack_extensions/migrations/0004_auto_20170126_1510.py
davec-edx: Still should be single quotes

course_discovery/apps/edx_haystack_extensions/migrations/0004_auto_20170126_1510.py
davec-edx: Single quotes

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -568,3 +529,81 @@ def test_typeahead_partner_filter(self):
...
+
+        search_results = SearchQuerySet().models(CourseRun).all()
+        self.assertEqual(2, len(search_results))
+        if expects_boost:
+            self.assertGreater(search_results[0].score, search_results[1].score)
+            self.assertEqual(test_record.title, search_results[0].title)
+        else:
+            self.assertEqual(search_results[0].score, search_results[1].score)
+
+    @ddt.data('MicroMasters', 'Professional Certificate')
```
rlucioni, line 600: Seeing this here makes me wonder if this test actually belongs in the `edx_catalog_extensions` app.

rlucioni (approved):

amangano-edx:
Note: We need to rebuild the search index after this is deployed.

PR #596, opened 2017-01-24
Allow live changes to JS and CSS in devstack

clintonb:
Fixes https://github.com/edx/devstack/issues/37


clintonb:
FYI @roderickmorales @bderusha

clintonb:
See https://github.com/edx/course-discovery/pull/593

PR #599, opened 2017-01-25
ECOM-4959 Enable configuration of synonyms for search and typeahead

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -54,6 +54,46 @@ class LoginMixin:
         self.client.login(username=self.user.username, password=USER_PASSWORD)
 
 
+class SynonymTestMixin:
+
+    def test_org_synonyms(self):
+        """ Test that synonyms work for organization names """
+        title = 'UniversityX'
+        authoring_organizations = [OrganizationFactory(name='University')]
+        CourseRunFactory(
+            title=title,
+            course__partner=self.partner,
+            authoring_organizations=authoring_organizations
+        )
+        ProgramFactory(title=title, partner=self.partner, authoring_organizations=authoring_organizations)
+        response1 = self.process_response({'q': title})
+        response2 = self.process_response({'q': 'University'})
+        self.assertDictEqual(response1, response2)
+
+    def test_title_synonyms(self):
+        """ Test that synonyms work for terms in the title """
+        CourseRunFactory(title='HTML', course__partner=self.partner)
+        ProgramFactory(title='HTML', partner=self.partner)
+        response1 = self.process_response({'q': 'HTML5'})
+        response2 = self.process_response({'q': 'HTML'})
+        self.assertDictEqual(response1, response2)
+
+    def test_special_character_synonyms(self):
+        """ Test that synonyms work with special characters (non ascii) """
+        ProgramFactory(title='spanish', partner=self.partner)
+        response1 = self.process_response({'q': 'spanish'})
+        response2 = self.process_response({'q': 'español'})
+        self.assertDictEqual(response1, response2)
+
+    def test_stemmed_synonyms(self):
+        """ Test that synonyms work with stemming from the snowball analyzer """
+        title = 'Running'
+        ProgramFactory(title=title, partner=self.partner)
+        response1 = self.process_response({'q': 'running'})
+        response2 = self.process_response({'q': 'jogging'})
+        self.assertDictEqual(response1, response2)
+
+
 class DefaultPartnerMixin:
     def setUp(self):
         super(DefaultPartnerMixin, self).setUp()
@@ -67,7 +107,7 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
     faceted_path = reverse('api:v1:search-course_runs-facets')
     list_path = reverse('api:v1:search-course_runs-list')
 
-    def get_search_response(self, query=None, faceted=False):
+    def get_response(self, query=None, faceted=False):
         qs = ''
 
         if query:
@@ -77,11 +117,16 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
         url = '{path}?{qs}'.format(path=path, qs=qs)
         return self.client.get(url)
 
+    def process_response(self, response):
+        response = self.get_response(response).json()
+        self.assertTrue(response['objects']['count'])
+        return response['objects']
+
     @ddt.data(True, False)
     def test_authentication(self, faceted):
         """ Verify the endpoint requires authentication. """
         self.client.logout()
-        response = self.get_search_response(faceted=faceted)
+        response = self.get_response(faceted=faceted)
         self.assertEqual(response.status_code, 403)
 
     def test_search(self):
@@ -105,7 +150,7 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
         # Generate data that should be indexed and returned by the query
         course_run = CourseRunFactory(course__partner=self.partner, course__title='Software Testing',
                                       status=CourseRunStatus.Published)
-        response = self.get_search_response('software', faceted=faceted)
+        response = self.get_response('software', faceted=faceted)
 
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
@@ -149,7 +194,7 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
         upcoming = CourseRunFactory(course__partner=self.partner, start=now + datetime.timedelta(days=61),
                                     end=now + datetime.timedelta(days=90), status=CourseRunStatus.Published)
 
-        response = self.get_search_response(faceted=True)
+        response = self.get_response(faceted=True)
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
 
@@ -228,7 +273,7 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
         ProgramFactory(courses=course_list, status=ProgramStatus.Active, excluded_course_runs=excluded_course_run_list)
 
         with self.assertNumQueries(6):
-            response = self.get_search_response('software', faceted=False)
+            response = self.get_response('software', faceted=False)
 
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
@@ -251,7 +296,7 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
         ProgramFactory(courses=[course_run.course], status=program_status)
 
         with self.assertNumQueries(8):
-            response = self.get_search_response('software', faceted=False)
+            response = self.get_response('software', faceted=False)
 
             self.assertEqual(response.status_code, 200)
             response_data = json.loads(response.content.decode('utf-8'))
@@ -268,14 +313,23 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
 
 
 @ddt.ddt
-class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase):
+class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin,
+                             SynonymTestMixin, APITestCase):
     path = reverse('api:v1:search-all-facets')
 
-    def get_search_response(self, querystring=None):
-        querystring = querystring or {}
-        qs = urllib.parse.urlencode(querystring)
+    def get_response(self, query=None):
+        qs = ''
+
+        if query:
+            qs = urllib.parse.urlencode(query)
+
         url = '{path}?{qs}'.format(path=self.path, qs=qs)
         return self.client.get(url)
+
+    def process_response(self, response):
+        response = self.get_response(response).json()
+        self.assertTrue(response['objects']['count'])
+        return response['objects']
 
     def test_results_only_include_published_objects(self):
         """ Verify the search results only include items with status set to 'Published'. """
@@ -286,7 +340,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
         course_run = CourseRunFactory(course__partner=self.partner, status=CourseRunStatus.Published)
         program = ProgramFactory(partner=self.partner, status=ProgramStatus.Active)
 
-        response = self.get_search_response()
+        response = self.get_response()
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
         self.assertListEqual(
@@ -301,7 +355,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
 
         self.assertEqual(CourseRun.objects.get(hidden=True), hidden_run)
 
-        response = self.get_search_response()
+        response = self.get_response()
         data = json.loads(response.content.decode('utf-8'))
         self.assertEqual(
             data['objects']['results'],
@@ -321,7 +375,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
         self.assertNotEqual(other_program.partner.short_code, self.partner.short_code)
         self.assertNotEqual(other_course_run.course.partner.short_code, self.partner.short_code)
 
-        response = self.get_search_response()
+        response = self.get_response()
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
         self.assertListEqual(
@@ -330,7 +384,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
         )
 
         # Filter results by partner
-        response = self.get_search_response({'partner': other_partner.short_code})
+        response = self.get_response({'partner': other_partner.short_code})
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
         self.assertListEqual(response_data['objects']['results'],
@@ -342,7 +396,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
         course_run = CourseRunFactory(course__partner=self.partner, status=CourseRunStatus.Published)
         program = ProgramFactory(partner=self.partner, status=ProgramStatus.Active)
 
-        response = self.get_search_response({'q': '', 'content_type': ['courserun', 'program']})
+        response = self.get_response({'q': '', 'content_type': ['courserun', 'program']})
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.content.decode('utf-8'))
         self.assertListEqual(response_data['objects']['results'],
@@ -358,7 +412,7 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
         upcoming = CourseRunFactory(course__partner=self.partner, start=now + datetime.timedelta(weeks=4))
         course_run_keys = [course_run.key for course_run in [archived, current, starting_soon, upcoming]]
 
-        response = self.get_search_response({"ordering": ordering})
+        response = self.get_response({"ordering": ordering})
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data['objects']['count'], 4)
 
@@ -368,24 +422,28 @@ class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin
 
 
 class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin, LoginMixin, ElasticsearchTestMixin,
-                               APITestCase):
+                               SynonymTestMixin, APITestCase):
     path = reverse('api:v1:search-typeahead')
 
-    def get_typeahead_response(self, query=None, partner=None):
-        qs = ''
-        query_dict = {'q': query, 'partner': partner or self.partner.short_code}
-        if query:
-            qs = urllib.parse.urlencode(query_dict)
+    def get_response(self, query=None, partner=None):
+        query_dict = query or {}
+        query_dict.update({'partner': partner or self.partner.short_code})
+        qs = urllib.parse.urlencode(query_dict)
 
         url = '{path}?{qs}'.format(path=self.path, qs=qs)
         return self.client.get(url)
+
+    def process_response(self, response):
+        response = self.get_response(response).json()
+        self.assertTrue(response['course_runs'] or response['programs'])
+        return response
 
     def test_typeahead(self):
         """ Test typeahead response. """
         title = "Python"
         course_run = CourseRunFactory(title=title, course__partner=self.partner)
         program = ProgramFactory(title=title, status=ProgramStatus.Active, partner=self.partner)
-        response = self.get_typeahead_response(title)
+        response = self.get_response({'q': title})
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
         self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
@@ -398,7 +456,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
         for i in range(RESULT_COUNT + 1):
             CourseRunFactory(title="{}{}".format(title, i), course__partner=self.partner)
             ProgramFactory(title="{}{}".format(title, i), status=ProgramStatus.Active, partner=self.partner)
-        response = self.get_typeahead_response(title)
+        response = self.get_response({'q': title})
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
         self.assertEqual(len(response_data['course_runs']), RESULT_COUNT)
@@ -417,7 +475,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
             title=title, authoring_organizations=authoring_organizations,
             status=ProgramStatus.Active, partner=self.partner
         )
-        response = self.get_typeahead_response(title)
+        response = self.get_response({'q': title})
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
         self.assertDictEqual(response_data, {'course_runs': [self.serialize_course_run(course_run)],
@@ -429,7 +487,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
         course_run = CourseRunFactory(title=title, course__partner=self.partner)
         program = ProgramFactory(title=title, status=ProgramStatus.Active, partner=self.partner)
         query = "Data Sci"
-        response = self.get_typeahead_response(query)
+        response = self.get_response({'q': query})
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
         expected_response_data = {
@@ -441,23 +499,25 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
     def test_unpublished_and_hidden_courses(self):
         """ Verify that typeahead does not return unpublished or hidden courses
         or programs that are not active. """
-        title = "Supply Chain"
+        title = "supply"
         course_run = CourseRunFactory(title=title, course__partner=self.partner)
-        CourseRunFactory(title=title + "_unpublished", status=CourseRunStatus.Unpublished, course__partner=self.partner)
-        CourseRunFactory(title=title + "_hidden", hidden=True, course__partner=self.partner)
+        CourseRunFactory(title=title + "unpublished", status=CourseRunStatus.Unpublished, course__partner=self.partner)
+        CourseRunFactory(title=title + "hidden", hidden=True, course__partner=self.partner)
         program = ProgramFactory(title=title, status=ProgramStatus.Active, partner=self.partner)
-        ProgramFactory(title=title + "_unpublished", status=ProgramStatus.Unpublished, partner=self.partner)
-        query = "Supply"
-        response = self.get_typeahead_response(query)
+        ProgramFactory(title=title + "unpublished", status=ProgramStatus.Unpublished, partner=self.partner)
+        query = "suppl"
+        response = self.get_response({'q': query})
         self.assertEqual(response.status_code, 200)
         response_data = response.json()
-        expected_response_data = {'course_runs': [self.serialize_course_run(course_run)],
-                                  'programs': [self.serialize_program(program)]}
+        expected_response_data = {
+            'course_runs': [self.serialize_course_run(course_run)],
+            'programs': [self.serialize_program(program)]
+        }
         self.assertDictEqual(response_data, expected_response_data)
 
     def test_exception(self):
         """ Verify the view raises an error if the 'q' query string parameter is not provided. """
-        response = self.get_typeahead_response()
+        response = self.get_response()
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.data, ["The 'q' querystring parameter is required for searching."])
 
@@ -468,7 +528,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
         program = ProgramFactory(authoring_organizations=authoring_organizations, partner=self.partner)
         partial_key = authoring_organizations[0].key[0:5]
 
-        response = self.get_typeahead_response(partial_key)
+        response = self.get_response({'q': partial_key})
         self.assertEqual(response.status_code, 200)
         expected = {
             'course_runs': [self.serialize_course_run(course_run)],
@@ -500,7 +560,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
             title='MIT Testing2',
             partner=self.partner
         )
-        response = self.get_typeahead_response('mit')
+        response = self.get_response({'q': 'mit'})
         self.assertEqual(response.status_code, 200)
         expected = {
             'course_runs': [self.serialize_course_run(mit_run),
@@ -523,7 +583,7 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
                 title=title, partner=partner,
                 status=ProgramStatus.Active
             ))
-        response = self.get_typeahead_response('partner', 'edx')
+        response = self.get_response({'q': 'partner'}, 'edx')
         self.assertEqual(response.status_code, 200)
         edx_course_run = course_runs[0]
         edx_program = programs[0]
--- a/course_discovery/apps/core/utils.py
+++ b/course_discovery/apps/core/utils.py
@@ -3,6 +3,8 @@ import logging
 
 from django.conf import settings
 
+from course_discovery.settings.process_synonyms import get_synonyms
+
 logger = logging.getLogger(__name__)
 
 
@@ -19,6 +21,7 @@ class ElasticsearchUtils(object):
             timestamp = datetime.datetime.utcnow().strftime("%Y%m%d%H%M%S")
             index = '{alias}_{timestamp}'.format(alias=alias, timestamp=timestamp)
             index_settings = settings.ELASTICSEARCH_INDEX_SETTINGS
+            index_settings['settings']['analysis']['filter']['synonym']['synonyms'] = get_synonyms(es)
             es.indices.create(index=index, body=index_settings)
             logger.info('...index [%s] created.', index)
 
--- a/course_discovery/apps/edx_haystack_extensions/backends.py
+++ b/course_discovery/apps/edx_haystack_extensions/backends.py
@@ -93,14 +93,21 @@ class ConfigurableElasticBackend(ElasticsearchSearchBackend):
 
     def build_schema(self, fields):
         content_field_name, mapping = super().build_schema(fields)
+        # Fields default to snowball analyzer, this keeps snowball functionality, but adds synonym functionality
+        snowball_with_synonyms = 'snowball_with_synonyms'
+        for field, value in mapping.items():
+            if value.get('analyzer') == 'snowball':
+                self.specify_analyzers(mapping=mapping, field=field,
+                                       index_analyzer=snowball_with_synonyms,
+                                       search_analyzer=snowball_with_synonyms)
         # Use the ngram analyzer as the index_analyzer and the lowercase analyzer as the search_analyzer
         # This is necessary to support partial searches/typeahead
         # If we used ngram analyzer for both, then 'running' would get split into ngrams like "ing"
         # and all words containing ing would come back in typeahead.
         self.specify_analyzers(mapping=mapping, field='title_autocomplete',
-                               index_analyzer='ngram_analyzer', search_analyzer='lowercase')
+                               index_analyzer='ngram_analyzer', search_analyzer=snowball_with_synonyms)
         self.specify_analyzers(mapping=mapping, field='authoring_organizations_autocomplete',
-                               index_analyzer='ngram_analyzer', search_analyzer='lowercase')
+                               index_analyzer='ngram_analyzer', search_analyzer=snowball_with_synonyms)
         return (content_field_name, mapping)
 
 
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -5,6 +5,9 @@ from django.conf import settings
 from haystack import connections as haystack_connections
 from haystack.management.commands.update_index import Command as HaystackCommand
 
+from course_discovery.settings.process_synonyms import get_synonyms
+
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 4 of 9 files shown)


course_discovery/settings/synonyms.txt
```diff
@@ -0,0 +1,55 @@
+# Comments must start with this character #
```
rlucioni: I would encourage you to use YAML for this. You can then use PyYAML to load it in without having to do any custom parsing.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,4 @@
+import os
+directory_path = os.path.dirname(os.path.realpath(__file__))
+with open(os.path.join(directory_path, 'synonyms.txt')) as f:
```
rlucioni: Have you tried just `open('synonyms.yml')` without this path manipulation?

course_discovery/settings/base.py
```diff
@@ -373,11 +374,22 @@
                 }
             },
             'analyzer': {
+                "snowball_with_synonyms": {
```
rlucioni: Single quotes here and everywhere, please!

course_discovery/settings/base.py
```diff
@@ -407,6 +419,11 @@
                     'type': 'nGram',
                     'min_gram': 2,
                     'max_gram': 15
+                },
+                'synonym' : {
+                  'type' : 'synonym',
+                  'ignore_case': 'true',
+                  'synonyms' : SYNONYMS_LIST
```
rlucioni: The linter isn't catching these because we have it configured to ignore settings files. Maybe we should change that. Run pylint against this file to catch cases like these (you have an extra space between the key and the colon).

course_discovery/settings/base.py
```diff
@@ -4,6 +4,7 @@
 from os.path import join, abspath, dirname
 from sys import path
 
+from .synonyms import SYNONYMS_LIST
```
rlucioni: We use absolute paths: `from course_discovery.settings.synonyms import SYNONYMS`.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,4 @@
...
+    SYNONYMS_LIST = [line.rstrip() for line in f if line[0] != '#']
```
rlucioni: You can just use `SYNONYMS` as the variable name.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +51,43 @@ def setUp(self):
...
+    def test_special_character_synonyms(self):
+        authoring_organizations = [OrganizationFactory(name='Valencia', key='ValenciaX')]
+        title='spanish'
+        ProgramFactory(title=title, partner=self.partner, authoring_organizations=authoring_organizations)
+        response1 = self.get_response({'q': 'español'}).json()
+        response2 = self.get_response({'q': 'Universitat Politècnica de València, València, Valencia'}).json()
+        if 'objects' in response1:
+            response1, response2 = response1['objects'], response2['objects']
+        self.assertTrue(len(response1))
+        self.assertTrue(len(response1))
```
rlucioni: Can you assert something more specific than "the response is not empty"?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +51,43 @@ def setUp(self):
         self.client.login(username=self.user.username, password=USER_PASSWORD)
 
 
+class SynonymsTestsMixin:
```
rlucioni: `SynonymTestMixin`

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +51,43 @@ def setUp(self):
...
+    def test_org_synonyms(self):
+        title = 'MITx'
+        authoring_organizations = [OrganizationFactory(name='Massachusetts Institute of Technology', key='MIT')]
```
rlucioni: If the synonyms file is changed, these tests may break. Is that desired?

rlucioni (changes requested):

course_discovery/settings/synonyms.txt
```diff
@@ -0,0 +1,55 @@
...
+ux, user experience
+# Other Terms
+HTML5, HTML
+autocad, auto cad
+and, &
+español, espanol, spanish
+hr, human resources
+iot, internet of things
+js, javascript
+excel, microsoft excel
```
rock345: maybe add msft excel

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -64,7 +101,7 @@ class CourseRunSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, Login
     faceted_path = reverse('api:v1:search-course_runs-facets')
     list_path = reverse('api:v1:search-course_runs-list')
 
-    def get_search_response(self, query=None, faceted=False):
+    def get_response(self, query=None, faceted=False):
```
clintonb, line 110: Why the change?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -326,14 +364,13 @@ def test_results_ordered_by_start_date(self, ordering):
...
-                               APITestCase):
+                               APITestCase, SynonymsTestsMixin):
     path = reverse('api:v1:search-typeahead')
 
-    def get_typeahead_response(self, query=None, partner=None):
-        qs = ''
-        query_dict = {'q': query, 'partner': partner or self.partner.short_code}
-        if query:
-            qs = urllib.parse.urlencode(query_dict)
+    def get_response(self, query=None, partner=None):
```
clintonb, line 428: Why the name change?

course_discovery/settings/synonyms.txt
```diff
@@ -0,0 +1,55 @@
...
+computer science, cs
+econ, economics
+ee, electrical engineering
+français, french
+it, information technology
+mis, management information systems
+psych, psychology
+seo, search engine optimization
+ux, user experience
+# Other Terms
```
clintonb: This would be cleaner with a newline between sections, or YAML as suggested in another comment.

course_discovery/settings/synonyms.txt
```diff
@@ -0,0 +1,55 @@
...
+js, javascript
+excel, microsoft excel
+microsoft, msft
+photo, photography
+management, mgmt
+css, cascading style sheets
+vb, visual basic
+vba, excel
+aws, amazon web services
+react, typescript, javascript, jquery, java script
```
clintonb: Should this be merged with the other line for "js, javascript"?

course_discovery/settings/synonyms.txt
```diff
@@ -0,0 +1,55 @@
...
+ee, electrical engineering
+français, french
+it, information technology
+mis, management information systems
+psych, psychology
+seo, search engine optimization
+ux, user experience
+# Other Terms
+HTML5, HTML
+autocad, auto cad
```
clintonb: Also, "cad"?

course_discovery/settings/synonyms.txt
```diff
@@ -0,0 +1,55 @@
+# Comments must start with this character #
+# Organizations
+MIT, MITx
+Berkeley, UC BerkeleyX, UCBerkeleyX
+The University of Texas System, UTx
+Microsoft, MicrosoftX
+ACCA, ACCA, ACCAx, ACCA-x
+NYIF, New York Institute of Finance, NYIFx
```
amangano-edx: It might make it easier for some people reading this file if these were sorted.

course_discovery/settings/synonyms.txt
MatthewPiatetsky: now using yaml/pyaml

course_discovery/settings/synonyms.py
MatthewPiatetsky: using synonyms now

course_discovery/settings/base.py
MatthewPiatetsky: fixed

course_discovery/settings/base.py
MatthewPiatetsky: fixed

course_discovery/apps/api/v1/tests/test_views/test_search.py

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: With the current implementation I am postprocessing the file to generate synonyms. Do you think it would be worth it to define a test synonyms file that would override the synonyms of base.py? It would only break if synonyms were removed (which seems like it would be uncommon)

course_discovery/settings/synonyms.txt
MatthewPiatetsky: added

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: I am using the SynonymMixin to test both search and typeahead. I wanted to be able to get the response and assert they are equal without knowing whether I am inside search or typeahead tests

course_discovery/settings/synonyms.txt
MatthewPiatetsky: added newlines

course_discovery/settings/synonyms.txt
MatthewPiatetsky: added

course_discovery/settings/synonyms.txt
MatthewPiatetsky: merged

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+    file_path = join(directory_path, 'synonyms.txt')
+    with open(file_path) as f:
+        return yaml.load(f)
+
+@lru_cache()
+def get_synonyms(host):
+    synonyms = get_synonym_lines()
+    analyze_string = '_analyze?analyzer=snowball&text='
+    synonyms = process_synonyms(host, analyze_string, synonyms)
+    return synonyms
```
clintonb: Add a newline.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+            response = requests.get(url).json()
+            synonym_tokens = ' '.join([item['token'] for item in response['tokens']])
+            processed_line.append(synonym_tokens)
+        processed_line = ','.join(processed_line)
+        processed_synonyms.append(processed_line)
+    return processed_synonyms
+
+def get_synonym_lines():
+    directory_path = dirname(realpath(__file__))
+    file_path = join(directory_path, 'synonyms.txt')
```
clintonb: `synonyms.yaml`

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+    and returns a list of lines (strings) in the format 'run, jog'
+    Preprocess synonyms to be in analyzed form (since they are defined in not-analyzed form.
+    """
+
+    processed_synonyms = []
+    for line in synonyms:
+        line = line.split(', ')
+        processed_line = []
+        for synonym in line:
+            url = '{}{}{}'.format(host, analyze_string, synonym)
```
clintonb: Use [`urllib.parse`](https://docs.python.org/3/library/urllib.parse.html) to build URLs.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
+from functools import lru_cache
+import requests
+from os.path import dirname, realpath, join
+import yaml
+
+def process_synonyms(host, analyze_string, synonyms):
+    """ This method takes list of lines (strings) in the form 'running, jogging'
```
clintonb: Follow the docstring format at http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +51,49 @@ def setUp(self):
...
+        CourseRunFactory(title='HTML', course__partner=self.partner)
+        ProgramFactory(title='HTML', partner=self.partner)
+        response1 = self.get_response({'q': 'HTML5'}).json()
+        response2 = self.get_response({'q': 'HTML'}).json()
+        if 'objects' in response1:
+            response1, response2 = response1['objects'], response2['objects']
+        self.assertDictEqual(response1, response2)
+
+    def test_special_character_synonyms(self):
+        ProgramFactory(title='spanish and french', partner=self.partner)
```
amangano-edx: Why does this only test `Program`s?

MatthewPiatetsky:
Understand the test failure, discussing with renzo (due to using wrong settings data)

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +51,49 @@ def setUp(self):
...
+        response1 = self.get_response({'q': 'español'}).json()
+        response2 = self.get_response({'q': 'français'}).json()
+        if 'objects' in response1:
+            response1, response2 = response1['objects'], response2['objects']
```
amangano-edx: Is it possible for these to be empty (maybe if neither search matched)? Maybe we should also verify that they contain something?

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+    Preprocess synonyms to be in analyzed form (since they are defined in not-analyzed form.
+    """
+
+    processed_synonyms = []
+    for line in synonyms:
+        line = line.split(', ')
+        processed_line = []
+        for synonym in line:
+            url = '{}{}{}'.format(host, analyze_string, synonym)
+            response = requests.get(url).json()
```
clintonb: This has the potential to overload Elasticsearch if there are a lot of synonyms. Is there a way to batch this operation or avoid it altogether?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +51,49 @@ def setUp(self):
...
+        self.assertDictEqual(response1, response2)
+
+    def test_stemmed_synonyms(self):
+        title = 'Biomedical Engineering'
+        ProgramFactory(title=title, partner=self.partner)
```
amangano-edx: Why only testing `Program`?

course_discovery/settings/synonyms.py
clintonb: We should use a setting (or some other editable mechanism) to define the location of this file. This should probably live in `edx_catalog_extensions`. The synonyms for edx.org may not be useful for Open edX users.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -225,10 +268,11 @@ def _test_exclude_program_types(self, program_status):
 
 
 @ddt.ddt
-class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase):
+class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase,
+                             SynonymTestMixin):
     path = reverse('api:v1:search-all-facets')
 
-    def get_search_response(self, querystring=None):
+    def get_response(self, querystring=None):
```
amangano-edx: This says it takes a `String` but all the Synonym tests pass in an dictionary. Could we maybe rename this param to communicate more clearly what is expected here?

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
+from functools import lru_cache
+import requests
```
clintonb: Group these appropriately.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+        line = line.split(', ')
+        processed_line = []
+        for synonym in line:
+            url = '{}{}{}'.format(host, analyze_string, synonym)
+            response = requests.get(url).json()
+            synonym_tokens = ' '.join([item['token'] for item in response['tokens']])
+            processed_line.append(synonym_tokens)
+        processed_line = ','.join(processed_line)
+        processed_synonyms.append(processed_line)
+    return processed_synonyms
```
amangano-edx: We should be able to store the synonyms in a format that does not require us to manually parse each line. Maybe CSV would be better?

course_discovery/settings/base.py
```diff
@@ -407,23 +418,17 @@
                     'type': 'nGram',
                     'min_gram': 2,
                     'max_gram': 15
+                },
+                'synonym' : {
+                  'type': 'synonym',
+                  'ignore_case': 'true',
+                  'synonyms': get_synonyms(HAYSTACK_HOST)
```
clintonb: This will not work as expected in production, or any other environment where `HAYSTACK_HOST` is overridden. This will need to be moved to a function that is called here in `base.py` and again in `production.py`, and `local.py` _after_ `HAYSTACK_HOST` is defined.

clintonb (changes requested):

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_synonym_lines():
+    directory_path = dirname(realpath(__file__))
+    file_path = join(directory_path, 'synonyms.txt')
+    with open(file_path) as f:
+        return yaml.load(f)
```
amangano-edx: Why bother using YAML if each line still has to be individually parsed? I think something like CSV would be better.

course_discovery/settings/base.py
MatthewPiatetsky: don't think that would work because production imports base, etc...
however, I realized I could just do this inside the update_index command

course_discovery/settings/synonyms.py
rlucioni: It is possible to use YAML for this without parsing each line:
```yaml
- - a
  - b
  - c
- - 1
  - 2
  - français
```
Yields:
```py
[['a', 'b', 'c'], [1, 2, 'français']]
```

course_discovery/settings/synonyms.py
MatthewPiatetsky: I think this is a valid docstring by that format, is this comment suggesting that I describe the arguments in an attributes section or something else?

course_discovery/settings/synonyms.py
MatthewPiatetsky: fixed

course_discovery/settings/synonyms.py
MatthewPiatetsky: fixed

course_discovery/settings/base.py
MatthewPiatetsky: I am now doing this in update_index and core/tests/mixins

course_discovery/settings/synonyms.py
MatthewPiatetsky: imo this version is more readable, since the line separation makes it easy to see which synonyms are grouped together. (the only postprocessing after reading the file though is splitting on comma. The bulk is parsing the response back from elasticsearch)

course_discovery/settings/synonyms.py
MatthewPiatetsky: I don't think there is a bulk operation in v 1.5.2
I think we want to process these with ES to ensure we see the same behavior as when we search (as opposed to some python library).
With this implementation this will only get called when rebuilding the index, do you still think that would be a problematic amount of load?

course_discovery/settings/synonyms.py
MatthewPiatetsky: is the change I made what you were suggesting

course_discovery/settings/synonyms.py
MatthewPiatetsky: Could you clarify how this would work? Currently I'm calling out to the function from apps/core/utils.py and edx_haystack_extensions/management/commands/update_index.py.

course_discovery/settings/synonyms.txt
MatthewPiatetsky: i sorted the lines alphabetically and change the order of some of the synonyms, sometimes i think the lines are more readable if they aren't alphabetical, but i changed this one

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: Do you think there would be value in testing both courses and programs?  
I think if I know that some synonyms work for both, it doesn't add value to run identical tests for both if we don't think there is a reason that one wouldn't work. I'm just looking at the same title field on both, not doing anything different

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +61,50 @@ def setUp(self):
...
+        response2 = self.get_response({'q': 'MIT'}).json()
+        if 'objects' in response1:
+            response1, response2 = response1['objects'], response2['objects']
+        self.assertDictEqual(response1, response2)
+
+    def test_title_synonyms(self, get_synonym_lines):  # pylint: disable=unused-argument
+        CourseRunFactory(title='HTML', course__partner=self.partner)
+        ProgramFactory(title='HTML', partner=self.partner)
+        response1 = self.get_response({'q': 'HTML5'}).json()
+        response2 = self.get_response({'q': 'HTML'}).json()
```
amangano-edx: The `get_response` methods all take different args and some have different names. I think we should go a little further in this refactor and change them all to have a consistent interface. Rather than `partner` as an argument for some of them and `faceted` as an argument to others, why not change them all so that they all just expect a query dictionary?

course_discovery/apps/core/utils.py
```diff
@@ -3,12 +3,15 @@
...
+from course_discovery.settings.synonyms import get_synonyms
+
+
 logger = logging.getLogger(__name__)
 
 
 class ElasticsearchUtils(object):
     @classmethod
-    def create_alias_and_index(cls, es, alias):
+    def create_alias_and_index(cls, es, alias, host=None):
```
clintonb: There is no need to pass `host` here. `es` is an Elasticsearch client. Use it. https://elasticsearch-py.readthedocs.io/en/master/

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -54,7 +58,7 @@ def set_alias(self, backend, alias, index):
         }
         backend.conn.indices.update_aliases(body)
 
-    def prepare_backend_index(self, backend):
+    def prepare_backend_index(self, backend, host):
```
clintonb: Again, no need for `host`. Use `backend`

course_discovery/settings/base.py
```diff
@@ -352,6 +351,18 @@
     'permission_denied_handler': 'course_discovery.apps.api.views.api_docs_permission_denied_handler'
 }
 
+# Haystack configuration (http://django-haystack.readthedocs.io/en/v2.5.0/settings.html)
```
clintonb: Nit: this doesn't need to be moved.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,38 @@
+from functools import lru_cache
+from os.path import dirname, realpath, join
+from urllib.parse import urljoin
+
+import requests
+import yaml
+
+def process_synonyms(host, analyze_string, synonyms):
+    """This method takes list of lines (strings) in the form 'running, jogging'
```
clintonb: List the `Args` according to the link provided earlier.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,38 @@
+from functools import lru_cache
+from os.path import dirname, realpath, join
+from urllib.parse import urljoin
+
+import requests
+import yaml
+
+def process_synonyms(host, analyze_string, synonyms):
```
clintonb: Pass the `Elasticsearch` client and use it, rather than `requests`.

course_discovery/settings/synonyms.yaml
```diff
@@ -0,0 +1,59 @@
...
+- HTML5, HTML
+- iot, internet of things
+- javascript, js, java script, react, typescript, jquery
+- microsoft, msft
+- management, mgmt
+- os, operating system
+- photo, photography
+- vb, visual basic
+- vba, excel
+- usa, united states of america
```
clintonb: Add a newline.

clintonb (changes requested):

course_discovery/settings/synonyms.yaml
MatthewPiatetsky: added

course_discovery/settings/synonyms.py
MatthewPiatetsky: This is parsing the response from es, not the file. Only line.split(', ') is parsing the file (that could be avoided, but I think the current format is much more readable)

course_discovery/settings/synonyms.py
MatthewPiatetsky: added attribuets section

course_discovery/apps/core/utils.py
MatthewPiatetsky: thanks, i couldn't find the analyze api on it earlier (it was under indices for some reason which doesn't make sense since you can call it without an index)

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
MatthewPiatetsky: fixed

course_discovery/settings/base.py
MatthewPiatetsky: fixed

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: I'll update some of the same things to be consistent (like query vs querystring). Though they may have the same name, they are methods with different functionality because they are in different classes. It makes it more dry to avoid the small bit of logic regarding faceted and partner to have that in the method rather than every test call

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: renamed to be consistent with the other search method

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: same reason mentioned above

course_discovery/settings/synonyms.py
amangano-edx: I was referring to the `line.split` piece. I think that functionality would be better handled by a CSV library, since that is how you're storing the data. I don't really understand the advantage to using YAML over CSV for this stuff.

MatthewPiatetsky: I think clinton and renzo suggested that so I could include comments and new lines for better readability. So regardless there will be one bit of text parsing for either format.

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: good point, i didn't want to explicitly look at the objects because this method uses both search and typeahead which return in different format. however, I realized i am only testing special characters here so if these did not work, this test could still pass. for that reason i've added a base case without special characters that we expect to work to compare against

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +61,52 @@ def setUp(self):
         self.client.login(username=self.user.username, password=USER_PASSWORD)
 
 
+@mock.patch('course_discovery.settings.synonyms.get_synonym_lines', return_value=SYNONYMS)
+class SynonymTestMixin:
+    def test_org_synonyms(self, get_synonym_lines):  # pylint: disable=unused-argument
```
clintonb: What's going on here with `get_synonym_lines`? It might be cleaner to move the mocking to a helper method.

clintonb: Add a docstring to these test methods explaining what they are testing.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +61,52 @@ def setUp(self):
...
+        authoring_organizations = [OrganizationFactory(name='Massachusetts Institute of Technology', key='MIT')]
+        CourseRunFactory(
+            title=title,
+            course__partner=self.partner,
+            authoring_organizations=authoring_organizations
+        )
+        ProgramFactory(title=title, partner=self.partner, authoring_organizations=authoring_organizations)
+        response1 = self.get_response({'q': title}).json()
+        response2 = self.get_response({'q': 'MIT'}).json()
+        if 'objects' in response1:
```
clintonb: `if` statements in test methods are a sign that the test method should really be two methods. I'm guessing this line is here because you are attempting to use the same method for two different endpoints. If that is the case, the mixin should contain the code that can be fully shared, and call an assertion helper that must be overridden by the individual test cases.

This should be done here and in the other test methods you have `if` statements.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -51,6 +61,52 @@ def setUp(self):
...
+            response1, response2 = response1['objects'], response2['objects']
```
amangano-edx: maybe the logic to normalize the responses could be DRY'd up by moving it into a helper method.

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: refactored the objects check into a helper and put an assertion that the response was not empty there

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: done

MatthewPiatetsky:
@amangano-edx @clintonb 
think I've addressed your existing comments (except @clintonb not sure about how to move the synonyms code into edx_catalog_extensions, could use some clarification on that)

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: maybe I can rename it get_synonym_lines_from_file. That method loads the file and this overrides it so the tests don't have to depend on the file

MatthewPiatetsky: renamed and added docstrings

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: added the process response to each individual class

course_discovery/settings/synonyms.yaml
clintonb: Additional synonyms: murika, murrrrrika, god bless, trumplonia

FYI @jcdyer @stroilova

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -268,15 +313,24 @@ def _test_exclude_program_types(self, program_status):
 
 
 @ddt.ddt
-class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase):
+class AggregateSearchViewSet(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin, APITestCase,
+                             SynonymTestMixin):
```
amangano-edx: Does the order of the mixins matter here?

clintonb: Yes. Mixins should come before the base class.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -77,11 +117,16 @@ def get_search_response(self, query=None, faceted=False):
         url = '{path}?{qs}'.format(path=path, qs=qs)
         return self.client.get(url)
 
+    def process_response(self, response):
+        response = self.get_response(response).json()
+        self.assertTrue(response['objects']['count'])
+        return response['objects']
```
amangano-edx, line 123: Does this contain a dictionary or a list? The Synonym tests all seem expect to get dictionaries (the use `self.assertDictEqual`).

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -93,14 +93,19 @@ def specify_analyzers(self, mapping, field, index_analyzer, search_analyzer):
 
     def build_schema(self, fields):
         content_field_name, mapping = super().build_schema(fields)
+        # Fields default to snowball analyzer, this keeps snowball functionality, but adds synonym functionality
+        synonym_snowball_fields = [field for field, value in mapping.items() if value.get('analyzer') == 'snowball']
+        for field in synonym_snowball_fields:
```
amangano-edx: You could do this in one loop (and avoid allocating another list object) by writing:
```python
for field, value in mapping.items():
    if value.get('analyzer') != 'snowball'
        continue
```

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,40 @@
...
+    applies the snowball analyzer and returns a list of synonyms in the format 'run, jog'.
+
+    Attributes:
+        es (client): client for making requests to es
+        synonyms (list): list of synonyms (each synonym group is a comma separated string)
+    """
+
+    processed_synonyms = []
+    for line in synonyms:
+        line = line.split(', ')
```
amangano-edx: Does this handle escaped commas correctly, like the one on https://github.com/edx/course-discovery/pull/599/files#diff-8fa1fa692c631251ec1a558a998d282aR13?

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: changed to have api test case last

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: dictionary

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: changed

course_discovery/settings/synonyms.py
amangano-edx: I think it would be good to have a test case for this.

course_discovery/apps/core/utils.py
```diff
@@ -3,6 +3,9 @@
 
 from django.conf import settings
 
+from course_discovery.settings.synonyms import get_synonyms
+
+
```
davec-edx: Extra blank line  (I think)

course_discovery/apps/core/utils.py
```diff
@@ -19,6 +22,7 @@ def create_alias_and_index(cls, es, alias):
             timestamp = datetime.datetime.utcnow().strftime("%Y%m%d%H%M%S")
             index = '{alias}_{timestamp}'.format(alias=alias, timestamp=timestamp)
             index_settings = settings.ELASTICSEARCH_INDEX_SETTINGS
+            index_settings['settings']['analysis']['filter']['synonym']['synonyms'] = get_synonyms(es)
```
davec-edx, line 24: Is this guaranteed to exist or there be a try/except around this?

course_discovery/settings/synonyms.py
MatthewPiatetsky: good point, i've updated the command to not split on those commas

course_discovery/apps/core/utils.py
MatthewPiatetsky: removed

course_discovery/apps/core/utils.py
MatthewPiatetsky: I create this in the base.py settings file, so I don't think I need to check for this

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -93,14 +93,20 @@ def specify_analyzers(self, mapping, field, index_analyzer, search_analyzer):
...
         # Use the ngram analyzer as the index_analyzer and the lowercase analyzer as the search_analyzer
         # This is necessary to support partial searches/typeahead
         # If we used ngram analyzer for both, then 'running' would get split into ngrams like "ing"
         # and all words containing ing would come back in typeahead.
         self.specify_analyzers(mapping=mapping, field='title_autocomplete',
-                               index_analyzer='ngram_analyzer', search_analyzer='lowercase')
+                               index_analyzer='ngram_analyzer', search_analyzer='snowball_with_synonyms')
         self.specify_analyzers(mapping=mapping, field='authoring_organizations_autocomplete',
-                               index_analyzer='ngram_analyzer', search_analyzer='lowercase')
+                               index_analyzer='ngram_analyzer', search_analyzer='snowball_with_synonyms')
```
amangano-edx: Nitpick: 'snowball_with_synonyms' is repeated a lot. You could save it to a variable first and DRY this up a bit.

course_discovery/settings/synonyms.py
```diff
@@ -0,0 +1,41 @@
...
+    applies the snowball analyzer and returns a list of synonyms in the format 'run, jog'.
+
+    Attributes:
+        es (client): client for making requests to es
+        synonyms (list): list of synonyms (each synonym group is a comma separated string)
+    """
+
+    processed_synonyms = []
+    for line in synonyms:
+        line = split(r'(?<!\\), ', line) # Split string on commas , except escaped commas \,
```
amangano-edx: I think there should be a test for this somewhere, to verify it works as expected.

course_discovery/apps/edx_haystack_extensions/backends.py
davec-edx: ehhh, leave it DAMP

MatthewPiatetsky: added variable

clintonb (approved):
You're missing "murika"!

rlucioni (approved):

MatthewPiatetsky:
I had a conversation regarding the synonym format with Anthony, he had some concerns about escaping characters like commas and just the general structure, I've changed the files to be lists of lists in python, that way we have to worry about less file parsing/escaping/comments etc.... Thoughts?

course_discovery/settings/synonyms_list.py
```diff
@@ -0,0 +1,64 @@
+# Note: Do not use synonyms with punctuation, search and typeahead do not yet fully support punctuation
+
+SYNONYMS = [
```
rlucioni: I'd prefer not to use Python modules to store configuration and data, but it seems to work in this case. Generally, using Python for config/data opens things up to unnecessary complexity. I still think my original YAML suggestion, one word per line, would be the cleanest option.

If you're going to go this route, please follow the same syntax standards we have elsewhere: single quotes, 4 space indentation. You may want to look into pointing the linter at these settings files so I don't have to keep making these comments 😄 .

course_discovery/settings/synonyms_list.py
```diff
@@ -0,0 +1,64 @@
+# Note: Do not use synonyms with punctuation, search and typeahead do not yet fully support punctuation
```
rlucioni: Keep it concise and just name the module `synonyms.py`. Same with `test_synonyms.py`.

course_discovery/settings/synonyms_list.py
```diff
@@ -0,0 +1,64 @@
...
+["hr", "human resources"],
+["HTML5", "HTML"],
+["iot", "internet of things"],
+["javascript", "js", "java script", "react", "typescript", "jquery"],
+["management", "mgmt"],
+["os", "operating system"],
+["photo", "photography"],
+["vb", "visual basic"],
+["vba", "excel"],
+["usa", "united states of america", "murika"]
```
rlucioni: Add a trailing comma so the next person to add an entry gets a clean diff.

course_discovery/settings/synonyms_list.py
MatthewPiatetsky: renamed (renamed the existing one process_synonyms.py)

course_discovery/settings/synonyms_list.py
MatthewPiatetsky: fixed

course_discovery/settings/synonyms_list.py
MatthewPiatetsky: It seems like everyone wanted slightly different, so it is hard to make everyone happy, but hopefully everyone is okish with this route.  
I updated the syntax, I think it is a good idea to update the linting to look at the settings, but I don't want to do that in this PR to confuse the diff. I can do that as soon as I merge this

rlucioni (approved):

course_discovery/settings/process_synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+        processed_line = []
+        for synonym in line:
+            response = es.indices.analyze(text=synonym, analyzer='snowball')
+            synonym_tokens = ' '.join([item['token'] for item in response['tokens']])
+            processed_line.append(synonym_tokens)
+        processed_line = ','.join(processed_line)
+        processed_synonyms.append(processed_line)
+    return processed_synonyms
+
+def get_synonym_lines_from_file():
```
amangano-edx, line 28: This probably doesn't need to be its own method anymore.

course_discovery/settings/process_synonyms.py
```diff
@@ -0,0 +1,36 @@
...
+    Attributes:
+        es (client): client for making requests to es
+        synonyms (list): list of synonyms (each synonym group is a comma separated string)
+    """
+
+    processed_synonyms = []
+    for line in synonyms:
+        processed_line = []
+        for synonym in line:
+            response = es.indices.analyze(text=synonym, analyzer='snowball')
```
amangano-edx, line 21: What should we do if this call fails? Do we care?

amangano-edx:
LGTM

PR #605, opened 2017-01-29
Stop metadata refresh from loading programs

rlucioni:
The catalog is now the source of truth for all program data. We don't want changes made to programs in the catalog being overridden by stale data from the deprecated programs service. Although use of it is removed, the program API data loader itself is left in place for now to give open source site operators a tool with which to migrate their own programs.

ECOM-4422

@edx/ecommerce


clintonb (changes requested):
Just remove the URL from `SiteConfiguration` for edx.org. This code should remain in place for those still using the program service (e.g. Open edX).

rlucioni:
@clintonb good call. Closing this.

PR #609, opened 2017-01-30
Add pytest-django-ordering

rlucioni:
This awesome plugin handles the same test sorting as our local conftest module. We can apply this plugin to other projects requiring the same treatment.

@edx/ecommerce FYI


cpennington:
I'd politely disagree, and suggest that we should be applying a randomized ordering instead.

rlucioni:
@cpennington I agree, and that is my ultimate goal. We have to start somewhere, though. Our first step is using pytest; [randomized ordering](https://github.com/adamchainz/pytest-randomly) will follow. The fact that we have test cases which reset the database by truncating tables in the first place means that we have ordering-dependent tests, even though we didn't realize it before.

Since you're here, do you have any thoughts on how to allow randomized execution of this project's [`LiveServerTestCase`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.LiveServerTestCase)? See this [docstring](https://github.com/django/django/blob/master/django/test/testcases.py#L1268-L1271).

cpennington:
I'm not sure that truncating the database actually makes the tests order dependent, though... The randomization shouldn't separate cases inside the same test class. Likewise, I think the `LiveServerTestCase` should work fine, as long as the tests themselves are doing all of the setup and teardown that they need.

rlucioni:
@cpennington right; the randomization will shuffle the order of the test classes themselves, which is where the problem presents itself. If either that `LiveServerTestCase` or another `TransactionTestCase` runs before all `TestCase` subclasses have completed, there is a high probability that the `TestCase` subclasses running afterwards will fail. Django masks this problem by running all `TestCase` subclasses before other `TransactionTestCase` subclasses.

The root of the problem is that many of this project's tests rely on initial data populated by migrations. That can most likely be corrected by using pytest's [fixtures](http://doc.pytest.org/en/latest/fixture.html).

cpennington:
Ah, I see.

PR #613, opened 2017-01-31
ECOM-7021 Convert organizations tags to typeahead and add tags to Course Runs

MatthewPiatetsky:
@edx/ecommerce 

tags4


rlucioni (approved):
@MatthewPiatetsky very nice.

@mikedikan this is a good example of an area where some pre-planning/technical discovery would have done us good. I think we spent an hour talking about this yesterday?

requirements/base.txt
```diff
@@ -18,6 +18,7 @@ django-sortedm2m==1.3.2
 django-stdimage==2.3.3
 django-storages==1.5.0
 django-taggit==0.20.2
```
clintonb: Let's take this opportunity to update this library: https://pypi.python.org/pypi/django-taggit.

clintonb (approved):

MatthewPiatetsky:
@rlucioni to be fair, i think we spent most of the time talking about the approach and figuring out that tags is what we wanted, and only a few minutes about whether tags was actually possible. Regardless, in the future I think if we decide we want a certain approach but aren't sure how long it could take we could just write "try to do x and come back to the team to discuss if it seems like it will be very complex"

requirements/base.txt
MatthewPiatetsky: updated

rlucioni:
@MatthewPiatetsky fair, that's a good point.

mikedikan:
@rlucioni I have received your feedback and appreciate your giving of it.

rlucioni:
@mikedikan 👍

PR #617, opened 2017-02-01
(WIP) Seats included in minimal course-run serializer.

hasnain-naveed:


clintonb (changes requested):
Before continuing, please explain why you need this. We have intentionally worked to minimize the the data returned by this serializer and API endpoint. This change will have performance consequences.

douglashall:
@clintonb We need the course price to display on the program detail page. Do you have a preferred way of making that happen?

clintonb:
@douglashall what is the performance impact of this change? It will need to be load-tested. The approach here seems correct. Another option is to add a flag (e.g. `include_course_run_seats` or `include_full_course_details`) that will trigger the use of an alternate serializer.

douglashall:
Thanks @clintonb. I like the separate serializer approach.

@hasnain-naveed Let's add a `CourseRunWithSeatsSerializer` which will be used when specifying an optional `include_course_run_seats` parameter on the API request.

clintonb:
@douglasshall what other changes do you intend to make to the API? I'd like to make sure we are thinking about these changes sooner than at code review time.

clintonb:
@douglashall ^

douglashall:
@clintonb We should only need the changes outlined above for the WL programs work we have in progress.

PR #619, opened 2017-02-01
ECOM-7021 Add tags to course run serializer

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -84,6 +84,7 @@ PREFETCH_FIELDS = {
         'staff',
         'staff__position',
         'staff__position__organization',
+        'tags',
         'transcript_languages',
     ],
     'course': [
@@ -91,6 +92,7 @@ PREFETCH_FIELDS = {
         'authoring_organizations__partner',
         'authoring_organizations__tags',
         'course_runs',
+        'course_runs__tags',
         'expected_learning_items',
         'level_type',
         'prerequisites',
@@ -363,11 +365,13 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
     instructors = serializers.SerializerMethodField(help_text='This field is deprecated. Use staff.')
     staff = PersonSerializer(many=True)
     level_type = serializers.SlugRelatedField(read_only=True, slug_field='name')
+    tags = TagListSerializerField()
 
     @classmethod
     def prefetch_queryset(cls):
         queryset = super().prefetch_queryset()
         return queryset.select_related('language', 'video').prefetch_related(
+            'tags',
             'transcript_languages',
             Prefetch('staff', queryset=PersonSerializer.prefetch_queryset()),
         )
@@ -376,7 +380,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
         fields = MinimalCourseRunSerializer.Meta.fields + (
             'course', 'full_description', 'announcement', 'video', 'seats', 'content_language',
             'transcript_languages', 'instructors', 'staff', 'min_effort', 'max_effort', 'modified',
-            'level_type', 'availability', 'mobile_available', 'hidden',
+            'level_type', 'availability', 'mobile_available', 'hidden', 'tags',
         )
 
     def get_instructors(self, obj):  # pylint: disable=unused-argument
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -289,9 +289,12 @@ class MinimalCourseRunSerializerTests(TestCase):
             'type': course_run.type,
         }
 
+    def create_course_run(self):
+        return CourseRunFactory()
+
     def test_data(self):
         request = make_request()
-        course_run = CourseRunFactory()
+        course_run = self.create_course_run()
         serializer = self.serializer_class(course_run, context={'request': request})
         expected = self.get_expected_data(course_run, request)
         self.assertDictEqual(serializer.data, expected)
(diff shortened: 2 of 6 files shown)


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -294,11 +295,13 @@ def test_data(self):
         course_run = CourseRunFactory()
         serializer = self.serializer_class(course_run, context={'request': request})
         expected = self.get_expected_data(course_run, request)
+        course_run.tags.add(self.TAG)
```
clintonb: This is not necessary.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -266,6 +266,7 @@ def test_published_course_runs_only(self, published_course_runs_only):
 
 class MinimalCourseRunSerializerTests(TestCase):
     serializer_class = MinimalCourseRunSerializer
+    TAG = 'test-tag'
```
clintonb: This is not necessary.

course_discovery/apps/api/serializers.py
```diff
@@ -334,6 +336,7 @@ class MinimalCourseRunSerializer(TimestampModelSerializer):
     def prefetch_queryset(cls):
         return CourseRun.objects.all().select_related('course').prefetch_related(
             'course__partner',
+            'tags',
```
clintonb: This shouldn't be here given that `MinimalCourseRunSerializer` does not serialize tags.

clintonb (changes requested):

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: Are you suggesting just removing the class attribute and hardcoding the string when I add the tag, or are you suggesting creating a tag factory and adding that as a post generation value to the course run factory?     
This is how the Organization tag tests do it, so I'd probably want to change those if I did this differently

clintonb: Neither `MinimalCourseRunSerializer` nor `MinimalCourseRunSerializerTests` deal with tags. Neither class should be modified.

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: removed

course_discovery/apps/api/serializers.py
MatthewPiatetsky: removed

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: but CourseRunSerializerTests also runs the test_data method defined in MinimalCourseRunSerializerTests and CourseRunSerializer does deal with tags

clintonb: Override `test_data` where necessary. There is no need to modify this class.

clintonb: Correction: there is no need to modify this class to make it aware of tags. In lieu of an override, you could also add a method to create the course run, that can also be overridden.

MatthewPiatetsky: got it, added create_course_run method

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -289,16 +289,21 @@ def get_expected_data(self, course_run, request):  # pylint: disable=unused-argu
             'type': course_run.type,
         }
 
+    def create_course_run(self):
+        course_run = CourseRunFactory()
```
clintonb: `return CourseRunFactory()`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -321,13 +326,19 @@ def get_expected_data(self, course_run, request):
             'modified': json_date_format(course_run.modified),  # pylint: disable=no-member
             'level_type': course_run.level_type.name,
             'availability': course_run.availability,
+            'tags': [self.TAG],
         })
 
         return expected
 
+    def create_course_run(self):
+        course_run = CourseRunFactory()
```
clintonb: `course_run = super(CourseRunSerializerTests, self).create_course_run()`

clintonb (approved):

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -39,7 +39,7 @@ def test_get(self):
         """ Verify the endpoint returns the details for a single course. """
         url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
 
-        with self.assertNumQueries(9):
```
amangano-edx, line 42: Why do some of these change by 1 and others by 2?

MatthewPiatetsky: will fix/explain shortly

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -140,8 +140,7 @@ def csv(self, request, id=None):  # pylint: disable=redefined-builtin,unused-arg
 
         # We use select_related and prefetch_related to decrease our database query count
         course_runs = course_runs.select_related(*serializers.SELECT_RELATED_FIELDS['course_run'])
-        prefetch_fields = ['course__' + field for field in serializers.PREFETCH_FIELDS['course']]
```
MatthewPiatetsky, line 143: Is this necessary? If so, should we be doing it here?
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views/course_runs.py#L54

PR #621, opened 2017-02-01
Add boosting config for enrollable CourseRuns

amangano-edx:
Adds config to boost any course that a user can enroll in.

JIRA: https://openedx.atlassian.net/browse/ECOM-7038
@edx/ecommerce please review.


MatthewPiatetsky:
Can you add some explanation regarding why we are doing this and what specifically you're changing in the config?

course_discovery/apps/edx_haystack_extensions/migrations/0005_auto_20170201_2043.py
```diff
@@ -0,0 +1,110 @@
...
+                    },
+
+                    # Boost function for enrollable CourseRuns.
+                    # We want to boost if:
+                    #   - enrollment_start and enrollment_end are unspecified
+                    #   - enrollment_start is unspecified and enrollment_end is in the future
+                    #   - enrollment_end is unspecified and enrollment_start is in the past
+                    #   - enrollment_start is in the past and enrollment_end is in the future
+                    # We apply a weight of 1.0 to match the boost given for self paced and enrollable paid courses.
+                    {
```
MatthewPiatetsky, line 59: is this the new section? is this essentially calculating whether something is enrollable in an ES query?
Why not just make it a property that we index like with your previous PR and just check that?

amangano-edx: Because it's dynamic. A course may not be enrollable today but may become enrollable tomorrow, if for instance, the enrollment_start date is tomorrow.

amangano-edx: Yes, this is using ES to check whether or not something is enrollable and applying boosting if it is.

amangano-edx:
@MatthewPiatetsky Product requested this in addition to the paid course boosting I just did with the goal of having courses that are enrollable appear above archived courses in search results. The last function entry in the config is completely new.

course_discovery/apps/edx_haystack_extensions/migrations/0005_auto_20170201_2043.py
```diff
@@ -0,0 +1,110 @@
...
+                                ],
+                                'should': [
+                                    {'bool': {'must_not': { 'exists': {'field': 'paid_seat_enrollment_end'}}}},
+                                    {'range': {'paid_seat_enrollment_end': {'gte': 'now'}}}
+                                ]
+                            }
+                        },
+                        'weight': 1.0
+                    },
+
```
amangano-edx, line 51: New config begins here.

rlucioni (commented):
@amangano-edx your config looks good. However, I'm hesitant to approve additional tests which rely on initial data in the database (the boosting config, in this case). I just sent an email about this same issue. 

That said, I know we're already doing things this way and don't want to hold this up just because of it. I'm fine with merging this if we can also look into ways of fixing this problem (e.g., switching to a config file instead of migrations).

amangano-edx:
@rlucioni I think moving to a config file is a great idea. I don't think that merging this would make the process of moving to a config file any more difficult, though, so I'd like to move forward with this for the time being with the expectation that it be refactored and moved into the config file when that work is done.

rlucioni (approved):
@amangano-edx I think the same goes for the tests being added here, so :shipit: .

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -667,3 +669,45 @@ def test_program_type_boosting(self, program_type):
...
+        self.build_normalized_course_run(title='test1')
+
+        # Create the test record
+        test_record = self.build_normalized_course_run(
+            title='test2',
+            enrollment_start=enrollment_start,
+            enrollment_end=enrollment_end
+        )
+
+        search_results = SearchQuerySet().models(CourseRun).all()
```
MatthewPiatetsky, line 707: based on what we saw today with the fact that this runs a slightly different query than search/typeahead which could affect boosting, i'm questioning whether we want to use the searchqueryset directly or query those views directly (to avoid having to question this as a factor in debugging issues). However, this is fine for the time being.

MatthewPiatetsky:
👍

PR #622, opened 2017-02-01
ECOM-7021 Add reporting type field

MatthewPiatetsky:
@edx/ecommerce 

in case it is confusing what the migrations are supposed to be doing, it is
1. add reporting type to models and add that to course run
2. add 'mooc' to reporting types
3. add more edX specific reporting types in edx_catalog_extensions migration
4. set reporting type field on course run to be required and default it to mooc

I could combine 2 and 3 in edx_catalog_extensions and put 4 in edx_catalog_extensions as well, but for some reason trying alter a field on a course_metadata courserun model from edx_catalog_extensions migration is giving me an error


course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
```diff
@@ -147,10 +147,10 @@ def test_courses(self):
         for course_run in excluded_runs:
             SeatFactory(course_run=course_run)
 
-        with self.assertNumQueries(42):
+        with self.assertNumQueries(30):
             response = self.client.get(url)
-            self.assertEqual(response.status_code, 200)
-            self.assertListEqual(response.data['results'], self.serialize_catalog_course(courses, many=True))
+        self.assertEqual(response.status_code, 200)
+        assert response.data['results'] == self.serialize_catalog_course(courses, many=True)
```
rlucioni: Hurray for pytest! You can change the line above to `assert response.status_code == 200`.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -66,21 +66,21 @@ def test_get_include_deleted_programs(self):
...
-        with self.assertNumQueries(20):
+        with self.assertNumQueries(13):
             response = self.client.get(url)
-            self.assertEqual(response.status_code, 200)
-            self.assertEqual(
-                response.data,
-                self.serialize_course_run(self.course_run, extra_context={'include_deleted_programs': True})
-            )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
```
rlucioni: You can use regular `assert` statements here, too, and throughout.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -136,8 +136,7 @@ def test_list_query(self):
         query = 'title:Some random title'
         url = '{root}?q={query}'.format(root=reverse('api:v1:course_run-list'), query=query)
 
-        with self.assertNumQueries(37):
```
rlucioni: Why get rid of this?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -120,7 +120,7 @@ def test_list_sorted_by_course_start_date(self):
         """ Verify the endpoint returns a list of all catalogs sorted by course start date. """
         url = '{root}?ordering=start'.format(root=reverse('api:v1:course_run-list'))
 
-        with self.assertNumQueries(11):
+        with self.assertNumQueries(12):
```
rlucioni: Are these incremented by one because it only takes one query to retrieve the new field, or because there's only one `ReportingType` record present? Given that the new field is a `ForeignKey`, I would have expected to see a `select_related` on the serializer, but I don't.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -65,8 +65,9 @@ def test_authentication(self):
     def test_retrieve(self):
         """ Verify the endpoint returns the details for a single program. """
         program = self.create_program()
-        with self.assertNumQueries(76):
-            self.assert_retrieve_success(program)
+        with self.assertNumQueries(32):
+            response = self.assert_retrieve_success(program)
+        self.assertEqual(response.data, self.serialize_program(program))
```
rlucioni: Consider switching the lines you're touching to use `assert`.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -75,16 +76,18 @@ def test_retrieve_with_sorting_flag(self, order_courses_by_start_date):
...
         self.assertEqual(course_list, list(program.courses.all()))  # pylint: disable=no-member
 
     def test_retrieve_without_course_runs(self):
         """ Verify the endpoint returns data for a program even if the program's courses have no course runs. """
         course = CourseFactory()
         program = ProgramFactory(courses=[course])
-        with self.assertNumQueries(49):
-            self.assert_retrieve_success(program)
+        with self.assertNumQueries(19):
```
rlucioni, line 88: For those who are wondering: these query counts are dropping because previously we were also counting the un-prefetched queries made by test methods like `serialize_program()`. We only care about the queries made by the view.

course_discovery/apps/course_metadata/models.py
```diff
@@ -381,6 +385,7 @@ class CourseRun(TimeStampedModel):
         default=False,
         help_text=_('Indicates whether the course relation has been manually overridden.')
     )
+    reporting_type = models.ForeignKey(ReportingType, null=False, blank=False, default=1)
```
rlucioni: Is the default here necessary? Hardcoding a primary key seems wrong. What happens if that record isn't present? During testing, this will fail MySQL's referential integrity checks if the record is cleared. I'd recommend leaving the default unspecified.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -199,6 +199,13 @@ def applicable_seat_types(self, create, extracted, **kwargs):
             add_m2m_data(self.applicable_seat_types, extracted)
 
 
+class ReportingTypeFactory(factory.django.DjangoModelFactory):
+    class Meta(object):
+        model = ReportingType
+
+    name = FuzzyText()
```
rlucioni: Use `Faker`, now included with factory_boy. These fuzzy utilities are deprecated.

rlucioni (commented):
I also mentioned this to @amangano-edx on https://github.com/edx/course-discovery/pull/621: we shouldn't be introducing new tests which only pass if some required initial data is present in the database. Tests should be creating (or verifying the existence of, via `get_or_create()`) their own data if they need it to pass.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
MatthewPiatetsky: This test performs an es query and locally there were only a few setup related queries and no other SQL queries. I think there are SQL queries coming back because I think the SerializationMixin may be mocking out the actual call to ES?

course_discovery/apps/api/v1/tests/test_views/test_programs.py
MatthewPiatetsky: switched

course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/factories.py
MatthewPiatetsky: changed, should we use factory.Faker('text') ?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
MatthewPiatetsky: I see, that makes sense. Changed serializers to select_related

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: Well as I mentioned above, I did look into specifying this in the edx_catalog_extensions app, but if we didn't have a default specified, the field would no have default, right? Is there a way we can not have the default and still satisfy the ask by analytics to default new courses to mooc?

course_discovery/apps/course_metadata/tests/factories.py
rlucioni: @MatthewPiatetsky `factory.Faker('word')` is closer to what you want.

course_discovery/apps/course_metadata/models.py
rlucioni: @MatthewPiatetsky sure, one possibility is modifying the `CourseRun` `save()` method so that the field is set to the right `ReportingType`.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
rlucioni: @MatthewPiatetsky it doesn't look like the serialization mixin was being used in this block to begin with. Don't we want to keep tabs on how many queries are being made by this view?

course_discovery/apps/course_metadata/models.py
```diff
@@ -330,6 +330,10 @@ def reindex_course_runs(self):
             index.update_object(course_run)
 
 
+class ReportingType(AbstractNamedModel):
```
clintonb: Why is this a separate model instead of a field? I'm in favor of using a field, and including a large comment explaining that this field is experimental and likely to be removed in the future. No need to try to put lipstick on the pig.

course_discovery/apps/course_metadata/models.py
clintonb: Remove the default and make the field nullable.

MatthewPiatetsky: Well we do want to allow setting the field to something else. I guess we could check on save if the field is not set, and set it to mooc if that is the case. Wouldn't we risk running into the same issue though if that record is removed like you said?

course_discovery/apps/api/serializers.py
```diff
@@ -363,11 +363,12 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
     instructors = serializers.SerializerMethodField(help_text='This field is deprecated. Use staff.')
     staff = PersonSerializer(many=True)
     level_type = serializers.SlugRelatedField(read_only=True, slug_field='name')
+    reporting_type = serializers.SlugRelatedField(slug_field='name', queryset=ReportingType.objects.all())
 
     @classmethod
     def prefetch_queryset(cls):
         queryset = super().prefetch_queryset()
-        return queryset.select_related('language', 'video').prefetch_related(
+        return queryset.select_related('language', 'video', 'reporting_type').prefetch_related(
```
rlucioni: @MatthewPiatetsky this is the right direction. Can you verify that there is a test case serializing multiple `CourseRuns` with linked `ReportingTypes`? We should make sure the `select_related` is actually working.

course_discovery/apps/course_metadata/models.py
rlucioni: @MatthewPiatetsky the code could be smart enough to not do anything if a filter for `ReportingTypes` with name `mooc` doesn't return any results.

course_discovery/apps/api/serializers.py
```diff
@@ -726,6 +727,18 @@ class Meta:
         )
 
 
+class ReportingTypeSerializer(serializers.ModelSerializer):
```
clintonb: This is not used, and should be deleted.

course_discovery/apps/api/serializers.py
```diff
@@ -17,8 +17,8 @@
 from course_discovery.apps.catalogs.models import Catalog
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
-    Course, CourseRun, Image, Organization, Person, Prerequisite, Seat, Subject, Video, Program, ProgramType, FAQ,
```
clintonb: Nit: no need to touch FAQ

clintonb (changes requested):

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: Ok I can change it to use a choice field, I did not get the impression that this field would be removed in the future from my conversation with analytics, but if we're ok hardcoding those terms that is fine. I thought that since we usually don't want to hardcoded edx specific things we'd want them in a catalog extensions migration like I have done here

clintonb: The plan I have been discussing is to take the simplest approach for now—choices. We will revisit this after we better understand all of the use cases. That could mean multiple fields. It could mean something else altogether.

clintonb: Eh...let's avoid the re-work on this PR. No solution will be ideal at this time. Keep the foreign key if that's what you want to go with.

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -63,7 +63,7 @@ def test_list(self):
 
         OrganizationFactory.create_batch(3)
 
-        with self.assertNumQueries(5):
+        with self.assertNumQueries(6):
```
MatthewPiatetsky: looking into why this changed

MatthewPiatetsky:
@rlucioni @clintonb i've done some refactoring based on @clintonb suggestion that should make most of your comments N/A

course_discovery/apps/course_metadata/models.py
```diff
@@ -332,6 +332,13 @@ def reindex_course_runs(self):
 
 class CourseRun(TimeStampedModel):
     """ CourseRun model. """
+    REPORTING_TYPE_CHOICES = (
```
clintonb: Use [django-choices](https://github.com/bigjason/django-choices).

course_discovery/apps/course_metadata/models.py
```diff
@@ -332,6 +332,13 @@ def reindex_course_runs(self):
...
+        ('mooc', _('mooc')),
```
clintonb: Unmark for migration. We don't want anyone else sharing our shame.

course_discovery/apps/api/serializers.py
```diff
@@ -357,6 +357,9 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
         read_only=True, slug_field='code', source='language',
         help_text=_('Language in which the course is administered')
     )
+    reporting_type = serializers.ChoiceField(
```
clintonb: You shouldn't need to specify this field explicitly. DRF knows how to validate choices.

MatthewPiatetsky:
@rlucioni @clintonb retagging

clintonb (changes requested):

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: can you clarify how to unmark for migration? Also, will the default value still work if the migration is not present?

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: changed to django-choices and moved to choices.py

course_discovery/apps/api/serializers.py
MatthewPiatetsky: removed

course_discovery/apps/course_metadata/choices.py
```diff
@@ -20,3 +20,11 @@ class ProgramStatus(DjangoChoices):
     Active = ChoiceItem('active', _('Active'))
     Retired = ChoiceItem('retired', _('Retired'))
     Deleted = ChoiceItem('deleted', _('Deleted'))
+
+
+class ReportingType(DjangoChoices):
+    mooc = ChoiceItem('mooc', _('mooc'))
```
clintonb: These don't need to be translated.

course_discovery/apps/course_metadata/models.py
```diff
@@ -389,6 +392,7 @@ class CourseRun(TimeStampedModel):
         default=False,
         help_text=_('Indicates whether the course relation has been manually overridden.')
     )
+    reporting_type = models.CharField(max_length=255, choices=ReportingType.choices, default='mooc')
```
clintonb: `ReportingType.mooc`

clintonb (changes requested):

course_discovery/apps/course_metadata/choices.py
MatthewPiatetsky: do i just need to do mooc = ChoiceItem('mooc', 'mooc')?
changed

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: changed

clintonb (approved):

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -53,8 +53,8 @@ def test_get_exclude_deleted_programs(self):
 
         with self.assertNumQueries(12):
             response = self.client.get(url)
-            self.assertEqual(response.status_code, 200)
-            self.assertEqual(response.data.get('programs'), [])
+        assert response.status_code == 200
+        self.assertEqual(response.data.get('programs'), [])
```
rlucioni: `assert` statement, please.

rlucioni (approved):

clintonb:

PR #630, opened 2017-02-03
WL-966 Allow for use of CourseRunSerializer when calling the programs list endpoint

afzaledx:
This PR allows for the ability to return fully serialized course runs from the program list API endpoint by including the `use_minimal_course_run_serializer` parameter. This is needed so that we can display course run price information on the WL program detail page.

**Sandbox URL**: https://discovery-edx.sandbox.edx.org/api/v1/programs/?use_minimal_course_run_serializer=0

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -559,7 +559,11 @@ class MinimalProgramCourseSerializer(MinimalCourseSerializer):
         if self.context.get('published_course_runs_only'):
             course_runs = [course_run for course_run in course_runs if course_run.status == CourseRunStatus.Published]
 
-        return MinimalCourseRunSerializer(
+        serializer_class = MinimalCourseRunSerializer
+        if self.context.get('use_full_course_serializer', False):
+            serializer_class = CourseRunSerializer
+
+        return serializer_class(
             course_runs,
             many=True,
             context={
@@ -610,6 +614,7 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
                 'exclude_utm': self.context.get('exclude_utm'),
                 'program': program,
                 'course_runs': course_runs,
+                'use_full_course_serializer': self.context.get('use_full_course_serializer', False),
             }
         )
 
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -832,6 +832,35 @@ class ProgramSerializerTests(MinimalProgramSerializerTests):
         expected = self.get_expected_data(program, request)
         self.assertDictEqual(serializer.data, expected)
 
+    def test_use_full_course_serializer(self):
+        """
+        Verify that we can use the `use_full_course_serializer` parameter to toggle the CourseRun
+        serializer used when returning program data.
+        """
+        request = make_request()
+        program = self.create_program()
+
+        # Assert that the seats are not available when fetched without any flag
+        program_without_seats = self.serializer_class(program, context={'request': request}).data
+        course_run_without_seats = program_without_seats['courses'][0]['course_runs'][0]
+        assert 'seats' not in course_run_without_seats
+
+        # Assert that the seats are not available when fetched with the use_full_course_serializer param set to 0
+        program_without_seats = self.serializer_class(program, context={
+            'request': request,
+            'use_full_course_serializer': 0
+        }).data
+        course_run_without_seats = program_without_seats['courses'][0]['course_runs'][0]
+        assert 'seats' not in course_run_without_seats
+
+        # Assert that the seats are available when fetched with the use_full_course_serializer param set to 1
+        program_with_seats = self.serializer_class(
+            program,
+            context={'request': request, 'use_full_course_serializer': 1}
+        ).data
+        course_run_with_seats = program_with_seats['courses'][0]['course_runs'][0]
+        assert 'seats' in course_run_with_seats
+
 
 class ProgramTypeSerializerTests(TestCase):
     serializer_class = ProgramTypeSerializer
--- a/course_discovery/apps/api/v1/views/programs.py
+++ b/course_discovery/apps/api/v1/views/programs.py
@@ -37,6 +37,7 @@ class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
         context.update({
             'published_course_runs_only': get_query_param(self.request, 'published_course_runs_only'),
             'exclude_utm': get_query_param(self.request, 'exclude_utm'),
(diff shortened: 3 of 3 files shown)


afzaledx:
@douglashall

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -37,6 +37,7 @@ def get_serializer_context(self, *args, **kwargs):
         context.update({
             'published_course_runs_only': get_query_param(self.request, 'published_course_runs_only'),
             'exclude_utm': get_query_param(self.request, 'exclude_utm'),
+            'course_run_with_seats': get_query_param(self.request, 'course_run_with_seats'),
```
douglashall: Rename to `include_course_run_seats`.

douglashall: Rename to `use_minimal_course_run_serializer` and default it to True.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -831,6 +831,32 @@ def test_data_without_course_sorting(self):
         expected = self.get_expected_data(program, request)
         self.assertDictEqual(serializer.data, expected)
 
+    def test_use_minimal_course_run_serializer(self):
+
```
douglashall: Remove extra whitespace.

douglashall:
@edx/ecommerce Please review.

course_discovery/apps/api/serializers.py
```diff
@@ -559,14 +559,24 @@ def get_course_runs(self, course):
...
 
-        return MinimalCourseRunSerializer(
-            course_runs,
-            many=True,
-            context={
-                'request': self.context.get('request'),
-                'exclude_utm': self.context.get('exclude_utm'),
-            }
-        ).data
+        if self.context.get('use_minimal_course_run_serializer', 1):
```
clintonb: Invert, please, to ensure that the default is always the minimal serializer: `use_full_course_serializer` default to `False`.

clintonb: Also, simplify:

```python
serializer_class = CourseRunSerializer if self.context.get('use_full_course_serializer', False) else MinimalCourseRunSerializer

return serializer_class(
  course_runs, 
  many=True,
  context={
    'request': self.context.get('request'),
    'exclude_utm': self.context.get('exclude_utm'),
  }
).data
```

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -831,6 +831,31 @@ def test_data_without_course_sorting(self):
         expected = self.get_expected_data(program, request)
         self.assertDictEqual(serializer.data, expected)
 
+    def test_use_minimal_course_run_serializer(self):
```
clintonb: Add a docstring.

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -70,6 +71,12 @@ def list(self, request, *args, **kwargs):
               type: integer
               paramType: query
               multiple: false
+            - name: use_minimal_course_run_serializer
```
clintonb: use_full_course_serializer

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -70,6 +71,12 @@ def list(self, request, *args, **kwargs):
...
+              description: Use the MinimalCourseRunSerializer when 1, CourseRunSerializer when 0.
```
clintonb: Return all serialized course information instead fo a minimal amount of information.

clintonb (changes requested):

rlucioni:
Why not make a request to the CourseRun API endpoint for the course runs you need more information about?

douglashall:
@rlucioni We are trying to avoid a round trip for every course run in a program. We want to use the program endpoint to get all the information we will need to render the program detail page in LMS (this includes the course price).

rlucioni:
@douglashall woah. First, you can make a single batch request for the course runs you care about. Second, if this is for the program detail page, why are you modifying the list view? Third, we already have an LMS program detail page. Why are you making a new one?

douglashall:
@rlucioni Where is the existing LMS program detail page?

rlucioni:
@douglashall existing program views in the LMS are at https://github.com/edx/edx-platform/blob/master/lms/djangoapps/learner_dashboard/views.py. They are being cleaned up as we speak as part of [ECOM-4422](https://openedx.atlassian.net/browse/ECOM-4422).

douglashall:
@rlucioni Ah, got it. Looks like we are using different definitions of "program detail page". The program detail page we are building for WL sites is the one for the marketing site that lives within LMS (WIP PRs [here](https://github.com/edx/edx-platform/pull/14339) and [here](https://github.com/edx/edx-platform/pull/14450)).

I think we would prefer to make a single catalog API request that contains all the information we need for the program detail page (both Program data and fully serialized CourseRun data). This way we don't have to make one call to get the Program data and then another to get the CourseRun data.

douglashall:
@clintonb Your comments have been addressed. This is ready for another pass. Thanks.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -831,6 +831,35 @@ def test_data_without_course_sorting(self):
...
+        Verify that we can use the `use_full_course_serializer` parameter to toggle the CourseRun
+        serializer used when returning program data.
+        """
+        request = make_request()
+        program = self.create_program()
+
+        # Assert that the seats are not available when fetched without any flag
+        program_without_seats = self.serializer_class(program, context={'request': request}).data
+        course_run_without_seats = program_without_seats['courses'][0]['course_runs'][0]
+        self.assertNotIn('seats', course_run_without_seats)
```
rlucioni: This repo uses pytest for testing. These unittest-style assertions will work, but could you please use plain `assert` statements instead? For example, this would become `assert 'seats' not in course_run_without_seats`.

rlucioni (commented):
@douglashall I see, thanks.

course_discovery/apps/api/tests/test_serializers.py
douglashall: Sounds good. Done.

clintonb (approved):
For future reference, please include the ticket number on the last line of the commit message. This helps us track down stories if we have questions later.

rlucioni (approved):
@douglashall please keep my team in the loop as you integrate use of this option elsewhere. Serialization of CourseRuns nested in Programs has suffered from performance issues in the past, and we'd like to prevent similar problems going forward.

PR #633, opened 2017-02-06
Fix value error in instructor lookup

tasawernawaz:
[ECOM-7083](https://openedx.atlassian.net/browse/ECOM-7083)

### DIFF
--- a/course_discovery/apps/course_metadata/lookups.py
+++ b/course_discovery/apps/course_metadata/lookups.py
@@ -66,13 +66,9 @@ class PersonAutocomplete(autocomplete.Select2QuerySetView):
         return []
 
     def get_result_label(self, result):
-        profile_image = result.profile_image_url
-        if hasattr(result.profile_image, 'url'):
-            profile_image = result.profile_image.url
-
         context = {
             'uuid': result.uuid,
-            'profile_image': profile_image,
+            'profile_image': result.get_profile_image_url,
             'full_name': result.full_name,
             'position': result.position if hasattr(result, 'position') else None
         }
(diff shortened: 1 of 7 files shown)


tasawernawaz:
@bderusha Please review this.

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -67,7 +67,7 @@ def get_queryset(self):
 
     def get_result_label(self, result):
         profile_image = result.profile_image_url
-        if hasattr(result.profile_image, 'url'):
+        if result.profile_image and hasattr(result.profile_image, 'url'):
```
waheedahmed: What if both fields have values then `result.profile_image.url` will replace the value of `result.profile_image_url`,  so which ones priority is high?

course_discovery/apps/publisher/forms.py
```diff
@@ -25,8 +25,11 @@ def label_from_instance(self, obj):
 
 class PersonModelMultipleChoice(forms.ModelMultipleChoiceField):
     def label_from_instance(self, obj):
+        profile_image = obj.profile_image_url
+        if obj.profile_image and hasattr(obj.profile_image, 'url'):
```
waheedahmed: Same question here.

waheedahmed (approved):
Just one question

course_discovery/apps/course_metadata/lookups.py
waheedahmed: @bderusha What do you think?

bderusha: @waheedahmed I think you're right to question this.  Probably makes more sense to give priority to `result.profile_image_url` since that's the current default.  Only if we don't have that should we fall back to the uploaded image

course_discovery/apps/publisher/forms.py
bderusha: same answer

bderusha (approved):
Once default swapped

PR #634, opened 2017-02-06
Added simple endpoint for PATCH for course_runs api to allow the modi…

davec-edx:
Addition of a PATCH verb to the API

ECOM-7024

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -39,7 +39,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         """ Verify the endpoint returns the details for a single course. """
         url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
 
-        with self.assertNumQueries(9):
+        with self.assertNumQueries(10):
             response = self.client.get(url)
 
         assert response.status_code == 200
@@ -51,7 +51,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
 
         url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
 
-        with self.assertNumQueries(12):
+        with self.assertNumQueries(13):
             response = self.client.get(url)
         assert response.status_code == 200
         assert response.data.get('programs') == []
@@ -66,7 +66,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
         url += '?include_deleted_programs=1'
 
-        with self.assertNumQueries(12):
+        with self.assertNumQueries(13):
             response = self.client.get(url)
         assert response.status_code == 200
         assert response.data == \
@@ -78,10 +78,10 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
 
         url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
 
-        with self.assertNumQueries(12):
+        with self.assertNumQueries(13):
             response = self.client.get(url)
             assert response.status_code == 200
-            self.assertEqual(response.data.get('programs'), [])
+            assert response.data.get('programs') == []
 
     def test_get_include_unpublished_programs(self):
         """
@@ -93,17 +93,47 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
         url += '?include_unpublished_programs=1'
 
-        with self.assertNumQueries(12):
+        with self.assertNumQueries(13):
             response = self.client.get(url)
         assert response.status_code == 200
         assert response.data == \
             self.serialize_course_run(self.course_run, extra_context={'include_unpublished_programs': True})
 
+    def test_partial_update(self):
+        """ Verify the endpoint supports partially updating a course_run's fields, provided user has permission. """
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
+
+        expected_min_effort = 867
+        expected_max_effort = 5309
+        data = {
+            'max_effort': expected_max_effort,
+            'min_effort': expected_min_effort,
+        }
+
+        # Update this course_run with the new info
+        response = self.client.patch(url, data, format='json')
+        assert response.status_code == 200
+
+        # refresh and make sure we have the new effort levels
+        self.course_run.refresh_from_db()
+
+        assert self.course_run.max_effort == expected_max_effort
+        assert self.course_run.min_effort == expected_min_effort
+
+    def test_partial_update_bad_permission(self):
+        """ Verify partially updating will fail if user doesn't have permission. """
+        user = UserFactory(is_staff=False, is_superuser=False)
+        self.client.force_authenticate(user)
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
+
+        response = self.client.patch(url, {}, format='json')
+        assert response.status_code == 403
+
     def test_list(self):
         """ Verify the endpoint returns a list of all catalogs. """
         url = reverse('api:v1:course_run-list')
 
-        with self.assertNumQueries(11):
+        with self.assertNumQueries(12):
             response = self.client.get(url)
 
         assert response.status_code == 200
@@ -116,7 +146,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         """ Verify the endpoint returns a list of all catalogs sorted by course start date. """
         url = '{root}?ordering=start'.format(root=reverse('api:v1:course_run-list'))
 
-        with self.assertNumQueries(11):
+        with self.assertNumQueries(12):
             response = self.client.get(url)
 
         assert response.status_code == 200
@@ -132,7 +162,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         query = 'title:Some random title'
         url = '{root}?q={query}'.format(root=reverse('api:v1:course_run-list'), query=query)
 
-        with self.assertNumQueries(37):
+        with self.assertNumQueries(38):
             response = self.client.get(url)
 
         actual_sorted = sorted(response.data['results'], key=lambda course_run: course_run['key'])
@@ -248,7 +278,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         url = '{}?{}'.format(reverse('api:v1:course_run-contains'), qs)
 
         response = self.client.get(url)
-        self.assertEqual(response.status_code, 400)
+        assert response.status_code == 400
 
     def test_contains_multiple_course_runs(self):
         qs = urllib.parse.urlencode({
@@ -281,4 +311,4 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         url = '{}?{}'.format(reverse('api:v1:course_run-contains'), qs)
 
         response = self.client.get(url)
-        self.assertEqual(response.status_code, 400)
+        assert response.status_code == 400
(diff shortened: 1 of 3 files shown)


clintonb (changes requested):
Permissions need to be fixed. We don't want any authenticated user to be able to modify course runs.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,29 @@ def test_get_include_unpublished_programs(self):
                 self.serialize_course_run(self.course_run, extra_context={'include_unpublished_programs': True})
             )
 
+    def test_partial_update(self):
+        """ Verify the endpoint supports partially updating a course_run's fields. """
```
MatthewPiatetsky: this works, but what if we used the CourseRun factory and updated all the fields, then checked that all of them got updated. If all the fields including m2m actually work, using the factory could be a simple way to ensure that new fields also get updated without having to update the test

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,29 @@ def test_get_include_unpublished_programs(self):
...
+        }
+
+        # Update this course_run with the new info
+        response = self.client.patch(url, data, format='json')
+        self.assertEqual(response.status_code, 200)
+
+        # re-read and make sure we have the newly calculated effort levels
+        updated_course_run = CourseRun.objects.get(key=self.course_run.key)
+
+        assert updated_course_run.max_effort == max_effort
```
MatthewPiatetsky: if you go with the factory you could probably have a loop of assert statements using the original_course_run.__dict__.items() and updated_course_run.__dict__.items() or something like that

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -23,7 +23,7 @@
 class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestCase):
     def setUp(self):
         super(CourseRunViewSetTests, self).setUp()
-        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.user = UserFactory(is_staff=False, is_superuser=True)
```
clintonb: This does not need to be changed.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,47 @@ def test_get_include_unpublished_programs(self):
                 self.serialize_course_run(self.course_run, extra_context={'include_unpublished_programs': True})
             )
 
+    def test_partial_update(self):
+        """ Verify the endpoint supports partially updating a course_run's fields, provided user has permission. """
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
+        self.request.user.is_staff = True
```
clintonb: This has no effect on the test.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,47 @@ def test_get_include_unpublished_programs(self):
...
+            'max_effort': max_effort,
+            'min_effort': min_effort,
+        }
+
+        # Update this course_run with the new info
+        response = self.client.patch(url, data, format='json')
+        self.assertEqual(response.status_code, 200)
+
+        # re-read and make sure we have the newly calculated effort levels
+        updated_course_run = CourseRun.objects.get(key=self.course_run.key)
```
clintonb: Use [`refresh_from_db`](https://docs.djangoproject.com/en/1.10/ref/models/instances/#django.db.models.Model.refresh_from_db).

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,47 @@ def test_get_include_unpublished_programs(self):
                 self.serialize_course_run(self.course_run, extra_context={'include_unpublished_programs': True})
             )
 
+    def test_partial_update(self):
+        """ Verify the endpoint supports partially updating a course_run's fields, provided user has permission. """
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
+        self.request.user.is_staff = True
+
+        # Get a course_run, and calculate new min_ and max_effort
+        original_course_run = CourseRun.objects.get(key=self.course_run.key)
```
clintonb: There is no need to reload this object from the database. Your test values don't need to depend on the original values.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,47 @@ def test_get_include_unpublished_programs(self):
...
+        max_effort = original_course_run.max_effort + 867
+        min_effort = original_course_run.min_effort + 5309
+        data = {
+            'max_effort': max_effort,
+            'min_effort': min_effort,
+        }
+
+        # Update this course_run with the new info
+        response = self.client.patch(url, data, format='json')
+        self.assertEqual(response.status_code, 200)
```
clintonb: `assert response.status_code == 200`

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -127,6 +128,13 @@ def list(self, request, *args, **kwargs):
         """
         return super(CourseRunViewSet, self).list(request, *args, **kwargs)
 
+    def partial_update(self, request, *args, **kwargs):
+        """ Update one, or more, fields for a course run. """
+        if self.request.user.is_authenticated() and self.request.user.is_staff:
```
clintonb: This is not how we handle permissions. See http://www.django-rest-framework.org/api-guide/permissions/.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,47 @@ def test_get_include_unpublished_programs(self):
...
+        updated_course_run = CourseRun.objects.get(key=self.course_run.key)
+
+        assert updated_course_run.max_effort == max_effort
+        assert updated_course_run.min_effort == min_effort
+
+    def test_partial_update_bad_permission(self):
+        """ Verify the endpoint supports partially updating will fail if user doesn't have permission. """
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
+
+        # Get a course_run, and calculate new min_ and max_effort
```
clintonb: This test should be much simpler. There is no need to worry about the data being sent to the API because we should not get to that level.

```python
response = self.client.patch(url,{}, format='json')
assert response.status_code == 403
```

clintonb (changes requested):

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
davec-edx: Everything was being tested with is_staff set, where most don't need anything more than being authenticated.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
davec-edx: In conjunction with not setting it for other tests, it needs to be set for this test (and unset for the corresponding negative test)

davec-edx:
@clintonb Please re-review -- thanks

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
davec-edx: Yes, I thought about that, but thought it would be overkill, I just chose a couple fields at random.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
clintonb: Superuser trumps staff. Given that all superusers are staff, this change is not necessary.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
clintonb: If you revert the change to `self.user` this line can go away.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
clintonb: Please clean this up. There is no need to load the object from the DB. Doing so is unnecessarily confusing.

```python
expected_max_effort = 1000
expected_min_effort = 100
data = {
  'max_effort': expected_max_effort,
  'min_effort': expected_min_effort,
}

response = self.client.patch(url, data, format='json')
assert response.status_code == 200

self.course_run.refresh_from_db()
assert self.course_run.max_effort == expected_max_effort
assert self.course_run.min_effort == expected_min_effort
```

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,39 @@ def test_get_include_unpublished_programs(self):
...
+        updated_course_run = CourseRun.objects.get(key=self.course_run.key)
+
+        assert updated_course_run.max_effort == max_effort
+        assert updated_course_run.min_effort == min_effort
+
+    def test_partial_update_bad_permission(self):
+        """ Verify the endpoint supports partially updating will fail if user doesn't have permission. """
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.course_run.key})
+
+        # Update this course_run with the new info, NOTE this should fail with a 404
```
clintonb: This is not valid.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -103,6 +103,39 @@ def test_get_include_unpublished_programs(self):
...
+        response = self.client.patch(url, {}, format='json')
+        assert response.status_code != 200
```
clintonb: `assert response.status_code == 403`

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -14,14 +16,14 @@
...
-class CourseRunViewSet(PartnerMixin, viewsets.ReadOnlyModelViewSet):
+class CourseRunViewSet(PartnerMixin, viewsets.ModelViewSet):
     """ CourseRun resource. """
     filter_backends = (DjangoFilterBackend, OrderingFilter)
     filter_class = filters.CourseRunFilter
     lookup_field = 'key'
     lookup_value_regex = COURSE_RUN_ID_REGEX
     ordering_fields = ('start',)
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (IsAuthenticated, )
```
clintonb: Revert

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -127,6 +129,11 @@ def list(self, request, *args, **kwargs):
         """
         return super(CourseRunViewSet, self).list(request, *args, **kwargs)
 
+    @method_decorator(staff_member_required)
```
clintonb: This is not what we discussed yesterday. You should be using [`DjangoModelPermissions`](http://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions).

clintonb (changes requested):

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -14,14 +14,14 @@
...
-class CourseRunViewSet(PartnerMixin, viewsets.ReadOnlyModelViewSet):
+class CourseRunViewSet(PartnerMixin, viewsets.ModelViewSet):
     """ CourseRun resource. """
     filter_backends = (DjangoFilterBackend, OrderingFilter)
     filter_class = filters.CourseRunFilter
     lookup_field = 'key'
     lookup_value_regex = COURSE_RUN_ID_REGEX
     ordering_fields = ('start',)
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (DRYPermissions,)
```
clintonb: `permission_classes = (IsAuthenticated, DjangoModelPermissions)`

We don't need instance-specific permissions, and we always want users to be authenticated. Use `DjangoModelPermissions`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -338,7 +339,7 @@ def reindex_course_runs(self):
             index.update_object(course_run)
 
 
-class CourseRun(TimeStampedModel):
+class CourseRun(ModelPermissionsMixin, TimeStampedModel):
```
clintonb: This is not necessary.

clintonb (changes requested):

clintonb (approved):

PR #641, opened 2017-02-08
ECOM-7079 Move ElasticSearchBoostConfig into settings

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -162,7 +162,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, APITestC
         query = 'title:Some random title'
         url = '{root}?q={query}'.format(root=reverse('api:v1:course_run-list'), query=query)
 
-        with self.assertNumQueries(38):
+        with self.assertNumQueries(36):
             response = self.client.get(url)
 
         actual_sorted = sorted(response.data['results'], key=lambda course_run: course_run['key'])
--- a/course_discovery/apps/edx_haystack_extensions/backends.py
+++ b/course_discovery/apps/edx_haystack_extensions/backends.py
@@ -1,6 +1,6 @@
 from haystack.backends.elasticsearch_backend import ElasticsearchSearchBackend, ElasticsearchSearchEngine
 
-from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
+from course_discovery.apps.edx_haystack_extensions.elasticsearch_boost_config import get_elasticsearch_boost_config
 
 
 class SimpleQuerySearchBackendMixin(object):
@@ -37,8 +37,7 @@ class SimpleQuerySearchBackendMixin(object):
         }
 
         # https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-function-score-query.html
-        function_score_config = ElasticsearchBoostConfig.get_solo().function_score
-
+        function_score_config = get_elasticsearch_boost_config()['function_score']
         function_score_config['query'] = {
             'query_string': simple_query
         }
--- a/course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
+++ b/course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
@@ -1,0 +1,83 @@
+def get_elasticsearch_boost_config():
+    elasticsearch_boost_config = {
+        'function_score': {
+            'boost_mode': 'sum',
+            'boost': 1.0,
+            'score_mode': 'sum',
+            'functions': [
+                {'filter': {'term': {'pacing_type_exact': 'self_paced'}}, 'weight': 1.0},
+                {'filter': {'term': {'type_exact': 'Professional Certificate'}}, 'weight': 1.0},
+                {'filter': {'term': {'type_exact': 'MicroMasters'}}, 'weight': 1.0},
+                {'linear': {'start': {'origin': 'now', 'decay': 0.95, 'scale': '1d'}}, 'weight': 5.0},
+
+                # Boost function for CourseRuns with enrollable paid Seats.
+                # We want to boost if:
+                #       - The course run has at least one enrollable paid Seat (has_enrollable_paid_seats is True)
+                # AND one of the following two conditions are true
+                #       - The paid_seat_enrollment_end is unspecified.
+                #       - The paid_seat_enrollment_end is in the future.
+                # We apply a weight of 1.0 to match the boost given for self paced courses.
+                {
+                    'filter': {
+                        'bool': {
+                            'must': [
+                                {'exists': {'field': 'has_enrollable_paid_seats'}},
+                                {'term': {'has_enrollable_paid_seats': True}}
+                            ],
+                            'should': [
+                                {'bool': {'must_not': {'exists': {'field': 'paid_seat_enrollment_end'}}}},
+                                {'range': {'paid_seat_enrollment_end': {'gte': 'now'}}}
+                            ]
+                        }
+                    },
+                    'weight': 1.0
+                },
+
+                # Boost function for enrollable CourseRuns.
+                # We want to boost if:
+                #   - enrollment_start and enrollment_end are unspecified
+                #   - enrollment_start is unspecified and enrollment_end is in the future
+                #   - enrollment_end is unspecified and enrollment_start is in the past
+                #   - enrollment_start is in the past and enrollment_end is in the future
+                # We apply a weight of 1.0 to match the boost given for self paced and enrollable paid courses.
+                {
+                    'filter': {
+                        'bool': {
+                            'should': [
+                                {'bool': {
+                                    'must_not': [
+                                        {'exists': {'field': 'enrollment_start'}},
+                                        {'exists': {'field': 'enrollment_end'}}
+                                    ]
+                                }},
+                                {'bool': {
+                                    'must_not': {'exists': {'field': 'enrollment_start'}},
+                                    'must': [
+                                        {'exists': {'field': 'enrollment_end'}},
+                                        {'range': {'enrollment_end': {'gt': 'now'}}}
+                                    ]
+                                }},
+                                {'bool': {
+                                    'must_not': {'exists': {'field': 'enrollment_end'}},
+                                    'must': [
+                                        {'exists': {'field': 'enrollment_start'}},
+                                        {'range': {'enrollment_start': {'lte': 'now'}}}
+                                    ]
+                                }},
+                                {'bool': {
+                                    'must': [
+                                        {'exists': {'field': 'enrollment_start'}},
+                                        {'exists': {'field': 'enrollment_end'}},
+                                        {'range': {'enrollment_start': {'lte': 'now'}}},
+                                        {'range': {'enrollment_end': {'gt': 'now'}}}
+                                    ]
+                                }}
+                            ]
+                        }
+                    },
+                    'weight': 1.0
+                }
+            ]
+        }
+    }
+    return elasticsearch_boost_config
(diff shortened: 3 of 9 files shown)


clintonb (changes requested):
Put the ticket number on the last line of your commit message.

How do we get edX.org information out of the base settings? As an OpenEdX user, I don't care about Professional Certificate.

MatthewPiatetsky:
@clintonb is there a way we can have settings in an edx only app override the base settings or something like that?  
(Though I'm also not sure if open edx users would care about other boosting either. Not all of it has our brand names, but all of it is specific to the behavior we want on our marketing site, no idea if others might want similar behavior).

clintonb:
@MatthewPiatetsky you could load the module dynamically like we do with fulfillment modules in edx/ecommerce:

* https://github.com/edx/ecommerce/blob/8fde125db17565c28cdf182fd35416c5c05b9dee/ecommerce/settings/_oscar.py#L80-L84
* https://github.com/edx/ecommerce/blob/8fde125db17565c28cdf182fd35416c5c05b9dee/ecommerce/extensions/fulfillment/api.py#L99-L100

Regardless of which implementation you use, remember to document this in the `docs/` directory.

course_discovery/apps/edx_haystack_extensions/tests/test_boosting.py
```diff
@@ -0,0 +1,147 @@
+import datetime
```
mikedikan, line 1: can you explain why all of this test code was moved?

MatthewPiatetsky: I'm in the middle of refactoring some things for this PR, but specifically Clinton suggested in an earlier comment to move the boosting config and functionality into an edx app since our config is specific to our needs

MatthewPiatetsky:
@clintonb @mikedikan did I address your comments

course_discovery/apps/edx_haystack_extensions/tests/mixins.py
```diff
@@ -37,7 +37,7 @@ class SimpleQuerySearchBackendMixinTestMixin(SearchBackendTestMixin):
     }
 
     def _default_function_score(self):
-        function_score = {'function_score': ElasticsearchBoostConfig.get_solo().function_score}
+        function_score = ELASTICSEARCH_BOOST_CONFIG
```
clintonb: Name this variable appropriately to avoid confusion. Alternatively, ditch the variable and use the constant directly on the line below.

clintonb (approved):

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -162,7 +162,7 @@ def test_list_query(self):
         query = 'title:Some random title'
         url = '{root}?q={query}'.format(root=reverse('api:v1:course_run-list'), query=query)
 
-        with self.assertNumQueries(38):
+        with self.assertNumQueries(36):
```
amangano-edx, line 165: Why do some of the query counts change by one and others by two?

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -1,6 +1,6 @@
 from haystack.backends.elasticsearch_backend import ElasticsearchSearchBackend, ElasticsearchSearchEngine
 
-from course_discovery.apps.edx_haystack_extensions.models import ElasticsearchBoostConfig
+from course_discovery.apps.edx_haystack_extensions.elasticsearch_boost_config import ELASTICSEARCH_BOOST_CONFIG
```
amangano-edx: Just curious, but the boost config was moved to an app because it is edx-specific and not everyone in the community will need it. Is the edx_haystack_extensions app also meant to be edx-specific?

course_discovery/apps/edx_haystack_extensions/tests/mixins.py
```diff
@@ -37,9 +37,8 @@ class SimpleQuerySearchBackendMixinTestMixin(SearchBackendTestMixin):
     }
 
     def _default_function_score(self):
-        function_score = {'function_score': ElasticsearchBoostConfig.get_solo().function_score}
-        function_score['function_score']['query'] = {'query_string': self.simple_query}
-        return function_score
+        ELASTICSEARCH_BOOST_CONFIG['function_score']['query'] = {'query_string': self.simple_query}
+        return ELASTICSEARCH_BOOST_CONFIG
```
amangano-edx: This is modifying the constant, which I believe may have consequences for tests that run later on in the suite (they may be flaky since the `ELASTICSEARCH_BOOST_CONFIG` was modified in a prior test and is no longer in a consistent state).

course_discovery/apps/edx_haystack_extensions/tests/mixins.py
```diff
@@ -70,27 +69,29 @@ def test_build_search_kwargs_specific_qs_no_filter(self):
...
-                            'type': 'micromasters'
-                        }
-                    },
-                    'weight': 10.0
-                }
-            ],
-            'boost': 5.0,
-            'score_mode': 'multiply',
-            'boost_mode': 'sum'
+        TEST_ELASTICSEARCH_BOOST_CONFIG = {
```
amangano-edx: Why change this to a constant if it's only used for this one test?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
MatthewPiatetsky: i looked at this test on master and it was getting the boost config twice, probably due to what we were doing in the backends.py file

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: i believe so  
it would be unfortunate if the name for a non edx specific app started with edx

amangano-edx: Aren't the edx_haystack_extensions required for the rest of the IDA to function properly? If you removed the edx_haystack_extensions app, would everything else continue to work properly, just without boosting? Isn't the synonym and typeahead behavior dependent on modifications made in that app?

The point I'm trying to make is maybe we might want to add the boost config here in a way that degrades gracefully if it's not available (in case someone uses course-discovery without our specific boosting).

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
davec-edx: course-runs was also changed for permissions so that could cause an extra query for perms

course_discovery/apps/edx_haystack_extensions/backends.py
clintonb: `edx_haystack_extensions` is not required, and is only for edx.org.

course_discovery/apps/edx_haystack_extensions/tests/mixins.py
```diff
@@ -37,9 +37,9 @@ class SimpleQuerySearchBackendMixinTestMixin(SearchBackendTestMixin):
     }
 
     def _default_function_score(self):
-        function_score = {'function_score': ElasticsearchBoostConfig.get_solo().function_score}
-        function_score['function_score']['query'] = {'query_string': self.simple_query}
-        return function_score
+        boost_config = ELASTICSEARCH_BOOST_CONFIG
+        boost_config['function_score']['query'] = {'query_string': self.simple_query}
```
amangano-edx: doesn't this still modify the constant (since `boost_config` is just a reference to the same dictionary as `ELASTICSEARCH_BOOST_CONFIG`?

amangano-edx (changes requested):

course_discovery/apps/edx_haystack_extensions/tests/mixins.py
amangano-edx: I think you need to create a deep copy of ELASTICSEARCH_BOOST_CONFIG to avoid modifying the original object.

MatthewPiatetsky: `X = 1
def changeX():
	X = 2
print (X)

1`

MatthewPiatetsky:
```
X = {'1': '2'}
def changeX():
	X['1'] = ['3']
print (X)

{'1': '2'}
```
I don't think this is an issue

MatthewPiatetsky: good point

course_discovery/apps/edx_haystack_extensions/backends.py
MatthewPiatetsky: @amangano-edx isn't the boost config only used in the edx_haystack_extensions app?  
Not sure about typeahead etc... but that probably wouldn't be part of this PR

MatthewPiatetsky:
@amangano-edx did I address your comments?

course_discovery/apps/edx_haystack_extensions/backends.py
amangano-edx: Yes, AFAIK the boost config is only used here. My comment about typeahead and synonyms was relevant to the question about whether this app (edx_haystack_extensions) is truly edx-specific or if it is always necessary for the IDA to run properly.

course_discovery/apps/edx_haystack_extensions/backends.py
```diff
@@ -37,7 +37,7 @@ def build_search_kwargs(self, *args, **kwargs):
         }
 
         # https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-function-score-query.html
-        function_score_config = ElasticsearchBoostConfig.get_solo().function_score
+        function_score_config = ELASTICSEARCH_BOOST_CONFIG['function_score']
```
amangano-edx: I think we need to create a deep copy of the config here. Otherwise, I believe we risk modifying it whenever a query is run (line 42) and leaving it in an inconsistent state.

course_discovery/apps/edx_haystack_extensions/tests/mixins.py
```diff
@@ -37,9 +37,9 @@ class SimpleQuerySearchBackendMixinTestMixin(SearchBackendTestMixin):
     }
 
     def _default_function_score(self):
-        function_score = {'function_score': ElasticsearchBoostConfig.get_solo().function_score}
-        function_score['function_score']['query'] = {'query_string': self.simple_query}
-        return function_score
+        boost_config = dict(ELASTICSEARCH_BOOST_CONFIG)
```
amangano-edx: I just want to point out that this makes a shallow copy of the config. All of the nested dictionaries and lists within the config would not be copied and would still refer to the original objects. This might not be a problem here, since we only modify the outermost dictionary in line 41. However, I think it'd be safer and less error prone to do a deep copy instead, in case more modifications need to be made to this in the future. https://docs.python.org/3.5/library/copy.html

MatthewPiatetsky:
@amangano-edx  
I just moved the variable into a function. That should address all the variable modification comments

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
```diff
@@ -0,0 +1,83 @@
+def get_elasticsearch_boost_config():
```
amangano-edx, line 1: This seems fine, though I wonder if it might make sense to just have this function become a query builder of sorts and move lines 41-43 from `course_discovery/apps/edx_haystack_extensions/backends.py` into here? It could take the query as an argument.

amangano-edx (approved):

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
MatthewPiatetsky: We use the config differently in the test files and could be differently in the future, I'd rather keep the specific modifications where they are needed and keep this generic

PR #642, opened 2017-02-08
ECOM-6820 Add marketable_course_runs_only_with_archived filter

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -514,6 +514,10 @@ class CourseWithProgramsSerializer(CourseSerializer):
             # closed - achieving this requires applying both the marketable and active filters.
             course_runs = course_runs.marketable().active()
 
+        if self.context.get('marketable_enrollable_course_runs_with_archived'):
+            # Same as "marketable_course_runs_only", but includes courses with an end date in the past
+            course_runs = course_runs.marketable().enrollable()
+
         if self.context.get('published_course_runs_only'):
             course_runs = course_runs.filter(status=CourseRunStatus.Published)
 
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -207,35 +207,26 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
         Verify that the marketable_course_runs_only option is respected, restricting returned
         course runs to those that are published, have seats, and can still be enrolled in.
         """
-        # Published course run with a seat, no enrollment start or end, and an end date in the future.
         enrollable_course_run = CourseRunFactory(
             status=CourseRunStatus.Published,
             end=datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=10),
             enrollment_start=None,
             enrollment_end=None,
+            course=self.course
         )
         SeatFactory(course_run=enrollable_course_run)
 
-        # Unpublished course run with a seat.
-        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished, course=self.course)
         SeatFactory(course_run=unpublished_course_run)
 
-        # Published course run with no seats.
-        no_seats_course_run = CourseRunFactory(status=CourseRunStatus.Published)
+        CourseRunFactory(status=CourseRunStatus.Published, course=self.course)
 
-        # Published course run with a seat and an end date in the past.
         closed_course_run = CourseRunFactory(
             status=CourseRunStatus.Published,
             end=datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=10),
+            course=self.course
         )
         SeatFactory(course_run=closed_course_run)
-
-        self.course.course_runs.add(
-            enrollable_course_run,
-            unpublished_course_run,
-            no_seats_course_run,
-            closed_course_run,
-        )
 
         serializer = self.serializer_class(
             self.course,
@@ -246,6 +237,46 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
             len(serializer.data['course_runs']),
             1 if marketable_course_runs_only else 4
         )
+
+    def test_marketable_enrollable_course_runs_with_archived(self):
+        """
+        Verify that the marketable_enrollable_course_runs_with_archived option is respected, restricting returned
+        course runs to those that are published, have seats, and can still be enrolled in
+        (including courses with an end date in the past.)
+        """
+        enrollable_course_run = CourseRunFactory(
+            status=CourseRunStatus.Published,
+            end=datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=10),
+            enrollment_start=None,
+            enrollment_end=None,
+            course=self.course
+        )
+        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished, course=self.course)
+        CourseRunFactory(status=CourseRunStatus.Published, course=self.course)
+        archived_course_run = CourseRunFactory(
+            status=CourseRunStatus.Published,
+            end=datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=10),
+            enrollment_start=None,
+            enrollment_end=None,
+            course=self.course
+        )
+        SeatFactory(course_run=unpublished_course_run)
+        SeatFactory(course_run=enrollable_course_run)
+        SeatFactory(course_run=archived_course_run)
+
+        context = {
+            'request': self.request,
+            'marketable_enrollable_course_runs_with_archived': 1
+        }
+
+        course_serializer = self.serializer_class(
+            self.course,
+            context=context
+        )
+
+        course_run_keys = [course_run['key'] for course_run in course_serializer.data['course_runs']]
+        # order doesn't matter
+        assert sorted(course_run_keys) == sorted([enrollable_course_run.key, archived_course_run.key])
 
     @ddt.data(0, 1)
     def test_published_course_runs_only(self, published_course_runs_only):
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -71,32 +71,27 @@ class CourseViewSetTests(SerializationMixin, APITestCase):
             end=datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=10),
             enrollment_start=None,
             enrollment_end=None,
+            course=self.course
         )
         SeatFactory(course_run=enrollable_course_run)
 
         # Unpublished course run with a seat.
-        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished, course=self.course)
         SeatFactory(course_run=unpublished_course_run)
 
         # Published course run with no seats.
-        no_seats_course_run = CourseRunFactory(status=CourseRunStatus.Published)
+        CourseRunFactory(status=CourseRunStatus.Published, course=self.course)
 
         # Published course run with a seat and an end date in the past.
         closed_course_run = CourseRunFactory(
             status=CourseRunStatus.Published,
             end=datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=10),
+            course=self.course
         )
         SeatFactory(course_run=closed_course_run)
 
-        self.course.course_runs.add(
-            enrollable_course_run,
-            unpublished_course_run,
-            no_seats_course_run,
-            closed_course_run,
-        )
-
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
-        url += '?marketable_course_runs_only={}'.format(marketable_course_runs_only)
+        url = '{}?marketable_course_runs_only={}'.format(url, marketable_course_runs_only)
         response = self.client.get(url)
 
         self.assertEqual(response.status_code, 200)
@@ -109,16 +104,45 @@ class CourseViewSetTests(SerializationMixin, APITestCase):
         )
 
     @ddt.data(1, 0)
+    def test_marketable_enrollable_course_runs_with_archived(self, marketable_enrollable_course_runs_with_archived):
+        """ Verify the endpoint filters course runs to those that are marketable and
+        enrollable, including archived course runs (with an end date in the past). """
+
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        CourseRunFactory(enrollment_start=None, enrollment_end=future, course=self.course)
+        CourseRunFactory(enrollment_start=None, enrollment_end=None, course=self.course)
+        CourseRunFactory(
+            enrollment_start=past, enrollment_end=future, course=self.course
+        )
+        CourseRunFactory(enrollment_start=future, course=self.course)
+        CourseRunFactory(enrollment_end=past, course=self.course)
+
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
+        url = '{}?marketable_enrollable_course_runs_with_archived={}'.format(
+            url, marketable_enrollable_course_runs_with_archived
+        )
+        response = self.client.get(url)
+
+        assert response.status_code == 200
+        assert response.data == self.serialize_course(
+            self.course,
+            extra_context={
+                'marketable_enrollable_course_runs_with_archived': marketable_enrollable_course_runs_with_archived
+            }
+        )
+
+    @ddt.data(1, 0)
     def test_get_include_published_course_run(self, published_course_runs_only):
         """
         Verify the endpoint returns hides unpublished programs if
         the 'published_course_runs_only' flag is set to True
         """
-        published_course_run = CourseRunFactory(status=CourseRunStatus.Published)
-        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
-        self.course.course_runs.add(published_course_run, unpublished_course_run)
+        CourseRunFactory(status=CourseRunStatus.Published, course=self.course)
+        CourseRunFactory(status=CourseRunStatus.Unpublished, course=self.course)
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
-        url += '?published_course_runs_only={}'.format(published_course_runs_only)
+        url = '{}?published_course_runs_only={}'.format(url, published_course_runs_only)
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -38,12 +38,10 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
 
     def get_serializer_context(self, *args, **kwargs):
         context = super().get_serializer_context(*args, **kwargs)
-        context.update({
-            'exclude_utm': get_query_param(self.request, 'exclude_utm'),
-            'include_deleted_programs': get_query_param(self.request, 'include_deleted_programs'),
-            'marketable_course_runs_only': get_query_param(self.request, 'marketable_course_runs_only'),
-            'published_course_runs_only': get_query_param(self.request, 'published_course_runs_only'),
-        })
+        query_params = ['exclude_utm', 'include_deleted_programs', 'marketable_course_runs_only',
+                        'marketable_enrollable_course_runs_with_archived', 'published_course_runs_only']
+        for query_param in query_params:
+            context[query_param] = get_query_param(self.request, query_param)
 
         return context
 
@@ -71,7 +69,14 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
               multiple: false
             - name: marketable_course_runs_only
               description: Restrict returned course runs to those that are published, have seats,
-                and can still be enrolled in.
+                and are enrollable or will be enrollable in the future
+              required: false
+              type: integer
+              paramType: query
+              mulitple: false
+            - name: marketable_enrollable_course_runs_with_archived
+              description: Restrict returned course runs to those that are published, have seats,
+                and can be enrolled in now. Includes archived courses.
               required: false
               type: integer
               paramType: query
--- a/course_discovery/apps/course_metadata/query.py
+++ b/course_discovery/apps/course_metadata/query.py
@@ -43,6 +43,27 @@ class CourseRunQuerySet(models.QuerySet):
             )
         )
 
+    def enrollable(self):
+        """ Returns course runs that are currently open for enrollment.
+
+        A course run is considered open for enrollment if its enrollment start date
+        has passed, is now or is None, AND its enrollment end date is in the future or is None.
+
+        Returns:
+            QuerySet
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        return self.filter(
+            (
+                Q(enrollment_end__gt=now) |
+                Q(enrollment_end__isnull=True)
+            ) & (
+                Q(enrollment_start__lte=now) |
+                Q(enrollment_start__isnull=True)
+            )
+
+        )
+
     def marketable(self):
         """ Returns CourseRuns that can be marketed to learners.
 
--- a/course_discovery/apps/course_metadata/tests/test_query.py
+++ b/course_discovery/apps/course_metadata/tests/test_query.py
@@ -68,6 +68,22 @@ class CourseRunQuerySetTests(TestCase):
 
         self.assertEqual(set(CourseRun.objects.active()), {active_enrollment_end, active_no_enrollment_end})
 
+    def test_enrollable(self):
+        """ Verify the method returns only course runs currently open for enrollment. """
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_start=past, enrollment_end=None)
+        enrollable_no_enrollment_start = CourseRunFactory(enrollment_start=None, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
+
+        # order doesn't matter
+        assert list(CourseRun.objects.enrollable().order_by('id')) == sorted([
+            enrollable, enrollable_no_enrollment_end, enrollable_no_enrollment_start
+        ], key=lambda x: x.id)
+
     def test_marketable(self):
         """ Verify the method filters CourseRuns to those with slugs. """
         course_run = CourseRunFactory()


course_discovery/apps/course_metadata/query.py
```diff
@@ -43,6 +44,23 @@ def active(self):
             )
         )
 
+    def enrollable(self):
+        """ Returns CourseRuns that have not yet ended and meet the following enrollment criteria:
+            - Open for enrollment
+            - OR will be open for enrollment in the future
+            - OR have no specified enrollment close date (e.g. self-paced courses)
```
amangano-edx: This conflicts with definitions of enrollable elsewhere (like the enrollable course boosting), which take into account enrollment_start date. Perhaps we should come up with a consensus for this term so that it may be used consistently within the code base.

MatthewPiatetsky: Could you spell out the definition in the boosting here?

course_discovery/apps/api/serializers.py
```diff
@@ -512,7 +512,7 @@ def get_course_runs(self, course):
             # should be excluded. As an unfortunate side-effect of the way we've marketed course
             # runs in the past - a course run could be marketed despite enrollment in that run being
             # closed - achieving this requires applying both the marketable and active filters.
-            course_runs = course_runs.marketable().active()
+            course_runs = course_runs.marketable().enrollable()
```
schenedx: I am worried that we decide to change the definition of this flag from `.active()` to `.enrollable()`. The Sailthru `marketable` variable is using this flag. Can you communicate with Patrick and Jeanie in Marketing to make sure they are OK with courses that ended to be `marketable` for email marketing?

MatthewPiatetsky: thanks, I didn't realize that sailthru was using this

course_discovery/apps/course_metadata/query.py
amangano-edx: A course_run is considered enrollable if:
- enrollment_start and enrollment_end are null
- enrollment_start is null and enrollment_end is in the future
- enrollment_start is in the past and enrollment_end is null
- enrollment_start is in the past and enrollment_end is in the future

MatthewPiatetsky: cool, jasper said we don't need courses with enrollment_start in future to show up, so i'll update enrollment to mean this

MatthewPiatetsky: updated with this definition

course_discovery/apps/api/serializers.py
MatthewPiatetsky: I am no longer modifying this query parameter, I will modify the marketing site to pass marketable=1&enrollable=1

schenedx:
@MatthewPiatetsky Could you please write a test case in the api view test to ensure the flag `is_enrollable` is returning with correct course_runs?

course_discovery/apps/course_metadata/query.py
```diff
@@ -43,6 +43,27 @@ def active(self):
...
+        Returns:
+            QuerySet
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        return self.filter(
+            (
+                Q(enrollment_end__gt=now) |
+                Q(enrollment_end__isnull=True)
+            ) & (
+                Q(enrollment_start__gt=now) |
```
amangano-edx: Maybe `gte` might be better for this?

course_discovery/apps/course_metadata/query.py
amangano-edx: The code looks good, but could you please also update the comment?

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: true

amangano-edx: sorry, this should actually be `lte`

amangano-edx: enrollment_start should be in the past

MatthewPiatetsky: thanks for helping clarify the lte/gte get confusing

MatthewPiatetsky:
FYI still in process of refactoring tests

course_discovery/apps/course_metadata/query.py
```diff
@@ -43,6 +43,26 @@ def active(self):
             )
         )
 
+    def enrollable(self):
+        """ Returns CourseRuns that meet the following enrollment criteria:
+            - Open for enrollment
+            - OR have no specified enrollment close date (e.g. self-paced courses)
```
amangano-edx: I think that this should also mention that enrollment_start must be unspecified or in the past.

schenedx (approved):
👍 
Can you please add this flag to the swagger list at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L53?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,59 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+            self.course,
+            context={
+                'request': self.request,
+                'marketable_enrollable_course_runs_with_archived': marketable_enrollable_course_runs_with_archived
+            }
+        )
+
+        self.assertEqual(
+            len(serializer.data['course_runs']),
+            2 if marketable_enrollable_course_runs_with_archived else 4
```
davec-edx: `assert len(serializer.data['course_runs']) == 2 if ... else 4`

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,6 +109,43 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        )
+
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
+        url += '?marketable_enrollable_course_runs_with_archived={}'.format(
+            marketable_enrollable_course_runs_with_archived
+        )
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
```
davec-edx: more self.asserts that should be changed

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,23 @@ def test_active(self):
...
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_end=None)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
+
+        self.assertEqual(
```
davec-edx: more self.asserts

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,23 @@ def test_active(self):
...
+    def test_enrollable(self):
+        """ Verify the method returns only course runs currently open for enrollment or opening in the future. """
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_end=None)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
```
davec-edx: Is there a need to explicitly add a start=None?  I see end=None but not start...

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/tests/test_views/test_courses.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/test_query.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/test_query.py
MatthewPiatetsky: added, there are 9 combinations for just these two fields, it wasn't really necessary to test all, but added this one

course_discovery/apps/api/serializers.py
```diff
@@ -514,6 +514,10 @@ def get_course_runs(self, course):
             # closed - achieving this requires applying both the marketable and active filters.
             course_runs = course_runs.marketable().active()
 
+        if self.context.get('marketable_enrollable_course_runs_with_archived'):
+            # Same as the querystring above, but does not filter out archived courses with an end date in the past
```
clintonb: Same as the "marketable_course_runs_only", but includes courses that have ended.

Don't assume this block of code will always follow the one above.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -224,17 +224,17 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
         no_seats_course_run = CourseRunFactory(status=CourseRunStatus.Published)
 
         # Published course run with a seat and an end date in the past.
-        closed_course_run = CourseRunFactory(
```
clintonb: Revert. "Archived" is more of a marketing term. As I've dug into the course lifecycle, it's clear that while an archived course run is a closed run (end date passed), a closed run is not necessarily archived (open for enrollment/review after end date).

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
         )
 
     @ddt.data(0, 1)
+    def test_marketable_enrollable_course_runs_with_archived(self, marketable_enrollable_course_runs_with_archived):
+        """
+        Verify that the test_marketable_enrollable_course_runs_with_archived option is respected, restricting returned
```
clintonb: s/test_marketable_enrollable_course_runs_with_archived/marketable_enrollable_course_runs_with_archived

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        course runs to those that are published, have seats, and can still be enrolled in.
+        Also includes archived courses with an end date in the past that the test_marketable_course_runs_only
```
clintonb: I don't know what this means.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        flag filters out.
+        """
+        # Published course run with a seat, no enrollment start or end, and an end date in the future.
```
clintonb: This isn't telling me anything that I can't see below. This line can go away.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        enrollable_course_run = CourseRunFactory(
+            status=CourseRunStatus.Published,
+            end=datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=10),
+            enrollment_start=None,
+            enrollment_end=None,
+        )
+        SeatFactory(course_run=enrollable_course_run)
+
+        # Unpublished course run with a seat.
```
clintonb: Remove.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        unpublished_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        SeatFactory(course_run=unpublished_course_run)
+
+        # Published course run with no seats.
```
clintonb: Remove

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        no_seats_course_run = CourseRunFactory(status=CourseRunStatus.Published)
```
clintonb: Nit: `course_run_without_seats`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+
+        # Published course run with a seat and an end date in the past.
```
clintonb: Remove.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        archived_course_run = CourseRunFactory(
```
clintonb: `closed_course_run`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -248,6 +248,57 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        serializer = self.serializer_class(
+            self.course,
+            context={
+                'request': self.request,
+                'marketable_enrollable_course_runs_with_archived': marketable_enrollable_course_runs_with_archived
+            }
+        )
+
+        assert len(serializer.data['course_runs']) == \
+            2 if marketable_enrollable_course_runs_with_archived else 4
```
clintonb: No if statements in tests. That's a sign that you need multiple tests. You don't need to test the case where the flag is not active. We already have tests for that. You do, however, need to ensure the correct data is returned, not just the number of course runs.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,6 +109,40 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
         )
 
     @ddt.data(1, 0)
+    def test_marketable_enrollable_course_runs_with_archived(self, marketable_enrollable_course_runs_with_archived):
+        """ Verify the endpoint filters course runs to those that are marketable and
+        enrollable, including archived courses (with an end date in the past). """
```
clintonb: s/courses/course_runs

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,6 +109,40 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_end=None)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
+
+        self.course.course_runs.add(
```
clintonb: Pass the course to the factories: `CourseRunFactory(course=self.course)`. Avoid creating other courses unnecessarily. Please correct the other tests in this file that are also creating course runs incorrectly.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -42,6 +42,9 @@ def get_serializer_context(self, *args, **kwargs):
             'exclude_utm': get_query_param(self.request, 'exclude_utm'),
```
clintonb: It's time to clean this up. There is a lost of duplication that can be avoided with a simple for-loop.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -71,7 +74,14 @@ def list(self, request, *args, **kwargs):
               multiple: false
             - name: marketable_course_runs_only
               description: Restrict returned course runs to those that are published, have seats,
-                and can still be enrolled in.
+                and can be enrolled in, now or in the future.
```
clintonb: If the course run can be enrolled in, by definition that is now or in the future. This change doesn't add much value and can be reverted.

course_discovery/apps/course_metadata/query.py
```diff
@@ -43,6 +43,26 @@ def active(self):
             )
         )
 
+    def enrollable(self):
+        """ Returns instructor paced course runs that are open for enrollment or self-paced course runs
```
clintonb: This has nothing to do with pacing.

clintonb (changes requested):

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: but based on the definitions in this ticket this course run is archived and not closed. https://openedx.atlassian.net/browse/ECOM-6819
I don't care what this course run is, but we should agree on a team on what these terms actually mean

clintonb: > ...but we should agree on a team on what these terms actually mean

This is why I don't want to touch any of this until we've defined the course lifecycle!

course_discovery/apps/api/v1/views/courses.py
MatthewPiatetsky: The enrollable filter only looks for enrollable course runs that are enrollable today, so the way we have defined enrollable there does not include courses which you can enroll in in the future. The existing active filter looks also includes course runs with an enrollment_start date in the future. They are not enrollable. If this is unclear I could change it to something like "and are enrollable or will be enrollable in the future"

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: ok will remove the docstring

course_discovery/apps/api/v1/views/courses.py
clintonb: I like your change. 👍

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,23 @@ def test_active(self):
 
         self.assertEqual(set(CourseRun.objects.active()), {active_enrollment_end, active_no_enrollment_end})
 
+    def test_enrollable(self):
+        """ Verify the method returns only course runs currently open for enrollment or opening in the future. """
```
amangano-edx: The `enrollable` method in query.py filters out CourseRuns for which enrollment is opening in the future.

MatthewPiatetsky: thanks, fixed

course_discovery/apps/api/v1/tests/test_views/test_courses.py
MatthewPiatetsky: Could you clarify why this approach is incorrect, I don't see the difference, or is it just that we use the other approach more often?

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: i don't mind changing this back, but philosophically I thought that us writing the definitions on the ticket and the grooming and planning it as a team meant that we agreed on the definitions as a team. Rather than pushing off defining some of these things into the future, why not make us of things we've already defined now?

course_discovery/apps/api/v1/tests/test_views/test_courses.py
clintonb: Calling `CourseRunFactory()` will create a `Course` if none is supplied. That's a bunch of wasted DB calls to create data that is not needed. Just pass the course and the factory will make the correct association at the time the `CourseRun` is created.

course_discovery/apps/api/tests/test_serializers.py
clintonb: We have not reached any agreement on these definitions or their implementation.

mikedikan: I would agree @MatthewPiatetsky - I thought this horse was already dead.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
MatthewPiatetsky: got it, thanks!

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -247,6 +238,47 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+            'request': self.request,
+            'marketable_enrollable_course_runs_with_archived': 1
+        }
+
+        course_serializer = self.serializer_class(
+            self.course,
+            context=context
+        )
+
+        assert len(course_serializer.data['course_runs']) == 2
```
clintonb: Less counting, more validating the actual data returned.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
clintonb: This still needs to be addressed.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,6 +109,40 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+            enrollable_no_enrollment_end,
+            enrollable,
+            enrollable_enrollment_start_in_past,
+        )
+
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
+        url += '?marketable_enrollable_course_runs_with_archived={}'.format(
```
clintonb: Use `.format()` for everything here instead of also mixing in concatenation.

course_discovery/apps/api/v1/views/courses.py
clintonb: This still needs to be addressed.

course_discovery/apps/course_metadata/query.py
```diff
@@ -43,6 +43,26 @@ def active(self):
             )
         )
 
+    def enrollable(self):
+        """ Returns course runs that can be enrolled in now.
```
clintonb: Returns course runs that are currently open for enrollment.

A course run is considered open for enrollment if its enrollment start date has passed or is `None`, AND its enrollment end date is in the future or is `None`.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,23 @@ def test_active(self):
...
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_end=None)
+        enrollable_no_enrollment_start = CourseRunFactory(enrollment_start=None, enrollment_end=None)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
+
+        assert set(CourseRun.objects.enrollable()) == {
```
clintonb: I am leery of casting to a set on the off-chance that duplicates are returned. I believe the ORM returns items in order of their ID values, if no `order_by` value is set on the model. Given that we don't set an `order_by`, you should be safe doing the following:

```python
assert list(CourseRun.objects.enrollable()) == [enrollable, enrollable_no_enrollment_end, enrollable_no_enrollment_start, enrollable_enrollment_start_in_past]
```

clintonb (changes requested):

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: I am validating in the statement below that the correct course runs below by the course run key. Do I want to verify all the fields even though we're already testing the serializer? The only issue I see with that is since we're using a different serializer to construct the get_expected_data I would need to add this querystring to the other serializer purely for test purposes. It seems bad to do that just for the tests, unless there's a better way to do it

course_discovery/apps/api/v1/views/courses.py
MatthewPiatetsky: ok will change, i thought you meant you liked the original lol

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -247,6 +238,47 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        course_run_keys = [course_run['key'] for course_run in course_serializer.data['course_runs']]
+        self.assertCountEqual(course_run_keys, [enrollable_course_run.key, archived_course_run.key])
```
amangano-edx: This seems to be testing that we got two course runs back, not that they are the correct two course runs.

MatthewPiatetsky: that's not what assertCountEqual does https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.assertCountEqual

however, since we are moving away from unittest i suppose I could change this to sorted()

amangano-edx: Also, I think we should be consistent with our approach to assertions. On line 278, we use a pytest style assertion but here we're using a unittest style assertion.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,6 +109,40 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
 
     @ddt.data(1, 0)
+    def test_marketable_enrollable_course_runs_with_archived(self, marketable_enrollable_course_runs_with_archived):
+        """ Verify the endpoint filters course runs to those that are marketable and
+        enrollable, including archived course runs (with an end date in the past). """
+
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_end=future)
```
amangano-edx: If we expect this to always be enrollable, we should set an explicit value for enrollment_start (even if it's just `None`). When it's left unspecified like this, the test may become flaky as the factory may sometimes set a value for enrollment_start that results in the CourseRun being un-enrollable.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,23 @@ def test_active(self):
 
         self.assertEqual(set(CourseRun.objects.active()), {active_enrollment_end, active_no_enrollment_end})
 
+    def test_enrollable(self):
+        """ Verify the method returns only course runs currently open for enrollment. """
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_end=future)
```
amangano-edx: I brought this up in one of the other tests, but I think you should explicitly set enrollment_start in addition to enrollment_end if you expect this course run to always be enrollable. Depending on the factory implementation, it may be possible for it to set enrollment_start in a way that would make the course_run unenrollable, which would result in flaky behavior from this test.

course_discovery/apps/course_metadata/tests/test_query.py
amangano-edx: Would it hurt to just add an order_by to the end of that query to ensure that the results are always returned consistently? Doing so would prevent flaky behavior from this test.

MatthewPiatetsky: I was using sorted, order by is just for queries

clintonb: Actually, we do have `ordering` (not `order_by` as I previously stated) set. We inherit `TimeStampedModel`, which returns items in reverse order of their modified and creation dates.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,23 @@ def test_active(self):
...
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_start=past, enrollment_end=None)
+        enrollable_no_enrollment_start = CourseRunFactory(enrollment_start=None, enrollment_end=future)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
+
+        assert sorted(CourseRun.objects.enrollable()) == sorted([
```
clintonb: These items are already sorted: `ordering = ('-modified', '-created',)`. There is no need to call `sorted()`.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -247,6 +238,47 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        }
+
+        course_serializer = self.serializer_class(
+            self.course,
+            context=context
+        )
+
+        assert len(course_serializer.data['course_runs']) == 2
+        course_run_keys = [course_run['key'] for course_run in course_serializer.data['course_runs']]
+        assert sorted(course_run_keys) == sorted([enrollable_course_run.key, archived_course_run.key])
```
clintonb: No need for `sorted()`.

course_discovery/apps/api/tests/test_serializers.py
clintonb: I don't quite follow you. If the existing methods aren't usable, don't use them. That being said, I mostly concerned that the correct course runs are concerned than the actual data returned for them.

course_discovery/apps/course_metadata/tests/test_query.py
MatthewPiatetsky: I'm using sorted to make it clear in the test that we don't care about the order. previously we'd were using assertCountEqual to signal the intention of the test.  
it seems to me like if we're using an assert statement in a test, we should the assert should only depend on what we're actually testing, even if other factors will coincidentally pass (the order).  
Otherwise everytime we see a list we'll have to figure out if the test cares about order or not

course_discovery/apps/api/tests/test_serializers.py
clintonb: If we are validating the course run keys, there is no need to validate the count. This line should go away.

course_discovery/apps/course_metadata/tests/test_query.py
clintonb: I'd prefer that to be in the form of a comment than assuming another developer will interpret the use of `sorted()` as not caring about order.

MatthewPiatetsky: i think that's why i liked assertCountEqual, but ok

MatthewPiatetsky: actually added sorted back in both places, though the query may have a set order, the expected results do not. I don't want the tests to only pass if we create the variables in a specific order. There were some boosting tests that were incorrectly passing for that reason previously. So rather than comment to people that they have to expect the variables back in the same order they were created, i'd rather use sorted and add a comment (which I did)

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -247,6 +238,48 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
             1 if marketable_course_runs_only else 4
         )
 
+    def test_marketable_enrollable_course_runs_with_archived(self):
+        """
+        Verify that the marketable_enrollable_course_runs_with_archived option is respected, restricting returned
+        course runs to those that are published, have seats, and can still be enrolled in.
+        Also includes courses with an end date in the past that the test_marketable_course_runs_only
```
clintonb: `test_marketable_course_runs_only` is not relevant here.

course_discovery/apps/api/tests/test_serializers.py
clintonb: This line is unnecessary given the assertion below.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,16 +104,45 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
+        CourseRunFactory(enrollment_start=None, enrollment_end=future, course=self.course)
+        CourseRunFactory(enrollment_end=None, course=self.course)
+        CourseRunFactory(
+            enrollment_start=past, enrollment_end=future, course=self.course
+        )
+        CourseRunFactory(enrollment_start=future, course=self.course)
+        CourseRunFactory(enrollment_end=past, course=self.course)
+
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
+        url += '?marketable_enrollable_course_runs_with_archived={}'.format(
```
clintonb: Use `.format()` here.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -38,12 +38,12 @@ def get_queryset(self):
...
-        context.update({
-            'exclude_utm': get_query_param(self.request, 'exclude_utm'),
-            'include_deleted_programs': get_query_param(self.request, 'include_deleted_programs'),
-            'marketable_course_runs_only': get_query_param(self.request, 'marketable_course_runs_only'),
-            'published_course_runs_only': get_query_param(self.request, 'published_course_runs_only'),
-        })
+        query_params = ['exclude_utm', 'include_deleted_programs', 'marketable_course_runs_only',
+                        'marketable_enrollable_course_runs_with_archived', 'published_course_runs_only']
+        for query_param in query_params:
+            context.update({
```
clintonb: This can be more efficient: `context[query_param] = get_query_param(self.request, query_param)`

course_discovery/apps/course_metadata/query.py
```diff
@@ -43,6 +43,26 @@ def active(self):
             )
         )
 
+    def enrollable(self):
+        """ Returns course runs that are currently open for enrollment.
+        A course run is considered open for enrollment if its enrollment start date
```
clintonb: Add a new line above this one.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,24 @@ def test_active(self):
...
+
+        enrollable = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_start=past, enrollment_end=None)
+        enrollable_no_enrollment_start = CourseRunFactory(enrollment_start=None, enrollment_end=future)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        CourseRunFactory(enrollment_start=future)
+        CourseRunFactory(enrollment_end=past)
+
+        # order doesn't matter
+        assert sorted(CourseRun.objects.enrollable()) == sorted([
```
clintonb: Use `.order_by` on `QuerySets`.

clintonb (approved):

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -68,6 +68,24 @@ def test_active(self):
...
 
+    def test_enrollable(self):
+        """ Verify the method returns only course runs currently open for enrollment. """
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        enrollable = CourseRunFactory(enrollment_start=past, enrollment_end=future)
+        enrollable_no_enrollment_end = CourseRunFactory(enrollment_start=past, enrollment_end=None)
+        enrollable_no_enrollment_start = CourseRunFactory(enrollment_start=None, enrollment_end=future)
+        enrollable_enrollment_start_in_past = CourseRunFactory(enrollment_start=past, enrollment_end=future)
```
amangano-edx: this test case is the same as the one on line 76.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -109,16 +104,45 @@ def test_marketable_course_runs_only(self, marketable_course_runs_only):
...
     @ddt.data(1, 0)
+    def test_marketable_enrollable_course_runs_with_archived(self, marketable_enrollable_course_runs_with_archived):
+        """ Verify the endpoint filters course runs to those that are marketable and
+        enrollable, including archived course runs (with an end date in the past). """
+
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        CourseRunFactory(enrollment_start=None, enrollment_end=future, course=self.course)
+        CourseRunFactory(enrollment_end=None, course=self.course)
```
amangano-edx: Should we have set enrollment_start here?

amangano-edx:
LGTM

PR #643, opened 2017-02-08
Added "america" as a synonym of "usa"

clintonb:


clintonb:
FYI @edx/ecommerce 

Approved in person by Jasper

MatthewPiatetsky:
what about central and south america? we don't want someone searching south america and getting results back about the US

clintonb:
From Jasper: we will worry about those if/when it becomes an issue.

PR #644, opened 2017-02-09
Fix console error on course run page

tasawernawaz:
[ECOM-7116](https://openedx.atlassian.net/browse/ECOM-7117)

* Moved JS google diff code to it own file.


tasawernawaz:
@waheedahmed @awais786 Please review.

waheedahmed (approved):

course_discovery/templates/publisher/course_detail.html
```diff
@@ -112,4 +112,5 @@
 {% block extra_js %}
```
awais786, line 112: Can you please add the {% compress js %} tag in base.html ?

awais786 (approved):

course_discovery/templates/publisher/course_detail.html
tasawernawaz: This needs more discovery leave where it is right now.

PR #646, opened 2017-02-10
Create remove_unused_indexes management command

mikedikan:
ECOM-7129

Creating management command to remove older indexes automatically, to clean up dangling elasticsearch indexes.

@edx/ecommerce for review

### DIFF
--- a/course_discovery/apps/core/utils.py
+++ b/course_discovery/apps/core/utils.py
@@ -10,21 +10,14 @@ logger = logging.getLogger(__name__)
 
 class ElasticsearchUtils(object):
     @classmethod
-    def create_alias_and_index(cls, es, alias):
+    def create_alias_and_index(cls, es_connection, alias):
         logger.info('Making sure alias [%s] exists...', alias)
 
-        if es.indices.exists_alias(name=alias):
+        if es_connection.indices.exists_alias(name=alias):
             # If the alias exists, and points to an open index, we are all set.
             logger.info('...alias exists.')
         else:
-            # Create an index with a unique (timestamped) name
-            timestamp = datetime.datetime.utcnow().strftime("%Y%m%d%H%M%S")
-            index = '{alias}_{timestamp}'.format(alias=alias, timestamp=timestamp)
-            index_settings = settings.ELASTICSEARCH_INDEX_SETTINGS
-            index_settings['settings']['analysis']['filter']['synonym']['synonyms'] = get_synonyms(es)
-            es.indices.create(index=index, body=index_settings)
-            logger.info('...index [%s] created.', index)
-
+            index = cls.create_index(es_connection=es_connection, prefix=alias)
             # Point the alias to the new index
             body = {
                 'actions': [
@@ -32,8 +25,29 @@ class ElasticsearchUtils(object):
                     {'add': {'alias': alias, 'index': index}},
                 ]
             }
-            es.indices.update_aliases(body)
+            es_connection.indices.update_aliases(body)
             logger.info('...alias updated.')
+
+    @classmethod
+    def create_index(cls, es_connection, prefix):
+        """
+        Creates a new index whose name is prefixed with the specified value.
+
+        Args:
+            es_connection (Elasticsearch): Elasticsearch connection - the connection object as created in the
+             ElasticsearchSearchBackend class - the 'conn' attribute
+            prefix (str): Alias for the connection, used as prefix for the index name
+
+        Returns:
+            index_name (str): Name of the new index.
+        """
+        timestamp = datetime.datetime.utcnow().strftime('%Y%m%d_%H%M%S')
+        index_name = '{alias}_{timestamp}'.format(alias=prefix, timestamp=timestamp)
+        index_settings = settings.ELASTICSEARCH_INDEX_SETTINGS
+        index_settings['settings']['analysis']['filter']['synonym']['synonyms'] = get_synonyms(es_connection)
+        es_connection.indices.create(index=index_name, body=index_settings)
+        logger.info('...index [%s] created.', index_name)
+        return index_name
 
 
 def get_all_related_field_names(model):
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
@@ -1,0 +1,72 @@
+import logging
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from haystack import connections as haystack_connections
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    backends = []
+    help = 'This command will purge the oldest indexes, freeing up disk space.  This command will never delete the ' \
+           'currently used index.'
+
+    def handle(self, **options):
+        self.backends = options.get('using')
+        if not self.backends:
+            self.backends = list(haystack_connections.connections_info.keys())
+
+        for backend_name in self.backends:
+            connection = haystack_connections[backend_name]
+            backend = connection.get_backend()
+            indices_client = backend.conn.indices
+            current_alias_name = backend.index_name
+            self.remove_unused_indexes(indices_client=indices_client, current_alias_name=current_alias_name)
+
+    def remove_unused_indexes(self, indices_client, current_alias_name):
+        """
+        Removes all but the newest (Elasticsearch) indexes, using the configured value to limit deletions
+
+        Args:
+            indices_client (IndicesClient): Elasticsearch Index API client, used to list/delete index
+            current_alias_name (str): The name of the configured alias, used for lookup
+
+        Returns:
+            None
+        """
+        sorted_indexes_by_timestamp = self.get_indexes_sorted_by_timestamp(indices_client=indices_client,
+                                                                           index_prefix=current_alias_name)
+        index_count = len(sorted_indexes_by_timestamp)
+        logger.info('Found {index_count} indexes'.format(index_count=index_count))
+
+        # Remove current index from list so we don't delete it
+        current_alias = indices_client.get_alias(name=current_alias_name)
+        sorted_indexes_by_timestamp = list(set(sorted_indexes_by_timestamp) - set(current_alias.keys()))
+
+        num_indices_to_remove = len(sorted_indexes_by_timestamp) - settings.HAYSTACK_INDEX_RETENTION_LIMIT
+
+        if num_indices_to_remove > 0:
+            indices_to_remove = sorted_indexes_by_timestamp[:num_indices_to_remove]
+            logger.info('Deleting indices %s...', indices_to_remove)
+            indices_client.delete(index=','.join(indices_to_remove))
+            logger.info('Successfully deleted indices %s.', indices_to_remove)
+        else:
+            logger.info('No indices to remove.')
+
+    def get_indexes_sorted_by_timestamp(self, indices_client, index_prefix):
+        """
+        Uses the haystack connection to fetch the (Elasticsearch) indexes, sorted by timestamp
+
+        Args:
+            indices_client (IndicesClient): Elasticsearch Index API client, used to fetch index info
+            index_prefix (str): The string prefix for the index, used to match the indexes fetched
+
+        Returns:
+            sorted_indexes_by_timestamp (list): The sorted listing of index names
+        """
+        all_index_settings = indices_client.get_settings()
+        all_indexes = list(all_index_settings.keys())
+        all_current_indexes = [index_name for index_name in all_indexes if index_name.startswith(index_prefix + '_')]
+        return sorted(all_current_indexes)
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -1,11 +1,9 @@
-import datetime
 import logging
 
-from django.conf import settings
 from haystack import connections as haystack_connections
 from haystack.management.commands.update_index import Command as HaystackCommand
 
-from course_discovery.settings.process_synonyms import get_synonyms
+from course_discovery.apps.core.utils import ElasticsearchUtils
 
 logger = logging.getLogger(__name__)
 
@@ -70,24 +68,6 @@ class Command(HaystackCommand):
                 index_name(str): Name of the newly-created index.
         """
         alias = backend.index_name
-        index_name = self.create_timestamped_index(backend, alias)
+        index_name = ElasticsearchUtils.create_index(backend.conn, alias)
         backend.index_name = index_name
         return alias, index_name
-
-    def create_timestamped_index(self, backend, prefix):
-        """
-        Creates a new index whose name is prefixed with the specified value.
-
-        Args:
-            backend (ElasticsearchSearchBackend): Backend through which to connect to Elasticsearch.
-            prefix (str): Prefix for the index name
-
-        Returns:
-            index_name (str): Name of the new index.
-        """
-        timestamp = datetime.datetime.utcnow().strftime('%Y%m%d_%H%M%S')
-        index_name = '{alias}_{timestamp}'.format(alias=prefix, timestamp=timestamp)
-        index_settings = settings.ELASTICSEARCH_INDEX_SETTINGS
-        index_settings['settings']['analysis']['filter']['synonym']['synonyms'] = get_synonyms(backend.conn)
-        backend.conn.indices.create(index=index_name, body=index_settings)
-        return index_name
--- a/course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
+++ b/course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
@@ -1,0 +1,60 @@
+import datetime
+
+from django.conf import settings
+from django.core.management import call_command
+from django.test import TestCase
+from freezegun import freeze_time
+
+from course_discovery.apps.core.utils import ElasticsearchUtils
+from course_discovery.apps.edx_haystack_extensions.tests.mixins import SearchIndexTestMixin
+
+
+class RemoveUnusedIndexesTests(SearchIndexTestMixin, TestCase):
+    backend = None
+
+    def test_handle(self):
+        """ Verify the command removes all but the newest indexes. """
+
+        # Create initial index with alias
+        ElasticsearchUtils.create_alias_and_index(es_connection=self.backend.conn, alias=self.backend.index_name)
+        # Use now as initial time, so indexes are created AFTER the current index so expected values are accurate
+        initial_time = datetime.datetime.now()
+        # Create 2 more indexes than we expect to exist after removal
+        for number in range(1, settings.HAYSTACK_INDEX_RETENTION_LIMIT + 2):
+            current_time = initial_time + datetime.timedelta(seconds=number)
+            freezer = freeze_time(current_time)
+            freezer.start()
+            ElasticsearchUtils.create_index(es_connection=self.backend.conn, prefix=self.backend.index_name)
+            freezer.stop()
+
+        # Prune indexes and confirm the right indexes are removed
+        call_command('remove_unused_indexes')
+        current_alias_name = self.backend.index_name
+        indices_client = self.backend.conn.indices
+        current_alias = indices_client.get_alias(name=current_alias_name)
+        indexes_to_keep = current_alias.keys()
+
+        # check that we keep the current indexes, which we don't want removed
+        all_indexes = self.get_current_index_names(indices_client=indices_client, index_prefix=self.backend.index_name)
+        assert set(all_indexes).issuperset(set(indexes_to_keep))
+
+        # check that other indexes are removed, excepting those that don't hit the retention limit
+        expected_count = settings.HAYSTACK_INDEX_RETENTION_LIMIT + len(indexes_to_keep)
+        assert len(all_indexes) == expected_count
+
+        # Attempt to prune indexes again and confirm that no indexes are removed
+        call_command('remove_unused_indexes')
+
+        # check that we keep the current indexes, which we don't want removed
+        all_indexes = self.get_current_index_names(indices_client=indices_client, index_prefix=self.backend.index_name)
+        assert set(all_indexes).issuperset(set(indexes_to_keep))
+
+        # check that index count remains the same as before
+        assert len(all_indexes) == expected_count
+
+    @staticmethod
+    def get_current_index_names(indices_client, index_prefix):
+        all_index_settings = indices_client.get_settings()
+        all_indexes = list(all_index_settings.keys())
+        all_current_indexes = [index_name for index_name in all_indexes if index_name.startswith(index_prefix + '_')]
+        return all_current_indexes
--- a/course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
@@ -4,8 +4,10 @@ from django.test import TestCase
 from elasticsearch import Elasticsearch
 from freezegun import freeze_time
 
+from course_discovery.apps.edx_haystack_extensions.tests.mixins import SearchIndexTestMixin
 
-class UpdateIndexTests(TestCase):
+
+class UpdateIndexTests(SearchIndexTestMixin, TestCase):
     @freeze_time('2016-06-21')
     def test_handle(self):
         """ Verify the command creates a timestamped index and repoints the alias. """
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -431,6 +431,7 @@ HAYSTACK_CONNECTIONS = {
 }
 
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
+HAYSTACK_INDEX_RETENTION_LIMIT = 3
 
 # Elasticsearch search query facet "size" option to increase from the default value of "100"
 # See  https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-facets-terms-facet.html#_accuracy_control
--- a/docs/edx_extensions.rst
+++ b/docs/edx_extensions.rst
@@ -11,3 +11,14 @@ until we can determine what other edX-specific components need to be extracted f
 
 edX developers should add ``'course_discovery.apps.edx_catalog_extensions'`` to the ``INSTALLED_APPS`` setting in a
 ``private.py`` settings file.
+
+Settings
+========
+
+``HAYSTACK_INDEX_RETENTION_LIMIT``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default: ``3``
+
+This field sets an upper bound for the number of indexes that will be retained after
+a purge triggered by the 'remove_unused_indexes' command.  This command will never delete the currently used index.
--- a/course_discovery/apps/edx_haystack_extensions/tests/mixins.py
+++ b/course_discovery/apps/edx_haystack_extensions/tests/mixins.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from elasticsearch.helpers import bulk
+from haystack import connections as haystack_connections
 from haystack.backends import BaseSearchBackend
 from mock import patch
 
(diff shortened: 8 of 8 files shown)


course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,64 @@
+import datetime
+import logging
+
+from collections import OrderedDict
```
clintonb: This is part of Python and should be grouped above.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,64 @@
...
+
+class Command(HaystackCommand):
+    backends = []
+
+    def handle(self, *items, **options):
+        self.backends = options.get('using')
+        if not self.backends:
+            self.backends = list(haystack_connections.connections_info.keys())
+
+        # Use the creation timestamp to identify oldest index
```
clintonb: This comment doesn't make sense here. Remove it.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,64 @@
...
+    def remove_unused_indexes(self, indices):
+        """
+        Removes all but the newest (Elasticsearch) indexes, using the configured value to limit deletions
+
+        Args:
+            indices (IndicesClient): Elasticsearch Index API client, used to list/delete index
+
+        Returns:
+            None
+        :param indices:
```
clintonb: Remove.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,64 @@
...
+    def handle(self, *items, **options):
+        self.backends = options.get('using')
+        if not self.backends:
+            self.backends = list(haystack_connections.connections_info.keys())
+
+        # Use the creation timestamp to identify oldest index
+        for backend_name in self.backends:
+            connection = haystack_connections[backend_name]
+            backend = connection.get_backend()
+            indices = backend.conn.indices
```
clintonb: Use `indicies_client` as the variable name to avoid confusion.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,64 @@
...
+
+            self.remove_unused_indexes(indices)
+
+    def remove_unused_indexes(self, indices):
```
clintonb: `indicies_client`

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Uses the haystack connection to fetch the (Elasticsearch) indexes, sorted by timestamp
+
+        Args:
+            indices (IndicesClient): Elasticsearch Index API client, used to fetch index info
+
+        Returns:
+            sorted_indexes_by_timestamp (OrderedDict): The sorted key:value pairing of timestamp:index_name
+        """
+        all_index_settings = indices.get_settings()
```
clintonb: All you need to do here is return a sorted list of index names. If you simply sort them alphabetically, you should be all set. No need to look at the settings or return a `dict`.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,63 @@
...
+        Returns:
+            None
+        :param indices:
+        """
+        index_index = 0  # ;)
+        sorted_indexes_by_timestamp = self.get_indexes_sorted_by_timestamp(indices)
+        index_count = len(sorted_indexes_by_timestamp)
+        for _, index_name in sorted_indexes_by_timestamp.items():
+            if index_index < index_count - settings.HAYSTACK_INDEX_RETENTION_LIMIT:
+                indices.delete(index=index_name)
```
clintonb: Create a comma-separated list and delete the indexes in bulk. See http://elasticsearch-py.readthedocs.io/en/1.9.0/api.html#elasticsearch.client.IndicesClient.delete. You shouldn't need a for-loop. It should be possible to use list slicing to pick which indexes need to be removed. Remember to *not* delete anything if there are `settings.HAYSTACK_INDEX_RETENTION_LIMIT`, or fewer, indexes.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,41 @@
...
+        connection = Elasticsearch(host)
+        response = connection.indices.get_settings()
+
+        # Prune indexes and confirm the count is correct
+        call_command('remove_unused_indexes')
+        response = connection.indices.get_settings()
+        expected_count = settings.HAYSTACK_INDEX_RETENTION_LIMIT
+        self.assertEqual(len(response), expected_count)
+
+    def reset_sessions(self):
```
clintonb: What are the errors?

course_discovery/settings/base.py
```diff
@@ -432,6 +432,7 @@
 }
 
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
+HAYSTACK_INDEX_RETENTION_LIMIT = 3
```
clintonb, line 434: Add documentation to https://github.com/edx/course-discovery/blob/master/docs/edx_extensions.rst.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def remove_unused_indexes(self, indices):
+        """
+        Removes all but the newest (Elasticsearch) indexes, using the configured value to limit deletions
+
+        Args:
+            indices (IndicesClient): Elasticsearch Index API client, used to list/delete index
+
+        Returns:
+            None
```
clintonb: No need to document this.

clintonb (changes requested):

course_discovery/settings/base.py
mikedikan: I don't see an obvious place to add documentation on this, did you have an idea of what to say so we avoid back and forth through PR comments?

clintonb: Add documentation at https://github.com/edx/course-discovery/blob/master/docs/edx_extensions.rst.

mikedikan: Sorry, what I mean is, i see the file - what exactly do you want me to say?  "Add documentation" is open ended and difficult to interpret.  The documentation you link has no natural spot for me to mention this variable, and the settings file has several other haystack variables. Are you asking me to describe this one only?  All of them? Some of them? Are you asking me to add a section in this file detailing the management commands and mention that one of them has a settings parameter?    What level of detail are you looking for?  Is there an example I can follow that shows me what your expectations are?

 I'm looking for guidance up front to avoid going through multiple rounds of revision.

clintonb: Add a new section, call it "Settings". Document the new setting. Since it only applies to this custom app, it seems logical to group it there.

docs/edx_extensions.rst
```diff
@@ -11,3 +11,8 @@ until we can determine what other edX-specific components need to be extracted f
 
 edX developers should add ``'course_discovery.apps.edx_catalog_extensions'`` to the ``INSTALLED_APPS`` setting in a
 ``private.py`` settings file.
+
+Settings
+========
+HAYSTACK_INDEX_RETENTION_LIMIT: This field sets an upper bound for the number of indexes that will be retained after
```
clintonb: Provide a bit better formatting:
* https://raw.githubusercontent.com/edx/edx-drf-extensions/master/docs/settings.rst
* http://edx-drf-extensions.readthedocs.io/en/latest/settings.html

docs/edx_extensions.rst
```diff
@@ -11,3 +11,8 @@ until we can determine what other edX-specific components need to be extracted f
...
+a purge triggered by the 'remove_unused_indexes' command.  This command will never delete the currently used index.
```
clintonb: The comment about the command not deleting current indexes belongs in the help text/docs for that command. It seems out of place here.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,68 @@
...
+from haystack import connections as haystack_connections
+from haystack.management.commands.update_index import Command as HaystackCommand
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(HaystackCommand):
+    backends = []
+
```
clintonb: Add `help`. See https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/#django.core.management.BaseCommand.help.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,68 @@
...
+        Args:
+            indices_client (IndicesClient): Elasticsearch Index API client, used to fetch index info
+
+        Returns:
+            sorted_indexes_by_timestamp (list): The sorted listing of index names
+        """
+        all_index_settings = indices_client.get_settings()
+        sorted_indexes_by_timestamp = sorted(list(all_index_settings.keys()))
+        return sorted_indexes_by_timestamp
```
clintonb: Merge with the line above.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,68 @@
...
+        Returns:
+            None
+        """
+        sorted_indexes_by_timestamp = self.get_indexes_sorted_by_timestamp(indices_client)
+        index_count = len(sorted_indexes_by_timestamp)
+        logger.debug('Found {index_count} indexes'.format(index_count=index_count))
+
+        # Remove current index from list so we don't delete it
+        current_alias = indices_client.get_alias(name=current_alias)
+        # We only have one index for the alias (so we can cheat and refer to the 0th index directly)
```
clintonb: Why? Just remove all them. There will almost always be one; but, we shouldn't make an assumption here.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,68 @@
...
+        Args:
+            indices_client (IndicesClient): Elasticsearch Index API client, used to list/delete index
+            current_alias (str): The name of the configured alias, used for lookup
+
+        Returns:
+            None
+        """
+        sorted_indexes_by_timestamp = self.get_indexes_sorted_by_timestamp(indices_client)
+        index_count = len(sorted_indexes_by_timestamp)
+        logger.debug('Found {index_count} indexes'.format(index_count=index_count))
```
clintonb: Debug-level log entries don't go to Splunk, making them useless outside of local development. Use `.info()`.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Remove current index from list so we don't delete it
+        current_alias = indices_client.get_alias(name=current_alias)
+        # We only have one index for the alias (so we can cheat and refer to the 0th index directly)
+        current_index_name = list(current_alias.keys())[0]
+        sorted_indexes_by_timestamp.remove(current_index_name)
+        logger.debug('Removing current index: {current_index} from cull list.'.format(current_index=current_index_name))
+        index_count = len(sorted_indexes_by_timestamp)
+
+        index_index = 0  # ;)
```
clintonb: Save the time/resources and just delete in bulk.

```python
num_indices_to_remove = len(sorted_indexes_by_timestamp) - settings.HAYSTACK_INDEX_RETENTION_LIMIT

if num_indices_to_remove > 0:
  indices_to_remove = sorted_indexes_by_timestamp[:num_indices_to_remove]
  logger.info('Deleting indices %s...', indices_to_remove)
  indices_client.delete(index=','.join(indices_to_remove))
  logger.info('Successfully deleted indices %s.', indices_to_remove)
else:
  logger.info('No indicies to remove.')
```

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,47 @@
...
+
+from course_discovery.apps.edx_haystack_extensions.management.commands.update_index import Command
+
+
+class RemoveUnusedIndexesTests(TestCase):
+    def test_handle(self):
+        """ Verify the command removes all but the newest indexes. """
+
+        # Create 2 more indexes than we expect to exist after removal
+        update_index_command = Command()
```
clintonb: There should not be a reliance on another command to test this command. Move shared code to [`ElasticsearchUtils`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/core/utils.py#L11) (and add the appropriate tests).

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,47 @@
...
+            current_time = initial_time + datetime.timedelta(seconds=number)
+            freezer = freeze_time(current_time)
+            freezer.start()
+            update_index_command.create_timestamped_index(backend=backend, prefix=backend.index_name)
+            freezer.stop()
+
+        # Get listing of indexes so we can count them
+        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
+        connection = Elasticsearch(host)
+        response = connection.indices.get_settings()
```
clintonb: This is never used.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Prune indexes and confirm the count is correct
+        call_command('remove_unused_indexes')
+        response = connection.indices.get_settings()
+        # +1 for the current index, which we don't want removed
+        expected_count = settings.HAYSTACK_INDEX_RETENTION_LIMIT + 1
+        self.assertEqual(len(response), expected_count)
```
clintonb: The purpose of these tests should be to ensure that the *correct* indexes are removed, not just a certain number. You want to ensure the currently aliased index is not removed. If the number of indexes on the host is less than the number of indexes to remove, you want to ensure no indexes are removed.

Also, you should be using `assert` now that we are using `pytest`.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,47 @@
...
+        for number in range(1, settings.HAYSTACK_INDEX_RETENTION_LIMIT + 2):
+            current_time = initial_time + datetime.timedelta(seconds=number)
+            freezer = freeze_time(current_time)
+            freezer.start()
+            update_index_command.create_timestamped_index(backend=backend, prefix=backend.index_name)
+            freezer.stop()
+
+        # Get listing of indexes so we can count them
+        host = settings.HAYSTACK_CONNECTIONS['default']['URL']
+        connection = Elasticsearch(host)
```
clintonb: You already have a backend. Use `backend.conn`.

clintonb (changes requested):

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
mikedikan: This seemed like it wasn't appropriate for splunk.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
mikedikan: What?

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
clintonb: If it's not going to Splunk, the statement is not that useful.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
mikedikan: I can add tests for those two scenarios you describe, but the count still seems like the appropriate way to do it.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
clintonb: We can easily account for the case where an alias might reference two distinct indices. Just do that instead of assuming the alias will only point to one index.

`sorted_indexes_by_timestamp = set(sorted_indexes_by_timestamp) - set(current_alias.keys())`

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
clintonb: The count doesn't assert that the indexes we expect to be removed are removed or that the live index remains untouched.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,47 @@
...
+        # Create 2 more indexes than we expect to exist after removal
+        update_index_command = Command()
+        backend = self.get_backend()
+        # Use now as initial time, so indexes are created AFTER the current index so expected values are accurate
+        initial_time = datetime.datetime.now()
+        for number in range(1, settings.HAYSTACK_INDEX_RETENTION_LIMIT + 2):
+            current_time = initial_time + datetime.timedelta(seconds=number)
+            freezer = freeze_time(current_time)
+            freezer.start()
+            update_index_command.create_timestamped_index(backend=backend, prefix=backend.index_name)
```
clintonb: Make sure you delete any additional indices created by these tests. `ElasticsearchTestMixin.delete_index()` has an example method. Ideally, the cleanup should happen in either the `tearDown()` or `tearDownClass()` method. You should be able to use a wildcard call to delete all indices matching `catalog_test_*`.

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
mikedikan: I disagree, i found it helpful during development, and think other developers might as well.

rock345:
@clintonb can you call out nits?

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
rlucioni: @mikedikan I can tell you from experience that this and other debug messages here will be useful to have in Splunk when something inevitably goes wrong. For example, take a look at how I did the logging for the LMS' program certificate [backpopulation command](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/programs/management/commands/backpopulate_program_credentials.py#L40).

Keep in mind that the people running these commands are often not intimately familiar with the code, and these messages can help to quickly diagnose problems.

rlucioni: Another way to say it: if you found it helpful during development, we're going to find it helpful in a production setting when we need to figure out what the command did. Detailed logging has proven to be useful. Please use `.info()`.

clintonb:
@rock345 the only item that might be considered a nit-pick is the use of `logging.debug()`. Other than that every other comment needs to be addressed before I approve this pull request.

mikedikan:
@clintonb This is ready for another pass.  Note there were several changes here due to the refactoring to use the shared index create util.  I suspect I can further reduce the code duplication, but I am concerned that this call might be doing something important:

https://github.com/edx/course-discovery/blob/master/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py#L30

If that is not important, the entire update_index command is redundant code that can be eliminated, see https://github.com/edx/course-discovery/blob/master/course_discovery/apps/core/management/commands/install_es_indexes.py

course_discovery/apps/core/utils.py
```diff
@@ -10,31 +10,44 @@
...
+            es_connection.indices.update_aliases(body)
             logger.info('...alias updated.')
 
+    @classmethod
+    def create_index(cls, es_connection, alias):
+        """
+        Creates a new index whose name is prefixed with the specified value.
+
+        Args:
+            es_connection (Elasticsearch): Elasticsearch connection - (ElasticsearchSearchBackend.conn)
```
clintonb: Nit: the parenthetical is confusing and should be removed.

course_discovery/apps/core/utils.py
```diff
@@ -10,31 +10,44 @@
...
                     {'remove': {'alias': alias, 'index': '*'}},
                     {'add': {'alias': alias, 'index': index}},
                 ]
             }
-            es.indices.update_aliases(body)
+            es_connection.indices.update_aliases(body)
             logger.info('...alias updated.')
 
+    @classmethod
+    def create_index(cls, es_connection, alias):
```
clintonb: s/alias/prefix

course_discovery/apps/core/utils.py
```diff
@@ -10,31 +10,44 @@
...
+        Creates a new index whose name is prefixed with the specified value.
+
+        Args:
+            es_connection (Elasticsearch): Elasticsearch connection - (ElasticsearchSearchBackend.conn)
+            alias (str): Alias for the connection, used as prefix for the index name
+
+        Returns:
+            index_name (str): Name of the new index.
+        """
+        timestamp = datetime.datetime.utcnow().strftime("%Y%m%d%H%M%S")
```
clintonb: Nit: single quotes

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -0,0 +1,73 @@
...
+            None
+        """
+        sorted_indexes_by_timestamp = self.get_indexes_sorted_by_timestamp(indices_client=indices_client,
+                                                                           index_prefix=current_alias_name)
+        index_count = len(sorted_indexes_by_timestamp)
+        logger.info('Found {index_count} indexes'.format(index_count=index_count))
+
+        # Remove current index from list so we don't delete it
+        current_alias = indices_client.get_alias(name=current_alias_name)
+        # We only have one index for the alias (so we can cheat and refer to the 0th index directly)
```
clintonb: This comment makes no sense with the current code.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,78 @@
...
+        call_command('remove_unused_indexes')
+        current_alias_name = self.backend.index_name
+        indices_client = self.backend.conn.indices
+        current_alias = indices_client.get_alias(name=current_alias_name)
+        indexes_to_keep = current_alias.keys()
+
+        # check that we keep the current indexes, which we don't want removed
+        all_indexes = self.get_current_index_names(indices_client=indices_client, index_prefix=self.backend.index_name)
+        for index_to_keep in indexes_to_keep:
+            assert (index_to_keep in all_indexes) is True
```
clintonb: No need for the parentheses or `is True`

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_current_index_names(indices_client, index_prefix):
+        all_index_settings = indices_client.get_settings()
+        all_indexes = list(all_index_settings.keys())
+        all_current_indexes = [index_name for index_name in all_indexes if index_name.startswith(index_prefix + '_')]
+        return all_current_indexes
+
+    def tearDown(self):
+        """ Remove the indexes we created """
+
+        self.backend.conn.indices.delete(index='catalog_test_*')
```
clintonb: This prefix should come from the Haystack settings: https://github.com/edx/course-discovery/blob/master/course_discovery/settings/shared/test.py#L7

`self.backend.index_name + '_*'`

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,78 @@
...
+        super().tearDown()
+
+    @staticmethod
+    def get_backend():
+        """ Get the backend from the configured connection """
+        backends = list(haystack_connections.connections_info.keys())
```
clintonb: The odds of using a connection other than `'default'` *for tests* are quite low. You can simplify this by simply calling `self.backend = haystack_connections['default'].get_backend()` in `setUp()`, and removing this method.

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
```diff
@@ -25,3 +30,17 @@ def test_handle(self):
             }
         }
         self.assertDictEqual(response, expected)
+
+    def tearDown(self):
+        """ Remove the indexes we created """
+
+        self.backend.conn.indices.delete(index='catalog_test_*')
```
clintonb: Move this setup/teardown behavior to a mixin to avoid duplication.

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
```diff
@@ -3,16 +3,21 @@
...
+        self.backend = self.get_backend()
+
     @freeze_time('2016-06-21')
     def test_handle(self):
         """ Verify the command creates a timestamped index and repoints the alias. """
         call_command('update_index')
 
         alias = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
-        index = '{alias}_20160621_000000'.format(alias=alias)
+        index = '{alias}_20160621000000'.format(alias=alias)
```
clintonb: The index naming scheme should not be changing.

course_discovery/apps/core/utils.py
clintonb: Keep the underscore between the date and time so the index names are easier for humans to parse: `'%Y%m%d_%H%M%S'`.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
clintonb: Also, this can be simplified: `assert set(all_indexes).issuperset(set(indexes_to_keep))`

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
```diff
@@ -0,0 +1,78 @@
...
+            assert (index_to_keep in all_indexes) is True
+
+        # check that other indexes are removed, excepting those that don't hit the retention limit
+        expected_count = settings.HAYSTACK_INDEX_RETENTION_LIMIT + len(indexes_to_keep)
+        assert len(all_indexes) == expected_count
+
+        # Attempt to prune indexes again and confirm that no indexes are removed
+        call_command('remove_unused_indexes')
+
+        # check that we keep the current indexes, which we don't want removed
```
clintonb, line 48: Same as above: `assert set(all_indexes).issuperset(set(indexes_to_keep))`.

clintonb (approved):
The call to `super(Command, self).handle(*items, **options)` in `update_index.py` is quite important—it does the actual indexing. The rest of the file is simply ensuring we are operating on an index with a timestamped name.

course_discovery/apps/edx_haystack_extensions/tests/test_remove_unused_indexes.py
mikedikan: How is that a simplification?

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -70,24 +68,7 @@ def prepare_backend_index(self, backend):
                 index_name(str): Name of the newly-created index.
         """
         alias = backend.index_name
-        index_name = self.create_timestamped_index(backend, alias)
+        # index_name = self.create_timestamped_index(backend, alias)
```
clintonb: This should be removed.

course_discovery/apps/core/utils.py
```diff
@@ -10,31 +10,45 @@
...
+
+        Args:
+            es_connection (Elasticsearch): Elasticsearch connection - the connection object as created in the
+             ElasticsearchSearchBackend class - the 'conn' attribute
+            prefix (str): Alias for the connection, used as prefix for the index name
+
+        Returns:
+            index_name (str): Name of the new index.
+        """
+        timestamp = datetime.datetime.utcnow().strftime("%Y%m%d_%H%M%S")
```
clintonb: Single quotes.

clintonb (approved):

PR #650, opened 2017-02-10
WL-766 Add ability to filter Program list by a comma-separated list of ProgramType names

douglashall:
WL-766

We will use this in the LMS marketing pages to filter the list of Programs by the ProgramTypes that should be display on a given site.

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -156,6 +156,7 @@ class CourseRunFilter(FilterSetMixin, django_filters.FilterSet):
 class ProgramFilter(FilterSetMixin, django_filters.FilterSet):
     marketable = django_filters.MethodFilter()
     type = django_filters.CharFilter(name='type__name', lookup_expr='iexact')
+    types = CharListFilter(name='type__slug', lookup_expr='in')
     uuids = UUIDListFilter()
 
     class Meta:
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -127,6 +127,18 @@ class ProgramViewSetTests(SerializationMixin, APITestCase):
         url = self.list_path + '?type=bar'
         self.assert_list_results(url, [], 4)
 
+    def test_filter_by_types(self):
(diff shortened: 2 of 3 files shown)


douglashall:
@edx/ecommerce Please review.

course_discovery/apps/api/filters.py
```diff
@@ -156,6 +156,7 @@ class Meta:
 class ProgramFilter(FilterSetMixin, django_filters.FilterSet):
     marketable = django_filters.MethodFilter()
     type = django_filters.CharFilter(name='type__name', lookup_expr='iexact')
+    types = CharListFilter(name='type__name', lookup_expr='in')
```
clintonb: You should filter by the slug field. The name field can have spaces and other characters that are not ideal for a query string. No need to modify the existing type field, but the reason for the difference should be documented in a comment.

Also, add this to the Swagger docs by modifying the docstring for `ProgramViewSet.list()`.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -127,6 +127,18 @@ def test_filter_by_type(self):
         url = self.list_path + '?type=bar'
         self.assert_list_results(url, [], 4)
 
+    def test_filter_by_types(self):
+        """ Verify that the endpoint filters programs to those matching the provided ProgramType names. """
+        expected = ProgramFactory.create_batch(2)
+        expected.reverse()
+        type_names = [str(p.type.name) for p in expected]
```
clintonb: You shouldn't need to cast this to a string. It's already a string.

clintonb (changes requested):

course_discovery/apps/api/filters.py
douglashall: Why would this need to be different from the `type` filter above?

clintonb: The previous filter was created before we have program type names with spaces, such as "Professional Certificate".

douglashall: Are you saying I should definitely not change the `type` filter or that it is not required for this PR to merge?

douglashall:
@clintonb Comments addressed. Good to go?

clintonb (approved):

rlucioni (approved):

PR #651, opened 2017-02-10
Added isort for import order validation

clintonb:


clintonb:
@edx/ecommerce please review

course_discovery/apps/api/pagination.py
```diff
@@ -1,7 +1,5 @@
-from rest_framework.pagination import (
-    PageNumberPagination as BasePageNumberPagination,
-    LimitOffsetPagination,
-)
+from rest_framework.pagination import PageNumberPagination as BasePageNumberPagination
+from rest_framework.pagination import LimitOffsetPagination
```
rlucioni, line 2: Can these stay on the same line so we're not importing from the same package twice? I'm surprised isort doesn't like this.

rlucioni (approved):
This is huge! Thanks.

bderusha (approved):
👍

course_discovery/apps/api/pagination.py
clintonb: The only way to allow that is to add an inline comment `# isort:skip`. Doing that however, would increase line length...requiring a comment to ignore that issue.

tl;dr it's not worth the hassle. Leaving as is.

PR #659, opened 2017-02-13
ECOM-6518 add logic to select CourseRun for indexing with Course

amangano-edx:
This PR adds the logic necessary to select the CourseRun to use for indexing a Course. This logic is not currently used anywhere, but will be in a subsequent PR.

JIRA: https://openedx.atlassian.net/browse/ECOM-6815

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -355,7 +355,7 @@ class Course(TimeStampedModel):
         - Purchasability: A user can purchase a paid Seat.
         """
         def sort_key(course_run):
-            """ 
+            """
             Return a numeric score indicating the priority of this CourseRun relative to others. Lower numbers
             indicate higher priority.
             """
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -448,6 +448,61 @@ class CourseRun(TimeStampedModel):
         return deadline
 
     @property
+    def is_publicly_visible(self):
+        """
+        Return True if the CourseRun should be visible to the general public, othwerise return False.
+
+        A CourseRun should be visible to the general public if it is published, has seats, and has not been hidden.
+        """
+        return self.status == CourseRunStatus.Published and not self.hidden and len(self.seats.all()) > 0
+
+    @property
+    def is_enrollable(self):
+        """
+        Return True if the CourseRun is currently enrollable, otherwise return False.
+
+        A CourseRun is enrollable if an unenrolled user may enroll in the course (enrollment has begun and
+        has not ended).
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        enrollment_has_begun = self.enrollment_start is None or self.enrollment_start <= now
+        enrollment_has_not_ended = self.enrollment_end is None or self.enrollment_end > now
+        return enrollment_has_begun and enrollment_has_not_ended
+
+    @property
+    def will_be_enrollable(self):
+        """
+        Return True if the CourseRun is not currently enrollable but will be in the future, otherwise return False.
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        return self.enrollment_start is not None and self.enrollment_start > now
+
+    @property
+    def is_consumable(self):
+        """
+        Return True if the CourseRun is currently consumable, otherwise return False.
+
+        A CourseRun is consumable if a user can consume course content (the course has started).
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        return self.start is None or self.start <= now
+
+    @property
+    def is_or_will_be_purchasable(self):
+        """
+        Return True if the CourseRun is currently purchasable or will be in the future, otherwise return False.
+
+        A CourseRun is purchasable if an unenrolled user may enroll and purchase a paid seat for the CourseRun.
+        """
+        has_enrollable_paid_seats = self.has_enrollable_paid_seats()
+        if not has_enrollable_paid_seats:
+            return False
+
+        now = datetime.datetime.now(pytz.UTC)
+        paid_seat_enrollment_end = self.get_paid_seat_enrollment_end()
+        return paid_seat_enrollment_end is None or paid_seat_enrollment_end > now
+
+    @property
     def program_types(self):
         """
         Exclude unpublished and deleted programs from list
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -382,8 +382,11 @@ class Course(TimeStampedModel):
             else:
                 return 7
 
-        course_runs = sorted(list(self.course_runs.all()), key=sort_key)
-        return course_runs[0] if len(course_runs) > 0 else None
+        course_runs = list(self.course_runs.all())
+        if course_runs:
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -345,6 +345,46 @@ def reindex_course_runs(self):
...
+                return 4
+            elif is_publicly_visible and will_be_enrollable:
+                return 5
+            elif is_publicly_visible:
+                return 6
+            else:
+                return 7
+
+        course_runs = sorted(list(self.course_runs.all()), key=sort_key)
+        return course_runs[0] if len(course_runs) > 0 else None
```
davec-edx: Can be written as:
`return course_runs[0] if course_runs else None`

course_discovery/apps/course_metadata/models.py
```diff
@@ -345,6 +345,46 @@ def reindex_course_runs(self):
...
+            Return a numeric score indicating the priority of this CourseRun relative to others. Lower numbers
+            indicate higher priority.
+            """
+            is_publicly_visible = course_run.is_publicly_visible
+            is_enrollable = course_run.is_enrollable
+            will_be_enrollable = course_run.will_be_enrollable
+            is_consumable = course_run.is_consumable
+            is_or_will_be_purchasable = course_run.is_or_will_be_purchasable
+
+            if is_publicly_visible and is_enrollable and is_consumable and is_or_will_be_purchasable:
```
MatthewPiatetsky, line 368: I think you could make this much more compact by using nested if statements, especially for is_publicly_visible

course_discovery/apps/course_metadata/models.py
```diff
@@ -448,6 +488,61 @@ def get_paid_seat_enrollment_end(self):
...
+
+    @property
+    def is_enrollable(self):
+        """
+        Return True if the CourseRun is currently enrollable, otherwise return False.
+
+        A CourseRun is enrollable if an unenrolled user may enroll in the course (enrollment has begun and
+        has not ended).
+        """
+        now = datetime.datetime.now(pytz.UTC)
```
MatthewPiatetsky, line 510: I had the other PR that added the enrollable filter. I'm wondering if we could use this property there or something

course_discovery/apps/course_metadata/models.py
```diff
@@ -448,6 +488,61 @@ def get_paid_seat_enrollment_end(self):
...
+        enrollment_has_not_ended = self.enrollment_end is None or self.enrollment_end > now
+        return enrollment_has_begun and enrollment_has_not_ended
+
+    @property
+    def will_be_enrollable(self):
+        """
+        Return True if the CourseRun is not currently enrollable but will be in the future, otherwise return False.
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        return self.enrollment_start is not None and self.enrollment_start > now
```
MatthewPiatetsky, line 521: You need to check both start and end to know if something will be enrollable, right?

course_discovery/apps/course_metadata/models.py
```diff
@@ -345,6 +345,46 @@ def reindex_course_runs(self):
...
+            """
+            is_publicly_visible = course_run.is_publicly_visible
+            is_enrollable = course_run.is_enrollable
+            will_be_enrollable = course_run.will_be_enrollable
+            is_consumable = course_run.is_consumable
+            is_or_will_be_purchasable = course_run.is_or_will_be_purchasable
+
+            if is_publicly_visible and is_enrollable and is_consumable and is_or_will_be_purchasable:
+                return 0
+            elif is_publicly_visible and is_enrollable and is_consumable:
```
MatthewPiatetsky, line 370: I would use is_enrollable in a nested if here also

course_discovery/apps/course_metadata/models.py
```diff
@@ -345,6 +345,46 @@ def reindex_course_runs(self):
...
+            elif is_publicly_visible and will_be_enrollable and is_or_will_be_purchasable:
+                return 4
+            elif is_publicly_visible and will_be_enrollable:
+                return 5
+            elif is_publicly_visible:
+                return 6
+            else:
+                return 7
+
+        course_runs = sorted(list(self.course_runs.all()), key=sort_key)
```
MatthewPiatetsky: would it make sense for the length check to be here, seems strange to do the extra operations on an empty list

amangano-edx: If the list is empty I would think there would be nothing for `sorted` to do, rendering it a noop. Is this not the case?

course_discovery/apps/course_metadata/models.py
amangano-edx: Yea, I considered that, but I felt it would made the logic a lot more difficult to understand. The way I've written it, each of the priority states are much more explicit and easier to follow. I think that's more important than keeping the code DRY in this case.

course_discovery/apps/course_metadata/models.py
amangano-edx: Thanks, I'll change to that.

course_discovery/apps/course_metadata/models.py
amangano-edx: Your filter operates at the DB level, whereas this property is expected to be used on records that have already been loaded. Therefor I don't think the two are interchangeable.

course_discovery/apps/course_metadata/models.py
amangano-edx: I suppose if you really want to be thorough, you could also check the end date here. But I'm not sure it's necessary, as we shouldn't have any CourseRuns with enrollment_end dates < the enrollment_start date, and we're already checking that enrollment_start is in the future.

course_discovery/apps/course_metadata/models.py
amangano-edx: I prefer the verbose logic here as it's easier to follow and understand than the nested DRY logic. In this case, since this method is quite complex I think it's more valuable to optimize for readability and clarity than for more compact code.

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: well i was suggesting this not to be dry but because i think nested if statements would be clearer. A clearer mental model to help understand which cases are higher priority than others would be like a binary tree with a higher and lower priority option.    
So with nested ifs we'd say, ok is there a public visible course run? if yes, is there an enrollable course run? It would make sense that the questions we ask first would be higher priority. but i don't care that much so if you disagree this is fine   
I think if we're thinking of it in our head, we will ask one question at a time and go down the mental tree, not 4 states at a time

course_discovery/apps/course_metadata/models.py
amangano-edx: Ohh nvm, I see that you're referring to the if check on line 386. Perhaps it could make sense to move it up here instead.

course_discovery/apps/course_metadata/models.py
amangano-edx: Changed.

course_discovery/apps/course_metadata/models.py
amangano-edx: changed.

PR #661, opened 2017-02-14
Inconsistent datetime formats in Course Catalog API

ahsan-ul-haq:
ECOM-4354

@schenedx please review

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -87,7 +87,7 @@ class CourseRunViewSet(PartnerMixin, viewsets.ModelViewSet):
               paramType: query
               multiple: false
             - name: hidden
-              description: Filter based on wether the course run is hidden from search.
+              description: Filter based on whether the course run is hidden from search.
               required: false
               type: Boolean
               paramType: query
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -334,7 +334,7 @@ REST_FRAMEWORK = {
         'user': '100/hour',
     },
     # NOTE: This is the same format as DRF's iso-8601, however the milliseconds are always rendered.
-    'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%S.%fZ',
+    'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%S.%f%zZ',
 }
 
 # NOTE (CCB): JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -333,6 +333,8 @@ REST_FRAMEWORK = {
     'DEFAULT_THROTTLE_RATES': {
         'user': '100/hour',
     },
+    # NOTE: This is the same format as DRF's iso-8601, however the milliseconds are always rendered.
+    'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%S.%fZ',
 }
 
 # NOTE (CCB): JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.


clintonb (changes requested):
We do NOT want fractional seconds in the API responses. Also, this should be handle via settings: http://www.django-rest-framework.org/api-guide/settings/#date-and-time-formatting

clintonb:
See https://github.com/tomchristie/django-rest-framework/blob/a68b37d8bc432fae37ef5880aec500002b59f565/rest_framework/utils/humanize_datetime.py#L10.

We want to use that format without the optional milliseconds. After discussing with @schenedx, always show milliseconds: `'YYYY-MM-DDThh:mm:ss.uuuuuu[+HH:MM|-HH:MM|Z]'`

ahsan-ul-haq:
@clintonb I think with %f it will always show milliseconds. Do I need to do something else?

clintonb:
Use the setting intead of modifying each serializer field.

ahsan-ul-haq:
@clintonb please review

course_discovery/apps/api/serializers.py
```diff
@@ -309,7 +309,7 @@ class SeatSerializer(serializers.ModelSerializer):
         max_digits=Seat.PRICE_FIELD_CONFIG['max_digits']
     )
     currency = serializers.SlugRelatedField(read_only=True, slug_field='code')
-    upgrade_deadline = serializers.DateTimeField()
+    upgrade_deadline = serializers.DateTimeField(format=settings.API_DATETIME_FORMAT)
```
clintonb: You should not need to change any of these serializers.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -87,7 +87,7 @@ def list(self, request, *args, **kwargs):
               paramType: query
               multiple: false
             - name: hidden
-              description: Filter based on wether the course run is hidden from search.
+              description: Filter based on weather the course run is hidden from search.
```
clintonb: whether

course_discovery/settings/base.py
```diff
@@ -479,3 +479,7 @@
         'debug_toolbar.panels.redirects.RedirectsPanel',
         'elastic_panel.panel.ElasticDebugPanel'
     ]
+
+
+# API Datetime Format
+API_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
```
clintonb: http://www.django-rest-framework.org/api-guide/settings/#date-and-time-formatting

```python
REST_FRAMEWORK = {
  # NOTE: This is the same format as DRF's iso-8601, however the milliseconds are always rendered.
  'DATETIME_FORMAT': 'YYYY-MM-DDThh:mm:ss.uuuuuu[+HH:MM|-HH:MM|Z]',
}
```

clintonb (changes requested):

ahsan-ul-haq:
@clintonb made the change

course_discovery/settings/base.py
```diff
@@ -333,6 +333,8 @@
     'DEFAULT_THROTTLE_RATES': {
         'user': '100/hour',
     },
+    # NOTE: This is the same format as DRF's iso-8601, however the milliseconds are always rendered.
+    'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%S.%fZ',
```
clintonb: Please use this exact string: `'YYYY-MM-DDThh:mm:ss.uuuuuu[+HH:MM|-HH:MM|Z]'` to avoid issues with time zones.

ahsan-ul-haq: @clintonb Doing that causes the dates time to appear like following
screen shot 2017-02-15 at 7 49 14 pm

ahsan-ul-haq:
I have made change for time zone as well

course_discovery/settings/base.py
schenedx: @clintonb I read the DRF documentation again. They require python strftime format string. See http://www.django-rest-framework.org/api-guide/settings/#datetime_format

So `YYYY-MM-DDThh:mm:ss.uuuuuu[+HH:MM|-HH:MM|Z]` is not a python strftime format. It's just a format they use on datetime string to replace `iso-8601`.

schenedx (approved):
👍

schenedx:
@ahsan-ul-haq You got 8 unit test failures, could you please investigate?

PR #666, opened 2017-02-15
Added publisher user creation form in admin.

waheedahmed:
[ECOM-6841](https://openedx.atlassian.net/browse/ECOM-6841)

As a Product Manager, I need a way to add new users to Publisher by using their username so that roles and permissions can be setup ahead of time. (Currently we only add users to Publisher when they login.)

**Implementation Details:**

- We can create our own custom user creation form with custom fields (without password) in django admin.

- Fields: username, groups

screen shot 2017-02-15 at 2 39 24 pm
screen shot 2017-02-15 at 2 39 40 pm


waheedahmed:

tasawernawaz (approved):

awais786 (approved):

course_discovery/apps/publisher/admin.py
```diff
@@ -58,3 +59,17 @@ class SeatAdmin(admin.ModelAdmin):
...
+    add_form_template = 'publisher/admin/add_user_form.html'
+    add_fieldsets = (
+        (None, {'fields': ('username', 'groups',)}),
+    )
+    add_form = PublisherUserCreationForm
+
+    change_password_form = None
+
+    def get_queryset(self, request):
+        return self.model.objects.filter(groups__in=Group.objects.all()).distinct()
```
clintonb: This returns all users in a group. Is that correct? Do you want users in a specific group?

course_discovery/apps/publisher/models.py
```diff
@@ -596,3 +596,10 @@ def change_state(self, state):
             self.published()
 
         self.save()
+
+
+class PublisherUser(User):
```
clintonb, line 601: Does this need a migration?

course_discovery/apps/publisher/admin.py
waheedahmed: @clintonb Yes its correct, we want all user's which belongs to any group.

course_discovery/apps/publisher/models.py
waheedahmed: @clintonb Created, but I can't think why does django needs an empty migration for a proxy model.

waheedahmed: Thanks for notifying!

clintonb: The migration is necessary to create the data associated with the new content type—publisher.PublisherUser. The existing user table will still be used.

PR #669, opened 2017-02-16
Course-run approval widget.

awais786:
ECOM-6076


course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -531,14 +533,31 @@ class ChangeCourseRunStateViewTests(TestCase):
...
         self.client.login(username=self.user.username, password=USER_PASSWORD)
 
+        language_tag = LanguageTag(code='te-st', name='Test Language')
+        language_tag.save()
+        self.course_run.transcript_languages.add(language_tag)
+        self.course_run.language = language_tag
+        self.course_run.save()
+
+        staff = PersonFactory()
+        self.course_run.staff = [staff]
```
waheedahmed: `self.course_run.staff.add(PersonFactory())`

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -86,6 +88,64 @@ def test_studio_url(self):
...
         )
 
+    def test_has_valid_staff(self):
+        """ Verify that property returns True if course-run must have a staff member
+        with bio and image.
+        """
+        self.assertFalse(self.course_run.has_valid_staff)
+        staff = PersonFactory()
+        self.course_run.staff = [staff]
+        self.course_run.save()
```
waheedahmed: No need for above three lines, just do this `self.course_run.staff.add(PersonFactory())`

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -86,6 +88,64 @@ def test_studio_url(self):
...
+        self.assertTrue(self.course_run.has_valid_staff)
+
+    @ddt.data('bio', 'profile_image')
+    def test_with_in_valid_staff(self, field):
+        """ Verify that property returns False staff has bio or image is missing."""
+        staff = PersonFactory(profile_image_url=None)
+        self.course_run.staff = [staff]
+        self.course_run.save()
```
waheedahmed: No need to call save just do this `self.course_run.staff.add(staff)`

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -489,13 +549,40 @@ class CourseRunStateTests(TestCase):
...
+        )
+
+    def setUp(self):
+        super(CourseRunStateTests, self).setUp()
+
+        language_tag = LanguageTag(code='te-st', name='Test Language')
+        language_tag.save()
+        self.course_run.transcript_languages.add(language_tag)
+        self.course_run.language = language_tag
+        self.course_run.is_micromasters = True
```
waheedahmed, line 575: I think you don't need to set value for `is_micromasters` and `micromasters_name` here, if `is_micromasters` is `False`, `is_valid_micromasters` will always return True.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -507,8 +594,54 @@ def test_change_state(self, state):
...
+        self.assertNotEqual(self.course_run_state.name, state)
+        self.course_run_state.change_state(state=state, user=self.user)
+        self.assertEqual(self.course_run_state.name, state)
 
-        self.run_state.change_state(state=state)
+    def test_with_invalid_parent_course_state(self):
+        """
+        Verify that method return False if parent course is not approved.
+        """
+        self.course.course_state.name = CourseStateChoices.Draft
```
waheedahmed: You should assert true at top to check that `can_send_for_review()` is returning `True`. Then make test assert False after changing values.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -507,8 +594,54 @@ def test_change_state(self, state):
...
+        self.course.save()
+        self.assertFalse(self.course_run_state.can_send_for_review())
+
+    def test_can_send_for_review_with_invalid_program_type(self):
+        """
+        Verify that method return False if program type is invalid.
+        """
+        self.course_run.is_micromasters = False
```
waheedahmed: Same here, you should assert true at top to check that `can_send_for_review()` is returning `True`. Then make test assert False after changing values.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -507,8 +594,54 @@ def test_change_state(self, state):
...
+        self.course_run.save()
+        self.assertFalse(self.course_run_state.can_send_for_review())
 
-        self.assertEqual(self.run_state.name, state)
+    def test_can_send_for_review_with_invalid_seat(self):
+        """
+        Verify that method return False if data is missing.
+        """
+        # seat type is verified but its price is 0
+        self.seat.price = 0
```
waheedahmed, line 624: Same here, you should assert true at top to check that `can_send_for_review()` is returning `True`. Then make test assert False after changing values.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -507,8 +594,54 @@ def test_change_state(self, state):
...
+        self.seat.save()
+        self.assertFalse(self.course_run_state.can_send_for_review())
+
+    def test_can_send_for_review_with_no_seat(self):
+        """
+        Verify that method return False if data is missing.
+        """
+        self.course_run.seats.all().first().delete()
```
waheedahmed, line 632: Same here, you should assert true at top to check that `can_send_for_review()` is returning `True`. Then make test assert False after changing values.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -507,8 +594,54 @@ def test_change_state(self, state):
...
+        self.assertFalse(self.course_run_state.can_send_for_review())
+
+    def test_can_send_for_review_without_language(self):
+        """
+        Verify that method return False if data is missing.
+        """
+        self.course_run.language = None
```
waheedahmed, line 639: Same here, you should assert true at top to check that `can_send_for_review()` is returning `True`. Then make test assert False after changing values.

course_discovery/apps/publisher/tests/test_model.py
```diff
@@ -507,8 +594,54 @@ def test_change_state(self, state):
...
+        self.course_run.save()
+        self.assertFalse(self.course_run_state.can_send_for_review())
+
+    def test_can_send_for_review_without_transcript_language(self):
+        """
+        Verify that method return False if data is missing.
+        """
+        self.course_run.transcript_languages = []
```
waheedahmed, line 647: Same here, you should assert true at top to check that `can_send_for_review()` is returning `True`. Then make test assert False after changing values.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1022,6 +1022,116 @@ def test_approval_widget_with_disable_switch(self):
         response = self.client.get(self.page_url)
         self.assertContains(response, '<div id="approval-widget" class="approval-widget hidden">')
 
+    def test_course_approval_widget_disabled_button(self):
```
waheedahmed: No need for this test case, you are already doing this in `test_course_run_approval_widget_for_course_team` at line 1075.

waheedahmed (changes requested):
First pass completed.

course_discovery/apps/publisher/tests/test_model.py
awais786: Done

course_discovery/apps/publisher/tests/test_model.py
awais786: Done

course_discovery/apps/publisher/tests/test_model.py
awais786: Done

course_discovery/apps/publisher/tests/test_model.py
awais786: Done

course_discovery/apps/publisher/models.py
```diff
@@ -332,12 +334,50 @@ def created_by(self):
...
+        if self.lms_course_id and self.course.partner and self.course.partner.studio_url:
             path = 'course/{lms_course_id}'.format(lms_course_id=self.lms_course_id)
             return urljoin(self.course.partner.studio_url, path)
 
         return None
 
+    @property
+    def has_valid_staff(self):
+        """ Check that each staff member has his bio data and image."""
+        if not self.staff.all().exists():
```
waheedahmed: Create a variable `staff_members = self.staff.all()` and use that variable on both places, it will reduce query count. Also you don't need to call `exists()`, you can simply do the following:
```
if not staff_members:
    return False
```

course_discovery/apps/publisher/tests/test_model.py
waheedahmed: You don't need to set this True for every test just move this line into its test case and remove next line.

waheedahmed (changes requested):

tasawernawaz (approved):

waheedahmed (approved):

waheedahmed (approved):
👍 when build passes.

PR #671, opened 2017-02-16
Add program total price to the price_ranges array ECOM-7023

schenedx:
@edx/ecommerce Please review
This is to calculate the total price of a program so we can render the price of the program on the program details page

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -810,11 +810,67 @@ class Program(TimeStampedModel):
     def seat_types(self):
         return set(seat.type for seat in self.seats)
 
+    def _get_total_price_by_currency(self):
+        """
+        This helper function returns the total program price indexed by the currency
+        """
+        currencies_with_total = defaultdict()
+        course_map = defaultdict(list)
+        for seat in self.seats:
+            course_uuid = seat.course_run.course.uuid
+            # Identify the most relevant course_run seat for a course.
+            # And use the price of the seat to represent the price of the course
+            selected_seats = course_map.get(course_uuid)
+            if not selected_seats:
+                # if we do not have this course yet, create the seats array
+                course_map[course_uuid] = [seat]
+            else:
+                add_seat = False
+                seats_to_remove = []
+                for selected_seat in selected_seats:
+                    if seat.currency != selected_seat.currency:
+                        # If the candidate seat has a different currency than the one in the array,
+                        # always add to the array
+                        add_seat = True
+                    elif ((seat.course_run.end is None or
+                           seat.course_run.end >= datetime.datetime.now(pytz.UTC)) and
+                          (seat.course_run.enrollment_start is None or
+                           seat.course_run.enrollment_start > selected_seat.course_run.enrollment_start and
+                           seat.course_run.enrollment_start < datetime.datetime.now(pytz.UTC))):
+                        # If the seat has same currency, the course has not ended,
+                        # and the course is enrollable, then choose the new seat associated with the course instead,
+                        # and mark the original seat in the array to be removed
+                        logger.info(
+                            "Use course_run {} instead of course_run {} for total program price calculation".format(
+                                seat.course_run.key,
+                                selected_seat.course_run.key
+                            )
+                        )
+                        add_seat = True
+                        seats_to_remove.append(selected_seat)
+
+                if add_seat:
+                    course_map[course_uuid].append(seat)
+                for seat in seats_to_remove:
+                    # Now remove the seats that should not be counted for calculation for program total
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -809,18 +809,70 @@ def seats(self):
     def seat_types(self):
         return set(seat.type for seat in self.seats)
 
+    def _get_total_price_by_currency(self):
+        '''
```
clintonb: Double-quotes

clintonb (commented):
1. Make sure we communicate to Marketing, et. al. that, if there are course runs with different prices, the displayed total may be incorrect. Some learners may feel this is bait-and-switch.
2. Your logic should also account for the start/end dates of course runs. If a a course run has ended, it should not be part of the price calculation.

schenedx:
@clintonb Took in your review feedback. Can you let me know if you are OK with merge with this change?

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,18 +810,75 @@ def seats(self):
...
 
+    def _get_total_price_by_currency(self):
+        """
+        This helper function returns the total program price indexed by the currency
+        """
+        currencies_with_total = defaultdict()
+        course_map = defaultdict(list)
+        for seat in self.seats:
+            course_uuid = seat.course_run.course.uuid
+            # identify the most relavent course_run seat for a course, per currency.
```
clintonb: relevent

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,18 +810,75 @@ def seats(self):
...
+            # And use the price of the seat to represent the price of the course
+            selected_seats = course_map.get(course_uuid)
+            if not selected_seats:
+                # if we do not have this course yet, create the seats array
+                course_map[course_uuid] = [seat]
+            else:
+                add_seat = False
```
clintonb, line 828: This is a bit confusing.

1. Group seats by currency.
2. Within currency groups, group by course.
3. For each course, select the appropriate course run.
4. Sum the cost of seats by course and currency.

Is the above approach correct?

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,18 +810,75 @@ def seats(self):
...
+                        add_seat = True
+                    elif ((seat.course_run.end is None or
+                           seat.course_run.end >= datetime.datetime.now(pytz.UTC)) and
+                          (seat.course_run.enrollment_start is None or
+                           seat.course_run.enrollment_start > selected_seat.course_run.enrollment_start and
+                           seat.course_run.enrollment_start < datetime.datetime.now(pytz.UTC))):
+                        # If the seat has same currency, the course has not ended,
+                        # and the course is enrollable, then choose the new seat associated with the course instead,
+                        # and mark the original seat in the array to be removed
+                        logger.debug(
```
clintonb: Would this be useful to have logged in production? If so, use `info()`.

clintonb (commented):
This seems fine, but it would be nice to see the code simplified a bit to make it more legible. I have found [itertools.groupby()](https://docs.python.org/2/library/itertools.html#itertools.groupby) to be useful.

course_discovery/apps/course_metadata/models.py
schenedx: What I did:
1. Group the seats, regardless of currencies into courses first.
2. Take all the selected seats in the group, and re-grouped them by currency.
3. Sum the cost of seats by course and currency

clintonb (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,18 +810,75 @@ def seats(self):
...
+                # if we do not have this course yet, create the seats array
+                course_map[course_uuid] = [seat]
+            else:
+                add_seat = False
+                seats_to_remove = []
+                for selected_seat in selected_seats:
+                    if seat.currency != selected_seat.currency:
+                        # If the candidate seat has a different currency than the one in the array,
+                        # always add to the array
+                        add_seat = True
```
davec-edx, line 834: It's unclear to me why, if the currency is different, we don't check that the course is enrollable.  I would think we'd want to check that ... maybe not...  could you add a docstring saying what this should do?

schenedx: @davec-edx 
Keep in mind, each seat has a 1 on 1 relationship with price per currency. This means, if the price is for English Pound, it is a different seat.
To calculate the total program price, we need to have at least 1 seat per currency in the array.
Even if the seat is no longer enrollable or ended, we should use that seat price.

PR #677, opened 2017-02-21
ECOM-7188 Add config model to control max threads for data loaders 

MatthewPiatetsky:
@edx/ecommerce


course_discovery/apps/course_metadata/models.py
```diff
@@ -902,3 +903,10 @@ class Meta(object):
 class PersonWork(AbstractValueModel):
     """ Person Works model. """
     person = models.ForeignKey(Person, related_name='person_works')
+
+
+class DataLoadersConfig(SingletonModel):
```
amangano-edx: Would this be more at home in it's own app? It doesn't feel like it belongs mixed in with the metadata models.

amangano-edx:
Is there any way we can add tests to verify that this works as advertised?

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: well refresh_course_metadata is in the course_metadata app, so I think this belongs in the same place. If we decide to move that it would go with it.

MatthewPiatetsky:
This gets tested by the refresh course metadata tests (the previous run actually failed because of them)

amangano-edx:
@MatthewPiatetsky, do those tests verify that the max thread count gets set correctly? If not could we add a test there that does?

course_discovery/apps/course_metadata/models.py
amangano-edx: Ok

MatthewPiatetsky:
The tests are currently checking that the max_workers_override is set to 7, I wasn't sure what to default the setting to, but since it seemed we were expecting 7, I set it to default to 7 and those now pass (when it was set to 1 they failed)

course_discovery/apps/course_metadata/migrations/0048_dataloadersconfig.py
```diff
@@ -0,0 +1,25 @@
...
+    dependencies = [
+        ('course_metadata', '0047_personwork'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DataLoadersConfig',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('max_threads_for_data_loaders', models.PositiveSmallIntegerField(default=7)),
```
amangano-edx: Nitpick: Does this field need to be this verbose or could we get away with just calling it `max_threads`? Note that DataLoaders is already in the model/table name.

amangano-edx: Also does it make sense for 7 to be the default value for this table or would it be better to default to something like 1 and manually set our config to 7 in a migration? I wonder if this table is relevant to openedx users and if 7 would make sense as a default in their setups?

amangano-edx:
LGTM

course_discovery/apps/course_metadata/models.py
```diff
@@ -902,3 +903,10 @@ class Meta(object):
...
+    """
+    Control number of threads that data loaders use through refresh_course_metadata.
+    """
+    max_threads_for_data_loaders = models.PositiveSmallIntegerField(default=7)
```
rlucioni: @MatthewPiatetsky where'd you get a default of 7 from? I'd expect the default to be `None`, to preserve existing behavior (letting the `ThreadPoolExecutor` decide how many threads to use).

course_discovery/apps/course_metadata/models.py
rlucioni: Not a huge deal, but I'd name the model `DataLoaderConfig` (singular).

course_discovery/apps/course_metadata/models.py
```diff
@@ -902,3 +903,10 @@ class Meta(object):
 class PersonWork(AbstractValueModel):
     """ Person Works model. """
     person = models.ForeignKey(Person, related_name='person_works')
+
+
+class DataLoadersConfig(SingletonModel):
+    """
+    Control number of threads that data loaders use through refresh_course_metadata.
```
rlucioni: This is too specific, the model can house any data loader config.

course_discovery/apps/course_metadata/models.py
rlucioni: The name of this field can be more concise. The model is clearly for data loader config; we don't need to mention that again in the field name. Consider sticking to the terminology used by the [`Executor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) class and naming this `max_workers`.

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: That is true, but why not have the docstring describe all the things the model is doing?  
Though I suppose at that point it would be listed by parameter, so I'll change it to something more generic

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: I believe existing behavior is 7 (ThreadPoolExecutor does not decide)  
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py#L67

MatthewPiatetsky: I guess this would make that argument do nothing, should the argument be removed?

rlucioni: @MatthewPiatetsky good catch, I forgot about that.

rlucioni: @MatthewPiatetsky actually, you should get rid of that argument now. It's not used. We want to be using the config instead.

MatthewPiatetsky: removed

rlucioni (approved):

rlucioni:
Please remember to squash your commits.

clintonb (commented):
Are you missing a migration? How does the table get created?

MatthewPiatetsky:
@clintonb when I changed variable name and recreated migration i missed committing it which is why tests failed, it is there now

clintonb (approved):

PR #678, opened 2017-02-22
ENT-158: Make sure staff users can access all catalogs.

saleem-latif:
Hi @mattdrayer , @asadiqbal08 , @clintonb 

This PR contains changes for [ENT-158](https://openedx.atlassian.net/browse/ENT-158), I have updated the permissions filter so that staff users are able to access catalogs. Kindly take a look and share your thoughts.


mattdrayer:
Small update that LGTM -- @clintonb @bderusha we are not exactly sure how to add a proper test for this particular change -- can you point us in the right direction?

mattdrayer:
According to @clintonb this PR can be closed in favor of a change he has made.

PR #687, opened 2017-02-24
Fix bug on total program price calculation

schenedx:
ECOM-7023
@clintonb Could you please quickly review this?


clintonb (commented):
Test?

schenedx:
@clintonb Good call. Added a new test case to cover this change

clintonb (approved):

PR #688, opened 2017-02-24
Fix second bug with Program total price calculation ECOM-7023

schenedx:
@clintonb Please help review
I made the mistake on the last bug fix to not ensure my unit test cover the bug.
This time. I made sure.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -810,6 +810,28 @@ class Program(TimeStampedModel):
     def seat_types(self):
         return set(seat.type for seat in self.seats)
 
+    def _select_for_total_price(self, selected_seat, candidate_seat):
+        """
+        A helper function to determine which course_run seat is best suitable to be used to calculate
+        the program total price. A seat is most suitable if the related course_run is now enrollable,
+        has not ended, and the enrollment_start date is most recent
+        """
+        end_valid = candidate_seat.course_run.end is None or \
+            candidate_seat.course_run.end >= datetime.datetime.now(pytz.UTC)
+
+        selected_enrollment_start = selected_seat.course_run.enrollment_start or \
+            pytz.utc.localize(datetime.datetime.min)
+
+        # Only select the candidate seat if the candidate seat has no enrollment start,
+        # or make sure the candidate course_run is enrollable and
+        # the candidate seat enrollment start is most recent
+        enrollment_start_valid = candidate_seat.course_run.enrollment_start is None or (
+            candidate_seat.course_run.enrollment_start > selected_enrollment_start and
+            candidate_seat.course_run.enrollment_start < datetime.datetime.now(pytz.UTC)
+        )
+
+        return end_valid and enrollment_start_valid
+
     def _get_total_price_by_currency(self):
         """
         This helper function returns the total program price indexed by the currency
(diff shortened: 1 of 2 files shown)


schenedx:
@edx/ecommerce Please help review.
Trying to fix this bug before it gets into production

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,6 +810,23 @@ def seats(self):
     def seat_types(self):
         return set(seat.type for seat in self.seats)
 
+    def _select_for_total_price(self, selected_seat, candidate_seat):
+        """
+        A helper function to determine if the existing seat should be
```
MatthewPiatetsky: I'm not sure what this comment means, rather than this comment could you just explain maybe a high level of what this method is doing in english?

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,6 +810,23 @@ def seats(self):
...
+        """
+        end_valid = candidate_seat.course_run.end is None or \
+            candidate_seat.course_run.end >= datetime.datetime.now(pytz.UTC)
+
+        selected_enrollment_start = selected_seat.course_run.enrollment_start or \
+            pytz.utc.localize(datetime.datetime.min)
+
+        enrollment_start_valid = candidate_seat.course_run.enrollment_start is None or \
+            candidate_seat.course_run.enrollment_start > selected_enrollment_start and \
+            candidate_seat.course_run.enrollment_start < datetime.datetime.now(pytz.UTC)
```
amangano-edx: This might be clearer with parenthesis.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -513,17 +513,20 @@ def create_program_with_multiple_course_runs(self, set_all_dates=True):
...
-        course_runs_same_course = factories.CourseRunFactory.create_batch(2, course=course)
+        course_runs_same_course = factories.CourseRunFactory.create_batch(3, course=course)
         for course_run in single_course_course_runs:
             factories.SeatFactory(type='audit', currency=currency, course_run=course_run, price=0)
             factories.SeatFactory(type='verified', currency=currency, course_run=course_run, price=10)
 
         day_diff = 1
         for course_run in course_runs_same_course:
-            if set_all_dates or day_diff > 1:
+            if set_all_dates or day_diff < 2:
```
MatthewPiatetsky: what does day_diff mean? Could you make the variable name more descriptive

MatthewPiatetsky:
@schenedx could you explain what the bug was and what the fix is?

course_discovery/apps/course_metadata/models.py
amangano-edx: Why does the candidate seat enrollment_start need to be in the past? What if it starts tomorrow?

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,6 +810,23 @@ def seats(self):
...
+
```
davec-edx: I would love to see a comment that just describes 'how' the determination is being calculated.  I missed that this was comparing 'selected' seat v. 'candidate seat'.  So just a short comment to help the reader understand would be great.

davec-edx (approved):
other than the nit about the comment, lgtm

course_discovery/apps/course_metadata/models.py
```diff
@@ -810,6 +810,23 @@ def seats(self):
...
+    def _select_for_total_price(self, selected_seat, candidate_seat):
+        """
+        A helper function to determine if the existing seat should be
+        replaced by the new candidate seat for total program price calculation
+        """
+        end_valid = candidate_seat.course_run.end is None or \
+            candidate_seat.course_run.end >= datetime.datetime.now(pytz.UTC)
+
+        selected_enrollment_start = selected_seat.course_run.enrollment_start or \
+            pytz.utc.localize(datetime.datetime.min)
```
schenedx, line 823: @MatthewPiatetsky Yes. The fix for the bug is on this line. Before this change, I didn't add the `pytz.utc.localize` function. This means the `datetime.datetime.min` was a datetime without any timezone specified. When we do comparison with this information in the statement below, it raises an error with unmatched type. See https://splunk.edx.org/en-US/app/search/search?earliest=%40w0&latest=now&q=search%20index%3Dstage-edx%20service_variant%3Ddiscovery%20%22Internal%20Server%20Error%22&display.page.search.mode=fast&dispatch.sample_ratio=1&sid=1488226085.4673769

course_discovery/apps/course_metadata/models.py
schenedx: @amangano-edx We don't want to use a course_run with enrollment_start in the future if possible.

schenedx:
@MatthewPiatetsky @amangano-edx Took in your feedback

PR #689, opened 2017-02-24
WIP - Very early first pass at computing distinct counts by aggregation key

amangano-edx:
This is necessary to allow us to filter duplicate course_runs from the response on the front end while still displaying accurate counts.

I had to touch a lot of code to get this to work, but I've tried to keep the changes as isolated as possible to minimize risk. There is still more work to do before this is ready for prod, but for now it would be helpful to get some feedback from the rest of the team on the approach.

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/v1/views/search.py
+++ b/course_discovery/apps/api/v1/views/search.py
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
     facet_serializer_class = serializers.AggregateFacetSearchSerializer
     serializer_class = serializers.AggregateSearchSerializer
 
+    # FIXME Should this be separated and included from an edx-specific app?
+    @list_route(methods=['get'], url_path='distinct_facets')
+    def distinct_facets(self, request):
+        from course_discovery.apps.edx_haystack_extensions.distinct_counts import (DistinctCountsSearchQuerySet,
+                DistinctCountsAggregateFacetSearchSerializer)
+
+        queryset = self.filter_facet_queryset(self.get_queryset())
+        for facet in request.query_params.getlist("selected_facets"):
+            if ":" not in facet:
+                continue
+
+            field, value = facet.split(":", 1)
+            if value:
+                queryset = queryset.narrow('%s:"%s"' % (field, queryset.query.clean(value)))
+
+        queryset = DistinctCountsSearchQuerySet.from_queryset(queryset).set_aggregation_key('aggregation_key_exact')
+
+        serializer_kwargs = {'context': self.get_serializer_context()}
+        serializer_kwargs['context'].update({'objects': queryset, 'many': False})
+        serializer = DistinctCountsAggregateFacetSearchSerializer(queryset.facet_counts(), **serializer_kwargs)
+        return Response(serializer.data)
+
 
 class TypeaheadSearchView(PartnerMixin, APIView):
     """ Typeahead for courses and programs. """
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -110,6 +110,13 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
 
     prerequisites = indexes.MultiValueField(faceted=True)
 
+    # A key that can be used to group related documents together to enable the computation of distinct facet and hit
+    # counts.
+    aggregation_key = indexes.CharField(faceted=True)
+
+    def prepare_aggregation_key(self, obj):
+        return 'course:{}'.format(obj.key)
+
     def prepare_course_runs(self, obj):
         return [course_run.key for course_run in obj.course_runs.all()]
 
@@ -152,6 +159,13 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     subject_uuids = indexes.MultiValueField()
     has_enrollable_paid_seats = indexes.BooleanField(null=False)
     paid_seat_enrollment_end = indexes.DateTimeField(null=True)
+
+    # A key that can be used to group related documents together to enable the computation of distinct facet and hit
+    # counts.
+    aggregation_key = indexes.CharField(faceted=True)
+
+    def prepare_aggregation_key(self, obj):
+        return 'courserun:{}'.format(obj.course.key)
 
     def prepare_has_enrollable_paid_seats(self, obj):
         return obj.has_enrollable_paid_seats()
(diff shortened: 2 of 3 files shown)


course_discovery/apps/api/v1/views/search.py
```diff
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
     facet_serializer_class = serializers.AggregateFacetSearchSerializer
     serializer_class = serializers.AggregateSearchSerializer
 
+    # FIXME Should this be separated and included from an edx-specific app?
+    @list_route(methods=['get'], url_path='distinct_facets')
```
clintonb, line 122: None of these method kwargs are necessary.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
     facet_serializer_class = serializers.AggregateFacetSearchSerializer
     serializer_class = serializers.AggregateSearchSerializer
 
+    # FIXME Should this be separated and included from an edx-specific app?
```
clintonb, line 121: Yeah, this needs to be in the `edx_catalog_extensions` app.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+    @list_route(methods=['get'], url_path='distinct_facets')
+    def distinct_facets(self, request):
+        from course_discovery.apps.edx_haystack_extensions.distinct_counts import (DistinctCountsSearchQuerySet,
+                DistinctCountsAggregateFacetSearchSerializer)
+
+        queryset = self.filter_facet_queryset(self.get_queryset())
+        for facet in request.query_params.getlist("selected_facets"):
```
clintonb, line 128: Single quotes

course_discovery/apps/api/v1/views/search.py
```diff
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+            if ":" not in facet:
```
clintonb, line 129: What does this mean? Add comments.

course_discovery/apps/api/v1/views/search.py
clintonb: Add a comment explaining what's going on here.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
     facet_serializer_class = serializers.AggregateFacetSearchSerializer
     serializer_class = serializers.AggregateSearchSerializer
 
+    # FIXME Should this be separated and included from an edx-specific app?
+    @list_route(methods=['get'], url_path='distinct_facets')
+    def distinct_facets(self, request):
+        from course_discovery.apps.edx_haystack_extensions.distinct_counts import (DistinctCountsSearchQuerySet,
```
clintonb, line 124: Add a docstring explaining the purpose of this endpoint, and an overview of what it's doing.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -118,6 +118,28 @@ class AggregateSearchViewSet(BaseHaystackViewSet):
...
+                continue
+
+            field, value = facet.split(":", 1)
+            if value:
+                queryset = queryset.narrow('%s:"%s"' % (field, queryset.query.clean(value)))
+
+        queryset = DistinctCountsSearchQuerySet.from_queryset(queryset).set_aggregation_key('aggregation_key_exact')
+
+        serializer_kwargs = {'context': self.get_serializer_context()}
+        serializer_kwargs['context'].update({'objects': queryset, 'many': False})
```
clintonb, line 139: This can be  cleaner:

```python
context = self.get_serializer_context()
context.update({'objects': queryset, 'many': False})
serializer = DistinctCountsAggregateFacetSearchSerializer(queryset.facet_counts(), context=context)
```

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -110,6 +110,13 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
 
     prerequisites = indexes.MultiValueField(faceted=True)
 
+    # A key that can be used to group related documents together to enable the computation of distinct facet and hit
```
clintonb, line 113: Why are you touching programs and courses if you're only dealing with course runs?

course_discovery/apps/edx_haystack_extensions/distinct_counts.py
```diff
@@ -0,0 +1,289 @@
+import elasticsearch
```
clintonb, line 1: This needs to be broken up appropriately:

* distinct_counts/backends.py
* distinct_counts/serializers.py
* distinct_counts/query.py

clintonb (commented):
I'm honestly not sure what's going on here. Before adding more code, we need to document the expected behavior of this new endpoint. What are the inputs? What's the expected output? Once you have that document what's going on with aggregations.

tl;dr Less code, more English.

clintonb:
P.S. Add the ticket number to your commit message.

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky, line 128: can you add a comment explaining this?

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky, line 124: i would put the imports at the top

course_discovery/apps/course_metadata/search_indexes.py
amangano-edx: because the marketing site expects to get back both courseruns and programs with the response, and in order for ES to compute the distinct counts correctly, all records need to have a value for the aggregation key.

course_discovery/apps/api/v1/views/search.py
amangano-edx: This is doing the same thing as the `facets` method in this file and in the drf_haystack mixins module.

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: why is getting facet results based on deduped data edx specific? or would you say that because you think search for multiple course runs isn't something we'll support now for open source users

course_discovery/apps/api/v1/views/search.py
amangano-edx: This follows what is being done in the drf_haystack `facets` endpoint method.

course_discovery/apps/api/v1/views/search.py
amangano-edx: This follows what is being done in the drf_haystack facets endpoint method.

course_discovery/apps/api/v1/views/search.py
amangano-edx: This follows what is being done in the drf_haystack facets endpoint method.

course_discovery/apps/api/v1/views/search.py
amangano-edx: I just put it here to keep all the hacky code together. I will move the imports to the top when I move this method into the edx_catalog_extensions app.

course_discovery/apps/api/v1/views/search.py
amangano-edx: It's an edx-specific hack that we do not want to be exposed to the general community, so I agree with Clinton that it should be in an edx-specific app.

amangano-edx:
replaced by https://github.com/edx/course-discovery/pull/704

PR #694, opened 2017-02-27
ECOM-7336 Add upgradeable filter to course runs

MatthewPiatetsky:
Here is the logic as I understood it for this [ticket](https://openedx.atlassian.net/browse/ECOM-7336).

For the course run dialog, _In addition to current filtering_:
1. Only include a course run if a course run has a verified or professional seat and that seat has an upgrade deadline that is none or in the future.

If we go with this approach, changing the marketing site will not be necessary, since we can modify the existing query param that the marketing site passes.

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -542,7 +542,7 @@ class CourseWithProgramsSerializer(CourseSerializer):
 
         if self.context.get('marketable_enrollable_course_runs_with_archived'):
             # Same as "marketable_course_runs_only", but includes courses with an end date in the past
-            course_runs = course_runs.marketable().enrollable()
+            course_runs = course_runs.marketable().enrollable().upgradeable()
 
         if self.context.get('published_course_runs_only'):
             course_runs = course_runs.filter(status=CourseRunStatus.Published)
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -26,7 +26,7 @@ from course_discovery.apps.core.tests.factories import UserFactory
 from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
-from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
+from course_discovery.apps.course_metadata.models import Course, CourseRun, Program, Seat
 from course_discovery.apps.course_metadata.tests.factories import (
     CorporateEndorsementFactory, CourseFactory, CourseRunFactory, EndorsementFactory, ExpectedLearningItemFactory,
     ImageFactory, JobOutlookItemFactory, OrganizationFactory, PersonFactory, PositionFactory, PrerequisiteFactory,
@@ -260,9 +260,9 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
             enrollment_end=None,
             course=self.course
         )
-        SeatFactory(course_run=unpublished_course_run)
-        SeatFactory(course_run=enrollable_course_run)
-        SeatFactory(course_run=archived_course_run)
+        SeatFactory(course_run=unpublished_course_run, upgrade_deadline=None, type=Seat.VERIFIED)
+        SeatFactory(course_run=enrollable_course_run, upgrade_deadline=None, type=Seat.VERIFIED)
+        SeatFactory(course_run=archived_course_run, upgrade_deadline=None, type=Seat.VERIFIED)
 
         context = {
             'request': self.request,
--- a/course_discovery/apps/course_metadata/query.py
+++ b/course_discovery/apps/course_metadata/query.py
@@ -83,6 +83,22 @@ class CourseRunQuerySet(models.QuerySet):
             status=CourseRunStatus.Published
         )
 
+    def upgradeable(self):
+        """ Returns course runs which have a verified or professional seat and do not have
+        an expired upgrade deadline.
+
+        Returns:
+            QuerySet
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        # Nested to avoid circular import.
+        from course_discovery.apps.course_metadata.models import Seat
+        return self.filter(
+            Q(seats__type__contains=Seat.VERIFIED) | Q(seats__type__contains=Seat.PROFESSIONAL)
+        ).exclude(
+            Q(seats__upgrade_deadline__lt=now)
+        )
+
 
 class ProgramQuerySet(models.QuerySet):
     def marketable(self):
--- a/course_discovery/apps/course_metadata/tests/test_query.py
+++ b/course_discovery/apps/course_metadata/tests/test_query.py
@@ -5,7 +5,7 @@ import pytz
 from django.test import TestCase
 
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
-from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
+from course_discovery.apps.course_metadata.models import Course, CourseRun, Program, Seat
 from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, ProgramFactory, SeatFactory
 
 
@@ -119,6 +119,33 @@ class CourseRunQuerySetTests(TestCase):
             course_run.save()
 
         self.assertEqual(CourseRun.objects.marketable().exists(), is_published)
+
+    def test_upgradeable(self):
+        """ Verify the method returns only course runs whose upgrade deadline has not passed
+        or is not defined for verified/professional modes. """
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        audit_mode = CourseRunFactory()
+        SeatFactory(upgrade_deadline=future, course_run=audit_mode, type=Seat.AUDIT)
+
+        upgrade_deadline_passed = CourseRunFactory()
+        SeatFactory(upgrade_deadline=past, course_run=upgrade_deadline_passed, type=Seat.VERIFIED)
+
+        upgrade_deadline_not_passed = CourseRunFactory()
+        SeatFactory(upgrade_deadline=future, course_run=upgrade_deadline_not_passed, type=Seat.VERIFIED)
+
+        upgrade_deadline_not_defined = CourseRunFactory()
+        SeatFactory(upgrade_deadline=None, course_run=upgrade_deadline_not_defined, type=Seat.VERIFIED)
+        SeatFactory(upgrade_deadline=future, course_run=upgrade_deadline_not_defined, type=Seat.AUDIT)
+
+        upgrade_deadline_not_defined_professional = CourseRunFactory()
+        SeatFactory(upgrade_deadline=None, course_run=upgrade_deadline_not_defined_professional, type=Seat.PROFESSIONAL)
+
+        # order doesn't matter
+        assert sorted(CourseRun.objects.upgradeable(), key=lambda x: x.id) == \
+            sorted([upgrade_deadline_not_passed, upgrade_deadline_not_defined,
+                    upgrade_deadline_not_defined_professional], key=lambda x: x.id)
 
 
 @ddt.ddt
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -76,7 +76,7 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
               mulitple: false
             - name: marketable_enrollable_course_runs_with_archived
               description: Restrict returned course runs to those that are published, have seats,
-                and can be enrolled in now. Includes archived courses.
+              can be enrolled in now and can be upgraded. Includes archived courses.
               required: false
               type: integer
               paramType: query


MatthewPiatetsky:
is codecov complaining about a publisher change?

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +83,32 @@ def marketable(self):
...
+            QuerySet
+        """
+        now = datetime.datetime.now(pytz.UTC)
+
+        upgradeable_course_runs, seat_types = [], set()
+
+        for course_run in self:
+            for seat in course_run.seats.all():
+                seat_types.add(seat.type)
+                if (seat.type in ['verified', 'professional'] and
```
amangano-edx: I wrote similar logic to this for the boosting config, except I filtered out 'credit' seats (which are not individually purchasable) instead of explicitly selecting 'verified' and 'professional'. I think it might be better to do that here as well, as it would keep the logic consistent. I added a constant to make selecting the Seats to filter out easier. See https://github.com/edx/course-discovery/blob/945215cfd243ed3fe4550ce1995974fccc5e90e2/course_discovery/apps/course_metadata/models.py#L595

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +83,32 @@ def marketable(self):
             status=CourseRunStatus.Published
         )
 
+    def upgradeable(self):
+        """ Returns course runs which do not have an upgrade deadline or
+        whose upgrade deadline has not passed for their verified/professional modes.
+
+        Returns all course runs if all modes on all course runs are audit.
+
+
```
amangano-edx: Nitpick: Do we need two newlines here?

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +83,32 @@ def marketable(self):
             status=CourseRunStatus.Published
         )
 
+    def upgradeable(self):
```
amangano-edx, line 86: This seems to violate the traditional queryset filter contract where queryset filters are used to build up the query to run but do not actually run the query themselves. Perhaps it would be better to add this to the model class instead?

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +83,32 @@ def marketable(self):
...
+        """ Returns course runs which do not have an upgrade deadline or
+        whose upgrade deadline has not passed for their verified/professional modes.
+
+        Returns all course runs if all modes on all course runs are audit.
+
+
+        Returns:
+            QuerySet
```
amangano-edx: Lines 108 and 110 do not seem to be returning Querysets.

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +83,32 @@ def marketable(self):
...
+        for course_run in self:
+            for seat in course_run.seats.all():
+                seat_types.add(seat.type)
+                if (seat.type in ['verified', 'professional'] and
+                        (seat.upgrade_deadline is None or seat.upgrade_deadline > now)):
+                    upgradeable_course_runs.append(course_run)
+
+        if upgradeable_course_runs:
+            return upgradeable_course_runs
+        elif seat_types == {'audit'}:
```
amangano-edx: This seems fragile. If we were to add a new seat type in the future that is not 'verified' or 'professional', this method could potentially return nothing.

course_discovery/apps/api/serializers.py
```diff
@@ -541,7 +541,7 @@ def get_course_runs(self, course):
 
         if self.context.get('marketable_enrollable_course_runs_with_archived'):
```
amangano-edx, line 543: I thought I saw documentation for these filters somewhere in another PR. If we're changing the functionality for this one, we should update those docs if they exist.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -260,9 +260,9 @@ def test_marketable_enrollable_course_runs_with_archived(self):
             enrollment_end=None,
             course=self.course
         )
-        SeatFactory(course_run=unpublished_course_run)
-        SeatFactory(course_run=enrollable_course_run)
-        SeatFactory(course_run=archived_course_run)
+        SeatFactory(course_run=unpublished_course_run, upgrade_deadline=None, type='verified')
+        SeatFactory(course_run=enrollable_course_run, upgrade_deadline=None, type='verified')
+        SeatFactory(course_run=archived_course_run, upgrade_deadline=None, type='verified')
```
amangano-edx: You should use the `Seat` type constants when specifying type. For example, `Seat.VERIFIED`. See https://github.com/edx/course-discovery/blob/945215cfd243ed3fe4550ce1995974fccc5e90e2/course_discovery/apps/course_metadata/models.py#L587

course_discovery/apps/course_metadata/query.py
amangano-edx: Use `Seat` type constants when specifying seat types. `Seat.AUDIT`.

course_discovery/apps/course_metadata/query.py
amangano-edx: Use `Seat` type constants when specifying Seat types. `Seat.VERIFIED`, `Seat.PROFESSIONAL`.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -120,6 +120,40 @@ def test_marketable_unpublished_exclusions(self, is_published):
 
         self.assertEqual(CourseRun.objects.marketable().exists(), is_published)
 
+    def test_upgradeable(self):
+        """ Verify the method returns only course runs whose upgrade deadline has not passed
+        or is not defined for verified/professional modes. """
+        past = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
+        future = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
+
+        upgrade_deadline_passed = CourseRunFactory()
```
amangano-edx: did you mean 'past'?

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -120,6 +120,40 @@ def test_marketable_unpublished_exclusions(self, is_published):
...
+        SeatFactory(upgrade_deadline=past, course_run=upgrade_deadline_passed, type='verified')
+        upgrade_deadline_not_passed = CourseRunFactory()
+        SeatFactory(upgrade_deadline=future, course_run=upgrade_deadline_not_passed, type='verified')
+        upgrade_deadline_not_defined = CourseRunFactory()
+        SeatFactory(upgrade_deadline=None, course_run=upgrade_deadline_not_defined, type='verified')
```
amangano-edx: Use `Seat` type constants when specifying Seat types.

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: removed

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: It seems to me the opposite would be fragile.  
The two cases I discussed with Lauren explicitly apply to verified/professional and to audit.  
We don't know what other seat types might be in the future and what the business rules would be for them, if we tried to guess what to do with other seat types we could very well guess wrong. I'd rather just implement the rules that we know we want.

MatthewPiatetsky: changed

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/tests/test_query.py
MatthewPiatetsky: i meant that the upgrade deadline passed

course_discovery/apps/course_metadata/tests/test_query.py
MatthewPiatetsky: changed

course_discovery/apps/course_metadata/query.py
amangano-edx: Maybe we should explicitly raise an error and log in that case then, so whomever is adding a new seat type knows right away that they will need to update this method as well?

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: good question, I'm not sure, is there somewhere in the django docs that references this contract?  

@rlucioni

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: If you think we should add a clearer comment that makes sense to me, but I'm not sure in what case we would want to raise an error.
We are defining what we want here by explicitly looking at these types. Other seat types wouldn't make our logic as defined incorrect if they were present. 
They are just not applicable so we ignore those seats and determine if the course run is upgradeable based on the rule we defined.

course_discovery/apps/api/serializers.py
MatthewPiatetsky: good point, updated

course_discovery/apps/course_metadata/query.py
amangano-edx: I think it's plausible that there could be a third state here where there are no upgradable seats but there are some non-audit seat types. For instance, maybe there are verified/professional seats for which the upgrade deadlines have passed? Maybe there's a new seat type besides 'verified' and 'professional' that is added in the future. I think we need to do something to deal with this state.

amangano-edx: If this state arises because of something we didn't plan for, like a new seat type, it might make sense to raise so that the implementer of that new seat type knows that they need to update this method. If you just add a comment they might not see it prior to shipping their code to prod.

MatthewPiatetsky: I'm not sure what you mean. If there are verified/professional seats for which the upgrade deadlines have passed, the logic we've defined in the ticket/in my discussions with Lauren explicitly says we shouldn't return any course runs. We can already have other seat types but they do not change the correct result according to the current logic as defined

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +83,32 @@ def marketable(self):
...
+            List
+        """
+        now = datetime.datetime.now(pytz.UTC)
+
+        upgradeable_course_runs, seat_types = [], set()
+
+        for course_run in self:
+            for seat in course_run.seats.all():
+                seat_types.add(seat.type)
+                from course_discovery.apps.course_metadata.models import Seat
```
MatthewPiatetsky: if i put this import at the top i get an error, any ideas why that is?

rlucioni: @MatthewPiatetsky hard to say for sure without actually seeing the error, but it's likely a circular import. The models also import from `course_discovery.apps.course_metadata.query`.

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky read the first paragraph [here](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#when-querysets-are-evaluated). I'd prefer to see this implemented using filters and `Q` objects, if possible.

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +84,22 @@ def marketable(self):
             status=CourseRunStatus.Published
         )
 
+    def upgradeable(self):
+        """ Returns course runs which do not have an upgrade deadline or
+        whose upgrade deadline has not passed for their verified/professional modes.
+
+        Returns all course runs if all modes on all course runs are audit.
```
amangano-edx: Is this still true?

MatthewPiatetsky: thanks, fixed

course_discovery/apps/course_metadata/query.py
```diff
@@ -4,6 +4,7 @@
 from django.db import models
 from django.db.models.query_utils import Q
 
+import course_discovery.apps.course_metadata.models
```
rlucioni: I'm surprised to see that this prevents the circular import problem from before. Regardless, the nested import you had before is more common, usually with a comment indicating that the import is nested to prevent a circular import.

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +84,21 @@ def marketable(self):
...
 
+    def upgradeable(self):
+        """ Returns course runs which have a verified or professional seat and do not have
+        an expired upgrade deadline.
+
+        Returns:
+            QuerySet
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        seat = course_discovery.apps.course_metadata.models.Seat
```
rlucioni: Lowercase `seat` usually implies a specific instance of `Seat`.

rlucioni (approved):
@MatthewPiatetsky looks good, though I would suggest making your acrobatics to get around the circular import more clear.

rlucioni:
Don't forget to squash before you merge!

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: http://stackoverflow.com/a/37126790  
I found this informative. I'm fine with changing, but an import at the top of the file seems standard to me and less hacky feeling than inside the function?  

what makes you say the nested import is more common, we have absolute imports without the from like this everywhere

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: good point, changed to seat_types

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky did you read the edit?

> The ImportError and AttributeError issues only occur in python 2. In python 3 the import machinery has been rewritten and all of these import statements (with the exception of 4) will work, even with circular dependencies.

Absolute imports like this are generally fine, but it's not very nice to the next person to disguise a circular import like this. You might now know it just by looking at this.

course_discovery/apps/course_metadata/query.py
```diff
@@ -83,6 +84,21 @@ def marketable(self):
...
 
+    def upgradeable(self):
+        """ Returns course runs which have a verified or professional seat and do not have
+        an expired upgrade deadline.
+
+        Returns:
+            QuerySet
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        seat_types = course_discovery.apps.course_metadata.models.Seat
```
rlucioni: @MatthewPiatetsky this isn't quite what I meant; `seat_types` is actually more misleading because you're really using the `Seat` model (not `SeatType`). I would do this:

```
# Nested to avoid circular import.
from course_discovery.apps.course_metadata.models import Seat

# Do things with Seat...
```

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: hmm..that is odd..my virtualenv has python 3 but I definitely get an error if i put the from import at the top

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: ok sure, fixed

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -120,6 +120,26 @@ def test_marketable_unpublished_exclusions(self, is_published):
...
+
+        audit_mode = CourseRunFactory()
+        SeatFactory(upgrade_deadline=future, course_run=audit_mode, type=Seat.AUDIT)
+        upgrade_deadline_passed = CourseRunFactory()
+        SeatFactory(upgrade_deadline=past, course_run=upgrade_deadline_passed, type=Seat.VERIFIED)
+        upgrade_deadline_not_passed = CourseRunFactory()
+        SeatFactory(upgrade_deadline=future, course_run=upgrade_deadline_not_passed, type=Seat.VERIFIED)
+        upgrade_deadline_not_defined = CourseRunFactory()
+        SeatFactory(upgrade_deadline=None, course_run=upgrade_deadline_not_defined, type=Seat.VERIFIED)
+        SeatFactory(upgrade_deadline=future, course_run=upgrade_deadline_not_defined, type=Seat.AUDIT)
```
amangano-edx: We should probably test that it returns professional seats as well.

amangano-edx: We should also test that it works with a mix of seat types.

amangano-edx:
I'd like to see a couple more tests, but other than that LGTM

PR #698, opened 2017-02-28
Ecom 6815 Add aggregation key to search indexes

amangano-edx:
This PR adds a new field, `aggregation_key`, to the CourseIndex, CourseRunIndex, and ProgramIndex and also includes it in the search serializers. This is necessary for several reasons:
- The `aggregation_key` field allows us to compute accurate counts for distinct records across content types, which will be necessary when we begin filtering duplicate cards from search results on the marketing site.
- By including the `aggregation_key` in the serializers, it will be available in the client-side js code on the marketing site and can be used to filter duplicate cards from the search results page in a way that is consistent with how the distinct counts are computed. We were going to have to add a field to the serializers anyway to support client-side filtering, as course_key is not currently included in the CourseRunSearchSerializer.

This will require running the update_index command after deploy.

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -13,7 +13,7 @@ from course_discovery.apps.api.fields import ImageField, StdImageSerializerField
 from course_discovery.apps.api.serializers import (
     AffiliateWindowSerializer, CatalogSerializer, ContainedCourseRunsSerializer, ContainedCoursesSerializer,
     CorporateEndorsementSerializer, CourseRunSearchSerializer, CourseRunSerializer, CourseRunWithProgramsSerializer,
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1276,8 +1300,9 @@ def test_data(self):
             'title': course_run.title,
             'orgs': [org.key for org in course_run.authoring_organizations.all()],
             'marketing_url': course_run.marketing_url,
+            'aggregation_key': 'courserun:{}'.format(course_run.course.key)
```
davec-edx, line 1303: Consider adding a comma at the end even though nothing follows.  It helps reduce changes if something is added after it.  I think you'll notice that the marketing_url already had one so you didn't need to add it when you added the aggregation key.

amangano-edx:
Replaced by https://github.com/edx/course-discovery/pull/699 and https://github.com/edx/course-discovery/pull/700

PR #699, opened 2017-02-28
ECOM-6815 Add aggregation_key to indices

amangano-edx:
This PR adds a new field, `aggregation_key`, to the CourseIndex, CourseRunIndex, and ProgramIndex. This is necessary as it allows us to compute accurate counts for distinct records across content types, which will be necessary when we begin filtering duplicate cards from search results on the marketing site.

This will require running the update_index command after deploy.

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -47,8 +47,11 @@ class OrganizationsMixin:
 class BaseIndex(indexes.SearchIndex):
     model = None
 
-    text = indexes.CharField(document=True, use_template=True)
+    # A key that can be used to group related documents together to enable the computation of distinct facet and hit
+    # counts.
+    aggregation_key = indexes.CharField()
     content_type = indexes.CharField(faceted=True)
+    text = indexes.CharField(document=True, use_template=True)
 
     def prepare_content_type(self, obj):  # pylint: disable=unused-argument
         return self.model.__name__.lower()
@@ -110,6 +113,9 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
 
     prerequisites = indexes.MultiValueField(faceted=True)
 
+    def prepare_aggregation_key(self, obj):
+        return 'course:{}'.format(obj.key)
+
     def prepare_course_runs(self, obj):
         return [course_run.key for course_run in obj.course_runs.all()]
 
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -110,6 +110,13 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
 
     prerequisites = indexes.MultiValueField(faceted=True)
 
+    # A key that can be used to group related documents together to enable the computation of distinct facet and hit
+    # counts.
+    aggregation_key = indexes.CharField()
```
rlucioni: Think it's worth keeping the fields and methods ordered alphabetically?

rlucioni (approved):

rlucioni:
@amangano-edx please add a more useful commit message before you merge this. The description you used in the first comment on this PR would work well as a second line. Please also include the ticket number.

course_discovery/apps/course_metadata/search_indexes.py
amangano-edx: They don't seem to be ordered like that in the other indices. I think that's a worthy thing to try to do but I don't think reordering all of the fields is in scope here.

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -153,6 +160,13 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     has_enrollable_paid_seats = indexes.BooleanField(null=False)
     paid_seat_enrollment_end = indexes.DateTimeField(null=True)
 
+    # A key that can be used to group related documents together to enable the computation of distinct facet and hit
+    # counts.
+    aggregation_key = indexes.CharField()
```
MatthewPiatetsky: I think you could define aggregation_key in the base index class and then override the prepare method in the subclasses

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -153,6 +160,13 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
...
+
+    def prepare_aggregation_key(self, obj):
+        return 'courserun:{}'.format(obj.course.key)
```
MatthewPiatetsky: why is this course key and not course run key?  
If this is intentional it needs to be explained in a comment

course_discovery/apps/course_metadata/search_indexes.py
amangano-edx: Good idea.

course_discovery/apps/course_metadata/search_indexes.py
amangano-edx: If we ever request search results with both courserun and course objects, I'm not sure if we'd want to dedup them using the same key. I introduced the course/courserun portion of the aggregation key to prevent a situation where a Course object is returned first in results and subsequent CourseRuns get filtered out because they have the same key.

amangano-edx: We use the course key here instead of the course_run key because we want to dedup course_runs that belong to the same course.

MatthewPiatetsky: Why would we ever want both courses and course runs in the same search results?  
Please add a comment explaining that we are aggregating course runs by course to deduplicate the course runs by course.

amangano-edx: No idea why we would ever want both, but just in case I added the content type piece to prevent issues in that case.

I added a comment in the most recent commit.

MatthewPiatetsky (approved):

PR #700, opened 2017-02-28
Ecom 6815 Add aggregation key to search serializers

amangano-edx:
This PR adds the new field, `aggregation_key`, to the search serializers. This is necessary so that it can be used on the marketing site to filter duplicate cards from the search results page in a way that is consistent with how the distinct counts are computed. We were going to have to add a field to the serializers anyway to support client-side filtering, as course_key is not currently included in the CourseRunSearchSerializer.

Depends on https://github.com/edx/course-discovery/pull/699

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -53,6 +53,7 @@ COURSE_RUN_SEARCH_FIELDS = (
     'enrollment_end', 'pacing_type', 'language', 'transcript_languages', 'marketing_url', 'content_type', 'org',
     'number', 'seat_types', 'image_url', 'type', 'level_type', 'availability', 'published', 'partner', 'program_types',
     'authoring_organization_uuids', 'subject_uuids', 'staff_uuids', 'mobile_available', 'logo_image_urls',
+    'aggregation_key',
 )
 
 PROGRAM_FACET_FIELD_OPTIONS = {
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1001,6 +1002,7 @@ class Meta:
 
 class CourseRunSearchSerializer(HaystackSerializer):
     availability = serializers.SerializerMethodField()
+    aggregation_key = serializers.CharField()
```
rlucioni: Should this just be added to `Meta.fields`?

rlucioni (approved):
As with https://github.com/edx/course-discovery/pull/699, please add a more useful message to the commit that this PR will end up merging.

course_discovery/apps/api/serializers.py
```diff
@@ -980,7 +981,7 @@ class CourseSearchSerializer(HaystackSerializer):
 
     class Meta:
         field_aliases = COMMON_SEARCH_FIELD_ALIASES
-        fields = ('key', 'title', 'short_description', 'full_description', 'text',)
+        fields = ('key', 'title', 'short_description', 'full_description', 'text', 'aggregation_key',)
```
MatthewPiatetsky, line 984: why add this to course if we will not be using it?

amangano-edx: Consistency. It is possible we could ask for all content types when making requests to the new endpoint and we should be able to handle course records the same way we handle everything else.

MatthewPiatetsky: personally i think if we wanted to use it in the future that we should add it then, but this is fine

MatthewPiatetsky:
👍

PR #702, opened 2017-03-01
Fixed course team user cannot change role assignment for CourseTeam role.

waheedahmed:
[ECOM-7353](https://openedx.atlassian.net/browse/ECOM-7353)

Fixed course team user cannot change role assignment for CourseTeam role. `CanViewAssociatedCourse` was enough to check permission for both type of internal and course team users, removed extra permission class to fix this bug.


awais786 (approved):

waheedahmed:
@bderusha Please review.

tasawernawaz (approved):

PR #704, opened 2017-03-01
WIP ECOM-6815 computing distinct search result counts

amangano-edx:
This is necessary to allow us to filter duplicate course_runs from the response on the front end while still displaying accurate counts.

This is the second version of this code, based on feedback from https://github.com/edx/course-discovery/pull/689.

@edx/ecommerce


course_discovery/apps/edx_catalog_extensions/views.py
```diff
@@ -0,0 +1,61 @@
...
+from course_discovery.apps.edx_catalog_extensions.serializers import DistinctCountsAggregateFacetSearchSerializer
+from course_discovery.apps.edx_haystack_extensions.distinct_counts.query import DistinctCountsSearchQuerySet
+
+
+class DistinctCountsAggregateSearchViewSet(AggregateSearchViewSet):
+    """
+    Provides a facets endpoint that returns distinct hit and facet counts with the rest of the query results.
+    
+    This ViewSet only exposes a single endpoint, `/search/all/facets`, which should be functionally identical to the
+    AggregateSearchViewSet facets endpoint except that it also include distinct hit and facet counts. The other
```
rlucioni: s/include/includes

course_discovery/apps/edx_catalog_extensions/views.py
```diff
@@ -0,0 +1,61 @@
...
+                pytype: str
+              required: false
+        """
+        return super(DistinctCountsAggregateSearchViewSet, self).facets(request)
+
+    def list(self, request):
+        """
+        Raise NotImplemented, as this endpoint is not currently supported.
+        """
+        raise NotImplemented('The list endpoint is not supported for this ViewSet.')
```
rlucioni: s/endpoint/action, s/supported for/supported by

course_discovery/apps/edx_catalog_extensions/views.py
```diff
@@ -0,0 +1,61 @@
+from course_discovery.apps.api.v1.views.search import AggregateSearchViewSet
+from course_discovery.apps.edx_catalog_extensions.serializers import DistinctCountsAggregateFacetSearchSerializer
+from course_discovery.apps.edx_haystack_extensions.distinct_counts.query import DistinctCountsSearchQuerySet
+
+
+class DistinctCountsAggregateSearchViewSet(AggregateSearchViewSet):
+    """
+    Provides a facets endpoint that returns distinct hit and facet counts with the rest of the query results.
+    
+    This ViewSet only exposes a single endpoint, `/search/all/facets`, which should be functionally identical to the
```
rlucioni: DRF refers to methods at this level as "actions" instead of as endpoints. You only have endpoints with URLs at the routing layer (urls.py).

course_discovery/urls.py
```diff
@@ -50,6 +50,11 @@
     url(r'^taggit_autosuggest/', include('taggit_autosuggest.urls')),
 ]
 
+if 'course_discovery.apps.edx_catalog_extensions' in settings.INSTALLED_APPS:
+    urlpatterns.append(
+        url(r'^api-edx/', include('course_discovery.apps.edx_catalog_extensions.urls'))
```
rlucioni: I'm not sure about this name. I'm concerned that it would be easy to confuse with `api.edx...`. Does this app even need it's own URL routing? What about registering this in `course_discovery.apps.api.v1.urls.py`?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -414,6 +414,23 @@ def test_results_ordered_by_start_date(self, ordering):
...
         self.assertEqual(response.data['objects']['results'], expected)
 
+    def test_results_include_aggregation_key(self):
+        """ Verify the search results only include the aggregation_key for each document. """
+        course_run = CourseRunFactory(course__partner=self.partner, status=CourseRunStatus.Published)
+        program = ProgramFactory(partner=self.partner, status=ProgramStatus.Active)
+
+        response = self.get_response()
+        assert response.status_code == 200
+        response_data = json.loads(response.content.decode('utf-8'))
```
mikedikan: why do you have to call out utf-8 here - is there no way to configure this globally?

course_discovery/urls.py
amangano-edx: I felt like this should be isolated from the rest of the normal api endpoints since it is an edx-specific hack. I wanted to make sure it wasn't available by default.

course_discovery/apps/edx_catalog_extensions/serializers.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        Format and return the query facet data so that it may be properly serialized.
+
+        Re-implements BaseHaystackFacetSerializer.format_query_facet_data so that distinct
+        count information may be included with the data.
+        """
+        query_data = {}
+        for field, options in getattr(self.Meta, 'field_queries', {}).items():  # pylint: disable=no-member
+            # This serializer is only expected to be used when we need to compute distinct counts.
+            # The custom DistinctCountstSearchQuery backend formats query facet counts as a dictionary
```
mikedikan: typo?  DistinctCountstSearchQuery = DistinctCountsSearchQuery

course_discovery/apps/edx_catalog_extensions/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.conf.urls import url
+
+from course_discovery.apps.edx_catalog_extensions.views import DistinctCountsAggregateSearchViewSet
+
+distinct_facets = DistinctCountsAggregateSearchViewSet.as_view({'get': 'facets'})
+
+urlpatterns = [
+    url(r'^v1/search/all/facets', distinct_facets, name='edx-search-all-facets')
```
MatthewPiatetsky: what is going on here?  
search/all/facets is the endpoint regular search uses, is this overriding that endpoint only when this app is enabled?

course_discovery/apps/edx_catalog_extensions/serializers.py
```diff
@@ -0,0 +1,117 @@
...
+    
+    This serializer is only expected to be used by the DistinctCountsAggregateFacetSearchSerializer, when
+    we need to compute distinct hit and facet counts.
+    """
+    distinct_count = serializers.SerializerMethodField()
+
+    def get_distinct_count(self, instance):
+        """Return the distinct count for this facet."""
+
+        # The DistinctCountsElasticsearchBackendWrapper formats field facet counts as a three tuple
```
MatthewPiatetsky: what is DistinctCountsElasticsearchBackendWrapper? I don't see it anywhere in this file

course_discovery/apps/edx_catalog_extensions/serializers.py
```diff
@@ -0,0 +1,117 @@
...
+        for field, options in getattr(self.Meta, 'field_queries', {}).items():  # pylint: disable=no-member
+            # This serializer is only expected to be used when we need to compute distinct counts.
+            # The custom DistinctCountstSearchQuery backend formats query facet counts as a dictionary
+            # with field_name mapping to a two tuple containing normal count and distinct count.
+            counts = query_facet_counts.get(field, (0, 0))
+            if counts[0]:
+                query_data[field] = {
+                    'field': field,
+                    'options': options,
+                    'count': counts[0],
```
mikedikan: i'm not a huge fan of [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) - can you use existing or create a new constant to define these index references?

course_discovery/apps/edx_catalog_extensions/serializers.py
MatthewPiatetsky: why not format things in a Named tuple or dictionary for clarity

course_discovery/apps/edx_catalog_extensions/views.py
```diff
@@ -0,0 +1,61 @@
...
+    AggregateSearchViewSet facets endpoint except that it also include distinct hit and facet counts. The other
+    AggregateSearchViewSet endpoints (list, etc) are not exposed here.
+    """
+
+    # Use DistinctCountsAggregateFacetSearchSerializer so that the distinct counts may be returned with the response.
```
MatthewPiatetsky: i don't think this comment is necessary

course_discovery/apps/edx_catalog_extensions/serializers.py
```diff
@@ -0,0 +1,117 @@
...
+    we need to compute distinct hit and facet counts.
+    """
+    distinct_count = serializers.SerializerMethodField()
+
+    def get_distinct_count(self, instance):
+        """Return the distinct count for this facet."""
+
+        # The DistinctCountsElasticsearchBackendWrapper formats field facet counts as a three tuple
+        # containing the field name, normal count and distinct count.
+        count = instance[2]
```
mikedikan: I appreciate the comment, but i feel like a constant would be better

mikedikan (commented):
A few minor (ignorable) comments - lgtm!

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def run_mlt(self, **kwargs):
+        """
+        Run a more-like-this query.
+
+        The DistinctCountsSearchQuery class is only meant to be used when distinct hit and facet counts are needed. The
+        logic necessary for computing distinct counts has not been implemented for more-like-this queries. Therefore,
+        this method will raise an error if called.
+        """
+        raise NotImplemented('DistinctCountsSearchQuery.run_mlt is not supported.')
```
MatthewPiatetsky: why is this necessary?  
If this is not here we will know a more-like-this query and any other queries are not implemented, right?  
We don't mention anything like this for our regular search endpoint

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+        method will raise an error if called.
+        """
+        raise NotImplemented('DistinctCountsSearchQuery.run_raw is not supported.')
+
+    def set_aggregation_key(self, aggregation_key):
+        """
+        Set the aggregation_key for this Query. The aggregation_key is the field that should
+        be used to group records when computing distinct counts.
+
+        The aggregation_key should be a field that is NOT analyzed by the index (like one of the
```
MatthewPiatetsky: why is this comment necessary here, it seems redundant with the one you had in backends and that place where you actually set it to not_analyzed seems like the appropriate place for the comment

course_discovery/apps/edx_catalog_extensions/urls.py
amangano-edx: No, this is installed at a different root path than the other endpoints. The other endpoints begin at `/api`. This begins at `/api-edx`

course_discovery/apps/edx_catalog_extensions/serializers.py
amangano-edx: Because this is consistent with how the superclass does it, and how the serializers expect it to be formatted. The superclass returns a two-tuple.

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
```diff
@@ -0,0 +1,44 @@
...
+        self._distinct_result_count = None
+
+    def distinct_count(self):
+        """
+        Return the distinct hit count.
+
+        Note: This will cause the query to run if it hasn't already.
+        """
+        if self._distinct_result_count is None:
+            self._distinct_result_count = self.query.get_distinct_count()
```
MatthewPiatetsky: this is called once, why are there 3 definitions of this method?

course_discovery/apps/api/v1/tests/test_views/test_search.py
amangano-edx: I think I copied that from another test.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Note: If this query includes facets, each facet will be converted to an aggregation
+        and the facets clause will be removed.
+        """
+        search_kwargs = self.backend.build_search_kwargs(*args, **kwargs)
+
+        # Setting precision to 1500 should result in a distinct hit count that is close to accurate
+        # when the query returns fewer than 1500 search results. This may be increased (with a performance
+        # penalty) when larger result sets become more common.
+        aggregations = self._build_cardinality_aggregation(precision=1500)
```
MatthewPiatetsky: this should probably be a constant

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+        if search_kwargs.get('facets'):
+            for facet_name, facet_config in search_kwargs['facets'].items():
+                new_facet = self._convert_facet_to_aggregation(facet_config)
+
+                # Setting precision to 1000 should result in a distinct hit count that is close to accurate
+                # when the query returns fewer than 1000 search results. This may be increased (with a performance
+                # penalty) when larger result sets become more common.
+                new_facet['aggs'] = self._build_cardinality_aggregation(precision=1000)
```
MatthewPiatetsky: why is this 1000 here and 1500 above

course_discovery/apps/edx_catalog_extensions/serializers.py
amangano-edx: This is how the superclass does it. Since this is a hack that required re-implementing many superclass methods, I'd rather leave the code as close to the original as possible.

course_discovery/apps/edx_catalog_extensions/views.py
davec-edx: FWIW, as in RESTful calls, the 'endpoint' is the URL.  The 'resource' is the entity that 'actions' are done against.  Sounds like a nit-picky correction but it's good to have a common vocabulary.  Implementation-wise, we should consider subclassing the NotImplemented with a general message that the resource (or action or some phrase) isn't implemented and leave it at that, raising the same exception always.  Returning too much text shows that we have code behind it and could lead to a security issue.  Although, this is open source so maybe that's not as much of a concern.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: It's necessary because this class inherits from the regular Haystact search query class, which implements this method. I do not want this method to be used because I haven't implemented the distinct counts functionality for it. This query class should only be used when we need distinct counts.

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
amangano-edx: I'm not sure which definitions you're referring to. The ones that I can think of are on the DistinctCountsSearchQuerySet and the DistinctCountsSearchQuery. It needs to be defined on both in order for it to be exposed to the end consumer. This is consistent with how other fields like `count` are made available to the consumer. The query object caches the value it receives from the backend and exposes it via a `get_` method. The consumer typically interacts with a SearchQuerySet, rather than a Query object, so the SearchQuerySet exposes a `get_` method, which has to call through to it's associated Query object to get the value.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def search(self, query_string, **kwargs):
+        """
+        Run a search query and return the results.
+
+        Re-implements the ElasticsearchSearchBackend.search method so that the logic necessary for computing
+        and processing distinct hit and facet counts can be added in the appropriate places.
+        """
+        if len(query_string) == 0:
+            return {
```
MatthewPiatetsky: it looks like below you're setting raw_results = {} and letting the process results call happen, why not do the same thing in both places?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def run_raw(self, **kwargs):
+        """
+        Run a raw search query.
+
+        The DistinctCountsSearchQuery class is only meant to be used when distinct hit and facet counts are needed. The
+        logic necessary for computing distinct counts has not been implemented for raw search queries. Therefore, this
+        method will raise an error if called.
+        """
+        raise NotImplemented('DistinctCountsSearchQuery.run_raw is not supported.')
```
MatthewPiatetsky: why is this necessary? you're not trying to run a raw query

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: We want to be able to build the aggregation with different values for precision. Theoretically, the hit count aggregation is more important that the facet counts aggregation (and less impactful on performance), so we'll probably want a higher precision value for that one relative to the facet aggregations.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+        DistinctCountsSearchBackend may be used to execute the query and so that the distinct hit
+        counts may be cached.
+        """
+        # This query class is only meant to be used when we need to compute distinct hit and facet counts.
+        # Therefore, it should always be configured with an aggregation_key.
+        if not self._aggregation_key:
+            raise RuntimeError('DistinctCountsSearchQuery cannot run without aggregation_key.')
+
+        final_query = self.build_query()
+        search_kwargs = self.build_params(spelling_query=spelling_query)
```
MatthewPiatetsky: what is a spelling_query?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Two reason:
1.) IMO (I plan to confirm with product), the hit count aggregation is more important than facet counts aggregation since it is what the user will likely see first on the page.
2.) There is only one hit count aggregation so we can use a higher precision value than we use with the facet aggregations (for which there are many) without incurring too much of a performance hit.

course_discovery/apps/edx_catalog_extensions/serializers.py
MatthewPiatetsky: you could still have a constant called NORMAL_COUNT_INDEX that can make this clear but still use a tuple for consistency

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
MatthewPiatetsky: ok, the reason for using different numbers makes sense, but you can still use two constants.  
HIT_COUNT_PRECISION and FACET_COUNT_PRECISION

course_discovery/apps/edx_catalog_extensions/serializers.py
```diff
@@ -0,0 +1,117 @@
...
+        return serializers.IntegerField(read_only=True).to_representation(count)
+
+
+class DistinctCountsQueryFacetFieldSerializer(QueryFacetFieldSerializer):
```
MatthewPiatetsky: what is the difference between this and DistinctCountsFacetFieldSerializer

course_discovery/apps/edx_catalog_extensions/views.py
```diff
@@ -0,0 +1,61 @@
...
+            - name: selected_query_facets
+              description: Query facets
+              paramType: query
+              allowMultiple: true
+              type: array
+              items:
+                pytype: str
+              required: false
+        """
+        return super(DistinctCountsAggregateSearchViewSet, self).facets(request)
```
MatthewPiatetsky: is this just the same as what the super class has?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+        super(DistinctCountsSearchQuery, self).__init__(**kwargs)
+        self._aggregation_key = None
+        self._distinct_hit_count = None
+
+    def _clone(self, **kwargs):
+        """
+        Create and return a new DistinctCountsSearchQuery with fields set to match those on the original
+        object.
+
+        Overrides ElasticsearchSearchQuery._clone to ensure that _aggregation_key and _distinct_hit_count
```
MatthewPiatetsky: I'm confused by this, this doesn't seem related to cloning, why not add the aggregation key and other field after cloning?

course_discovery/apps/edx_catalog_extensions/serializers.py
amangano-edx: Most of this is re-implemented from the superclass method. I'd prefer to keep it as close to that implementation as possible and only modify the lines necessary for the functionality we're adding.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: It doesn't hurt anything. It could be helpful to someone working in this file but not in the place where it is actually used by the consumer.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Most of this is re-implemented from the superclass method. I'd prefer to keep it as close to that implementation as possible and only modify the lines necessary for the functionality we're adding.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: It's necessary because I don't want people to be able to use this Query object in an unexpected way. I'd prefer it to fail loudly if/when someone uses it like this so that that person sees the issue right away.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: This is part of the superclass method.

course_discovery/apps/edx_catalog_extensions/serializers.py
amangano-edx: This is used for query facets, which are formatted differently from field facets.

course_discovery/apps/edx_catalog_extensions/views.py
amangano-edx: Yes, but I thought it would be good to add documentation.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: The superclass _clone method works by creating a new instance of the Query and then copying all of the existing field values to it. We're doing the same thing here, and making sure that our custom fields also get copied over correctly.

course_discovery/apps/api/v1/tests/test_views/test_search.py
bderusha: https://docs.python.org/3/library/stdtypes.html#bytes.decode

Default is `utf-8`

course_discovery/apps/edx_catalog_extensions/serializers.py
bderusha: I'm okay leaving it if it matches the original implementation, but can we call out the addition in a comment just to make it clear which part is ours?

I know you mention it in the method comment, but I think a quick comment inline would also be beneficial in this case.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
bderusha: For things like this I find it helpful when the comment links to the source of the original method

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
bderusha: Is there any reason to believe that we would want to update this dynamically?  Are we okay releasing when this needs to be changed?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
bderusha: is `aggs` our choice or is this key chosen for us?  If it's our choice prefer the full word `aggregations`; otherwise ignore

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Only the 'field' and 'size' options are supported for conversion. If other options are present, an error
+        will be raised. See the elasticsearch docs for more information on converting terms facets to aggregations.
+        https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-facets-migrating-to-aggs.html#_simple_cases
+        """
+        supported_options = {'field', 'size'}
+        for option, value in facet_config['terms'].items():
+            if option not in supported_options:
+                msg = 'Cannot convert terms facet to aggregation: Unsupported option ({})'.format(option)
+                raise RuntimeError(message)
```
bderusha: `msg`

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def _process_results(self, raw_results, **kwargs):
+        """
+        Construct and return a dictionary of query result information from the raw query results.
+
+        Overrides and re-implements the ElasticsearchSearchBackend._process_results method so that
+        the distinct count information may be extracted and included in the processed results dictionary.
+        """
+        results = self.backend._process_results(raw_results, **kwargs)
+        aggs = raw_results['aggregations']
```
bderusha: NIT: full var name `aggregations`

course_discovery/urls.py
bderusha: I understand the intent, but I wonder if it's necessary (or if it ends up being more confusing) to separate this out.  I don't know that I have a fully formed opinion on it right now, but I get a weird feeling about this.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: `aggs` is an elasticsearch shortcut for `aggregations`. I believe we can use `aggregations` if that's what we prefer.

amangano-edx:
replaced by https://github.com/edx/course-discovery/pull/725 and https://github.com/edx/course-discovery/pull/726

PR #705, opened 2017-03-01
ECOM-7338 Deduplicate course runs in typeahead by course

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -473,6 +473,26 @@ class TypeaheadSearchViewTests(DefaultPartnerMixin, TypeaheadSerializationMixin,
         self.assertEqual(len(response_data['course_runs']), RESULT_COUNT)
         self.assertEqual(len(response_data['programs']), RESULT_COUNT)
 
+    def test_typeahead_deduplicate_course_runs(self):
+        """ Verify the typeahead response will only include the first course run per course. """
+        RESULT_COUNT = TypeaheadSearchView.RESULT_COUNT
+        title = "Test"
+        course1 = CourseFactory(partner=self.partner)
+        course2 = CourseFactory(partner=self.partner)
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="{}{}{}".format(title, course1.title, i), course=course1)
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="{}{}{}".format(title, course2.title, i), course=course2)
+        response = self.get_response({'q': title})
+        assert response.status_code == 200
+        response_data = response.json()
+
+        # There are many runs for both courses, but only one from each will be included
+        course_runs = response_data['course_runs']
+        assert len(course_runs) == 2
+        # compare course titles embedded in course run title to ensure that course runs belong to different courses
+        assert course_runs[0]['title'][4:-1] != course_runs[1]['title'][4:-1]
+
     def test_typeahead_multiple_authoring_organizations(self):
         """ Test typeahead response with multiple authoring organizations. """
         title = "Design"
--- a/course_discovery/apps/api/v1/views/search.py
+++ b/course_discovery/apps/api/v1/views/search.py
@@ -134,7 +134,17 @@ class TypeaheadSearchView(PartnerMixin, APIView):
             SQ(authoring_organizations_autocomplete=clean_query)
         )
         course_runs = course_runs.filter(published=True).exclude(hidden=True).filter(partner=partner.short_code)
-        course_runs = course_runs[:self.RESULT_COUNT]
+
+        # Get first three results after deduplicating by course run
+        course_runs_list, course_keys_set = [], set()
+        for course_run in course_runs:
+            course_key = course_run.course_key
+            if course_key in course_keys_set:
+                continue
+            course_keys_set.add(course_key)
+            course_runs_list.append(course_run)
+            if len(course_runs_list) == 3:
+                break
 
         programs = sqs.models(Program).filter(
             SQ(title_autocomplete=clean_query) |
@@ -143,7 +153,7 @@ class TypeaheadSearchView(PartnerMixin, APIView):
         programs = programs.filter(status=ProgramStatus.Active).filter(partner=partner.short_code)
         programs = programs[:self.RESULT_COUNT]
 
-        return course_runs, programs
+        return course_runs_list, programs
 
     def get(self, request, *args, **kwargs):
         """


course_discovery/apps/api/v1/views/search.py
```diff
@@ -134,7 +134,16 @@ def get_results(self, query, partner):
             SQ(authoring_organizations_autocomplete=clean_query)
         )
         course_runs = course_runs.filter(published=True).exclude(hidden=True).filter(partner=partner.short_code)
-        course_runs = course_runs[:self.RESULT_COUNT]
+
+        # Get first three results after deduplicating by course run
+        course_runs_list, course_keys_list = [], set()
```
amangano-edx: It's a little confusing for the set to be named '_list'. Maybe use '_set' instead?

course_discovery/apps/api/v1/views/search.py
```diff
@@ -134,7 +134,16 @@ def get_results(self, query, partner):
...
+        for course_run in course_runs:
+            course_key = course_run.course_key
+            if course_key not in course_keys_list:
```
amangano-edx: You could save a layer of nesting by using `continue` if the course_key is already in the list instead of using `not in`.

```python
if course_key in course_key_list:
    continue
```

course_discovery/apps/api/v1/views/search.py
```diff
@@ -134,7 +134,17 @@ def get_results(self, query, partner):
             SQ(authoring_organizations_autocomplete=clean_query)
         )
         course_runs = course_runs.filter(published=True).exclude(hidden=True).filter(partner=partner.short_code)
-        course_runs = course_runs[:self.RESULT_COUNT]
+
+        # Get first three results after deduplicating by course run
+        course_runs_list, course_keys_set = [], set()
+        for course_run in course_runs:
+            course_key = course_run.course_key
```
amangano-edx, line 141: I wish this could be done by aggregation_key so that it can be consistent with how the other endpoints will be deduplicating course_runs, but I suppose this is fine.

MatthewPiatetsky: We're not using an aggregation query here, and we use course_key everywhere else, so that seems more appropriate to me.  
Especially if it is called aggregation_key I would rather not expand the scope of it beyond using it for aggregations (and course_key is more understandable in this case)

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -456,6 +456,23 @@ def test_typeahead_multiple_results(self):
...
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="{}{}".format(title, i), course=course1)
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="{}{}".format(title, i), course=course2)
+        response = self.get_response({'q': title})
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+
+        # There are many runs for both courses, but only one from each will be included
+        self.assertEqual(len(response_data['course_runs']), 2)
```
amangano-edx: We should use pytest-style assertions.

Also, it would probably be good to verify that the two records we get back have different course keys.

course_discovery/apps/api/v1/views/search.py
amangano-edx: It is going to be used for deduplication on the client side. Since it is what is being used for computing the counts it makes sense that it would also be used for deduplicating records.

MatthewPiatetsky: sure but it makes sense there because you're using an aggregation query

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: changed

course_discovery/apps/api/v1/views/search.py
amangano-edx: What if we want to dedup by more than just course key at some point? By using aggregation_key, the logic can be isolated there and we won't need to worry about tracking down all the other places we did deduplication in other ways.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -456,6 +456,26 @@ def test_typeahead_multiple_results(self):
...
+        RESULT_COUNT = TypeaheadSearchView.RESULT_COUNT
+        title = "Test"
+        course1 = CourseFactory(partner=self.partner)
+        course2 = CourseFactory(partner=self.partner)
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="{}{}{}".format(title, course1.title, i), course=course1)
+        for i in range(RESULT_COUNT):
+            CourseRunFactory(title="{}{}{}".format(title, course2.title, i), course=course2)
+        response = self.get_response({'q': title})
+        self.assertEqual(response.status_code, 200)
```
amangano-edx: We should use a pytest-style assertion here.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -456,6 +456,26 @@ def test_typeahead_multiple_results(self):
...
+        response_data = response.json()
+
+        # There are many runs for both courses, but only one from each will be included
+        course_runs = response_data['course_runs']
+        assert len(course_runs) == 2
+        # compare course titles embedded in course run title to ensure that course runs belong to different courses
+        assert course_runs[0]['title'][4:-1] != course_runs[1]['title'][4:-1]
```
amangano-edx, line 494: Wouldn't both of these substrings also include the value of `i` from lines 465 and 467? Couldn't they still be different because of that even if the courses are the same?

Why not just compare the course keys?

course_discovery/apps/api/v1/views/search.py
amangano-edx: For the record, I'm fine with using course key if that's how you want to do it. But I think there could be value in using aggregation_key if available.

course_discovery/apps/api/v1/tests/test_views/test_search.py
MatthewPiatetsky: that's why i'm using the -1 here.  
Alternatively I could use the opaque keys library to try to parse out the course key but this seems simpler

amangano-edx: I thought -1 when used like this includes the last character of the string?

amangano-edx: NVM, looks like it doesn't include the last character.

MatthewPiatetsky: if you use array indexing then [-1] will select just the last character  
with slicing you have [first character (default first) : last character (default last) : step ]

course_discovery/apps/api/v1/views/search.py
```diff
@@ -134,7 +134,17 @@ def get_results(self, query, partner):
             SQ(authoring_organizations_autocomplete=clean_query)
         )
         course_runs = course_runs.filter(published=True).exclude(hidden=True).filter(partner=partner.short_code)
-        course_runs = course_runs[:self.RESULT_COUNT]
+
+        # Get first three results after deduplicating by course run
+        course_runs_list, course_keys_set = [], set()
+        for course_run in course_runs:
```
amangano-edx, line 140: Will this scale? I believe this will result in us loading every matching course run with each typeahead request, whereas before we were only loading RESULT_COUNT records.

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: personally i'd rather use course_key here, I'm always hesitant to try to design things by trying to do what we might need in the future, because then we end up assuming something already works even if it is not doing exactly what we want

course_discovery/apps/api/v1/views/search.py
MatthewPiatetsky: hmm, I ran some short queries a bunch of times both on this branch and on master typeahead requests take me 200ms +- 50ms (after the first one which takes longer on both for some reason)
Regardless, I don't think it should be an issue. 
I think typeahead gets about 2 rps which isn't a lot, and the bottleneck should be ES not python runtime or memory.

amangano-edx: From looking at the SearchQuerySet code, it doesn't seem like it loads everything into memory all at once. It requests and caches a single page of results at a time as it iterates. If you break before it has iterated through all the results from the first page, it would have only loaded a single page of records. See https://github.com/django-haystack/django-haystack/blob/master/haystack/query.py#L130

amangano-edx:
LGTM

PR #706, opened 2017-03-02
lms_course_id set as required=False for django-admin

awais786:
I did't make any change at the model level. Instead I have updated the django-admin form and make this field optional there.

[ECOM-6931](https://openedx.atlassian.net/browse/ECOM-6931)


awais786:
screen shot 2017-03-01 at 10 10 50 pm

awais786:
@waheedahmed @tasawernawaz Please review this PR.

awais786:
We don't need to override the clean method in admin. Because now without data form will not submit.

course_discovery/apps/publisher/models.py
```diff
@@ -189,7 +189,7 @@ class CourseRun(TimeStampedModel, ChangedByMixin):
     )
 
     course = models.ForeignKey(Course, related_name='publisher_course_runs')
-    lms_course_id = models.CharField(max_length=255, unique=True, null=True, blank=True)
+    lms_course_id = models.CharField(max_length=255, unique=True, null=True)
```
tasawernawaz: I think this will effect course/course_run edit pages as well, we will not able to submit form with providing value of `lms_course_id`. Can you confirm ?

awais786: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/forms.py#L237

Its not required in forms.

course_discovery/apps/publisher/forms.py
```diff
@@ -384,3 +384,21 @@ def clean(self):
             )
 
         return self.cleaned_data
+
+
+class CourseRunAdminForm(forms.ModelForm):
+    lms_course_id = forms.CharField(required=True)
```
tasawernawaz: @awais786 No need to make `lms_course_id` required, I think only we need to change it to `None` if `lms_course_id` is not provided because we don't want to save it as empty string.

@bderusha You thoughts ?

awais786: @tasawernawaz in that we need to override the clean method only.

tasawernawaz: @awais786 Yes that's enough I think.

bderusha: Per our conversation offline, please make `required=False`

bderusha (changes requested):

awais786:
@waheedahmed Please review.

waheedahmed (approved):

awais786:
@bderusha please review.

awais786:
@bderusha please review.

awais786:
@bderusha please review.

bderusha (approved):

PR #708, opened 2017-03-02
Preview available email.

waheedahmed:
[ECOM-6084](https://openedx.atlassian.net/browse/ECOM-6084)

As a Publisher tool owner, I'd like an email template created which will be used to send emails to Course Team when a Preview link has been added by the Publisher.

screen shot 2017-03-02 at 5 09 42 pm


waheedahmed:

course_discovery/apps/publisher/api/serializers.py
```diff
@@ -87,6 +87,7 @@ def validate(self, data):
 
         return validated_values
 
+    @transaction.atomic
```
bderusha, line 90: Can you comment (here, not in the code) as to why this was added.  Just curious

course_discovery/templates/publisher/email/course_run/preview_available.html
```diff
@@ -4,17 +4,17 @@
...
         {% blocktrans with link_start='<a href="' link_middle='">' link_end='</a>' trimmed %}
-            Preview for {{ link_start }}{{ page_url }}{{ link_middle }} {{ course_name }} {{ link_end }} is available for review.
+            A preview is now available for the {{ link_start }}{{ page_url }}{{ link_middle }} {{ course_run_number }} course run {{ link_end }} of {{ course_name }}. Visit {{ link_start }}{{ preview_link }}{{ link_middle }} link to preview {{ link_end }} to approve or decline the preview for this course run.
         {% endblocktrans %}
     </p>
 
     {% comment %}Translators: It's closing of mail.{% endcomment %}
     {% trans "Thanks," %}<br>
-    {% trans "The edX team" %}
+    edX {{ sender_role }}
```
bderusha: Same note here as the other PR. Please use dynamic site name here

bderusha (changes requested):

course_discovery/apps/publisher/api/serializers.py
waheedahmed: @bderusha This was added if some error occurred during sending email, then user will not get an email notification and the user who is updating course_key or preview_url will think he has done his part but the course team user did not get any email, so he does not know a preview is available or studio instance is created for the course.

So I'm rolling back the whole transaction if some error occurred to let user know that there was an error during updating course_key or preview_url.

course_discovery/templates/publisher/email/course_run/preview_available.html
waheedahmed: @bderusha Yeah it should be, but not mentioned in any ticket.

awais786 (approved):

course_discovery/templates/publisher/email/course_run/preview_available.html
waheedahmed: I think we should create a ticket to fix it every where in publisher.

waheedahmed: Done.

waheedahmed:
@bderusha Feedback addressed, please take another look.

bderusha (approved):

PR #709, opened 2017-03-02
Added published email content.

waheedahmed:
[ECOM-6083](https://openedx.atlassian.net/browse/ECOM-6083)

As a Publisher tool owner, I'd like an email template created which will be used to send emails to Course Team when a Course-run has been published.

screen shot 2017-03-02 at 6 37 29 pm


course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -681,6 +682,8 @@ def test_course_published(self):
         self.run_state.name = CourseRunStateChoices.Approved
         self.run_state.preview_accepted = True
         self.run_state.save()
+        self.course_run.lms_course_id = 'course-v1:testX+test45+2017T2'
```
bderusha: Will it break other test cases to add lms_course_id in the setup method?

course_discovery/templates/publisher/email/course_run/published.html
```diff
@@ -8,13 +8,13 @@
...
         {% blocktrans with link_start='<a href="' link_middle='">' link_end='</a>' trimmed %}
-            Your course {{ link_start }}{{ page_url }}{{ link_middle }} {{ course_name }} {{ link_end }} is now live.
+            The {{ link_start }}{{ page_url }}{{ link_middle }} {{ course_run_number }} course run {{ link_end }} of {{ course_name }} has been published. Visit {{ link_start }}{{ preview_url }}{{ link_middle }}link to live course run page{{ link_end }} to view the course run.
         {% endblocktrans %}
     </p>
 
     {% comment %}Translators: It's closing of mail.{% endcomment %}
     {% trans "Thanks," %}<br>
-    {% trans "The edX team" %}
+    edX {{ sender_role }}
```
bderusha: Just noticing this now, but shouldn't `edX` also be dynamic for platforms which are not edX?

course_discovery/apps/publisher/api/tests/test_views.py
waheedahmed: @bderusha No, its only being used in published email to extract the run from it.

course_discovery/templates/publisher/email/course_run/published.html
waheedahmed: @bderusha Yeah it should be, but not mentioned in any ticket.

waheedahmed:
@bderusha Feedback addressed, please take another look.

tasawernawaz (approved):
LGTM

course_discovery/apps/publisher/api/tests/test_views.py
bderusha: Then, can we move this into the setup of the course_run?

bderusha (approved):

course_discovery/apps/publisher/api/tests/test_views.py
waheedahmed: Actually I just realized its already added in setup recently, so we don't need it here now.

PR #711, opened 2017-03-03
Awais786/ecom 7231 sort dropdowns bug

awais786:
Fixing ordering using JS.

### DIFF
--- a/course_discovery/apps/publisher/api/views.py
+++ b/course_discovery/apps/publisher/api/views.py
@@ -1,3 +1,5 @@
+from django.db.models import CharField
+from django.db.models.functions import Concat, Lower, Value
 from rest_framework.generics import ListAPIView, RetrieveAPIView, UpdateAPIView, get_object_or_404
 from rest_framework.permissions import IsAuthenticated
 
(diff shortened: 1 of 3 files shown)


course_discovery/apps/publisher/api/views.py
```diff
@@ -23,7 +25,12 @@ class OrganizationGroupUserView(ListAPIView):
 
     def get_queryset(self):
         org_extension = get_object_or_404(OrganizationExtension, organization=self.kwargs.get('pk'))
-        queryset = User.objects.filter(groups__name=org_extension.group).order_by('full_name', 'username')
+        queryset = User.objects.filter(groups__name=org_extension.group).annotate(
+            display_name=Concat(
+                Lower('full_name'), Value(' '), Lower('username'), output_field=CharField()
+            )
+        ).order_by('display_name')
```
bderusha, line 32: Why can't we just `order_by('full_name', 'username')`?  Shouldn't the original code do what we want?

course_discovery/apps/publisher/forms.py
```diff
@@ -150,7 +151,11 @@ def __init__(self, *args, **kwargs):
             org_extension = OrganizationExtension.objects.get(organization=organization)
             self.declared_fields['team_admin'].queryset = User.objects.filter(
                 groups__name=org_extension.group
-            ).order_by('full_name', 'username')
```
bderusha, line 153: what was wrong with this?

course_discovery/apps/publisher/api/views.py
awais786: Yes I am considering to go back. I did this because if full_name is '' then it disturbs order.
I am going back to standard query.

awais786:
@bderusha I think we don't need to fix this issue at the moment. If user will give the valid full_name then this issue will not appear.

bderusha:
Sounds good

PR #712, opened 2017-03-03
Update remove index approach to match AWS

mikedikan:
ECOM-7297

AWS ES doesn't support the settings request. We need to update the way we fetch the indexes to use something else that is supported.

@edx/ecommerce & @feanil for review

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
@@ -66,7 +66,10 @@ class Command(BaseCommand):
         Returns:
             sorted_indexes_by_timestamp (list): The sorted listing of index names
         """
-        all_index_settings = indices_client.get_settings()
-        all_indexes = list(all_index_settings.keys())
+        # Elasticsearch in AWS is not a full implementation of ES, and we need to use the (more verbose) status
+        # endpoint instead of the (more succinct) settings endpoint. For more information, see
+        # http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es_version_1_5.html
+        all_index_status = indices_client.status()
+        all_indexes = list(all_index_status['indices'].keys())
         all_current_indexes = [index_name for index_name in all_indexes if index_name.startswith(index_prefix + '_')]
         return sorted(all_current_indexes)


course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -66,7 +66,7 @@ def get_indexes_sorted_by_timestamp(self, indices_client, index_prefix):
         Returns:
             sorted_indexes_by_timestamp (list): The sorted listing of index names
         """
-        all_index_settings = indices_client.get_settings()
-        all_indexes = list(all_index_settings.keys())
+        all_index_status = indices_client.status()
```
rlucioni: Could you add a comment explaining why we're using `.status()`? Something like "Amazon's Elasticsearch service doesn't support `.get_settings()`."

rlucioni (approved):

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
rlucioni: Extra credit for a link to AWS or Haystack documentation explaining why this is necessary.

mikedikan: 👍

course_discovery/apps/edx_haystack_extensions/management/commands/remove_unused_indexes.py
```diff
@@ -66,7 +66,9 @@ def get_indexes_sorted_by_timestamp(self, indices_client, index_prefix):
         Returns:
             sorted_indexes_by_timestamp (list): The sorted listing of index names
         """
-        all_index_settings = indices_client.get_settings()
-        all_indexes = list(all_index_settings.keys())
+        # Elastic Search in AWS is not a full implementation of ES, and we need to use the (more verbose) status
```
rlucioni: Looks good. (s/Elastic Search/Elasticsearch)

feanil (approved):
Verified that the endpoint works from a discovery machine in stage.

mikedikan:
Thanks @feanil for checking!

PR #716, opened 2017-03-06
Remove upgradeable filter from serializer  

MatthewPiatetsky:
After some conversations with Jasper and Alasdair I've uncovered some use cases that were not 100% clear previously that may require regrooming/replanning. We will most likely still need the upgradeable filter I added in the previous PR, but I am removing the changes to actually use the filter until we resolve the open questions regarding the story

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -542,7 +542,7 @@ class CourseWithProgramsSerializer(CourseSerializer):
 
         if self.context.get('marketable_enrollable_course_runs_with_archived'):
             # Same as "marketable_course_runs_only", but includes courses with an end date in the past
-            course_runs = course_runs.marketable().enrollable().upgradeable()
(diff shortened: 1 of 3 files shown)


rlucioni (approved):
@MatthewPiatetsky can you include a comment on the `upgradeable()` filter itself so that others understand why it's there but unused? Extra credit for including a ticket number.

clintonb (changes requested):
Remove the filter completely if it's not used. If it needs to be restored in the future, it's a matter of reverting. We should not have dead code in the repository.

MatthewPiatetsky:
@clintonb I could remove this, but the filter has functionality in itself without being used in the serializer. You can still pass it as a query parameter to determine which runs are upgradeable.  
Does that portion of this work not deliver independent value?  
(Also, I think there are other filters in query.py not used in serializers at the moment)

clintonb:
No dead code. If we aren't using it, get rid of it.

PR #722, opened 2017-03-08
ECOM-7346 Turn off index updates on model save

MatthewPiatetsky:
Tested this by running refresh with es off before changes (which gave indexing errors) and after changes (which did not give errors)

plan was to merge this today and then revert today after running refresh_course_metadata

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -333,13 +333,14 @@ class Course(TimeStampedModel):
         ids = [result.pk for result in results]
         return cls.objects.filter(pk__in=ids)
 
-    def save(self, *args, **kwargs):
-        super(Course, self).save(*args, **kwargs)
-        try:
-            self.reindex_course_runs()
-        except Exception:  # pylint: disable=broad-except
-            logger.exception("An error occurred while attempting to reindex the course runs"
-                             "of Course with key: [{key}].".format(key=self.key))
+    # This is commented out for a temporary experiment for ECOM-7346
+    # def save(self, *args, **kwargs):
+    #     super(Course, self).save(*args, **kwargs)
+    #     try:
+    #         self.reindex_course_runs()
+    #     except Exception:  # pylint: disable=broad-except
+    #         logger.exception("An error occurred while attempting to reindex the course runs"
+    #                          "of Course with key: [{key}].".format(key=self.key))
 
     def reindex_course_runs(self):
         index = connections['default'].get_unified_index().get_index(CourseRun)
@@ -938,7 +939,8 @@ class Program(TimeStampedModel):
                 publisher.publish_program(self)
         else:
             super(Program, self).save(*args, **kwargs)
-        self.reindex_courses()
+        # This is commented out for a temporary experiment for ECOM-7346
+        # self.reindex_courses()
 
     def reindex_courses(self):
         try:
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -3,6 +3,7 @@ import datetime
 from urllib.parse import urlencode
 
 import ddt
+import pytest
 import pytz
 from django.test import TestCase
 from haystack.query import SearchQuerySet
@@ -1191,6 +1192,7 @@ class CourseSearchSerializerTests(TestCase):
 
 
 class CourseRunSearchSerializerTests(TestCase):
+    @pytest.mark.skip(reason="turning off temporarily for experiment - ECOM-7346")
     def test_data(self):
         course_run = CourseRunFactory(transcript_languages=LanguageTag.objects.filter(code__in=['en-us', 'zh-cn']),
                                       authoring_organizations=[OrganizationFactory()])
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -3,6 +3,7 @@ import json
 import urllib.parse
 
 import ddt
+import pytest
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from haystack.query import SearchQuerySet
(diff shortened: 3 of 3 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -335,11 +336,13 @@ def search(cls, query):
 
     def save(self, *args, **kwargs):
         super(Course, self).save(*args, **kwargs)
-        try:
-            self.reindex_course_runs()
-        except Exception:  # pylint: disable=broad-except
-            logger.exception("An error occurred while attempting to reindex the course runs"
-                             "of Course with key: [{key}].".format(key=self.key))
+        # only enable index updates in test environment for temporary expirement
```
rlucioni: s/expirement/experiment. Also, it would be a good idea to include the JIRA ticket number.

course_discovery/apps/course_metadata/models.py
```diff
@@ -938,7 +941,9 @@ def save(self, *args, **kwargs):
                 publisher.publish_program(self)
         else:
             super(Program, self).save(*args, **kwargs)
-        self.reindex_courses()
+        # only enable index updates in test environment for temporary expirement
+        if settings.HAYSTACK_SIGNAL_PROCESSOR == 'haystack.signals.RealtimeSignalProcessor':
```
rlucioni: It bothers me seeing the same string hardcoded twice. Even though this is meant to be an experiment, I would still use a variable for the path to the `RealtimeSignalProcessor`.

rlucioni (commented):
@MatthewPiatetsky how temporary are we talking here? Is the plan to deploy this, run the refresh, then revert this change?

You may need to tell coverage to ignore this method in order to merge.

MatthewPiatetsky:
@rlucioni  
yeah the plan was to deploy see if this fixes the issue and then revert
how would I tell codecov to ignore this section?

rlucioni:
@MatthewPiatetsky http://coverage.readthedocs.io/en/coverage-4.2/excluding.html

codecov.yml
```diff
@@ -7,3 +7,5 @@ coverage:
     project:
       default:
         target: 95
+ignore:
+  - "apps/course_metadata/models.py"  # ignore this file
```
rlucioni: @MatthewPiatetsky don't do this. Use coverage's `pragma: no cover`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -335,11 +338,13 @@ def search(cls, query):
 
     def save(self, *args, **kwargs):
         super(Course, self).save(*args, **kwargs)
-        try:
-            self.reindex_course_runs()
-        except Exception:  # pylint: disable=broad-except
-            logger.exception("An error occurred while attempting to reindex the course runs"
-                             "of Course with key: [{key}].".format(key=self.key))
+        # only enable index updates in test environment for temporary experiment
```
rlucioni: It would be a good idea to include the JIRA ticket number here.

course_discovery/apps/course_metadata/models.py
```diff
@@ -335,11 +338,13 @@ def search(cls, query):
...
+        if settings.HAYSTACK_SIGNAL_PROCESSOR == REALTIME_SIGNAL_PROCESSOR:
```
clintonb: This is odd, and unnecessary. Just comment out the `save()` method.

clintonb (changes requested):

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: then a bunch of tests will fail. they depend on this code

codecov.yml
MatthewPiatetsky: didn't know this, thanks, changed

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: added

course_discovery/apps/course_metadata/models.py
clintonb: Skip the tests.

MatthewPiatetsky: how do I do that?

clintonb:
Alternate approach: instead of putting this on master, build an AMI from a branch and deploy it manually.

course_discovery/apps/course_metadata/models.py
rlucioni: http://lmgtfy.com/?q=pytest+skip+test

MatthewPiatetsky:
@rlucioni do you know how to do that?

rlucioni:
@MatthewPiatetsky yes, but I would suggest going the test skipping route or leaving what you have now in place (checking the signal processor) instead of trying an unorthodox release.

course_discovery/apps/course_metadata/models.py
```diff
@@ -30,6 +31,8 @@
 
 logger = logging.getLogger(__name__)
 
+REALTIME_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
```
clintonb: Delete.

course_discovery/apps/course_metadata/models.py
```diff
@@ -333,13 +336,13 @@ def search(cls, query):
...
         return cls.objects.filter(pk__in=ids)
 
-    def save(self, *args, **kwargs):
-        super(Course, self).save(*args, **kwargs)
-        try:
-            self.reindex_course_runs()
-        except Exception:  # pylint: disable=broad-except
-            logger.exception("An error occurred while attempting to reindex the course runs"
-                             "of Course with key: [{key}].".format(key=self.key))
+    # def save(self, *args, **kwargs):
```
clintonb: Add a comment, explaining why this is commented out.

clintonb (approved):

rlucioni (approved):

PR #725, opened 2017-03-08
ECOM 6815 Add ability to compute distinct hit and facet counts

amangano-edx:
This PR adds support for running elasticsearch cardinality aggregations along with search queries so that distinct hit and facet counts may be computed.

Replaces portions of https://github.com/edx/course-discovery/pull/704

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
+++ b/course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
@@ -1,0 +1,303 @@
+import elasticsearch
+
+from django.conf import settings
+from haystack.backends.elasticsearch_backend import ElasticsearchSearchQuery
+from haystack.models import SearchResult
+
+
+class DistinctCountsSearchQuery(ElasticsearchSearchQuery):
+    """ Custom Haystack Query class that computes and caches distinct hit and facet counts for a query."""
+
+    def __init__(self, **kwargs):
+        """
+        Create and return a new instance of DistinctCountsSearchQuery.
+
+        Overrides BaseSearchQuery.__init__ from:
+        https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/__init__.py#L443
+        """
+        super(DistinctCountsSearchQuery, self).__init__(**kwargs)
+        self.aggregation_key = None
+        self._distinct_hit_count = None
+
+    def _clone(self, **kwargs):
+        """
+        Create and return a new DistinctCountsSearchQuery with fields set to match those on the original object.
+
+        Overrides BaseSearchQuery._clone from:
+        https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/__init__.py#L981
+        """
+        clone = super(DistinctCountsSearchQuery, self)._clone(**kwargs)
+        if isinstance(clone, DistinctCountsSearchQuery):
+            clone.aggregation_key = self.aggregation_key
+            clone._distinct_hit_count = self._distinct_hit_count  # pylint: disable=protected-access
+        return clone
+
+    def get_distinct_count(self):
+        """
+        Return the distinct hit count for this query. Calling this method will cause the query to execute if
+        it hasn't already been run.
+        """
+        if self._distinct_hit_count is None:
+            self.run()
+        return self._distinct_hit_count
+
+    def run(self, spelling_query=None, **kwargs):
+        """
+        Run the query and cache the results.
+
+        Overrides and re-implements ElasticsearchSearchQuery.run from:
+        https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/elasticsearch_backend.py#L941
+        """
+        # Make sure that the Query is valid before running it.
+        self.validate()
+
+        final_query = self.build_query()
+        search_kwargs = self.build_params(spelling_query)
+
+        if kwargs:
+            search_kwargs.update(kwargs)
+
+        # Use the DistinctCountsElasticsearchBackendWrapper to execute the query so that distinct hit and query
+        # counts may be computed.
+        backend = DistinctCountsElasticsearchBackendWrapper(self.backend, self.aggregation_key)
+        results = backend.search(final_query, **search_kwargs)
+
+        self._results = results.get('results', [])
+        self._hit_count = results.get('hits', 0)
+        self._distinct_hit_count = results.get('distinct_hits', 0)
+        self._facet_counts = self.post_process_facets(results)
+        self._spelling_suggestion = results.get('spelling_suggestion', None)
+
+    def validate(self):
+        """ Verify that all Query options are valid and supported by this custom Query class."""
+        if self._more_like_this:
+            raise RuntimeError('DistinctCountsSearchQuery does not support more_like_this queries.')
+
+        if self._raw_query:
+            raise RuntimeError('DistinctCountsSearchQuery does not support raw queries.')
+
+        if self.date_facets:
+            raise RuntimeError('DistinctCountsSearchQuery does not support date facets.')
+
+        if self.facets:
+            for field, options in self.facets.items():
+                self._validate_field_facet_options(field, options)
+
+        if self.aggregation_key is None:
+            raise RuntimeError('aggregation_key is required.')
+
+    def _validate_field_facet_options(self, field, options):
+        """ Verify that the provided field facet options are valid and can be converted to an aggregation."""
+        supported_options = DistinctCountsElasticsearchBackendWrapper.SUPPORTED_FIELD_FACET_OPTIONS
+        for option, __ in options.items():
+            if option not in supported_options:
+                msg = (
+                    'DistinctCountsSearchQuery only supports a limited set of field facet options.'
+                    'Field: {field}, Supported Options: ({supported}), Provided Options: ({provided})'
+                ).format(field=field, supported=','.join(supported_options), provided=','.join(options.keys()))
+                raise RuntimeError(msg)
+
+    def more_like_this(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """ Raise an exception since we do not currently want/need to support more_like_this queries."""
+        raise RuntimeError('DistinctCountsSearchQuery does not support more_like_this queries.')
+
+    def run_mlt(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """ Raise an exception since we do not currently want/need to support more_like_this queries."""
+        raise RuntimeError('DistinctCountsSearchQuery does not support more_like_this queries.')
+
+    def raw_search(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """ Raise an exception since we do not currently want/need to support raw queries."""
+        raise RuntimeError('DistinctCountsSearchQuery does not support raw queries.')
+
+    def run_raw(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """ Raise an exception since we do not currently want/need to support raw queries."""
+        raise RuntimeError('DistinctCountsSearchQuery does not support raw queries.')
+
+    def add_date_facet(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """ Raise an exception since we do not currently want/need to support date facets."""
+        raise RuntimeError('DistinctCountsSearchQuery does not support date facets.')
+
+    def add_field_facet(self, field, **options):
+        """
+        Add a field facet to the Query. Raise an error if any unsupported options are provided.
+
+        Overrides BaseSearchQuery.add_field_facet from:
+        https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/__init__.py#L897
+        """
+        self._validate_field_facet_options(field, options)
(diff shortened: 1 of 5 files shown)


course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def _clone(self, **kwargs):
+        """ Create and return a new DistinctCountsSearchQuery with fields set to match those on the original object."""
+        clone = super(DistinctCountsSearchQuery, self)._clone(**kwargs)
+        if isinstance(clone, DistinctCountsSearchQuery):
+            clone._aggregation_key = self._aggregation_key  # pylint: disable=protected-access
+            clone._distinct_hit_count = self._distinct_hit_count  # pylint: disable=protected-access
+        return clone
+
+    def set_aggregation_key(self, aggregation_key):
```
rlucioni: Use of getters and setters is discouraged in Python. Use a property instead.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        Set the aggregation_key for this Query. The aggregation_key is the field that should
+        be used to group records when computing distinct counts.
+        """
+        self._aggregation_key = aggregation_key
+
+    def get_distinct_count(self):
```
rlucioni: Similar to the above, use of getters and setters is discouraged in Python. You can use a `distinct_hit_count` property instead.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+            search_kwargs.update(kwargs)
+
+        # Use the DistinctCountsElasticsearchBackendWrapper to execute the query so that distinct hit and query
+        # counts may be computed.
+        backend = DistinctCountsElasticsearchBackendWrapper(self.backend, self._aggregation_key)
+        results = backend.search(final_query, **search_kwargs)
+
+        self._results = results.get('results', [])
+        self._hit_count = results.get('hits', 0)
+        self._distinct_hit_count = results.get('distinct_hits', 0)
```
rlucioni, line 67: In what cases would these keys be missing from the results dict?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if self._aggregation_key is None:
+            raise RuntimeError('aggregation_key is required.')
+
+    def _validate_field_facet_options(self, field, options):
+        """ Verify that the provided field facet options are valid and can be converted to an aggregation."""
+        supported_options = DistinctCountsElasticsearchBackendWrapper.SUPPORTED_FIELD_FACET_OPTIONS
+        for option, _ in options.items():
+            if option not in supported_options:
+                msg = 'DistinctCountsSearchQuery only supports a limited set of field facet options.'
```
rlucioni: Use parentheses when you need to join together a long string broken across multiple lines:

```py
>>> msg = (
    'this is the first line. '
    'this is the second line.'
)
>>> msg
'this is the first line. this is the second line.'
```

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+    SUPPORTED_FIELD_FACET_OPTIONS = {'size'}
+
+    # The default size for field facet aggregations. This is the same value used by haystack.
+    DEFAULT_FIELD_FACET_SIZE = 100
+
+    def __init__(self, backend, aggregation_key):
+        """
+        Initialize a new instance of the DistinctCountsElasticsearchBackendWrapper.
+
+        -- Parameters --
```
rlucioni: You'll see this written as follows throughout other edX code, including this repo.
```
Arguments:
    backend (ElasticsearchSearchBackend): ...
    aggregation_key (string): ...
```
I'd stick to that format for consistency.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # Precision settings for the elasticsearch cardinality aggregations. We use a higher value for the distinct hit
+    # precision for two reasons:
+    #   1.) The hit count for the current query is more relevant to users than the facet counts.
+    #   2.) The performance penalty is lower for distinct hit count precision than it is for facet count precision,
+    #       since the distinct hit count requires only a single aggregation.
+    # See the elasticsearch docs for more information about the cardinality aggregation and precision settings.
+    # https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-aggregations-metrics-cardinality-aggregation.html
+    DISTINCT_HIT_PRECISION = 1500
+    DISTINCT_FACET_PRECISION = 250
```
rlucioni: These seem like they belong in the settings module.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+            search backend and will result in records being grouped by substrings of the aggregation_key field.
+        """
+        self.backend = backend
+        self.aggregation_key = aggregation_key
+        self.aggregation_name = 'distinct_{}'.format(aggregation_key)
+
+    def search(self, query_string, **kwargs):
+        """ Run a search query and return the results."""
+        if len(query_string) == 0:
+            return {'results': [], 'hits': 0, 'distinct_hits': 0}
```
rlucioni: What do we gain from this optimization?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+                body=search_kwargs,
+                index=self.backend.index_name,
+                doc_type='modelresult',
+                _source=True
+            )
+        except elasticsearch.TransportError as e:
+            if not self.backend.silently_fail:
+                raise
+
+            self.backend.log.error('Failed to query Elasticsearch using "%s": %s', query_string, e, exc_info=True)
```
rlucioni, line 198: Is `self.backend.log` an Elasticsearch construct, or is it just a Python logger under the hood? If so, you should use `log.exception`, in which case you won't need to pass `exc_info` or store the exception in `e`.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+
+        search_kwargs = self._build_search_kwargs(query_string, **kwargs)
+        search_kwargs['from'] = kwargs.get('start_offset', 0)
+
+        order_fields = set()
+        for order in search_kwargs.get('sort', []):
+            for key in order.keys():
+                order_fields.add(key)
+
+        geo_sort = '_geo_distance' in order_fields
```
rlucioni, line 180: You only seem to use this variable when calling `_process_results()`. Consider doing away with the variable and determining the value inline.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+        aggregation = {self.aggregation_name: {'cardinality': {'field': self.aggregation_key}}}
+        if precision is not None:
+            aggregation[self.aggregation_name]['cardinality']['precision_threshold'] = precision
+        return aggregation
+
+    def _build_field_facet_aggregations(self, facet_dict, precision=None):
+        """ Build and return a dictionary of aggregations for field facets."""
+        aggregations = {}
+        for facet_fieldname, opts in facet_dict.items():
+            for opt, _ in opts.items():
```
rlucioni: We typically use a double underscore for this. Single underscore is reserved for `ugettext()`.

Tangent: are the error messages in here taken from Haystack? If so, I don't think we want to mark them for translation, but if they're custom to us we might want to think about it.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
```diff
@@ -0,0 +1,298 @@
...
+            }
+        return aggregations
+
+    def _build_query_facet_aggregations(self, facet_list, precision=None):
+        """ Build and return a dictionary of aggregations for query facets."""
+        aggregations = {}
+        for facet_fieldname, value in facet_list:
+            aggregations[facet_fieldname] = {
+                'filter': {'query': {'query_string': {'query': value}}},
+                "aggregations": self._build_cardinality_aggregation(precision=precision),
```
rlucioni: Minor: single-quotes for consistency, please.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: This is done this way to be consistent with the rest of the Haystack Query interface (count is exposed via `get_count` rather than a `count` property).

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Most of this method is re-implemented logic from https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/__init__.py#L555. When adding custom logic, I tried to do things in a way that is consistent with how it is done in the original implementation.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: This is re-implemented logic from https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/elasticsearch_backend.py#L496. I just added `'distinct_hits'` to the returned dictionary.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: This is re-implemented logic from https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/elasticsearch_backend.py#L513.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: This is re-implemented logic from https://github.com/django-haystack/django-haystack/blob/v2.5.0/haystack/backends/elasticsearch_backend.py#L529.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
rlucioni: @amangano-edx I see, thanks for the context. That's fair. A comment or warning to that effect (maybe at the top of the module), might make it clearer that this is largely borrowed from Haystack.

mikedikan (approved):
I didn't see anything major to comment on, but one general question I have is if it makes sense for you to demarcate which code is 'inherited' from the original source and which is code you wrote, so future developers understand the code that should be changed and the code that shouldn't.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: These errors are custom. How do we mark strings for translation?

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Changed.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Changed.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Changed.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Changed.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Changed the underscores. Will punt on the translations since these errors should only be visible to developers.

course_discovery/apps/edx_haystack_extensions/distinct_counts/backends.py
amangano-edx: Changed.

amangano-edx:
@rlucioni I believe I've addressed all of your concerns. Please take another look when you have a minute.

@mikedikan I added links to the original implementation of overridden method to their docstrings. Does that alleviate your concern?

rlucioni (approved):

PR #726, opened 2017-03-08
ECOM-6815 add endpoint to return results with distinct hit and facet counts

amangano-edx:
Adds a new endpoint that can be used to request search results with distinct hit and facet counts.

Depends on https://github.com/edx/course-discovery/pull/725

Replaces portions of https://github.com/edx/course-discovery/pull/704

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/edx_catalog_extensions/api/serializers.py
+++ b/course_discovery/apps/edx_catalog_extensions/api/serializers.py
@@ -1,0 +1,83 @@
+# pylint: disable=abstract-method
+from drf_haystack.fields import FacetDictField, FacetListField
+from drf_haystack.serializers import FacetFieldSerializer
+from rest_framework import serializers
+from rest_framework.fields import DictField
+
+from course_discovery.apps.api.serializers import AggregateFacetSearchSerializer, QueryFacetFieldSerializer
+
+
+class DistinctCountsAggregateFacetSearchSerializer(AggregateFacetSearchSerializer):
+    """ Custom AggregateFacetSearchSerializer which includes distinct hit and facet counts."""
+
+    def get_fields(self):
+        """ Return the field_mapping needed for serializing the data."""
+        # Re-implement the logic from the superclass methods, but make sure to handle field and query facets properly.
+        # https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L950
+        # https://github.com/inonit/drf-haystack/blob/master/drf_haystack/serializers.py#L373
+        field_data = self.instance.pop('fields', {})
+        query_data = self.format_query_facet_data(self.instance.pop('queries', {}))
+        field_mapping = super(DistinctCountsAggregateFacetSearchSerializer, self).get_fields()
+
+        field_mapping['fields'] = FacetDictField(
+            child=FacetListField(child=DistinctCountsFacetFieldSerializer(field_data), required=False)
+        )
+
+        field_mapping['queries'] = DictField(
+            query_data, child=DistinctCountsQueryFacetFieldSerializer(), required=False
+        )
+
+        if self.serialize_objects:
+            field_mapping.move_to_end('objects')
+
+        self.instance['fields'] = field_data
+        self.instance['queries'] = query_data
+        return field_mapping
+
+    def get_objects(self, instance):
+        """ Return the objects that should be serialized along with the facets."""
+        data = super(DistinctCountsAggregateFacetSearchSerializer, self).get_objects(instance)
+        data['distinct_count'] = self.context['objects'].distinct_count()
+        return data
+
+    def format_query_facet_data(self, query_facet_counts):
+        """ Format and return the query facet data so that it may be properly serialized."""
+        # Re-implement the logic from the superclass method, but make sure to handle changes to the raw query facet
+        # data and extract the distinct counts.
+        # https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L966
+        query_data = {}
+        for field, options in getattr(self.Meta, 'field_queries', {}).items():  # pylint: disable=no-member
+            # The query facet data is expected to be formatted as a dictionary with fields mapping to a two-tuple
+            # containing count and distinct count.
+            count, distinct_count = query_facet_counts.get(field, (0, 0))
+            if count:
+                query_data[field] = {
+                    'field': field,
+                    'options': options,
+                    'count': count,
+                    'distinct_count': distinct_count,
+                }
+        return query_data
+
+
+class DistinctCountsFacetFieldSerializer(FacetFieldSerializer):
+    """ Custom FacetFieldSerializer which includes distinct counts."""
+    distinct_count = serializers.SerializerMethodField()
+
+    def get_distinct_count(self, instance):
+        """ Return the distinct count for this facet."""
+        # The instance is expected to be formatted as a three tuple containing the field name, normal count and
+        # distinct count. This is consistent with the superclass implementation here:
+        # https://github.com/inonit/drf-haystack/blob/master/drf_haystack/serializers.py#L321
+        count = instance[2]
+        return serializers.IntegerField(read_only=True).to_representation(count)
+
+
+class DistinctCountsQueryFacetFieldSerializer(QueryFacetFieldSerializer):
+    """ Custom QueryFacetFieldSerializer which includes distinct counts."""
+    distinct_count = serializers.SerializerMethodField()
+
+    def get_distinct_count(self, instance):
+        """ Return the distinct count for this facet."""
+        count = instance['distinct_count']
+        return serializers.IntegerField(read_only=True).to_representation(count)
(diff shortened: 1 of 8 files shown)


course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_backends.py
```diff
@@ -0,0 +1,275 @@
...
+    def test_get_distinct_count_returns_cached_value(self):
+        """ Verify that get_distinct_count returns the distinct_count from the cache when present."""
+        query = DistinctCountsSearchQuery()
+        query._distinct_hit_count = 123
+        assert query.get_distinct_count() == 123
+
+    def test_get_distinct_count_runs_query_when_cache_empty(self):
+        """ Verify that get_distinct_count runs the query and caches/returns the distinct_count."""
+        course = CourseFactory()
+        run_1 = CourseRunFactory(title='foo', course=course)
```
clintonb: Use a constant for `'foo'`.

course_discovery/apps/edx_catalog_extensions/api/serializers.py
```diff
@@ -0,0 +1,83 @@
...
+class DistinctCountsFacetFieldSerializer(FacetFieldSerializer):
+    """ Custom FacetFieldSerializer which includes distinct counts."""
+    distinct_count = serializers.SerializerMethodField()
+
+    def get_distinct_count(self, instance):
+        """ Return the distinct count for this facet."""
+        # The instance is expected to be formatted as a three tuple containing the field name, normal count and
+        # distinct count. This is consistent with the superclass implementation here:
+        # https://github.com/inonit/drf-haystack/blob/master/drf_haystack/serializers.py#L321
+        count = instance[2]
```
davec-edx, line 72: Is this a dictionary?  Can we use the key instead of position to access it?

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,257 @@
...
+
+class DistinctCountsAggregateSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, LoginMixin,
+                                                ElasticsearchTestMixin, SynonymTestMixin, APITestCase):
+    path = reverse('api:v1:extensions-search-all-facets')
+
+    def get_response(self, query=None):
+        if query:
+            query = urllib.parse.urlencode(query)
+        else:
+            query = ''
```
davec-edx: Nit: Might be clearer with a ternary

course_discovery/apps/edx_catalog_extensions/api/serializers.py
amangano-edx: This is a tuple.

course_discovery/apps/api/v1/urls.py
```diff
@@ -33,3 +34,9 @@
 router.register(r'search/programs', search_views.ProgramSearchViewSet, base_name='search-programs')
 
 urlpatterns += router.urls
+
+# Add the catalog api v1 extension urls if edx_catalog_extensions is installed.
+if 'course_discovery.apps.edx_catalog_extensions' in settings.INSTALLED_APPS:
```
clintonb: I thought we agreed to keep this *out of the api/v/1 URL-space*.

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,257 @@
...
+        """ Build a Program that will be visible in search results."""
+        kwargs.update({'status': ProgramStatus.Active})
+        return ProgramFactory(**kwargs)
+
+    def assert_url_path_and_params(self, url, expected_path, expected_query_params):
+        """ Verify that the URL contains the expected path and query parameters."""
+        assert expected_path in url
+        for param_mapping in expected_query_params:
+            assert urllib.parse.urlencode(param_mapping) in url
+
```
davec-edx: Do you need to check that there aren't unexpected params in the url's query string?

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,257 @@
...
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseRunFactory, ProgramFactory
+from course_discovery.apps.edx_catalog_extensions.api.serializers import DistinctCountsAggregateFacetSearchSerializer
+
+
+class DistinctCountsAggregateSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, LoginMixin,
+                                                ElasticsearchTestMixin, SynonymTestMixin, APITestCase):
+    path = reverse('api:v1:extensions-search-all-facets')
+
+    def get_response(self, query=None):
+        if query:
```
clintonb: This can be simplified:

```python
query = query or {}
query = urllib.parse.urlencode(query)
```

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,257 @@
...
+        now = datetime.datetime.now()
+        current = (now - datetime.timedelta(days=1), now + datetime.timedelta(days=1))
+        starting_soon = (now + datetime.timedelta(days=1), now + datetime.timedelta(days=2))
+        upcoming = (now + datetime.timedelta(days=61), now + datetime.timedelta(days=62))
+        archived = (now - datetime.timedelta(days=2), now - datetime.timedelta(days=1))
+
+        for dates in [current, starting_soon, upcoming, archived]:
+            course = CourseFactory(partner=self.partner)
+            self.build_courserun(start=dates[0], end=dates[1], course=course)
+            self.build_courserun(start=dates[0], end=dates[1], course=course)
```
davec-edx: Is this to create 2 course runs for each date?  Can you add a comment just to point out it isn't a duplicate line

course_discovery/apps/api/v1/urls.py
amangano-edx: We agreed to keep the code out of the api app, but we didn't decide anything about url/path. Do you have any suggestions as to where it should go?

I think it makes sense to use the api/v1 path since it is an api extension. Putting it here also allows us to avoid having to make changes to the `api-manager` which maps requests to api.edx.org to 'api/v1'.

davec-edx: We're changing the contract here.  I thought we weren't going to do this.

clintonb: I care more about the path. Placing it under the api/v1 path says we commit to supporting this for the lifetime of API v1. We don't want to make that commitment. The marketing site doesn't use the API proxy, so there is no need to add this to the proxy.

davec-edx: Ignore my concern, Clinton commented while I was reviewing too.

amangano-edx: This endpoint will not be published publicly, so I don't really understand the concern that we would have to commit to supporting it for the lifetime of API v1.

davec-edx: @clintonb @amangano-edx where I have similar concerns about API, versioning, maintenance, and the terms we use to call out the portions of an API.  Can I suggest taking this offline to talk about it?

clintonb: Everything at /api/v1/ is public, even if it's not part of the gateway (api.edx.org).

amangano-edx: What I meant was that we don't advertise that it's available or encourage its use. Therefore there is no guarantee that it will be supported for any amount of time and anyone using it is doing so at their own risk.

clintonb: We don't *advertise* anything. Our code is open source. If the endpoint is out there, someone may use it. We don't put hacky endpoints in our APIs. This belongs elsewhere.

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
amangano-edx: Changed.

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
amangano-edx: Changed.

course_discovery/apps/api/v1/urls.py
amangano-edx: Changed.

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
amangano-edx: Changed.

course_discovery/apps/edx_catalog_extensions/api/v1/tests/test_views.py
amangano-edx: Added.

course_discovery/apps/api/v1/urls.py
amangano-edx: Changed.

amangano-edx:
@clintonb , @davec-edx I believe I've addressed your concerns. Please take another look when you have a chance.

clintonb (approved):

PR #730, opened 2017-03-08
ECOM-7395 Disable real time and on-save index updates

MatthewPiatetsky:


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1190,11 +1190,12 @@ def serialize_course(self, course):
         return serializer
 
 
-class CourseRunSearchSerializerTests(TestCase):
+class CourseRunSearchSerializerTests(TestCase, ElasticsearchTestMixin):
```
clintonb: Mixins come first.

clintonb (approved):

course_discovery/settings/shared/test.py
```diff
@@ -7,6 +7,7 @@
         'INDEX_NAME': 'catalog_test',
     },
 }
+HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
```
rlucioni: @MatthewPiatetsky is this still necessary given your changes to the tests?

course_discovery/settings/base.py
```diff
@@ -430,7 +430,7 @@
     },
 }
 
-HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
+HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.BaseSignalProcessor'
```
rlucioni: Could you add a comment here explaining that `RealtimeSignalProcessor` overloads ES and that we use the `BaseSignalProcessor` and scheduled index updates to get past that?

course_discovery/settings/shared/test.py
MatthewPiatetsky: pretty sure nearly all the tests would fail without this since creating the objects would not update the index

course_discovery/settings/base.py
MatthewPiatetsky: does this comment look good?

course_discovery/settings/shared/test.py
rlucioni: @MatthewPiatetsky okay. Could you please add a comment to that effect here?

course_discovery/settings/base.py
rlucioni: @MatthewPiatetsky LGTM.

rlucioni (approved):

course_discovery/settings/shared/test.py
MatthewPiatetsky: added

PR #735, opened 2017-03-09
Clean up typeahead dedup

rlucioni:
Result count shouldn't be hardcoded, deduplication is by course not by run, and there's no benefit to using a set if you're performing a membership check before adding a new element.


rlucioni:
@MatthewPiatetsky FYI.

PR #738, opened 2017-03-09
ECOM-7262 Expose course run seats on program endpoint

MatthewPiatetsky:
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -383,6 +383,7 @@ class NestedProgramSerializer(serializers.ModelSerializer):
 class MinimalCourseRunSerializer(TimestampModelSerializer):
     image = ImageField(read_only=True, source='card_image_url')
     marketing_url = serializers.SerializerMethodField()
+    seats = SeatSerializer(many=True)
 
     @classmethod
     def prefetch_queryset(cls):
@@ -393,7 +394,7 @@ class MinimalCourseRunSerializer(TimestampModelSerializer):
 
     class Meta:
         model = CourseRun
-        fields = ('key', 'uuid', 'title', 'image', 'short_description', 'marketing_url',
+        fields = ('key', 'uuid', 'title', 'image', 'short_description', 'marketing_url', 'seats',
                   'start', 'end', 'enrollment_start', 'enrollment_end', 'pacing_type', 'type',)
 
     def get_marketing_url(self, obj):
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -318,6 +318,7 @@ class MinimalCourseRunSerializerTests(TestCase):
             'enrollment_end': json_date_format(course_run.enrollment_end),
             'pacing_type': course_run.pacing_type,
             'type': course_run.type,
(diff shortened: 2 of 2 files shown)


davec-edx:
Looks ok to me, my only question was why change from 'seats' to 'staff' in the test, but that got answered.

course_discovery/apps/api/serializers.py
```diff
@@ -383,6 +383,7 @@ class Meta:
 class MinimalCourseRunSerializer(TimestampModelSerializer):
     image = ImageField(read_only=True, source='card_image_url')
     marketing_url = serializers.SerializerMethodField()
+    seats = SeatSerializer(many=True)
```
clintonb, line 386: Why are we adding items to the minimal serializer? That goes against the minimal idea.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -871,26 +872,26 @@ def test_use_full_course_serializer(self):
         request = make_request()
         program = self.create_program()
 
-        # Assert that the seats are not available when fetched without any flag
```
clintonb: It seems your find-replace operation did too much.

clintonb (changes requested):

clintonb (commented):
How will this data be used?

MatthewPiatetsky:
We need the upgrade deadlines on seats for both the course about and program about pages to implement all the upgrade deadline logic for the multi course runs  
this data is already available in the courses endpoint, just not the program endpoint

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: are you asking why i changed seats to staff?
this test is testing the "use_full_course_serializer" argument by checking if seats are present or not, so if i add them to the minimal serializer i need a different field for this check

course_discovery/apps/api/serializers.py
MatthewPiatetsky: if we need this data for the program detail page we could use a different serializer, but then nothing would use this serializer. Minimal should just be the smallest set of an actual existing use case right?

course_discovery/apps/api/tests/test_serializers.py
clintonb: Yeah...this test is not implemented well. If you want to verify the `CourseSerializer` is used, just use that class directly. Check to see what data *isn't* in the serialized value is odd and, as we see now, brittle. `test_course_ordering` offers a good example of cleanly validating the courses.

```python
actual = self.serializer_class(program, context={'request': request, 'use_full_course_serializer': 1}).data['courses']
expected = ProgramCourseSerializer(
    program.courses,
    many=True,
    context={'request': request, 'program': program, 'course_runs': list(program.course_runs)}
).data
self.assertDictEqual(actual, expected)
```

clintonb (changes requested):

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: ok I changed the test to use the CourseRunSerializer

clintonb (approved):

PR #739, opened 2017-03-09
ECOM-7262 Add marketable_enrollable_course_runs_with_archived param to programs  

MatthewPiatetsky:
Please review this PR first.
https://github.com/edx/course-discovery/pull/738

@edx/ecommerce 

We pass a query param to apply marketable/enrollable filters to course runs for the course about page/courses endpoint.  
We don't do this for the program about page/programs endpoint (but we should to have consistent behavior on a program page and its course run pages)

I need to apply the filters before the courses are sorted (because at that point they are a list rather than a queryset)

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -642,10 +642,18 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
         read_only_fields = ('uuid', 'marketing_url', 'banner_image')
 
     def get_courses(self, program):
+        course_runs = list(program.course_runs)
+
+        if self.context.get('marketable_enrollable_course_runs_with_archived'):
+            marketable_enrollable_course_runs = set()
+            for course in program.courses.all():
+                marketable_enrollable_course_runs.update(course.course_runs.marketable().enrollable())
+            course_runs = list(set(course_runs).intersection(marketable_enrollable_course_runs))
+
         if program.order_courses_by_start_date:
-            courses, course_runs = self.sort_courses(program)
+            courses = self.sort_courses(program, course_runs)
         else:
-            courses, course_runs = program.courses.all(), list(program.course_runs)
+            courses = program.courses.all()
 
         course_serializer = MinimalProgramCourseSerializer(
             courses,
@@ -662,7 +670,7 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
 
         return course_serializer.data
 
-    def sort_courses(self, program):
+    def sort_courses(self, program, course_runs):
         """
         Sorting by enrollment start then by course start yields a list ordered by course start, with
         ties broken by enrollment start. This works because Python sorting is stable: two objects with
@@ -672,7 +680,6 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
         course_runs should never be empty. If it is, key functions in this method attempting to find the
         min of an empty sequence will raise a ValueError.
         """
-        course_runs = list(program.course_runs)
 
         def min_run_enrollment_start(course):
             # Enrollment starts may be empty. When this is the case, we make the same assumption as
@@ -715,7 +722,7 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
         courses.sort(key=min_run_enrollment_start)
         courses.sort(key=min_run_start)
 
-        return courses, course_runs
+        return courses
 
 
 class ProgramSerializer(MinimalProgramSerializer):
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -892,6 +893,35 @@ def test_use_full_course_serializer(self):
...
+        """
+        course = CourseFactory()
+        CourseRunFactory(status=CourseRunStatus.Unpublished, course=course)
+        marketable_enrollable_run = CourseRunFactory(
+            status=CourseRunStatus.Published,
+            end=datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=10),
+            enrollment_start=None,
+            enrollment_end=None,
+            course=course,
+            slug='notempty'
```
clintonb: If you're not using this value, this can go away.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -892,6 +893,35 @@ def test_use_full_course_serializer(self):
...
+
+        serializer = self.serializer_class(
+            program,
+            context={
+                'request': request,
+                'marketable_enrollable_course_runs_with_archived': marketable_enrollable
+            }
+        )
+
+        self.assertEqual(len(serializer.data['courses'][0]['course_runs']), 2 - marketable_enrollable)
```
clintonb: Check the actual data being returned, not the count. This will probably mean splitting this into two distinct test methods, and using helper methods to de-dupe.

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -34,11 +34,10 @@ def get_queryset(self):
...
         context = super().get_serializer_context(*args, **kwargs)
-        context.update({
-            'published_course_runs_only': get_query_param(self.request, 'published_course_runs_only'),
-            'exclude_utm': get_query_param(self.request, 'exclude_utm'),
-            'use_full_course_serializer': get_query_param(self.request, 'use_full_course_serializer'),
-        })
+        query_params = ['exclude_utm', 'use_full_course_serializer', 'published_course_runs_only',
+                        'marketable_enrollable_course_runs_with_archived']
+        for query_param in query_params:
+            context[query_param] = get_query_param(self.request, query_param)
```
clintonb, line 41: Update the docstring for `list()`

course_discovery/apps/api/serializers.py
```diff
@@ -642,10 +642,22 @@ class Meta:
         read_only_fields = ('uuid', 'marketing_url', 'banner_image')
 
     def get_courses(self, program):
+        course_runs = list(program.course_runs)
```
clintonb, line 645: The variable names are a bit confusing when taken altogether. Also, you can use some set magic to simplify this.

```python
course_runs = list(program.course_runs)

if self.context.get('marketable_enrollable_course_runs_with_archived'):
  marketable_enrollable_course_runs = set()
  for course in program.courses.all():
    marketable_enrollable_course_runs.union(set(list(course.course_runs.marketable().enrollable())))

  course_runs = list(set(course_runs).intersection(marketable_enrollable_course_runs))
```

clintonb (changes requested):

course_discovery/apps/api/tests/test_serializers.py

course_discovery/apps/api/tests/test_serializers.py
MatthewPiatetsky: added comparison of course run data

course_discovery/apps/api/v1/views/programs.py
MatthewPiatetsky: updated

course_discovery/apps/api/serializers.py
MatthewPiatetsky: changed to something similar

clintonb: TIL: `set.update()`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -892,6 +893,48 @@ def test_use_full_course_serializer(self):
...
+        expected = MinimalProgramCourseSerializer(
+            [course],
+            many=True,
+            context={
+                'request': request,
+                'program': program,
+                'course_runs': [unpublished_course_run, marketable_enrollable_run]
+            }
+        ).data
+        expected[0]['course_runs'] = [course_run for course_run in expected[0]['course_runs'] if
```
clintonb: `if` statements just make debugging tests unnecessarily difficult. Split this into two test cases. Actually...you don't need to test the case where the filter is disabled. That will be tested by every other test case.

clintonb (approved):
Fix the test, and this is good to go!

PR #740, opened 2017-03-10
Fix Courses Without End Date Not Visible via API. ECOM-7444

awaisdar001:
### Description
We have noticed that there is a number of open and available courses missing from csv catalogs generated by our catalog API by using either

### Fix
Courses were filtered which were missing `end` dates, I have changed the `active` method to include those runs as well. 

### Reviewers
- [x] Style and readability review: @Ayub-Khan 
- [x] Code review: @clintonb 
- [x] Code review: @schenedx 

FYI: @adampalay @rlucioni 

### Post-review

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_query.py
+++ b/course_discovery/apps/course_metadata/tests/test_query.py
@@ -46,26 +46,33 @@ class CourseQuerySetTests(TestCase):
 class CourseRunQuerySetTests(TestCase):
     def test_active(self):
         """ Verify the method returns only course runs currently open for enrollment or opening in the future. """
+        now = datetime.datetime.now(pytz.UTC)
+        active_course_end = now + datetime.timedelta(days=60)
+        inactive_course_end = now - datetime.timedelta(days=15)
+        open_enrollment_end = now + datetime.timedelta(days=30)
+        closed_enrollment_end = now - datetime.timedelta(days=30)
+
         # Create course with end date in future and enrollment_end in past.
-        end = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=2)
-        enrollment_end = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=1)
-        CourseRunFactory(end=end, enrollment_end=enrollment_end)
+        CourseRunFactory(end=active_course_end, enrollment_end=closed_enrollment_end)
 
         # Create course with end date in past and no enrollment_end.
-        end = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=2)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -44,29 +44,42 @@ def test_active(self):
 
 @ddt.ddt
 class CourseRunQuerySetTests(TestCase):
+    def setUp(self):
+        super(CourseRunQuerySetTests, self).setUp()
+
+        now = datetime.datetime.now(pytz.UTC)
+        self.active_course_end = now + datetime.timedelta(days=60)
```
clintonb: Move these to the methods that need them rather than creating them for every test.

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -44,29 +44,42 @@ def test_active(self):
...
+        self.inactive_course_end = now - datetime.timedelta(days=15)
+        self.open_enrollment_end = now + datetime.timedelta(days=30)
+        self.closed_enrollment_end = now - datetime.timedelta(days=30)
+
     def test_active(self):
-        """ Verify the method returns only course runs currently open for enrollment or opening in the future. """
```
clintonb: This change is not necessary.

clintonb: Actually...none of your changes to this method are necessary. All you need to do here is add `active_no_end_date = CourseRunFactory(end=None, enrollment_end=self.open_enrollment_end)` to the end of this method and update the assertion.

clintonb (changes requested):

schenedx:
@rlucioni I always had a question why course_run `active` must have course_end date defined. Was there a consideration when that was done? Do you think the change here is valid and won't break course multi-run select dialog? 
This change does have an impact to the Sailthru marketing. But I can manage that impact

rlucioni:
@schenedx I was actually just talking about the discrepancy between the Course and CourseRun `active` filters with @clintonb. I think this is the right change to make; end date isn't necessary for a run to be considered "active."

schenedx (changes requested):
Unit test can be cleaned up according to @clintonb's comment

awaisdar001:
@clintonb @schenedx PR is updated with requested changes. Please have a look.

clintonb (approved):

schenedx (approved):
👍

schenedx:
@awaisdar001 Are you going to merge this PR?

awaisdar001:
@Ayub-Khan Thank you for doing this, while I was away.

Ayub-Khan:
👍

PR #743, opened 2017-03-10
Fixed search/add staff.

waheedahmed:
[ECOM-7416](https://openedx.atlassian.net/browse/ECOM-7416)

Course team user cannot select or add new Course Staff.


waheedahmed:
@tasawernawaz @awais786 please review.

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -204,3 +204,32 @@ def test_instructor_autocomplete_with_invalid_uuid(self):
...
+        non_staff_user = UserFactory()
+        self.client.logout()
+        self.client.login(username=non_staff_user.username, password=USER_PASSWORD)
+
+        response = self.client.get(
+            reverse('admin_metadata:person-autocomplete') + '?q={q}'.format(q='ins')
+        )
+        self._assert_response(response, 2)
+
+        # update first instructor's name
```
tasawernawaz: You can remove this part of test (which is already tested in `test_instructor_autocomplete`),  because this test only related to permission.

tasawernawaz (approved):
Just a minor comment.

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -57,7 +57,7 @@ def get_queryset(self):
 
 class PersonAutocomplete(autocomplete.Select2QuerySetView):
     def get_queryset(self):
-        if self.request.user.is_authenticated() and self.request.user.is_staff:
+        if self.request.user.is_authenticated():
```
awais786: I would like to add a suggestion here apart from checking is_authentication also check that user must be part of any group. With that check this API remains safe from common users testing.

waheedahmed: @awais786 I think we should not add any publisher related check in course_metadata, @bderusha what do you think?

awais786 (approved):

course_discovery/apps/course_metadata/lookups.py
clintonb: If your view requires authentication, use [`LoginRequiredMixin`](https://docs.djangoproject.com/en/1.9/topics/auth/default/#the-loginrequired-mixin). The same goes for permissions. Avoid using special data—empty array in this case—to indicate errors. That's what exceptions and HTTP error codes are for.

clintonb: There is also [`staff_member_required`](https://docs.djangoproject.com/en/1.9/ref/contrib/admin/#django.contrib.admin.views.decorators.staff_member_required).

waheedahmed: We don't need staff check here. Its being used in Publisher app so user will not be a staff member always.

PR #748, opened 2017-03-13
Exclude courses without active, marketable runs from catalogs

rlucioni:
The catalog API should only list courses which have at least one active and marketable course run.

ECOM-6473

@edx/ecommerce FYI.

### DIFF
--- a/course_discovery/apps/api/v1/views/catalogs.py
+++ b/course_discovery/apps/api/v1/views/catalogs.py
@@ -84,14 +84,12 @@ class CatalogViewSet(viewsets.ModelViewSet):
         """
         Retrieve the list of courses contained within this catalog.
 
-        Only courses with active course runs are returned. A course run is considered active if it is currently
-        open for enrollment, or will open in the future.
+        Only courses with at least one active and marketable course run are returned.
         ---
         serializer: serializers.CourseSerializerExcludingClosedRuns
         """
-
         catalog = self.get_object()
-        queryset = catalog.courses().active()
+        queryset = catalog.courses().active().marketable()
         queryset = prefetch_related_objects_for_courses(queryset)
 
         page = self.paginate_queryset(queryset)
--- a/course_discovery/apps/course_metadata/query.py
+++ b/course_discovery/apps/course_metadata/query.py
@@ -9,18 +9,40 @@ from course_discovery.apps.course_metadata.choices import CourseRunStatus, Progr
 
 class CourseQuerySet(models.QuerySet):
     def active(self):
-        """ Filters Courses to those with CourseRuns that are either currently open for enrollment,
-        or will be open for enrollment in the future. """
-
+        """
+        Filter Courses to those with CourseRuns that have not yet ended and are
+        either open for enrollment or will be open for enrollment in the future.
+        """
         now = datetime.datetime.now(pytz.UTC)
         return self.filter(
             (
-                Q(course_runs__end__gt=now) | Q(course_runs__end__isnull=True)
+                Q(course_runs__end__gt=now) |
+                Q(course_runs__end__isnull=True)
             ) &
             (
-                Q(course_runs__enrollment_end__gt=now) | Q(course_runs__enrollment_end__isnull=True)
+                Q(course_runs__enrollment_end__gt=now) |
+                Q(course_runs__enrollment_end__isnull=True)
             )
         )
+
+    def marketable(self):
+        """
+        Filter Courses to those with CourseRuns that can be marketed. A CourseRun
+        is deemed marketable if it has a defined slug, has seats, and has been published.
+        """
+        # exclude() is intentionally avoided here. We want Courses to be included
+        # in the resulting queryset if only one of their runs matches our "marketable"
+        # criteria. For example, consider a Course with two CourseRuns; one of the
+        # runs is published while the other is not. If you used exclude(), the Course
+        # would be dropped from the queryset even though it has one run which matches
+        # our criteria.
+        return self.filter(
+            Q(course_runs__slug__isnull=False) & ~Q(course_runs__slug='')
+        ).filter(
+            course_runs__seats__isnull=False
+        ).filter(
+            course_runs__status=CourseRunStatus.Published
+        ).distinct()
 
 
 class CourseRunQuerySet(models.QuerySet):
--- a/course_discovery/apps/course_metadata/tests/test_query.py
+++ b/course_discovery/apps/course_metadata/tests/test_query.py
@@ -11,35 +11,67 @@ from course_discovery.apps.course_metadata.tests.factories import CourseRunFacto
 
 class CourseQuerySetTests(TestCase):
     def test_active(self):
-        """ Verify the method filters the Courses to those with active course runs. """
+        """
+        Verify the method filters Courses to those with active CourseRuns.
+        """
         now = datetime.datetime.now(pytz.UTC)
         active_course_end = now + datetime.timedelta(days=60)
         inactive_course_end = now - datetime.timedelta(days=15)
         open_enrollment_end = now + datetime.timedelta(days=30)
         closed_enrollment_end = now - datetime.timedelta(days=30)
 
-        # Create an active enrollable course
+        # Create an active, enrollable course run
         active_course = CourseRunFactory(enrollment_end=open_enrollment_end, end=active_course_end).course
 
-        # Create an active unenrollable course
+        # Create an active, unenrollable course run
         CourseRunFactory(enrollment_end=closed_enrollment_end, end=active_course_end, course__title='ABC Test Course 2')
 
-        # Create an inactive unenrollable course
+        # Create an inactive, unenrollable course run
         CourseRunFactory(enrollment_end=closed_enrollment_end, end=inactive_course_end)
 
-        # Create an active course with unrestricted enrollment
+        # Create an active course run with an unspecified enrollment end
         course_without_enrollment_end = CourseRunFactory(enrollment_end=None, end=active_course_end).course
 
-        # Create an inactive course with unrestricted enrollment
+        # Create an inactive course run with an unspecified enrollment end
         CourseRunFactory(enrollment_end=None, end=inactive_course_end)
 
-        # Create course with end date is NULL
+        # Create an enrollable course run with an unspecified end date
         course_without_end = CourseRunFactory(enrollment_end=open_enrollment_end, end=None).course
 
-        self.assertEqual(
-            set(Course.objects.active()),
-            {active_course, course_without_enrollment_end, course_without_end}
-        )
+        assert set(Course.objects.active()) == {
+            active_course, course_without_enrollment_end, course_without_end
+        }
+
+    def test_marketable(self):
+        """
+        Verify the method filters Courses to those with marketable CourseRuns.
+        """
+        # Courses whose runs have null or empty slugs are excluded, even if
+        # those runs are published and have seats.
+        for invalid_slug in (None, ''):
+            excluded_course_run = CourseRunFactory(slug=invalid_slug)
+            SeatFactory(course_run=excluded_course_run)
+
+        # Courses whose runs have no seats are excluded, even if those runs
+        # are published and have valid slugs.
+        CourseRunFactory()
+
+        # Courses whose runs are unpublished are excluded, even if those runs
+        # have seats and valid slugs.
+        excluded_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        SeatFactory(course_run=excluded_course_run)
+
+        # Courses with at least one run that is published and has seats and a valid
+        # slug are included.
+        included_course_run = CourseRunFactory()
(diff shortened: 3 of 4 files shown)


mikedikan (approved):

course_discovery/apps/course_metadata/query.py
```diff
@@ -9,19 +9,44 @@
...
+        # exclude() is intentionally avoided here. We want Courses to be included
+        # in the resulting queryset if only one of their runs matches our "marketable"
+        # criteria. For example, consider a Course with two CourseRuns; one of the
+        # runs is published while the other is not. If you used exclude(), the Course
+        # would be dropped from the queryset even though it has one run which matches
+        # our criteria.
+        return self.filter(
+            course_runs__slug__isnull=False
+        ).filter(
+            course_runs__slug__gt=''
```
MatthewPiatetsky: imo it might be clearer to do the null and blank conditions separately if you're trying to check both. gt seems to me like it could be confusing

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -11,35 +11,67 @@
...
         inactive_course_end = now - datetime.timedelta(days=15)
         open_enrollment_end = now + datetime.timedelta(days=30)
         closed_enrollment_end = now - datetime.timedelta(days=30)
 
-        # Create an active enrollable course
+        # Create an active, enrollable course run
         active_course = CourseRunFactory(enrollment_end=open_enrollment_end, end=active_course_end).course
 
-        # Create an active unenrollable course
+        # Create a active, unenrollable course run
```
MatthewPiatetsky: an is correct

clintonb: an

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -84,14 +84,14 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin,unused
...
+        Only courses with at least one active and marketable course run are returned.
+        A course run is considered active if it is currently open for enrollment,
+        or will open in the future.
         ---
         serializer: serializers.CourseSerializerExcludingClosedRuns
         """
-
         catalog = self.get_object()
-        queryset = catalog.courses().active()
+        queryset = catalog.courses().active().marketable()
```
MatthewPiatetsky, line 92: while this solves the problem of courses with no runs, are we sure that we want the additional checks regarding slug/seats/etc... here

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -11,35 +11,67 @@
...
+            active_course, course_without_enrollment_end, course_without_end
+        }
+
+    def test_marketable(self):
+        """
+        Verify the method filters Courses to those with marketable CourseRuns.
+        """
+        # Courses whose runs have null or empty slugs are excluded, even if
+        # those runs are published and have seats.
+        for invalid_slug in (None, ''):
```
MatthewPiatetsky, line 51: can/do they actually have both?

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -84,14 +84,14 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin,unused
         """
         Retrieve the list of courses contained within this catalog.
 
-        Only courses with active course runs are returned. A course run is considered active if it is currently
-        open for enrollment, or will open in the future.
+        Only courses with at least one active and marketable course run are returned.
+        A course run is considered active if it is currently open for enrollment,
+        or will open in the future.
```
MatthewPiatetsky: should we also define marketable here or not define active?

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -11,35 +11,67 @@
...
+            excluded_course_run = CourseRunFactory(slug=invalid_slug)
+            SeatFactory(course_run=excluded_course_run)
```
clintonb, line 53: `SeatFactory(course_run__slug=invalid_slug)`

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -11,35 +11,67 @@
...
+
+        # Courses whose runs have no seats are excluded, even if those runs
+        # are published and have valid slugs.
+        CourseRunFactory()
+
+        # Courses whose runs are unpublished are excluded, even if those runs
+        # have seats and valid slugs.
+        excluded_course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        SeatFactory(course_run=excluded_course_run)
```
clintonb, line 62: Combine, similar to above.

course_discovery/apps/course_metadata/query.py
```diff
@@ -9,19 +9,44 @@
...
+        ).filter(
+            # This will only include course runs with seats.
+            course_runs__seats__isnull=False
```
MatthewPiatetsky: would seats__len__gt=0 work here? might be clearer and eliminate need for comment

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -11,35 +11,67 @@
...
+
+        # Courses with at least one run that is published and has seats and a valid
+        # slug are included.
+        included_course_run = CourseRunFactory()
+        SeatFactory(course_run=included_course_run)
```
clintonb, line 67: Combine, similar to above.

course_discovery/apps/course_metadata/query.py
```diff
@@ -9,19 +9,44 @@
...
+                Q(course_runs__end__isnull=True)
             ) &
             (
-                Q(course_runs__enrollment_end__gt=now) | Q(course_runs__enrollment_end__isnull=True)
+                Q(course_runs__enrollment_end__gt=now) |
+                Q(course_runs__enrollment_end__isnull=True)
             )
         )
 
+    def marketable(self):
```
schenedx, line 28: These duplicate the logic in the `CourseRunQuerySet.marketable()`. Is there a way we can centralize these two? I can't think of a way myself.

schenedx (approved):
👍

course_discovery/apps/course_metadata/tests/test_query.py
rlucioni: @MatthewPiatetsky a course run may have no slug, or it may have the empty string as its slug. A course run can't have both a null and empty string slug at the same time. Does that answer your question?

course_discovery/apps/api/v1/views/catalogs.py
rlucioni: @MatthewPiatetsky I'll remove the definition of active here. Better to centralize those definitions in docstrings on the query methods themselves instead of repeating those definitions all over the place.

course_discovery/apps/api/v1/views/catalogs.py
rlucioni: @MatthewPiatetsky yes, take a look at the serializer used by this view, [`CourseSerializerExcludingClosedRuns`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L573). If we don't filter courses based on active and marketable runs at this point, we're going to end up with empty courses in the response, in which case we wouldn't have meet the acceptance criteria on [ECOM-6473](https://openedx.atlassian.net/browse/ECOM-6473).

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky can you think of a cleaner way to do the null and blank checks separately without using `exclude()`?

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky it would, but I'd prefer to be consistent with our use of isnull throughout this module. On second thought, this is pretty clear, and that comment provides no value; I'll remove it.

course_discovery/apps/course_metadata/tests/test_query.py
rlucioni: @clintonb nice, thanks for the tip. I forgot you could do that.

course_discovery/apps/api/v1/views/catalogs.py
MatthewPiatetsky: I get that, and we probably want these filters as well.  
My point was the ticket specifically says filter out courses with no runs, not filter out courses that are not marketable (which will also filter out courses with no runs)
I don't know the requirements for the ticket, but in theory maybe we want courses with no marketing slug to show up in the catalogs for example. Just want to clarify if these are in fact the requirements we want

course_discovery/apps/course_metadata/tests/test_query.py
MatthewPiatetsky: I was just wondering how they get both?  
Do different data loaders set them differently?  
I imagine if data is being set in one form not setting anything will only be None or ''

course_discovery/apps/course_metadata/tests/test_query.py
rlucioni: @clintonb I take it back, this doesn't seem to work. I get a `TypeError` when I try it.
```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/factory/base.py", line 69, in __call__
    return cls.create(**kwargs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/factory/base.py", line 623, in create
    return cls._generate(True, attrs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/factory/base.py", line 548, in _generate
    obj = cls._prepare(create, **attrs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/factory/base.py", line 523, in _prepare
    return cls._create(model_class, *args, **kwargs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/factory/django.py", line 181, in _create
    return manager.create(*args, **kwargs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/django/db/models/manager.py", line 122, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/django/db/models/query.py", line 399, in create
    obj = self.model(**kwargs)
  File "/Users/renzolucioni/.pyenv/versions/discovery/lib/python3.5/site-packages/django/db/models/base.py", line 451, in __init__
    raise TypeError("'%s' is an invalid keyword argument for this function" % list(kwargs)[0])
TypeError: 'course_run__slug' is an invalid keyword argument for this function
```

course_discovery/apps/course_metadata/query.py
MatthewPiatetsky: could you do something like `.filter(Q(alias__isnull=False) & ~Q(alias__exact=''))` ?

course_discovery/apps/course_metadata/query.py
rlucioni: @schenedx the queries are similar, but they're not exactly the same. Note the use of `distinct()` and the lack of `exclude()` here. We could try pulling the field names we're querying on out into a single, centralized data structure, then use that to build up the queries in both methods, but I think that would wind up being more complex than necessary.

course_discovery/apps/course_metadata/tests/test_query.py
rlucioni: @MatthewPiatetsky I'm not sure, but I need to guard against both because we made the mistake of setting `null=True` on this `CharField`.
```py
```
This is why the [Django docs](https://docs.djangoproject.com/en/1.10/ref/models/fields/#null) warn against setting `null=True` on `CharField`s; it means we have to deal with two possible representations of "no data", null and the empty string. Django convention is to use the empty string.

course_discovery/apps/api/v1/views/catalogs.py
rlucioni: @MatthewPiatetsky this is how we planned the ticket. See [ECOM-7379](https://openedx.atlassian.net/browse/ECOM-7379).

course_discovery/apps/course_metadata/query.py
schenedx: That's fine. Just want to see if you have considered it.

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky https://www.youtube.com/watch?v=cEeWLDMDqpk

(For reference, the lookup type is assumed to be `exact` when no lookup type is provided.)

rlucioni:
@MatthewPiatetsky I've addressed your comments.

MatthewPiatetsky (approved):

PR #749, opened 2017-03-13
Prevent serializer from discarding prefetched data

rlucioni:
Chained calls to queryset methods like filter() and exclude() imply a new database query and will cause prefetched results to be ignored. This change is an incremental improvement which more than halves the query count for requests which don't use the marketable_course_runs_only or marketable_enrollable_course_runs_with_archived querystring parameters (from 51 to 22 queries on a system with 1 course and 3 nested course runs). Reducing the query count for requests which do use those parameters requires filtering in Python which mimics the active, marketable, and enrollable queryset methods.

@edx/ecommerce remember this PR when you're making changes to serializers that are meant to operate on prefetched data.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -258,10 +258,11 @@ class PersonSerializer(serializers.ModelSerializer):
         return person
 
     def get_social_network_url(self, url_type, obj):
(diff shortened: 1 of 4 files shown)


rlucioni:
@clintonb would you mind taking a quick look at this when you have a chance? This is the fix we paired yesterday.

clintonb (approved):

PR #757, opened 2017-03-15
Courses appearing in catalogs must be available

rlucioni:
A course is considered to be "available" if it contains at least one course run that can be enrolled in immediately, is ongoing or yet to start, and appears on the marketing site. The catalog API should only list courses which are available. An earlier attempt at this fell short because it filtered courses to those which contained at least one active *or* marketable run.

ECOM-6473

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -477,8 +477,12 @@ class MinimalCourseSerializer(TimestampModelSerializer):
     image = ImageField(read_only=True, source='card_image_url')
 
     @classmethod
-    def prefetch_queryset(cls):
-        return Course.objects.all().select_related('partner').prefetch_related(
+    def prefetch_queryset(cls, queryset=None):
+        # Explicitly check for None to avoid returning all Courses when the
+        # queryset passed in happens to be empty.
+        queryset = queryset if queryset is not None else Course.objects.all()
+
+        return queryset.select_related('partner').prefetch_related(
             'authoring_organizations',
             Prefetch('course_runs', queryset=MinimalCourseRunSerializer.prefetch_queryset()),
         )
@@ -501,8 +505,12 @@ class CourseSerializer(MinimalCourseSerializer):
     marketing_url = serializers.SerializerMethodField()
 
     @classmethod
-    def prefetch_queryset(cls):
-        return Course.objects.all().select_related('level_type', 'video', 'partner').prefetch_related(
+    def prefetch_queryset(cls, queryset=None):
+        # Explicitly check for None to avoid returning all Courses when the
+        # queryset passed in happens to be empty.
+        queryset = queryset if queryset is not None else Course.objects.all()
+
+        return queryset.select_related('level_type', 'video', 'partner').prefetch_related(
             'expected_learning_items',
             'prerequisites',
             'subjects',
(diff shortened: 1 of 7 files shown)


course_discovery/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+
+import pytest
+import pytz
+
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
+from course_discovery.apps.course_metadata.tests.factories import SeatFactory
+
+
+@pytest.fixture(scope='class')
+def course_run_states(request):
```
rlucioni, line 13: @clintonb I ended up writing a fixture!

course_discovery/apps/api/serializers.py
```diff
@@ -581,12 +589,23 @@ class Meta(CourseSerializer.Meta):
...
+    """
     course_runs = serializers.SerializerMethodField()
 
     def get_course_runs(self, course):
-        return CourseRunSerializer(course.course_runs.active().marketable(), many=True, context=self.context).data
+        return CourseRunSerializer(
+            # TODO: These queryset methods chain filter() and exclude() calls,
+            # causing prefetched results to be discarded. They should be replaced
+            # with Python-based filtering that preserves the prefetched data.
+            course.course_runs.active().enrollable().marketable(),
```
MatthewPiatetsky, line 605: this is probably not a big deal, but active and enrollable apply some of the same filters

course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
```diff
@@ -130,8 +133,8 @@ def test_create_with_new_user(self):
 
         catalog = Catalog.objects.latest()
         latest_user = User.objects.latest()
-        self.assertEqual(latest_user.username, new_viewer_username)
-        self.assertListEqual(list(catalog.viewers), [existing_viewer, latest_user])
+        assert latest_user.username == new_viewer_username
+        assert set(catalog.viewers) == {existing_viewer, latest_user}
```
MatthewPiatetsky, line 137: good idea to do this instead of trying to ignore order

course_discovery/apps/api/serializers.py
MatthewPiatetsky: also are you trying to mimic available here, didn't you create an actual available filter?

course_discovery/apps/course_metadata/query.py
```diff
@@ -8,41 +8,49 @@
...
+                Q(course_runs__enrollment_start__lte=now) |
+                Q(course_runs__enrollment_start__isnull=True)
             ) &
             (
                 Q(course_runs__enrollment_end__gt=now) |
                 Q(course_runs__enrollment_end__isnull=True)
             )
         )
 
-    def marketable(self):
```
MatthewPiatetsky, line 28: i'm confused from this diff if you are you removing this? is this because it is not used? If so, are you sure it is not used?

course_discovery/apps/api/serializers.py
rlucioni: @MatthewPiatetsky I noticed that overlap between `active()` and `enrollable()`, too. I don't think it's a problem.

I created an `available()` filter for filtering a course queryset, not a course run queryset. We only need to filter courses by availability. However, we currently have a use for independent course run filters. For example, you may want to get `marketable()` runs without limiting yourself to runs which are also `active()` and `enrollable()`. That said, I think we may eventually centralize on an equivalent `available()` filter for course runs. That's not my goal here, though.

course_discovery/apps/course_metadata/query.py
rlucioni: @MatthewPiatetsky that's right, I'm removing this because it's not used. I added it a few days ago and now I'm generalizing it and the `active()` filters into the one filter we actually wanted in the first place.

MatthewPiatetsky (approved):

PR #759, opened 2017-03-17
Cache search results when requesting facet counts

amangano-edx:
Our search/facets endpoints currently execute three search queries for every one search request. When the drf_haystack facets action initializes its serializer, it calls SearchQuerySet.facet_counts, which executes the query but doesn't cache any of the results. Later, when the serializer creates a paginator object, SearchQuerySet.count is called, which executes another query and doesn't cache any results. Finally, when the paginator object requests the results for the page, a third query is run. This time, the results get cached but by this point three queries have already run.

This fix might not be the most elegant, but it is quick and would allow us to significantly reduce search load. I've already run some loadtests for the new distinct_counts endpoint with these changes and found response times to be 25% lower than they were in prior tests (186 ms vs 263 ms on average). With these changes, the distinct_counts endpoint is actually 8% faster than the original search/all/facets endpoint (186 ms vs 202 ms on average). Without them, it is 30% slower.

@rlucioni 
@edx/ecommerce

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
+++ b/course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from haystack.query import SearchQuerySet
 from course_discovery.apps.edx_haystack_extensions.distinct_counts.backends import DistinctCountsSearchQuery
 
@@ -49,3 +50,25 @@ class DistinctCountsSearchQuerySet(SearchQuerySet):
         if self._distinct_result_count is None:
             self._distinct_result_count = self.query.get_distinct_count()
         return self._distinct_result_count
+
+    def facet_counts(self):
+        """
+        Return the facet counts. Note: this will cause the query to run if it hasn't already.
+
+        Override the original implementation so that if we're forced to run the query, we can
+        cache the results that come back with it and avoid having to make another request to get
+        them later. Original implementation:
+        https://github.com/django-haystack/django-haystack/blob/master/haystack/query.py#L532
+        """
+        if self.query.has_run():
+            return self.query.get_facet_counts()
+        else:
+            # Force the query to run and fill the cache with the first page of results.
+            # This will cause the facet_counts to be cached along with the rest of the results
+            # and could potentially reduce the number of queries required to complete a search
+            # request.
+            #
+            # Note: If there are fewer than count results for the query, ES will simply return what it
+            # has found without raising an exception.
+            self._fill_cache(0, settings.DISTINCT_COUNTS_QUERY_CACHE_WARMING_COUNT)
+            return self.query.get_facet_counts()
--- a/course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
+++ b/course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
@@ -1,4 +1,5 @@
 import datetime
+import mock
 
 import pytest
 from django.test import TestCase
@@ -123,3 +124,39 @@ class DistinctCountsSearchQuerySetTests(ElasticsearchTestMixin, TestCase):
         hidden_count, hidden_distinct_count = facet_counts['queries']['hidden']
         assert hidden_count == 2
         assert hidden_distinct_count == 1
+
+    def test_facet_counts_caches_results(self):
+        """ Verify that facet_counts cache results when it is forced to run the query."""
+        course = CourseFactory()
+        runs = [
+            CourseRunFactory(title='foo', pacing_type='self_paced', hidden=True, course=course),
+            CourseRunFactory(title='foo', pacing_type='self_paced', hidden=True, course=course),
+            CourseRunFactory(title='foo', pacing_type='instructor_paced', hidden=False, course=course),
+        ]
+
+        queryset = SearchQuerySet().filter(title='foo').models(CourseRun)
+        queryset = queryset.facet('pacing_type').query_facet('hidden', 'hidden:true')
+        dc_queryset = DistinctCountsSearchQuerySet.from_queryset(queryset).with_distinct_counts('aggregation_key')
+
+        # This should force the query to run and the results to be cached
+        facet_counts = dc_queryset.facet_counts()
+
+        with mock.patch.object(DistinctCountsSearchQuery, 'run') as mock_run:
+            # Calling facet_counts again shouldn't result in an additional query
+            cached_facet_counts = dc_queryset.facet_counts()
+            assert not mock_run.called
+            assert facet_counts == cached_facet_counts
+
+            # Calling count shouldn't result in another query, as we should have already cached it with the
+            # first request.
+            count = dc_queryset.count()
+            assert not mock_run.called
+            assert count == len(runs)
+
+            # Fetching the results shouldn't result in another query, as we should have already cached them
+            # with the initial request.
+            results = dc_queryset[:]
+            assert not mock_run.called
+            expected = {run.key for run in runs}
+            actual = {run.key for run in results}
+            assert expected == actual
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -454,6 +454,11 @@ SEARCH_FACET_LIMIT = 10000
 DISTINCT_COUNTS_HIT_PRECISION = 1500
 DISTINCT_COUNTS_FACET_PRECISION = 250
 
+# The number of records that should be requested when warming the SearchQuerySet cache. Set this to equal the
+# number of records typically requested with each search query in order to reduce the number of queries that need
+# to be executed.
+DISTINCT_COUNTS_QUERY_CACHE_WARMING_COUNT = 20
+
 DEFAULT_PARTNER_ID = None
 
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#site-id


course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
```diff
@@ -49,3 +49,25 @@ def distinct_count(self):
...
+            return self.query.get_facet_counts()
+        else:
+            # Force the query to run and fill the cache with the first page of results.
+            # This will cause the facet_counts to be cached along with the rest of the results
+            # and could potentially reduce the number of queries required to complete a search
+            # request.
+            #
+            # Note: If there are fewer than 10 results for the query, ES will simply return what it
+            # has found without raising an exception.
+            self[0:10]
```
amangano-edx: Potentially we could pass the pagination params through somehow and eliminate an unnecessary second query for paginated requests.

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
```diff
@@ -49,3 +49,25 @@ def distinct_count(self):
...
             self._distinct_result_count = self.query.get_distinct_count()
         return self._distinct_result_count
+
+    def facet_counts(self):
+        """
+        Return the facet counts. Note: this will cause the query to run if it hasn't already.
+
+        Override the original implementation so that if we're forced to run the query, we can
+        cache the results that come back with it and avoid having to make another request to get
+        them later. Original implimentation:
```
rlucioni: s/implimentation/implementation

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
```diff
@@ -49,3 +49,25 @@ def distinct_count(self):
...
+            return self.query.get_facet_counts()
+        else:
+            # Force the query to run and fill the cache with the first page of results.
+            # This will cause the facet_counts to be cached along with the rest of the results
+            # and could potentially reduce the number of queries required to complete a search
+            # request.
+            #
+            # Note: If there are fewer than 10 results for the query, ES will simply return what it
+            # has found without raising an exception.
+            self[0:10]  # pylint: disable=pointless-statement
```
rlucioni: Is it necessary to use slicing like this to force queryset execution? Can you do something more conventional (and less arcane) like computing the length? For example:

```py
len(self)
```

Haystack's source appears to do this in several places. Also, is calling `_fill_cache()` directly an option? If so, that seems easier to understand.

course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
```diff
@@ -123,3 +124,37 @@ def test_facet_counts_includes_distinct_counts(self):
...
+            assert not mock_run.called
+            assert facet_counts == cached_facet_counts
+
+            # Calling count shouldn't result in another query, as we should have already cached it with the
+            # first request.
+            count = dc_queryset.count()
+            assert not mock_run.called
+            assert count == 3
+
+            # Fetching the results shouldn't result in another query, as weh should have already cached them
```
rlucioni: s/weh/we

course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
```diff
@@ -123,3 +124,37 @@ def test_facet_counts_includes_distinct_counts(self):
...
+            # with the initial request.
+            results = dc_queryset[:]
```
rlucioni: Is this necessary? Why can't you use `dc_queryset` directly?

course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
```diff
@@ -123,3 +124,37 @@ def test_facet_counts_includes_distinct_counts(self):
...
+            assert not mock_run.called
+            expected = sorted([run_1.key, run_2.key, run_3.key])
+            actual = sorted([run.key for run in results])
```
rlucioni: Can you use sets instead of sorting these lists?

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
amangano-edx: Calling _fill_cache directly is probably a better bet. I don't think `len()` caches the results.

amangano-edx:
On second thought, I think it might be better to do this in the view right before the serializer is created. This would allow us to populate the cache based on pagination params, which would reduce the amount of queries required for paginated requests.

course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
amangano-edx: This will evaluate the query if it hasn't been run already.

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
```diff
@@ -1,9 +1,12 @@
+from django.conf import settings
 from haystack.query import SearchQuerySet
 from course_discovery.apps.edx_haystack_extensions.distinct_counts.backends import DistinctCountsSearchQuery
 
 
 class DistinctCountsSearchQuerySet(SearchQuerySet):
     """Custom SearchQuerySet class that can compute and cache distinct hit and facet counts for a query."""
+    # The number of records that should be requested when warming the SearchQuerySet cache.
+    DEFAULT_QUERY_CACHE_WARMING_COUNT = 10
```
rlucioni: I'd suggest avoiding hardcoding defaults into application code when possible. The settings file should be the only place you need to look to understand how to configure the service. I'd rather the app blow up if required configuration is missing than make a silent assumption and continue to work.

course_discovery/settings/base.py
```diff
@@ -454,6 +454,11 @@
 DISTINCT_COUNTS_HIT_PRECISION = 1500
 DISTINCT_COUNTS_FACET_PRECISION = 250
 
+# The number of records that should be requested when warming the SearchQuerySet cache. Set this to equal the
+# number of records typically requested with each search query in order to reduce the number of queries that need
+# to be executed.
+DISTINCT_COUNTS_QUERY_CACHE_WARMING_COUNT = 9
```
rlucioni: 9 seems like an odd default. I know it's what we use on the marketing site, but the default should match the API's default page size, which I think is 20?

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
amangano-edx: I did use the settings file, but this is a fallback in case that is not set.

course_discovery/settings/base.py
amangano-edx: This isn't a default. This is an explicit setting that reflects how we use the api. I defined the default in the query class, which can be used if an explicit setting isn't found.

rlucioni: This is a default. It is the application's base settings file. If we need to customize a default (say, to use 9 instead of the default, 20), that should be done using the configuration repo.

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
rlucioni: I understand. I'm asking you to not hardcode fallbacks for settings into the application. If a setting like this is required, it should appear in base.py. If it's missing, I would rather the application not work than silently make an assumption about what I wanted.

course_discovery/apps/edx_haystack_extensions/distinct_counts/query.py
```diff
@@ -49,3 +50,26 @@ def distinct_count(self):
...
+            return self.query.get_facet_counts()
+        else:
+            # Force the query to run and fill the cache with the first page of results.
+            # This will cause the facet_counts to be cached along with the rest of the results
+            # and could potentially reduce the number of queries required to complete a search
+            # request.
+            #
+            # Note: If there are fewer than count results for the query, ES will simply return what it
+            # has found without raising an exception.
+            count = settings.DISTINCT_COUNTS_QUERY_CACHE_WARMING_COUNT
```
rlucioni: Minor: you don't need to assign this to a variable anymore.

course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
```diff
@@ -123,3 +124,37 @@ def test_facet_counts_includes_distinct_counts(self):
         hidden_count, hidden_distinct_count = facet_counts['queries']['hidden']
         assert hidden_count == 2
         assert hidden_distinct_count == 1
+
+    def test_facet_counts_caches_results(self):
+        """ Verify that facet_counts cache results when it is forced to run the query."""
+        course = CourseFactory()
+        run_1 = CourseRunFactory(title='foo', pacing_type='self_paced', hidden=True, course=course)
+        run_2 = CourseRunFactory(title='foo', pacing_type='self_paced', hidden=True, course=course)
+        run_3 = CourseRunFactory(title='foo', pacing_type='instructor_paced', hidden=False, course=course)
```
rlucioni: You can simplify this test by using a list; you don't seem to actually need to assign these runs to variables. If you do that, you could also easily use more than three course runs for testing, it there's a benefit to doing that.

course_discovery/settings/base.py
```diff
@@ -454,6 +454,11 @@
 DISTINCT_COUNTS_HIT_PRECISION = 1500
 DISTINCT_COUNTS_FACET_PRECISION = 250
 
+# The number of records that should be requested when warming the SearchQuerySet cache. Set this to equal the
+# number of records typically requested with each search query in order to reduce the number of queries that need
+# to be executed.
+DISTINCT_COUNTS_QUERY_CACHE_WARMING_COUNT = 20
```
rlucioni, line 460: Like you mentioned earlier, we typically request 9 results per page. Would it be a problem if we left this as 20, or is it important for us to override this setting?

amangano-edx: I think leaving it at 20 is less than ideal as we will end up loading 11 extra records from ES with every request. I will look into overriding it in the configuration repo, as you mentioned. I don't think it would be terrible to leave it at 20 though.

amangano-edx: We're going to stick with 20 for now, since any loss of efficiency from loading the extra 11 records should be offset by a gain in efficiency from only needing to do a single query when the second page of results is requested.

course_discovery/apps/edx_haystack_extensions/tests/test_distinct_counts/test_query.py
```diff
@@ -123,3 +124,39 @@ def test_facet_counts_includes_distinct_counts(self):
...
+            # Calling facet_counts again shouldn't result in an additional query
+            cached_facet_counts = dc_queryset.facet_counts()
+            assert not mock_run.called
+            assert facet_counts == cached_facet_counts
+
+            # Calling count shouldn't result in another query, as we should have already cached it with the
+            # first request.
+            count = dc_queryset.count()
+            assert not mock_run.called
+            assert count == 3
```
rlucioni:
```py
assert count == len(runs)
```

rlucioni (approved):

course_discovery/settings/base.py
mikedikan: can we make it dynamic based on the request?

mikedikan (approved):

course_discovery/settings/base.py
amangano-edx: To do that, we would need access to the pagination information. I tried doing that in https://github.com/edx/course-discovery/pull/762 but ultimately decided against that approach since I didn't have the time to implement the pagination logic generically.

PR #762, opened 2017-03-17
Proposal: Eliminate excess search queries in search facets views

amangano-edx:
Our search/facets endpoints currently execute three search queries for every one search request. When the drf_haystack facets action initializes its serializer, it calls SearchQuerySet.facet_counts, which executes the query but doesn't cache any of the results. Later, when the serializer creates a paginator object, SearchQuerySet.count is called, which executes another query and doesn't cache any results. Finally, when the paginator object requests the results for the page, a third query is run. This time, the results get cached but by this point three queries have already run.

This fix might not be the most elegant, but it is quick and would allow us to significantly reduce search load. 

@rlucioni

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import urllib.parse
+from unittest import mock
 
 import ddt
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/views/search.py
```diff
@@ -66,8 +66,53 @@ def facets(self, request):
...
+            if ":" not in facet:
+                continue
+
+            field, value = facet.split(":", 1)
+            if value:
+                queryset = queryset.narrow('%s:"%s"' % (field, queryset.query.clean(value)))
+
+        # Run the query and populate the cache before passing the queryset to the serializer, to prevent extra
+        # queries from running.
+        self.populate_queryset_cache(queryset, request)
```
amangano-edx, line 87: All of the code in this method, except for this one line, was borrowed from the superclass version. See https://github.com/inonit/drf-haystack/blob/master/drf_haystack/mixins.py#L47

course_discovery/apps/api/v1/views/search.py
```diff
@@ -66,8 +66,53 @@ def facets(self, request):
               items:
                 pytype: str
               required: false
+
+        Overrides and re-implement FacetsMixin.facets so that we can populate the queryset cache before passing it to
+        the serializer and avoid making extra search queries. Original implimentation:
```
rlucioni, line 71: s/implimentation/implementation

rlucioni (approved):
@amangano-edx nice. Is this something that can be contributed back to drf-haystack? If so, I would encourage you to go that route and push the maintainer to release an updated version of the package. Are there situations in which you wouldn't want to do this?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 import json
 import urllib.parse
+from unittest import mock
```
clintonb, line 4: `import mock`

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -437,6 +438,41 @@ def test_results_include_aggregation_key(self):
         )
         assert expected == actual
 
+    @ddt.data(
+        (5, {'q': 'cs'}, 5),
```
clintonb, line 442: De-dupe. `record_count` is always 5. `query_params` always has `{'q': 'cs'}`.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -437,6 +438,41 @@ def test_results_include_aggregation_key(self):
...
+        (5, {'q': 'cs', 'page': 1, 'page_size': 2}, 2),
+        (5, {'q': 'cs', 'page': 3, 'page_size': 2}, 1),
+        (5, {'q': 'cs', 'page': 'last', 'page_size': 5}, 5),  # 'last' can be used to request the last page of results
+    )
+    @ddt.unpack
+    def test_faceted_search_executes_a_single_query(self, record_count, query_params, expected_results_count):
+        """ Verify the view only needs to execute a single query to return results and facets. """
+        for __ in range(record_count):
+            course = CourseFactory(partner=self.partner)
+            CourseRunFactory(course=course, status=CourseRunStatus.Published, title='cs')
```
clintonb, line 455: `CourseRunFactory.create_batch(record_count, course__partner=self.partner, status=CourseRunStatus.Published, title='cs')`

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -437,6 +438,41 @@ def test_results_include_aggregation_key(self):
...
+
+        with mock.patch.object(self.backend, 'search', wraps=self.backend.search) as mock_search:
+            response = self.get_response(query_params)
+            assert response.status_code == 200
+            assert mock_search.call_count == 1
+            assert response.data['objects']['count'] == record_count
+            assert len(response.data['objects']['results']) == expected_results_count
+
+    @ddt.data(
+        {'q': 'cs', 'page': -1},
```
clintonb, line 465: De-dupe. The only data needed here is the value of `page`.

course_discovery/apps/api/v1/views/search.py
clintonb: Suggested redux: This method is copied exactly from `FacetMixin.facets()`. The only difference is the inclusion of the call to `self.populate_queryset_cache()`.

course_discovery/apps/api/v1/views/search.py
clintonb: Put this behind a Waffle switch in case we need to disable it.

clintonb (changes requested):
I will approve after Waffle switch added, and comments addressed.

amangano-edx:
Decided to go with the original approach after all, as this one assumes that the endpoints use PageNumberPagination.

https://github.com/edx/course-discovery/pull/759

PR #763, opened 2017-03-19
Optimize queries made by the catalog courses endpoint

rlucioni:
filter() and exclude() calls made to narrow the set of course runs after prefetching were causing prefetched data to be discarded. This resulted in the endpoint making many duplicate queries. These expensive duplicate queries are eliminated by prefetching the filtered set of course runs instead of prefetching all course runs and then trying to filter them.

ECOM-6473

I'll be giving the `CourseWithProgramsSerializer` a similar treatment in a subsequent PR. It makes the same mistake as the serializer corrected here: filtering a prefetched result instead of prefetching the filtered result directly.

@edx/ecommerce

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -387,8 +387,12 @@ class MinimalCourseRunSerializer(TimestampModelSerializer):
     seats = SeatSerializer(many=True)
 
     @classmethod
-    def prefetch_queryset(cls):
-        return CourseRun.objects.all().select_related('course').prefetch_related(
+    def prefetch_queryset(cls, queryset=None):
+        # Explicitly check for None to avoid returning all CourseRuns when the
+        # queryset passed in happens to be empty.
+        queryset = queryset if queryset is not None else CourseRun.objects.all()
+
+        return queryset.select_related('course').prefetch_related(
             'course__partner',
             Prefetch('seats', queryset=SeatSerializer.prefetch_queryset()),
         )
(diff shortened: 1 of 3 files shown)


schenedx (approved):
👍

course_discovery/apps/api/serializers.py
```diff
@@ -597,12 +602,34 @@ class CatalogCourseSerializer(CourseSerializer):
     """
     course_runs = serializers.SerializerMethodField()
 
+    @classmethod
```
mikedikan, line 605: Is the idea to add into the prefetch all of the possible things that might be used as filters later in the query?

rlucioni: @mikedikan not quite. The idea is to determine the filtered set required for serialization and prefetch that. Any `filter()` or `exclude()` performed after the fact will cause prefetched results to be discarded.

mikedikan: ahh, so if we build the prefetch using the proper fields it does the filtering for us?  Should we not ever be using queryset filters?

rlucioni: You should never use the queryset filters *after the queryset has been prefetched*. The correct thing to do is to apply those filters, then prefetch the results.

mikedikan:
Right, so either use the queryset filters exclusively, or use the prefetch to construct your query if the query is a performance issue - yes?

rlucioni:
@mikedikan yes. As long as you don't use the filters after prefetching.

PR #768, opened 2017-03-21
Fixed role widgets for parent course.

waheedahmed:
[ECOM-7500](https://openedx.atlassian.net/browse/ECOM-7500)

As a publisher user, I only want to see people that are involved in the approval process of a parent course, so that I don't get distracted with roles that are not involved in the approval of a course

**AC:**

1. Only display:
    - Course team
    - Marketing

screen shot 2017-03-21 at 2 14 50 pm


waheedahmed:

awais786 (approved):

PR #769, opened 2017-03-21
Don't show history widget for single object.

waheedahmed:
[ECOM-7501](https://openedx.atlassian.net/browse/ECOM-7501)

As a publisher user, I don't want to see the revision history module unless there is actual history so that I don't get confused with a wrong display.

**AC:**

- Validate that the revision history drop down is only shown when there is more than one version.


waheedahmed:

waheedahmed:
@bderusha Please review.

bderusha (approved):

tasawernawaz (approved):

PR #771, opened 2017-03-22
make current user default course team admin

tasawernawaz:
[ECOM-7496](https://openedx.atlassian.net/browse/ECOM-7496)


tasawernawaz:
@waheedahmed @awais786 Need your review.

PR #772, opened 2017-03-22
Course run submit to PC for review instead of MKTG.

waheedahmed:
[ECOM-7526](https://openedx.atlassian.net/browse/ECOM-7526)

As a CT admin, I should be sending Course runs for review to Project Coordinators instead of Marketing Reviewer.


waheedahmed:

tasawernawaz (approved):

PR #775, opened 2017-03-23
Added instructor marketing profile URL .

awais786:
On course run detail page Instructor 
Name is link now that links to the instructor page (on Drupal)
https://openedx.atlassian.net/browse/ECOM-7558

screen shot 2017-03-23 at 3 42 38 pm

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -228,6 +228,18 @@ class Person(TimeStampedModel):
 
         return url
 
+    @property
+    def profile_url(self):
+        url = None
+        marketing_site_url_root = self.partner.marketing_site_url_root
+        if marketing_site_url_root:
+            path = '{marketing_site_url_root}bio/{full_name}'.format(
+                marketing_site_url_root=marketing_site_url_root, full_name='-'.join((self.given_name, self.family_name))
+            )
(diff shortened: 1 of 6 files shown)


awais786:
@waheedahmed  @tasawernawaz Please review.

waheedahmed (approved):

awais786:
@bderusha please review.

awais786:
@bderusha please review.

course_discovery/apps/course_metadata/models.py
```diff
@@ -228,6 +228,18 @@ def get_profile_image_url(self):
 
         return url
 
+    @property
+    def profile_url(self):
```
bderusha, line 232: This is not always true.

```
url: "https://stage.edx.org/bio/piotr-mitros-0",
```

We need to actually get the URL from the node and store it

bderusha (changes requested):

course_discovery/apps/course_metadata/models.py
bderusha: The marketing team has the ability to change these URLs at will

awais786: @bderusha then what do u suggest?
This might need to be done in data-loaders.

PR #776, opened 2017-03-23
Temporarily remove updating index alias

MatthewPiatetsky:
@rlucioni @mikedikan @edx/ecommerce 

Renzo and I had the thought to make sure we don't update to a broken index and then test update index command with higher verbosity


rlucioni (approved):
The plan is to deploy this change and re-run the update index command with increased verbosity. This will help us decide if it's safe to re-enable the index update and delete jobs.

PR #777, opened 2017-03-23
ECOM-7572 Add sanity check before moving index alias

MatthewPiatetsky:
@rlucioni @mikedikan @edx/ecommerce 
thoughts on this approach?

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -1,5 +1,7 @@
 import logging
 
+from django.conf import settings
+from django.core.management import CommandError
 from haystack import connections as haystack_connections
 from haystack.management.commands.update_index import Command as HaystackCommand
 
@@ -10,6 +12,9 @@ logger = logging.getLogger(__name__)
 
 class Command(HaystackCommand):
     backends = []
+
+    def get_record_count(self, conn, index_name):
+        return conn.count(index_name).get('count')
 
     def handle(self, *items, **options):
         self.backends = options.get('using')
@@ -22,6 +27,7 @@ class Command(HaystackCommand):
         for backend_name in self.backends:
             connection = haystack_connections[backend_name]
             backend = connection.get_backend()
+            record_count = self.get_record_count(backend.conn, backend.index_name)
             alias, index_name = self.prepare_backend_index(backend)
             alias_mappings.append((backend, index_name, alias))
 
@@ -29,7 +35,33 @@ class Command(HaystackCommand):
 
         # Set the alias (from settings) to the timestamped catalog.
         for backend, index, alias in alias_mappings:
+            record_count_is_sane, index_info_string = self.sanity_check_new_index(backend.conn, index, record_count)
+            if not record_count_is_sane:
+                raise CommandError('Sanity check failed for new index. ' + index_info_string)
             self.set_alias(backend, alias, index)
+
+    def percentage_change(self, current, previous):
+        try:
+            return abs(current - previous) / previous
+        except ZeroDivisionError:
+            # pick large percentage for division by 0
+            # This is done to fail the sanity check
+            return 1
+
+    def sanity_check_new_index(self, conn, index, previous_record_count):
+        """ Ensure that we do not point to an index that looks like it has missing data. """
+        current_record_count = conn.count(index).get('count')
+        percentage_change = self.percentage_change(current_record_count, previous_record_count)
+
+        # Verify there was not a big shift in record count
+        record_count_is_sane = percentage_change < settings.INDEX_SIZE_CHANGE_THRESHOLD
+        index_info_string = (
+            'The previous index contained [{}] records. '
+            'The new index contains [{}] records, a [{:.2f}%] change.'.format(
+                previous_record_count, current_record_count, percentage_change * 100
+            )
+        )
+        return record_count_is_sane, index_info_string
 
     def set_alias(self, backend, alias, index):
         """
--- a/course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
@@ -1,17 +1,24 @@
+import mock
+import pytest
 from django.conf import settings
-from django.core.management import call_command
-from django.test import TestCase
+from django.core.management import CommandError, call_command
+from django.test import TestCase, override_settings
 from elasticsearch import Elasticsearch
 from freezegun import freeze_time
 
+from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
+from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory
 from course_discovery.apps.edx_haystack_extensions.tests.mixins import SearchIndexTestMixin
 
 
-class UpdateIndexTests(SearchIndexTestMixin, TestCase):
+@override_settings(HAYSTACK_SIGNAL_PROCESSOR='haystack.signals.BaseSignalProcessor')
+class UpdateIndexTests(ElasticsearchTestMixin, SearchIndexTestMixin, TestCase):
     @freeze_time('2016-06-21')
     def test_handle(self):
         """ Verify the command creates a timestamped index and repoints the alias. """
-        call_command('update_index')
+        with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                        'update_index.Command.sanity_check_new_index', return_value=(True, '')):
+            call_command('update_index')
 
         alias = settings.HAYSTACK_CONNECTIONS['default']['INDEX_NAME']
         index = '{alias}_20160621_000000'.format(alias=alias)
@@ -27,3 +34,32 @@ class UpdateIndexTests(SearchIndexTestMixin, TestCase):
             }
         }
         self.assertDictEqual(response, expected)
+
+    def test_sanity_check_error(self):
+        """ Verify the command raises a CommandError if new index fails the sanity check. """
+        CourseRunFactory()
+        record_count = 2
+        additional_runs = int(100 * settings.INDEX_SIZE_CHANGE_THRESHOLD + 1)
+        CourseRunFactory.create_batch(additional_runs)
+
+        # Ensure that an error is raised if the sanity check does not pass
+        with pytest.raises(CommandError):
+            with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                            'update_index.Command.set_alias', return_value=True):
+                with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                                'update_index.Command.get_record_count', return_value=record_count):
+                    call_command('update_index')
+
+    def test_sanity_check_success(self):
+        """ Verify the command does not raise a CommandError error if the new index passes the sanity check. """
+        CourseRunFactory.create_batch(30)
+        record_count = 60
+        additional_runs = int(10 * settings.INDEX_SIZE_CHANGE_THRESHOLD - 1)
+        CourseRunFactory.create_batch(additional_runs)
+
+        # Ensure that no error is raised and the sanity check passes the second time
+        with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                        'update_index.Command.set_alias', return_value=True):
+            with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                            'update_index.Command.get_record_count', return_value=record_count):
+                call_command('update_index')
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -435,6 +435,10 @@ HAYSTACK_CONNECTIONS = {
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.BaseSignalProcessor'
 HAYSTACK_INDEX_RETENTION_LIMIT = 3
 
+# Update Index Settings
+# Make sure the size of the new index does not change by more than this percentage
+INDEX_SIZE_CHANGE_THRESHOLD = .1
+
 # Elasticsearch search query facet "size" option to increase from the default value of "100"
 # See  https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-facets-terms-facet.html#_accuracy_control
 SEARCH_FACET_LIMIT = 10000


course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +28,36 @@ def handle(self, *items, **options):
         for backend_name in self.backends:
             connection = haystack_connections[backend_name]
             backend = connection.get_backend()
+            record_count = backend.conn.count(backend.index_name).get('count')
```
MatthewPiatetsky: I'm not sure how to tell if our prod elasticsearch exposes this count endpoint.  
If it does not, we could just do a search and look at hits.get('total')

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -1,12 +1,18 @@
...
 from haystack.management.commands.update_index import Command as HaystackCommand
 
 from course_discovery.apps.core.utils import ElasticsearchUtils
 
 logger = logging.getLogger(__name__)
 
+# Make sure the new index has at least this many records
+SANITY_CHECK_COUNT = 1000
+# Make sure the size of the new index does not change by more than this percentage
+SANITY_CHECK_PERCENTAGE_CHANGE = 10
```
rlucioni: This percentage check on its own should be sufficient. I don't think there's value in overcomplicating this and also checking the record count.

Please make this a float (so you don't need to multiply by 100 - it's standard to see percentage settings specified as floats) and make it a proper setting by moving it to base.py.

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
mikedikan: http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es_version_1_5.html

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +28,36 @@ def handle(self, *items, **options):
...
+        except ZeroDivisionError:
+            raise CommandError('The current index is empty')
+
+    def sanity_check_new_index(self, conn, index, previous_record_count):
+        """ Ensure that we do not point to an index that looks like it has missing data. """
+        current_record_count = conn.count(index).get('count')
+        percentage_change = self.percentage_change(current_record_count, previous_record_count)
+
+        # Verify there was not a big shift in record count
+        condition1 = percentage_change < SANITY_CHECK_PERCENTAGE_CHANGE
```
rlucioni: Give this a useful name. For example, `is_record_count_sane`, or something along those lines.

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +28,36 @@ def handle(self, *items, **options):
...
+        # Verify that the record count is not unexpectedly low
+        condition2 = current_record_count > SANITY_CHECK_COUNT
+
+        if not (condition1 and condition2):
+            raise CommandError('Sanity check failed for new index. Previous count: '
+                               '{} Current Count: {}'.format(previous_record_count, current_record_count))
```
rlucioni: Could you also print the percentage change here?

rlucioni (changes requested):

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
mikedikan: Sorry, that used to have a table showing the available endpoints, it doesn't seem to any longer, let me see if I can find you that.

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
```diff
@@ -13,7 +14,9 @@ class UpdateIndexTests(SearchIndexTestMixin, TestCase):
     @freeze_time('2016-06-21')
     def test_handle(self):
         """ Verify the command creates a timestamped index and repoints the alias. """
-        call_command('update_index')
+        with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                        'update_index.Command.sanity_check_new_index', return_value=True):
```
rlucioni: @MatthewPiatetsky you do plan to write tests for the sanity check, right? We can't trust it to work like this without tests.

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
mikedikan: http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-supported-es-operations.html

mikedikan: It does appear that _count is a supported operation

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +24,40 @@ def handle(self, *items, **options):
         for backend_name in self.backends:
             connection = haystack_connections[backend_name]
             backend = connection.get_backend()
-            alias, index_name = self.prepare_backend_index(backend)
+            record_count = backend.conn.count(backend.index_name).get('count')
+            alias = connection.options['INDEX_NAME']
```
MatthewPiatetsky: it seems if i do backend.index_name which this was previously doing, and run the update_index command twice in my test, the second time it will return the current index_name (catalog_test_20170323_194951), not the alias catalog_test (which causes set_alias to fail because an index with that name already exists).

If we didn't want to do this i could mock out set_alias to pass on the second run of this command in the test

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
MatthewPiatetsky: see the tests i added, i added a comment about one

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +24,41 @@ def handle(self, *items, **options):
...
         # Set the alias (from settings) to the timestamped catalog.
         for backend, index, alias in alias_mappings:
+            self.sanity_check_new_index(backend.conn, index, record_count)
             self.set_alias(backend, alias, index)
 
+    def percentage_change(self, current, previous):
+        try:
+            return abs(current - previous) / previous
+        except ZeroDivisionError:
+            raise CommandError('The current index is empty')
```
rlucioni: This seems too defensive. Is it really a problem if the current index is empty and you're trying to replace it with a populated index? I would pass in this case.

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +24,41 @@ def handle(self, *items, **options):
...
+
+    def sanity_check_new_index(self, conn, index, previous_record_count):
```
rlucioni: This is a side-effecting function. A more Pythonic approach is to name this something like `is_new_index_sane()` and return a Boolean (i.e., return `percentage_change < settings.SANITY_CHECK_PERCENTAGE_CHANGE`), then use that in the `handle()` method to decide whether to raise a `CommandError` or set the alias.

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
rlucioni: @MatthewPiatetsky if I'm understanding correctly, you're making this change on account of a test? If so, I would prefer to see the tests mock as necessary rather than change the code under test.

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
```diff
@@ -27,3 +32,22 @@ def test_handle(self):
             }
         }
         self.assertDictEqual(response, expected)
+
+    def test_sanity_check_error(self):
+        """ Verify the command raises a CommandError if the previous index is empty. """
```
rlucioni: I don't see a test verifying that the command refuses to set the alias if the sanity check doesn't pass when the previous index is non-empty. That's the more likely case.

rlucioni (changes requested):

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +24,41 @@ def handle(self, *items, **options):
...
+
+        # Verify there was not a big shift in record count
+        record_count_is_sane = percentage_change < settings.SANITY_CHECK_PERCENTAGE_CHANGE
+
+        if not record_count_is_sane:
+            raise CommandError(
+                'Sanity check failed for new index. '
+                'The previous index contained [{}] records. '
+                'The new index contains [{}] records, a [{}%] change.'.format(
+                    previous_record_count, current_record_count, percentage_change * 100
```
rlucioni: You should round the percentage to the nearest two decimal places when printing it. For example:
```py
>>>  '{0:.2f}%'.format(0.123456*100)
'12.35%'
```

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
MatthewPiatetsky: that seems strange. pass if the previous index is empty but fail if it has one record?

course_discovery/settings/base.py
```diff
@@ -435,6 +435,10 @@
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.BaseSignalProcessor'
 HAYSTACK_INDEX_RETENTION_LIMIT = 3
 
+# Update Index Settings
+# Make sure the size of the new index does not change by more than this percentage
+SANITY_CHECK_PERCENTAGE_CHANGE = .1
```
rlucioni: The name for this setting can make its purpose a little more clear. Suggestion: `INDEX_SIZE_CHANGE_THRESHOLD`.

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
rlucioni: Why would the check fail if the previous index had one record? I just don't see how failing because the previous index is empty is useful here. If I'm trying to use this command to populate a new index, and I happen to have an empty index in place, why refuse to update the index?

MatthewPiatetsky: this check is just if the previous index is empty. With the current code if there was any number of records that did not meet the threshold, including fewer records, it would fail.  
i don't think we should always assume more records is right

MatthewPiatetsky:
@rlucioni did I address your comments?

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
```diff
@@ -27,3 +34,32 @@ def test_handle(self):
...
+        with pytest.raises(CommandError):
+            with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                            'update_index.Command.set_alias', return_value=True):
+                with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                                'update_index.Command.get_record_count', return_value=record_count):
+                    call_command('update_index')
+
+    def test_sanity_check_success(self):
+        """ Verify the command does not raise a CommandError error if the new index passes the sanity check. """
+        CourseRunFactory.create_batch(100)
```
rlucioni: Is it really necessary to create 100 runs? Could you do this with an order of magnitude less?

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
rlucioni: Can you update this docstring to reflect the behavior the test is covering now?

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
```diff
@@ -27,3 +34,32 @@ def test_handle(self):
...
+        """ Verify the command raises a CommandError if the previous index is empty. """
+        CourseRunFactory()
+        record_count = 2
+        additional_runs = int(100 * settings.INDEX_SIZE_CHANGE_THRESHOLD + 1)
+        CourseRunFactory.create_batch(additional_runs)
+
+        # Ensure that an error is raised if the sanity check does not pass
+        with pytest.raises(CommandError):
+            with mock.patch('course_discovery.apps.edx_haystack_extensions.management.commands.'
+                            'update_index.Command.set_alias', return_value=True):
```
rlucioni, line 48: Why is it necessary to mock `set_alias` here? Aren't you expecting the sanity check to fail, meaning this method will never be called?

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -22,15 +27,41 @@ def handle(self, *items, **options):
...
+            record_count_is_sane, index_info_string = self.sanity_check_new_index(backend.conn, index, record_count)
+            if not record_count_is_sane:
+                raise CommandError('Sanity check failed for new index. ' + index_info_string)
             self.set_alias(backend, alias, index)
 
+    def percentage_change(self, current, previous):
+        try:
+            return abs(current - previous) / previous
+        except ZeroDivisionError:
+            # pick large percentage for division by 0
```
rlucioni, line 47: I'd elaborate a bit on your choice here to make it clear that this is done to force the sanity check to fail.

rlucioni (approved):
@MatthewPiatetsky looks good, though I would make a few changes to the tests.

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
MatthewPiatetsky: well if i did 10 then 10 * .1 - 1 will be 0 additional courses. 
I can do 30.

course_discovery/apps/edx_haystack_extensions/tests/test_update_index.py
MatthewPiatetsky: due to this issue i mentioned earlier where i had to change update_index if i didn't do this, because it was trying to make the alias for the new index the name of the existing index which would break. the important part here though is that the sanity check passes/fails.  
What happens after doesn't matter as much

PR #778, opened 2017-03-23
Update comment reference to point to current documentation

mikedikan:
@edx/ecommerce for review


rlucioni (approved):
@mikedikan this is a change that could merge without review. Thanks for updating this.

PR #780, opened 2017-03-24
WIP Filter program list by partner short code

douglashall:
Allows for filtering of the program list API endpoint by
partner short code using a partner_short_code query parameter.

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -156,6 +156,7 @@ class CourseRunFilter(FilterSetMixin, django_filters.FilterSet):
 
 class ProgramFilter(FilterSetMixin, django_filters.FilterSet):
     marketable = django_filters.MethodFilter()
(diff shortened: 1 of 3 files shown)


clintonb:
The service has pretty much everything it needs to facilitate multi-tenancy except updated API endpoints. It would be preferable to complete that functionality rather than add filtering ad hoc.

douglashall:
@clintonb I'm not sure what you mean. Do you have a document that explains how you expect multi-tenancy to work for the catalog service? At the moment, I'm most interested in understanding how you would expect an Open EdX site to display programs specific to that site.

clintonb:
All of the data is linked to a partner. What is missing for multi-tenancy is linking the partners to sites, and automatically filtering the data at the API endpoints by the requesting site's partner. If we do that, we solve this problem all of the data—courses, programs, organizations, people.

douglashall:
@clintonb Awesome. Have you talked with @mjfrey about planning that work? It will be required for the program certs enablement.

clintonb:
No. I am not actively working on certificates.

PR #787, opened 2017-03-28
[LEARNER-275] Enable eligible program functionality

iivic:
Ticket: https://openedx.atlassian.net/browse/LEARNER-275

admin-program-filter
admin-one-click-purchase-field


course_discovery/apps/course_metadata/utils.py
```diff
@@ -8,6 +8,39 @@
...
+    Returns:
+        bool: Indicator for the check result.
+    """
+    for course in program.courses.all():
+        if course.course_runs.count() != 1:
+            return False
+    return True
+
+
+def is_program_eligible(program):
```
clintonb: This should be a property on the model. Also, eligible for what? The name is too ambiguous.

course_discovery/apps/course_metadata/models.py
```diff
@@ -728,6 +728,7 @@ class Program(TimeStampedModel):
         help_text=_('The description of credit redemption for courses in program'),
         blank=True, null=True
     )
+    one_click_purchase = models.BooleanField(default=False)
```
clintonb: Vague name is vague. one_click_purchase_enabled

course_discovery/apps/course_metadata/utils.py
```diff
@@ -8,6 +8,39 @@
...
+    """
+    Helper method for checking if the program contains only courses with one course run.
+
+    Args:
+        program (Program): Program instance for check.
+
+    Returns:
+        bool: Indicator for the check result.
+    """
+    for course in program.courses.all():
```
clintonb: Use prefetch to get all data in a single DB call. Also, you need to account for excluded course runs. @rlucioni can help with this if you need it.

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,45 @@
...
+            if is_program_eligible(program):
+                program_ids.add(program.id)
+        else:
+            if not is_program_eligible(program):
+                program_ids.add(program.id)
+    return queryset.filter(pk__in=program_ids)
+
+
+class ProgramEligibilityFilter(admin.SimpleListFilter):
+    title = _('is eligible for one click purchase')
```
clintonb: Eligible for one-click purchase

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,45 @@
...
+    parameter_name = 'is_eligible'
```
clintonb: eligible_for_one_click_purchase

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,45 @@
+from django.contrib import admin
+from django.utils.translation import ugettext_lazy as _
+
+from course_discovery.apps.course_metadata.utils import is_program_eligible
+
+
+def _get_program_queryset(queryset, display_eligible_programs):
```
clintonb: This seems like it should be a method in the class below. It should be the overridden `queryset()` method.

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Args:
+        queryset (queryset): Queryset of programs that should be filtered.
+        display_eligible_programs (bool): Indicates how the queryset should be filtered.
+
+    Returns:
+        queryset of filtered programs.
+    """
+    program_ids = set()
+    for program in queryset:
```
clintonb: Again, look at prefetch options. As-is this is O(n^3) DB queries.

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,45 @@
...
+    parameter_name = 'is_eligible'
+
+    def lookups(self, request, model_admin):
+        return (
+            (1, _('Yes')),
+            (0, _('No'))
+        )
+
+    def queryset(self, request, queryset):
+        if self.value() == '1':
```
clintonb:
```python
value = bool(int(self.value()))
return _get_program_queryset(queryset, value)
```

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,45 @@
+from django.contrib import admin
+from django.utils.translation import ugettext_lazy as _
```
clintonb: No need for `_lazy`.

clintonb (changes requested):

course_discovery/apps/course_metadata/admin.py
```diff
@@ -92,7 +93,7 @@ class ProgramAdmin(admin.ModelAdmin):
     fields = (
         'title', 'subtitle', 'status', 'type', 'partner', 'banner_image', 'banner_image_url', 'card_image_url',
         'marketing_slug', 'overview', 'credit_redemption_overview', 'video', 'weeks_to_complete',
-        'min_hours_effort_per_week', 'max_hours_effort_per_week',
+        'min_hours_effort_per_week', 'max_hours_effort_per_week', 'one_click_purchase'
     )
     fields += (
```
clintonb: Please merge this with the tuple above. Not sure why this is here.

course_discovery/apps/course_metadata/utils.py
clintonb: * https://docs.djangoproject.com/en/1.10/ref/models/querysets/#select-related
* https://docs.djangoproject.com/en/1.10/ref/models/querysets/#prefetch-related

course_discovery/apps/course_metadata/utils.py
```diff
@@ -8,6 +8,39 @@
...
+        if course.course_runs.count() != 1:
+            return False
+    return True
+
+
+def is_program_eligible(program):
+    """
+    Checks if the program is eligible for one click purchase.
+    To pass the check the program must have one_click_purchase field enabled and
+    all his courses must contain only one course run.
```
rlucioni: s/his/its

course_discovery/apps/course_metadata/utils.py
rlucioni: Good catch @clintonb. @iivic please use the Django debug toolbar to understand how the query counts for the program list and detail views are impacted by this change. Most of the serializers, including the `MinimalProgramSerializer`, include a [`prefetch_queryset()`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L671) method. Any prefetching should occur in this method. Any manipulation of data should happen after that method is called so that you're working with cached data. Be careful not to perform any queryset filtering after prefetching; doing that will cause the cached data to be discarded.

rlucioni: Follow-up: in this case, it looks like the serializer may already be prefetching course runs for you. Regardless, you should use the debug toolbar to make sure you understand how query counts are changing. Unit tests should fail if you cause the query count to change.

course_discovery/apps/api/serializers.py
```diff
@@ -683,7 +684,7 @@ class Meta:
         model = Program
         fields = (
             'uuid', 'title', 'subtitle', 'type', 'status', 'marketing_slug', 'marketing_url', 'banner_image',
-            'courses', 'authoring_organizations', 'card_image_url',
+            'courses', 'authoring_organizations', 'card_image_url', 'is_eligible_for_one_click_purchase'
```
clintonb: Add a trailing comma.

course_discovery/apps/api/serializers.py
```diff
@@ -665,6 +665,7 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     authoring_organizations = MinimalOrganizationSerializer(many=True)
     banner_image = StdImageSerializerField()
     courses = serializers.SerializerMethodField()
+    is_eligible_for_one_click_purchase = serializers.SerializerMethodField()
```
clintonb: There should be no need to declare this here. If you must declare it, declare it as a Boolean field that is read-only.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -81,7 +82,7 @@ class ProgramAdmin(admin.ModelAdmin):
     form = ProgramAdminForm
     inlines = [FaqsInline, IndividualEndorsementInline, CorporateEndorsementsInline]
     list_display = ('id', 'uuid', 'title', 'type', 'partner', 'status',)
-    list_filter = ('partner', 'type', 'status',)
+    list_filter = ('partner', 'type', 'status', ProgramEligibilityFilter)
```
clintonb: Add a trailing comma.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -92,9 +93,7 @@ class ProgramAdmin(admin.ModelAdmin):
     fields = (
         'title', 'subtitle', 'status', 'type', 'partner', 'banner_image', 'banner_image_url', 'card_image_url',
         'marketing_slug', 'overview', 'credit_redemption_overview', 'video', 'weeks_to_complete',
-        'min_hours_effort_per_week', 'max_hours_effort_per_week',
-    )
-    fields += (
+        'min_hours_effort_per_week', 'max_hours_effort_per_week', 'one_click_purchase_enabled',
```
clintonb: Avoid messing up users' current workflows. Move the new field to the bottom of the page/end of the list.

course_discovery/apps/course_metadata/models.py
```diff
@@ -728,12 +728,39 @@ class Program(TimeStampedModel):
...
+        """
+        Helper method for checking if the program contains only courses with one course run.
+
+        Args:
+            self (Program): Program instance for check.
+
+        Returns:
+            bool: Indicator for the check result.
+        """
+        for course in self.courses.all():
```
clintonb: Move this to the method below.

course_discovery/apps/course_metadata/models.py
```diff
@@ -728,12 +728,39 @@ class Program(TimeStampedModel):
...
+            if course.course_runs.count() != 1:
```
clintonb: Remember to account for excluded course runs.

clintonb (changes requested):
If you don't make changes to Tex in JS files, there is no need to commit updated djangojs translation files.

course_discovery/apps/course_metadata/models.py
iivic: Can you elaborate? Currently it counts all course runs that are linked to the course regardless of the program excluded course runs list.

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,32 @@
...
+        The queryset can be filtered to contain programs that are eligible for
+        one click purchase or to exclude them.
+        """
+        value = self.value()
+        if value is None:
+            return queryset
+
+        program_ids = set()
+        for program in queryset.prefetch_related('courses__course_runs'):
+            if value == '1':
```
vkaracic: I don't see ``value`` changing with each (any?) iteration so maybe avoid doing the same operation in each iteration.

vkaracic: I don't see value changing with each (any?) iteration so maybe avoid doing the same operation in each iteration.

course_discovery/apps/course_metadata/models.py
clintonb: Let's say you encounter a course with two course runs. If one of those course runs is excluded, the course run count should for that course should be one, not two.

iivic: I asked how to count the course runs on our standup yesterday. Maybe I wasn't clear but this is exactly what I was referring to. I'll inform the team on todays standup.

course_discovery/apps/course_metadata/models.py
```diff
@@ -728,12 +728,39 @@ class Program(TimeStampedModel):
...
+        return True
+
+    @property
+    def is_program_eligible_for_one_click_purchase(self):
+        """
+        Checks if the program is eligible for one click purchase.
+        To pass the check the program must have one_click_purchase field enabled and
+        all its courses must contain only one course run.
+        """
+        if self.one_click_purchase_enabled and self._program_course_and_course_run_mapping():
```
vkaracic: You don't need the rest of the code in this property, just do `` return elf.one_click_purchase_enabled and self._program_course_and_course_run_mapping()``.

course_discovery/apps/course_metadata/models.py
clintonb: I am technically out of the office all week, so I haven't been at standup. It's easier for me to have a conversation either here on the PR or on the original ticket in JIRA.

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,28 @@
...
+        """
+        The queryset can be filtered to contain programs that are eligible for
+        one click purchase or to exclude them.
+        """
+        value = self.value()
+        if value is None:
+            return queryset
+
+        program_ids = set()
+        for program in queryset.prefetch_related('courses__course_runs'):
```
clintonb: Move this to a separate line: `queryset = queryset.prefetch_related('courses__course_runs')`

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,28 @@
+from django.contrib import admin
+from django.utils.translation import ugettext as _
+
+
+class ProgramEligibilityFilter(admin.SimpleListFilter):
```
clintonb: This seems more at home in `admin.py`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -728,12 +728,29 @@ class Program(TimeStampedModel):
         help_text=_('The description of credit redemption for courses in program'),
         blank=True, null=True
     )
+    one_click_purchase_enabled = models.BooleanField(default=False)
```
clintonb: Add help text: Allow allow courses in this program to be purchased in a single transaction.

course_discovery/apps/course_metadata/models.py
```diff
@@ -728,12 +728,29 @@ class Program(TimeStampedModel):
...
+        """
+        Checks if the program is eligible for one click purchase.
+        To pass the check the program must have one_click_purchase field enabled and
+        all its courses must contain only one course run.
+        """
+        if not self.one_click_purchase_enabled:
+            return False
+
+        for course in self.courses.all():
+            if course.course_runs.exclude(id__in=self.excluded_course_runs.values_list('id', flat=True)).count() != 1:
```
clintonb: This will most likely result in a database call. You should do the filtering in Python:

```python
excluded_course_run_ids = set(self.excluded_course_runs.values_list('id', flat=True))

for course in self.courses.all():
  course_run_ids = set(course.course_runs.values_list('id', flat=True)) - excluded_course_run_ids
  if len(course_run_ids) != 1
    return False
```

course_discovery/conf/locale/eo/LC_MESSAGES/djangojs.po
```diff
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-03-31 15:15+0500\n"
```
clintonb: You aren't changing any JS text, so there is no need to commit changes to `djanogojs.*`.

course_discovery/apps/course_metadata/admin_filters.py
```diff
@@ -0,0 +1,28 @@
...
+    title = _('eligible for one-click purchase')
```
clintonb: Capitalize: Eligible

course_discovery/apps/course_metadata/admin_filters.py
clintonb: This also needs tests.

clintonb (changes requested):

course_discovery/apps/course_metadata/admin_filters.py
clintonb: Please ensure the tests also ensure a linear (relative to the number of programs) query count.

course_discovery/apps/course_metadata/admin_filters.py
iivic: "By" will get prepended so capitalising would look silly. You can see how it looks in the description screenshot.

course_discovery/apps/course_metadata/admin_filters.py
iivic: What is the benefit of doing so?

clintonb: The code is clearer to read.

iivic:
You can review one more time.

course_discovery/apps/course_metadata/migrations/0050_program_one_click_purchase_enabled.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('course_metadata', '0049_courserun_eligible_for_financial_aid'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='program',
+            name='one_click_purchase_enabled',
+            field=models.BooleanField(default=False),
```
clintonb: This needs to be updated with the help text.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,50 @@ def test_program_update(self):
...
+        self.program_filter = ProgramEligibilityFilter(None, {}, None, None)
+        self.course_runs = factories.CourseRunFactory.create_batch(4)
+        self.courses = [course_run.course for course_run in self.course_runs]
+        self.one_click_purchase_eligible_program = factories.ProgramFactory(
+            courses=self.courses,
+            one_click_purchase_enabled=True
+        )
+        self.one_click_purchase_ineligible_program = factories.ProgramFactory(courses=self.courses)
+        self.programs = [self.one_click_purchase_ineligible_program, self.one_click_purchase_eligible_program]
+        self.programs_queryset = Program.objects.filter(id__in=[pr.id for pr in self.programs])
```
clintonb: Just use `Program.objects.all()` where this variable is used.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,50 @@ def test_program_update(self):
...
+    def setUp(self):
+        super(ProgramEligibilityFilterTests, self).setUp()
+        self.program_filter = ProgramEligibilityFilter(None, {}, None, None)
+        self.course_runs = factories.CourseRunFactory.create_batch(4)
+        self.courses = [course_run.course for course_run in self.course_runs]
+        self.one_click_purchase_eligible_program = factories.ProgramFactory(
+            courses=self.courses,
+            one_click_purchase_enabled=True
+        )
+        self.one_click_purchase_ineligible_program = factories.ProgramFactory(courses=self.courses)
```
clintonb: This doesn't need to be an instance variable. The same goes for pretty much all of these variables. Move the data to the tests where they actually belong. This will make the tests much easier to read, and avoid introducing dependencies that make adding/modifying tests difficult in the future.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -405,6 +405,22 @@ def create_program_with_seats(self):
...
+        self.assertTrue(self.program.is_program_eligible_for_one_click_purchase)
+
+    def test_one_click_purchase_ineligible(self):
+        """ Verify that program is one click purchase ineligible. """
+        # This is the case when one_click_purchase_enabled is set to False or a
+        # course with multiple course runs is included in a program or both.
+        self.program.one_click_purchase_enabled = False
+        self.assertFalse(self.program.is_program_eligible_for_one_click_purchase)
+
+        self.program.one_click_purchase_enabled = True
```
clintonb: You need more conditions/tests and comments. Conditions I can think of (that are not tested):

1. Course has multiple course runs.
2. Course has multiple course runs, and all but one is excluded from the program.
3. One course has multiple runs, other courses have one.
4. Course runs are not enrollable/purchaseable.

clintonb (changes requested):

course_discovery/apps/course_metadata/tests/test_models.py
iivic: 4. Is this a condition that should be meet by an eligible program? If not how is it connected to these tests? What is required for a course run to be enrollable, enrollment_start/end dates?

clintonb: We don't want to have a one-click purchase scenario where the learner ends up with a fraction of the program. The simplest method of determining availability will be to look at the upgrade deadlines on the eligibility seats.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -405,6 +405,54 @@ def create_program_with_seats(self):
...
+        factories.CourseRunFactory(course=program.courses.first())
+        self.assertFalse(program.is_program_eligible_for_one_click_purchase)
+
+        # Program has one_click_purchase_enabled set to True and
+        # contains one course with multiple course runs
+        course_run = factories.CourseRunFactory()
+        factories.CourseRunFactory(course=course_run.course)
+        program = factories.ProgramFactory(courses=[course_run.course], one_click_purchase_enabled=True)
+        self.assertFalse(program.is_program_eligible_for_one_click_purchase)
+
```
clintonb: You still need to test for the case when a course run is excluded from the program.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,48 @@ def test_program_update(self):
...
+        course_run = factories.CourseRunFactory()
+        one_click_purchase_eligible_program = factories.ProgramFactory(
+            courses=[course_run.course],
+            one_click_purchase_enabled=True
+        )
+        one_click_purchase_ineligible_program = factories.ProgramFactory(courses=[course_run.course])
+        with self.assertNumQueries(1):
+            self.assertEqual(
+                list(program_filter.queryset({}, Program.objects.all())),
+                [one_click_purchase_ineligible_program, one_click_purchase_eligible_program]
```
clintonb, line 375: `Program.objects.all()`

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,48 @@ def test_program_update(self):
...
+        """ Verify that one click purchase eligible programs pass the filter. """
+        program_filter = ProgramEligibilityFilter(None, {self.parameter_name: 1}, None, None)
+        course_run = factories.CourseRunFactory()
+        one_click_purchase_eligible_program = factories.ProgramFactory(
+            courses=[course_run.course],
+            one_click_purchase_enabled=True
+        )
+        with self.assertNumQueries(8):
+            self.assertEqual(
+                program_filter.queryset({}, Program.objects.all()).first(),
```
clintonb: Drop `first()`. There shouldn't be any more objects returned. Cast to a `list` as you did above.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,48 @@ def test_program_update(self):
...
+                one_click_purchase_eligible_program
+            )
+
+    def test_queryset_method_returns_ineligible_programs(self):
+        """ Verify that one click purchase ineligible programs pass the filter. """
```
clintonb: Verify programs ineligible for one-click purchase do not pass the filter.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,48 @@ def test_program_update(self):
...
+        program_filter = ProgramEligibilityFilter(None, {self.parameter_name: 0}, None, None)
+        course_run = factories.CourseRunFactory()
```
clintonb: These two lines can be simplified: `program = factories.ProgramFactory(one_click_purchase_enabled=False)`.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -349,3 +350,48 @@ def test_program_update(self):
...
+        one_click_purchase_ineligible_program = factories.ProgramFactory(courses=[course_run.course])
+        with self.assertNumQueries(6):
+            self.assertEqual(
+                program_filter.queryset({}, Program.objects.all()).first(),
```
clintonb: Same comment as above regarding `first()`.

course_discovery/apps/course_metadata/tests/test_admin.py
iivic: An additional database call will be made after this change. I prefer leaving it as is.

course_discovery/apps/course_metadata/models.py
```diff
@@ -437,6 +438,28 @@ def get_paid_seat_enrollment_end(self):
...
+            2. There is a verified seat linked to the course run
+            3. The verified seats upgrade deadline has not passed
+        """
+        if self.enrollment_start and self.enrollment_start < timezone.now():
+            return False
+
+        if self.enrollment_end and timezone.now() < self.enrollment_end:
+            return False
+
+        verified_seats = self.seats.filter(type=Seat.VERIFIED)
```
clintonb: Calling `filter()` here will result in an extra call to the DB that should be avoided by using Python filtering.

course_discovery/apps/course_metadata/models.py
```diff
@@ -437,6 +438,28 @@ def get_paid_seat_enrollment_end(self):
         return deadline
 
     @property
+    def is_purchasable(self):
+        """
+        Course run is purchasable when it satisfies the following requirements:
+            1. The course run is open for enrollment
```
clintonb: You also need to account for whether the course run itself is open. See `start` and `end`.

course_discovery/apps/course_metadata/models.py
clintonb: We don't want to hard-code `Seat.VERIFIED` here. You might want to modify the method signature:

```python
def enrollable_seats(self, types):
  Returns seats, of the given type(s), that can be enrolled in/purchased.

  Arguments:
    types (SeatType[]): Type of seats to limit the returned value to.

  Returns:
    SeatType[]
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -743,16 +766,17 @@ def __str__(self):
...
+        not excluded course run must be purchasable.
         """
         if not self.one_click_purchase_enabled:
             return False
 
         excluded_course_run_ids = set(self.excluded_course_runs.values_list('id', flat=True))
         for course in self.courses.all():
             course_run_ids = set(course.course_runs.values_list('id', flat=True)) - excluded_course_run_ids
-            if len(course_run_ids) != 1:
+            if len(course_run_ids) != 1 or not CourseRun.objects.get(id=course_run_ids.pop()).is_purchasable:
```
clintonb: This is going to result in multiple calls to the database that should be avoided.

```python
excluded_course_runs = set(self.excluded_course_runs)

for course in self.courses.all():
  course_runs = set(course.course_runs) - excluded_course_runs

  if len(course_runs) != 1:
    return False

  if not course_runs.pop().is_purchaseable:
    return False
```

clintonb (commented):
This is a great purchaseable property. It needs some tweaks before it's good to go.

course_discovery/apps/course_metadata/models.py
iivic: Regarding the filter() I need to get the seats from the database in both cases. I don't see how to avoid that.

clintonb: When using `prefetch`/`select_related`, you have all of the data cached after the first call. If you need to filter, do so in Python (rather than the DB/Django ORM) to avoid a call to the database.

```python
verified_seats = [seat if seat.type == Seat.VERIFIED for seat in self.seats.all()]
```

iivic:
@clintonb Please review the latest changes.

iivic:
@clintonb @marjev

marjev (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -436,6 +437,45 @@ def get_paid_seat_enrollment_end(self):
...
+
+        if self.enrollment_start and self.enrollment_start > now:
+            return enrollable_seats
+
+        if self.enrollment_end and now > self.enrollment_end:
+            return enrollable_seats
+
+        for seat in self.seats.all():
+            if seat.type in types:
+                if not seat.upgrade_deadline or now < seat.upgrade_deadline:
```
clintonb: Combine with the statement above.

course_discovery/apps/course_metadata/models.py
```diff
@@ -436,6 +437,45 @@ def get_paid_seat_enrollment_end(self):
...
+                    enrollable_seats.append(seat)
+
+        return enrollable_seats
+
+    @property
+    def is_purchasable(self):
+        """
+        Indicates weather or not the course run contains a purchasable seat
```
clintonb: whether

course_discovery/apps/course_metadata/models.py
```diff
@@ -580,6 +620,8 @@ class Seat(TimeStampedModel):
     PROFESSIONAL = 'professional'
     CREDIT = 'credit'
 
+    PURCHASABLE_SEAT_TYPES = [PROFESSIONAL, VERIFIED]
```
clintonb: This is not needed.

course_discovery/apps/course_metadata/models.py
```diff
@@ -436,6 +437,45 @@ def get_paid_seat_enrollment_end(self):
...
+
+        for seat in self.seats.all():
+            if seat.type in types:
+                if not seat.upgrade_deadline or now < seat.upgrade_deadline:
+                    enrollable_seats.append(seat)
+
+        return enrollable_seats
+
+    @property
+    def is_purchasable(self):
```
clintonb: This is not needed.

clintonb (changes requested):
`CourseRun.is_purchaseable()` is not needed.

iivic:
@clintonb Comments addressed.

course_discovery/apps/course_metadata/models.py
```diff
@@ -729,12 +761,41 @@ class Program(TimeStampedModel):
...
+
+        excluded_course_runs = set(self.excluded_course_runs.all())
+
+        for course in self.courses.all():
+            course_runs = set(course.course_runs.all()) - excluded_course_runs
+
+            if len(course_runs) != 1:
+                return False
+
+            if not course_runs.pop().enrollable_seats([Seat.VERIFIED, Seat.PROFESSIONAL]):
```
clintonb: Stop hard-coding seat types. Use self.type.applicable_seat_types.

clintonb (changes requested):

course_discovery/apps/course_metadata/models.py
iivic: I understood that the the seat type must be buyable (learner must pay for it). That's the reason I hardcoded verified and professional because those are buyable. `applicable_seat_types` can be any seat type so this is different from what I was doing.

clintonb: Applicable seat types is what matters. If, in the future, we decide to allow audit seats or some new seat type, we should not need to modify this code.

marjev: So `self.type.applicable_seat_types` currently doesn't contain non-purchasable seat types like Audit?

clintonb: Nope. Even if it did, that would be okay, because that's what the creators deemed applicable for completing a program.

course_discovery/apps/course_metadata/models.py
```diff
@@ -463,6 +463,9 @@ def enrollable_seats(self, types):
             return enrollable_seats
 
         for seat in self.seats.all():
+            print(types)
```
clintonb: Remove these print statements.

course_discovery/apps/course_metadata/models.py
```diff
@@ -784,14 +787,15 @@ def is_program_eligible_for_one_click_purchase(self):
...
+        applicable_seat_types = [seat_type.name for seat_type in self.type.applicable_seat_types.all()]
 
         for course in self.courses.all():
             course_runs = set(course.course_runs.all()) - excluded_course_runs
 
             if len(course_runs) != 1:
                 return False
 
-            if not course_runs.pop().enrollable_seats([Seat.VERIFIED, Seat.PROFESSIONAL]):
+            if not course_runs.pop().enrollable_seats(applicable_seat_types):
```
clintonb: `list(self.type.applicable_seat_types.all())`

iivic: This will return a list of SeatType objects and thus not work.

course_discovery/apps/course_metadata/models.py
```diff
@@ -436,6 +437,37 @@ def get_paid_seat_enrollment_end(self):
 
         return deadline
 
+    def enrollable_seats(self, types):
+        """
+        Returns seats, of the given type(s), that can be enrolled in/purchased.
+
+        Arguments:
+            types (list of SeatTypes): Type of seats to limit the returned value to.
```
clintonb: list of seat type names

clintonb (approved):
See my last comment on the docstring. Remember to squash your commits.

iivic:
@clintonb FYI I noticed that the applicable_seat_types are uppercase and added https://github.com/edx/course-discovery/pull/787/files#diff-a79b6f6b0e3ecfd5c59847f0e39d59d2R787

PR #788, opened 2017-03-28
Increase Haystack connection timeout

rlucioni:
This was previously unspecified, causing us to use the default of 10 seconds. We've seen sporadic Elasticsearch timeouts while updating new indices. In an attempt to resolve this, we've lowered the batch size used for bulk updates. Increasing this timeout should further help reduce timeouts.

LEARNER-370

### DIFF
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -427,6 +427,7 @@ HAYSTACK_CONNECTIONS = {
         'ENGINE': 'course_discovery.apps.edx_haystack_extensions.backends.EdxElasticsearchSearchEngine',
         'URL': 'http://localhost:9200/',
         'INDEX_NAME': 'catalog',
+        'TIMEOUT': 30,
     },
 }
 


clintonb:
Why do we have ES requests lasting more than 10s?

rlucioni:
@clintonb great question. I don't have a good answer for you beyond the fact that we were doing bulk updates 1000 records at a time. See these [console logs](https://tools-edx-jenkins.edx.org/job/Catalog/job/prod-edx-update-index/35/console) if you're interested. I'm open to other ideas for addressing this if you have them. This change is coupled with a smaller batch size, introduced in https://github.com/edx/edx-internal/pull/26.

clintonb:
Is this change required? It seems logical to test the bulk change first, and only merge this if that change does not resolve the issues.

rlucioni:
@clintonb sure, that seems reasonable. I'll keep this open so the change is queued up if we need it.

rlucioni:
The batch size change alone seems to have been sufficient. Closing.

PR #798, opened 2017-04-03
Upgraded to jQuery 3.2.1

clintonb:
This avoids a security vulnerability in older versions.

LEARNER-456

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -13,7 +13,7 @@
     "clipboard": "1.5.12",
     "datatables": "1.10.12",
     "edx-pattern-library": "~0.18.0",
-    "jquery": "^3.1.1",
+    "jquery": "^3.2.1",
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/fedx-team

attritionorg:
Upgrading from 3.1.1 to 3.2.1, can you clarify which jQuery vulnerability this addresses?

clintonb:
https://openedx.atlassian.net/wiki/display/EdxOps/jQuery+XSS+vulnerability+discovery+and+recommendations

This service was actually not affected since we have always used jQuery 3.x.

attritionorg:
Unfortunately that requires authentication. If it references the recent: https://nodesecurity.io/advisories/329, that is fixed by 1.9.0 and presumably doesn't impact later trees. That is what made me curious.

PR #801, opened 2017-04-04
Sync person profile url from drupal.

awais786:
1. Add new field in people model.
2. Update data-loader and sync the profile url field.

ECOM-7587

screen shot 2017-04-04 at 4 46 25 pm


awais786:
@tasawernawaz @waheedahmed @bderusha Please review.

course_discovery/apps/course_metadata/models.py
```diff
@@ -205,6 +205,7 @@ class Person(TimeStampedModel):
         },
     )
     slug = AutoSlugField(populate_from=('given_name', 'family_name'), editable=True)
+    profile_url = models.URLField(null=True, blank=True)
```
waheedahmed, line 208: Do we also need a default value for this?

waheedahmed (approved):
Just one question, otherwise 👍

course_discovery/apps/course_metadata/models.py
awais786: It has null=True so it will not effect the current records. 
@clintonb your thoughts ?

clintonb: No default is needed.

clintonb (approved):

PR #804, opened 2017-04-04
Display applicable seat types in serializer

vkaracic:
We need the applicable seat types information exposed so that on the marketing site we know which seats' SKUs to use to create the link that would add the seats to the user basket.
https://openedx.atlassian.net/browse/LEARNER-475

@clintonb could this solve our problem?

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -820,6 +820,7 @@ class ProgramSerializer(MinimalProgramSerializer):
     )
     subjects = SubjectSerializer(many=True)
     staff = PersonSerializer(many=True)
+    applicable_seat_types = serializers.SerializerMethodField()
 
     @classmethod
     def prefetch_queryset(cls):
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -848,14 +849,18 @@ def prefetch_queryset(cls):
             Prefetch('individual_endorsements', queryset=EndorsementSerializer.prefetch_queryset()),
         )
 
+    def get_applicable_seat_types(self, obj):
+        types = ProgramType.objects.get(program=obj).applicable_seat_types.all()
```
clintonb: `return obj.program_type.applicable_seat_types.all().values_list('slug', flat=True)`

clintonb (commented):
Yes, this is on the right track.

course_discovery/apps/api/serializers.py
```diff
@@ -848,14 +849,17 @@ def prefetch_queryset(cls):
             Prefetch('individual_endorsements', queryset=EndorsementSerializer.prefetch_queryset()),
         )
 
+    def get_applicable_seat_types(self, obj):
+        return list(obj.type.applicable_seat_types.values_list('slug', flat=True))
```
vkaracic, line 853: I've changed this to explicitly be a list and not a queryset so that I don't have to change every single test in test_serializers.py.

vkaracic:
@clintonb this is ready for a code review.

clintonb (approved):

vkaracic:
Someone else from @edx/helio please review.

mjfrey (approved):

PR #808, opened 2017-04-06
 Add download image url for staff

awais786:
Adding download images url.
Adding copy paste links.

https://openedx.atlassian.net/browse/ECOM-7555


awais786:
@roderickmorales I am adding download image link and adding copy icons to the name and organisations and position.
Also we need to add a indication this is newly added image and not synced with drupal. 
So i need some design guide lines i tried to add icons but every thing looks very messy. 

screen shot 2017-04-06 at 6 46 03 pm

PR #809, opened 2017-04-06
LEARNER-498 Fix marketing site client login process

MatthewPiatetsky:
When you logged in to the marketing site, you used to be taken to the page with the same url, but now you are redirected straight to the admin (so the publishing code was logging in successfully but the expected_url was wrong).
Not sure if someone changed this

@edx/learner-growth 

[](https://openedx.atlassian.net/browse/LEARNER-498)

### DIFF
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -96,12 +96,10 @@ class MarketingSiteAPIClient(object):
             'op': 'Log in',
         }
         response = session.post(login_url, data=login_data)
-        expected_url = '{root}/users/{username}'.format(root=self.api_url, username=self.username)
         admin_url = '{root}/admin'.format(root=self.api_url)
-        # Temporary way of checking whether the user has been logged into marketing site until
-        # the marketing site login flow is fixed
-        can_access_admin = session.get(admin_url)
-        if not (can_access_admin.status_code == 200 and response.url == expected_url):
+        # This is not a RESTful API so checking the status code is not enough
+        # We also check that we were redirected to the admin page
+        if not (response.status_code == 200 and response.url == admin_url):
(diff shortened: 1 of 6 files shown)


course_discovery/apps/course_metadata/utils.py
```diff
@@ -96,12 +96,11 @@ def init_session(self):
             'op': 'Log in',
         }
         response = session.post(login_url, data=login_data)
-        expected_url = '{root}/users/{username}'.format(root=self.api_url, username=self.username)
         admin_url = '{root}/admin'.format(root=self.api_url)
         # Temporary way of checking whether the user has been logged into marketing site until
         # the marketing site login flow is fixed
         can_access_admin = session.get(admin_url)
-        if not (can_access_admin.status_code == 200 and response.url == expected_url):
+        if not (can_access_admin.status_code == 200 and response.url == admin_url):
```
rlucioni: Why does the `admin_url` need to be checked? I think you can just check the status code.

rlucioni (changes requested):

course_discovery/apps/course_metadata/utils.py
MatthewPiatetsky: I think the issue is, this is a form not a restful API, so if the form renders, but your login failed for instance, response code will stil be 200 (the can_access_admin is redundant at this point though so will remove that)

rlucioni: I see, a comment to that effect would be nice.

clintonb (approved):

rlucioni (approved):

mikedikan:
This is why this happened:

https://openedx.atlassian.net/browse/ECOM-6496

mikedikan (approved):

rlucioni:
Nice find @mikedikan.

PR #819, opened 2017-04-11
Displaying error message on course-edit page in case of error.

awais786:
https://openedx.atlassian.net/browse/ECOM-7666

screen shot 2017-04-11 at 2 47 19 pm

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -398,6 +398,11 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
         messages.success(self.request, _('Course  updated successfully.'))
         return HttpResponseRedirect(self.get_success_url())
 
+    def form_invalid(self, form):
+        # pylint: disable=no-member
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -398,6 +398,10 @@ def form_valid(self, form):
         messages.success(self.request, _('Course  updated successfully.'))
         return HttpResponseRedirect(self.get_success_url())
 
+    def form_invalid(self, form):
+        messages.error(self.request, _('Please fill all required fields.'))
```
waheedahmed: Translations does not have this message, I think you need to re-generate.

waheedahmed: Also quality will fail, you need to disable pylint no-member for this `# pylint: disable=no-member`

waheedahmed (approved):
Couple of comments, 👍 once addressed.

srpearce:
@awais786 Please change the error message to the following:

Make sure to provide all required information.

PR #821, opened 2017-04-11
Added flag to disable record change limit for update_index command

clintonb:
This limit makes sense for existing installations, but not so much for
new ones (e.g. a fresh devstack). This flag allows us to disable the
check as needed.


clintonb:
FYI @edx/learner @edx/docker-devstack-working-group

PR #822, opened 2017-04-12
Remove html tags before comparing course fields

tasawernawaz:
[ECOM-7694](https://openedx.atlassian.net/browse/ECOM-7694)

* Only text will be compared
* If value is "Not yet added" then don't compare.

**Course Detail Page**

screen shot 2017-04-12 at 2 20 21 pm

**Revision History Page**
screen shot 2017-04-12 at 3 34 08 pm


tasawernawaz:
Please review this. @waheedahmed @awais786

awais786 (approved):

PR #824, opened 2017-04-12
Publish course run status changes to the marketing site

rlucioni:
Generalizes program publication code so that it can be applied to any model, including course runs.

LEARNER-419

@edx/learner-growth FYI


rlucioni:
@MatthewPiatetsky I'm still finishing up unit tests, but this is very close to being done.

course_discovery/apps/course_metadata/exceptions.py
```diff
@@ -1,15 +1,37 @@
...
-    """ The exception thrown during the program publishing process to marketing site """
+class MarketingSitePublisherException(Exception):
+    pass
 
-    def __init__(self, message):
-        super(ProgramPublisherException, self).__init__(message)
-        suffix = 'The program data has not been saved. Please check your marketing site configuration'
-        self.message = '{exception_msg} {suffix}'.format(exception_msg=message, suffix=suffix)
+
+class AliasCreateError(MarketingSitePublisherException):
```
MatthewPiatetsky, line 9: Is there a reason we need all of these different errors?

course_discovery/apps/course_metadata/signals.py
```diff
@@ -3,12 +3,16 @@
...
+from course_discovery.apps.course_metadata.publishers import ProgramMarketingSitePublisher
 
 
 @receiver(pre_delete, sender=Program)
 def delete_program(sender, instance, **kwargs):  # pylint: disable=unused-argument
-    if waffle.switch_is_active('publish_program_to_marketing_site') and \
-            instance.partner.has_marketing_site:
-        publisher = MarketingSitePublisher()
-        publisher.delete_program(instance)
+    is_publishable = (
```
MatthewPiatetsky, line 11: is this publishing when a program is deleted (what about when course runs are deleted or when course runs/programs are saved

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -1,254 +1,82 @@
...
-
-    @responses.activate
-    def test_publish_delete_xseries(self):
-        self.program = ProgramFactory(type=ProgramType.objects.get(name='XSeries'))
-        publisher = MarketingSitePublisher()
-        publisher.delete_program(self.program)
-        self.assert_responses_call_count(0)
+        pass
+
+    def test_edit_node(self):
```
MatthewPiatetsky: Are you still working on these tests?

rlucioni: No, I think `pass` gives us sufficient coverage.



(Yes, I still need to finish these.)

course_discovery/apps/course_metadata/exceptions.py
rlucioni: Clarity. I think this is preferable to the sledgehammer-like approach of using the same exception everywhere and changing the associated message each time. It's easier to debug problems if the exception raised is specific to the problem.

course_discovery/apps/course_metadata/signals.py
rlucioni: This only triggers when programs are deleted. Look at the save methods in `models.py` to see what happens when courses and programs are saved.

MatthewPiatetsky: what about when courses are deleted?

course_discovery/apps/course_metadata/exceptions.py
MatthewPiatetsky: but you're not actually calling these anywhere atm right?  
I'm not sure if I agree regarding debugging.  
If we can see a stacktrace of where the error happened, it seems that would give us the same information but in a clearer way. (but i don't care that much so w/e)

course_discovery/apps/course_metadata/signals.py
rlucioni: Nothing happens. The goal of the story is only to publish course run status when their status changes. Publishing other course run data is out of scope.

course_discovery/apps/course_metadata/exceptions.py
rlucioni: I am raising all of them in specific places. Look at `publishers.py`.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -1,171 +1,377 @@
...
+
+        response = self.client.api_session.post(self.node_api_base, data=node_data)
 
-    def _create_node(self, api_client, node_data):
-        node_url = '{root}/node.json'.format(root=api_client.api_url)
-        response = api_client.api_session.post(node_url, data=json.dumps(node_data))
         if response.status_code == 201:
-            response_json = response.json()
-            return response_json['id']
+            # TODO: Should this be nid?
```
MatthewPiatetsky: you have a todo here

course_discovery/apps/course_metadata/signals.py
MatthewPiatetsky: i meant course runs rather than courses, but wouldn't a course run being deleted count as a change in status?

rlucioni: No, they're separate. Publication status is independent of deletion. If you delete a course run and we mark it unpublished on the marketing site, the data loaders will re-create the run. If we delete the run from the marketing site, too, we're making workflow changes that aren't being asked for (yet).

rlucioni:
@MatthewPiatetsky this is ready to go!

MatthewPiatetsky (approved):
Seems good to me, assuming you've tested that publishing new programs and publishing a change to course run/program still works

rlucioni:
@MatthewPiatetsky all verified.

PR #828, opened 2017-04-14
LEARNER-799 Add more data to search results

MatthewPiatetsky:
fyi @edx/learner-growth

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -140,6 +140,7 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     announcement = indexes.DateTimeField(model_attr='announcement', null=True)
     min_effort = indexes.IntegerField(model_attr='min_effort', null=True)
     max_effort = indexes.IntegerField(model_attr='max_effort', null=True)
+    weeks_to_complete = indexes.IntegerField(model_attr='weeks_to_complete', null=True)
     language = indexes.CharField(null=True, faceted=True)
     transcript_languages = indexes.MultiValueField(faceted=True)
     pacing_type = indexes.CharField(model_attr='pacing_type', null=True, faceted=True)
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -815,7 +814,7 @@ def is_program_eligible_for_one_click_purchase(self):
 
         return True
 
-    @cached_property
+    @property
```
MatthewPiatetsky: This was causing tests to fail when i added these fields to the index

rlucioni: Do you know why? What is the cost of removing this?

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -235,6 +239,22 @@ class ProgramIndex(BaseIndex, indexes.Indexable, OrganizationsMixin):
...
     published = indexes.BooleanField(null=False, faceted=True)
+    min_hours_effort_per_week = indexes.IntegerField(null=True)
+    max_hours_effort_per_week = indexes.IntegerField(null=True)
+    weeks_to_complete_min = indexes.IntegerField(null=True)
+    weeks_to_complete_max = indexes.IntegerField(null=True)
+
+    def prepare_min_hours_effort_per_week(self, obj):
+        return obj.min_hours_effort_per_week
+
+    def prepare_max_hours_effort_per_week(self, obj):
```
rlucioni: Do you need these `prepare` methods if you're not doing anything special to the fields before adding them to the index?

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: I think it might have something to do with the program object being indexed before the course object or something? I'm not entirely sure. The cost is that this method will be run twice for the weeks_to_complete_min and weeks_to_complete_max

rlucioni: What was the nature of the test failures you saw? Do you have a stacktrace? It'd be nice to keep this, if possible, and I'd prefer not to see us making code changes like these solely on account of tests.

MatthewPiatetsky: ok i changed it back you can see the error in the travis log

MatthewPiatetsky: any ideas what causes this error? i believe the first time the property is hit during the real time indexing it is empty for some reason and then when hit again during the request it is cached

rlucioni: I'd have to dig in to get an answer I'm confident in, but it sounds like you could be on the right track. I would try invalidating the cached property in your test setup as explained [here](http://stackoverflow.com/a/23490487).

MatthewPiatetsky: thanks doing that fixed it

rlucioni (approved):

PR #838, opened 2017-04-18
Catalog API pagination URL domains should be consistant

adeelkhan:
ECOM-5122

Generates pagination url based on gateway passed HOST header params instead of underlying hostname/server e.g. stage-edx-discovery.

### DIFF
--- a/acceptance_tests/config.py
+++ b/acceptance_tests/config.py
@@ -1,5 +1,9 @@
 import os
 
+API_COURSES_ROOT = os.environ.get('API_COURSES_ROOT')
+if not API_COURSES_ROOT:
+    raise RuntimeError('API_COURSE_ROOT (e.g. https://api.stage.edx.org/) must be supplied!')
+
 API_GATEWAY_CATALOG_ROOT = os.environ.get('API_GATEWAY_CATALOG_ROOT')
 if not API_GATEWAY_CATALOG_ROOT:
     raise RuntimeError('API_GATEWAY_CATALOG_ROOT (e.g. https://api.stage.edx.org/catalog/v1) must be supplied!')
--- a/acceptance_tests/test_xforwarding_host.py
+++ b/acceptance_tests/test_xforwarding_host.py
@@ -1,0 +1,50 @@
+""" Tests to validate X-Forwarded-Host is enable. """
+import requests
+from django.test import TestCase
+
+from acceptance_tests.config import API_ACCESS_TOKEN, API_COURSES_ROOT
+
+
+class XForwardedHostTest(TestCase):
+    forwarded_host_name = 'api.stage.edx.org'
+    path = 'api/v1/courses'
+
+    def get_discovery_api_url(self, path):
+        """ Returns a complete URL for the given path. """
+        return '{root}/{path}'.format(root=API_COURSES_ROOT.rstrip('/'), path=path)
+
+    def get_expected_paginated_url(self):
+        return "http://{forwarded_host_name}/{path}/".format(forwarded_host_name=self.forwarded_host_name,
+                                                             path=self.path)
+
+    def assert_pagination_url(self, path, expected_status_code=200, **headers):
+        """
+        Verify the API returns HTTP 200 and next and previous pagination url
+        generated with forwarded_host_name.
+
+        Arguments:
+            path(str) -- Path of the API endpoint to call.
+            expected_status_code (int) -- Expected HTTP status code of the API response.
+            headers (dict) -- Headers to pass with the request.
+        """
+        url = self.get_discovery_api_url(path)
+        response = requests.get(url, headers=headers)
+        courses_list = response.json()
+
+        self.assertEqual(response.status_code, expected_status_code)
+
+        if courses_list['next']:
+            self.assertEqual(courses_list['next'].split('?')[0], self.get_expected_paginated_url())
+
+        if courses_list['previous']:
+            self.assertEqual(courses_list['previous'].split('?')[0], self.get_expected_paginated_url())
+
+    def test_xforwarded_generated_pagination(self):
+        """
+        Verify the endpoint returns HTTP 200 for request and generates
+        correct pagination url based on x-forwarded-host header param. """
+        headers = {
+            'Authorization': 'JWT {token}'.format(token=API_ACCESS_TOKEN),
+            'X-Forwarded-Host': self.forwarded_host_name
+        }
+        self.assert_pagination_url(self.path, **headers)
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -18,6 +18,7 @@ DEBUG = False
 
 ALLOWED_HOSTS = []
 
+USE_X_FORWARDED_HOST = True
 # Application definition
 
 INSTALLED_APPS = [


course_discovery/settings/local.py
```diff
@@ -1,7 +1,8 @@
 from course_discovery.settings.base import *
 
 DEBUG = True
-ALLOWED_HOSTS = ['*']
+ALLOWED_HOSTS = ['api.stage.edx.org']
```
clintonb: This should never be in our code base.

course_discovery/settings/local.py
```diff
@@ -1,7 +1,8 @@
...
+USE_X_FORWARDED_HOST = True
```
clintonb: This belongs in base.py. Local settings will *never* be read in production.

clintonb (changes requested):

course_discovery/settings/local.py
schenedx: What are you trying to achieve by making this change?

course_discovery/settings/local.py
schenedx: Please move this line to  https://github.com/edx/course-discovery/blob/master/course_discovery/settings/base.py#L20

schenedx (changes requested):
Please see my comments

adeelkhan:
@simon I just wanted to share changes with you for testing purpose. Therefore I hardcoded the ALLOWED_HOST value. This wasn’t meant for external review. After agreeing on changes I would ask for further reviews. Sorry for not mentioning this before.

schenedx (approved):
👍

ahsan-ul-haq (approved):

schenedx:
@clintonb Since you are back, can you review this PR quickly? One line change only

clintonb (changes requested):
I would like to see a test to confirm this will work as expected before we deploy it.

clintonb:
By test, I mean unit/acceptance test. I want to ensure that we don't accidentally break this functionality in the future.

acceptance_tests/config.py
```diff
@@ -1,6 +1,10 @@
 import os
 
-API_GATEWAY_CATALOG_ROOT = os.environ.get('API_GATEWAY_CATALOG_ROOT')
+API_COURSES_ROOT = os.environ.get('API_COURSES_ROOT', 'http://localhost:18381')
```
ahsan-ul-haq: Put a default value for stage not your local.

acceptance_tests/config.py
```diff
@@ -1,6 +1,10 @@
...
+if not API_COURSES_ROOT:
+    raise RuntimeError('API_COURSE_ROOT (e.g. https://api.stage.edx.org/api/v1/courses) must be supplied!')
```
ahsan-ul-haq: As you are putting a default value this block of code will not run.

acceptance_tests/config.py
```diff
@@ -1,6 +1,10 @@
...
+
+API_GATEWAY_CATALOG_ROOT = os.environ.get('API_GATEWAY_CATALOG_ROOT','http://localhost:18381')
 if not API_GATEWAY_CATALOG_ROOT:
     raise RuntimeError('API_GATEWAY_CATALOG_ROOT (e.g. https://api.stage.edx.org/catalog/v1) must be supplied!')
```
ahsan-ul-haq: Same here

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,49 @@
+""" Tests to validate X-Forwarded-Host is enable. """
+import ddt
+import requests
+from django.test import TestCase
+
+from acceptance_tests.config import API_ACCESS_TOKEN, API_COURSES_ROOT
+
+
+@ddt.ddt
```
ahsan-ul-haq: what is the use of ddt here?

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,49 @@
...
+class XForwardedHostTest(TestCase):
+    forwarded_host_name = 'api.stage.edx.org'
+    path = 'api/v1/courses/'
```
ahsan-ul-haq: why `api/v1/courses/` instead of `api/v1`

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,49 @@
...
+            path(str) -- Path of the API endpoint to call.
+            expected_status_code (int) -- Expected HTTP status code of the API response.
+            headers (dict) -- Headers to pass with the request.
+        """
+        url = self.get_discovery_api_url(path)
+        response = requests.get(url, headers=headers)
+        courses_list = response.json()
+        self.assertEqual(response.status_code, expected_status_code)
+
+        if courses_list['next'] is not None:
```
ahsan-ul-haq: you can simplify it to 
```
if not courses_list['next']:
```

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,49 @@
...
+            self.assertEqual(courses_list['next'].split('?')[0],
+                             "http://{forwarded_host_name}/{path}".format(forwarded_host_name=self.forwarded_host_name,
+                                                                          path=self.path))
```
ahsan-ul-haq: why not using `get_discovery_api_url` here?

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if courses_list['previous'] is not None:
```
ahsan-ul-haq: Same here.

ahsan-ul-haq (changes requested):

acceptance_tests/test_xforwarding_host.py
awais786: reverse('api:v1:course-list')

schenedx: Please take in this feedback

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,49 @@
...
+
+from acceptance_tests.config import API_ACCESS_TOKEN, API_COURSES_ROOT
+
+
+class XForwardedHostTest(TestCase):
+    forwarded_host_name = 'api.stage.edx.org'
+    path = 'api/v1/courses/'
+
+    def get_discovery_api_url(self, path):
+        """ Returns a complete URL for the given path. """
```
schenedx: @adeelkhan I feel like you can generalize this function by doing `Returns a complete URL for given path and root`. This way, you can use the caller code to specify the domain, and then the path.
This way, can you reuse this function in the `assertEqual` statements below

schenedx:
@clintonb You think the new acceptance test works?

adeelkhan:
@ahsan-ul-haq @clintonb Please review acceptance test.

acceptance_tests/config.py
```diff
@@ -1,5 +1,9 @@
 import os
 
+API_COURSES_ROOT = os.environ.get('API_COURSES_ROOT')
```
clintonb, line 3: This is not necessary given `API_GATEWAY_CATALOG_ROOT` below.

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,50 @@
...
+from django.test import TestCase
+
+from acceptance_tests.config import API_ACCESS_TOKEN, API_COURSES_ROOT
+
+
+class XForwardedHostTest(TestCase):
+    forwarded_host_name = 'api.stage.edx.org'
+    path = 'api/v1/courses'
+
+    def get_discovery_api_url(self, path):
```
clintonb, line 12: `get_api_gateway_url`

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,50 @@
+""" Tests to validate X-Forwarded-Host is enable. """
+import requests
+from django.test import TestCase
+
+from acceptance_tests.config import API_ACCESS_TOKEN, API_COURSES_ROOT
+
+
+class XForwardedHostTest(TestCase):
+    forwarded_host_name = 'api.stage.edx.org'
```
clintonb, line 9: This should not be hard-coded.

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,50 @@
...
+    path = 'api/v1/courses'
+
+    def get_discovery_api_url(self, path):
+        """ Returns a complete URL for the given path. """
+        return '{root}/{path}'.format(root=API_COURSES_ROOT.rstrip('/'), path=path)
+
+    def get_expected_paginated_url(self):
+        return "http://{forwarded_host_name}/{path}/".format(forwarded_host_name=self.forwarded_host_name,
```
clintonb, line 17: This should basically be `self.get_discovery_api_url(self.path)`.

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,50 @@
...
+
+        self.assertEqual(response.status_code, expected_status_code)
+
+        if courses_list['next']:
+            self.assertEqual(courses_list['next'].split('?')[0], self.get_expected_paginated_url())
+
+        if courses_list['previous']:
+            self.assertEqual(courses_list['previous'].split('?')[0], self.get_expected_paginated_url())
+
+    def test_xforwarded_generated_pagination(self):
```
clintonb, line 42: This belongs in `test_api_gateway.py`.

clintonb: This can be simplified:

```python
from edx_rest_api_client.client import EdxRestApiClient

def test_pagination_urls(self):
    client = EdxRestApiClient(API_GATEWAY_CATALOG_ROOT, jwt=API_ACCESS_TOKEN)
    response = client.catalogs(CATALOG_ID).courses.get()

    pagination_urls = (response.get('next'), response.get('previous'))
    pagination_urls = [url for url in pagination_urls if url]

    if not pagination_urls:
        self.fail('The API response does not include pagination URLs.')

    for url in pagination_urls:
        self.assertTrue(url.startswith(API_GATEWAY_CATALOG_ROOT))

```

acceptance_tests/test_xforwarding_host.py
```diff
@@ -0,0 +1,50 @@
+""" Tests to validate X-Forwarded-Host is enable. """
+import requests
+from django.test import TestCase
+
+from acceptance_tests.config import API_ACCESS_TOKEN, API_COURSES_ROOT
+
+
+class XForwardedHostTest(TestCase):
+    forwarded_host_name = 'api.stage.edx.org'
+    path = 'api/v1/courses'
```
clintonb, line 10: This path is not exposed by the API gateway. Only a subset of catalog paths are exposed, as described in https://github.com/edx/course-discovery/blob/master/api.yaml:

* /catalogs/
* /catalogs/{id}/
* /catalogs/{id}/courses/

clintonb (changes requested):
This will not work as expected.

Take `https://prod-edx-discovery.edx.org/api/v1/catalogs/`. The gateway path is `https://api.stage.edx.org/catalog/v1/catalogs`. Simply changing the host will result a URL pointing to path that does not exist, breaking pagination for all clients.

PR #839, opened 2017-04-19
Script to fix orgs

clintonb:


clintonb:
@schenedx this script should do what we want. I'll let you take over from here, if you're in the office tomorrow.

FYI @edx/devops we'll want to run this in a Django shell after review

schenedx:
@clintonb Can you please review my edits?

clintonb:
@schenedx your changes look good to me.

clintonb:
👍

PR #845, opened 2017-04-20
Auto populate existing run and seat data on new run form.

waheedahmed:
[ECOM-7732](https://openedx.atlassian.net/browse/ECOM-7732)

As a CT Admin, when I create a new Course-run, I would like all fields from the previous run to be copied over, except the dates, lms_course_id and preview_url fields, so that I can save time creating new runs.


waheedahmed:

course_discovery/apps/publisher/views.py
```diff
@@ -472,19 +473,37 @@ def get_parent_course(self):
...
+        try:
+            latest_run = parent_course.course_runs.latest('created')
+            initial_run_data = model_to_dict(latest_run)
+            del initial_run_data['id'], initial_run_data['start'], initial_run_data['end']
+            del initial_run_data['preview_url'], initial_run_data['lms_course_id'], initial_run_data['changed_by']
+
+            latest_seat = latest_run.seats.latest('created')
+            initial_seat_data = model_to_dict(latest_seat)
+            del initial_seat_data['id'], initial_seat_data['course_run'], initial_seat_data['changed_by']
+        except (CourseRun.DoesNotExist, Seat.DoesNotExist):
```
tasawernawaz: Is this possible? because every course has at least one seat and course run.

I think it will be better to handle exception separately for both course run and seat because if seat is missing at least course run should be copied and vice versa.

tasawernawaz (changes requested):

course_discovery/apps/publisher/views.py
waheedahmed: @tasawernawaz Every course will have at least one course-run but may not have a seat thats why I'm generating course-run initial data first, If there is not seat and exception occured course-run data will be copied.

waheedahmed:
@tasawernawaz Comments addressed please take another look.

tasawernawaz (approved):

PR #848, opened 2017-04-20
Update the api gateway to provide x-forwarded-host header value

adeelkhan:
ECOM-7686 - Sub task of ECOM-5122
 api gateway now forwards x-forwarded-host header.
WIP, Not for external reviewers.

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -50,6 +50,12 @@ auth_header: &auth_header
   required: true
   type: "string"
 
+host_header: &host_header
+  name: "Host"
+  in: "header"
+  required: true
+  type: "string"
+
 limit_qs_parameter: &limit_qs_parameter
   in: querystring
   name: limit
@@ -70,6 +76,7 @@ x-amazon-apigateway-integration: &apigateway_integration
   type: "http"
   requestParameters:
     integration.request.header.Authorization: "method.request.header.Authorization"
+    integration.request.header.X-Forwarded-Host: "method.request.header.Host"
 
 x-amazon-apigateway-integration-with-id: &apigateway_integration_with_id_parameter
   responses: *apigateway_responses
@@ -77,6 +84,7 @@ x-amazon-apigateway-integration-with-id: &apigateway_integration_with_id_paramet
   type: "http"
   requestParameters:
     integration.request.header.Authorization: "method.request.header.Authorization"
+    integration.request.header.X-Forwarded-Host: "method.request.header.Host"
     integration.request.path.id: "method.request.path.id"
 
 endpoints:
@@ -88,6 +96,7 @@ endpoints:
           produces: *produces
           parameters:
             - *auth_header
+            - *host_header
             - *limit_qs_parameter
             - *offset_qs_parameter
           operationId: "get_catalogs"
@@ -102,6 +111,7 @@ endpoints:
           produces: *produces
           parameters:
             - *auth_header
+            - *host_header
             - *id_parameter
           operationId: "get_catalog_by_id"
           responses: *responses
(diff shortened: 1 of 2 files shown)


schenedx (commented):
@adeelkhan I don't really have domain knowledge on how the changes here would affect the behavior of api gateway. I did notice that on api-manager file, the reference https://github.com/edx/api-manager/blob/master/swagger/api.yaml#L33 refer to a specific version. Do we need to update the api-manager swagger api.yaml as well?

efagin:
@schenedx yes, once this merges we'll need to update the api-manager reference accordingly. @mattdrayer's team has a development instance of the gateway that you could potentially use if you want, before trying to update and merge everything.

api-compact.yaml
```diff
@@ -50,6 +50,12 @@ auth_header: &auth_header
   required: true
   type: "string"
 
+host_header: &host_header
+  name: "HOST"
```
efagin: I would prefer that this is defined as {{Host}}, instead of {{HOST}}. Header names are typically not defined as all-caps. Please try to remain consistent with how the {{Authorization}} header is defined.

api-compact.yaml
```diff
@@ -70,6 +76,7 @@ x-amazon-apigateway-integration: &apigateway_integration
   type: "http"
   requestParameters:
     integration.request.header.Authorization: "method.request.header.Authorization"
+    integration.request.header.X-Forwarded-Host: "method.request.header.HOST"
```
efagin: Please see above regarding {{Host}} instead of {{HOST}}.

api-compact.yaml
```diff
@@ -78,6 +85,7 @@ x-amazon-apigateway-integration-with-id: &apigateway_integration_with_id_paramet
   requestParameters:
     integration.request.header.Authorization: "method.request.header.Authorization"
     integration.request.path.id: "method.request.path.id"
+    integration.request.header.X-Forwarded-Host: "method.request.header.HOST"
```
efagin: Please see above regarding {{Host}} instead of {{HOST}}.

mattdrayer:
@saleem-latif @asadiqbal08 @zubair-arbi FYI

api-compact.yaml
efagin: Also, please group the {{integration.request.header}} items together instead of just appending this to the end of the list, as it's easier to review and understand that way.

efagin (changes requested):

efagin:
@mattdrayer also FYI that you're likely going to want to use this Host header proxying for the ENT catalog endpoints as well (or any other endpoint that has paginated responses). We first need to ensure that this method works, but once it's established I recommend doing something similar.

efagin (approved):

schenedx (approved):
👍 Let's see if this does the job on stage.

clintonb:
> Let's see if this does the job on stage.

Has this not been tested locally?

clintonb:
By "this" I mean the usage of the X-Forwarded-For header to change pagination URL. That seems like it should be pretty quick to test locally with cURL.

schenedx:
@clintonb "The usage of the X-Forwarded-For header to change pagination URL" was actually already tested on PR https://github.com/edx/course-discovery/pull/838. The PR there is to realize that update.
This PR is really to configure the API manager behavior to provide the Hostname value in x-forwarded-for header

api-compact.yaml
```diff
@@ -69,14 +75,18 @@ x-amazon-apigateway-integration: &apigateway_integration
   httpMethod: "GET"
   type: "http"
   requestParameters:
-    integration.request.header.Authorization: "method.request.header.Authorization"
+    integration.request.header:
+      Authorization: "method.request.header.Authorization"
```
efagin: I don't think the Swagger spec allows nesting of request parameters.

efagin: On second thought, that's part of the AWS specific detail, so I'm not sure that's a Swagger issue, but I suspect it'll fail at the AWS upload level.

PR #856, opened 2017-04-25
LEARNER-692 Populate short description override from drupal course runs if present

MatthewPiatetsky:
fyi @edx/learner-growth

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -472,6 +472,7 @@ class CourseMarketingSiteDataLoader(AbstractMarketingSiteDataLoader):
             'mobile_available': data.get('field_course_enrollment_mobile') or False,
             'video': course.video,
             'course': course,
+            'short_description_override': self.clean_html(data['field_course_sub_title_short']) or None,
         }
 
         if weeks_to_complete:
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -524,6 +524,7 @@ def assert_course_run_loaded(self, data):
             'pacing_type': self.loader.get_pacing_type(data),
             'hidden': self.loader.get_hidden(data),
             'mobile_available': data['field_course_enrollment_mobile'] or False,
+            'short_description_override': self.loader.clean_html(data['field_course_sub_title_short']) or None,
```
rlucioni, line 527: Talk about a change-detector test, huh?

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -472,6 +472,7 @@ def format_course_run_data(self, data, course):
             'mobile_available': data.get('field_course_enrollment_mobile') or False,
             'video': course.video,
             'course': course,
+            'short_description_override': self.clean_html(data['field_course_sub_title_short']) or None,
```
rlucioni, line 475: Is this something that should only be taken from the canonical course run?

MatthewPiatetsky: I need to double check with patrick but during the launch they seemed very confused by the fact that they would make edits on the new run and it would not change on the program list view. I am fairly sure they preferred it to behave this way, but i will double check.

MatthewPiatetsky: we want images and short descriptions to come from the course run corresponding to the enroll button if they are available

MatthewPiatetsky:
bump

rlucioni (approved):

PR #857, opened 2017-04-25
Cache program viewset responses

rlucioni:
@clintonb @mjfrey @mikedikan @MatthewPiatetsky this is a proof-of-concept showing how we can cache DRF responses using only utilities provided by Django. [cache_page()](https://docs.djangoproject.com/en/1.11/topics/cache/#django.views.decorators.cache.cache_page) uses the request received by the view to construct a cache key, taking headers and querystring parameters into account when doing so.

### DIFF
--- a/course_discovery/apps/api/v1/views/programs.py
+++ b/course_discovery/apps/api/v1/views/programs.py
@@ -1,6 +1,7 @@
 from rest_framework import mixins, viewsets
 from rest_framework.filters import DjangoFilterBackend
 from rest_framework.permissions import IsAuthenticated
+from rest_framework_extensions.cache.mixins import CacheResponseMixin
 
 from course_discovery.apps.api import filters, serializers
 from course_discovery.apps.api.pagination import ProxiedPagination
@@ -9,7 +10,7 @@ from course_discovery.apps.course_metadata.models import ProgramType
 
 
 # pylint: disable=no-member
-class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
+class ProgramViewSet(CacheResponseMixin, viewsets.ReadOnlyModelViewSet):
     """ Program resource. """
     lookup_field = 'uuid'
     lookup_value_regex = '[0-9a-f-]+'
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -308,7 +308,6 @@ LOGGING = {
     }
 }
 
-
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': (
         'rest_framework.authentication.SessionAuthentication',
@@ -333,6 +332,12 @@ REST_FRAMEWORK = {
     'DEFAULT_THROTTLE_RATES': {
         'user': '100/hour',
     },
+}
+
+# http://chibisov.github.io/drf-extensions/docs/
+REST_FRAMEWORK_EXTENSIONS = {
+    'DEFAULT_CACHE_ERRORS': False,
+    'DEFAULT_CACHE_RESPONSE_TIMEOUT': 60,
 }
 
 # NOTE (CCB): JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.
--- a/.travis/docker-compose-travis.yml
+++ b/.travis/docker-compose-travis.yml
@@ -14,9 +14,9 @@ services:
   es:
     image: elasticsearch:1.5.2
     container_name: es
-  memcache:
+  memcached:
     image: memcached:1.4.24
-    container_name: memcache
+    container_name: memcached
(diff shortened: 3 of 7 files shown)


course_discovery/settings/base.py
```diff
@@ -309,6 +309,8 @@
 }
 
 
+RESPONSE_CACHE_TTL = 60
```
rlucioni: If we had a good way to clear and/or rebuild specific cache entries when certain models are saved, this could be set to `None` to cache entries indefinitely.

rlucioni: Doing this would require a way of finding the cache entries associated with each model. For example, when a program is saved, we'd need to know which cache keys to clear and/or set. In theory, we could do this by crafting request objects and passing them to the function Django uses to construct the view cache keys. Not pretty.

rlucioni:
@mikedikan @clintonb @MatthewPiatetsky what are your thoughts on starting small and introducing simple, TTL-based caching for the programs endpoint as shown here? Even a short TTL (e.g. 1 minute, 5 minutes) could help protect the discovery service (and by extension, other services like the LMS and ecommerce) when discovery experiences significant throughput spikes.

mikedikan:
I'd be fine with a 1 minute TTL - we just should make it clear to our django admin users (tatiana especially)

rlucioni:
@mikedikan sure. We could even add a simple cache-clearing view like the one @clintonb implemented for the credentials service, so that we have a way to clear the cache in case of emergencies. Though that doesn't seem as critical if we're only using 1 to 5 minute TTLs.

MatthewPiatetsky:
for 1 minute I don't think we need to be able to clear it

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -41,6 +44,7 @@ def get_serializer_context(self, *args, **kwargs):
 
         return context
 
+    @method_decorator(cache_page(settings.RESPONSE_CACHE_TTL))
```
clintonb: How does this work for authenticated, non-authenticated, and error responses? 

https://www.technomancy.org/python/django-cache_page-useless/

We should use drf-extensions to avoid these issues: https://chibisov.github.io/drf-extensions/docs/#caching

clintonb (changes requested):

course_discovery/apps/api/v1/views/programs.py
rlucioni: @clintonb what I've got here handles authentication fine, but it doesn't give us the option to skip caching of error responses, nor does it provide us with many of the tools drf-extensions provides (e.g., key functions). I will convert this to using drf-extensions to avoid re-inventing the wheel.

rlucioni:
@clintonb updated to use drf-extensions, which provides a variety of tools we can use to implement response caching. This change now adds caching with a short 1 minute TTL for the ProgramViewSet's list and retrieve actions. Error responses are not cached. This will provide some protection against significant increases in throughput. We can build on this later to add a longer TTL and invalidation on model changes.

rlucioni:
Also, one point of clarification: we don't yet have a Memcached cluster configured for discovery. The plan is to move ahead with this change and use in-memory, per-worker caches while I work with @feanil to write the Terraform necessary to provision ourselves an ElastiCache Memcached cluster. Once that's done, we should be able to apply the same patterns to other services which need a Memcached cluster.

rlucioni:
Terraform PR to provision ElastiCache clusters is taking shape at https://github.com/edx-ops/terraform/pull/415.

clintonb (approved):

rlucioni:
@clintonb thanks for the review. I've added some changes to make sure that we're using memcached when running tests, and modified tests to verify that we're actually using the cache as expected.

clintonb (approved):

rlucioni:
I'm not sure what to make of these [defaults](https://github.com/edx/configuration/blob/master/playbooks/roles/discovery/defaults/main.yml#L46-L52) in the configuration repo. Unless they're overridden in secure config so that we're using an in-memory cache, I don't think I can deploy these changes without breaking things (because we don't actually have a Memcached cluster to use yet). We've gotten by to this point because we're not using the cache yet.

@feanil I think the right thing to do here is going to be to move the cache settings from configuration to edx-internal, and then make sure they're specified correctly once we've set up the ElastiCache clusters. Can you confirm? I will hold off on merging and deploying this until I know things aren't going to break.

rlucioni:
Rather than make a config change to use per-worker, in-memory caches, then another to switch to memcached, I'm going to hold off on merging this until Terraform from https://github.com/edx-ops/terraform/pull/415 is applied and config updates in https://github.com/edx/edx-internal/pull/57 are merged and deployed.

PR #863, opened 2017-04-26
Invalidate API cache on course_metadata model changes

rlucioni:
Builds on https://github.com/edx/course-discovery/pull/857.

This is a proof-of-concept demonstrating how to invalidate cached actions on a DRF viewset when dependent models change.

LEARNER-401

@clintonb @mikedikan thoughts? This stands to be cleaned up a bit and/or generalized for viewsets other than the ProgramViewSet, but it's an example of what I was describing earlier today.

### DIFF
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -337,7 +337,9 @@ REST_FRAMEWORK = {
 # http://chibisov.github.io/drf-extensions/docs/
 REST_FRAMEWORK_EXTENSIONS = {
     'DEFAULT_CACHE_ERRORS': False,
-    'DEFAULT_CACHE_RESPONSE_TIMEOUT': 60,
+    'DEFAULT_CACHE_RESPONSE_TIMEOUT': None,
+    'DEFAULT_LIST_CACHE_KEY_FUNC': 'course_discovery.apps.api.cache.timestamped_list_key_constructor',
+    'DEFAULT_OBJECT_CACHE_KEY_FUNC': 'course_discovery.apps.api.cache.timestamped_object_key_constructor',
 }
 
 # NOTE (CCB): JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -1,0 +1,53 @@
+import logging
+import time
+
+from django.core.cache import cache
+from rest_framework_extensions.key_constructor.bits import KeyBitBase
+from rest_framework_extensions.key_constructor.constructors import (
+    DefaultListKeyConstructor, DefaultObjectKeyConstructor
+)
+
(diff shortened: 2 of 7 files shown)


rlucioni:
Theoretically, we could use something like [this](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/modelviz.py) to automatically determine relationships between models and register the correct signal handlers for each viewset.

mikedikan:
Looks like the signals handler clears the list api if any program, course, or course run is updated/deleted.  Why not org or any of the other nested metadata?

rlucioni:
@mikedikan only because this is a proof-of-concept/WIP. You're right that we would want to clear the cached response if any of those nested models change. If we go this route, I'd like those dependent models to be found and associated with the signal handler automatically.

mikedikan:
got it - thanks

rlucioni:
I've updated this so that the API cache will be invalidated whenever any model in the course metadata app is saved or deleted. Given how interconnected our data is and how infrequently our models change (data loading aside), this is a clean and simple way to ensure correctness of the API while providing closer-to-optimal cache TTLs. All that's required to apply this caching strategy to other viewsets is inheriting from the `CacheResponseMixin` and setting the key constructors on the viewset.

clintonb (approved):

course_discovery/settings/base.py
```diff
@@ -337,7 +337,9 @@
 # http://chibisov.github.io/drf-extensions/docs/
 REST_FRAMEWORK_EXTENSIONS = {
     'DEFAULT_CACHE_ERRORS': False,
-    'DEFAULT_CACHE_RESPONSE_TIMEOUT': 60,
+    'DEFAULT_CACHE_RESPONSE_TIMEOUT': None,
+    'DEFAULT_LIST_CACHE_KEY_FUNC': 'course_discovery.apps.api.cache.timestamped_list_key_constructor',
+    'DEFAULT_OBJECT_CACHE_KEY_FUNC': 'course_discovery.apps.api.cache.timestamped_object_key_constructor',
```
rlucioni, line 342: These caching functions are now specified in default settings, so you don't need to reapply them every time you want to apply caching to a new viewset. Just inherit from `CacheResponseMixin` and you're good to go. @clintonb FYI.

rlucioni:
This is cleaned up and ready to go! I've added logging to give us visibility into cache invalidation. I've also tweaked the `refresh_course_metadata` command so that the API response cache is only invalidated once, after data loading completes.

At the moment, the `ProgramViewSet` continues to be the only cached viewset. My plan is to deploy this on Monday morning and, assuming things looks good, expand use of the `CacheResponseMixin` to other viewsets as appropriate.

clintonb (approved):

clintonb:
Tests?

rlucioni:
Tests added.

PR #865, opened 2017-04-28
Prototype code for import course-metadata courses into publisher app.

awais786:
**WIP: This Code is just for team review will not be merge into master.**
This PR contains Prototype code for importing course-metadata into publisher app.

1. I made some changes in code instead of using any organizaion i think its good to do this in batches. e.g start_id = 1 to end_id= 500
2. Second the right mapping of fields.

ECOM-7712

Following ticket contains detailed information.
https://openedx.atlassian.net/browse/ECOM-7755


awais786:
@bderusha what do think about this PR. I want to discuss this with you next week also.

awais786:
**Fields Mapping questions**
Metadata has authoring_organizations and sponsoring_organizations. But publisher has only organizations. So which organization from metadata will go into publisher and it should be one only?

awais786:
Is there any field in metadata for course-image ? I did't find any one.

clintonb:
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L277

awais786:
In course table there is a URL field card_image_url. We don't have any field in publisher to save this field. We have image field which saves the image as file. So i think we need to add another URL field to save this value.

PR #867, opened 2017-05-02
[DO NOT MERGE][BLOCKED][WIP][LEARNER-687] Import course instructors' data from the LMS

marjev:
### Reviewers:
@edx/helio 

### JIRA ticket:
https://openedx.atlassian.net/browse/LEARNER-687

### Related PRs:
https://github.com/edx/edx-platform/pull/14986

# To-do:
- [ ] Write tests
- [ ] Figure out a way to provide instructor's UUID

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -448,7 +448,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
     transcript_languages = serializers.SlugRelatedField(many=True, read_only=True, slug_field='code')
     video = VideoSerializer()
     seats = SeatSerializer(many=True)
-    instructors = serializers.SerializerMethodField(help_text='This field is deprecated. Use staff.')
+    instructors = PersonSerializer(many=True)
     staff = PersonSerializer(many=True)
     level_type = serializers.SlugRelatedField(read_only=True, slug_field='name')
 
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -448,7 +448,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
     transcript_languages = serializers.SlugRelatedField(many=True, read_only=True, slug_field='code')
     video = VideoSerializer()
     seats = SeatSerializer(many=True)
-    instructors = serializers.SerializerMethodField(help_text='This field is deprecated. Use staff.')
```
clintonb: This field was deprecated for a reason. Why are you restoring it?

clintonb (changes requested):

course_discovery/apps/api/serializers.py
marjev: To differentiate between staff members and instructors.

Each time a `refresh_course_metadata` command is ran, staff field is cleared and updated. To refresh instructors' data in the same manner, I planned to restore this field.

Is there a better way to handle this?

clintonb: Do we need that separation? I would prefer to mimic the edx.org behavior as much as possible. We do not make such a separation there.

marjev: edx.org only displays the staff members on the Program marketing page, but the WL site uses both the staff member and the instructors.

If we decide to keep the data in a single Many2Many field (staff), we would have to handle the refreshing of the course metadata, since the field is currently cleared before being filled again. This is OK now, because there's a single source of data (Marketing API). But, how would we handle the case when both Marketing API and LMS Courses API provide the data during the execution of `refresh_course_metadata` for the same course?

marjev:
@clintonb I've addressed the comment you left. Could you please take another look to verify that this is the way to go?

PR #872, opened 2017-05-03
Add option to serialize only program UUIDs

rlucioni:
ProgramViewSet.list now returns a flat list of program UUIDs when `uuids_only=1` is included in the querystring. This allows clients which cache each program in an individual entry (e.g., LMS) to more quickly identify the programs they need to get from their cache.

LEARNER-382

@edx/learner-growth FYI.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -137,6 +137,28 @@ class ProgramViewSetTests(SerializationMixin, APITestCase):
         # Verify that repeated list requests use the cache.
         self.assert_list_results(self.list_path, expected, 2)
 
+    def test_uuids_only(self):
+        """
+        Verify that the list view returns a simply list of UUIDs when the
+        uuids_only query parameter is passed.
+        """
+        active = ProgramFactory.create_batch(3)
+        retired = [ProgramFactory(status=ProgramStatus.Retired)]
+        programs = active + retired
+
+        querystring = {'uuids_only': 1}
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -137,6 +137,28 @@ def test_list(self):
         # Verify that repeated list requests use the cache.
         self.assert_list_results(self.list_path, expected, 2)
 
+    def test_uuids_only(self):
+        """
+        Verify that the list view returns a simply list of UUIDs when the
+        uuids_only query parameter is passed.
+        """
+        active = ProgramFactory.create_batch(3)
+        retired = [ProgramFactory(status=ProgramStatus.Retired)]
```
clintonb, line 146: Should we iterate over all statuses to cover our bases?

clintonb (approved):

course_discovery/apps/api/v1/tests/test_views/test_programs.py
rlucioni: The most complete approach here would be to extend each existing test of the list view's filtering to verify that it's possible to get just UUIDs back. Do you feel strongly about this? A lightweight check that filtering is supported seemed sufficient to me.

clintonb: I don't really care. It's just strange to see the retired status get special treatment, but not the other statuses.

rlucioni: Got it. It could be any status, I just happened to pick this one. I'm going to leave this as-is for now to avoid sinking time into rewriting the existing tests.

PR #874, opened 2017-05-04
LEARNER-834 Include hidden course runs

MatthewPiatetsky:


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -59,6 +59,29 @@ def test_get_include_deleted_programs(self):
                 self.serialize_course(self.course, extra_context={'include_deleted_programs': True})
             )
 
+    def test_get_include_hidden_course_runs(self):
+        """
+        Verify the endpoint returns associated hidden course runs
+        with the 'include_hidden_course_runs' flag set to True
```
mikedikan, line 65: nitpick: when the ... flag is set to True

mikedikan (approved):

course_discovery/apps/api/v1/tests/test_views/test_courses.py
MatthewPiatetsky: i copied this from above, but i think this also makes sense

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -33,7 +33,10 @@ def get_queryset(self):
             queryset = Course.search(q)
             queryset = self.get_serializer_class().prefetch_queryset(queryset=queryset)
         else:
-            course_runs = CourseRun.objects.exclude(hidden=True)
+            if not get_query_param(self.request, 'include_hidden_course_runs'):
```
rlucioni: Minor, but I would flip this for the sake of readability:
```py
if get_query_param(self.request, 'include_hidden_course_runs'):
    course_runs = CourseRun.objects.all()
else:
    course_runs = CourseRun.objects.exclude(hidden=True)
```
There's no need to involve a negation here.

rlucioni (approved):

PR #875, opened 2017-05-04
Removed unused partner parameter from Swagger view for programs list endpoint

clintonb:
We haven't used this parameter for quite some time. It doesn't work, and should not be shown on the Swagger page.


clintonb:
FYI @mjfrey

mjfrey (approved):

PR #876, opened 2017-05-05
Updated program_types endpoint

clintonb:
- Organized view in the same manner as all other views
- Using a ReadOnlyViewSet
- Returning applicable_seat_types in serialized response

LEARNER-435


clintonb:
@edx/helio please review

FYI @edx/learner

schenedx (approved):
👍

PR #877, opened 2017-05-05
Create only course object from new course page, removed run and seat related code.

waheedahmed:
[ECOM-7760](https://openedx.atlassian.net/browse/ECOM-7760)

As a CT Admin, I'd like to create a new Course from the dashboard or course-list page.

**AC:**

1. Validate if I clicked on Create new course button from dashboard, I'm taken to a new course page with only new course fields (same as when editing a course).
2. Validate I can save just the course fields
3. Validate there is a checkbox above the "Create New Course" button whereby checking on this checkbox will create the course and then automatically (provided there are no errors) take the user to the Add a run page where they can enter the run related fields.
4. Validate if the checkbox is not clicked, the course is created and the user is taken to the parent course page with a success message (same way it works now)

screencapture-localhost-8004-publisher-courses-new-1493971927694


waheedahmed:

tasawernawaz (approved):
LGTM

PR #880, opened 2017-05-08
Minimum effort and maximum effort field validations

asadazam93:
Learner-852

Field validation added. Minimum effort cannot be greater than Maximum effort.


waheedahmed (approved):

awais786:
Where is test case ?

asadazam93:
@awais786 adding test case

PR #881, opened 2017-05-08
Allow course runs to be saved when publication fails

rlucioni:
Preventing course runs from being saved unless they are are successfully published to the marketing site makes the Django admin difficult to use for those who don't need data published to Drupal. It's not reasonable to couple saving and publication without also supporting node creation. We shouldn't expect everyone using the admin to create a node on the marketing site before saving course run status changes.

[LEARNER-904](https://openedx.atlassian.net/browse/LEARNER-904)

@brittneyexline @mikedikan @schenedx

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -608,9 +608,12 @@ class CourseRun(TimeStampedModel):
             publisher = CourseRunMarketingSitePublisher(self.course.partner)
             previous_obj = CourseRun.objects.get(id=self.id) if self.id else None
 
-            with transaction.atomic():
-                super(CourseRun, self).save(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


schenedx (approved):
👍 
Do we have unit tests we need to update?

rlucioni:
@schenedx yes. I'll update them if we decide this is what we want to do.

The downside of making this change is that if there's a problem publishing like there was today, the save will succeed anyway, leaving course run statuses out of sync. People using the admin won't know there was a problem until they go to the marketing site and see that data hasn't been updated. I don't know if we want to open ourselves to that possibility.

rlucioni:
I'm going to hold off on making this change for now. We don't want publication to fail silently in production. @brittneyexline it sounds like you're planning to follow up on this with @mikedikan. I think we'll be able to make a safer change when we better understand what you're trying to do.

PR #883, opened 2017-05-09
Create course run from dashboard.

waheedahmed:
[ECOM-7758](https://openedx.atlassian.net/browse/ECOM-7758)

As a CT admin, I'd like to create a course run from the Dashboard and specify the parent course.

**AC:**

1. Validate, if the new course run flow was started from the dashboard, there's a way for the user to specify the Parent course for the course run by entering either parent course id or title (type ahead).
2. Validate After the course id or title is provided, then I can create the course run the same way I create it from the Course detail page.
3. Validate Studio instance creation is also part of this flow.
4. Validate the emails get sent out the same way as before

screencapture-localhost-8004-publisher-1494333460893

----------------------------------------------------------------------------------------------------

screencapture-localhost-8004-publisher-course_runs-new-1494333482689


waheedahmed:
@awais786 @tasawernawaz @asadazam93 Please review.

awais786 (approved):

PR #884, opened 2017-05-10
Course Editing with revision changes 

awais786:
If user selected any previous revision then load the edit form with changes comparison in tinymce.
ECOM-7790

screen shot 2017-05-11 at 12 55 25 pm

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -339,10 +339,19 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
 
     def get_context_data(self, **kwargs):
         context = super(CourseEditView, self).get_context_data(**kwargs)
-        context.update(
-            {'is_internal_user': is_internal_user(self.request.user)}
-        )
+        history_id = self.request.GET.get('history_id', None)
 
+        try:
+            history_object = self.object.history.get(history_id=history_id) if history_id else None
+        except:  # pylint: disable=bare-except
+            history_object = None
+
+        context.update(
+            {
+                'is_internal_user': is_internal_user(self.request.user),
+                'history_object': history_object
+            }
+        )
         return context
 
     def form_valid(self, form):
--- a/course_discovery/templates/publisher/_history_widget.html
+++ b/course_discovery/templates/publisher/_history_widget.html
@@ -11,7 +11,14 @@
                     {% endblocktrans %}
                 </option>
             {% else %}
-                <option data-revision-url="{% url 'publisher:api:course_revisions' history.history_id  %}" value="{% url 'publisher:publisher_course_revision' course.id history.history_id %}" data-revert-url="{% url 'publisher:api:course_revision_revert' history.history_id  %}">
+                {% url 'publisher:api:course_revisions' history.history_id as revision_url  %}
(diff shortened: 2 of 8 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -338,10 +338,15 @@ def get_form_kwargs(self):
 
     def get_context_data(self, **kwargs):
         context = super(CourseEditView, self).get_context_data(**kwargs)
+        history_id = self.request.GET.get('history_id', None)
+        history_object = self.object.history.filter(history_id=history_id) if history_id else None
```
waheedahmed: Why don't you used get here instead of filter, it will look more cleaner and better.

course_discovery/settings/test.py
```diff
@@ -43,3 +43,5 @@
 LOGGING['handlers']['local'] = {'class': 'logging.NullHandler'}
 
 PUBLISHER_FROM_EMAIL = 'test@example.com'
+COMPRESS_ENABLED = False
```
waheedahmed: Revert these changes.

course_discovery/templates/publisher/_history_widget.html
```diff
@@ -11,7 +11,7 @@ <h5 class="hd-5 emphasized course-runs-heading">{% trans "REVISION HISTORY" %}</
                     {% endblocktrans %}
                 </option>
             {% else %}
-                <option data-revision-url="{% url 'publisher:api:course_revisions' history.history_id  %}" value="{% url 'publisher:publisher_course_revision' course.id history.history_id %}" data-revert-url="{% url 'publisher:api:course_revision_revert' history.history_id  %}">
+                <option data-revision-url="{% url 'publisher:api:course_revisions' history.history_id  %}" value="{% url 'publisher:publisher_course_revision' course.id history.history_id %}" data-revert-url="{% url 'publisher:api:course_revision_revert' history.history_id  %}" data-revision-id="{% url 'publisher:publisher_courses_edit' pk=object.id %}?history_id={{ history.history_id }}">
```
waheedahmed: Line is going very long and not easy to read, make all these url's a variable like this `{% url 'publisher:api:course_revisions' history.history_id as course_revisions_url %}` and then put them in option attrs.

waheedahmed (changes requested):
First pass completed.

course_discovery/templates/publisher/_history_widget.html
```diff
@@ -11,7 +11,14 @@ <h5 class="hd-5 emphasized course-runs-heading">{% trans "REVISION HISTORY" %}</
                     {% endblocktrans %}
                 </option>
             {% else %}
-                <option data-revision-url="{% url 'publisher:api:course_revisions' history.history_id  %}" value="{% url 'publisher:publisher_course_revision' course.id history.history_id %}" data-revert-url="{% url 'publisher:api:course_revision_revert' history.history_id  %}">
+                {% url 'publisher:api:course_revisions' history.history_id as revision_url  %}
+                {% url 'publisher:publisher_course_revision' course.id history.history_id as value %}
```
waheedahmed: Use more appropriate variable name instead `value`.

waheedahmed (approved):
Just one comment after that 👍

PR #887, opened 2017-05-11
Removed CourseTeam from organization role dropdown choices.

waheedahmed:
LEARNER-810

'Course Team' role option should not appear in organization role dropdown. The provisions to assign 'course team' role to a user should only be provided through Publisher app.


waheedahmed:
@awais786 @tasawernawaz Please review.

awais786 (approved):

PR #894, opened 2017-05-15
Filtered instructors by org on course-run edit

asadazam93:
ECOM-7774

When searching and adding instructors to a Course-run, the type ahead search only returns instructors from the same Organization as the Course team logged in.


waheedahmed:
Add some tests as well, with and without internal user.

waheedahmed (approved):

PR #899, opened 2017-05-16
Average length calculate based on canonical course run

ahsan-ul-haq:
LEARNER-578

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -818,7 +818,7 @@ class Program(TimeStampedModel):
 
     @cached_property
     def _course_run_weeks_to_complete(self):
-        return [course_run.weeks_to_complete for course_run in self.course_runs
+        return [course_run.weeks_to_complete for course_run in self.canonical_course_runs
                 if course_run.weeks_to_complete is not None]
 
     @property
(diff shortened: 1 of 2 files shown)


tasawernawaz (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -851,6 +851,15 @@ def course_runs(self):
                     yield run
 
     @property
+    def canonical_course_runs(self):
+        excluded_course_run_ids = [course_run.id for course_run in self.excluded_course_runs.all()]
+
+        for course in self.courses.all():
```
schenedx, line 857: I am glad you are thinking about the excluded_course_run case.
What if canonical_course_run is excluded, instead of not using it, shouldn't we choose another course_run from the course? It makes it slightly complicated. But I feel we should protected the system with that logic.

schenedx (changes requested):
Please consider my suggestion. If my suggestion is not clear, message me.

course_discovery/apps/course_metadata/models.py
ahsan-ul-haq: Canonical course run is the current run of the course, I don't think it will make sense to do the calculation based on previous or future course runs. So if canonical course run is excluded the average length still will be calculated based on the other course runs in the program.

schenedx: Got it. Thanks for the response

schenedx (approved):
👍

PR #900, opened 2017-05-17
Feature to accept/reject changes individual changes in tinyMCE.

awais786:
https://openedx.atlassian.net/browse/ECOM-7779

screen shot 2017-05-17 at 2 38 49 pm


course_discovery/static/sass/publisher/publisher.scss
```diff
@@ -423,6 +423,20 @@
...
+    position: relative;
+
+    #btn_course_edit {
+        @include right(4px);
+        position: absolute;
+        margin: 0 auto;
+        top: 42px;
+        font-weight: 600;
+        font-size: 0.875rem;
+        margin-right: 50px;
```
waheedahmed: Add RTL support.

course_discovery/static/js/publisher/tinymce-init.js
```diff
@@ -3,22 +3,89 @@ $(document).ready(function(){
...
+                editor.on('NodeChange', function(e) {
+                    console.log(e.element.getAttribute('class'));
+
+                });
+            },
+            content_style: '.added {background-color: greenyellow;} .removed {background-color: red;}'
+
     };
 
+    function getParentInsDelElement(elem)
```
waheedahmed: It looks like this function returns `INS` or `DEL` elements, so name should be `getInsDelElement`

waheedahmed (approved):
Couple of comments, otherwise 👍

PR #901, opened 2017-05-18
Remove CT from reviewed email when PC approves course-run.

waheedahmed:
ECOM-7767

**AC:**

- Validate when PC approves the course-run and it's sent to the Publisher role to create preview, no email is sent to the CT


waheedahmed:
@tasawernawaz @awais786 Please review

awais786 (approved):

PR #903, opened 2017-05-18
upgrade django to 1.10.7

ahsan-ul-haq:
# Summary
Upgrade Django to the newer version

## Description
As Django has been upgraded to 1.11.X, this is one step moving towards the newer version.

LEARNER-946

### DIFF
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -73,7 +73,6 @@ class CourseFactory(factory.DjangoModelFactory):
     level_type = factory.SubFactory(LevelTypeFactory)
     card_image_url = FuzzyURL()
     video = factory.SubFactory(VideoFactory)
-    marketing_url = FuzzyText(prefix='https://example.com/test-course-url')
     partner = factory.SubFactory(PartnerFactory)
 
     class Meta:
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -178,7 +178,7 @@ class CreateCourseViewTests(TestCase):
 
         self.assertContains(
             response,
-            '<select class="field-input input-select" id="id_organization" name="organization">'
+            '<select class="field-input input-select" id="id_organization" name="organization" required>'
         )
 
         new_organization_extension = factories.OrganizationExtensionFactory()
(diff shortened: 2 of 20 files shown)


course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -73,7 +73,6 @@ class CourseFactory(factory.DjangoModelFactory):
     level_type = factory.SubFactory(LevelTypeFactory)
     card_image_url = FuzzyURL()
     video = factory.SubFactory(VideoFactory)
-    marketing_url = FuzzyText(prefix='https://example.com/test-course-url')
```
clintonb, line 76: Why was this removed?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -178,7 +178,7 @@ def test_create_form_with_multiple_organization(self):
 
         self.assertContains(
             response,
-            '<select class="field-input input-select" id="id_organization" name="organization">'
+            '<select class="field-input input-select" id="id_organization" name="organization" required>'
```
clintonb, line 181: Why is this changed?

course_discovery/apps/course_metadata/tests/factories.py
ahsan-ul-haq: marketing_url is a property in Course Model and it is not required to be here. But the reason I removed it was that when we create factory object in Django 1.10.7 it gives an error that the marketing URL is not a field.

course_discovery/apps/publisher/tests/test_views.py
ahsan-ul-haq: I am not quite sure here as the tests were failing and I checked and find out that the select input was required so changing it fixed the test.

clintonb (approved):
Squash your commits, and follow commit message pattern outlined at https://openedx.atlassian.net/wiki/display/ECOM/Developer+Onboarding.

schenedx (approved):
👍 
Can you tell us what is your plan to validate this change on stage? Can we do some integration tests with the stage marketing site, and, run at least the data load on stage?

ahsan-ul-haq:
@schenedx my plan for testing this on stage is to run following test cases:

- Create new program on stage and make sure it is published on marketing site
- Check that all the APIs are working fine 
- Create course run and make sure it is published on marketing
- As I don't have access to run data loader, I would appreciate you would do that.
- Run the whole process of publisher app at least once with help of @waheedahmed or @awais786 to make sure nothing is broken on their side.
- On the marketing side, make sure the data fetched from course-discovery is being rendered properly.

Let me know if there is any other test case you want me to check.
@clintonb your input on this would also be encouraged.

rlucioni:
@ahsan-ul-haq this PR is missing a migration for the service's User model. Directly from the [1.10 release notes](https://docs.djangoproject.com/en/1.11/releases/1.10/#abstractuser-username-max-length-increased-to-150):

> If you have a custom user model inheriting from AbstractUser, you’ll need to generate and apply a database migration for your user model.

Please open a PR adding the missing migration. Also, please let me know once you've validated this change on stage. I intend to deploy discovery to production on Monday morning.

PR #906, opened 2017-05-21
Added Custom migration for user model

ahsan-ul-haq:
A custom migration needs to be created in Django 1.10 for the user model inherided from AbstractUser


ahsan-ul-haq:
My apology for missing this part while updating Django to 1.10.7. @rlucioni thanks for pointing this out.

clintonb (approved):
No need to apologize. We need to do a better job of detecting missing migrations. This can be done by running `./manage.py makemigrations --check --dry-run`.

clintonb:
I have filed [LEARNER-1085](https://openedx.atlassian.net/browse/LEARNER-1085) to update Travis to detect missing migrations.

rlucioni (approved):
Please remember to squash your commits!

PR #909, opened 2017-05-22
Respect all querystring parameters when caching list responses

rlucioni:
The DefaultListKeyConstructor includes the PaginationKeyBit. While it does subclass QueryParamsKeyBit, it also bypasses logic which includes all query params in the cache key, restricting the set of query params that end up in the cache key to those that are used for page number pagination. This causes cache collisions when other query params are involved.

LEARNER-1072

@edx/learner FYI.


clintonb (approved):

schenedx (approved):
👍

PR #910, opened 2017-05-23
Fixed same role can be added multiple times against an Org or Course.

waheedahmed:
ECOM-7823

**AC:**

1. Validate that user cannot add same role multiple time for an organization in OrganizationUserRole
2. Validate that user cannot add same role multiple time for a course in CourseUserRole


waheedahmed:
@awais786 @tasawernawaz Please review.

awais786 (approved):

PR #911, opened 2017-05-23
Resolution of deprecation warnings and preparations for Django upgrade

ahsan-ul-haq:
LEARNER-947


course_discovery/apps/api/tests/test_views.py
```diff
@@ -1,7 +1,7 @@
 import ddt
 from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import PermissionDenied
-from django.core.urlresolvers import reverse
+from django.core.urlresolvers import reverse_lazy
```
clintonb: Why are you changing this everywhere? There is no mention of this change in [the release notes](https://docs.djangoproject.com/en/1.11/releases/1.11/).

requirements/base.txt
```diff
@@ -45,5 +45,3 @@ python-dateutil==2.5.3
 pytz==2016.6.1
 # Pinning `requests` at 2.9.1 because the latest version of the package causes test failures with mocked requests
 requests==2.9.1
-
-git+https://github.com/edx/django-sortedm2m.git@1.3.4#egg=django-sortedm2m==1.3.4
```
clintonb: Remember to delete this fork.

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -108,7 +108,7 @@ def test_retrieve(self):
         self.assertEqual(response.status_code, 200)
         self.assert_response_data_valid(response, organization, many=False)
 
-    @ddt.data(123, uuid.uuid4())
+    @ddt.data('31bb7f66-4c43-11e7-b114-b2f933d5fe66', uuid.uuid4())
```
clintonb: Why was this changed? If `123` is not acceptable, get rid of `ddt`.

requirements/base.txt
```diff
@@ -1,23 +1,23 @@
 beautifulsoup4==4.5.1
 boto==2.42.0
 cryptography==1.7.1
-django==1.10.7
+django==1.11.2
```
clintonb: Please follow the checklist at https://openedx.atlassian.net/wiki/display/OpenDev/Django+Upgrade+Checklist. Move this to a separate PR that we will deploy on its own.

ahsan-ul-haq:
@schenedx @clintonb @rlucioni could you please review.

clintonb (changes requested):
Correct your commit message and PR title. This is not the Django upgrade. This is the resolution of deprecation warnings, and other preparations, for the upgrade.

ahsan-ul-haq:
@clintonb corrected the title and commit message

clintonb (approved):

schenedx (approved):
👍

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -108,10 +108,9 @@ def test_retrieve(self):
         self.assertEqual(response.status_code, 200)
         self.assert_response_data_valid(response, organization, many=False)
 
-    @ddt.data(123, uuid.uuid4())
-    def test_retrieve_not_found(self, organization_uuid):
+    def test_retrieve_not_found(self):
```
rlucioni, line 109: @ahsan-ul-haq what prompted this change? If there's a good reason to make the change, `ddt` is no longer used in this module, so please remove it.

ahsan-ul-haq: The reason behind this is change is that in the newer version of Django `123` is no longer considered as valid `UUID` and the test fails in Django 1.11 with the error message that 123 is not valid UUID. I didn't notice that this is the only place where `ddt` was used so I will remove that.

ahsan-ul-haq:
Verified on stage

PR #912, opened 2017-05-23
display critical fields in admin

asadazam93:
ECOM 7824

All critical fields of each model are visible in the publisher admin site.
Also one is able to search using certain fields.


awais786 (approved):

asadazam93:
I have verified these changes on stage, they are ready for deployment.

PR #915, opened 2017-05-24
Fixed bug in Program.enrollable_seats

clintonb:
This method should not care about course run start dates. That date is irrelevant to determining if a course run is enrollable.

LEARNER-1151


clintonb:
I opted for a simpler fix rather than completely overhauling the query.

mjfrey (approved):

PR #920, opened 2017-05-25
Support program language faceting

rlucioni:
Indexes program languages and includes the new field in search responses.

LEARNER-340

@edx/learner-growth FYI


MatthewPiatetsky:
is this to be able to facet on program languages as well? 
have you tested this locally?
if so, 👍 once tests pass

rlucioni:
@MatthewPiatetsky yes to both. I'll merge once tests pass.

PR #929, opened 2017-05-29
Learner-780 -Display price of only active courses on prog cert page

Ayub-Khan:
Learner-780

**Description:**
prof cert program pages are displaying costs in the right sidebar under "price" for inactive, published runs (from years past), giving current or prospective learners confusing information.

**Expected Behavior:**
pricing in the right sidebar should display cost information associated only with current / published runs

**Acceptance Criteria:**
The desired behavior on the Prof Cert landing page is to display a price range for all course runs in the program, excluding course runs that are either unpublished or published, but whose start and end dates are in the past.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -452,6 +452,8 @@ class ProgramTests(TestCase):
         currency = Currency.objects.get(code='USD')
 
         course_run = factories.CourseRunFactory()
+        course_run.course.canonical_course_run = course_run
+        course_run.course.save()
         factories.SeatFactory(type='audit', currency=currency, course_run=course_run, price=0)
         factories.SeatFactory(type='credit', currency=currency, course_run=course_run, price=600)
         factories.SeatFactory(type='verified', currency=currency, course_run=course_run, price=100)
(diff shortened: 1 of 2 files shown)


schenedx (approved):
👍

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -651,6 +653,24 @@ def test_canonical_course_runs(self):
         # Verify only canonical course runs are returned in set
         self.assertEqual(set(self.program.canonical_course_runs), set(expected_canonical_runs))
 
+    def test_canonical_course_seats(self):
+        """ Test canonical course seats returns only canonical course run's applicable seats """
+        currency = Currency.objects.get(code='USD')
+
+        course = factories.CourseFactory()
+        course_runs_same_course = factories.CourseRunFactory.create_batch(3, course=course)
```
ahsan-ul-haq, line 661: Can't we use course runs created in the setup?

ahsan-ul-haq (changes requested):

course_discovery/apps/course_metadata/tests/test_models.py
Ayub-Khan: yeah we could have. But this is more suitable to test the right behavior. Because course runs created in setup are all from different courses. But We get double values for single course (one from expired and one from canonical ) if there are two. So i created the test like that it tests that it tests the canonical course seats for single course course runs.

ahsan-ul-haq (approved):
👍

schenedx:
@Ayub-Khan Please test your changes here on course discovery stage now. And report your findings on this PR. Thank you

PR #931, opened 2017-05-29
Remove un-necessary code form the command.

awais786:
I have remove the org and group creation code. 
Also remove the permissions assignment code.

https://openedx.atlassian.net/browse/ECOM-7844


awais786:
@waheedahmed Please review this PR.

waheedahmed:
Why removing that code? Are you going to right another command for that?

waheedahmed (approved):

awais786:
Yes i have divided the task into multiple task. It will avoid the significant load.

PR #934, opened 2017-05-30
Add hidden boolean field to Program

ibrahimahmed443:
Add hidden boolean field to Program model, search index and as a filter.


course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -226,6 +226,17 @@ def test_filter_by_status(self):
...
+        """ Endpoint should filter programs by their hidden attribute value. """
+        hidden = ProgramFactory(hidden=True)
+        not_hidden = ProgramFactory(hidden=False)
+
+        url = self.list_path + '?hidden=True'
+        self.assert_list_results(url, [hidden], 8)
+
+        url = self.list_path + '?hidden=False'
+        self.assert_list_results(url, [not_hidden], 8)
+
```
vkaracic, line 239: Could you please add a test to verify that the filter field is actually non-nullable. For example if you would filter with `?hidden=None` it shouldn't return any results.

clintonb: Frankly, I don't care about any values besides 0, 1, true, and false. As long as the API endpoint returns properly for those, we are fine.

clintonb: Please add tests with the field query string set to 0 and 1.

clintonb (changes requested):
Update to your commit message: `Added hidden filter to Programs API list endpoint`.

course_discovery/apps/api/filters.py
```diff
@@ -156,17 +156,15 @@ class Meta:
...
-
-
 class ProgramFilter(FilterSetMixin, django_filters.FilterSet):
     marketable = django_filters.MethodFilter()
     status = django_filters.MultipleChoiceFilter(choices=ProgramStatus.choices)
     type = django_filters.CharFilter(name='type__name', lookup_expr='iexact')
     types = CharListFilter(name='type__slug', lookup_expr='in')
     uuids = UUIDListFilter()
-    hidden = NonNullableBooleanFilter(name='hidden')
+    hidden = django_filters.TypedChoiceFilter(
```
clintonb: Yeah...this is too much of a hack. The first thing you should do is upgrade django-filter to the latest version to determine if the new package corrects issues with `BooleanFilter`. Also, see https://github.com/encode/django-rest-framework/issues/2122 for ideas.

clintonb: See f624a914050aee8748b74220fc27fb51e83f4e8c.

ibrahimahmed443:
@clintonb I fixed some issues with FilterSetMixin. Are we looking good now?

clintonb:
Squash the last four commits.

course_discovery/apps/course_metadata/models.py
```diff
@@ -780,7 +780,9 @@ class Program(TimeStampedModel):
         default=False,
         help_text=_('Allow courses in this program to be purchased in a single transaction')
     )
-    hidden = models.BooleanField(default=False, db_index=True)
+    hidden = models.BooleanField(
+        default=False, db_index=True,
+        help_text=_('Hide program on marketing site landing and search pages. This program MAY have a detail page.'))
```
clintonb: Ideally, this should be part of the first commit.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -226,6 +226,17 @@ def test_filter_by_status(self):
         url = self.list_path + '?status=active&status=retired'
         self.assert_list_results(url, [retired, active], 8)
 
+    def test_filter_by_hidden(self):
```
clintonb, line 229: This should be part of the last commit, which deals with API filtering.

course_discovery/apps/api/filters.py
```diff
@@ -161,6 +161,7 @@ class ProgramFilter(FilterSetMixin, django_filters.FilterSet):
     type = django_filters.CharFilter(name='type__name', lookup_expr='iexact')
     types = CharListFilter(name='type__slug', lookup_expr='in')
     uuids = UUIDListFilter()
+    hidden = django_filters.BooleanFilter(name='hidden')
```
clintonb: This belongs in the last commit.

clintonb (approved):

course_discovery/apps/course_metadata/migrations/0056_auto_20170531_0500.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0055_program_hidden'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
clintonb: Merge this with the migration above.

clintonb: All you need to do is move the `help_text` kwarg into the first commit.

clintonb:
@ibrahimahmed443 please remember to delete merged branches.

PR #936, opened 2017-05-31
Associate Site model with Partner

mjfrey:
Associate Site model with Partner model to allow for multi-tenant support.

LEARNER-1117


course_discovery/apps/core/models.py
```diff
@@ -78,6 +79,7 @@ class Partner(TimeStampedModel):
     oidc_key = models.CharField(max_length=255, null=True, verbose_name=_('OpenID Connect Key'))
     oidc_secret = models.CharField(max_length=255, null=True, verbose_name=_('OpenID Connect Secret'))
     studio_url = models.URLField(max_length=255, null=True, blank=True, verbose_name=_('Studio URL'))
+    site = models.OneToOneField(Site, null=True, blank=True)
```
vkaracic: A partner can be without a site attached to it?

course_discovery/apps/publisher_comments/tests/factories.py
```diff
@@ -1,16 +1,11 @@
 import factory
 from django.contrib.sites.models import Site
 
-from course_discovery.apps.core.tests.factories import UserFactory
+from course_discovery.apps.core.tests.factories import SiteFactory, UserFactory
 from course_discovery.apps.publisher.tests.factories import CourseRunFactory
 from course_discovery.apps.publisher_comments.models import Comments
 
 
-class SiteFactory(factory.DjangoModelFactory):  # pylint: disable=missing-docstring
```
vkaracic, line 9: Why the move to the upper file?

course_discovery/apps/core/models.py
mjfrey: Yes, for now, until we fix up the data.

course_discovery/apps/publisher_comments/tests/factories.py
mjfrey: Should be in core and is not specific to publisher.

course_discovery/apps/core/models.py
vkaracic: Is there a ticket for that fix? I'd like to see what data needs to be fixed, cause I might be able to do it in the same migration file.

mjfrey: We can just do this manually as there will only be a handful.  Can be done via the django admin.

course_discovery/apps/core/admin.py
```diff
@@ -59,4 +59,4 @@ class PartnerAdmin(admin.ModelAdmin):
     )
     list_display = ('name', 'short_code',)
     ordering = ('name', 'short_code',)
-    search_fields = ('name', 'short_code',)
+    search_fields = ('name', 'short_code', 'site')
```
clintonb: This field should be in the list display. Given the handful of sites, there is no need for search.

course_discovery/apps/core/migrations/0004_partner_site.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ('sites', '0002_alter_domain_unique'),
+        ('core', '0003_auto_20170522_0856'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='partner',
+            name='site',
+            field=models.OneToOneField(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='sites.Site'),
```
clintonb: Cascade is not good. We don't want to delete all of the data. We should disable this.

course_discovery/apps/core/tests/factories.py
```diff
@@ -1,12 +1,21 @@
...
 
+
 USER_PASSWORD = 'password'
 
 
+class SiteFactory(factory.DjangoModelFactory):
+    class Meta(object):
+        model = Site
+
+    domain = factory.fuzzy.FuzzyText(suffix='.fake')
```
clintonb: Fuzzy is deprecated. Use Faker.

course_discovery/apps/core/tests/factories.py
```diff
@@ -1,12 +1,21 @@
...
 
 from course_discovery.apps.core.models import Partner, User
 from course_discovery.apps.core.tests.utils import FuzzyUrlRoot
 
+
 USER_PASSWORD = 'password'
 
 
+class SiteFactory(factory.DjangoModelFactory):
+    class Meta(object):
```
clintonb: No need for the parenthetical in Python 3.5.

clintonb (changes requested):

course_discovery/apps/core/migrations/0004_partner_site.py
clintonb: https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete

PROTECT is what we want here.

vkaracic:
@clintonb comments addressed.

course_discovery/apps/core/tests/factories.py
```diff
@@ -1,12 +1,21 @@
...
 
+
 USER_PASSWORD = 'password'
 
 
+class SiteFactory(factory.DjangoModelFactory):
+    class Meta:
+        model = Site
+
+    domain = factory.fuzzy.FuzzyText(suffix='.fake')
```
clintonb: Same comment as before: Fuzzy is deprecated. Use Faker. See https://github.com/edx/credentials/blob/master/credentials/apps/core/tests/factories.py#L25-L30.

course_discovery/apps/core/tests/factories.py
```diff
@@ -31,11 +40,12 @@ class PartnerFactory(factory.DjangoModelFactory):
     programs_api_url = '{root}/api/programs/v1/'.format(root=FuzzyUrlRoot().fuzz())
     marketing_site_api_url = '{root}/api/courses/v1/'.format(root=FuzzyUrlRoot().fuzz())
     marketing_site_url_root = '{root}/'.format(root=FuzzyUrlRoot().fuzz())
-    marketing_site_api_username = FuzzyText().fuzz()
-    marketing_site_api_password = FuzzyText().fuzz()
+    marketing_site_api_username = factory.fuzzy.FuzzyText().fuzz()
```
clintonb: Use Faker. See https://github.com/edx/credentials/blob/master/credentials/apps/core/tests/factories.py#L33 for examples.

clintonb (approved):
Use Faker.

PR #942, opened 2017-05-31
Refine start field boosting

rlucioni:
Decay scales that are too small can cause scores to quickly drop to 0, leaving you with no start field boosting at all. This change also switches to Gaussian start field decay, which better suits our boosting needs. The Gaussian function decays slowly, then rapidly, then slowly again. This creates a cluster of high-scoring results whose start field is near the present, and a spread of results whose starts are further from the present, either in the past or in future.

LEARNER-993

@edx/learner-growth @bderusha @amangano-edx FYI.

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
+++ b/course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
@@ -1,4 +1,25 @@
+# pylint: disable=line-too-long
 def get_elasticsearch_boost_config():
+    """
+    Custom boosting config used to control relevance scores.
+
+    For a good primer on the theory behind relevance scoring, read
+    https://www.elastic.co/guide/en/elasticsearch/guide/1.x/scoring-theory.html.
+
+    If you're trying to tweak this config locally with a small dataset and are
+    seeing strange relevance scores, keep in mind that Elasticsearch computes
+    shard-local relevance scores. This causes small discrepancies between relevance
+    scores across shards that become more pronounced with small data sets. For more
+    on this, see https://www.elastic.co/blog/understanding-query-then-fetch-vs-dfs-query-then-fetch.
+
+    Use search_type=dfs_query_then_fetch to counteract this effect when querying
+    Elasticsearch while debugging. The search_type parameter must be passed in the
+    querystring. For more, see https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-body.html
+    and https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-search-type.html.
+
+    To see how a given hit's score was computed, use the explain parameter:
+    https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-explain.html
+    """
     elasticsearch_boost_config = {
         'function_score': {
             'boost_mode': 'sum',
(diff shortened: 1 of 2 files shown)


course_discovery/settings/devstack.py
```diff
@@ -11,14 +11,7 @@
...
 
-
-HAYSTACK_CONNECTIONS = {
-    'default': {
-        'ENGINE': 'haystack.backends.elasticsearch_backend.ElasticsearchSearchEngine',
-        'URL': 'http://edx.devstack.elasticsearch:9200/',
-        'INDEX_NAME': 'catalog',
-    },
-}
+HAYSTACK_CONNECTIONS['default']['URL'] = 'http://edx.devstack.elasticsearch:9200/'
```
rlucioni, line 14: This threw me for a loop when debugging. All we need to change when developing locally with Docker is the Elasticsearch URL. This was previously doing too much by also changing the `ENGINE` option, causing our boosting config to be excluded from search queries.

rlucioni:
Bonus clip of me finally understanding Elasticsearch:

course_discovery/settings/devstack.py
MatthewPiatetsky: is this the settings file for docker devstack version of course discovery?

rlucioni: @MatthewPiatetsky yes.

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
```diff
@@ -8,7 +25,34 @@ def get_elasticsearch_boost_config():
...
+                # Be careful with scales less than 30 days! Scales that are too
+                # small can cause scores to quickly drop to 0, leaving you with
+                # no start field boosting at all.
+                #
+                # For more on how decay functions work, especially if you're thinking
+                # about changing this, read https://www.elastic.co/guide/en/elasticsearch/guide/1.x/decay-functions.html
+                # and https://www.elastic.co/guide/en/elasticsearch/reference/1.5/query-dsl-function-score-query.html#_decay_functions.
+                #
+                # For help visualizing the effect different decay functions can
+                # have on relevance scores, try https://codepen.io/xyu/full/MyQYjN.
```
MatthewPiatetsky, line 49: so is the rule of thumb that it will take 6 months for the boost to go to 0 and about 7.5 weeks to go to 1/2?  
maybe we could add something like that as a comment

rlucioni: That's roughly true. I'd rather not leave a comment like that which may go stale quickly and might lead someone to jump to the wrong conclusion about how the overall boosting is being done. If you need to answer a question about how boosting changes over time, you should be inspecting the computed scores.

MatthewPiatetsky: ok that's fine but i think it would be useful to know this general number to help debugging similar issues in the future

MatthewPiatetsky:
👍

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
rlucioni: I'll leave a note about the `explain` parameter.

bderusha:
Belated 👍

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
```diff
@@ -8,7 +29,34 @@ def get_elasticsearch_boost_config():
...
+                {
+                    'gauss': {
+                        'start': {
+                            'origin': 'now',
+                            'decay': 0.95,
+                            'scale': '30d'
```
bderusha, line 55: Won't this mean that we don't decay the score until each 30 day window has passed?  That seems a bit long to me.  Or am I misunderstanding how the model works?

I'm sure you tested it for reasonable inputs and outputs, just curious since it seems like too many things would fit into the 60 day window on either side of `now`

rlucioni: @bderusha no, the decay function is continuous, not a step function. A 30-day scale yields a curve with a slope that's less steep than what you get when you use a 1-day scale. The behavior you're describing is closer to how the [`offset`](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/query-dsl-function-score-query.html#_decay_functions) parameter works. I thought [this tool](https://codepen.io/xyu/full/MyQYjN) did a nice job helping to visualize the effect these parameters have. If you have other questions or this isn't clear, let me know; I'm happy to talk through it!

PR #950, opened 2017-06-02
Added Dockerfile

clintonb:
- Added simplified Dockerfile that does not rely on Ansible
- Updated devstack settings to read from environment variables


feanil:
@clintonb is this still relevant? Should this be closed?

PR #955, opened 2017-06-05
Remove XSeries data loader

rlucioni:
XSeries will now be managed using discovery's Django admin like all other program types. Changes to XSeries program may now trigger publication to the marketing site.

LEARNER-1148

@edx/learner-growth FYI


MatthewPiatetsky:
code looks fine to me.  
Can you verify that all the fields that should be published from catalog to drupal get published correctly?

rlucioni:
@MatthewPiatetsky I've done so. This is ready to go, but I won't merge it until [LEARNER-1257](https://openedx.atlassian.net/browse/LEARNER-1257) is completed.

PR #959, opened 2017-06-06
Only check eligible runs that are published

mjfrey:
Programs can contain course runs that are no longer published. We therefore only want to check if published  runs are eligible not all of them.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_admin.py
+++ b/course_discovery/apps/course_metadata/tests/test_admin.py
@@ -387,7 +387,7 @@ class ProgramEligibilityFilterTests(TestCase):
             courses=[course_run.course],
             one_click_purchase_enabled=True,
         )
-        with self.assertNumQueries(10):
+        with self.assertNumQueries(11):
             self.assertEqual(
                 list(program_filter.queryset({}, Program.objects.all())),
                 [one_click_purchase_eligible_program]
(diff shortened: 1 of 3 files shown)


clintonb (commented):
You need tests and a ticket number in your commit message.

mjfrey:
@edx/helio please review

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -387,7 +387,7 @@ def test_queryset_method_returns_eligible_programs(self):
             courses=[course_run.course],
             one_click_purchase_enabled=True,
         )
-        with self.assertNumQueries(10):
+        with self.assertNumQueries(11):
```
clintonb, line 390: Nit: you can maintain the existing query count by filtering via list comprehension rather than the database.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -523,6 +523,30 @@ def test_one_click_purchase_eligible(self):
         )
         self.assertTrue(program.is_program_eligible_for_one_click_purchase)
 
+    def test_one_click_purchase_eligible_with_unpublished_runs(self):
+        """ Verify that program with unpublished course runs is one click purchase eligible. """
+
+        verified_seat_type, __ = SeatType.objects.get_or_create(name=Seat.VERIFIED)
+        program_type = factories.ProgramTypeFactory(applicable_seat_types=[verified_seat_type])
+        published_course_run = factories.CourseRunFactory(
+            end=None,
```
clintonb, line 532: The dates don't matter. It's the `status` field that should be explicitly set here.

clintonb (approved):

PR #960, opened 2017-06-07
Removed RequireJS

clintonb:
RequireJS is not actually used anywhere. Ultimately Webpack will be used instead.

ECOM-5776


clintonb:
FYI @edx/learner

PR #961, opened 2017-06-07
Displayed specs for add new instructor image and bio

asadazam93:
ECOM-7802

On the add new instructor pop up page displayed specs for instructor image and bio.
screen shot 2017-06-07 at 3 00 15 pm


awais786 (commented):
Looks good

PR #977, opened 2017-06-13
Filter API data by partner

vkaracic:
LEARNER-1119

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -369,8 +369,8 @@ class OrganizationSerializer(TaggitSerializer, MinimalOrganizationSerializer):
     tags = TagListSerializerField()
 
     @classmethod
-    def prefetch_queryset(cls):
-        return Organization.objects.all().select_related('partner').prefetch_related('tags')
+    def prefetch_queryset(cls, partner):
+        return Organization.objects.filter(partner=partner).select_related('partner').prefetch_related('tags')
 
     class Meta(MinimalOrganizationSerializer.Meta):
         fields = MinimalOrganizationSerializer.Meta.fields + (
@@ -551,18 +551,18 @@ class CourseSerializer(MinimalCourseSerializer):
     marketing_url = serializers.SerializerMethodField()
 
     @classmethod
-    def prefetch_queryset(cls, queryset=None, course_runs=None):
+    def prefetch_queryset(cls, queryset=None, course_runs=None, partner=None):
         # Explicitly check for None to avoid returning all Courses when the
         # queryset passed in happens to be empty.
-        queryset = queryset if queryset is not None else Course.objects.all()
+        queryset = queryset if queryset is not None else Course.objects.filter(partner=partner)
 
         return queryset.select_related('level_type', 'video', 'partner').prefetch_related(
             'expected_learning_items',
             'prerequisites',
             'subjects',
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
-            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
-            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
+            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
         )
 
     class Meta(MinimalCourseSerializer.Meta):
@@ -586,20 +586,20 @@ class CourseWithProgramsSerializer(CourseSerializer):
     programs = serializers.SerializerMethodField()
 
     @classmethod
-    def prefetch_queryset(cls, queryset=None, course_runs=None):
+    def prefetch_queryset(cls, queryset=None, course_runs=None, partner=None):
         """
         Similar to the CourseSerializer's prefetch_queryset, but prefetches a
         filtered CourseRun queryset.
         """
-        queryset = queryset if queryset is not None else Course.objects.all()
+        queryset = queryset if queryset is not None else Course.objects.filter(partner=partner)
 
         return queryset.select_related('level_type', 'video', 'partner').prefetch_related(
             'expected_learning_items',
             'prerequisites',
             'subjects',
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
-            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
-            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
+            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
         )
 
     def get_course_runs(self, course):
@@ -634,20 +634,20 @@ class CatalogCourseSerializer(CourseSerializer):
     course_runs = serializers.SerializerMethodField()
 
     @classmethod
-    def prefetch_queryset(cls, queryset=None, course_runs=None):
+    def prefetch_queryset(cls, queryset=None, course_runs=None, partner=None):
         """
         Similar to the CourseSerializer's prefetch_queryset, but prefetches a
         filtered CourseRun queryset.
         """
-        queryset = queryset if queryset is not None else Course.objects.all()
+        queryset = queryset if queryset is not None else Course.objects.filter(partner=partner)
 
         return queryset.select_related('level_type', 'video', 'partner').prefetch_related(
             'expected_learning_items',
             'prerequisites',
             'subjects',
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
-            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
-            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
+            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
         )
 
     def get_course_runs(self, course):
@@ -703,8 +703,8 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     type = serializers.SlugRelatedField(slug_field='name', queryset=ProgramType.objects.all())
 
     @classmethod
-    def prefetch_queryset(cls):
-        return Program.objects.all().select_related('type', 'partner').prefetch_related(
+    def prefetch_queryset(cls, partner):
+        return Program.objects.filter(partner=partner).select_related('type', 'partner').prefetch_related(
             'excluded_course_runs',
             # `type` is serialized by a third-party serializer. Providing this field name allows us to
             # prefetch `applicable_seat_types`, a m2m on `ProgramType`, through `type`, a foreign key to
@@ -828,7 +828,7 @@ class ProgramSerializer(MinimalProgramSerializer):
     applicable_seat_types = serializers.SerializerMethodField()
 
     @classmethod
-    def prefetch_queryset(cls):
+    def prefetch_queryset(cls, partner):
         """
         Prefetch the related objects that will be serialized with a `Program`.
 
@@ -836,7 +836,7 @@ class ProgramSerializer(MinimalProgramSerializer):
         chain of related fields from programs to course runs (i.e., we want control over
         the querysets that we're prefetching).
         """
-        return Program.objects.all().select_related('type', 'video', 'partner').prefetch_related(
+        return Program.objects.filter(partner=partner).select_related('type', 'video', 'partner').prefetch_related(
             'excluded_course_runs',
             'expected_learning_items',
             'faq',
@@ -847,9 +847,9 @@ class ProgramSerializer(MinimalProgramSerializer):
             'type__applicable_seat_types',
             # We need the full Course prefetch here to get CourseRun information that methods on the Program
             # model iterate across (e.g. language). These fields aren't prefetched by the minimal Course serializer.
-            Prefetch('courses', queryset=CourseSerializer.prefetch_queryset()),
-            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
-            Prefetch('credit_backing_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
+            Prefetch('courses', queryset=CourseSerializer.prefetch_queryset(partner=partner)),
+            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('credit_backing_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
             Prefetch('corporate_endorsements', queryset=CorporateEndorsementSerializer.prefetch_queryset()),
             Prefetch('individual_endorsements', queryset=EndorsementSerializer.prefetch_queryset()),
         )
--- a/course_discovery/apps/api/tests/mixins.py
+++ b/course_discovery/apps/api/tests/mixins.py
@@ -1,12 +1,15 @@
-from django.conf import settings
-from django.contrib.sites.models import Site
+from django.test import RequestFactory
 
 from course_discovery.apps.core.tests.factories import PartnerFactory, SiteFactory
 
 
-class PartnerMixin(object):
+class SiteMixin(object):
     def setUp(self):
-        super(PartnerMixin, self).setUp()
-        Site.objects.all().delete()
-        self.site = SiteFactory(id=settings.SITE_ID)
+        super(SiteMixin, self).setUp()
+        domain = 'testserver.fake'
+        self.client = self.client_class(SERVER_NAME=domain)
+        self.site = SiteFactory(domain=domain)
         self.partner = PartnerFactory(site=self.site)
+
+        self.request = RequestFactory(SERVER_NAME=self.site.domain).get('')
+        self.request.site = self.site
--- a/course_discovery/apps/api/tests/mixins.py
+++ b/course_discovery/apps/api/tests/mixins.py
@@ -1,0 +1,12 @@
+from django.conf import settings
+from django.contrib.sites.models import Site
+
+from course_discovery.apps.core.tests.factories import PartnerFactory, SiteFactory
+
+
+class PartnerMixin(object):
+    def setUp(self):
+        super(PartnerMixin, self).setUp()
+        Site.objects.all().delete()
+        self.site = SiteFactory(id=settings.SITE_ID)
+        self.partner = PartnerFactory(site=self.site)
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -21,6 +21,7 @@ from course_discovery.apps.api.serializers import (
     ProgramSerializer, ProgramTypeSerializer, SeatSerializer, SubjectSerializer, TypeaheadCourseRunSearchSerializer,
     TypeaheadProgramSearchSerializer, VideoSerializer
 )
+from course_discovery.apps.api.tests.mixins import PartnerMixin
 from course_discovery.apps.catalogs.tests.factories import CatalogFactory
 from course_discovery.apps.core.models import User
 from course_discovery.apps.core.tests.factories import UserFactory
@@ -96,7 +97,7 @@ class CatalogSerializerTests(ElasticsearchTestMixin, TestCase):
         self.assertEqual(User.objects.filter(username=username).count(), 0)  # pylint: disable=no-member
 
 
-class MinimalCourseSerializerTests(TestCase):
+class MinimalCourseSerializerTests(PartnerMixin, TestCase):
     serializer_class = MinimalCourseSerializer
 
     def get_expected_data(self, course, request):
@@ -113,8 +114,8 @@ class MinimalCourseSerializerTests(TestCase):
 
     def test_data(self):
         request = make_request()
-        organizations = OrganizationFactory()
-        course = CourseFactory(authoring_organizations=[organizations])
+        organizations = OrganizationFactory(partner=self.partner)
+        course = CourseFactory(authoring_organizations=[organizations], partner=self.partner)
         CourseRunFactory.create_batch(2, course=course)
         serializer = self.serializer_class(course, context={'request': request})
         expected = self.get_expected_data(course, request)
@@ -177,9 +178,10 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
     def setUp(self):
         super().setUp()
         self.request = make_request()
-        self.course = CourseFactory()
+        self.course = CourseFactory(partner=self.partner)
         self.deleted_program = ProgramFactory(
             courses=[self.course],
+            partner=self.partner,
             status=ProgramStatus.Deleted
         )
 
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -21,7 +21,7 @@ from course_discovery.apps.api.serializers import (
     ProgramSerializer, ProgramTypeSerializer, SeatSerializer, SubjectSerializer, TypeaheadCourseRunSearchSerializer,
     TypeaheadProgramSearchSerializer, VideoSerializer
 )
-from course_discovery.apps.api.tests.mixins import PartnerMixin
+from course_discovery.apps.api.tests.mixins import SiteMixin
 from course_discovery.apps.catalogs.tests.factories import CatalogFactory
 from course_discovery.apps.core.models import User
 from course_discovery.apps.core.tests.factories import UserFactory
@@ -97,7 +97,7 @@ class CatalogSerializerTests(ElasticsearchTestMixin, TestCase):
         self.assertEqual(User.objects.filter(username=username).count(), 0)  # pylint: disable=no-member
 
 
-class MinimalCourseSerializerTests(PartnerMixin, TestCase):
+class MinimalCourseSerializerTests(SiteMixin, TestCase):
     serializer_class = MinimalCourseSerializer
 
     def get_expected_data(self, course, request):
--- a/course_discovery/apps/api/v1/tests/test_views/mixins.py
+++ b/course_discovery/apps/api/v1/tests/test_views/mixins.py
@@ -4,6 +4,7 @@ import json
 
 import responses
 from django.conf import settings
+from rest_framework.test import APITestCase as RestAPITestCase
 from rest_framework.test import APIRequestFactory
 
 from course_discovery.apps.api.serializers import (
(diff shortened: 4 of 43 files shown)


vkaracic:
@edx/helio please review.

.travis/run_tests.sh
```diff
@@ -28,4 +28,5 @@ make validate_translations
 # Compile assets and run validation
 make clean_static
 make static
-xvfb-run make validate
```
clintonb: Revert.

course_discovery/apps/api/serializers.py
```diff
@@ -551,25 +551,25 @@ class CourseSerializer(MinimalCourseSerializer):
...
+            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
         )
 
     class Meta(MinimalCourseSerializer.Meta):
         model = Course
         fields = MinimalCourseSerializer.Meta.fields + (
             'short_description', 'full_description', 'level_type', 'subjects', 'prerequisites',
-            'expected_learning_items', 'video', 'sponsors', 'modified', 'marketing_url',
+            'expected_learning_items', 'video', 'sponsors', 'modified', 'marketing_url', 'partner'
```
clintonb: Why is the partner field being added to the API response?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -96,9 +97,12 @@ def test_invalid_data_user_create(self):
         self.assertEqual(User.objects.filter(username=username).count(), 0)  # pylint: disable=no-member
 
 
-class MinimalCourseSerializerTests(TestCase):
+class MinimalCourseSerializerTests(PartnerMixin, TestCase):
     serializer_class = MinimalCourseSerializer
 
+    def setUp(self):
```
clintonb: Remove. This should not be necessary.

course_discovery/apps/api/v1/tests/test_views/mixins.py
```diff
@@ -88,3 +91,27 @@ def mock_user_info_response(self, user, status=200):
             content_type='application/json',
             status=status
         )
+
+
+class PartnerMixin(object):
+    def setUp(self):
+        super(PartnerMixin, self).setUp()
+
+        site = Site.objects.get_current()
```
clintonb: Why are you doing this? Just make the `Site` object. Actually, use `PartnerFactory` to make everything.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -27,16 +27,21 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
     pagination_class = ProxiedPagination
 
     def get_queryset(self):
+        partner = get_current_site(self.request).partner
```
clintonb: `partner = self.request.site.partner`

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -27,16 +27,21 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
...
 
         if q:
             queryset = Course.search(q)
-            queryset = self.get_serializer_class().prefetch_queryset(queryset=queryset)
+            queryset = self.get_serializer_class().prefetch_queryset(queryset=queryset, partner=partner)
         else:
             if get_query_param(self.request, 'include_hidden_course_runs'):
-                course_runs = CourseRun.objects.all()
+                course_runs = CourseRun.objects.filter(
+                    course__partner=partner
```
clintonb: One line.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -27,16 +27,21 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
...
+                )
             else:
-                course_runs = CourseRun.objects.exclude(hidden=True)
+                course_runs = CourseRun.objects.filter(
+                    course__partner=partner
```
clintonb: One line.

course_discovery/apps/api/v1/views/organizations.py
```diff
@@ -1,3 +1,4 @@
+from django.contrib.sites.shortcuts import get_current_site
```
clintonb: You should not need this function anywhere. Use [`CurrentSiteMiddleware`](https://docs.djangoproject.com/en/1.11/ref/middleware/#django.contrib.sites.middleware.CurrentSiteMiddleware).

course_discovery/conf/locale/eo/LC_MESSAGES/djangojs.po
```diff
@@ -7,14 +7,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-06-09 21:46+0500\n"
```
clintonb: If you didn't actually change any translations, there is no need to commit translation files.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -21,6 +21,7 @@
     ProgramSerializer, ProgramTypeSerializer, SeatSerializer, SubjectSerializer, TypeaheadCourseRunSearchSerializer,
     TypeaheadProgramSearchSerializer, VideoSerializer
 )
+from course_discovery.apps.api.v1.tests.test_views.mixins import PartnerMixin
```
clintonb: The `api` module should not use code from `api.v1`. Move the shared code to `api`.

course_discovery/apps/api/v1/tests/test_views/mixins.py
```diff
@@ -88,3 +91,27 @@ def mock_user_info_response(self, user, status=200):
...
+            site = SiteFactory()
+
+        if hasattr(site, 'partner'):
+            self.partner = site.partner
+        else:
+            # DEFAULT_PARTNER_ID is used explicitly here to avoid issues with differences in
+            # auto-incrementing behavior across databases. Otherwise, it's not safe to assume
+            # that the partner created here will always have id=DEFAULT_PARTNER_ID.
+            self.partner = PartnerFactory(
+                id=settings.DEFAULT_PARTNER_ID,
```
clintonb: We no longer need `DEFAULT_PARTNER_ID` since everything should be filtered.

clintonb (changes requested):

course_discovery/apps/api/v1/views/courses.py
vkaracic: I've tried that already and keep getting `'Request' object has no attribute 'site'` errors.

afzaledx:
@vkaracic the api/v1/programs/?uuids_only=1 is not filtering by partner.

course_discovery/apps/api/v1/views/courses.py
vkaracic: Never mind, works after adding the middleware.

afzaledx:
SITE_ID = None in settings (base.py)?

vkaracic:
@clintonb comments addressed!

.travis/run_tests.sh
```diff
@@ -28,4 +28,4 @@ make validate_translations
...
+xvfb-run make validate
```
clintonb: Add a new line and/or revert all changes to this file.

course_discovery/apps/api/tests/mixins.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib.sites.models import Site
+
+from course_discovery.apps.core.tests.factories import PartnerFactory
+
+
+class PartnerMixin(object):
+    def setUp(self):
+        super(PartnerMixin, self).setUp()
+        site = Site.objects.get_current()
```
clintonb: Stop using `Site.objects.get_current()` for tests! Anything that relies on a global request or other global variable in tests should be avoided like the plague. Use `SiteFactory`, and update it to create a related partner (if it doesn't already).

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -313,10 +306,13 @@ def _test_exclude_program_types(self, program_status):
 
 
 @ddt.ddt
-class AggregateSearchViewSetTests(DefaultPartnerMixin, SerializationMixin, LoginMixin, ElasticsearchTestMixin,
+class AggregateSearchViewSetTests(SerializationMixin, LoginMixin, ElasticsearchTestMixin,
                                   SynonymTestMixin, APITestCase):
     path = reverse('api:v1:search-all-facets')
 
+    def setUp(self):
```
clintonb: This is not necessary.

course_discovery/apps/core/views.py
```diff
@@ -15,6 +15,18 @@
 User = get_user_model()
 
 
+def get_database_status():
```
clintonb: Why are you touching in this file? None of it is related to the API. If this is necessary, it should be moved to a separate pull request.

course_discovery/conf/locale/en/LC_MESSAGES/djangojs.po
```diff
@@ -7,14 +7,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-06-14 17:09+0500\n"
```
clintonb: If you didn't actually touch strings, there is no need to commit translations.

course_discovery/settings/base.py
```diff
@@ -79,10 +79,11 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
+    'django.contrib.sites.middleware.CurrentSiteMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'social_django.middleware.SocialAuthExceptionMiddleware',
     'waffle.middleware.WaffleMiddleware',
-    'simple_history.middleware.HistoryRequestMiddleware',
```
clintonb: Revert and restore the trailing comma.

clintonb (changes requested):

course_discovery/apps/core/views.py
vkaracic: It's related to the new middleware addition. The first database interaction with the new middleware is fetching the site (IIRC), not the one in this function, so [this test](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/core/tests/test_views.py#L24-L27) fails now.

vkaracic:
@clintonb comments addressed!

afzaledx:
@vkaracic @clintonb , don't we require SITE_ID = None in settings for multi tenancy to work?

course_discovery/apps/api/tests/mixins.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from django.contrib.sites.models import Site
+
+from course_discovery.apps.core.tests.factories import PartnerFactory, SiteFactory
+
+
+class PartnerMixin(object):
+    def setUp(self):
+        super(PartnerMixin, self).setUp()
+        Site.objects.all().delete()
```
clintonb: Why? If this is absolutely needed, add a comment explaining why.

course_discovery/apps/api/tests/mixins.py
```diff
@@ -0,0 +1,12 @@
...
+        self.site = SiteFactory(id=settings.SITE_ID)
```
clintonb: We shouldn't be specifying the site in settings. Also, look at the SiteMixin in edx/ecommerce. Call SiteConfigurationFactory, and pull Site and Partner from it.

course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
```diff
@@ -185,8 +185,7 @@ def test_courses(self):
                 # Any course appearing in the response must have at least one serialized run.
                 assert len(response.data['results'][0]['course_runs']) > 0
             else:
-                with self.assertNumQueries(3):
```
clintonb, line 188: Why was this removed?

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -27,17 +26,19 @@ def test_authentication(self):
 
     def assert_response_data_valid(self, response, organizations, many=True):
         """ Asserts the response data (only) contains the expected organizations. """
-
         actual = response.data
+        serializer_data = self.serialize_organization(organizations, many=many)
+
         if many:
             actual = actual['results']
+            actual = sorted(actual, key=lambda k: k['uuid'])
```
clintonb: What changed to cause us to need to sort the data? This is a code smell because the data should be returned with a stable sort.

course_discovery/apps/core/views.py
clintonb: Replicate what we've done for ecommerce and other services: patch the middleware. See https://github.com/edx/ecommerce/blob/master/ecommerce/core/tests/test_views.py.

clintonb: The boilerplate should ideally be consistent across services. Ultimately, we will move it all to a common Django app.

clintonb (changes requested):
Make sure that SITE_ID is not set anywhere in settings. If it is, we will have problems in production.

course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
vkaracic: It started to behave flaky. When I add `with self.assertNumQueries(4)` I get `AssertionError: 3 != 4 : 3 queries executed, 4 expected`. When I add `with self.assertNumQueries(3)` I would get `AssertionError: 4 != 3 : 4 queries executed, 3 expected`. The query that differs is related to the `django_site` table.

course_discovery/apps/api/tests/mixins.py
vkaracic: There is no `SiteConfiguration` in Course Discovery.

course_discovery/apps/api/tests/mixins.py
vkaracic: There already was a site object in (not all) tests. I just wanted to make sure that the site that is created here is the only one. Something similar that is being done [here](https://github.com/edx/ecommerce/blob/master/ecommerce/tests/mixins.py#L244). If another site is required it could be added in the actual test case.

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
vkaracic: The only change that could have affected this is the added filter [here](https://github.com/edx/course-discovery/pull/977/files#diff-651fda19c2550b81fbdc3c7de19a0dbbR373). I'll add an explicit `order_by` clause there and see if this ordering is still needed.

vkaracic:
@clintonb I stumbled on a problem here. When I remove `SITE_ID` I would get errors like:
```
____________________________________________________________________________ AdminTests.test_comment_detail_form _____________________________________________________________________________

self = <course_discovery.apps.publisher_comments.tests.test_admin.AdminTests testMethod=test_comment_detail_form>

    def test_comment_detail_form(self):
        """ Verify in admin panel comment detail form contain the custom modified field. """
        # pylint: disable=no-member
>       resp = self.client.get(reverse('admin:publisher_comments_comments_change', args=(self.comment.id,)))

course_discovery/apps/publisher_comments/tests/test_admin.py:25: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../venvs/discovery/lib/python3.5/site-packages/django/test/client.py:529: in get
    **extra)
../venvs/discovery/lib/python3.5/site-packages/django/test/client.py:333: in get
    return self.generic('GET', path, secure=secure, **r)
../venvs/discovery/lib/python3.5/site-packages/django/test/client.py:409: in generic
    return self.request(**r)
../venvs/discovery/lib/python3.5/site-packages/django/test/client.py:494: in request
    six.reraise(*exc_info)
../venvs/discovery/lib/python3.5/site-packages/django/utils/six.py:686: in reraise
    raise value
../venvs/discovery/lib/python3.5/site-packages/django/core/handlers/exception.py:42: in inner
    response = get_response(request)
../venvs/discovery/lib/python3.5/site-packages/django/core/handlers/base.py:244: in _legacy_get_response
    response = middleware_method(request)
../venvs/discovery/lib/python3.5/site-packages/django/contrib/sites/middleware.py:12: in process_request
    request.site = get_current_site(request)
../venvs/discovery/lib/python3.5/site-packages/django/contrib/sites/shortcuts.py:15: in get_current_site
    return Site.objects.get_current(request)
../venvs/discovery/lib/python3.5/site-packages/django/contrib/sites/models.py:67: in get_current
    return self._get_site_by_request(request)
../venvs/discovery/lib/python3.5/site-packages/django/contrib/sites/models.py:44: in _get_site_by_request
    SITE_CACHE[host] = self.get(domain__iexact=host)
../venvs/discovery/lib/python3.5/site-packages/django/db/models/manager.py:85: in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <QuerySet [<Site: example.com>]>, args = (), kwargs = {'domain__iexact': 'testserver'}, clone = <QuerySet []>, num = 0

    def get(self, *args, **kwargs):
        """
            Performs the query and returns a single object matching the given
            keyword arguments.
            """
        clone = self.filter(*args, **kwargs)
        if self.query.can_filter() and not self.query.distinct_fields:
            clone = clone.order_by()
        num = len(clone)
        if num == 1:
            return clone._result_cache[0]
        if not num:
            raise self.model.DoesNotExist(
                "%s matching query does not exist." %
>               self.model._meta.object_name
            )
E           django.contrib.sites.models.DoesNotExist: Site matching query does not exist.

../venvs/discovery/lib/python3.5/site-packages/django/db/models/query.py:385: DoesNotExist
```
If you take a look at the stacktrace you'll see that it eventually hits `Site.objects.get_current(request)`, which in turn [gets the site from either `settings.SITE_ID` or the request](https://github.com/django/django/blob/master/django/contrib/sites/models.py#L50).
On Ecommerce this seems to be done by [setting the global request variable](https://github.com/edx/ecommerce/blob/master/ecommerce/tests/mixins.py#L264) which we want to avoid here.
Any ideas how to set the site in this case?

clintonb:
See the `SiteMixin` in edx/ecommerce: https://github.com/edx/ecommerce/blob/0a614c778146cf99ac03a7b4868078ee92ae0d7a/ecommerce/tests/mixins.py#L240-L242

vkaracic:
@clintonb please review the last changes.

course_discovery/apps/api/tests/mixins.py
```diff
@@ -0,0 +1,10 @@
+from course_discovery.apps.core.tests.factories import PartnerFactory, SiteFactory
+
+
+class SiteMixin(object):
+    def setUp(self):
+        super(SiteMixin, self).setUp()
+        domain = 'testserver'
+        self.client = self.client_class(SERVER_NAME=domain)
+        self.site = SiteFactory(domain=domain, name='Test server')
```
clintonb: Don't override fields unless you actually need to. Ditch the `name` kwarg.

course_discovery/apps/api/tests/mixins.py
```diff
@@ -0,0 +1,10 @@
+from course_discovery.apps.core.tests.factories import PartnerFactory, SiteFactory
+
+
+class SiteMixin(object):
+    def setUp(self):
+        super(SiteMixin, self).setUp()
+        domain = 'testserver'
```
clintonb: Nit: testserver.fake

Use a domain name that is obviously fake and cannot be confused with Django's defaults.

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
clintonb: Any update on this? We should have a defined, stable, sort. Tests should not need to sort the results.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -190,7 +191,8 @@ def test_new_program_without_courses(self):
         self.assertEqual(response.status_code, 200)
 
 
-class ProgramAdminFunctionalTests(LiveServerTestCase):
+@override_settings(SITE_ID=1)
```
clintonb: Why is this necessary? You probably want the following in `SiteMixin.setUp()`:

```python
domain = 'testserver.fake'
self.client = self.client_class(SERVER_NAME=domain)
```

See the following in edx/ecommerce:
* https://github.com/edx/ecommerce/blob/3873cad9fdbd846f3ede7077dd6bc4d5ae6fa724/ecommerce/tests/mixins.py#L241-L242
* https://github.com/edx/ecommerce/blob/57bcc22b5337a86e4cff3baa75cf6aeb18bcd331/ecommerce/tests/testcases.py

course_discovery/apps/publisher/api/serializers.py
```diff
@@ -40,7 +40,8 @@ def update(self, instance, validated_data):
         former_user = instance.user
         instance = super(CourseUserRoleSerializer, self).update(instance, validated_data)
         if not instance.role == PublisherUserRole.CourseTeam:
-            send_change_role_assignment_email(instance, former_user)
+            request = self.context.get('request')
```
clintonb: What happens if this returns `None`? It may be best to use `self.context['request']` and make the failure more obvious.

course_discovery/apps/publisher/api/serializers.py
```diff
@@ -104,17 +105,18 @@ def update(self, instance, validated_data):
         instance = super(CourseRunSerializer, self).update(instance, validated_data)
         preview_url = validated_data.get('preview_url')
         lms_course_id = validated_data.get('lms_course_id')
+        request = self.context.get('request')
```
clintonb: Same comment as above. Use `self.context['request']`.

course_discovery/apps/publisher/api/tests/test_serializers.py
```diff
@@ -255,6 +257,7 @@ def setUp(self):
         self.request = RequestFactory()
```
clintonb: `RequestFactory()` returns a `RequestFactory`. `RequestFactory().get('')` will return a `Request`. It will be best if you simply create `self.request` in `SiteMixin` so that all tests benefit. See https://github.com/edx/ecommerce/blob/3873cad9fdbd846f3ede7077dd6bc4d5ae6fa724/ecommerce/tests/mixins.py#L261-L263.

course_discovery/apps/publisher/api/tests/test_serializers.py
```diff
@@ -299,6 +302,7 @@ def setUp(self):
...
```
clintonb: Same as above.

clintonb (commented):
1. Work with @edx/devops to ensure the `SITE_ID` setting is not set in secure configuration. I have confirmed it is not set in edx/edx-internal.
2. Resolve the sorting question.
3. Squash your commits.

course_discovery/apps/course_metadata/tests/test_admin.py
vkaracic: I don't think that will work in this case since this is a selenium test where ``self.client`` isn't used at all.

clintonb: Add the following to your `setUp()`:

```python
self.site.domain = self.live_server_url.strip('http://')
self.site.save()
```

This will create a `Site` corresponding to the hostname used by the live server.

feanil:
@vkaracic @clintonb `SITE_ID` is not set in any configuration.  It looks like it was never something that ansible knew about.  https://github.com/edx/configuration/search?utf8=%E2%9C%93&q=SITE_ID&type=

It's not a configuration we could override it looks like.

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
vkaracic: I've spent the better half of today trying to understand why this happens and I still can't figure out. I've tried adding a `order_by` clause to the query on the serializer, that didn't work. I've tried adding it to both serializer query and the query in the test which also didn't work. 
Tried casting them to lists in the assertion, didn't work. Tried removing all the organizations in the `setUp`, didn't work. Tried adding a `.filter(partner=self.partner)` in the [couple](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_organizations.py#L68) of [places](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_organizations.py#L98) where Organizations are retrieve in the test, that didn't work either.

What's strange is that when I tried casting them to lists, I would run only the `test_organizations.py` test file and all the tests would pass, and my first thought was that it must be somehow related to Django's lazy evaluation, but when running the whole test suite they would fail (because of different order of objects), so I was back to the drawing board.

I'll need some help with this. Can you take a quick look, please?

vkaracic: @clintonb ping ^

clintonb: Try adding the following to `assert_list_uuid_filter`: `organizations = sorted(organizations, key=lambda o: o.created)`

vkaracic: That doesn't work either.

clintonb (approved):
Digging into that sorting issue will take more time than I have. I recommend sorting in the specific test, and adding a comment explaining the issue.

vkaracic:
@marjev found a more appropriate assertion for our issue. @clintonb please take a final(ish) look :)

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -27,17 +27,17 @@ def test_authentication(self):
...
     def assert_response_data_valid(self, response, organizations, many=True):
         """ Asserts the response data (only) contains the expected organizations. """
-
         actual = response.data
+        serializer_data = self.serialize_organization(organizations, many=many)
         if many:
             actual = actual['results']
 
-        self.assertEqual(actual, self.serialize_organization(organizations, many=many))
+        self.assertCountEqual(actual, serializer_data)
```
clintonb, line 35: I disagree with this change. I would be more comfortable with you sorting the data and leaving a comment explaining why. Simply looking at the number of items doesn't validate the correct data is actually returned. If I modified the endpoint to simply return an array of empty dicts this test would pass.

vkaracic: Actually the name of the assertion is misleading. [ `Test that sequence first contains the same elements as second, regardless of their order.` ](https://docs.python.org/3.2/library/unittest.html#unittest.TestCase.assertCountEqual)

PR #983, opened 2017-06-14
Include this repo in Open edX tagging

nedbat:


nedbat:
@rlucioni amirite?

rlucioni (approved):
Yup.

PR #984, opened 2017-06-14
Tune Elasticsearch boost config

rlucioni:
Reward course runs with enrollable, paid seats. Penalize those without them. Slightly increase boost given to MicroMasters and Professional Certificate programs. Tweak self-paced and enrollable boosts to match. Taken together, this modified boost configuration yields search results which tend to promote course runs with enrollable, paid seats above those without them.

LEARNER-1374

@edx/learner-growth FYI.

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/backends.py
+++ b/course_discovery/apps/edx_haystack_extensions/backends.py
@@ -36,7 +36,7 @@ class SimpleQuerySearchBackendMixin(object):
             'auto_generate_phrase_queries': True,
         }
 
-        # https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-function-score-query.html
+        # https://www.elastic.co/guide/en/elasticsearch/reference/1.5/query-dsl-function-score-query.html
         function_score_config = get_elasticsearch_boost_config()['function_score']
         function_score_config['query'] = {
             'query_string': simple_query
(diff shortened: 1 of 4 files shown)


MatthewPiatetsky:
i'll review in depth tomorrow but i think it would be helpful to add a few more tests to the test boosting file (will be helpful resource for people to understand how things should be ordered)

rlucioni:
@MatthewPiatetsky good idea. Those tests are still accurate as they are, but I'll see if it makes sense to extend them to cover penalizing of course runs without enrollable, paid seats.

MatthewPiatetsky:
I'm not entirely sure why you made certain changes like boosting micromasters more, but if jasper said the search results were better then that's good.
perhaps in addition to adding a test specifically testing the new deboosting you added you could add one test that creates like 15 courses/programs all in order labeled like self_paced_enrollable_micromasters_course, self_paced_micromasters_course, instructor_paced_prof_cert_course, etc..., something like that would probably go a lot further to helping people understand how our boosting/search results work than a comment or doc
otherwise 👍

rlucioni:
@MatthewPiatetsky the program boosting tweaks keep those program types from being pushed too far down relative to course runs with enrollable, paid seats, and are meant to bring them up a bit in the results (Jasper and I noticed they were being pushed down a little too far as-is). I'll look into adding a larger test case which covers a mix of course runs and programs separately.

PR #986, opened 2017-06-15
Pre-Django 1.11 Updates

ahsan-ul-haq:
LEARNER-947

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_admin.py
+++ b/course_discovery/apps/course_metadata/tests/test_admin.py
@@ -1,6 +1,7 @@
 import itertools
 
 import ddt
+from bs4 import BeautifulSoup
 from django.contrib.contenttypes.models import ContentType
 from django.test import LiveServerTestCase, TestCase
 from django.urls import reverse
@@ -98,11 +99,12 @@ class AdminTests(TestCase):
         # add some new courses and course runs
         factories.CourseRunFactory.create_batch(2)
         response = self.client.get(reverse('admin_metadata:update_course_runs', args=(self.program.id,)))
-        html = '<input checked="checked" id="id_excluded_course_runs_0" '
-        html += 'name="excluded_course_runs" type="checkbox" value="{id}" />'.format(
-            id=self.excluded_course_run.id
+        response_content = BeautifulSoup(response.content)
+        attribute = response_content.find(
+            "input", {"value": self.excluded_course_run.id, "type": "checkbox", "name": "excluded_course_runs"}
         )
-        self.assertContains(response, html)
+        assert attribute is not None
+
         for run in self.course_runs:
             self.assertContains(response, run.key)
 
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -4,6 +4,8 @@ from datetime import datetime, timedelta
 
 import ddt
 import factory
+
+from bs4 import BeautifulSoup
 from django.conf import settings
 from django.contrib.auth.models import Group
 from django.contrib.sites.models import Site
@@ -174,11 +176,12 @@ class CreateCourseViewTests(TestCase):
         organization_extension = factories.OrganizationExtensionFactory()
         self.user.groups.add(organization_extension.group)
         response = self.client.get(reverse('publisher:publisher_courses_new'))
+        response_content = BeautifulSoup(response.content)
 
-        self.assertContains(
-            response,
-            '<select class="field-input input-select" id="id_organization" name="organization" required>'
+        organization_attribute = response_content.find(
+            "select", {"id": "id_organization", "name": "organization", "class": "field-input"}
         )
+        assert organization_attribute is not None
 
         new_organization_extension = factories.OrganizationExtensionFactory()
         response = self.client.get(reverse('publisher:publisher_courses_new'))
@@ -468,21 +471,20 @@ class CreateCourseRunViewTests(TestCase):
         response = self.client.get(
             reverse('publisher:publisher_course_runs_new', kwargs={'parent_course_id': self.course.id})
         )
+        response_content = BeautifulSoup(response.content)
+
+        pacing_type_attribute = response_content.find(
+            "input", {"value": latest_run.pacing_type, "type": "radio", "name": "pacing_type"}
+        )
+        seat_type_attribute = response_content.find("option", {"value": latest_seat.type})
+        price_attribute = response_content.find(
+            "input", {"value": latest_seat.price, "id": "id_price", "step": "0.01", "type": "number"}
+        )
 
         # Verify that existing course run and seat values auto populated on form.
-        expected_pacing = '<input checked="checked" class="field-input input-checkbox" '
-        'id="id_pacing_type_0" name="pacing_type" type="radio" value="{pacing}" />'.format(
-            pacing=latest_run.pacing_type
-        )
-        self.assertContains(response, expected_pacing)
-
-        expected_seat_type = '<option value="{seat_type}" selected="selected">'.format(seat_type=latest_seat.type)
-        self.assertContains(response, expected_seat_type)
-
-        expected_seat_price = 'id="id_price" name="price" step="0.01" type="number" value="{price}" />'.format(
-            price=latest_seat.price
-        )
-        self.assertContains(response, expected_seat_price)
+        assert pacing_type_attribute is not None
+        assert seat_type_attribute is not None
+        assert price_attribute is not None
 
     def test_credit_type_without_price(self):
         """ Verify that without credit price course-run cannot be created with credit seat type. """
@@ -2778,8 +2780,12 @@ class CourseRunEditViewTests(TestCase):
             OrganizationExtension.EDIT_COURSE_RUN, self.organization_extension.group, self.organization_extension
         )
         response = self.client.get(self.edit_page_url)
+        response_content = BeautifulSoup(response.content)
 
-        self.assertContains(response, '<input id="id_lms_course_id" name="lms_course_id" type="hidden"')
+        course_id_attribute = response_content.find(
+            "input", {"id": "id_lms_course_id", "type": "hidden", "name": "lms_course_id"}
+        )
+        assert course_id_attribute is not None
 
         self.assertContains(response, 'Studio URL')
         self.assertContains(response, 'STUDIO URL')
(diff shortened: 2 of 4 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -588,8 +589,11 @@ def post(self, request, *args, **kwargs):
         user = request.user
         parent_course = self.get_parent_course()
 
+        mutable = self.request.POST._mutable
```
clintonb: This is a big code smell. If times need to be parsed, update the forms to do so. Django is wisely not allowing you to touch the original data. Don't work around that.

course_discovery/apps/publisher/views.py
```diff
@@ -731,8 +736,11 @@ def post(self, request, *args, **kwargs):
         course_run = context.get('course_run')
         lms_course_id = course_run.lms_course_id
 
+        mutable = self.request.POST._mutable
```
clintonb: Same comment as above.

clintonb (changes requested):

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -98,11 +98,7 @@ def test_page_loads_only_course_related_runs(self):
         # add some new courses and course runs
         factories.CourseRunFactory.create_batch(2)
         response = self.client.get(reverse('admin_metadata:update_course_runs', args=(self.program.id,)))
-        html = '<input checked="checked" id="id_excluded_course_runs_0" '
-        html += 'name="excluded_course_runs" type="checkbox" value="{id}" />'.format(
-            id=self.excluded_course_run.id
-        )
-        self.assertContains(response, html)
+        self.assertContains(response, 'value="{id}'.format(id=self.excluded_course_run.id))
```
schenedx: Why we need to simplify the assert here? I believe the original intention for the assert is to make sure the check box is included in the HTML. With the change here, we are no longer making sure a checkbox is rendered.

course_discovery/apps/publisher/views.py
```diff
@@ -588,8 +587,6 @@ def post(self, request, *args, **kwargs):
         user = request.user
         parent_course = self.get_parent_course()
 
-        self.request.POST['start'] = parse_datetime_field(self.request.POST.get('start'))
```
schenedx, line 582: If we remove these, won't the data fail the condition needed to format `start` time at  course_discovery/apps/publisher/forms.py line 343?

schenedx (changes requested):
If we are no longer able to check the existence of HTML element, can you figure out and explain here why?

course_discovery/apps/course_metadata/tests/test_admin.py
ahsan-ul-haq: In the new release of Django 1.11 the HTML content does not always render the same way. For example, this HTML could be rendered as following
```
<input checked="checked" id="id_excluded_course_runs_0" name="excluded_course_runs" type="checkbox" value="{id}" />
```
and 
```
<input type="checkbox" checked="checked" name="excluded_course_runs"  id="id_excluded_course_runs_0" value="{id}" />
```
Therefore I decided to check only the ids.

course_discovery/apps/publisher/views.py
ahsan-ul-haq: No, it won't.

ahsan-ul-haq:
@clintonb @schenedx please review.

ahsan-ul-haq:
@clintonb @schenedx please review

course_discovery/apps/course_metadata/tests/test_admin.py
clintonb: At the very least use a regex, BeautifulSoup, or some other mechanism to ensure you have the correct element (e.g. based on ID). As is, this test seems like a potential source of flakiness should the UI ever changed. In fact, if we had done proper HTML parsing when this test was first written, we wouldn't be having this conversation now.

course_discovery/apps/publisher/forms.py
```diff
@@ -340,6 +340,16 @@ def clean_lms_course_id(self):
 
         return None
 
+    def clean_start(self):
```
clintonb: This is not necessary. `DateTimeField` does parsing already. See https://docs.djangoproject.com/en/1.11/ref/forms/fields/#datetimefield.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -177,7 +177,7 @@ def test_create_form_with_multiple_organization(self):
 
         self.assertContains(
             response,
-            '<select class="field-input input-select" id="id_organization" name="organization" required>'
```
clintonb: Same comment as above. Use an HTML parser.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2853,10 +2848,10 @@ def test_studio_instance_with_project_coordinator(self):
 
         self.assertContains(response, self.new_course_run.lms_course_id)
         self.assertContains(
-            response, '<input class="field-input input-text" id="id_lms_course_id" name="lms_course_id"'
+            response, 'id="id_lms_course_id"'
         )
 
-        self.assertNotContains(response, '<a class="studio-link"')
+        self.assertNotContains(response, 'class="studio-link"')
```
clintonb: Use an HTML parse for the entire file.

Unrelated: holy cow! 3300 lines of test code seems a bit much.

requirements/base.txt
```diff
@@ -1,14 +1,14 @@
 beautifulsoup4==4.5.1
 boto==2.42.0
 cryptography==1.7.1
-django==1.10.7
+django==1.11.2
```
clintonb: None of the changes you've made are specific to Django 1.11. They are backward-compatible with Django 1.10. It will be much safer if we do the actual Django upgrade in its own commit that we can safely deploy by itself. See https://openedx.atlassian.net/wiki/display/OpenDev/Django+Upgrade+Checklist.

clintonb (changes requested):

course_discovery/apps/course_metadata/tests/test_admin.py
schenedx: @ahsan-ul-haq I agree what @clintonb is saying. Let's use BeautifulSoup to parse the html elements and asserts base on those.
The changes here would make the testing less rigorous, hence allow the code to be broken undetected.

requirements/base.txt
schenedx: @clintonb I don't know about that tho. If Ahsan upgraded to django 1.11 directly and merged without his changes in here, won't we suffer from flaky tests at least?

clintonb: I'm advocating for merging all of the changes except this one line. Release those changes. Run another release with only this Django upgrade.

schenedx: @clintonb Ah.. Thank you for the explanation. I agree with your proposed plan.

ahsan-ul-haq:
@clintonb made the suggested change

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -98,11 +100,9 @@ def test_page_loads_only_course_related_runs(self):
...
         factories.CourseRunFactory.create_batch(2)
         response = self.client.get(reverse('admin_metadata:update_course_runs', args=(self.program.id,)))
-        html = '<input checked="checked" id="id_excluded_course_runs_0" '
-        html += 'name="excluded_course_runs" type="checkbox" value="{id}" />'.format(
-            id=self.excluded_course_run.id
-        )
-        self.assertContains(response, html)
+        response_content = BeautifulSoup(response.content)
+
+        self.assertIsNotNone(response_content.find("input", {"value": self.excluded_course_run.id}))
```
schenedx: I believe this assert is not good enough. Can you please make sure the input is of type `checkbox` with proper html id?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -174,11 +176,9 @@ def test_create_form_with_multiple_organization(self):
         organization_extension = factories.OrganizationExtensionFactory()
         self.user.groups.add(organization_extension.group)
         response = self.client.get(reverse('publisher:publisher_courses_new'))
+        response_content = BeautifulSoup(response.content)
 
-        self.assertContains(
-            response,
-            '<select class="field-input input-select" id="id_organization" name="organization" required>'
-        )
+        self.assertIsNotNone(response_content.find("select", {"id": "id_organization"}))
```
schenedx: Again, please validate more elements on the HTML element. Like make sure the name is `organization`

schenedx (changes requested):
Please improve the strength of the test validations by doing more asserts and checks on the HTML elements.
rest LGTM

clintonb (approved):
Squash your commits, and correct the commit message and PR title.

schenedx (approved):
👍

rlucioni:
@doctoryes and I glanced at this in the context of a separate conversation and noticed that the version of DRF being installed remains at 3.3.3. 3.3.3 was only [tested up to Django 1.9](https://github.com/encode/django-rest-framework/blob/0c42c742290400e3ee742e77268681b6911604ed/tox.ini#L18). 3.6.3 was the first version to be tested against a stable version of Django 1.11. Notably, DRF doesn't require Django for installation in an attempt to support users running a forked version of Django: https://github.com/encode/django-rest-framework/issues/2390#issuecomment-69193177. That missing `install_requires` directive is how we ended up in this position.

Given that this project's tests pass, it seems like the parts of DRF used here are compatible with 1.11, but we should be aware that we're running DRF alongside a version of Django with which it wasn't tested (and doesn't promise to support).

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -1,6 +1,8 @@
 import itertools
 
 import ddt
+
```
rlucioni: Is isort requiring you to add this newline (and the ones below)? We don't typically separate direct module imports into a separate group like this.

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -98,11 +100,11 @@ def test_page_loads_only_course_related_runs(self):
...
         factories.CourseRunFactory.create_batch(2)
         response = self.client.get(reverse('admin_metadata:update_course_runs', args=(self.program.id,)))
-        html = '<input checked="checked" id="id_excluded_course_runs_0" '
-        html += 'name="excluded_course_runs" type="checkbox" value="{id}" />'.format(
-            id=self.excluded_course_run.id
-        )
-        self.assertContains(response, html)
+        response_content = BeautifulSoup(response.content)
+
+        self.assertIsNotNone(response_content.find(
```
rlucioni: We use pytest in this project. Use regular assertions when you add new tests, here and below.
```py
assert response_content.find(...) != None
```

jibsheet:
has anyone run showmigrations and sqlmigrate and examined what will change in this upgrade?

course_discovery/apps/course_metadata/tests/test_admin.py
ahsan-ul-haq: yes seemingly isort require me to do so.

rlucioni: This isn't a big deal, but could you try pushing up a commit without these newlines? Our version of isort is pinned and it shouldn't suddenly start requiring different (and unwanted) behavior.

clintonb:
You didn't fix your commit message.

ahsan-ul-haq:
@clintonb oh missed that

PR #999, opened 2017-06-19
Use set comprehension instead of list comprehension when collecting C…

douglashall:
…ourse model primary keys.

This is an attempt to fix an issue we are seeing with unreliable search results. Also, we are temporarily adding some additional logging to help diagnose the issue.

ENT-444


clintonb (changes requested):
Have you tested this locally? I fail to see how this fixes anything. Databases don't usually care how many times a clause is met; they will only return one row.

douglashall:
@clintonb Agreed that it shouldn't matter, but technically this should be a set comprehension and not a list comprehension. Do you see any downsides to making this change? If not, I'd like to roll this out to see if it has any effect on the issue.

Unfortunately I am unable to currently reproduce the search flakiness issue locally (waiting on DevOps for a snapshot of PROD discovery elasticsearch).

clintonb:
There are no downsides to deploying, but there are no upsides either. It's a change we know won't fix anything. Why would we exert the resources to deploy it?

douglashall:
@clintonb Sometimes life surprises you when you least expect it. Do you see deploying the discovery service as something that takes a lot of resources?

course_discovery/apps/course_metadata/models.py
```diff
@@ -334,7 +334,14 @@ def search(cls, query):
         """
         query = clean_query(query)
         results = SearchQuerySet().models(cls).raw_search(query)
-        ids = [result.pk for result in results]
+        ids = {result.pk for result in results}
+
+        # TODO: Remove this logging after running in production to collect some samples
+        logger.info('Course search query {query} returned the following ids: {course_ids}'.format(
```
mattdrayer: Rather than remove, maybe we can make it a configuration thing so we can turn on/off when further troubleshooting is needed?

douglashall:
@mattdrayer Thanks for the suggestion. I have added the waffle switch.

mattdrayer:
Thanks!  👍 -- good to go after we've got all the checks passing

clintonb:
Please remember to remove this once your investigation is complete.

douglashall:
@clintonb Will do. As expected, the set comprehension had no effect, but the logging proves that our problem lies with the elasticsearch query. I'm getting back completely different result sets with each query to elasticsearch.

mattdrayer:
Possible explanation: https://github.com/elastic/elasticsearch/issues/11037

PR #1001, opened 2017-06-20
ENT-444 Use single shard for catalog index.

douglashall:
This change will cause newly created catalog indices to be created with a shard size of 1.

The catalog index is small in terms of number of documents (6732) and data
size (87.2mb). There is no reason to prematurely shard the index. In addition,
sharding seems to be affecting our ability to properly score search
results with such a small data set causing inconsistent results when
performing queries which match many documents in the index.

https://aws.amazon.com/blogs/database/get-started-with-amazon-elasticsearch-service-how-many-shards-do-i-need/

Number of shards = 87.2MB / 30GB = 1

ENT-444

### DIFF
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -363,8 +363,15 @@ SWAGGER_SETTINGS = {
 # We are adding the lowercase analyzer and tweaking the ngram analyzers here,
 # so we need to use these settings rather than the index defaults.
 # We are making these changes to enable autocomplete for the typeahead endpoint.
+# In addition we are specifying the number of shards and replicas that indices
+# will be created with as recommended here:
+# https://aws.amazon.com/blogs/database/get-started-with-amazon-elasticsearch-service-how-many-shards-do-i-need/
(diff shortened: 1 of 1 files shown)


clintonb:
Is there evidence to supporting that this resolves the issue?

We know for a fact that course run searches are stable whereas course searches are not. Have you considered updating the code to use the same mechanisms for executing both of these searches?

* [Course](https://github.com/edx/course-discovery/blob/dfe712e38eb78fd04f85aa7f4d3dd8639f7d0cbf/course_discovery/apps/course_metadata/models.py#L326-L338)
* [CourseRun](https://github.com/edx/course-discovery/blob/dfe712e38eb78fd04f85aa7f4d3dd8639f7d0cbf/course_discovery/apps/course_metadata/models.py#L591-L592)

douglashall:
@clintonb I would like to get this onto stage so we can test if it fixes the inconsistency with Course search.

In general though, this seems like a best practice.

clintonb (approved):
Update the comment above the setting explaining this change and linking to the related documentation.

PR #1003, opened 2017-06-21
Call load_all on Course SearchQuerySet.

douglashall:
Force haystack to make a single elasticsearch query to retrieve
Course search results.

ENT-444


douglashall:
FYI @clintonb

PR #1004, opened 2017-06-21
Update the script to add short_description_override

awais786:
Update the import script so that it can fetch the short_description_override.
Property added also to make sure if this value has the priority over the course short description.

https://openedx.atlassian.net/browse/EDUCATOR-704

### DIFF
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -204,6 +204,15 @@ class Course(TimeStampedModel, ChangedByMixin):
 
         return None
 
+    @property
+    def course_short_description(self):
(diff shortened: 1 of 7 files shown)


awais786:
@waheedahmed Please review.

course_discovery/apps/publisher/models.py
```diff
@@ -204,6 +204,15 @@ def course_image_url(self):
 
         return None
 
+    @property
+    def short_description_override(self):
```
waheedahmed: Property name should be only `short_description`.

waheedahmed (approved):
One comment otherwise looks good.

PR #1005, opened 2017-06-21
updated organization user role save method

asadazam93:
Educator-665

When a new user is assigned to a role in organization_user_roles via the publisher admin the courses_user_roles for that organization are also updated accordingly. i.e the new user role is assigned in all the courses for that organization.

### DIFF
--- a/course_discovery/apps/publisher/admin.py
+++ b/course_discovery/apps/publisher/admin.py
@@ -53,10 +53,14 @@ class OrganizationUserRoleAdmin(admin.ModelAdmin):
     def save_model(self, request, obj, form, change):
         obj.save()
         publisher_courses = obj.organization.publisher_courses
+
         courses_without_role = publisher_courses.exclude(course_user_roles__role=obj.role)
+
         CourseUserRole.objects.bulk_create(
             [CourseUserRole(course=course, user=obj.user, role=obj.role) for course in courses_without_role]
         )
+
+        CourseUserRole.objects.filter(course__organizations__in=[obj.organization], role=obj.role).update(user=obj.user)
 
         # Assign user a group according to its role.
         group = Group.objects.get(name=self.role_groups_dict.get(obj.role))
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/admin.py
```diff
@@ -56,9 +56,11 @@ def save_model(self, request, obj, form, change):
         obj.save()
         publisher_courses = obj.organization.publisher_courses
         courses_without_role = publisher_courses.exclude(course_user_roles__role=obj.role)
+        courses_with_role = publisher_courses.filter(course_user_roles__role=obj.role)
         CourseUserRole.objects.bulk_create(
             [CourseUserRole(course=course, user=obj.user, role=obj.role) for course in courses_without_role]
         )
+        CourseUserRole.objects.filter(course__in=courses_with_role).update(user=obj.user)
```
awais786: Add role condition.

course_discovery/apps/publisher/admin.py
```diff
@@ -53,11 +53,20 @@ class OrganizationUserRoleAdmin(admin.ModelAdmin):
...
         courses_without_role = publisher_courses.exclude(course_user_roles__role=obj.role)
+        courses_with_role = publisher_courses.filter(course_user_roles__role=obj.role)
+
         CourseUserRole.objects.bulk_create(
             [CourseUserRole(course=course, user=obj.user, role=obj.role) for course in courses_without_role]
         )
 
+        course_user_roles = CourseUserRole.objects.filter(course__in=courses_with_role, role=obj.role)
+
+        for course_user_role in course_user_roles:
```
waheedahmed: You can achieve this with single line instead of looping on queryset and saving every object individually and also no need for extra query at L58:
```
CourseUserRole.objects.filter(course__organizations__in=[obj.organization], role=obj.role).update(user=obj.user)
```

waheedahmed (changes requested):

waheedahmed (approved):

PR #1006, opened 2017-06-21
Revert "Call load_all on Course SearchQuerySet."

douglashall:
Reverts edx/course-discovery#1003

This change did not resolve ENT-444 so I am rolling it back.


douglashall:
FYI @clintonb

PR #1007, opened 2017-06-21
Increase HAYSTACK_ITERATOR_LOAD_PER_QUERY setting.

douglashall:
This change will cause Haystack to load up to 5000 results with each
elasticsearch query instead of making multiple round trips to collect
all query results.

ENT-444


douglashall:
FYI @clintonb

clintonb (commented):
What's your hypothesis for this change? Why is it necessary? Have actually experimented with it? You can try these out with simple configuration changes in edx/edx-internal.

douglashall:
@clintonb My hypothesis here is that making multiple elasticsearch requests to page through the result set and collect all results is resulting in variable complete result sets.

Also, it seems inefficient to make multiple round trips to elasticsearch when we know we need the complete result set.

Unfortunately, I cannot produce the issue locally, so the only place I can test this hypothesis is on stage. I will roll this change back after I have tested my hypothesis.

PR #1015, opened 2017-06-23
Update Django comments

vkaracic:
In the 1.8 version they have moved from using the `SITE_ID` settings attribute which we want to remove because of multi-tenancy in https://github.com/edx/course-discovery/pull/977

https://github.com/django/django-contrib-comments/pull/108
https://openedx.atlassian.net/browse/LEARNER-1583

@edx/helio please review.


clintonb (approved):
Cleanup your commit message, so that it's less confusing to future viewers. The package name is `django-contrib-comments`, not Django comments. Explain why the update is necessary in the commit message.

PR #1017, opened 2017-06-23
Revert "Update django-contrib-comments"

vkaracic:
This reverts commit e5d26878e169501f846238206524322adc37cd9f.


clintonb:
Why the revert?

vkaracic:
It causes an error when commenting in the publisher service. I'd like some time to explore that service and possibly add a relevant test alongside the upgrade.

clintonb (approved):

PR #1018, opened 2017-06-23
Removed autocomplete for the video field on program admin

clintonb:
The autocomplete does not allow for the video to be deleted. Since autocomplete is overkill for this field, I have removed it and replaced it with Django's built-in search functionality over raw_id_fields.


clintonb:
@edx/learner please review

mjfrey (approved):

schenedx (approved):
👍

PR #1020, opened 2017-06-23
Add warning logs to MarketingSitePublisher 

schenedx:
LEARNER-1602
I am trying to make it easy to understand publishing problem by adding logging around publishing.

Tested on local machine against production marketing site

@rlucioni @MatthewPiatetsky Please review

### DIFF
--- a/course_discovery/apps/course_metadata/publishers.py
+++ b/course_discovery/apps/course_metadata/publishers.py
@@ -102,7 +102,7 @@ class BaseMarketingSitePublisher:
         if response.status_code == 200:
             return response.json()['list'][0]['nid']
         else:
-            raise NodeLookupError
+            raise NodeLookupError({'response_text': response.text, 'response_status': response.status_code})
 
     def create_node(self, node_data):
         """
@@ -124,7 +124,7 @@ class BaseMarketingSitePublisher:
         if response.status_code == 201:
             return response.json()['id']
         else:
-            raise NodeCreateError
+            raise NodeCreateError({'response_text': response.text, 'response_status': response.status_code})
 
     def edit_node(self, node_id, node_data):
         """
@@ -143,7 +143,13 @@ class BaseMarketingSitePublisher:
         response = self.client.api_session.put(node_url, data=node_data)
 
         if response.status_code != 200:
-            raise NodeEditError
+            raise NodeEditError(
+                {
+                    'node_id': node_id,
+                    'response_text': response.text,
+                    'response_status': response.status_code
+                }
+            )
 
     def delete_node(self, node_id):
         """
@@ -160,7 +166,13 @@ class BaseMarketingSitePublisher:
         response = self.client.api_session.delete(node_url)
 
         if response.status_code != 200:
-            raise NodeDeleteError
+            raise NodeDeleteError(
+                {
+                    'node_id': node_id,
+                    'response_text': response.text,
+                    'response_status': response.status_code
+                }
+            )
 
 
 class CourseRunMarketingSitePublisher(BaseMarketingSitePublisher):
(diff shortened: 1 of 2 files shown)


MatthewPiatetsky:
👍 this seems fine to me, but why did you add logs rather than just adding the message to the exception? (the exception still gets logged in splunk, so this way the info will just get logged twice) 
is it for the test?

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -102,6 +105,11 @@ def node_id(self, obj):
         if response.status_code == 200:
             return response.json()['list'][0]['nid']
         else:
+            logger.warning(
```
rlucioni: We're already using exception handlers to catch and log these exceptions in model save and signal receivers. I second @MatthewPiatetsky's suggestion to use this message as the exception message, and log it once where the exceptions are already being logged.

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -128,8 +142,18 @@ def test_create_node(self):
         self.mock_api_client()
         self.mock_node_create(response_data, 500)
 
-        with pytest.raises(NodeCreateError):
-            self.publisher.create_node(node_data)
+        with LogCapture(publisher_logger.name) as log_capture:
```
rlucioni: These checks you're adding here are change-detectors. I would remove them. For more, see https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html.

rlucioni (changes requested):

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -102,6 +105,11 @@ def node_id(self, obj):
...
+                'Get marketing site node failed. HTTP status code [%s], and response text is %s',
+                response.status_code,
+                response.text
```
rlucioni: @schenedx when you were running this locally, did you check to see if this actually contains anything useful? I'm worried it might not end up helping.

MatthewPiatetsky: @rlucioni i checked that response.text has a useful message sometimes

course_discovery/apps/course_metadata/publishers.py
rlucioni: @schenedx to clarify, all you need to do is set these strings as the exception messages. We're already logging them.

schenedx:
@rlucioni Is this what you are suggesting now?

awais786:
@MatthewPiatetsky @schenedx  you can log the actual error in this block also.
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views/people.py#L50

course_discovery/apps/course_metadata/publishers.py
schenedx: @rlucioni In our existing bug case, the error message is `403 Access Denied: CSRF validation failed`

rlucioni (approved):
@schenedx that works. Please remember to squash your commits.

rlucioni:
@awais786 we're working with a different publication process here (i.e., courses and programs).

awais786:
@rlucioni On ticket the logs shows the messages from this block. That's why I suggested that.

`Jun 23 14:14:08 ip-10-2-110-233 [service_variant=discovery][course_discovery.apps.api.v1.views.people] ERROR [ip-10-2-110-233  1566] [/edx/app/discovery/discovery/course_discovery/apps/api/v1/views/people.py:53] - An error occurred while adding the person [?]-[?] to the marketing site.`

schenedx:
This change is already validated on stage. Feel free to deploy to production

PR #1022, opened 2017-06-24
Adding tabular inline data in admin for a course with course-runs and…

awais786:
Adding inline items on admin course detail page.

- Course Runs
- Course UserRoles

https://openedx.atlassian.net/browse/EDUCATOR-739

### DIFF
--- a/course_discovery/apps/publisher/admin.py
+++ b/course_discovery/apps/publisher/admin.py
@@ -80,11 +80,29 @@ class CourseRunStateAdmin(admin.ModelAdmin):
     ordering = ['id']
 
 
+class CourseUserRoleInline(admin.TabularInline):
+    model = CourseUserRole
+    extra = 1
+    raw_id_fields = ('user',)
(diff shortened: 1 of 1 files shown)


awais786:
screen shot 2017-06-24 at 3 20 33 pm

awais786:
@asadazam93 Please review this.

Ayub-Khan (approved):
Looks good to me.

clintonb (changes requested):
What are you trying to accomplish here? Your ticket has zero details. Any Django admin component using the user model is a bad idea. That's a good way to bring down the database.

awais786:
@clintonb but I am using the raw-id for user and also exclude the changed-by.

awais786:
I m closing this PR for time being.

clintonb:
The moment someone clicks the search icon, a window will open that loads the entire user table. That will cause an issue with the database.

PR #1024, opened 2017-06-26
Update django-contrib-comments

vkaracic:
Version 1.8.0 of django-contrib-comments moved away from
using the SITE_ID settings attribute in favor of supporting
multi-tenancy.

LEARNER-1583 + EDUCATOR-733

@edx/helio and @awais786 please review.

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -158,7 +158,7 @@ class CourseRunDetailView(mixins.LoginRequiredMixin, mixins.PublisherPermissionM
         context['object'] = course_run
         context['comment_object'] = course_run
 
-        # this ulr is used for the comments post back redirection.
+        # this URL is used for the comments post back redirection.
         context['post_back_url'] = reverse('publisher:publisher_course_run_detail', kwargs={'pk': course_run.id})
 
         context['can_edit'] = mixins.check_course_organization_permission(
--- a/course_discovery/apps/publisher_comments/api/tests/test_views.py
+++ b/course_discovery/apps/publisher_comments/api/tests/test_views.py
@@ -5,8 +5,38 @@ from rest_framework.reverse import reverse
 
 from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
 from course_discovery.apps.publisher.tests import JSON_CONTENT_TYPE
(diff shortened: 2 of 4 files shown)


clintonb (changes requested):
Test? The last time we merged this, it broke. How do we know the breakage has been addressed?

vkaracic:
@clintonb test added!

course_discovery/apps/publisher_comments/tests/test_forms.py
```diff
@@ -0,0 +1,36 @@
+from django.test import TestCase
+
+from course_discovery.apps.publisher.tests.factories import CourseRunFactory
+from course_discovery.apps.publisher_comments.forms import CommentsForm
+from course_discovery.apps.publisher_comments.models import Comments
+
+
+class CommentsFormTests(TestCase):
+    """Tests for the CommentsForm form."""
```
clintonb: I've always found these class-level comments worthless. Feel free to remove this.

course_discovery/apps/publisher_comments/tests/test_forms.py
```diff
@@ -0,0 +1,36 @@
...
+        data = {
+            'name': 'Tester',
+            'email': 'tester@example.com',
+            'comment': 'Test comment'
+        }
+        data.update(f.initial)
+        return data
+
+    def test_get_comment_model(self):
+        """Retrieves the Comments model."""
```
clintonb: What retrieves the model? This doc string doesn't really help me understand what you're testing or why.

course_discovery/apps/publisher_comments/tests/test_forms.py
```diff
@@ -0,0 +1,36 @@
...
+        model = self.form.get_comment_model()
+        self.assertEqual(model, Comments)
+
+    def test_get_comment_create_data(self):
+        """Sets the right site_id value."""
```
clintonb: Same as above. Help me understand why this test is useful.

vkaracic:
@clintonb I have somewhat addressed your comments. I've created a test on a more suitable place. The error that happens when upgrading this package happens when data is posted to the comment post endpoint hence this test. And I have confirmed that when I remove the `site_id` parameter this new test fails.

clintonb (approved):

PR #1025, opened 2017-06-26
Mark the courseRun.save with signature from refresh_course_metadata 

schenedx:
LEARNER-1606
Do not publish if the course_run save call comes from refresh_course_metadata command

@rlucioni @MatthewPiatetsky Please review

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -141,7 +141,7 @@ class CoursesApiDataLoader(AbstractDataLoader):
 
     def update_course_run(self, course_run, body):
         validated_data = self.format_course_run_data(body)
-        self._update_instance(course_run, validated_data)
+        self._update_instance(course_run, validated_data, suppress_publication=True)
 
         logger.info('Processed course run with UUID [%s].', course_run.uuid)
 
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -604,7 +604,8 @@ def __str__(self):
     def save(self, *args, **kwargs):
         is_publishable = (
             self.course.partner.has_marketing_site and
-            waffle.switch_is_active('publish_course_runs_to_marketing_site')
+            waffle.switch_is_active('publish_course_runs_to_marketing_site') and
+            not kwargs.pop('refresh_course_metadata', False)
```
rlucioni: Is there a need to pop here? I would do it like this:
```
kwargs.get('suppress_publication', False)
```

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -184,7 +184,7 @@ def update_course(self, course, body):
     def _update_instance(self, instance, validated_data):
         for attr, value in validated_data.items():
             setattr(instance, attr, value)
-        instance.save()
+        instance.save(refresh_course_metadata=True)
```
rlucioni: I would try to generalize the name of this kwarg to make your intent and the kwarg's function more clear. For example:
```py
instance.save(suppress_publication=True)
```

rlucioni (approved):

course_discovery/apps/course_metadata/models.py
schenedx: Yes. If we don't do `pop`, the `super(CourseRun, self).save()` call below would fail on validation. I'll leave a comment

PR #1027, opened 2017-06-27
Add cache bust Querystring to get the CSRF token

schenedx:
LEARNER-1602

## Description
Drupal actually expect a different CSRF token whenever we make write level API calls. 
Before this change, we cache and re-use CSRF token.
After this change, we no longer cache the CSRF token, and ensure we get a new CSRF token with every request to Drupal

## Tested
Manually tested from local discovery environment connecting to the stage environment. 
Was able to reproduce the original error on stage and saw this change fixes the problem

Please review @mikedikan @rlucioni

### DIFF
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -1,3 +1,6 @@
+import random
+import string
+
 import requests
 from django.utils.functional import cached_property
 from stdimage.models import StdImageFieldFile
@@ -100,20 +103,32 @@ class MarketingSiteAPIClient(object):
         # This is not a RESTful API so checking the status code is not enough
         # We also check that we were redirected to the admin page
         if not (response.status_code == 200 and response.url == admin_url):
-            raise MarketingSiteAPIClientException('Marketing Site Login failed!')
+            raise MarketingSiteAPIClientException(
+                {
+                    'message': 'Marketing Site Login failed!',
+                    'status': response.status_code,
+                    'url': response.url
+                }
+            )
         return session
 
-    @cached_property
+    @property
     def api_session(self):
         self.init_session.headers.update(self.headers)
         return self.init_session
 
-    @cached_property
+    @property
     def csrf_token(self):
-        token_url = '{root}/restws/session/token'.format(root=self.api_url)
+        # We need to make sure we can bypass the Varnish cache.
+        # So adding a random salt into the query string to cache bust
+        random_qs = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
+        token_url = '{root}/restws/session/token?cachebust={qs}'.format(root=self.api_url, qs=random_qs)
         response = self.init_session.get(token_url)
         if not response.status_code == 200:
-            raise MarketingSiteAPIClientException('Failed to retrieve Marketing Site CSRF token!')
+            raise MarketingSiteAPIClientException({
+                'message': 'Failed to retrieve Marketing Site CSRF token!',
+                'status': response.status_code,
+            })
         token = response.content.decode('utf8')
         return token
 
@@ -127,7 +142,7 @@ class MarketingSiteAPIClient(object):
         user_id = response.json()['list'][0]['uid']
         return user_id
 
-    @cached_property
+    @property
     def headers(self):
         return {
             'Content-Type': 'application/json',


mikedikan (approved):

course_discovery/apps/course_metadata/utils.py
```diff
@@ -100,20 +103,32 @@ def init_session(self):
...
+            raise MarketingSiteAPIClientException(
+                {
+                    'message': 'Marketing Site Login failed!',
+                    'status': response.status_code,
+                    'url': response.url
+                }
+            )
         return session
 
-    @cached_property
```
rlucioni, line 106: @schenedx what is getting rid of this caching going to do to the data loader runtime?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -100,20 +103,32 @@ def init_session(self):
...
         self.init_session.headers.update(self.headers)
         return self.init_session
 
-    @cached_property
+    @property
     def csrf_token(self):
-        token_url = '{root}/restws/session/token'.format(root=self.api_url)
+        # We need to make sure we can bypass the cloud front cache.
+        # So adding a random salt into the query string to cache bust
+        random_qs = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
```
rlucioni: @schenedx what changed to necessitate this? We haven't had to do this in the past.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -100,20 +103,32 @@ def init_session(self):
...
+    @property
     def api_session(self):
         self.init_session.headers.update(self.headers)
         return self.init_session
 
-    @cached_property
+    @property
     def csrf_token(self):
-        token_url = '{root}/restws/session/token'.format(root=self.api_url)
+        # We need to make sure we can bypass the cloud front cache.
```
rlucioni: We don't use CloudFront on the marketing site anymore. Separately, my understanding is that the querystring is used to bypass Varnish caching.

course_discovery/apps/course_metadata/utils.py
schenedx: Getting rid of this cache will have no impact to the data loader runtime. 
Data loader do not call marketing site publisher

rlucioni (commented):
@schenedx I'd like to make sure we understand why this change is necessary (i.e., why is this necessary now, when it wasn't before), and what impact (if any) it will have on the data loaders.

course_discovery/apps/course_metadata/utils.py
schenedx: I don't know what changed in the Drupal's world. Since last Friday, we started to experience this problem in Drupal. The only event co-incided with this bug is the fact that Acquia is doing "unscheduled maintenance" in our platform. The maintenance is still ongoing.
I think cache busting to get the new CSRF token is the right way to fix this.

course_discovery/apps/course_metadata/utils.py
rlucioni: @schenedx this API client is also used by the data loaders.

course_discovery/apps/course_metadata/utils.py
schenedx: I can update the Comment to say `bypass the Varnish cache`

course_discovery/apps/course_metadata/utils.py
rlucioni: @schenedx this isn't the right way to fix this. It's a hack. The right way to fix this starts with understanding what changed. (It's most likely unrelated, but we did change CDN providers last Wednesday. It's possible that publishing wasn't attempted until Friday.)

schenedx:
@rlucioni I don' know what really changed on the drupal side. I did investigation and couldn't see any obvious triggers from past releases. See https://github.com/edx/edx-mktg/compare/release-2017-06-21-16.20...release-2017-06-26-20.45
That said, I believe it could be the OS upgrade Acquia is doing on our marketing site platform that may be the trigger.

course_discovery/apps/course_metadata/utils.py
schenedx: @rlucioni Publishing is definitely attempted before Friday. I dealt with a bug related to this already.
The CDN change maybe a possible cause for this issue, but what would be the fix about the CDN change? Shouldn't we update our code to conform the new environment?

schenedx:
@rlucioni You have a point on it's impact to data loaders. I can update the solution to only update the CSRF token during the PUT/POST/DELETE api calls

course_discovery/apps/course_metadata/utils.py
rlucioni: @schenedx like I said, the CDN change is most likely unrelated, it's just something I know about that changed.

rlucioni:
@schenedx have you tried timing the data loaders before and after this change? The marketing site data loaders cache an instance of this client, so you may not have to handle HTTP methods in different ways.

I'm reacting strongly here because it doesn't look like impact on the data loaders was considered. If they're unaffected then I don't have grounds on which to object to this, but it does make me sad that we need to do this.

schenedx:
To follow up on the offline conversation, here is what we are going to do:

1. Merge this PR and test the dataloader impact on stage. We suspect there shouldn't be any data loader impact because of the cached session at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/data_loaders/marketing_site.py#L39
2. Talk to devops about cloudflare setup, and see if there are configurations that are different from cloudflare and cloud front. Leave the ticket open but lower it's priority.

schenedx:
@rlucioni I did a stage data loader run with this change deployed. The data loader run finished within 13 min, which is the norm compared to other runs.
I believe this change had no effect on data loader performance

PR #1029, opened 2017-06-28
Updated Django Waffle to 0.12.0

clintonb:


clintonb:
FYI @edx/learner

schenedx:
@clintonb Looks like some management command teardown is broken?

nedbat:
If this is important for Ginkgo, it will have to be cherry-picked.

nedbat:
What's the status of this?

clintonb:
@nedbat I doubt this will land anytime soon. @schenedx does your team have capacity to resolve this issue?

schenedx:
@clintonb Not at the moment. And frankly, I don't even remember the context here. Can I get some in this PR?

clintonb:
The Django 1.11 upgraded was pushed out with a version of Django Waffle that is not fully compatible with Django 1.11—management commands don't work. This PR upgrades Django Waffle; however, there is an issue with a test resulting in MySQL going away.

nedbat:
This PR is mentioned on the Ginkgo wiki page. I'll mark it there as deferred. Thanks.

PR #1037, opened 2017-07-06
optimize queries for courses list page

muhammad-ammar:
[**EDUCATOR-828**](https://openedx.atlassian.net/browse/EDUCATOR-828)

These changes will improve the publisher courses list page(`/publisher/courses`) load time. There is still room for improvement but for now this good enough.

-----------
**Publisher courses page load time with queries executed for `1242` courses** 

Before Optimisation
——————————
1371.67 ms (8875 queries including 8869 duplicates )

After Optimisation
—————————
561.99 ms (3912 queries including 3901 duplicates )

-----------

**Unit test stats shows queries executed before and after the optimization**

|               test                                            | before    |  after|
----------------------------------------|----------|-------|
|test_page_with_enable_waffle_switch    |     13         |   15|
|test_page_with_disable_waffle_switch   |      21        |    21|
|test_courses_with_permission               |    100         |  64|
|test_courses_with_course_user_role      |    60          |  32|
|test_courses_with_admin                        |     58         |   31 |

### DIFF
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -118,6 +118,19 @@ class Course(TimeStampedModel, ChangedByMixin):
         return user_emails
 
     @property
+    def organization_name(self):
+        """
+        Returns organization name for a course.
+        """
+        organization_name = ''
+        try:
+            organization_name = self.organizations.only('key').first().key
+        except AttributeError:
+            pass
+
+        return organization_name
+
+    @property
     def keywords_data(self):
         keywords = self.keywords.all()
         if keywords:
@@ -131,7 +144,7 @@ class Course(TimeStampedModel, ChangedByMixin):
         Return course project coordinator user.
         """
         try:
-            return self.course_user_roles.get(role=PublisherUserRole.ProjectCoordinator).user
+            return self.course_user_roles.only('user').get(role=PublisherUserRole.ProjectCoordinator).user
         except CourseUserRole.DoesNotExist:
             return None
 
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -1300,13 +1300,17 @@ class DashboardTests(TestCase):
         """ Verify that user from un-authorize group can access only that group courses. """
         self.client.logout()
         self.client.login(username=UserFactory(), password=USER_PASSWORD)
-        response = self.assert_dashboard_response(studio_count=0, published_count=0, progress_count=0, preview_count=0)
+        response = self.assert_dashboard_response(
+            studio_count=0, published_count=0, progress_count=0, preview_count=0, queries_executed=11
+        )
         self._assert_tabs_with_roles(response)
 
     @ddt.data('progress', 'preview', 'studio', 'published')
     def test_with_internal_group(self, tab):
         """ Verify that internal user can see courses assigned to the groups. """
-        response = self.assert_dashboard_response(studio_count=2, published_count=1, progress_count=2, preview_count=1)
+        response = self.assert_dashboard_response(
+            studio_count=2, published_count=1, progress_count=2, preview_count=1, queries_executed=23
+        )
         self.assertContains(response, '<li role="tab" id="tab-{tab}" class="tab"'.format(tab=tab))
 
     def test_with_permissions(self):
@@ -1319,7 +1323,9 @@ class DashboardTests(TestCase):
         assign_perm(OrganizationExtension.VIEW_COURSE, self.organization_extension.group, self.organization_extension)
         self.course_run_1.course.organizations.add(self.organization_extension.organization)
 
-        response = self.assert_dashboard_response(studio_count=0, published_count=0, progress_count=0, preview_count=0)
+        response = self.assert_dashboard_response(
+            studio_count=0, published_count=0, progress_count=0, preview_count=0, queries_executed=11
+        )
         self._assert_tabs_with_roles(response)
 
     def test_with_permissions_with_data(self):
@@ -1342,18 +1348,24 @@ class DashboardTests(TestCase):
             OrganizationExtension.VIEW_COURSE, self.organization_extension.group, self.organization_extension
         )
 
-        response = self.assert_dashboard_response(studio_count=0, published_count=0, progress_count=2, preview_count=1)
+        response = self.assert_dashboard_response(
+            studio_count=0, published_count=0, progress_count=2, preview_count=1, queries_executed=21
+        )
         self._assert_tabs_with_roles(response)
 
     def test_studio_request_course_runs_as_pc(self):
         """ Verify that PC user can see only those courses on which he is assigned as PC role. """
-        response = self.assert_dashboard_response(studio_count=2, published_count=1, progress_count=2, preview_count=1)
+        response = self.assert_dashboard_response(
+            studio_count=2, published_count=1, progress_count=2, preview_count=1, queries_executed=23
+        )
         self._assert_tabs_with_roles(response)
 
     def test_studio_request_course_runs_without_pc_group(self):
         """ Verify that PC user can see only those courses on which he is assigned as PC role. """
         self.user1.groups.remove(self.group_project_coordinator)
-        response = self.assert_dashboard_response(studio_count=0, published_count=1, progress_count=2, preview_count=1)
+        response = self.assert_dashboard_response(
+            studio_count=0, published_count=1, progress_count=2, preview_count=1, queries_executed=20
+        )
         self._assert_tabs_with_roles(response)
 
     def test_without_studio_request_course_runs(self):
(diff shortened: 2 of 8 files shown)


course_discovery/apps/publisher/models.py
```diff
@@ -118,6 +118,29 @@ def get_course_users_emails(self):
...
+        """
+        Return course project coordinator user name.
+
+        Note: This method is specifically written for views.py:CourseListView
+              to reduce the number of queries executed.
+        """
+        if self.user_role == PublisherUserRole.ProjectCoordinator:
+            return self.course_user_name
+
+        return ''
```
muzaffaryousaf: Can we have single `return` here.
```
course_user_name = ''
if condition:
   course_user_name = 'blah'
return course_user_name
```

course_discovery/apps/publisher/models.py
```diff
@@ -118,6 +118,29 @@ def get_course_users_emails(self):
...
+        except AttributeError:
+            return ''
+
+    @property
+    def project_coordinator_name(self):
+        """
+        Return course project coordinator user name.
+
+        Note: This method is specifically written for views.py:CourseListView
+              to reduce the number of queries executed.
```
muzaffaryousaf: Is it specific to that view only, if yes then i think we don't put this as property here.

course_discovery/apps/publisher/models.py
```diff
@@ -118,6 +118,29 @@ def get_course_users_emails(self):
...
 
     @property
+    def organization_name(self):
+        """
+        Returns organization name for a course.
+        """
+        try:
+            return self.organizations.only('key').first().key
+        except AttributeError:
+            return ''
```
muzaffaryousaf: Same, let's try to have single return.

course_discovery/apps/publisher/models.py
```diff
@@ -118,6 +118,29 @@ def get_course_users_emails(self):
...
+
+    @property
+    def project_coordinator_name(self):
+        """
+        Return course project coordinator user name.
```
muzaffaryousaf: NIT: Return --> Returns

course_discovery/apps/publisher/models.py
```diff
@@ -131,7 +154,7 @@ def project_coordinator(self):
         Return course project coordinator user.
         """
         try:
-            return self.course_user_roles.get(role=PublisherUserRole.ProjectCoordinator).user
+            return self.course_user_roles.only('user').get(role=PublisherUserRole.ProjectCoordinator).user
```
muzaffaryousaf, line 147: Does this make any difference in query time ? Have you tried personally ?

course_discovery/apps/publisher/views.py
```diff
@@ -797,10 +798,18 @@ class CourseListView(mixins.LoginRequiredMixin, ListView):
 
     def get_queryset(self):
         user = self.request.user
+        courses = Course.objects.select_related().prefetch_related(
+            'organizations', 'course_state', 'publisher_course_runs', 'course_user_roles'
+        ).annotate(
+            course_user_name=F('course_user_roles__user__full_name')
+        ).annotate(
+            user_role=F('course_user_roles__role')
+        )
```
muzaffaryousaf: Can we add test coverage for query counts ?

muzaffaryousaf: I'm assuming that existing tests are already covering this logic in terms of data set. Can you confirm if that's the case.

course_discovery/apps/publisher/views.py
```diff
@@ -797,10 +798,18 @@ class CourseListView(mixins.LoginRequiredMixin, ListView):
 
     def get_queryset(self):
         user = self.request.user
+        courses = Course.objects.select_related().prefetch_related(
```
muzaffaryousaf: Nice Shot for using `prefetch_related` 💯

course_discovery/apps/publisher/views.py
```diff
@@ -797,10 +798,18 @@ class CourseListView(mixins.LoginRequiredMixin, ListView):
...
+            'organizations', 'course_state', 'publisher_course_runs', 'course_user_roles'
+        ).annotate(
+            course_user_name=F('course_user_roles__user__full_name')
```
muzaffaryousaf: Why do we need these annotates and how much they're adding to the performance improvement.

muzaffaryousaf (changes requested):
@muhammad-ammar Done with first pass, let me know when you're done.

course_discovery/apps/publisher/views.py
muzaffaryousaf: `There may be some situations where you wish to call select_related() with a lot of related objects, or where you don’t know all of the relations. In these cases it is possible to call select_related() with no arguments. This will follow all non-null foreign keys it can find - nullable foreign keys must be specified. This is not recommended in most cases as it is likely to make the underlying query more complex, and return more data, than is actually needed.`

@muhammad-ammar Hope you've seen this from Django docs.

course_discovery/apps/publisher/models.py
muhammad-ammar: yes, I have verified this. we are only selecting a single field instead of all table columns.

course_discovery/apps/publisher/views.py
muhammad-ammar: I have removed this.

course_discovery/apps/publisher/views.py
muhammad-ammar: This will have a huge impact. Currently for each course separate database hits are made to calculate the name of project coordinator. So after annotation no database query will be made in the loop.

muhammad-ammar:
@awais786 @waheedahmed Can anyone of you review the changes?

course_discovery/apps/publisher/models.py
```diff
@@ -118,6 +118,33 @@ def get_course_users_emails(self):
...
+    @property
+    def project_coordinator_name(self):
+        """
+        Returns course project coordinator user name.
+
+        Note: This method is specifically written for views.py:CourseListView
+              to reduce the number of queries executed.
+        """
+        course_user_name = ''
+        if self.user_role == PublisherUserRole.ProjectCoordinator:
```
awais786: Make sure course has the role otherwise it will throw the exception.
Also there is existing property for [PC](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/models.py#L129)

muhammad-ammar:
@awais786 Can you have a quick look? It would be great if we can merge this today.

muhammad-ammar:
@muzaffaryousaf Can you have a final look?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1555,43 +1555,46 @@ class CourseListViewTests(TestCase):
 
     def setUp(self):
         super(CourseListViewTests, self).setUp()
-        self.course = factories.CourseFactory()
+        self.courses = [factories.CourseFactory() for _ in range(10)]
```
muzaffaryousaf, line 1600: Ideally, we should import the `CourseFactory` in first place instead of using `factories`.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1601,11 +1604,21 @@ def test_page_with_enable_waffle_switch(self):
...
-        toggle_switch('publisher_hide_features_for_pilot', True)
+        factories.CourseUserRoleFactory(course=self.course, user=self.user, role=PublisherUserRole.CourseTeam)
+        organization_extension = factories.OrganizationExtensionFactory()
+        self.course.organizations.add(organization_extension.organization)
+        self.user.groups.add(organization_extension.group)
+        assign_perm(OrganizationExtension.VIEW_COURSE, organization_extension.group, organization_extension)
+        assign_perm(OrganizationExtension.EDIT_COURSE, organization_extension.group, organization_extension)
         response = self.client.get(self.courses_url)
-        self.assertNotIn(response.content.decode('UTF-8'), 'Edit')
+        self.assertContains(response, 'Edit')
```
muzaffaryousaf, line 1656: Earlier we were not expecting 'Edit' in response but now we do. Reason ?

muzaffaryousaf: Is it not too much broad assertion ?

muzaffaryousaf: I see you're check the assertion later.

course_discovery/apps/publisher/views.py
```diff
@@ -797,10 +797,14 @@ class CourseListView(mixins.LoginRequiredMixin, ListView):
 
     def get_queryset(self):
         user = self.request.user
+        courses = Course.objects.prefetch_related(
+            'organizations', 'course_state', 'publisher_course_runs', 'course_user_roles'
+        )
```
muzaffaryousaf: can we add `objects.all()`

muzaffaryousaf (approved):
LGTM 👍  except couple of NITs. 💯

awais786:
👍 Good Job @muhammad-ammar.

course_discovery/apps/publisher/tests/test_views.py
muhammad-ammar: yes, but this what is happening now but we can refactor it later in a separate PR

course_discovery/apps/publisher/tests/test_views.py
muhammad-ammar: It is but i will prefer to update the test in a separate PR.

michaelroytman:
Hello,

I saw that this commit went to stage this morning. I got a bunch of pages this morning for stage discovery.

You can see the NewRelic alerts below.

https://rpm.newrelic.com/accounts/88178/incidents/30400026

https://rpm.newrelic.com/accounts/88178/applications/15120497/traced_errors/8a5f2e9e-68a4-11e7-8968-0242ac11000f_16360_21050/similar_errors?original_error_id=8a5f2e9e-68a4-11e7-8968-0242ac11000f_16360_21050

Could you take a look at this before it goes to production? Please coordinate with the Learner - Spartan team, if need be.

Best,
Michael

muzaffaryousaf:
@MichaelRoytman I've verified from New Relic https://rpm.newrelic.com/accounts/88178/applications/15120497/traced_errors/8a5f2e9e-68a4-11e7-8968-0242ac11000f_16360_21050?original_error_id=8a5f2e9e-68a4-11e7-8968-0242ac11000f_16360_21050
This PR did't changed anything related to this error, something else must be. Looks like something is wrong with Publisher's data.

muzaffaryousaf:
@sstack22 This is live on prod, you can check the improvements in PR header.

PR #1038, opened 2017-07-06
Changed the family name default to be an empty string.

afzaledx:
LEARNER-1701


afzaledx:
@edx/helio

mjfrey (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -222,7 +222,7 @@ def __str__(self):
 
     @property
     def full_name(self):
-        return ' '.join((self.given_name, self.family_name,))
+        return ' '.join((self.given_name, self.family_name,)) if self.family_name else self.given_name
```
clintonb: Don't do this. Set the default value of `family_name` to an empty string and make the field non-nullable.

clintonb (changes requested):

clintonb:
See the note on commit messages at https://openedx.atlassian.net/wiki/display/ECOM/Developer+Onboarding. Shorten your commit summaries.

afzaledx:
@clintonb , can you please review again?

clintonb:
Let's discuss this on the ticket. Fixing this correctly will involve a bit more work, and I'm not convinced this is a real issue we have with our current data.

clintonb:
This is not a priority. We'll get to it...eventually.

PR #1043, opened 2017-07-11
extras=reST doesn't work, so force docutils to get installed.

nedbat:
I think the extra_requires spec in django-rest-swagger simply doesn't work (https://github.com/pypa/pip/issues/4617).  Force docutils to get installed.  This can all be removed when we upgrade to 2.x of django-rest-swagger.


nedbat:
@clintonb @rlucioni I think this is the right (temporary) fix.

clintonb (approved):

rlucioni (approved):

PR #1046, opened 2017-07-12
Removed inlines for FAQ and endorsements on Program admin

clintonb:
Using the default UI for SortedManyToManyField since it properly exposes the sorting capability.

LEARNER-1863


schenedx:
@clintonb Can you attach a screenshot of the program editor page after your change?

clintonb:
@schenedx see https://github.com/gregmuellegger/django-sortedm2m#admin

schenedx (approved):
👍

PR #1049, opened 2017-07-12
Upgrade to DRF 3.4.7

rlucioni:
DRF 3.4.7 is the same version run by edx/credentials. Release notes are at http://www.django-rest-framework.org/topics/release-notes/#34x-series. Highlights include:

1. [Support for Django 1.10](https://github.com/encode/django-rest-framework/pull/4158) - sigh
2. [Support for schema generation](https://github.com/encode/django-rest-framework/pull/4179) - required for newer versions of django-rest-swagger
3. A change that [prevents paginated views from re-running queries](https://github.com/encode/django-rest-framework/issues/4201) when count queries return 0 - this explains the expected query count changes made in tests

LEARNER-1590

@edx/learner @nedbat FYI. I will be following this with a PR similar to https://github.com/edx/credentials/pull/281 to fix API docs broken by the Django 1.11 upgrade.


schenedx (approved):
👍

clintonb (approved):

PR #1050, opened 2017-07-13
Fix API docs

rlucioni:
The API docs were broken by the Django 1.11 upgrade. This change includes an upgrade to the most recent version of django-rest-swagger that supports DRF 3.4. Newer versions of django-rest-swagger require DRF 3.5.

This change is complicated by a decision made by the maintainers of the django-filter package to avoid subclassing DRF's DjangoFilterBackend. For more on that decision, see https://github.com/carltongibson/django-filter/pull/576. The SchemaGenerator from DRF 3.4.7 expects instances of DjangoFilterBackend to have a get_fields() method. The DjangoFilterBackend from django-filter 1.0.4 doesn't have this method, instead replacing it with the get_schema_fields() used by the SchemaGenerator in DRF 3.5. To work around this, I've replaced our one use django-filter's DjangoFilterBackend with DRF's DjangoFilterBackend. Note that DRF 3.5 removes its implementation of DjangoFilterBackend in favor of django-filter's, so we will have to change how we import DjangoFilterBackend when we upgrade to 3.5.

LEARNER-1590

@edx/learner @nedbat FYI.

### DIFF
--- a/course_discovery/apps/core/tests/test_throttles.py
+++ b/course_discovery/apps/core/tests/test_throttles.py
@@ -1,9 +1,10 @@
+from django.conf import settings
 from django.core.cache import cache
 from django.urls import reverse
 from rest_framework.test import APITestCase
 
 from course_discovery.apps.core.models import UserThrottleRate
-from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, PartnerFactory, UserFactory
 from course_discovery.apps.core.throttles import OverridableUserRateThrottle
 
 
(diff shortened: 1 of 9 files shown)


requirements/base.txt
```diff
@@ -25,11 +25,7 @@ djangorestframework==3.4.7
 djangorestframework-csv==1.4.1
 djangorestframework-jwt==1.8.0
 djangorestframework-xml==1.3.0
-django-rest-swagger[reST]==0.3.10
-# django-rest-swagger[reST] should install docutils, but doesn't
-# (see https://github.com/pypa/pip/issues/4617), so we force it here.
-docutils>=0.8
```
rlucioni, line 31: @nedbat I've removed this forced install because django-rest-swagger==2.0.7 doesn't require docutils anymore.

course_discovery/apps/core/tests/test_throttles.py
```diff
@@ -14,7 +15,10 @@ class RateLimitingTest(APITestCase):
 
     def setUp(self):
         super(RateLimitingTest, self).setUp()
-        self.url = reverse('django.swagger.resources.view')
+
+        PartnerFactory(pk=settings.DEFAULT_PARTNER_ID)
```
clintonb, line 19: This setting will/should go away when multi-tenancy is completed.

clintonb (approved):

PR #1058, opened 2017-07-17
Made the course run length field Required

asadazam93:
[EDUCATOR-881](https://openedx.atlassian.net/browse/EDUCATOR-881)
**Field update**
Model: Course Run
Field name: length
Current: Optional field
Changed to: Required field


noraiz-anwar (approved):
👍 
Once all tests are passed.

awaisdar001:
@asadazam93  What would happen to existing course runs that already have been sent for a review?

asadazam93:
@awaisdar001 there are no course runs sent for review with this field missing so it would not create any problem. However we have a 190 course runs in the 'published' state missing this required field so we would have some inconsistent data in the database as there would be published course runs missing a required field. This has been discussed on the ticket and they are happy to go forward with this field change anyway.

awaisdar001:
Great, Thanks for the confirmation.

awaisdar001 (approved):

PR #1059, opened 2017-07-17
Filter API data by partner

vkaracic:
Re-adding https://github.com/edx/course-discovery/pull/1042

LEARNER-1119 + LEARNER-1909

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -1,6 +1,5 @@
 import logging
 
-from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db.models import QuerySet
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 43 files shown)


vkaracic:
@clintonb before re-merging this we need to discuss edx/devstack implications. There needs to be a default site created when provisioning a new docker devstack or else 'Site missing' errors will occur. The problem is that the site domain needs to be either `localhost` or `localhost:18381` in order to open the discovery site properly, but then commands like `cache_programs` on the LMS won't work because they access the discovery service through `edx.devstack.discovery` domain. 
That command would give a 500 error (missing Site on discovery side). Adding a new site with `edx.devstack.discovery` domain fixed the error, but I wouldn't get any program back because all of them were created with the partner who's connected to the `localhost` site.

PR #1061, opened 2017-07-17
Add default site id.

vkaracic:
We are adding a default site ID setting via
the edx-django-sites-extensions package because
requests from outside of saved sites' domains,
like the AWS health check, would return an error.


mjfrey (approved):

douglashall (approved):

vkaracic:
Added as part of https://github.com/edx/course-discovery/pull/1065

PR #1062, opened 2017-07-17
LEARNER-1925 add the PIL setting to allow image of smaller size

schenedx:
[LEARNER-1925](https://openedx.atlassian.net/browse/LEARNER-1925)

We want to allow banner image of smaller size than 1440X480.
Currently we suffer from the "IOError: image file is truncated" error when adding new programs or updating existing program with a banner image smaller than that size.
All new programs are having banner images at 1440X260. This change would allow this banner image size to be saved with the program. See https://github.com/python-pillow/Pillow/blob/master/PIL/ImageFile.py#L227 and 

@rlucioni @clintonb please review

### DIFF
--- a/course_discovery/apps/course_metadata/apps.py
+++ b/course_discovery/apps/course_metadata/apps.py
@@ -1,4 +1,5 @@
 from django.apps import AppConfig
+from PIL import ImageFile
 
 
 class CourseMetadataConfig(AppConfig):
@@ -7,5 +8,9 @@ class CourseMetadataConfig(AppConfig):
 
     def ready(self):
         super(CourseMetadataConfig, self).ready()
+        # We need to add this setting because, since MM programs we are accepting banner
+        # images with height less than 480 max. In order to accept those images, we need
+        # to allow PIL to work with these images correctly by setting this variable true
+        ImageFile.LOAD_TRUNCATED_IMAGES = True
         # noinspection PyUnresolvedReferences
         import course_discovery.apps.course_metadata.signals  # pylint: disable=unused-variable


course_discovery/apps/course_metadata/apps.py
```diff
@@ -9,3 +9,5 @@ def ready(self):
         super(CourseMetadataConfig, self).ready()
         # noinspection PyUnresolvedReferences
         import course_discovery.apps.course_metadata.signals  # pylint: disable=unused-variable
+        from PIL import ImageFile
```
clintonb: Why is the import statement here?

clintonb (commented):
How will loading images of this new size affect the cropped versions of the image?

course_discovery/apps/course_metadata/apps.py
schenedx: It doesn't. I have moved to the top of the file. Thanks for the reminder.

rlucioni:
@schenedx I'm concerned that this setting isn't the best way to solve the problem you're dealing with. If we want to change the size of program banner images, why aren't you changing the required variations on the [banner_image](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L761-L772) field? More generally, how do the places that use the "large" 1440x480 version of the image handle a 1440x260 image? It looks like this problem could be avoided short-term by just uploading a 1440x480 image.

rlucioni:
@schenedx and I caught up about this offline. Turns out 1440x260 images have been uploaded for months at this point. @clintonb the cropped versions of the image come out smaller (shorter) than advertised by the API. For example, the "small" 435x145 version is actually 435x75. @schenedx has a ticket for fixing this.

rlucioni (approved):

clintonb (approved):
How were improperly-sized images uploaded?

rlucioni:
@clintonb seemingly through the admin, from what @schenedx showed me. I'm surprised that the size of the image uploaded wasn't validated (i.e., that an image smaller than the largest required variant was allowed), but I haven't dug in to figure out why.

schenedx:
@clintonb Just like what @rlucioni said. We have been allowing uploading of image height smaller than 480px in discovery django admin for a while since MM programs. I found out about these since the creation of LEARNER-501. Tatiana told me that MM and Prof Cert programs all have the banner image resolution at 1440X260. It's only since July 12th, she started to experience the errors documented in LEARNER-1925. I tried to figure out what changed that would cause the error, but couldn't pin point a reason. Figured I'd first fix this problem for her.

PR #1065, opened 2017-07-19
[LEARNER-1119] Filter API data by partner

vkaracic:


course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -61,6 +61,8 @@ def test_create_with_authentication(self):
 
     def test_create_without_drupal_client_settings(self):
         """ Verify that if credentials are missing api will return the error. """
+        self.partner.marketing_site_api_username = None
+        self.partner.save()
```
vkaracic: @clintonb the test will now (probably) fail. But if you look at the change in the commit before the last one, that one passes.

vkaracic: The problem here is that:

1. `self.partner` defined in [SiteMixin](https://github.com/edx/course-discovery/pull/1065/files#diff-a878a57d76623de3c005fb11a0d961ceR15) has a random value for `marketing_site_api_username` attribute.
2. This part here does remove that value (I've printed out all the attribute after the .save() and it did in fact remove the value)
3. The partner in the request [here](https://github.com/edx/course-discovery/pull/1065/files#diff-4884aa315e28a2d72645182fc6b11a31R32) has the value, even though it has the same ID as the one that has been changed not to have a value for `marketing_site_api_username`

I don't understand why is that happening.

vkaracic: Specifically this part [here](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/utils.py#L84) needs to fail in order for the test to pass.

vkaracic:
@edx/helio please review.

mjfrey (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -847,9 +847,9 @@ def prefetch_queryset(cls):
...
             # model iterate across (e.g. language). These fields aren't prefetched by the minimal Course serializer.
-            Prefetch('courses', queryset=CourseSerializer.prefetch_queryset()),
-            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
-            Prefetch('credit_backing_organizations', queryset=OrganizationSerializer.prefetch_queryset()),
+            Prefetch('courses', queryset=CourseSerializer.prefetch_queryset(partner=partner)),
+            Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('credit_backing_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
             Prefetch('corporate_endorsements', queryset=CorporateEndorsementSerializer.prefetch_queryset()),
             Prefetch('individual_endorsements', queryset=EndorsementSerializer.prefetch_queryset()),
         )
```
marjev, line 855: Why is the partner optional parameter in some `prefetch_related` methods, while being required in others?

Is filtering optional or required?

vkaracic: Can't really remember anymore. But I've addressed your comment. Please take another look.

marjev (approved):

PR #1069, opened 2017-07-21
Serialize additional fields for Affiliate Window

rlucioni:
These fields aren't required, but give additional data to affiliates using the API.

LEARNER-1700

@edx/learner-growth FYI

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
@@ -100,24 +100,24 @@ class AffiliateWindowViewSetTests(ElasticsearchTestMixin, SerializationMixin, AP
 
     def assert_product_xml(self, content, seat):
         """ Helper method to verify product data in xml format. """
-        self.assertEqual(content.find('pid').text, '{}-{}'.format(self.course_run.key, seat.type))
-        self.assertEqual(content.find('name').text, self.course_run.title)
-        self.assertEqual(content.find('desc').text, self.course_run.short_description)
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
```diff
@@ -100,20 +101,21 @@ def test_with_closed_enrollment(self):
...
+        assert content.find('currency').text == seat.currency.code
+        assert content.find('category').text == AffiliateWindowSerializer.CATEGORY
+
+    @pytest.mark.skip(reason='AWIN has no current DTD file that we can use to validate our XML.')
     def test_dtd_with_valid_data(self):
         """ Verify the XML data produced by the endpoint conforms to the DTD file. """
         response = self.client.get(self.affiliate_url)
         self.assertEqual(response.status_code, 200)
-        filename = abspath(join(dirname(dirname(__file__)), 'affiliate_window_product_feed.1.4.dtd'))
+        filename = abspath(join(dirname(dirname(__file__)), 'affiliate_window_product_feed.1.5.dtd'))
```
rlucioni: This is the most recent DTD file distributed by AWIN, but it's out of date relative to their [documentation](http://wiki.awin.com/index.php/Product_Feed_Advanced_File_Structure). This test will be skipped until we can get a more recent DTD file. The fields being introduced in this PR will be validated by setting up a new feed.

clintonb (approved):

rlucioni:
I'm going to wait to merge this until we hear back from AWIN about the fields being added. I don't want to break the existing feed. Without an up-to-date DTD file or word back from them we don't for sure which field names are correct.

PR #1072, opened 2017-07-24
LEARNER-1816 Updated name for program banner image

Ayub-Khan:
**[LEARNER-1816](https://openedx.atlassian.net/browse/LEARNER-1816)** XSeries banner and card images displaying incorrectly	

**Description**
Updated the Name with the date. So that Cached image of cloud front is replaced with the new image we upload.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_utils.py
+++ b/course_discovery/apps/course_metadata/tests/test_utils.py
@@ -1,4 +1,4 @@
-import os
+import re
 
 import ddt
 import responses
@@ -20,18 +20,18 @@ class UploadToFieldNamePathTests(TestCase):
         self.program = ProgramFactory()
 
     @ddt.data(
-        ('/media/program', 'uuid', '.jpeg'),
-        ('/media/program', 'title', '.jpeg'),
-        ('/media', 'uuid', '.jpeg'),
-        ('/media', 'title', '.txt'),
+        ('/media/program/', 'uuid', '.jpeg'),
+        ('/media/program/', 'title', '.jpeg'),
+        ('/media/', 'uuid', '.jpeg'),
+        ('/media/', 'title', '.txt'),
         ('', 'title', ''),
     )
     @ddt.unpack
     def test_upload_to(self, path, field, ext):
         upload_to = utils.UploadToFieldNamePath(populate_from=field, path=path)
         upload_path = upload_to(self.program, 'name' + ext)
-        expected = os.path.join(path, str(getattr(self.program, field)) + ext)
-        self.assertEqual(upload_path, expected)
+        regex = re.compile(path + str(getattr(self.program, field)) + '-[a-f0-9]{12}' + ext)
+        self.assertTrue(regex.match(upload_path))
 
 
 class MarketingSiteAPIClientTests(MarketingSiteAPIClientTestMixin):
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -1,5 +1,6 @@
 import random
 import string
+import uuid
 
 import requests
 from django.utils.functional import cached_property
@@ -46,8 +47,9 @@ class UploadToFieldNamePath(UploadTo):
 
     def __call__(self, instance, filename):
         field_value = getattr(instance, self.populate_from)
+        # Update name with Random string of 12 character at the end example '-ba123cd89e97'
         self.kwargs.update({
-            'name': field_value
+            'name': str(field_value) + str(uuid.uuid4())[23:]
         })
         return super(UploadToFieldNamePath, self).__call__(instance, filename)
 


course_discovery/apps/course_metadata/utils.py
```diff
@@ -47,7 +48,7 @@ def __init__(self, populate_from, **kwargs):
     def __call__(self, instance, filename):
         field_value = getattr(instance, self.populate_from)
         self.kwargs.update({
-            'name': field_value
+            'name': str(field_value) + "{:-%d-%m-%Y}".format(datetime.datetime.now())
```
schenedx: I don't really like using the current date for this purpose. Is there another choice we can use to make this unique?
Plus, with this implementation, if two files are uploaded in the same date, they would still have the same name.
Why can't we use a random set of string if we just want cache busting? For random string, we can use regex to unit test and validate, right?

schenedx (changes requested):
@Ayub-Khan I have questions about this approach.

clintonb (changes requested):
This seems to be treating a symptom, not the actual issue. The images stored in the Discovery Service are correct. The issue is that the downstream client—the marketing site—is improperly caching the images. Fix the client, not the server.

clintonb:
Correction: It looks like this service is in fact caching an invalid image: https://prod-edx-discovery.edx.org/api/v1/programs/ef7f3b09-5660-466c-8283-0810077f353a/. How can we invalidate the cache when an image is changed without necessarily changing the image's name?

clintonb:
Perhaps the solution is to not worry about invalidation, but inform users to upload real images rather than placeholders.

course_discovery/apps/course_metadata/utils.py
clintonb: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ReplacingObjectsSameName.html

I agree with @schenedx. Use a portion of a GUID. The resulting file name should be:

`{program_uuid}.{size}.{cache_guid}.jpg`
`ef7f3b09-5660-466c-8283-0810077f353a.small.abc123.jpg`

schenedx:
@clintonb The root cause is that, cloudfront would cache the image stored on S3. Every time we upload a new image, the image reference to S3 remains the same. 
See https://dphy0qlkz2ftb.cloudfront.net/media/programs/banner_images/a87e5eac-3c93-45a1-a8e1-4c79ca8401c8.large.jpg VS https://dphy0qlkz2ftb.cloudfront.net/media/programs/banner_images/a87e5eac-3c93-45a1-a8e1-4c79ca8401c8.large.jpg?gdaf=4345

The idea is that, we should use a different name to reference this S3 image every time a new file has been uploaded to get around the image cache.

clintonb:
@schenedx @Ayub-Khan what's the plan for cleaning up old images?

schenedx:
@clintonb Now I remember why we didn't create a unique name for each banner image in the first place. I don't know a good plan to clean old images... Aaarrrgg

clintonb:
You could write a management command that cleans up older files after a certain period of time. We would need to establish an SLA with API clients (especially affiliates) so that they understand that image links are only good for 30 days, for example. Ultimately, we could do something similar for other static assets—CSS and JS—on the frontend.

schenedx:
@Ayub-Khan @clintonb 
I think it's not worth the trouble to implement this management command which clean up old banner images. We only add very few old banner images for a long periods of time (years). It's not a problem we need to worry about at the moment I say.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -20,18 +20,18 @@ def setUp(self):
...
+        ('/media/', 'title', '.txt'),
         ('', 'title', ''),
     )
     @ddt.unpack
     def test_upload_to(self, path, field, ext):
         upload_to = utils.UploadToFieldNamePath(populate_from=field, path=path)
         upload_path = upload_to(self.program, 'name' + ext)
-        expected = os.path.join(path, str(getattr(self.program, field)) + ext)
-        self.assertEqual(upload_path, expected)
+        regex = re.compile(path + str(getattr(self.program, field)) + '-............' + ext)
```
clintonb: Drop the dots. Use `-[a-f0-9]{12}`.

clintonb (approved):

course_discovery/apps/course_metadata/utils.py
```diff
@@ -47,7 +48,7 @@ def __init__(self, populate_from, **kwargs):
     def __call__(self, instance, filename):
         field_value = getattr(instance, self.populate_from)
         self.kwargs.update({
-            'name': field_value
+            'name': str(field_value) + str(uuid.uuid4())[23:]
```
schenedx: So, we use the first 12 characters? Can you please add a comment?

schenedx (approved):
Minor comment, rest LGTM 👍

course_discovery/apps/course_metadata/utils.py
Ayub-Khan: last 12 characters

PR #1074, opened 2017-07-24
Updated Publisher base template to use correct block name for CSS

clintonb:
ECOM-5776


clintonb:
This fixes the Publisher CSS issue introduced by https://github.com/edx/course-discovery/pull/474.

FYI @muhammad-ammar

PR #1075, opened 2017-07-25
Using Django Compressor for Publisher CSS files

clintonb:


clintonb:
This should _finally_ resolve the errors introduced by https://github.com/edx/course-discovery/pull/474.

PR #1079, opened 2017-07-31
fix serializer fields

muhammad-ammar:


Qubad786 (approved):
👍 once the tests pass.

muzaffaryousaf (approved):
👍

Qubad786 (approved):
fit 👍

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1956,6 +1956,18 @@ def test_context(self):
             self.assertEqual(response.context_data['publisher_courses_url'], reverse('publisher:publisher_courses'))
             self.assertEqual(response.context_data['allowed_page_sizes'], json.dumps(COURSES_ALLOWED_PAGE_SIZES))
 
+    @mock.patch('course_discovery.apps.publisher.models.CourseState.course_team_status', new_callable=mock.PropertyMock)
+    @mock.patch('course_discovery.apps.publisher.models.CourseState.internal_user_status',
+                new_callable=mock.PropertyMock)
+    def test_course_state_statuses(self, mocked_internal_user_status, mocked_course_team_status):
+        """ Verify that course_state statuses raise no exception. """
```
mushtaqak, line 1964: I m much inclined to docstrings on new line, but it's up to you if you wan tot change.

course_discovery/apps/publisher/serializers.py
```diff
@@ -61,14 +61,17 @@ def get_course_team_status(self, course):
...
         try:
             course_team_status = course.course_state.course_team_status
         except ObjectDoesNotExist:
-            return {
-                'status': '',
-                'date': ''
-            }
+            return default_status
 
+        course_team_status = default_status if course_team_status is None else course_team_status
```
mushtaqak, line 74: This may return an object accessed before assigned error? may be not. But I always prefer to initialise a variable in outer block, so that if accessed in inner block or/ outer block, we don't get access error.  

May be `default_status ` may not be needed if we initialise `course_team_status` with `{'status': '', 'date': ''}` value. In this way, we don't have to worry below, if try-except does ended in exception then we return the `course_team_status` otherwise `course_team_status` would be updated.

course_discovery/apps/publisher/serializers.py
```diff
@@ -79,14 +82,17 @@ def get_internal_user_status(self, course):
...
         try:
             internal_user_status = course.course_state.internal_user_status
         except ObjectDoesNotExist:
-            return {
-                'status': '',
-                'date': ''
-            }
+            return default_status
 
+        internal_user_status = default_status if internal_user_status is None else internal_user_status
```
mushtaqak, line 95: same as `course_team_status`

mushtaqak (changes requested):
@muhammad-ammar Just a few suggestions

mushtaqak (approved):
After discussion,  👍

PR #1080, opened 2017-07-31
Added support for filtering programs by marketing slug

clintonb:
This will ultimately be used by the marketing site to retrieve program data without having to store a separate reference (e.g. UUID).

LEARNER-2053


clintonb:
FYI @edx/learner

PR #1081, opened 2017-08-01
ownership and course state does not change on pc edit

asadazam93:
### Description
There are fields in the course that only PCs have the ability to fill in namely (About Video Link), but since the project coordinator has no role in the actual workflow of the course so when the project coordinator makes an edit to the course the ownership is changed to PC along with the course state changed to Draft and that disturbs the actual workflow between the course team and the marketing team.
Solution 
So now when a project coordinator edits a course the course state will not be changed and neither the ownership (however a history object for that edit would be created) hence the workflow of the course that involves the course team and marketing team would not be disturbed. A project coordinator is only making edits on behalf of the course team, he is not a part of the workflow of sending the course for review neither marking the course as reviewed so it does not make sense to transfer ownership to the PC.

### How to Test?

- Login as project coordinator of a course.
- Edit the course
- Confirm that the ownership and state of the course did not change but a history object for that edit was created

**Sandbox**
- https://discovery-educator-973.sandbox.edx.org/publisher/courses/3/

### Reviewers
- [ ] Code review: @attiyaIshaque 
- [ ] Code review: @awaisdar001  

FYI: @sstack22 

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -2612,6 +2612,40 @@ class CourseEditViewTests(SiteMixin, TestCase):
         course_state = CourseState.objects.get(id=self.course.course_state.id)
         self.assertEqual(course_state.name, CourseStateChoices.Draft)
 
+    def test_edit_course_with_project_coordinator(self):
+        """
+        Verify that on editing the course as project coordinator the ownership and
+        course state does not change.
+        """
+        project_coordinator = UserFactory()
+        project_coordinator.groups.add(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))
+        project_coordinator.groups.add(Group.objects.get(name=PROJECT_COORDINATOR_GROUP_NAME))
+        factories.CourseUserRoleFactory(course=self.course, role=PublisherUserRole.ProjectCoordinator,
+                                        user=project_coordinator)
+        self.client.logout()
+        self.client.login(username=project_coordinator.username, password=USER_PASSWORD)
+        self._assign_permissions(self.organization_extension)
+
+        self.course.course_state.name = CourseStateChoices.Review
+        self.course.course_state.owner_role = PublisherUserRole.MarketingReviewer
+        self.course.course_state.save()
+
+        post_data = self._post_data(self.organization_extension)
+        post_data['team_admin'] = self.course_team_role.user.id
+
+        response = self.client.post(self.edit_page_url, data=post_data)
+
+        self.assertRedirects(
+            response,
+            expected_url=reverse('publisher:publisher_course_detail', kwargs={'pk': self.course.id}),
+            status_code=302,
+            target_status_code=200
+        )
+
+        course_state = CourseState.objects.get(id=self.course.course_state.id)
+        self.assertEqual(course_state.name, CourseStateChoices.Review)
+        self.assertEqual(course_state.owner_role, PublisherUserRole.MarketingReviewer)
+
     def test_edit_course_with_ownership_changed(self):
         """
         Verify that on editing course state changed to `Draft` and ownership changed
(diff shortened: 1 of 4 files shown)


course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2612,6 +2612,40 @@ def test_update_course_with_state(self):
         course_state = CourseState.objects.get(id=self.course.course_state.id)
         self.assertEqual(course_state.name, CourseStateChoices.Draft)
 
+    def test_edit_course_with_project_coordinator(self):
+        """
+        Verify that on editing the course as project coordinator the ownership and
+        course state does not change.
+        """
+        project_coordinator = UserFactory()
+        project_coordinator.groups.add(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))
```
attiyaIshaque, line 2621: You can also use this:
```project_coordinator.groups.add(Group.objects.get(name= for name in [INTERNAL_USER_GROUP_NAME, PROJECT_COORDINATOR_GROUP_NAME]))```

Comment (asadazam93) [reply course_discovery/apps/publisher/tests/test_views.py:2621]: I tried but get cannot return multiple groups.

Comment (attiyaIshaque) [reply course_discovery/apps/publisher/tests/test_views.py:2621]: Okay. :)

**Review by attiyaIshaque — APPROVED** [review]
@asadazam93 code good looks to me, aslo i have tested on sandbox. 👍

File: course_discovery/apps/publisher/views.py [hunk 2]
```diff
@@ -406,13 +406,15 @@ def form_valid(self, form):
             # Change course state to draft if marketing not yet reviewed or
             # if marketing person updating the course.
             if not self.object.course_state.marketing_reviewed or user_role.role == PublisherUserRole.MarketingReviewer:
-                if self.object.course_state.name != CourseStateChoices.Draft:
```
awaisdar001: What is `self.object`? This seems a very genric name to me. Please update it to whatever type of object that is. I.e. `course` or `course_run`

course_discovery/apps/publisher/views.py
```diff
@@ -406,13 +406,15 @@ def form_valid(self, form):
...
+                if (self.object.course_state.name != CourseStateChoices.Draft and
+                        user_role.role != PublisherUserRole.ProjectCoordinator):
                     self.object.course_state.change_state(
                         state=CourseStateChoices.Draft, user=user, site=self.request.site
                     )
 
                 # Change ownership if user role not equal to owner role.
-                if self.object.course_state.owner_role != user_role.role:
+                if (self.object.course_state.owner_role != user_role.role and
```
awaisdar001: If you are checking that `self.object.course_state.owner_role` should not be equal to 2 or more variables do this.
`self.object.course_state.owner_role not in  (user_role.role, PublisherUserRole.ProjectCoordinator)`

course_discovery/apps/publisher/views.py
```diff
@@ -406,13 +406,15 @@ def form_valid(self, form):
             # Change course state to draft if marketing not yet reviewed or
             # if marketing person updating the course.
             if not self.object.course_state.marketing_reviewed or user_role.role == PublisherUserRole.MarketingReviewer:
-                if self.object.course_state.name != CourseStateChoices.Draft:
+                if (self.object.course_state.name != CourseStateChoices.Draft and
```
awaisdar001: is `self.object` is a keyword here which is required for backend oprations? Or it is just a normal veriable name? If it is just a veriable, can we change this somthing meaningful ? maybe `course` or `course_run`? then the following statement can become more readable. 

`self.object.course_state.name != CourseStateChoices.Draft` => `course.course_state_name != CourseStateChoices.Draft` 

Note the `course_state_name` property method.

awaisdar001 (changes requested):

course_discovery/apps/publisher/views.py
asadazam93: self.object is an object instance the view is operating on, i.e it is an instance of a django model in this case (it is inherited from the mixin we are using). So that is why we cannot change it's name.

course_discovery/apps/publisher/views.py
asadazam93: These are two different conditions. First check is that the the owner role is not equal to the logged in user role and the second condition is that the logged in user is not a project coordinator.

PR #1083, opened 2017-08-02
Add start date, end date, and CEUs to Programs

tuchfarber:
In order to make WL certificates dynamic, we need this program information to be available to the template.

[WL-1119]
[WL-1120]


tuchfarber:
@clintonb I don't see start and end date being an issue, but we haven't broached where CEU/CMEs should be stored. Wanted a Learner perspective on it. The other option I can think of is to store them at the course level and bubble them up in the ProgramSerializer somewhere. Thoughts?

course_discovery/apps/course_metadata/models.py
```diff
@@ -755,6 +755,11 @@ class Program(TimeStampedModel):
         null=True, blank=True,
         help_text=_('This field is now deprecated (ECOM-6021).'
                     'Estimated number of weeks needed to complete a course run belonging to this program.'))
+    start_date = models.DateField(null=True, blank=True, db_index=True)
```
clintonb, line 758: None of these new fields need to be indexed unless you intend on actually querying the DB with them.

course_discovery/apps/course_metadata/models.py
```diff
@@ -755,6 +755,11 @@ class Program(TimeStampedModel):
...
+    end_date = models.DateField(null=True, blank=True, db_index=True)
+    continuing_education_units = models.DecimalField(
```
clintonb, line 760: Why is this a decimal field?

course_discovery/apps/course_metadata/migrations/0060_auto_20170802_2006.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
clintonb, line 1: You should only have one migration.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -121,9 +121,9 @@ class ProgramAdmin(admin.ModelAdmin):
     # ordering the field display on admin page.
     fields = (
         'uuid', 'title', 'subtitle', 'status', 'type', 'partner', 'banner_image', 'banner_image_url', 'card_image_url',
-        'marketing_slug', 'overview', 'credit_redemption_overview', 'video', 'weeks_to_complete',
-        'min_hours_effort_per_week', 'max_hours_effort_per_week', 'courses', 'order_courses_by_start_date',
-        'custom_course_runs_display', 'excluded_course_runs', 'authoring_organizations',
+        'marketing_slug', 'overview', 'credit_redemption_overview', 'video', 'weeks_to_complete', 'start_date',
```
clintonb, line 124: Chat with UX and Tatiana in Marketing about grouping these fields in some logical manner. See https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets.

clintonb (changes requested):
Tests? Squash your commits.

clintonb:
FYI @edx/learner

course_discovery/apps/course_metadata/models.py
tuchfarber: Noted. Will update.

course_discovery/apps/course_metadata/models.py
tuchfarber: Needed to store half credits. The current Sys Eng program we're running is 8.5 CEUs.

clintonb:
Should CEUs be a property of course runs, courses, or programs?

course_discovery/apps/course_metadata/models.py
douglashall: I would store the `continuing_education_unit` value at the most atomic level possible. I could see the next request coming in from folks being the display of CEUs on the course about page. If you add the `continuing_education_unit` field to the Course model you could calculate Program CEUs.

The problem with that approach though is that there would not be a good way for course staff to author the CEU data at the present time. Course staff do not have access to Django admin in the discovery service, so someone here at edX (project coordinator) would have to take on this step when setting up new courses.

clintonb:
@tuchfarber it will be best to split this into two PRs—one for dates, another for CEUs. The dates change is fine, as long as you communicate it to PCs. The CEUs change will require more discussion.

tuchfarber:
It's been decided that CEU metadata should fall under Courses, not programs.

clintonb:
@tuchfarber delete your branches after closing PRs

PR #1089, opened 2017-08-11
Removed duplicate staff and subject UUIDs from programs index

clintonb:
LEARNER-2229


clintonb:
@edx/learner please review

clintonb:
@edx/learner please review

schenedx (approved):
👍

PR #1090, opened 2017-08-11
Disabled Haystack's constant updating of index mappings

clintonb:
Haystack attempts to always ensure the index mapping is up-to-date. 
There is a bug in this code (at least when using Elasticsearch 1.x) that
results in an invalid mapping comparison. Given that our mapping is 
static, this feature is simply a waste of resources since we always 
make three API calls to Elasticsearch for every search call.

This commit disables this mapping check/update feature when executing a 
search query.

LEARNER-2188


clintonb:
FYI @edx/learner @edx/devops @ormsbee

schenedx (commented):
👍

PR #1097, opened 2017-08-21
Add subject list/details endpoints

mikedikan:
ECOM-7905

Adding endpoints to get individual subjects and a full subject listing.  The subjects will be used on the marketing site to render the subject landing pages and eventually the all-subjects page.

@edx/learner-growth & @edx/learner-heart  FYI

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -164,9 +164,13 @@ class FAQSerializer(serializers.ModelSerializer):
 class SubjectSerializer(serializers.ModelSerializer):
     """Serializer for the ``Subject`` model."""
 
+    @classmethod
+    def prefetch_queryset(cls):
+        return Subject.objects.filter()
+
     class Meta(object):
         model = Subject
-        fields = ('name', 'subtitle', 'description', 'banner_image_url', 'card_image_url', 'slug',)
+        fields = ('name', 'subtitle', 'description', 'banner_image_url', 'card_image_url', 'slug', 'uuid')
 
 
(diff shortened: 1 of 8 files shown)


course_discovery/apps/api/v1/tests/test_views/test_subjects.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.client.logout()
+        response = self.client.get(self.list_path)
+        assert response.status_code == 403
+
+    def test_list(self):
+        """ Verify the endpoint returns a list of all subjects. """
+        SubjectFactory.create_batch(4)
+        expected = Subject.objects.all()
+        with self.assertNumQueries(4):
```
bderusha: Does this mean it will take 30 queries to list all subjects? Seems...less than ideal

mikedikan: Changed to 8 subjects in factory batch for clarity 👍

course_discovery/apps/api/serializers.py
```diff
@@ -164,9 +164,13 @@ class Meta(object):
 class SubjectSerializer(serializers.ModelSerializer):
     """Serializer for the ``Subject`` model."""
 
+    @classmethod
+    def prefetch_queryset(cls):
```
MatthewPiatetsky, line 168: I think usually this method is created when you're using a prefetch_related or select_related, which you're not doing here. I suppose if we're fine having the same subjects for different partners it doesn't really matter

MatthewPiatetsky:
👍

PR #1100, opened 2017-08-30
Updated create_or_update_partner management command to modify Sites

clintonb:
The management command now creates and updates Site instances in  addition to Partner instances. This is necessary for proper multi-tenancy.

PLAT-1715


schenedx (approved):
👍

clintonb:
See #1101

PR #1101, opened 2017-08-30
Partner.site is now required

clintonb:
Partner instances are now required to have an associated Site for multi-tenancy.

PLAT-1715


clintonb:
The commit from https://github.com/edx/course-discovery/pull/1100 is included here.

schenedx (approved):
👍

PR #1108, opened 2017-09-06
LEARNER-2409 Add currency view

MatthewPiatetsky:
Currently uses https://openexchangerates.org (I created an API key that gives use 1000 free uses per month. Assuming the caching works correctly that should be plenty (we should only need ~30 per month))
Adding the API key to the secure repo is still a TODO

I was initially adding another python package to get the currency symbols (and potentially another for the alpha3 currency codes).
However, since all we need for this at the moment is just the mapping and not more advanced functionality, I think a regular dict should be sufficient and extra dependencies are not necessary.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_currency.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_currency.py
@@ -1,0 +1,74 @@
+import mock
+from django.core.cache import cache
+from django.test import override_settings
+from django.urls import reverse
+
+from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase
+from course_discovery.apps.api.v1.views.currency import CurrencyView
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
+
+
+class CurrencyViewTests(APITestCase):
+    list_path = reverse('api:v1:currency')
+
+    def setUp(self):
+        super(CurrencyViewTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.request.user = self.user
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+
+        # Clear the cache between test cases, so they don't interfere with each other.
+        cache.clear()
+
+    def test_get(self):
+        """ Verify the endpoint returns the right currency data and uses the cache. """
+        rates = {"GBP": 0.766609, "CAD": 1.222252, "CNY": 6.514431, "EUR": 0.838891}
+        currencies = {
+            'GBR': {'code': 'GBP', 'symbol': u'£'},
+            'CHN': {'code': 'CNY', 'symbol': u'¥'},
+            'CAN': {'code': 'CAD', 'symbol': '$'}
+        }
+        eurozone_countries = ['FRA']
+        get_data_return_value = [rates, currencies, eurozone_countries]
+
+        expected = {
+            "GBR": {"code": "GBP", "symbol": u"£", "rate": 0.766609},
+            "CAN": {"code": "CAD", "symbol": "$", "rate": 1.222252},
+            "CHN": {"code": "CNY", "symbol": u"¥", "rate": 6.514431},
+            "FRA": {"code": "EUR", "symbol": "€", "rate": 0.838891}
+        }
+
+        with mock.patch.object(CurrencyView, 'get_data', return_value=get_data_return_value) as mock_get_rates:
+            response = self.client.get(self.list_path)
+            self.assertDictEqual(response.data, expected)
+            self.assertEqual(mock_get_rates.call_count, 1)
+
+            # next request hits the cache
+            response = self.client.get(self.list_path)
+            self.assertEqual(mock_get_rates.call_count, 1)
+
+            # clearing the cache calls means the function gets called again
+            cache.clear()
+            response = self.client.get(self.list_path)
+            self.assertEqual(mock_get_rates.call_count, 2)
+
+    def test_no_api_key(self):
+        response = self.client.get(self.list_path)
+        self.assertEqual(response.json(), {})
+
+    @override_settings(OPENEXCHANGERATES_API_KEY='test')
+    def test_get_rates(self):
+        def mocked_requests_get(*args, **kwargs):  # pylint: disable=unused-argument
+            class MockResponse:
+                def __init__(self, json_data, status_code, text):
+                    self.json_data = json_data
+                    self.status_code = status_code
+                    self.text = text
+
+                def json(self):
+                    return self.json_data
+            return MockResponse({"bad": "data"}, 500, "baddata")
+        with mock.patch('course_discovery.apps.api.v1.views.currency.requests.get', side_effect=mocked_requests_get):
+            response = self.client.get(self.list_path)
+            response_json = response.json()
+            self.assertEqual(response_json, {})
--- a/course_discovery/apps/api/v1/views/currency.py
+++ b/course_discovery/apps/api/v1/views/currency.py
@@ -1,0 +1,69 @@
+import logging
+
+import requests
+from django.conf import settings
+from rest_framework import views
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework_extensions.cache.decorators import cache_response
+
+
+class CurrencyView(views.APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get_rates(self):
+        try:
+            app_id = settings.OPENEXCHANGERATES_API_KEY
+            if app_id:
+                url = 'https://openexchangerates.org/api/latest.json'
+                response = requests.get(url, params={'app_id': app_id}, timeout=2)
+                response_json = response.json()
+                result = response_json['rates']
+                return result
+            else:
+                logging.warning('No app id available for openexchangerates')
+                return {}
+        except Exception as e:  # pylint: disable=broad-except
+            response_text = '' if not isinstance(response, object) else response.text
+            message = 'Exception Type {}. Message {}. Response {}.'.format(
+                type(e).__name__, e, response_text
+            )
+            logging.error('Could not retrieve rates from openexchangerates. ' + message)
+            return {}
+
+    def get_data(self):
+        rates = self.get_rates()
+        # ISO 3166-1 alpha-3 codes
+        currencies = {
+            'IND': {'code': 'INR', 'symbol': u'₹'},
+            'BRA': {'code': 'BRL', 'symbol': 'R$'},
+            'MEX': {'code': 'MXN', 'symbol': '$'},
+            'GBR': {'code': 'GBP', 'symbol': u'£'},
+            'AUS': {'code': 'AUD', 'symbol': '$'},
+            'CHN': {'code': 'CNY', 'symbol': u'¥'},
+            'COL': {'code': 'COP', 'symbol': '$'},
+            'PER': {'code': 'PEN', 'symbol': 'S/.'},
+            'CAN': {'code': 'CAD', 'symbol': '$'}
+        }
+        eurozone_countries = [
+            'AUT', 'BEL', 'CYP', 'EST', 'FIN', 'FRA', 'DEU', 'GRC', 'IRL',
+            'ITA', 'LVA', 'LTU', 'LUX', 'MLT', 'NLD', 'PRT', 'SVK', 'SVN', 'ESP'
+        ]
+        return [rates, currencies, eurozone_countries]
+
+    # Cache exchange rates for 1 day
+    @cache_response(60 * 60 * 24)
+    def get(self, request, *args, **kwargs):
+        rates, currencies, eurozone_countries = self.get_data()
+        if not rates:
+            return Response({})
+
+        for country, currency in currencies.items():
+            currency_name = currency['code']
+            currencies[country]['rate'] = rates.get(currency_name)
+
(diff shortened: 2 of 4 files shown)


course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,55 @@
...
+from rest_framework import views
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework_extensions.cache.decorators import cache_response
+
+
+class CurrencyView(views.APIView):
+    permission_classes = (IsAuthenticated,)
+
+    EUROZONE_COUNTRIES = [
```
mikedikan: Should this be in this file and not in settings?

course_discovery/apps/api/v1/tests/test_views/test_currency.py
```diff
@@ -0,0 +1,68 @@
...
+    def setUp(self):
+        super(CurrencyViewTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.request.user = self.user
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+
+        # Clear the cache between test cases, so they don't interfere with each other.
+        cache.clear()
+
+    def test_list(self):
```
mikedikan: Is there a way to have this in an external file?

mikedikan (commented):
Small comments - looks good so far 👍

course_discovery/apps/api/v1/tests/test_views/test_currency.py
MatthewPiatetsky: i just shortened the test data instead and refactored the files a bit to allow easily mocking all three files

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: imo this and the other variable in this file are pretty specific implementation things rather than settings that may be changed/reused. (i could also see this code being replaced by a package in the future)
i think if we end up wanting to reuse this anywhere else though then it should be a setting

mikedikan (approved):
:shipit:

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,58 @@
...
+
+class CurrencyView(views.APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get_rates(self):
+        try:
+            # TODO add key to secure repo
+            api_key = settings.OPENEXCHANGERATES_API_KEY
+            url = 'https://openexchangerates.org/api/latest.json?app_id='
+            response = requests.get(url + api_key)
```
jdmulloy: What happens when this service is down or slow to respond. What is the timeout?

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,58 @@
...
+from rest_framework_extensions.cache.decorators import cache_response
+
+
+class CurrencyView(views.APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get_rates(self):
+        try:
+            # TODO add key to secure repo
+            api_key = settings.OPENEXCHANGERATES_API_KEY
```
jdmulloy: What happens if the key has not been set, such as in a sandbox or devstack?

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,58 @@
...
+            url = 'https://openexchangerates.org/api/latest.json?app_id='
+            response = requests.get(url + api_key)
+            result = response.json()['rates']
+            return result
+        except:  # pylint: disable=bare-except
+            logging.error('Could not retrieve rates from openexchangerates')
```
jdmulloy: For debugging and troubleshooting purposes when failures occur logging the exception type and message here would be very useful.

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class CurrencyView(views.APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get_rates(self):
+        try:
+            # TODO add key to secure repo
+            api_key = settings.OPENEXCHANGERATES_API_KEY
+            url = 'https://openexchangerates.org/api/latest.json?app_id='
```
jdmulloy: It might also be nice to have the URL in a config variable in case it needs to change in the future, or if we make this more modular to allow other users of our code to use other providers.

Nit: It would be cleaner and less fragile to pass the app_id parameter as an argument to the requests.get method instead of constructing a string. See http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls for details.

Perhaps something like the following.
```
url = 'https://openexchangerates.org/api/latest.json'
response = requests.get(url, params={'api_key':  api_key})
```

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,58 @@
...
+            response = requests.get(url + api_key)
+            result = response.json()['rates']
+            return result
+        except:  # pylint: disable=bare-except
+            logging.error('Could not retrieve rates from openexchangerates')
+
+    def get_data(self):
+        rates = self.get_rates()
```
jdmulloy: What happens when get_rates() returns None, such as when an exception is caught? In the method get() below "rates" is assumed to be a dictionary. If it's None then we will get stack traces saying something like```'NoneType' object has no attribute 'get'```

jdmulloy (changes requested):

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: added an if statement

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: the code wouldn't work as it stands with another provider. imo i would keep it like this and move it to config if a second use case comes up. added the params dict though

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: added a 2 second timeout.
i figured we don't want to wait too long, but at the same time this should only run once a day, which is why i said 2 seconds over 1.

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: added message/status code/exception type to the log

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: changed get_rates to return dict

MatthewPiatetsky:
@jdmulloy still need to update tests and make sure all my new code works, but otherwise did I address your comments?

AlasdairSwan:
@MatthewPiatetsky I have a question about the format of the returned data. So the response would be in the following format;
```
"MEX": [
    "MXN",
    "$",
    17.795193
]
```
How about making it something like;
```
"MEX": {
    country: "MXN",
    unit: "$",
    rate: 17.795193
}
```
So it's easier to read the code that would use it. For example;
```
var displayPrice = MEX[1] + ' ' + MEX[0] + ' ' + (price * MEX[2]);
```
isn't as easy to read as
```
var displayPrice = MEX.unit + ' ' + MEX.country + ' ' + (price * MEX.rate);
```

MatthewPiatetsky:
@AlasdairSwan i was just keeping down the size of the response, but that's fine, I can change it to that

AlasdairSwan:
Thanks @MatthewPiatetsky, I just think it would make it easier to work with. These may be clearer names;
```
"MEX": {
    code: "MXN",
    symbol: "$",
    rate: 17.795193
}
```

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -14,13 +14,22 @@ class CurrencyView(views.APIView):
...
-            logging.error('Could not retrieve rates from openexchangerates')
+            app_id = settings.OPENEXCHANGERATES_API_KEY
+            if app_id:
+                url = 'https://openexchangerates.org/api/latest.json'
+                response = requests.get(url, params={'app_id': app_id}, timeout=2)
+                response_json = response.json()
+                result = response_json['rates']
+                return result
+            else:
+                logging.error('No app id available for openexchangerates')
```
feanil: Is this an error?  Won't this be true for all non edx deployments to start with?

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -14,13 +14,22 @@ class CurrencyView(views.APIView):
...
-            response = requests.get(url + api_key)
-            result = response.json()['rates']
-            return result
-        except:  # pylint: disable=bare-except
-            logging.error('Could not retrieve rates from openexchangerates')
+            app_id = settings.OPENEXCHANGERATES_API_KEY
+            if app_id:
+                url = 'https://openexchangerates.org/api/latest.json'
+                response = requests.get(url, params={'app_id': app_id}, timeout=2)
+                response_json = response.json()
```
feanil, line 20: This call can fail if the response is not JSON.  If you try to use `response_json.get('message')` in the except section, that will probably just return `None`.  You should handle this case.  You should do something different if requests fails to parse JSON from the response.

feanil (changes requested):

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: running this code without an app id would mean you'd get a 403 response back.
so shouldn't this be an error if this code will not work without an app id?
if your question is really about non edx deployments, i'm not sure this is a feature that would work well with non edx deployments, since it assumes they need to make a key with this service.
would you suggest making this code fully edx specific?

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: what additional granularity would this provide, since we would see a json decode error in the exception type even if message was none.
IMO catching specific exceptions would be required if I was following some API docs, but they don't have an official python API client, so if there was an error, we don't know exactly what it would mean regardless, so the goal here was just to have a catch all

course_discovery/apps/api/v1/views/currency.py
jdmulloy: Is using this feature going to become a requirement for non edx deployments? If this feature is optional then there should be a flag to enable/disable it. Either way I would argue that this should be a warning and the code should still operate, just without the exchange rate information.

course_discovery/apps/api/v1/views/currency.py
```diff
@@ -0,0 +1,69 @@
...
+                result = response_json['rates']
+                return result
+            else:
+                logging.error('No app id available for openexchangerates')
+                return {}
+        except Exception as e:  # pylint: disable=broad-except
+            message = 'Status {}. Exception Type {}. Message {}.'.format(
+                response.status_code, response_json.get('message'), type(e).__name__
```
jdmulloy: Depending on where the failure occurs in the try block ```response.status_code``` or ```response_json.get()``` may not exist, which will cause this line to throw an exception. Logging just the exception type name and message is probably sufficient here. If you want to log the HTTP status code and/or the response body, you'll need to check that they exist before trying to use them.

jdmulloy (changes requested):

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: this is a standalone endpoint at the moment not called by anything.
The feature we're building would not need to be used by non edx deployments. However, we may end up building the code to call this as an npm package only installed for edx, or some javascript included in the edx templates, so it won't necessarily need a feature flag to be edx only.
If we log an error the code will still operate. This is not like raising an exception, correct?
However, we'd want to know that something broke so we could look into it

course_discovery/apps/api/v1/views/currency.py
feanil: @MatthewPiatetsky if you don't catch the json decode exception differently, you will end up throwing another exception in your except clause because it assumes that the `response_json` variable is defined and is a python dict.  When you do `response_json.get()` it probably throw a name error.

course_discovery/apps/api/v1/views/currency.py
feanil: @MatthewPiatetsky I think it makes sense to log this but I thin you should log at the info or warning level rather than error.

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: @jdmulloy suggested above to only use the exception message and type, so I am no longer logging those

feanil: You may want to log `respons.text` as this might have more information on why the call failed. eg. The server keeps sending you their 500 page content.

course_discovery/apps/api/v1/views/currency.py
MatthewPiatetsky: changed to log exception type name and message

feanil: You want `response.text` which will work even if the content is not json.   looking into the json won't if the content is not json.

MatthewPiatetsky: that is what i'm currently using

feanil (approved):

jdmulloy (approved):
This looks good.

PR #1109, opened 2017-09-07
Fix enables person family name in discovery to be empty.

adeelkhan:
Adding person in discovery throws 500 if family name is
empty. Person expects family name to generate full name.
This Fix would enable family name to be empty as its not a
required field.

LEARNER-1701


schenedx (approved):
👍

Ayub-Khan (approved):

PR #1111, opened 2017-09-08
[ENT-629] Add `description` field to ProgramType

UmanShahzad:
We need to display program type information (in particular, a description for a program type) on the program enrollment landing page for [ENT-629](https://openedx.atlassian.net/browse/ENT-629). This PR adds that model field and gets it serialized.

**JIRA tickets**: [ENT-629](https://openedx.atlassian.net/browse/ENT-629)

**Merge deadline**: ~2 weeks after this PR opened.

**Testing instructions**: 

1. Checkout this branch in your devstack course-discovery and apply the migration.
1. Make a program type in course-discovery, filling in the description field and the slug. Note the slug.
1. Go to `/api/v1/program_type` and see that in the list of Program Types, you also see the `description `field for all program types.
1. Go to `/api/v1/program_type/{slug}` using the slug you noted above, and see that the `description` field shows for the one you made.


douglashall (approved):
@clintonb @mikedikan @bderusha Any objections to adding this field?

clintonb (commented):
In the future, let's talk about these changes before the PR phase.

1. How will this text get translated?
2. Are you aware that the same description will be used across all white label sites, should they choose to use this field?
3. Given that this is marketing text, why isn't it on the marketing site?
4. Why are Enterprise learners not using the same marketing site as other learners?

douglashall:
@UmanShahzad We decided to not do this. I will follow up with details on ENT-629.

PR #1102, opened 2017-09-03
Added Publisher API endpoint to publish data to sources of truth

clintonb:
LEARNER-2472

PR #1117, opened 2017-09-12
Publishing course runs to Studio upon creation

clintonb:
LEARNER-2469


clintonb:
@schenedx review and merge at your convenience

clintonb:
This was handled by #1102

PR #1121, opened 2017-09-18
Updated Publisher publish endpoint to return detailed status

clintonb:
The endpoint now returns the publication status of each service to which it published. This will better enable the debugging of errors that occur when publication to one, or more, services fails.

LEARNER-2472


clintonb:
FYI @schenedx

schenedx (commented):
👍 This is a good api enhancement

PR #1129, opened 2017-09-20
Add translations to the Subjects model

BenjiLee:
@edx/learner-heart 

Phase 1 (https://openedx.atlassian.net/browse/LEARNER-2615) of https://openedx.atlassian.net/browse/LEARNER-2314

1. Add django-parler to requirements
2. Install Django-parler (this step may be different since we configure prod in a different repo)
3. Create migration to add translations table for subjects
4. Copy data to new table.
5. Modify api to look up and return new fields without changing the response structure
6. Update admin interface to make the unused fields to be uneditable. 

Accessing the subjects list with the query parameter "language" will get you the language or default. 

Language available: 'en', 'es'
Default: 'en'

 http://localhost:18381/api/v1/subjects/?language_code=es Returns es
 http://localhost:18381/api/v1/subjects/ Returns default en
 http://localhost:18381/api/v1/subjects/?language_code=es-419 return es
 http://localhost:18381/api/v1/subjects/?language_code=not-a-language returns en

TODO:
- [x]  Unit Tests


course_discovery/apps/api/v1/views/subjects.py
```diff
@@ -22,7 +22,7 @@ class SubjectViewSet(viewsets.ReadOnlyModelViewSet):
     pagination_class = ProxiedPagination
 
     def get_queryset(self):
-        return serializers.SubjectSerializer.prefetch_queryset()
+        return serializers.SubjectSerializer.prefetch_queryset(self.request.query_params.get('language', 'en'))
```
amangano-edx: We should use the default from `django.settings`, rather than hardcoding this to `en`, as open source users may default to something other than en.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -220,7 +221,8 @@ class OrganizationAdmin(admin.ModelAdmin):
 
 
 @admin.register(Subject)
-class SubjectAdmin(admin.ModelAdmin):
+class SubjectAdmin(TranslatableAdmin):
+    exclude = ('name', 'subtitle', 'description')
```
amangano-edx: Could you add a comment explaining why we need to exclude these?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -220,7 +221,8 @@ class OrganizationAdmin(admin.ModelAdmin):
...
     list_display = ('uuid', 'name', 'slug',)
```
amangano-edx: name should probably be excluded as well.

course_discovery/apps/course_metadata/migrations/0058_auto_20170919_1927.py
```diff
@@ -0,0 +1,34 @@
...
+                ('master', models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='translations', to='course_metadata.Subject')),
+            ],
+            options={
+                'verbose_name': 'Subject model translations',
+            },
+        ),
+        migrations.AlterUniqueTogether(
+            name='subjecttranslation',
+            unique_together=set([('language_code', 'master')]),
+        ),
```
amangano-edx: Should we also build an index on language code since we'll be querying by it a lot?

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
```diff
@@ -0,0 +1,58 @@
...
+from django.db import migrations
+from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
+
+
+def forwards_func(apps, schema_editor):
+    Subject = apps.get_model('course_metadata', 'Subject')
+    SubjectTranslation = apps.get_model('course_metadata', 'SubjectTranslation')
+
+    for object in Subject.objects.all():
```
amangano-edx: Is object a special word in python?

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
```diff
@@ -0,0 +1,58 @@
...
+def backwards_func(apps, schema_editor):
+    Subject = apps.get_model('course_metadata', 'Subject')
+    SubjectTranslation = apps.get_model('course_metadata', 'SubjectTranslation')
+
+    for object in Subject.objects.all():
+        translation = _get_translation(object, SubjectTranslation)
+        object.name = translation.name_t
+        object.subtitle = translation.subtitle_t
+        object.description = translation.description_t
+        object.save()   # Note this only calls Model.save()
```
amangano-edx: The `forward_func` didn't make any modifications to the master object, so I don't see why the `backwards_func` would need to. The `backwards_func` should probably just clear the translations for each master object from the translations table.

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
```diff
@@ -0,0 +1,58 @@
...
+
+def _get_translation(object, SubjectTranslation):
+    translations = SubjectTranslation.objects.filter(master_id=object.pk)
+    try:
+        # Try default translation
+        return translations.get(language_code=settings.LANGUAGE_CODE)
+    except ObjectDoesNotExist:
+        try:
+            # Try default language
+            return translations.get(language_code=settings.PARLER_DEFAULT_LANGUAGE_CODE)
```
amangano-edx: What is `PARLER_DEFAULT_LANGUAGE_CODE` and how is it different from our default language code?

course_discovery/apps/course_metadata/models.py
```diff
@@ -110,15 +111,15 @@ class LevelType(AbstractNamedModel):
     pass
 
 
-class Subject(TimeStampedModel):
+class Subject(TranslatableModel, TimeStampedModel):
```
amangano-edx, line 116: Does mixin order matter here?

course_discovery/apps/course_metadata/models.py
```diff
@@ -127,12 +128,23 @@ def __str__(self):
...
+class SubjectTranslation(TranslatedFieldsModel):
+    master = models.ForeignKey(Subject, related_name='translations', null=True)
+
+    name_t = models.CharField(max_length=255, blank=False, null=False)
+    subtitle_t = models.CharField(max_length=255, blank=True, null=True)
+    description_t = models.TextField(blank=True, null=True)
+
+    class Meta:
+        unique_together = ('language_code', 'master')
+        verbose_name = _("Subject model translations")
```
amangano-edx: nit: inconsistent quote usage.

course_discovery/settings/base.py
```diff
@@ -108,9 +109,22 @@
 }
 
 # Internationalization
-# https://docs.djangoproject.com/en/dev/topics/i18n/
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#language-code
+# http://www.i18nguy.com/unicode/language-identifiers.html
+LANGUAGE_CODE = 'en'
 
-LANGUAGE_CODE = 'en-us'
```
amangano-edx, line 115: We should be careful about making this change, as it might affect how strings are extracted into .po files and made available to transifex.

course_discovery/settings/base.py
```diff
@@ -108,9 +109,22 @@
...
+PARLER_DEFAULT_LANGUAGE_CODE = LANGUAGE_CODE
+
+PARLER_LANGUAGES = {
```
BenjiLee, line 120: This should be moved to edx configuration repo

course_discovery/settings/devstack.py
```diff
@@ -22,6 +22,8 @@
 COMPRESS_OFFLINE = False
 COMPRESS_ENABLED = False
 
+LANGUAGE_CODE = 'en'
```
BenjiLee: This will be removed. Temporarily placed for devstack usage

BenjiLee:
@amangano-edx I tried to name the fields in the new table to the same ones in the original table. SQL allows this, but Django parler throws an error because it's sharing the same field name.

course_discovery/apps/course_metadata/models.py
BenjiLee: I don't believe it should. Both models have super's on their init method and do not share method namespace.

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
BenjiLee: Addressed

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
BenjiLee: It's the default language code that parler uses. By default it is the same as the LANGAUGE CODE but I think it's better to be explicit about it since we are using it for parler.

amangano-edx:
@BenjiLee do you have any insight as to why parlor doesn't allow the field names to be the same?

course_discovery/apps/course_metadata/migrations/0058_auto_20170919_1927.py
BenjiLee: I don't believe there will be any queries for the language code specifically on the Subject Translations table. The query accesses the Subjects model, when then references the Subjects Translations model by ID.

BenjiLee:
@amangano-edx It's probably because of the usability we gain from Parler. It allows us to reference the SubjectTranslations model directly from the Subjects model. 

Example:
```
>>> subject = Subject.objects.all()[0]
>>> subject.get_current_language() 
'en'
>>> subject.name
'english name in subjects table'
>>> subject.name_t
'english name in subjects translations table'
>>> subject.set_current_language('es') 
>>> subject.name
'spanish name in subjects table'
>>> subject.name_t
'spanish name in subjects translations table'
```

course_discovery/apps/course_metadata/migrations/0058_auto_20170919_1927.py
```diff
@@ -0,0 +1,46 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='SubjectTranslation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('language_code', models.CharField(db_index=True, max_length=15, verbose_name='Language')),
+                ('name_t', models.CharField(max_length=255)),
+                ('subtitle_t', models.CharField(blank=True, max_length=255, null=True)),
+                ('description_t', models.TextField(blank=True, null=True)),
+                ('master', models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='translations', to='course_metadata.Subject')),
```
bderusha: I'm assuming`master` is a parler default field name?

course_discovery/apps/course_metadata/migrations/0058_auto_20170919_1927.py
```diff
@@ -0,0 +1,46 @@
...
+            ],
+            options={
+                'verbose_name': 'Subject model translations',
+            },
+        ),
+        migrations.AlterField(
+            model_name='subject',
+            name='slug',
+            field=django_extensions.db.fields.AutoSlugField(blank=True, editable=False,
```
bderusha: Will this do the right thing?  Since the `name_t` field lives in another table and now has multiple values per "master" record?

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
```diff
@@ -0,0 +1,60 @@
...
+
+def forwards_func(apps, schema_editor):
+    Subject = apps.get_model('course_metadata', 'Subject')
+    SubjectTranslation = apps.get_model('course_metadata', 'SubjectTranslation')
+
+    for subject in Subject.objects.all():
+
+        SubjectTranslation.objects.create(
+            master_id=subject.pk,
+            language_code=settings.LANGUAGE_CODE,
```
bderusha: NIT: should this be the `PARLER_DEFAULT_LANGUAGE_CODE`

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
```diff
@@ -0,0 +1,60 @@
...
+        # Try default translation
+        return translations.get(language_code=settings.LANGUAGE_CODE)
+    except ObjectDoesNotExist:
+        try:
+            # Try default language
+            return translations.get(language_code=settings.PARLER_DEFAULT_LANGUAGE_CODE)
+        except ObjectDoesNotExist:
+            # Maybe the object was translated only in a specific language?
+            # Hope there is a single translation
+            return translations.get()
```
bderusha: If no translations exist for some reason this migration will not be able to run.  Should probably put in another `try...except` block and log an error if this happens but let the migration continue

course_discovery/apps/course_metadata/models.py
```diff
@@ -127,11 +129,28 @@ def __str__(self):
...
-            ('partner', 'name'),
             ('partner', 'slug'),
             ('partner', 'uuid'),
         )
 
+    def validate_unique(self, *args, **kwargs):
+        super(Subject, self).validate_unique(*args, **kwargs)
+        qs = Subject.objects.filter(partner=self.partner_id)
+        if qs.filter(translations__name_t=self.name_t).exists():
+            raise ValidationError({'name_t': ['Subject with this Name and Parger already exists', ]})
```
bderusha: SpellCheck:  `Parger` VS `Partner` :)

course_discovery/settings/base.py
```diff
@@ -110,9 +111,22 @@
...
+    1: (
+        {'code': 'en', },
+        {'code': 'es', },
```
bderusha: Makes sense to have the default be English in this repo...but probably not Spanish

course_discovery/settings/base.py
```diff
@@ -110,9 +111,22 @@
...
+    ),
+    'default': {
+         'fallbacks': ['en'],  # defaults to PARLER_DEFAULT_LANGUAGE_CODE
```
bderusha: do we just want to use the default here then?

bderusha (changes requested):

course_discovery/apps/course_metadata/models.py
BenjiLee: Definitely Parger

course_discovery/settings/base.py
BenjiLee: https://github.com/edx/configuration/pull/4108

course_discovery/settings/base.py
BenjiLee: Verified that transifex is using 'en'

course_discovery/apps/course_metadata/migrations/0058_auto_20170919_1927.py
BenjiLee: It will use the default language.

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
BenjiLee: I don't think there is a way to create a subject without the default language. Let me test this

Update: There must exist a translation when creating a subject. I don't think this can return an error. If the translation object exists, there exists a translation.

course_discovery/apps/course_metadata/models.py
```diff
@@ -127,11 +129,28 @@ def __str__(self):
...
 
+    def validate_unique(self, *args, **kwargs):
+        super(Subject, self).validate_unique(*args, **kwargs)
+        qs = Subject.objects.filter(partner=self.partner_id)
+        if qs.filter(translations__name_t=self.name_t).exists():
+            raise ValidationError({'name_t': ['Subject with this Name and Partner already exists', ]})
+
+
+class SubjectTranslation(TranslatedFieldsModel):
+    master = models.ForeignKey(Subject, related_name='translations', null=True)
```
bderusha, line 145: Prefer `subject` to `master` as best practice

BenjiLee: Django-parler requires that we use the keyword `master`

bderusha: gross 👍 :shipit:

course_discovery/settings/base.py
BenjiLee: removed here. added in the configuration repo

course_discovery/settings/base.py
BenjiLee: fixed

course_discovery/apps/course_metadata/migrations/0059_make_translations_table.py
BenjiLee: updated

bderusha (approved):
:shipit:

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -948,8 +948,8 @@ def test_data(self):
         serializer = SubjectSerializer(subject)
 
         expected = {
-            'name': subject.name,
-            'description': subject.description,
+            'name': subject.name_t,
+            'description': subject.description_t,
             'banner_image_url': subject.banner_image_url,
             'card_image_url': subject.card_image_url,
             'subtitle': subject.subtitle,
```
amangano-edx: Should this be set to `subject.subtitle_t`?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -220,11 +221,13 @@ class OrganizationAdmin(admin.ModelAdmin):
 
 
 @admin.register(Subject)
-class SubjectAdmin(admin.ModelAdmin):
-    list_display = ('uuid', 'name', 'slug',)
+class SubjectAdmin(TranslatableAdmin):
+    # These fields are excluded here because they will be removed in favor of the translated fields.
+    exclude = ('name', 'subtitle', 'description')
```
amangano-edx, line 226: Are we able to set `name_t`, `subtitle_t`, and `description_t`?

course_discovery/apps/course_metadata/migrations/0061_make_translations_table.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
```
amangano-edx: Nit: The names of these two migration files are confusing. This one is called `_make_translations_table`, but it's actually just filling in the existing table with data. The other migration is called '_update_subject_models', but it's actually creating a completely new model/index.

course_discovery/apps/course_metadata/migrations/0061_make_translations_table.py
```diff
@@ -0,0 +1,52 @@
...
+def backwards_func(apps, schema_editor):
+    Subject = apps.get_model('course_metadata', 'Subject')
+    SubjectTranslation = apps.get_model('course_metadata', 'SubjectTranslation')
+
+    for subject in Subject.objects.all():
+        translation = _get_translation(subject, SubjectTranslation)
+        subject.name = translation.name_t
+        subject.subtitle = translation.subtitle_t
+        subject.description = translation.description_t
+        subject.save()  # Note this only calls Model.save()
```
amangano-edx: I don't see why we're doing anything to the original subjects table in the backwards function. The original subjects table wasn't touched in the forwards function at all, only the translations table was. We should be reversing whatever was done to the translations table in the forwards_func here.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
```diff
@@ -0,0 +1,47 @@
...
+    dependencies = [
+        ('course_metadata', '0059_auto_20171002_1705'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='SubjectTranslation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('language_code', models.CharField(db_index=True, max_length=15, verbose_name='Language')),
```
amangano-edx: Just curious, but why did you decide to add an index here? I thought we came to the conclusion that an index wouldn't make sense for this field since there would not be much data in this table.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
```diff
@@ -0,0 +1,47 @@
...
+                ('name_t', models.CharField(max_length=255)),
+                ('subtitle_t', models.CharField(blank=True, max_length=255, null=True)),
```
amangano-edx: Any chance of subtitle being > 255 characters?

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
```diff
@@ -0,0 +1,47 @@
...
+        ('course_metadata', '0059_auto_20171002_1705'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='SubjectTranslation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('language_code', models.CharField(db_index=True, max_length=15, verbose_name='Language')),
+                ('name_t', models.CharField(max_length=255)),
```
amangano-edx: Why can't name_t be blank/null?

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
```diff
@@ -0,0 +1,47 @@
...
+                ('subtitle_t', models.CharField(blank=True, max_length=255, null=True)),
+                ('description_t', models.TextField(blank=True, null=True)),
+                ('master',
+                 models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='translations',
```
amangano-edx: Should we disallow nulls here? It would be weird to have a translation with no master subject.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
```diff
@@ -0,0 +1,47 @@
...
+            options={
+                'verbose_name': 'Subject model translations',
+            },
+        ),
+        migrations.AlterField(
+            model_name='subject',
+            name='slug',
+            field=django_extensions.db.fields.AutoSlugField(blank=True, editable=False,
+                                                            help_text='Leave this field blank to have the value generated automatically.',
+                                                            populate_from='name_t'),
```
amangano-edx: Shouldn't the slug be based on the canonical name instead of the translated name?

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        migrations.AlterUniqueTogether(
+            name='subject',
+            unique_together=set([('partner', 'slug'), ('partner', 'uuid')]),
```
amangano-edx: Why do we need to change the constraints on the original table?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -6,20 +6,22 @@
...
 from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.core.utils import SearchQuerySetWrapper
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
     FAQ, AbstractMediaModel, AbstractNamedModel, AbstractValueModel, CorporateEndorsement, Course, CourseRun,
-    Endorsement, Seat, SeatType
-)
+    Endorsement, Seat, SeatType,
+    Subject)
```
amangano-edx: Nit: line 23 seems really short compared to line 22, so I think we can probably just leave subject on line 23.

course_discovery/apps/api/tests/test_serializers.py
BenjiLee: Will fix

course_discovery/apps/course_metadata/admin.py
BenjiLee: We are

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: This was automatically created, but I will take a look.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: Maybe, but I think it should intentionally be short since description is also available. If anything I would argue that it should be < 255 characters. I usually see these types of fields used in marketing spot where you can only fit x amount of characters. 

Examples:
meta description when showing up in a google search is ~ 160 char
android where you only have 80 characters to describe your app before needing to open up the full description

Not sure if where subtitle is or ever will be displayed but thats how I expect it to be used.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: You need to have a name_t in order to create the subject.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: This was suggested by their documentation. I'll look into this.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: The translated name will eventually be the canonical name. If we used the subject.name instead of subject.name_t, when a subject is created, the slug will be 'null'. In the clean up phase, it will be changed back to name from name_t.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: These constraints are the ones that are being kept. The one we are dropping is partner<->name_t which now exists on different tables. The dropped constraints are validated via an overriden `validate_unique` method.

course_discovery/apps/course_metadata/tests/test_models.py
BenjiLee: will do. Auto imported wanted it this way. If it fails isort, i will set it to as is.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
amangano-edx: Ok, just making sure we've thought about that, as the DB would limit this field to 255 characters based on the current settings.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
amangano-edx: Hmm, does this complicate things from a subject-creation standpoint? Doesn't it mean that the subject-translation will need to be created and saved before the master, as otherwise name_t wouldn't be populated yet?

BenjiLee: That's the magic of parler. In the admin, when creating a subject, the fields shown to you are combined from the Subjects and Subjects translations model. It looks like the original  subjects form.

course_discovery/apps/course_metadata/migrations/0061_make_translations_table.py
BenjiLee: updated.

course_discovery/apps/course_metadata/migrations/0061_make_translations_table.py
BenjiLee: After this PR is merged, when a user creates a subject, the data goes in the subject translations model and no longer the original Subjects model. When migrating backwards, it will take the default language and set the original subject models now deprecated fields.

Example:

Before PR:
subject.name -> The name of the subject

After PR:
subject.name -> No longer used
subject.name_t -> The new required field.

When creating a new subject, subject.name does not get populated anymore, instead subject.name_t is populated. When migrating backwards, subject.name_t will be migrated in the backwards_func() to subject.name to preserve any new subjects that were created after implementing this code.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: Looks like it's explicitly set in the django-parler package. Leaving this as is for now.

course_discovery/apps/course_metadata/migrations/0060_update_subject_models.py
BenjiLee: I agree with you and I'm not sure why it's in the documentation so I asked a question https://github.com/django-parler/django-parler/issues/196

course_discovery/apps/course_metadata/migrations/0061_migrate_subjects_data.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+
+
+def backwards_func(apps, schema_editor):
+    Subject = apps.get_model('course_metadata', 'Subject')
+    SubjectTranslation = apps.get_model('course_metadata', 'SubjectTranslation')
+
+    for subject in Subject.objects.all():
+        try:
+            translation = SubjectTranslation.objects.filter(master_id=subject.pk)
```
amangano-edx, line 34: Should we scope this to just the `PARLER_DEFAULT_LANGUAGE_CODE`? What if a subject has more than one translation?

course_discovery/apps/course_metadata/migrations/0061_migrate_subjects_data.py
```diff
@@ -0,0 +1,51 @@
...
+            subject.name = translation.name_t
+            subject.subtitle = translation.subtitle_t
+            subject.description = translation.description_t
+            subject.save()  # Note this only calls Model.save()
+        except ObjectDoesNotExist:
+            # nothing to migrate
+            logger.exception('Migrating data from SubjectTranslation for master_id={} DoesNotExist')
```
amangano-edx: I think you forgot to call `format` on this string.

course_discovery/apps/course_metadata/models.py
```diff
@@ -128,11 +130,31 @@ def __str__(self):
 
     class Meta:
         unique_together = (
-            ('partner', 'name'),
             ('partner', 'slug'),
             ('partner', 'uuid'),
         )
 
+    def validate_unique(self, *args, **kwargs):
```
amangano-edx, line 137: Note that this might not run in some cases, such as when loading data directly into the database via mySql

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -25,7 +25,7 @@ def test_model_change(self, mock_set_api_timestamp):
         # connecting to. We want to test each of them.
         for model in apps.get_app_config('course_metadata').get_models():
             # Ignore models that aren't exposed by the API or are only used for testing.
-            if model == DataLoaderConfig or 'abstract' in model.__name__.lower():
+            if model in [DataLoaderConfig, SubjectTranslation] or 'abstract' in model.__name__.lower():
```
amangano-edx, line 28: Why is this necessary?

course_discovery/settings/base.py
amangano-edx: Is this just default config now?

course_discovery/settings/base.py
```diff
@@ -110,9 +111,21 @@
...
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#language-code
+# http://www.i18nguy.com/unicode/language-identifiers.html
+LANGUAGE_CODE = 'en'
 
-LANGUAGE_CODE = 'en-us'
+PARLER_DEFAULT_LANGUAGE_CODE = LANGUAGE_CODE
+
+PARLER_LANGUAGES = {
+    1: (
+        {'code': 'en', },
```
amangano-edx: Can we reference a constant for this?

amangano-edx (approved):
A few more questions and minor stuff. Otherwise LGTM

course_discovery/settings/base.py
BenjiLee: Correct. Since it's a new config, the presence of this config should not affect anything else.

course_discovery/settings/base.py
BenjiLee: Maybe `LANGUAGE_CODE`?

course_discovery/apps/course_metadata/tests/test_signals.py
BenjiLee: This test throws an error for a missing factory.
`SubjectTranslation` does not need a factory for testing.

course_discovery/apps/course_metadata/migrations/0061_migrate_subjects_data.py
BenjiLee: shh...

course_discovery/apps/course_metadata/models.py
BenjiLee: Yep. Well aware of this. There is no database level validation across foreign keys :(

course_discovery/apps/course_metadata/migrations/0061_migrate_subjects_data.py
BenjiLee: Fetching a subject returns you the default by default.

bderusha:
🔥 🎸 🎆

PR #1131, opened 2017-09-20
Removed Course.verification_deadline from Publisher

clintonb:
This field is not used, and makes no sense. Only course runs have verification deadlines.

LEARNER-2621


schenedx (commented):
👍

PR #1132, opened 2017-09-22
Create course about page when course_run saves

schenedx:
Create the course about page on marketing site when course_run saves
This does not include the change for URL redirect manipulation yet. Will add later.
@clintonb When you are available, feel free to review

### DIFF
--- a/course_discovery/apps/course_metadata/publishers.py
+++ b/course_discovery/apps/course_metadata/publishers.py
@@ -201,6 +201,11 @@ class CourseRunMarketingSitePublisher(BaseMarketingSitePublisher):
             node_data = self.serialize_obj(obj)
 
             self.edit_node(node_id, node_data)
+        elif not previous_obj:
+            # This is a brand new course_run object
+            # let's create it on the marketing site
+            node_data = self.serialize_obj(obj)
+            self.create_node(node_data)
 
     def serialize_obj(self, obj):
         """
@@ -217,6 +222,10 @@ class CourseRunMarketingSitePublisher(BaseMarketingSitePublisher):
         return {
             **data,
             'status': 1 if obj.status == CourseRunStatus.Published else 0,
+            'field_course_uuid': str(obj.uuid),
+            'title': obj.title,
+            'field_course_id': obj.key,
+            'type': 'course',
         }
 
 
--- a/course_discovery/apps/course_metadata/tests/test_publishers.py
+++ b/course_discovery/apps/course_metadata/tests/test_publishers.py
@@ -199,16 +199,19 @@ class CourseRunMarketingSitePublisherTests(MarketingSitePublisherTestMixin):
 
         self.obj = CourseRunFactory()
 
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'serialize_obj', return_value='data')
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'create_node')
+    def test_publish_obj_create(self, mock_create_node, *args):  # pylint: disable=unused-argument
+        self.publisher.publish_obj(self.obj)
+        mock_create_node.assert_called_with('data')
+
     @mock.patch.object(CourseRunMarketingSitePublisher, 'node_id', return_value='node_id')
     @mock.patch.object(CourseRunMarketingSitePublisher, 'serialize_obj', return_value='data')
     @mock.patch.object(CourseRunMarketingSitePublisher, 'edit_node', return_value=None)
-    def test_publish_obj(self, mock_edit_node, *args):  # pylint: disable=unused-argument
+    def test_publish_obj_edit(self, mock_edit_node, *args):  # pylint: disable=unused-argument
         """
         Verify that the publisher attempts to publish when course run status changes.
         """
-        # No previous object. No editing should occur.
-        self.publisher.publish_obj(self.obj)
-        assert not mock_edit_node.called
 
         # A previous object is provided, but the status hasn't changed.
         # No editing should occur.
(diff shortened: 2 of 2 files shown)


course_discovery/apps/course_metadata/publishers.py
```diff
@@ -201,6 +201,11 @@ def publish_obj(self, obj, previous_obj=None):
             node_data = self.serialize_obj(obj)
 
             self.edit_node(node_id, node_data)
+        elif not previous_obj:
+            # This is a brand new course_run object
+            # let's create it on the marketing site
+            node_data = self.serialize_obj(obj)
+            self.create_node(node_data)
 
     def serialize_obj(self, obj):
```
clintonb, line 210: s/obj/course_run

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -217,6 +222,10 @@ def serialize_obj(self, obj):
         return {
             **data,
             'status': 1 if obj.status == CourseRunStatus.Published else 0,
+            'field_course_uuid': str(obj.uuid),
+            'title': obj.title_override or obj.course.title,
```
clintonb: Use the `title` property: `course_run.title`

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -199,16 +199,20 @@ def setUp(self):
 
         self.obj = CourseRunFactory()
 
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'serialize_obj', return_value='data')
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'create_node', return_value='node_id')
+    def test_publish_obj_create(self, mock_create_node, *args):  # pylint: disable=unused-argument
+        # No previous object. creation should occur
```
clintonb: This comment is not necessary.

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -199,16 +199,20 @@ def setUp(self):
 
         self.obj = CourseRunFactory()
 
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'serialize_obj', return_value='data')
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'create_node', return_value='node_id')
+    def test_publish_obj_create(self, mock_create_node, *args):  # pylint: disable=unused-argument
```
clintonb: If `*args` is not used, why is it here?

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -199,16 +199,20 @@ def setUp(self):
 
         self.obj = CourseRunFactory()
 
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'serialize_obj', return_value='data')
+    @mock.patch.object(CourseRunMarketingSitePublisher, 'create_node', return_value='node_id')
```
clintonb: Is the `return_value` kwarg necessary given that you aren't using it?

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -199,16 +199,20 @@ def setUp(self):
...
+    def test_publish_obj_create(self, mock_create_node, *args):  # pylint: disable=unused-argument
+        # No previous object. creation should occur
+        self.publisher.publish_obj(self.obj)
+        mock_create_node.assert_called_with('data')
+
     @mock.patch.object(CourseRunMarketingSitePublisher, 'node_id', return_value='node_id')
     @mock.patch.object(CourseRunMarketingSitePublisher, 'serialize_obj', return_value='data')
     @mock.patch.object(CourseRunMarketingSitePublisher, 'edit_node', return_value=None)
-    def test_publish_obj(self, mock_edit_node, *args):  # pylint: disable=unused-argument
+    def test_publish_obj_edit(self, mock_edit_node, *args):  # pylint: disable=unused-argument
```
clintonb, line 211: Same as above. Get rid of `*args`.

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -231,8 +235,11 @@ def test_serialize_obj(self):
         actual = self.publisher.serialize_obj(self.obj)
         expected = {
             'field_course_id': self.obj.key,
+            'field_course_uuid': str(self.obj.uuid),
+            'title': self.obj.title_override or self.obj.course.title,
             'author': {'id': self.user_id},
             'status': 1,
+            'type': 'course'
```
clintonb: Add a trailing comma.

course_discovery/apps/course_metadata/publishers.py
```diff
@@ -217,6 +222,10 @@ def serialize_obj(self, obj):
...
+            'field_course_id': obj.key,
+            'type': 'course'
```
clintonb: Add a trailing comma.

course_discovery/apps/course_metadata/publishers.py
schenedx: The method was inherited from the super class above called `BaseMarketingSitePublisher`. Updating it to course_run would defeat the purpose of multiple inheritance here

course_discovery/apps/course_metadata/tests/test_publishers.py
schenedx: If we get rid of *args, unit tests would fail because the test would try to call the `test_publish_obj_edit` with 4 arguments, but the function would only accept 2 arguments. So the `*args` are used to catch those extra arguments

PR #1136, opened 2017-09-25
Create the correct URL for course about page during creation

schenedx:
Also, do not create course about page on Drupal if it is already created.
@clintonb Please review


course_discovery/apps/course_metadata/publishers.py
```diff
@@ -228,6 +361,12 @@ def serialize_obj(self, obj):
             'type': 'course',
         }
 
+    def get_marketing_slug(self, obj):
+        return slugify(obj.title)
```
clintonb: `CourseRun` instances have a `slug` field.

clintonb (changes requested):
Make sure you use the appropriate slug field.

clintonb (approved):

PR #1137, opened 2017-09-26
Updated date publishing for E-Commerce API calls

clintonb:
- Verification deadline now defaults to the course end date
- Product expiration date (upgrade deadline) now defaults to course a few days before the course end date

LEARNER-2621

### DIFF
--- a/course_discovery/apps/publisher/api/utils.py
+++ b/course_discovery/apps/publisher/api/utils.py
@@ -1,0 +1,28 @@
+import datetime
+
+from django.conf import settings
+
+from course_discovery.apps.core.utils import serialize_datetime
+from course_discovery.apps.publisher.models import Seat
+
+
+def serialize_seat_for_ecommerce_api(seat):
+    upgrade_deadline = seat.upgrade_deadline
+    if not upgrade_deadline:
+        upgrade_deadline = seat.course_run.end - datetime.timedelta(days=settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
+
+    return {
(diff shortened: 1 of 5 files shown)


course_discovery/settings/base.py
```diff
@@ -501,6 +501,10 @@
 
 PUBLISHER_FROM_EMAIL = None
 
+# If no upgrade deadline is specified for a course run seat, when the course is published the deadline will default to
+# the course run end date minus the specified number of days.
+PUBLISHER_UPGRADE_DEADLINE_DAYS = 10
```
schenedx, line 506: This is not used in the PR. Maybe you should use it in the `utils.py` file?

course_discovery/apps/publisher/api/utils.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+
+from course_discovery.apps.core.utils import serialize_datetime
+from course_discovery.apps.publisher.models import Seat
+
+
+def serialize_seat_for_ecommerce_api(seat):
+    upgrade_deadline = seat.upgrade_deadline or (seat.course_run.end - datetime.timedelta(days=3))
```
schenedx: What if the `seat.course_run.end` is `None`, like in the case of self-paced course? How is that case handled

schenedx (changes requested):
Got 2 questions.

course_discovery/apps/publisher/api/utils.py
clintonb: This field is required.

schenedx (approved):
👍

PR #1139, opened 2017-09-26
Updated enrollment date publishing to Studio

clintonb:
- When creating a new course run, enrollment dates are no longer copied from the previous run
- Temporary code in signal receiver has been removed. The enrollment dates will always be empty when creating a re-run.
- Added additional exception handling around API call to create images. Errors from this operation will no longer prevent a course run from being saved.

LEARNER-2469

### DIFF
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -250,8 +250,11 @@ class Course(TimeStampedModel, ChangedByMixin):
 
     @cached_property
     def discovery_counterpart(self):
-        course_key = '{org}+{number}'.format(org=self.organizations.first().key, number=self.number)
-        return DiscoveryCourse.objects.get(partner=self.partner, key=course_key)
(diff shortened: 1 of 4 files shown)


course_discovery/apps/publisher/models.py
```diff
@@ -250,8 +250,11 @@ def course_title(self):
 
     @cached_property
     def discovery_counterpart(self):
-        course_key = '{org}+{number}'.format(org=self.organizations.first().key, number=self.number)
-        return DiscoveryCourse.objects.get(partner=self.partner, key=course_key)
+        return DiscoveryCourse.objects.get(partner=self.partner, key=self.key)
+
+    @cached_property
+    def key(self):
```
schenedx, line 256: Is this property created to help simplify unit testing?

schenedx (approved):
Just one question. 👍

course_discovery/apps/publisher/models.py
clintonb: It's also used in production code. I got tired of computing it.

PR #1140, opened 2017-09-27
updated people api create to use group authentication

asadazam93:
### Description
https://github.com/edx/course-discovery/blob/fa251757dae29dbff2380728790116be9a498fe2/course_discovery/apps/api/v1/views/people.py#L26
previously this endpoint allowed any authenticated user to create a new person entity.
I have added one more method of authentication to create a new person. i.e group. The authenticated user has to be part of any group to be able to create a new person. As course team users of the publisher are the people who mostly use this endpoint and they are neither staff nor superusers so using both these permissions was not going be sufficient. The course team users are part of a group so I have used groups as a method of authentication (any non verified or non staff member is not a part of any group)

**Sandbox**
- N/A

### Reviewers
- [x] @awaisdar001  
- [x] @attiyaIshaque  

### Post-review

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_people.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_people.py
@@ -1,6 +1,7 @@
 # pylint: disable=redefined-builtin,no-member
 import ddt
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
 from django.db import IntegrityError
 from mock import mock
 from rest_framework.reverse import reverse
@@ -15,6 +16,8 @@ from course_discovery.apps.course_metadata.models import Person
 from course_discovery.apps.course_metadata.people import MarketingSitePeople
 from course_discovery.apps.course_metadata.tests import toggle_switch
 from course_discovery.apps.course_metadata.tests.factories import OrganizationFactory, PersonFactory, PositionFactory
+from course_discovery.apps.publisher.constants import INTERNAL_USER_GROUP_NAME
+from course_discovery.apps.publisher.permissions import logger as permission_logger
 
 User = get_user_model()
 
@@ -27,6 +30,7 @@ class PersonViewSetTests(SerializationMixin, SiteMixin, APITestCase):
     def setUp(self):
         super(PersonViewSetTests, self).setUp()
         self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.user.groups.add(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))
         self.client.force_authenticate(self.user)
         self.person = PersonFactory(partner=self.partner)
         self.organization = OrganizationFactory(partner=self.partner)
@@ -110,6 +114,23 @@ class PersonViewSetTests(SerializationMixin, SiteMixin, APITestCase):
         response = self.client.post(self.people_list_url, {}, format='json')
         self.assertEqual(response.status_code, 403)
         self.assertEqual(Person.objects.count(), 0)
+
+    def test_create_without_group(self):
+        """ Verify group is required when creating a person. """
+        self.user.groups.remove(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))
+        current_people_count = Person.objects.count()
+
+        with LogCapture(permission_logger.name) as log_capture:
+            response = self.client.post(self.people_list_url, {}, format='json')
+            self.assertEqual(response.status_code, 403)
+            self.assertEqual(Person.objects.count(), current_people_count)
+            log_capture.check(
+                (
+                    permission_logger.name,
+                    'INFO',
+                    'Permission denied. User [{}] has no groups'.format(self.user.username),
+                )
+            )
 
     def test_get(self):
         """ Verify the endpoint returns the details for a single person. """
--- a/course_discovery/apps/publisher/permissions.py
+++ b/course_discovery/apps/publisher/permissions.py
@@ -1,0 +1,16 @@
+import logging
+
+from rest_framework import permissions
+
+logger = logging.getLogger(__name__)
+
+
+class UserHasGroup(permissions.BasePermission):
+    """
+    Global permission to check if request.user has any group
+    """
+    def has_permission(self, request, view):
+        if request.user.groups.all():
+            return True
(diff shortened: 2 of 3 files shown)


course_discovery/apps/api/decorators.py
```diff
@@ -0,0 +1,18 @@
+import logging
+
+from rest_framework import status
+from rest_framework.response import Response
+
+logger = logging.getLogger(__name__)
+
+
+def checkgroup(create_func):
```
attiyaIshaque: Please try finding a more appropriate name for this method. Maybe like `checkusergroup`.

course_discovery/apps/api/decorators.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+     Modifies a create function such that if the request.user has no groups it denies the creation of object
+    """
+    def _checkgroup_wrapper(self, request, *args, **kwargs):
+        if request.user.groups.all():
+            return create_func(self, request, *args, **kwargs)
+        logger.exception('Permission denied. User has no groups')
```
attiyaIshaque: Please add username in the logger.exception.

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -27,6 +29,7 @@ class PersonViewSetTests(SerializationMixin, SiteMixin, APITestCase):
     def setUp(self):
         super(PersonViewSetTests, self).setUp()
         self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.user.groups.add(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))
```
attiyaIshaque, line 33: you can do it in this way
```
group = Group.objects.get(name=INTERNAL_USER_GROUP_NAME)
```
and also use it here `self.user.groups.remove(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))`

attiyaIshaque (changes requested):

clintonb (changes requested):
These permissions are too permissive. Users should be part of a specific group/groups, not just any group. Be explicit about who is allowed to access the endpoint. Also, use a proper [permissions class](http://www.django-rest-framework.org/api-guide/permissions/), not a custom decorator.

asadazam93:
@clintonb the course team users can be a part of any group (150+ groups exist and more groups are being added with time) so I am not sure how can we use specific groups. Please share your thoughts. And I will replace the decorator with a custom permission class as soon as this groups issue is resolved.

clintonb:
Find some commonality in the groups. Associate them with an organization, set a common prefix (not a great idea). Do something that is more secure and properly uses Django permissions.

asadazam93:
@clintonb one possible solution that I have in mind is that we create a new group (say `course_team`) and assign all course team users to this group. This way we will be able to identify course team users and set a custom permission class for that group. 

Currently there is no simpler way to identify a course team member but the fact that they are assigned to some group (the group list is a long mutable list with no commonality so we can not be sure of exactly which group)

clintonb:
The new group should be called "Course Creators". How do you intend to keep this group updated?

asadazam93:
@clintonb we will firstly need to write a management to add the current course team users to this new group and then all new course team users will also need to be added to this group.It would be more secure but at the same time it would complicate the solution. I was going for a more simpler solution by checking if the user is part of any group because a user which is a part of a group is in almost all cases a staff member or some internal user. And since only staff members or superusers can add or change group it seems to be secure enough. What are your thoughts about this?

clintonb:
My thoughts are that we should go for the correct solution when it comes to security. There is no telling what will change in the future regarding groups and who is in the groups. Permissions should be explicit, not implicit.

schenedx:
@clintonb Asking the escalation team to solve the mess we have with Publisher Permissions is too huge of an undertaking in this context, in my opinion. 
We had a conversation about this in our Educator Triage meeting. We will create a ticket to track the need of your proposal. Shelby will schedule that. We will just take the easiest route here to fix the permission problem. If you have concerns on this plan, please come and talk with Scott and me.

course_discovery/apps/publisher/permissions.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class UserHasGroup(permissions.BasePermission):
+    """
+    Global permission to check if request.user has any group
+    """
+    def has_permission(self, request, view):
+        if request.user.groups.all():
+            return True
+        logger.exception('Permission denied. User [%s] has no groups', request.user.username)
```
awaisdar001: maybe logger.info or error?

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -111,6 +114,15 @@ def test_create_without_authentication(self):
         self.assertEqual(response.status_code, 403)
         self.assertEqual(Person.objects.count(), 0)
 
+    def test_create_without_group(self):
+        """ Verify group is required when creating a person. """
+        self.user.groups.remove(Group.objects.get(name=INTERNAL_USER_GROUP_NAME))
+        Person.objects.all().delete()
```
awaisdar001: Why you are deleting the objects?

awaisdar001 (changes requested):

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -111,6 +115,23 @@ def test_create_without_authentication(self):
         self.assertEqual(response.status_code, 403)
         self.assertEqual(Person.objects.count(), 0)
 
+    def test_create_without_group(self):
+        """ Verify group is required when creating a person. """
```
awaisdar001, line 119: Only one nit in the docstring. looks good to me.  https://openedx.atlassian.net/wiki/spaces/EDUCATOR/pages/251528559/Coding+Practices

awaisdar001 (approved):

attiyaIshaque (approved):
Looks good to me. 👍

schenedx:
The changes here blocked marketing site from accessing instructors API.
Updated the implementation of this on https://github.com/edx/course-discovery/pull/1207

PR #1141, opened 2017-09-27
Add a waffle switch so we can dynamically control the auto creation

schenedx:
Add the waffle flag to stop or start the auto-creation of course about page on marketing site

@clintonb Please review

### DIFF
--- a/course_discovery/apps/course_metadata/migrations/0058_auto_course_about_page_creation_switch.py
+++ b/course_discovery/apps/course_metadata/migrations/0058_auto_course_about_page_creation_switch.py
@@ -1,0 +1,22 @@
+from django.db import migrations
+
+
+def create_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.get_or_create(name='auto_course_about_page_creation', defaults={'active': False})
+
+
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/migrations/0058_auto_course_about_page_creation_switch.py
```diff
@@ -0,0 +1,24 @@
+from django.db import migrations
+
+
+def create_switch(apps, schema_editor):
+    """Create the enable_publisher_email_notifications switch if it does not already exist."""
```
clintonb: These docstrings are wrong. Rather than fix them, get rid of them. We know what the functions do.

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -216,6 +222,7 @@ def test_publish_obj_create_successful(
         mock_create_node,
         *args
     ):  # pylint: disable=unused-argument
+        toggle_switch('auto_course_about_page_creation', True)
```
clintonb, line 225: FYI: Waffle has a decorator, `@override_switch()`, that does the same thing.

clintonb (approved):

PR #1142, opened 2017-09-27
Publisher Updates

clintonb:


clintonb:
Dependent on https://github.com/edx/course-discovery/pull/1143

clintonb:
I have removed a couple test cases because it's not clear that they provide any value. Sorting is handled by a third-party library. There is no need to test it to such extreme levels.

course_discovery/apps/publisher/templates/publisher/email/studio_instance_needed.txt
```diff
@@ -1,18 +0,0 @@
-{% load i18n %}
```
schenedx, line 1: Don't we still need this message for ASU? Or for those organization which do not want auto studio instance creation?

schenedx (commented):
Just one question, do we want to delete those studio emails?

course_discovery/apps/publisher/templates/publisher/email/studio_instance_needed.txt
clintonb: This template was never used by any code. It seems to be have been created unnecessarily.

PR #1143, opened 2017-09-27
Updated Publisher admin and course list view

clintonb:
- Updated Django admin to show history for Publisher models
- Added course number to course list view


schenedx (approved):
👍

PR #1144, opened 2017-09-28
Split Travis tasks into multiple calls

clintonb:
This separation allows us to assess how much time is consumed by each operation in our build. This data can be used to tune the build and seek out areas for improvement.


clintonb:
FYI @edx/learner this brings discovery in line with ecommerce, allowing us to get more granular data on times for each step of the build. Currently, it is obvious that testing takes the most time; however, if we notice increased times elsewhere, we now have the ability to find out where.

Makefile
```diff
@@ -32,18 +32,19 @@ clean: ## Delete generated byte code and coverage reports
...
 production-requirements: ## Install Python and JS requirements for production
 	pip install -r requirements.txt
 	npm install --production
-	$(NODE_BIN)/bower install --production
+	$(NODE_BIN)/bower install --allow-root --production
 
 test: clean ## Run tests and generate coverage report
-	coverage run -m pytest --durations=25
+	## The node_modules .bin directory is added to ensure we have access to Geckodriver.
+	PATH="$(NODE_BIN):$(PATH)" coverage run -m pytest --durations=25
```
schenedx, line 47: Nice. I always wondered why my browser test doesn't work because the lack of `Geckodriver`

schenedx (approved):
👍

PR #1151, opened 2017-10-02
Updated Publisher to push E-Commerce seat info to Discovery

clintonb:
This ensures we don't have to wait for a data refresh to pull data from
the E-Commerce Service.

LEARNER-2719


schenedx (approved):
👍

PR #1152, opened 2017-10-02
Updated refresh metadata command to pull course outcome from Drupal node

clintonb:
LEARNER-2463


course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -373,53 +373,6 @@ def test_create_course_run_and_seat_with_errors(self):
                     )
                 )
 
-    def test_create_course_run_and_seat(self):
```
schenedx: Why was this test case deleted? Seems unrelated, right?

schenedx: I see, you are trying to delete a flaky test.
Why the unit tests are broken by this change?

PR #1154, opened 2017-10-03
Added synonyms for Global Freshman Academy

Ayub-Khan:
**[LEARNER-2579](https://openedx.atlassian.net/browse/LEARNER-2579)** : 
**Description**:
As a learner, when I search "GFA" or "Global Freshman Academy", I expect to see all courses provided by the institution ASUx that are credit-eligible. These would be the same as the courses that appear on the GFA landing page www.edx.org/gfa.
the reason GFA & Global Freshman Academy doesn't return in search results today, is that those words aren't present in the course description, or title.
To get the behavior that Kathleen is requesting in her second point, the easiest would be to modify the course description to include the words "GFA" & "Global Freshman Academy", in addition to setting up a synonym for GFA = Global Freshman Academy.
Some data, in the past year, ~15 million unique searches were performed on our site. Global Freshman Academy was searched for 242 times and GFA was searched for 360 times.

### DIFF
--- a/course_discovery/settings/synonyms.py
+++ b/course_discovery/settings/synonyms.py
@@ -9,6 +9,7 @@ SYNONYMS = [
     ['ASU', 'ASUx', 'Arizona State', 'Arizona State University'],
     ['Berkeley', 'UC BerkeleyX', 'UCBerkeleyX'],
     ['Georgia Institute of Technology', 'Georgia Tech', 'GTx'],
+    ['GFA', 'Global Freshman Academy'],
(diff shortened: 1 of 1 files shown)


schenedx:
@Ayub-Khan You should rebase with master to fix your build issue

schenedx:
@Ayub-Khan Why close this PR?
I think we can still use this synonym addition. 
But this is a useless change until we figure out the ES index issue.

Ayub-Khan:
@schenedx  im going to reopen. ES index issue is resolved. we can move forward with adding synonym. I thought it would be useless if course already have gfa and global freshman academy in name

schenedx (approved):
👍

PR #1156, opened 2017-10-03
Remove video_translation_languages field from course_run object. 

schenedx:
The field name is not accurate and the data is not being used.
@clintonb Please help review
Will create the PR for the migration file after this is merged.

### DIFF
--- a/course_discovery/apps/publisher/api/v1/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/v1/tests/test_views.py
@@ -153,7 +153,6 @@ class CourseRunViewSetTests(APITestCase):
         assert discovery_course_run.learner_testimonials == publisher_course.learner_testimonial
         expected = set(publisher_course_run.transcript_languages.all())
         assert set(discovery_course_run.transcript_languages.all()) == expected
-        assert set(discovery_course_run.video_translation_languages.all()) == {publisher_course_run.video_language}
         assert set(discovery_course_run.staff.all()) == set(publisher_course_run.staff.all())
 
         assert discovery_course.canonical_course_run == discovery_course_run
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -406,8 +406,6 @@ class CourseRun(TimeStampedModel):
     syllabus = models.ForeignKey(SyllabusItem, default=None, null=True, blank=True)
     card_image_url = models.URLField(null=True, blank=True)
     video = models.ForeignKey(Video, default=None, null=True, blank=True)
-    video_translation_languages = models.ManyToManyField(
```
clintonb: You can’t get rid of the field, yet. That comes in the second/last PR.

clintonb (changes requested):
Revert the model change.

course_discovery/apps/course_metadata/models.py
schenedx: @clintonb My understanding is that, we need to remove this from the models.py first, then create the migration file on the second PR.
See a recent example at https://github.com/edx/edx-platform/pull/15863/files and https://github.com/edx/edx-platform/pull/15911/files

clintonb: Those were wrong. The committed models and migrations should always be in sync. Remove all references to the field, except the field itself first. After that, remove the field in both code and migration. See https://engineering.edx.org/django-migration-donts-f4588fd11b64.

schenedx: Hmm... OK.. I'll do what you recommended, and let's see.

clintonb (approved):
Squash your commits.

PR #1157, opened 2017-10-03
Course images pushed to Studio only if they exist

clintonb:
If a course has no associated image, a warning will be logged rather than attempting to access a non-existent file.

EDUCATOR-1462


schenedx (approved):
👍

PR #1158, opened 2017-10-04
Fix the discovery build by removing the unnecessary wait

schenedx:
@clintonb Please review.
I believe this will fix the build


clintonb (commented):
I don’t think this will fix anything. The wait is to ensure the browser is available. If it’s failing, that means the browser hasn’t started.

schenedx:
@clintonb I do think this change fixes the unit tests.
After we created the web driver out of firefox, I don't believe we need to call `implicitly_wait` at all. I study the code at https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L119, it give me concerns what exactly `implicitly_wait` does.
Then I read https://stackoverflow.com/questions/10404160/when-to-use-explicit-wait-vs-implicit-wait-in-selenium-webdriver and further agreed that we do not need that call here.

clintonb (approved):

PR #1159, opened 2017-10-04
Updated Publisher upgrade deadline logic

clintonb:
The upgrade deadline published to external services now follows edX business logic. Only verified seats have an upgrade deadline. If no value is set, one is calculated based on the course run's end date.

EDUCATOR-1465


schenedx (approved):
👍

clintonb:
Merged via https://github.com/edx/course-discovery/pull/1160.

PR #1160, opened 2017-10-04
Updated Publisher to push pacing type to Studio

clintonb:
This change ensures that self-paced course runs are marked appropriately in Studio.

EDUCATOR-1464


clintonb:
See last commit.

clintonb:
This functionality is dependent on https://github.com/edx/edx-platform/pull/16157; however, this change can be merged without that pull request first being merged.

schenedx (approved):
👍

PR #1161, opened 2017-10-04
Added missed migration

BenjiLee:
@clintonb Did you miss a migration by any chance? Master branch is showing me that there are still migrations to be made.


clintonb (changes requested):
Make sure you've pulled the latest image and code. These fields were added by https://github.com/edx/course-discovery/pull/1143. If you are using an outdated version of Django Simple History, that might explain why Django is attempting to remove these fields.

BenjiLee:
👍  Currently using my own image for my current ticket.

PR #1162, opened 2017-10-04
Added management command to update course images

clintonb:
This command downloads images from the course's card_image_url field
value, and stores them in the image field.

LEARNER-2463

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/download_course_images.py
+++ b/course_discovery/apps/course_metadata/management/commands/download_course_images.py
@@ -1,0 +1,62 @@
+import logging
+
+import requests
+from django.core.files.base import ContentFile
+from django.core.management import BaseCommand
+
+from course_discovery.apps.course_metadata.models import Course
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/management/commands/download_course_images.py
```diff
@@ -0,0 +1,61 @@
...
+        courses = Course.objects.filter(card_image_url__isnull=False).exclude(card_image_url='').order_by('key')
+
+        if not options['overwrite_existing']:
+            courses = courses.filter(image='')
+
+        count = courses.count()
+        if count:
+            logger.info('Retrieving images for [%d] courses...', count)
+        else:
+            logger.info('All courses are up to date.')
```
schenedx: `return` after this statement?

schenedx (commented):
Why creating a new management command, but not having this as part of `refresh_course_metadata` management command?

clintonb:
I opted for a new command so that we could perform this task optionally. Realistically, we only need to run it once.

schenedx (approved):
👍

PR #1164, opened 2017-10-05
Fix the Course run filters in *In Development* tab.

attiyaIshaque:
### Description
This PR fixes that filters of course run in *In development* tab not only showing the count also shows the data. 

### How to Test?
**Production**
https://prod-edx-discovery.edx.org/publisher/

- Click on the filters
- Observe there is no data.

**Sandbox**
- https://discovery-escalation-discovery.sandbox.edx.org/publisher/
-  Click on the filters
- Observe filters worked correctly

_username : Attiya
Password: attiya_

### Reviewers
- [x] @noraiz-anwar 
- [x] @waheedahmed  

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -114,9 +114,9 @@ class CourseRunListView(mixins.LoginRequiredMixin, ListView):
         else:
             studio_request_courses = []
 
-        context['studio_request_courses'] = [CourseRunWrapper(course_run) for course_run in studio_request_courses]
-        context['unpublished_course_runs'] = [CourseRunWrapper(course_run) for course_run in unpublished_course_runs]
-        context['published_course_runs'] = [CourseRunWrapper(course_run) for course_run in published_course_runs]
+        context['studio_request_courses'] = CourseRunWrapper.get_course_run_wrappers(studio_request_courses)
+        context['unpublished_course_runs'] = CourseRunWrapper.get_course_run_wrappers(unpublished_course_runs)
+        context['published_course_runs'] = CourseRunWrapper.get_course_run_wrappers(published_course_runs)
         context['default_published_days'] = self.default_published_days
 
         in_progress_course_runs = course_runs.filter(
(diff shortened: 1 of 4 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -127,24 +127,71 @@ def get_context_data(self, **kwargs):
...
+            course_runs:
+        Returns:
+            returns the internal user status
+        """
+        def _filter_internal_user(course_run):
+            return str(course_run.internal_user_status)
+
+        return self.get_filters_status(course_runs, _filter_internal_user)
+
+    def get_filters_status(self, course_runs, filter_func):
```
waheedahmed: I don't think it's worth doing all this stuff to just get already defined statuses, how much time it will take if there are 2K+ course runs loading on dashboard?

Instead you can simply store these statuses in some variables and use them on both places. https://github.com/edx/course-discovery/blob/549a00b4ba7dccce376da88e11b8c9914825bebb/course_discovery/apps/publisher/wrappers.py#L226

waheedahmed (changes requested):

course_discovery/apps/publisher/views.py
attiyaIshaque: @waheedahmed Sure I will change it in some variables. Thanks for explaining it.

attiyaIshaque:
@waheedahmed I have addressed your requested changes.

course_discovery/apps/publisher/wrappers.py
```diff
@@ -247,3 +258,13 @@ def preview_declined(self):
             object_pk=self.wrapped_obj.id,
             comment_type=CommentTypeChoices.Decline_Preview
         ).exists()
+
+    @classmethod
+    def get_courses_run_wrapper(cls, course_runs):
```
waheedahmed: I think more appropriate name for this is `get_course_run_wrappers`.

course_discovery/apps/publisher/views.py
```diff
@@ -127,21 +127,26 @@ def get_context_data(self, **kwargs):
...
         site = Site.objects.first()
         context['site_name'] = 'edX' if 'edx' in site.name.lower() else site.name
 
-        context['course_team_count'] = in_progress_course_runs.filter(
+        course_team_course_runs = in_progress_course_runs.filter(
             course_run_state__owner_role=PublisherUserRole.CourseTeam
-        ).count()
-        context['internal_user_count'] = in_progress_course_runs.exclude(
+        )
+        context['course_team_status'] = '%s|%s' % (CourseRunWrapper.Draft, CourseRunWrapper.AwaitingCourseTeamReview)
```
waheedahmed: Use format.

course_discovery/apps/publisher/views.py
```diff
@@ -127,21 +127,26 @@ def get_context_data(self, **kwargs):
...
+        context['course_team_count'] = course_team_course_runs.count()
+
+        internal_user_course_runs = in_progress_course_runs.exclude(
```
waheedahmed: I think you can reduce one query by combining count query here, same with course_team_course_runs.

waheedahmed (changes requested):

waheedahmed (approved):
:+1:
Please remember to squash your commits

noraiz-anwar (approved):
👍

PR #1165, opened 2017-10-05
Updated course and course run API endpoints to use stored images

clintonb:
If a course has an image on the image field, that image's URL will be
used in the API response. Course runs now use the related course's image.

LEARNER-2751

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -294,6 +294,7 @@ class Course(TimeStampedModel):
     )
     slug = AutoSlugField(populate_from='key', editable=True)
     video = models.ForeignKey(Video, default=None, null=True, blank=True)
+
     # TODO Remove this field.
     number = models.CharField(
         max_length=50, null=True, blank=True, help_text=_(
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -633,7 +647,7 @@ def save(self, *args, **kwargs):
 
         if is_publishable:
             publisher = CourseRunMarketingSitePublisher(self.course.partner)
-            previous_obj = CourseRun.objects.get(id=self.id) if self.id else None
+            previous_obj = None#CourseRun.objects.get(id=self.id) if self.id else None
```
schenedx: Please review the comment

course_discovery/apps/course_metadata/models.py
```diff
@@ -644,6 +658,7 @@ def save(self, *args, **kwargs):
                 super(CourseRun, self).save(*args, **kwargs)
                 publisher.publish_obj(self, previous_obj=previous_obj)
         else:
+            logger.info('Course run [%s] is not publishable.', self.key)
```
schenedx: Should we explain a bit what `publishable` means? Like a course not in `published` status or something?

schenedx (approved):
👍

PR #1166, opened 2017-10-05
Minor conveniences for Publisher and publishing

clintonb:
- Improved logging
- Switch to disable Publisher permissions

LEARNER-2458


clintonb:
FYI @schenedx

PR #1167, opened 2017-10-06
Updated image download code to continue if a download fails for a course

clintonb:
Failure to download an image for one course should not stop execution of downloads for other courses. Exception handling is now in place to ensure we do not prematurely exit the program.

LEARNER-2463


clintonb:
FYI @schenedx @jdmulloy

PR #1169, opened 2017-10-06
Added license field to Course Run model, API endpoint, and search index

clintonb:
The license is now ingested from the Course API on LMS and saved to the Course Run model. This value is in turn exposed on this service's course run and search API endpoints.

LEARNER-2791


clintonb:
Data ingestion relies on https://github.com/edx/edx-platform/pull/16189 being released.

clintonb:
FYI @edx/learner

PR #1171, opened 2017-10-10
Updated tests for currency API endpoint

clintonb:
The tests now use py.test.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,24 @@
+import logging
+
+import pytest
+from django.core.cache import cache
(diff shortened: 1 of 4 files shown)


MatthewPiatetsky:
👍 can you clarify the change? did i write the tests with unittest and they should have been using pytest?

clintonb:
I don't believe py.test was the standard when you wrote these. This is the start of my converting all of the tests so that we can ultimately take advantage of running them in parallel to cut down build time.

PR #1172, opened 2017-10-10
Converted tests to py.test

clintonb:


clintonb:
FYI @edx/learner I've converted a number of tests to use the py.test style. This is a precursor to running them in parallel (https://github.com/edx/course-discovery/pull/976).

PR #1175, opened 2017-10-11
Updated production requirements to latest versions

clintonb:


clintonb:
FYI @edx/learner @edx/devops

feanil:
@clintonb did you already review the changelogs?  Can you link to them?

PR #1176, opened 2017-10-12
Refactor LMSAPIClient to use discovery service user instead of request user.

douglashall:
ENT-657

Fix for https://github.com/edx/course-discovery/pull/1145#issuecomment-336181110.

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -4,6 +4,7 @@ import itertools
 from urllib.parse import urlencode
 
 import ddt
+import mock
 import pytest
 import pytz
 import responses
@@ -33,7 +34,7 @@ from course_discovery.apps.api.serializers import (AffiliateWindowSerializer, Ca
                                                    VideoSerializer, get_utm_source_for_user)
 from course_discovery.apps.api.tests.mixins import SiteMixin
 from course_discovery.apps.catalogs.tests.factories import CatalogFactory
-from course_discovery.apps.core.models import User
+from course_discovery.apps.core.models import Partner, User
 from course_discovery.apps.core.tests.factories import PartnerFactory, UserFactory
 from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin, LMSAPIClientMixin
@@ -1409,7 +1410,8 @@ class TestGetUTMSourceForUser(LMSAPIClientMixin, TestCase):
         self.partner = PartnerFactory.create()
 
     @override_switch('use_company_name_as_utm_source_value', active=False)
-    def test_with_waffle_switch_turned_off(self):
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    def test_with_waffle_switch_turned_off(self, mock_access_token):  # pylint: disable=unused-argument
         """
         Verify that `get_utm_source_for_user` returns User's username when waffle switch
         `use_company_name_as_utm_source_value` is turned off.
@@ -1417,7 +1419,8 @@ class TestGetUTMSourceForUser(LMSAPIClientMixin, TestCase):
 
         assert get_utm_source_for_user(self.partner, self.user) == self.user.username
 
-    def test_with_missing_lms_url(self):
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    def test_with_missing_lms_url(self, mock_access_token):  # pylint: disable=unused-argument
         """
         Verify that `get_utm_source_for_user` returns default value if
         `Partner.lms_url` is not set in the database.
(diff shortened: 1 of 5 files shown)


course_discovery/apps/core/api_client/lms.py
```diff
@@ -18,10 +18,8 @@ class LMSAPIClient(object):
     API Client for communication between discovery and LMS.
     """
 
-    def __init__(self, site, user):
-        self.site = site
-        self.user = user
-        self.client = EdxRestApiClient(self.site.partner.lms_url, oauth_access_token=user.access_token)
+    def __init__(self, site):
+        self.client = EdxRestApiClient(site.partner.lms_url, oauth_access_token=site.partner.access_token)
```
clintonb: Partner's use JWT access tokens, not bearer tokens (expected by `oauth_access_token`).

clintonb (changes requested):

course_discovery/apps/core/api_client/lms.py
douglashall: Thanks. Pushing fix now.

course_discovery/apps/core/api_client/lms.py
```diff
@@ -51,14 +49,9 @@ def get_api_access_request(self, user):
...
-        # So, we will filter by username for staff users.
-        if user.is_staff:
-            query_parameters = {
-                'user__username': user.username
-            }
+        query_parameters = {
+            'user__username': user.username
+        }
 
         cache_key = get_cache_key(username=user.username, resource=resource)
```
clintonb, line 56: Since you're here, please remove the comment on line 74 as it is wrong. We don't require an access request for all clients. For example, none of the service users have requests. I, personally, don't have an access request. That comment will be confusing in the future.

douglashall: Done.

course_discovery/apps/core/api_client/lms.py
```diff
@@ -78,7 +71,6 @@ def get_api_access_request(self, user):
             except (SlumberBaseException, ConnectionError, Timeout):
                 logger.exception('Failed to fetch API Access Request from LMS for user "%s".', user.username)
             except (IndexError, KeyError):
-                # This should not happen as user must always have at-least one api-access-request.
-                logger.exception('APIAccessRequest model not found for user [%s].', user.username)
+                logger.warning('APIAccessRequest model not found for user [%s].', user.username)
```
clintonb: This should be at the info level.

clintonb (approved):

saleem-latif (approved):
LGTM

course_discovery/apps/core/tests/mixins.py
```diff
@@ -42,7 +42,26 @@ def reindex_courses(self, program):
...
-    def mock_api_access_request(self, lms_url, status=200, api_access_request_overrides=None):
+    def mock_access_token(self, partner, jwt_token="Test JWT"):
+        """
+        Mock the api access requests endpoint response of the LMS.
+        """
+        data = {
+            "access_token": jwt_token,
+            "token_type": "JWT",
+            "expires_in": 31535999,
+            "scope": "API Permissions"
```
clintonb: That’s not a proper scope.

course_discovery/apps/core/tests/mixins.py
```diff
@@ -42,7 +42,26 @@ def reindex_courses(self, program):
 
 
 class LMSAPIClientMixin(object):
-    def mock_api_access_request(self, lms_url, status=200, api_access_request_overrides=None):
+    def mock_access_token(self, partner, jwt_token="Test JWT"):
```
clintonb: This is overkill. See https://github.com/edx/course-discovery/blob/5e45a193854a9ba15a782b27ed7d862d635fdc78/course_discovery/apps/publisher/api/v1/tests/test_views.py#L94-L95.

clintonb (changes requested):
Your test class should be a py.test style class, if a class at all. See https://docs.pytest.org/en/latest/contents.html.

course_discovery/apps/core/tests/mixins.py
saleem-latif: yes, I tried to use that at first but I had to use that decorator for almost all the new test cases, that resulted in a lot of repetition. I think using it with mixin gives us nice contained mocking and adding new test cases to this class also becomes easier.

saleem-latif:
@clintonb I have incorporated your feedback, please take another look

course_discovery/apps/core/tests/mixins.py
clintonb: You only have four test cases. The mock will be cleaner, and slightly faster, than using `responses`.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1397,7 +1398,8 @@ def serialize_program(self, program):
         return serializer
 
 
-class TestGetUTMSourceForUser(LMSAPIClientMixin, TestCase):
+@pytest.mark.django_db
+class TestGetUTMSourceForUser(LMSAPIClientMixin, unittest.TestCase):
```
clintonb: The point of using `pytest` is to _not_ use `unittest`. See some example tests in https://github.com/edx/course-discovery/pull/1172.

PR #1177, opened 2017-10-13
Subject API Translations clean up

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-2616

Phase 1 is here https://github.com/edx/course-discovery/pull/1129/files

Phase 2:  
- Delete deprecated fields in the Subjects model.
- Create field names for the SubjectTranslations table that were deprecated in the Subjects model during phase 1.
- Migrate the data from the old field name, to the new field name.
- Update code to use the original field name

TODO:
- [x] Fix unit tests


course_discovery/apps/course_metadata/models.py
```diff
@@ -137,13 +140,17 @@ class Meta:
...
         qs = Subject.objects.filter(partner=self.partner_id)
-        if qs.filter(translations__name_t=self.name_t).exists():
+        if qs.filter(translations__name_t=self.name_t).exclude(pk=self.pk).exists():
             raise ValidationError({'name_t': ['Subject with this Name and Partner already exists', ]})
 
 
 class SubjectTranslation(TranslatedFieldsModel):
     master = models.ForeignKey(Subject, related_name='translations', null=True)
 
+    name = models.CharField(max_length=255, blank=False, null=False, default="TODO")
```
amangano-edx: Rather than populate this with fake default data, why not allow blanks/nulls for now until all the data has been migrated, then alter the field to add constrains to prevent blanks/nulls in the future?

course_discovery/apps/course_metadata/models.py
```diff
@@ -137,13 +135,17 @@ class Meta:
     def validate_unique(self, *args, **kwargs):
         super(Subject, self).validate_unique(*args, **kwargs)
         qs = Subject.objects.filter(partner=self.partner_id)
-        if qs.filter(translations__name_t=self.name_t).exists():
+        if qs.filter(translations__name_t=self.name_t).exclude(pk=self.pk).exists():
```
amangano-edx: Should this reference `name` instead since we're removing `name_t`?

course_discovery/apps/course_metadata/migrations/0064_migrate_subjecttranslations_data.py
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            translations = SubjectTranslation.objects.filter(master_id=subject.pk)
+            for translation in translations:
+                translation.name = translation.name_t
+                translation.subtitle = translation.subtitle_t
+                translation.description = translation.description_t
+                translation.save()  # Note this only calls Model.save()
+        except ObjectDoesNotExist:
+            # nothing to migrate
+            logger.exception('Migrating data from SubjectTranslation for master_id={} DoesNotExist'.format(subject.pk))
```
amangano-edx: I think you might be able to do this in one SQL command. See https://stackoverflow.com/questions/6308594/how-can-i-copy-data-from-one-column-to-another-in-the-same-table for inspiration.

course_discovery/apps/course_metadata/migrations/0064_migrate_subjecttranslations_data.py
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            translations = SubjectTranslation.objects.filter(master_id=subject.pk)
+            for translation in translations:
+                translation.name_t = translation.name
+                translation.subtitle_t = translation.subtitle
+                translation.description_t = translation.description
+                translation.save()  # Note this only calls Model.save()
+        except ObjectDoesNotExist:
+            # nothing to migrate
+            logger.exception('Migrating data from SubjectTranslation for master_id={} DoesNotExist'.format(subject.pk))
```
amangano-edx: Same here, I think you can probably do this in one SQL query.

course_discovery/apps/course_metadata/models.py
```diff
@@ -137,13 +135,17 @@ class Meta:
     def validate_unique(self, *args, **kwargs):
```
amangano-edx, line 135: It might be good to add a test for this if it was broken before.

amangano-edx (approved):
Some small comments/questions. Otherwise LGTM.

course_discovery/apps/course_metadata/models.py
BenjiLee: yes

course_discovery/apps/course_metadata/models.py
BenjiLee: will do

course_discovery/apps/course_metadata/models.py
BenjiLee: added

PR #1181, opened 2017-10-18
Remove unused fields for Subject via migration

BenjiLee:
https://openedx.atlassian.net/browse/LEARNER-2617

Final step of https://openedx.atlassian.net/browse/LEARNER-2314

Run the migration for the unused fields.


amangano-edx:
LGTM

PR #1182, opened 2017-10-19
Added ability to specify total hours of effort for a program

McKenzieW:
[LEARNER-1941](https://openedx.atlassian.net/browse/LEARNER-1941)


McKenzieW:
@edx/learner-heart check it out

course_discovery/apps/course_metadata/models.py
```diff
@@ -809,6 +809,9 @@ class Program(TimeStampedModel):
     excluded_course_runs = models.ManyToManyField(CourseRun, blank=True)
     partner = models.ForeignKey(Partner, null=True, blank=False)
     overview = models.TextField(null=True, blank=True)
+    total_hours_of_effort = models.PositiveSmallIntegerField(
+        null=True, blank=True,
```
bderusha, line 813: Why `blank=True`?

bderusha (commented):
Do we need to do anything to have this field included in the serializer or do we get that for free?

course_discovery/apps/course_metadata/models.py
McKenzieW: It probably shouldn't be. I just immediately followed up writing null=True with blank=True. Will change.

McKenzieW: Also very good question about the serializer. I do not know the answer and will have to look into it.

course_discovery/apps/course_metadata/models.py
```diff
@@ -809,6 +809,9 @@ class Program(TimeStampedModel):
...
+        help_text='Total estimated time needed to complete all courses belonging to this program.')
```
amangano-edx: Nit: It might be good to mention that this field is intended to be used on certificates.

amangano-edx (approved):
LGTM

course_discovery/apps/course_metadata/models.py
McKenzieW: The other integer fields have this attribute, and when I removed it it broke tests. Is there a particular reason hours of effort shouldn't have blank=True?

bderusha (approved):

course_discovery/apps/course_metadata/models.py
bderusha: If that's the pattern we are using that's fine.  Just wanted to ask the question.  Given that programs are created in django admin it's fine, but in theory this could result in some mixed data of `null` and `empty string` which in this case mean the same thing.  In practice django admin will use empty string when submitting the form so we should never end up using `null`

PR #1183, opened 2017-10-19
LEARNER-2941: set default language for update_index

BenjiLee:
FYI @hamzamunir7300 

https://openedx.atlassian.net/browse/LEARNER-2941

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -24,6 +24,9 @@ class Command(HaystackCommand):
         return conn.count(index_name).get('count')
 
     def handle(self, *items, **options):
+        from django.utils import translation
+        translation.activate(settings.LANGUAGE_CODE)
+
         self.backends = options.get('using')
         if not self.backends:
             self.backends = list(haystack_connections.connections_info.keys())


hamzamunir7300:
@BenjiLee So  Without Including the  following line.
```python
translation.activate(settings.LANGUAGE_CODE)
```
i was able to reproduce the issue locally writing this line in the code. at the start of the command.

```python
Subject.objects.create()
```
I have another question about the PR don't we need to use at the end of the command.
```python
translation.deactivate()
```

course_discovery/apps/core/utils.py
```diff
@@ -2,10 +2,12 @@
 import logging
 
 from django.conf import settings
+from django.utils import translation
 
 from course_discovery.settings.process_synonyms import get_synonyms
 
 logger = logging.getLogger(__name__)
+translation.activate(settings.LANGUAGE_CODE)
```
clintonb: Utility libraries should _not_ have side effects such as this. This will come back to bite us in some unexpected manner.

clintonb (changes requested):

BenjiLee:
@clintonb Ready for another pass. I don't believe any of the other management commands need the translations to be activated.

clintonb (approved):
This is good for getting things fixed now (and can be merged), but please add a test to validate the fix. I am surprised this issue didn't break any existing tests.

BenjiLee:
I wonder if it is because it is dependent on prod data

or

if the tests have translations activated/have a site_id.

clintonb:
I suspect the test simply doesn't exist. I don't see any tests that fully exercise the management command. Please ensure one is created.

PR #1184, opened 2017-10-19
Updated Publisher to always create an audit seat for verified course runs

clintonb:
The SeatForm implements the logic to create an audit seat for verified
and credit course runs. It also removes audit seats for professional
course runs.

In the longterm, this should be implemented on the UI and exposed to the
end user to avoid future confusion.

LEARNER-2876

### DIFF
--- a/course_discovery/apps/publisher/forms.py
+++ b/course_discovery/apps/publisher/forms.py
@@ -1,5 +1,7 @@
 import html
+import logging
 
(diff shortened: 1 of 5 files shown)


clintonb:
The second commit includes a management command to add audit seats to existing course runs.

schenedx (approved):
👍

PR #1187, opened 2017-10-23
Updated marketing site school loader

clintonb:
The UUID is now used as the primary identifier when refreshing schools. Usage of the title as the key poses issues with schools whose titles and keys differ on LMS (e.g. UC BerkeleyX vs. BerkeleyX). This change ensures that we can still pull changes to images and descriptions, while also supporting manual edits to keys.

EDUCATOR-1531


schenedx (approved):
👍

PR #1188, opened 2017-10-23
Add the ability to access and filter courses by UUID and not just key.

mjfrey:
Add to the course API endpoint the ability to access a course directly via the UUID.
Also added a uuid course filter that adds the ability to provide a list of UUIDs and get back the information for the courses.

LEARNER-2882

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -1,4 +1,7 @@
+import re
+
 from django.db.models.functions import Lower
+from django.shortcuts import get_object_or_404
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import viewsets
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -16,15 +19,37 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+        if self.course_key_regex.match(key):
+            filter_key = 'key'
+        elif self.course_uuid_regex.match(key):
+            filter_key = 'uuid'
+
+        filter_kwargs = {filter_key: key}
+        obj = get_object_or_404(queryset, **filter_kwargs)
+
+        # May raise a permission denied
```
staubina, line 47: Should we be concerned with this raising the permission denied exception?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -16,15 +19,37 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
     """ Course resource. """
     filter_backends = (DjangoFilterBackend,)
     filter_class = filters.CourseFilter
+    # We should eventually transition over to UUID as the PK.
```
staubina: Is this comment of value in here? I find that these get lost and the value is diminishing.  Should we create a backlog Story or Epic instead?

staubina (commented):
Some small questions and comments.  The changes seem straight forward.

course_discovery/apps/api/v1/views/courses.py
mjfrey: no, this is what the base DRF method does.

bderusha (approved):

staubina (approved):

PR #1191, opened 2017-10-26
Fix the bug left in the course run publish module

schenedx:
The courseRun object do not have `marketing_slug` property. We should've caught this last time around. 

@clintonb Please review.


clintonb (approved):

clintonb:
Test?

schenedx:
@clintonb Added tests.

PR #1193, opened 2017-10-27
Fix the course image resolution so the aspect ratio on course image is maintained

schenedx:
Currently, the course image uploaded to publisher is of resolution 2120X1192
This translates to 318X179 instead of the old 318X210
We will display this image to the front end and we should have the image size correctly.

@clintonb Please review


clintonb (approved):

schenedx:
This should be deployed together with https://github.com/edx/edx-mktg/pull/1712

PR #1194, opened 2017-10-27
ENT-687 Add indexed has_enrollable_seats field to CourseRun model.

douglashall:
This field will make it possible for clients of the search/all
endpoint to filter CourseRun results to those that have enrollable
seats.

ENT-687

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -501,7 +501,7 @@ class CourseRun(TimeStampedModel):
 
         return deadline
 
-    def enrollable_seats(self, types):
+    def enrollable_seats(self, types=None):
         """
         Returns seats, of the given type(s), that can be enrolled in/purchased.
 
@@ -523,11 +523,19 @@ class CourseRun(TimeStampedModel):
         if self.enrollment_end and now > self.enrollment_end:
             return enrollable_seats
 
+        types = types or Seat.SEAT_TYPES
         for seat in self.seats.all():
             if seat.type in types and (not seat.upgrade_deadline or now < seat.upgrade_deadline):
                 enrollable_seats.append(seat)
 
         return enrollable_seats
+
+    @property
+    def has_enrollable_seats(self):
+        """
+        Return a boolean indicating whether or not enrollable Seats are available for this CourseRun.
+        """
+        return len(self.enrollable_seats()) > 0
 
     @property
     def image_url(self):
@@ -701,6 +709,8 @@ class Seat(TimeStampedModel):
     VERIFIED = 'verified'
     PROFESSIONAL = 'professional'
     CREDIT = 'credit'
+
+    SEAT_TYPES = [HONOR, AUDIT, VERIFIED, PROFESSIONAL, CREDIT]
 
     # Seat types that may not be purchased without first purchasing another Seat type.
     # EX: 'credit' seats may not be purchased without first purchasing a 'verified' Seat.
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -5,6 +5,7 @@ from decimal import Decimal
 import ddt
 import mock
 import pytest
+import pytz
 from dateutil.parser import parse
 from django.conf import settings
 from django.core.exceptions import ValidationError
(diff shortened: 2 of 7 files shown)


asadiqbal08:
So instead of calling [is_course_run_enrollable here](https://github.com/edx/edx-enterprise/pull/229/files#diff-4bcae564bc272f5c2e6753dc6021eb1cR946), We will filter the contents on the basis of this [has_enrollable_seats](has_enrollable_seats) ? right?

douglashall:
@asadiqbal08 Correct, we will also need to add the filter [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L909).

asadiqbal08 (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -501,6 +501,14 @@ def get_paid_seat_enrollment_end(self):
 
         return deadline
 
+    @cached_property
+    def has_enrollable_seats(self):
+        """
+        Return a boolean indicating whether or not enrollable Seats are available for this CourseRun.
+        """
+        seat_types = [choice[0] for choice in Seat.SEAT_TYPE_CHOICES]
+        return len(self.enrollable_seats(seat_types)[:1]) > 0
```
clintonb: Why are you excluding a seat type? Also, relying on the order of `Seat.SEAT_TYPE_CHOICES` to remain consistent makes this code quite brittle.

clintonb: Instead of doing the unnecessary filtering, update `enrollable_seats` to return all seats if a list of seat types is not specified.

clintonb (changes requested):

course_discovery/apps/course_metadata/models.py
```diff
@@ -501,7 +501,14 @@ def get_paid_seat_enrollment_end(self):
...
 
-    def enrollable_seats(self, types):
+    @cached_property
+    def has_enrollable_seats(self):
+        """
+        Return a boolean indicating whether or not enrollable Seats are available for this CourseRun.
+        """
+        return len(self.enrollable_seats()[:1]) > 0
+
+    def enrollable_seats(self, types=Seat.SEAT_TYPES):
```
douglashall: @clintonb How does this look?

course_discovery/apps/course_metadata/models.py
```diff
@@ -501,7 +501,14 @@ def get_paid_seat_enrollment_end(self):
 
         return deadline
 
-    def enrollable_seats(self, types):
+    @cached_property
+    def has_enrollable_seats(self):
+        """
+        Return a boolean indicating whether or not enrollable Seats are available for this CourseRun.
+        """
+        return len(self.enrollable_seats()[:1]) > 0
```
clintonb: Why are you filtering?

douglashall: I was following the pattern in [has_enrollable_paid_seats](https://github.com/edx/course-discovery/pull/1194/files#diff-a79b6f6b0e3ecfd5c59847f0e39d59d2R475). I'm guessing someone thought calculating the length of the one-item list is slightly faster than calculating the length of the whole list. Think I should change this?

clintonb: I don’t know the reasoning behind that. It makes no obvious sense, and should not be replicated.

douglashall: Sounds good to me.

douglashall:
@clintonb Anything else to address here?

clintonb (commented):
Yeah, you’re missing something really important: tests! The new method needs to be tested, along with your changes to the existing method.

douglashall:
@clintonb Good call. How about now?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -71,20 +72,37 @@ def test_enrollable_seats(self):
...
+        course_run.start = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=1)
         course_run.save()
         self.assertEqual(
             course_run.enrollable_seats([Seat.VERIFIED, Seat.PROFESSIONAL]),
             [verified_seat, professional_seat]
         )
 
+        # Enrollable seats of any type should be returned
+        # when no type parameter is specified.
+        self.assertEqual(
```
clintonb: New code should use standard `assert` as part of testing with `py.test`.

```python
assert course_run.enrollable_seats() == [verified_seat, professional_seat, honor_seat]
```

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -71,20 +72,37 @@ def test_enrollable_seats(self):
...
+            course_run.enrollable_seats(),
+            [verified_seat, professional_seat, honor_seat]
+        )
+
+    def test_has_enrollable_seats(self):
+        """ Verify the expected value of has_enrollable_seats is returned. """
+        course_run = factories.CourseRunFactory(start=None, end=None, enrollment_start=None, enrollment_end=None)
+        verified_seat = factories.SeatFactory(course_run=course_run, type=Seat.VERIFIED, upgrade_deadline=None)
+        self.assertTrue(course_run.has_enrollable_seats())
```
clintonb: `assert course_run.has_enrollable_seats()`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -71,20 +72,37 @@ def test_enrollable_seats(self):
...
+
+        course_run.end = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=1)
+        course_run.save()
+        self.assertFalse(course_run.has_enrollable_seats())
```
clintonb: `assert not course_run.has_enrollable_seats()`

clintonb (approved):

PR #1196, opened 2017-10-30
Set time to 23:59:59 for Publisher's default upgrade deadline

clintonb:
The default upgrade deadline for Publisher course runs now has the time set to 23:59:59.99999 to give learners the entire day to upgrade.

LEARNER-3056


clintonb:
FYI @schenedx

schenedx (approved):
👍

PR #1198, opened 2017-10-31
EDUCATOR-1523 Fix Instructor photos size and dimensions.

attiyaIshaque:
### Description
This PR fixes that uploaded photos for instructors would be  110px x 110px and no other dimensions are allowed and file size should be less than 256kb.

### How to Test?
**Stage**
https://stage-edx-discovery.edx.org/publisher/course_runs/3771/edit/
- Click on Add new Instructor
- upload photo of any dimensions
- Save the instructor
- it will save the instructor

**Sandbox**
https://discovery-escalation-discovery.sandbox.edx.org/publisher/
-  Click on Add new Instructor
- upload photo of any dimensions
- it wouldn't allow you to add any kind of dimension and size > 256kb

_username : Attiya
Password: attiya_

### Reviewers
- [x] @Rabia23 
- [x] @asadazam93  

### Post-review


asadazam93 (approved):

course_discovery/static/js/publisher/instructors.js
```diff
@@ -96,22 +96,37 @@ $(document).ready(function () {
...
-            clearModalError();
-            reader.onload = function (e) {
-                $('.select-image').attr('src', e.target.result).addClass('image-updated');
-            };
+function loadSelectedImage (input) {
+    var maxFileSize = 256, // Size in KB's
+        imageFile = input.files[0],
+        imageDimension = 110,
+        imgPath = 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==';
+    if (input.files && imageFile) {
```
Rabia23: Why there is need of `input.files` in `if` condition? I think only `imageFile` is sufficient to check whether file exists or not.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -96,22 +96,37 @@ $(document).ready(function () {
...
-        } else {
-            var reader = new FileReader();
-
-            clearModalError();
-            reader.onload = function (e) {
-                $('.select-image').attr('src', e.target.result).addClass('image-updated');
-            };
+function loadSelectedImage (input) {
+    var maxFileSize = 256, // Size in KB's
+        imageFile = input.files[0],
```
Rabia23, line 101: What if `input.files` is empty? It will throw an error because you are accessing an element from an empty array.

Rabia23 (changes requested):

course_discovery/static/js/publisher/instructors.js
attiyaIshaque: oh, my bad. I will remove the input.files condition

course_discovery/static/js/publisher/instructors.js
attiyaIshaque: Actually, function is calling on, onChange event. so it will never throw the exception

attiyaIshaque:
@Rabia23 I have addressed your requested change.

course_discovery/static/js/publisher/instructors.js
Rabia23: alright.

Rabia23 (approved):

Rabia23:
👍

schenedx:
@attiyaIshaque The instructions says, the image must be less than 110X110px. That's in contrast of the message you put on the PR description where the image must only be 110X110 pixel. I think we should have a PR to update the message properly.

attiyaIshaque:
@schenedx oh I got your point. I will update it in my other PR. :)

PR #1200, opened 2017-10-31
Adding CourseEntitlement model to discovery

McKenzieW:
This new model is to Courses as Seats are to CourseRuns.
It stores information about the individual products available for a course, which is [returned to discovery from ecommerce on Publish](https://openedx.atlassian.net/browse/LEARNER-3065).

[LEARNER-3064](https://openedx.atlassian.net/browse/LEARNER-3064)


McKenzieW:
@edx/helio check it out

mjfrey:
Need some tests.

Look at the test_course_runs.py and test_courses.py

course_discovery/apps/course_metadata/migrations/0066_auto_20171031_1758.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
Mjloturco: This migration could use a more descriptive name

course_discovery/apps/course_metadata/models.py
```diff
@@ -746,6 +746,26 @@ class Meta(object):
...
+    }
+    course = models.ForeignKey(Course, related_name='entitlements')
+    mode = models.ForeignKey(SeatType)
+    price = models.DecimalField(**PRICE_FIELD_CONFIG)
+    currency = models.ForeignKey(Currency)
+    sku = models.CharField(max_length=128, null=True, blank=True)
+
+    class Meta(object):
+        unique_together = (
+            ('course', 'mode', 'currency')
```
Mjloturco: Should 'currency' be a part of the unique_together? can we have multiple entitlements for various currencies?

Mjloturco (commented):
Could maybe use a test the creates a CourseEntitlement (just to verify that making one works), for the most part LGTM

course_discovery/apps/course_metadata/migrations/0066_auto_20171031_1758.py
McKenzieW: Yeah it was automatically generated using manage.py makemigrations. I don't know what our edX standard practice is for renaming generated migrations

mjfrey: You can add comments to generated migrations.

course_discovery/apps/course_metadata/models.py
McKenzieW: Good question. I'm not quite sure if we're wanting to change that for Entitlements, but currency is part of the unique_together for Seats, so I figured it would be the same for this analogous model.

course_discovery/apps/course_metadata/migrations/0066_auto_20171031_1758.py
McKenzieW: @Mjloturco what is the information that you think ought to be conveyed about this migration, just that the new model is being created, or a description of what the model is and what it's used for as well?

course_discovery/apps/course_metadata/models.py
```diff
@@ -746,6 +746,26 @@ class Meta(object):
...
+class CourseEntitlement(TimeStampedModel):
+    """ Model storing product metadata for a Course. """
+    PRICE_FIELD_CONFIG = {
+        'decimal_places': 2,
+        'max_digits': 10,
+        'null': False,
+        'default': 0.00,
+    }
+    course = models.ForeignKey(Course, related_name='entitlements')
+    mode = models.ForeignKey(SeatType)
```
staubina, line 778: I am not familiar with the Ecommerce data models, but @mjfrey do we use Mode in Ecommerce elsewhere?  Or is it more common to use seat?

We should stick with whatever convention is used in Ecommerce.

staubina (commented):
Just a quick question.

course_discovery/apps/course_metadata/models.py
McKenzieW: @mjfrey @staubina any input on the necessity of `'currency'` in the `unique_together` constraint for `CourseEntitlements`?

McKenzieW:
@edx/helio I am currently working on some updates to the (EcommerceApiDataLoader)[https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/data_loaders/api.py#L250] that will populate / update the new CourseEntitlement model when refresh_course_metadata is run, but it looks like they will be dependent on the delivery of [LEARNER-3065](https://openedx.atlassian.net/browse/LEARNER-3065) as the updates include retrieving Course Entitlement data from the Ecommerce API.

mjfrey:
I think this looks right.  We will want to coordinate with the product API endpoint so we can support removal of nonexistent course entitlements.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -266,30 +268,47 @@ def ingest(self):
...
+                for future in [executor.submit(self._request_course_runs, page) for page in pagerange]:
                     response = future.result()
-                    self._process_response(response)
+                    self._process_course_runs(response)
+                for future in [executor.submit(self._request_entitlments, page) for page in pagerange]:
+                    response = future.result()
+                    self._process_entitlements(response)
 
-        logger.info('Retrieved %d course seats from %s.', count, self.partner.ecommerce_api_url)
+        logger.info('Retrieved %d course seats and % course entitlements from %s.', course_runs['count'],
```
staubina: Should that second % be a %d or a %s?

McKenzieW: oh good catch! It should be %d

staubina:
Overall looks good to me. We should test integration before merge.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -340,6 +359,43 @@ def update_seat(self, course_run, product_body):
...
+        except Currency.DoesNotExist:
+            msg = 'Could not find currency {}'.format(currency_code)
+            logger.warning(msg)
+            return None
+
+        mode_name = attributes.get('certificate_type')
+        try:
+            mode = SeatType.objects.get(name=mode_name)
+        except SeatType.DoesNotExist:
+            msg = 'Could not find course entitlement mode {}'.format(mode_name)
```
bderusha: Prefer named variables.  `'Could not find course entitlement mode {mode}'.format(mode=mode_name)`

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -340,6 +359,43 @@ def update_seat(self, course_run, product_body):
...
+            logger.warning(msg)
+            return None
+
+        defaults = {
+            'price': price,
+            'currency': currency,
+            'sku': sku
+        }
+        course.entitlements.update_or_create(mode=mode, defaults=defaults)
```
bderusha, line 399: Do we need to worry about cleaning up entitlements.  When we make a mistake and a professional mode entitlement is updated to be a verified entitlement the current code will result in us having both in catalog despite only having one in ecommerce.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -351,6 +353,45 @@ def mock_api(self):
...
+        bodies = [
+            {
+                "structure": "child",
+                "product_class": "Course Entitlement",
+                "price": "10.00",
+                "expires": None,
+                "attribute_values": [
+                    {
+                        "name": "certificate_type",
+                        "value": alt_mode if alt_mode else "verified",
```
bderusha, line 369: any reason not to put the default `verified` alt_mode in the method signature?

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -351,6 +353,45 @@ def mock_api(self):
...
+                    },
+                    {
+                        "name": "UUID",
+                        "value": alt_course if alt_course else str(course.uuid),
+                    }
+                ],
+                "is_available_to_buy": True,
+                "stockrecords": [
+                    {
+                        "price_currency": alt_currency if alt_currency else "USD",
```
bderusha, line 379: same question as above

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -351,6 +353,45 @@ def mock_api(self):
...
+                        "price_excl_tax": "10.00",
+                        "partner_sku": "sku132",
+                    }
+                ]
+            }
+        ]
+        url = self.api_url + 'products/'
```
bderusha: prefer `'{url}products/'.format...`

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -393,12 +434,35 @@ def assert_seats_loaded(self, body):
...
+            self.assertEqual(entitlement.sku, sku)
+
     @responses.activate
     def test_ingest(self):
         """ Verify the method ingests data from the E-Commerce API. """
-        api_data = self.mock_api()
-        loaded_course_run_data = api_data[:-1]
-        loaded_seat_data = api_data[:-2]
+        courses_api_data = self.mock_courses_api()
+        loaded_course_run_data = courses_api_data[:-1]
```
bderusha, line 463: A comment about what/why we are slicing here would be helpful to future devs

bderusha:
Some minor fixes, but one real question that we should know the answer to is how we handle cleaning up entitlements

bderusha:
@McKenzieW Here is an example of how we handle this for seats
https://github.com/edx/course-discovery/blob/1a7c216aa5b459a767424610c923b9410d92a8a6/course_discovery/apps/course_metadata/data_loaders/api.py#L326

course_discovery/apps/course_metadata/data_loaders/api.py
McKenzieW: I am currently working on the update that will include cleaning up old entitlements. Because of how the products endpoint is currently structured it would be quite a bit more complicated than the way we remove stale course seats, but we're going to update the endpoint to nest child products under their parents so that essentially the same procedure can be applied to entitlements as seats.

mjfrey (approved):
Remember to squash

PR #1204, opened 2017-11-02
Update the instructor API to provide just one profile image property

schenedx:
Make sure the API `profile_image` property reflect the profile image associated with instructor.

@clintonb Please review.
I am changing the API signature here. But I think no one is using this part of the API. So, I feel ok.
If you don't agree, I can add the `profile_image_url` property back into the API

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -247,7 +247,7 @@ class PositionSerializer(serializers.ModelSerializer):
 class PersonSerializer(serializers.ModelSerializer):
     """Serializer for the ``Person`` model."""
     position = PositionSerializer(required=False)
-    profile_image = StdImageSerializerField(required=False)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -247,7 +247,7 @@ class Meta(object):
 class PersonSerializer(serializers.ModelSerializer):
     """Serializer for the ``Person`` model."""
     position = PositionSerializer(required=False)
-    profile_image = StdImageSerializerField(required=False)
+    profile_image = serializers.SerializerMethodField()
```
clintonb: `profile_image = serializers.CharField(read_only=True, source='get_profile_image_url')`

See http://www.django-rest-framework.org/api-guide/fields/#source

clintonb (approved):

course_discovery/apps/api/serializers.py
schenedx: Nice. Thanks for the pointer.

PR #1206, opened 2017-11-03
(IT WILL NOT MERGE) WL-1295 | Added management command to create the sites and their part…

hasnain-naveed:
…ners.

### DIFF
--- a/course_discovery/apps/core/management/commands/create_partners_and_sites.py
+++ b/course_discovery/apps/core/management/commands/create_partners_and_sites.py
@@ -1,0 +1,89 @@
+""" Creates sites and partners """
+
+import logging
+
+from django.contrib.sites.models import Site
+from django.core.management import BaseCommand
+
+from course_discovery.apps.core.models import Partner
+
+logger = logging.getLogger(__name__)
+
+SITES = {
+    "mitxpro": {
+        "partner_code": "mitxpro",
+        "site": "mitxpro"
+    },
+    "hms": {
+        "partner_code": "hms",
+        "site": "hms"
+    },
+    "wharton": {
+        "partner_code": "wharton",
+        "site": "wharton"
+    },
+    "harvardx": {
+        "partner_code": "harvardx",
+        "site": "harvardx"
+    },
+
+}
+
+
+class Command(BaseCommand):
+    help = 'Create new sites and associated Partners.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dns-name',
+            action='store',
+            type=str,
+            dest='dns_name',
+            required=True,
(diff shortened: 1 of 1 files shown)


course_discovery/apps/core/management/commands/create_partners_and_sites.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+            ecommerce_api_url = "https://ecommerce-{site}-{dns_name}.sandbox.edx.org/".format(
+                site=site,
+                dns_name=dns_name
+            )
+            organizations_api_url = "https://{site}-{dns_name}.sandbox.edx.org/api/organizations/v0/".format(
+                site=site,
+                dns_name=dns_name
+            )
+            __, created = Partner.objects.update_or_create(
```
ibrahimahmed443, line 75: @hasnain-naveed please also add a partner for Site: "example.com". This site is created by default in the sandbox, so we need to add a partner, otherwise we will get a 500 error.

course_discovery/apps/core/management/commands/create_partners_and_sites.py
```diff
@@ -0,0 +1,89 @@
...
+
+            logger.info("dns_name: {dns_name}".format(dns_name=dns_name))
+
+            site_obj, created = Site.objects.get_or_create(
+                domain="{domain}-{dns_name}.sandbox.edx.org".format(domain=site, dns_name=dns_name),
+                defaults={"name": site}
+            )
+            studio_url = "https://studio-{dns_name}.sandbox.edx.org".format(dns_name=dns_name)
+            oidc_url_root = "https://{site}-{dns_name}.sandbox.edx.org/oauth2".format(site=site, dns_name=dns_name)
+            oidc_key = "{dns_name}-discovery-key".format(dns_name=dns_name)
```
ibrahimahmed443, line 61: @hasnain-naveed  Lets make the key and secret unique by appending site name.

course_discovery/apps/core/management/commands/create_partners_and_sites.py
```diff
@@ -0,0 +1,89 @@
...
+import logging
+
+from django.contrib.sites.models import Site
+from django.core.management import BaseCommand
+
+from course_discovery.apps.core.models import Partner
+
+logger = logging.getLogger(__name__)
+
+SITES = {
```
bill-filler, line 12: Same comment here that I left on https://github.com/edx/edx-platform/pull/16227/files
It might be best to move the site configurations into edx-themes in a separate file for each theme.
For example edx-themes/edx-platform/mitxpro.mit.edu/sandbox_config.json.
Then this code could iterate the themes dir and for each load the values in sandbox_config.json if they exist..

PR #1213, opened 2017-11-08
EDUCATOR-1705 Fix when profile_image is not available.

attiyaIshaque:
### Description
This PR fixes that when profile image of instructor is not available it would n't raises any error


course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
```diff
@@ -1,6 +1,6 @@
 <div class="instructor" id="instructor_{{ staff.id }}">
     <input type="hidden" id="staff_{{ staff.id }}" name="staff" value="{{ staff.id }}">
-    <div><img src="{% if staff.profile_image_url %}{{ staff.profile_image_url }}{% else %}{{ staff.profile_image.url }}{% endif %}"></div>
+    <div><img src="{% if staff.profile_image %}{{ staff.profile_image.url }}{% else %}{{ staff.profile_image_url }}{% endif %}"></div>
```
schenedx: You don't need the `if...else` statement here anymore. Please just use `{{ staff.get_profile_image_url }}`

schenedx (changes requested):
See my comment

attiyaIshaque:
@schenedx i have addressed your requested change.

schenedx (approved):
👍

PR #1218, opened 2017-11-13
Add instructor outside organization using org_override

asadazam93:
### Description
Allow instructors to be added outside of organization using the organization_override field.

**Sandbox**
- https://discovery-escalation-discovery.sandbox.edx.org/publisher/course_runs/5/edit/
Sandbox login. 
username: margaret
password: hamilton
User login
username : asadazam
password : asadazam

### Reviewers
- [ ] @schenedx    
- [x] @noraiz-anwar

### Post-review

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -238,7 +238,7 @@ class PositionSerializer(serializers.ModelSerializer):
 
     class Meta(object):
         model = Position
-        fields = ('title', 'organization_name', 'organization', 'organization_id')
+        fields = ('title', 'organization_name', 'organization', 'organization_id', 'organization_override')
         extra_kwargs = {
             'organization': {'write_only': True}
         }
@@ -301,7 +301,8 @@ class PersonSerializer(serializers.ModelSerializer):
         urls_data = validated_data.pop('urls', {})
 
         instance.position.title = position_data['title']
-        instance.position.organization = position_data['organization']
+        instance.position.organization = position_data.get('organization')
+        instance.position.organization_override = position_data.get('organization_override')
         instance.position.save()
 
         for url_type in [PersonSocialNetwork.FACEBOOK, PersonSocialNetwork.TWITTER, PersonSocialNetwork.BLOG]:
--- a/course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
+++ b/course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
@@ -9,82 +9,93 @@
             <fieldset class="form-group">
                 <div class="staff-image-icon">
                     <span>{% trans "Click here to upload your image" %}</span>
-                    <img src="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==" class="select-image" />
+                    <img src="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==" class="select-image"/>
                     <input name="staff-image" id="staffImageSelect" type="file" accept="image/*" onchange="loadSelectedImage(this);">
                 </div>
                 <ul>
-                <li><span class="required"> {% trans "The image dimensions must be less than 110 x 110" %}</span></li>
-                <li><span class="required"> {% trans "The image size must be less than 256KB" %}</span></li>
+                    <li><span class="required"> {% trans "The image dimensions must be less than 110 x 110" %}</span>
+                    </li>
+                    <li><span class="required"> {% trans "The image size must be less than 256KB" %}</span></li>
                 </ul>
                 <label class="field-label" for="given-name">{% trans "First Name" %}
                     <span class="required">* {% trans "required" %}</span>
                 </label>
-                <input class="field-input input-text" type="text" id="given-name" name="given-name" />
+                <input class="field-input input-text" type="text" id="given-name" name="given-name"/>
 
                 <label class="field-label" for="family-name">{% trans "Last Name" %}
                     <span class="required">* {% trans "required" %}</span>
                 </label>
-                <input class="field-input input-text" type="text" id="family-name" name="family-name" />
+                <input class="field-input input-text" type="text" id="family-name" name="family-name"/>
 
                 <label class="field-label" for="title">{% trans "Title" %}
                     <span class="required">* {% trans "required" %}</span>
                 </label>
-                <input class="field-input input-text" type="text" id="title" name="title" />
+                <input class="field-input input-text" type="text" id="title" name="title"/>
 
                 <label class="field-label" for="email">{% trans "Email" %}
                     <span class="required">* {% trans "required" %}</span>
                 </label>
                 <div class="email-field">
-                    <input class="field-input input-text" aria-describedby="email-hint" type="email" id="email" name="email" />
+                    <input class="field-input input-text" aria-describedby="email-hint" type="email" id="email"
+                           name="email"/>
                     <span id="email-hint">{% trans "Institution email is preferred" %}</span>
                 </div>
 
                 <label class="field-label">{% trans "Organization" %}
                     <span class="required">* {% trans "required" %}</span>
                 </label>
+                <div id="org_container">
+                    {% if organizations.count > 1 %}
+                        <select class="field-input input-select" id="id_organization" name="organization">
+                            <option value="------">-----</option>
+                            {% for organization in organizations %}
+                                <option value="{{ organization.id }}">{{ organization.name }}</option>
+                            {% endfor %}
+                        </select>
+                    {% else %}
+                        {% with organizations|first as organization %}
+                            <span class="read-only-field">{{ organization.name }}</span>
+                            <input id="id_organization" name="organization" type="hidden" value="{{ organization.id }}">
+                        {% endwith %}
+                    {% endif %}
+                </div>
+                <div id="org_override_container">
+                    <input class="field-input input-text" type="text" id="organization_override" name="organization_override"/>
+                </div>
+                <span id="override_toggle" class="read-only-field"> <a>{% trans "Enter a different organization" %}</a></span>
 
-                {% if organizations.count > 1 %}
-                    <select class="field-input input-select" id="id_organization" name="organization">
-                        <option value="------">-----</option>
-                        {% for organization in organizations %}
-                            <option value="{{ organization.id }}">{{ organization.name }}</option>
-                        {% endfor%}
-                    </select>
-                {% else %}
-                    {% with organizations|first as organization %}
-                        <span class="read-only-field">{{ organization.name }}</span>
-                        <input id="id_organization" name="organization" type="hidden" value="{{ organization.id }}">
-                    {% endwith %}
-                {% endif %}
 
                 <label class="field-label" for="bio">{% trans "Bio" %}
                     <span class="required">* {% trans "required (1-2 paragraphs)" %}</span>
                 </label>
-                <textarea class="field-textarea input-textarea textarea-medium textarea-block" id="bio" name="bio"></textarea>
+                <textarea class="field-textarea input-textarea textarea-medium textarea-block" id="bio"
+                          name="bio"></textarea>
 
                 <label class="field-label" for="facebook">{% trans "Facebook URL" %}
                     <span class="optional"> {% trans "optional" %}</span>
                 </label>
-                <input class="field-input input-text" type="text" id="facebook" name="facebook" />
+                <input class="field-input input-text" type="text" id="facebook" name="facebook"/>
 
                 <label class="field-label" for="twitter">{% trans "Twitter URL" %}
                     <span class="optional"> {% trans "optional" %}</span>
                 </label>
-                <input class="field-input input-text" type="text" id="twitter" name="twitter" />
+                <input class="field-input input-text" type="text" id="twitter" name="twitter"/>
 
                 <label class="field-label" for="blog">{% trans "Blog URL" %}
                     <span class="optional"> {% trans "optional" %}</span>
                 </label>
-                <input class="field-input input-text" type="text" id="blog" name="blog" />
+                <input class="field-input input-text" type="text" id="blog" name="blog"/>
 
                 <label class="field-label" for="majorWorks">{% trans "Major Works" %}
                     <span class="optional">({% trans "optional - one per line" %} )</span>
                 </label>
-                <textarea class="field-textarea input-textarea textarea-medium textarea-block" id="majorWorks" name="major-works"></textarea>
+                <textarea class="field-textarea input-textarea textarea-medium textarea-block" id="majorWorks"
+                          name="major-works"></textarea>
             </fieldset>
             <div class="actions">
                 <a class="btn-cancel closeModal" href="#">{% trans "Cancel" %}</a>
-                <button class="btn-brand btn-base btn-save" type="button" data-url="{% url 'api:v1:person-list' %}" id="add-instructor-btn">{% trans "Add Staff Member" %}</button>
+                <button class="btn-brand btn-base btn-save" type="button" data-url="{% url 'api:v1:person-list' %}"
+                        id="add-instructor-btn">{% trans "Add Staff Member" %}</button>
             </div>
         </form>
     </div>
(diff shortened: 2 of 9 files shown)


course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -9,82 +9,95 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
...
                 <div class="email-field">
-                    <input class="field-input input-text" aria-describedby="email-hint" type="email" id="email" name="email" />
+                    <input class="field-input input-text" aria-describedby="email-hint" type="email" id="email"
+                           name="email"/>
                     <span id="email-hint">{% trans "Institution email is preferred" %}</span>
                 </div>
 
                 <label class="field-label">{% trans "Organization" %}
                     <span class="required">* {% trans "required" %}</span>
+                    <span id="override_toggle"> <a>Override </a></span>
```
schenedx: I don't think the link should be named `Override`. How about `Manual Input` Or `Text Input`?
The link text here needs to be translated. Please use `{% trans "Manual Input" %}`

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -9,82 +9,95 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
...
                 </label>
+                <div id="org_container" style="display: block;">
```
schenedx: Please use class names instead of inline style like `display: block;`

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -9,82 +9,95 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
...
+                        </select>
+                    {% else %}
+                        {% with organizations|first as organization %}
+                            <span class="read-only-field">{{ organization.name }}</span>
+                            <input id="id_organization" name="organization" type="hidden" value="{{ organization.id }}">
+                        {% endwith %}
+                    {% endif %}
+                </div>
+                <div id="org_override_container" style="display: none;">
+                    <label class="field-label">{% trans "Organization Override" %}
```
schenedx: Instead of calling this text field `Organization Override`. It sounds better to just be `Organization`

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -9,82 +9,95 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
...
+                            {% endfor %}
+                        </select>
+                    {% else %}
+                        {% with organizations|first as organization %}
+                            <span class="read-only-field">{{ organization.name }}</span>
+                            <input id="id_organization" name="organization" type="hidden" value="{{ organization.id }}">
+                        {% endwith %}
+                    {% endif %}
+                </div>
+                <div id="org_override_container" style="display: none;">
```
schenedx: I think it's better to initiate the `display:none;` state of this element in javascript, rather than using inline style here

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
```diff
@@ -4,10 +4,12 @@
     <div><a class="delete" id="{{ staff.id }}" href="#"><i class="fa fa-trash-o fa-fw"></i></a>
         {% if course_user_role == 'course_team' %}
             {% if staff.position %}
-                {% if staff.position.organization_id in organizations_ids and staff.uuid %}
+                {% if staff.position.organization_id is None or staff.position.organization_id in organizations_ids and staff.uuid %}
                     <a class="edit" id="{{ staff.uuid }}" href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>
                 {% endif %}
             {% endif %}
+        {% elif staff.position.organization_id is None %}
```
schenedx: Are you sure this is the right place for this `elif` statement? Shouldn't this `elif` be after line 9?

schenedx: Or is this even needed?

course_discovery/static/js/publisher/instructors.js
```diff
@@ -156,10 +171,13 @@ function renderSelectedInstructor(id, name, image, uuid, organization_id, edit_i
 
 
     if (course_user_role == "course_team") {
-        if ($.inArray(parseInt(organization_id), JSON.parse(user_organizations_ids)) > -1 && uuid) {
+        if ($.inArray(parseInt(organization_id), JSON.parse(user_organizations_ids)) > -1 && uuid || organization_id == "None") {
             controlOptions += '<a class="edit" id="' + uuid + '"href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>';
         }
     }
+    else if (organization_id == "None") {
```
schenedx: This statement repeats the condition above

course_discovery/static/js/publisher/instructors.js
```diff
@@ -34,16 +39,26 @@ $(document).ready(function () {
             addModalError(gettext("Please upload a instructor image."));
             return;
         }
+        if ($('#organization_override').val()) {
```
schenedx: Please move this into a new function. Like `get_form_instructor_organization`

schenedx (changes requested):
Please see my inline review comments

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
asadazam93: This elif is for the case when the logged in user is not course team so it needs to be there for all internal users to be able to edit the instructor

course_discovery/static/js/publisher/instructors.js
asadazam93: The above condition is for course team user. This condition is for when someone other than the course team is logged in.

course_discovery/apps/api/serializers.py
```diff
@@ -301,7 +301,10 @@ def update(self, instance, validated_data):
         urls_data = validated_data.pop('urls', {})
 
         instance.position.title = position_data['title']
-        instance.position.organization = position_data['organization']
+        organization = position_data.get('organization')
+        organization_override = position_data.get('organization_override')
+        instance.position.organization = organization if organization else None
```
schenedx: I don't think you need those temporary variables.
You can do `instance.position.organization = position_data.get('organization')`directly, and that achieve the same.

course_discovery/apps/api/serializers.py
```diff
@@ -301,7 +301,10 @@ def update(self, instance, validated_data):
...
+        instance.position.organization_override = organization_override if organization_override else None
```
schenedx: This is the same thing. You just need one line to achieve the same result:
`instance.position.organization_override = position_data.get('organization_override')`

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
schenedx: @sstack22 Question:
Should we allow ANY internal user to edit instructors on Publisher?
Currently, we only allow course team members to do so. Should we open the access to edit up to others?

course_discovery/static/js/publisher/instructors.js
```diff
@@ -34,16 +41,15 @@ $(document).ready(function () {
             addModalError(gettext("Please upload a instructor image."));
             return;
         }
+        var position = getFormInstructorPosition($('#organization_override').val())
```
schenedx: Instead of passing in the value like this, just call the function without parameters, and call statement `$('#organization_override').val()` inside this method. You already do similar things in this method anyways. Let's make it consistent.

course_discovery/static/js/publisher/instructors.js
schenedx: I am really having trouble with this if else block.
To me, those conditions can be written as a separate function, and then we just have to write ` controlOptions += '<a class="edit" id="' + uuid + '"href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>';` once.
Also, why do we go through so many logic to render this edit icon anyways?

schenedx (changes requested):
Reviewed.
Maybe we can talk about the if else condition face to face?

asadazam93:
@schenedx I have addressed your comments. About the if condition maybe we can talk about it in the meeting today?

sstack22:
@asadazam93  - it seems there is a bug in the sandbox. I don't see a way to choose an existing organization - it looks like manual input is the only option. CC: @srpearce as an FYI and also for thoughts on text.

screen shot 2017-11-15 at 8 02 45 am

course_discovery/static/js/publisher/instructors.js
```diff
@@ -96,6 +102,19 @@ $(document).ready(function () {
     });
 });
 
+function getFormInstructorPosition () {
+    if ($('#organization_override').val()) {
```
schenedx, line 106: #nitpick This value is being used twice. Once here and once on line 109. Can you get this value into a temporary variable and use the variable instead on line 109?

asadazam93:
@sstack22 maybe because the user you are logging in with is not related to any organization. One can only create instructors for organizations you are related with.

srpearce:
@asadazam93 You mentioned that the username/password are asadazam/asadazam. Where do we enter this? When I sign in to the edX platform, I have to use an e-mail address (so I just used the standard staff@example.com with edx as the password).

srpearce:
I have a few suggestions for this UI. 

Here is what the UI label looks like currently before the user clicks **Manual Input**. There's no field or drop-down under **Organization**:
screenshot 47
**Note: This may be different if the user is signed in as someone who is a member of an organization. But the only way I know how to sign in is with the staff@example.com / edx credentials; I'm not aware of a place to enter just a username that isn't an email address.**

And here's what it looks like after you click **Manual Input**. There are two **Organization** labels but just one field:
screenshot 46

Again, I'm guessing that being able to sign in as asadazam will fix some of these problems.

However, for now I'll just say that I think the option should look like this before the user clicks **Enter a different organization**:
(Note that I cobbled these together with SnagIt, so they're not 100% what this should look like.)

screenshot 46a

Notes:
* **Manual Input** has changed to **Enter a different organization** and has moved under the drop-down list.
* **Required** has parentheses instead of an asterisk, and is in normal text instead of bold. (I'd recommend this change for all instances of "* required" on the page—it would look more professional.)

And I think it should look like this after the user clicks **Enter a different organization** (I've entered some sample text here):
screenshot 46b

Notes:
* The arrow that indicates the field is a drop-down list is gone because this is now a text field, and the cursor automatically appears inside the text field.
* The **Enter a different organization** option is no longer visible. (If we want the user to be able to enter another organization, we should keep this option visible.)

These are just suggestions; if you have any questions, please let me know.

asadazam93:
@srpearce You have to login via the admin. https://discovery-escalation-discovery.sandbox.edx.org/admin
I have Implemented your suggestions. Thank you for your input.

asadazam93:
@schenedx while testing that single if statement I discovered that we needed that None check for organization after all, because without that check the course team will not be able to edit instructors that are not related to any organization.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -96,6 +102,19 @@ $(document).ready(function () {
...
+        return {
+            title: $('#title').val(),
+            organization_override: $('#organization_override').val()
```
schenedx: We should also return the dictionary with organization set to null
we should return 
```
{
    title: $('#title').val(),
    organization_override: $('#organiztion_override').val(),
    organization: null
}
```
If don't do this, when course teams are editing an existing instructor who had organization specified before, to update it to have org override, we will have the instructor's organization and org override both specified, which will lead to confusion.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -96,6 +102,19 @@ $(document).ready(function () {
...
+        };
+    }
+    return {
+        title: $('#title').val(),
+        organization: parseInt($('#id_organization').val())
```
schenedx: You should make sure you explicitly set the organization override value to be `null` in this dictionary return

course_discovery/static/js/publisher/instructors.js
```diff
@@ -149,21 +168,19 @@ $(document).on('click', '.selected-instructor a.delete', function (e) {
...
-    }
-    else {
-        if (edit_instructor) {
-            controlOptions += '<a class="edit" id="' + uuid + '"href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>';
-        }
+    var user_is_course_team = course_user_role === "course_team";
+    var user_is_in_similar_org_as_instructor = $.inArray(parseInt(organization_id), JSON.parse(user_organizations_ids)) > -1;
+    var org_is_none = organization_id === "None";
+
+    if((user_is_course_team && user_is_in_similar_org_as_instructor && uuid || org_is_none) || is_internal_user || edit_instructor){
```
schenedx: I think the bracket should be:
`if(user_is_course_team && ((user_is_in_similar_org_as_instructor && uuid) || org_is_none) || is_internal_user...)`

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
```diff
@@ -2,12 +2,10 @@
...
     <div><img src="{{ staff.get_profile_image_url }}"></div>
     <div><a class="delete" id="{{ staff.id }}" href="#"><i class="fa fa-trash-o fa-fw"></i></a>
-        {% if course_user_role == 'course_team' %}
-            {% if staff.position %}
-                {% if staff.position.organization_id in organizations_ids and staff.uuid %}
-                    <a class="edit" id="{{ staff.uuid }}" href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>
-                {% endif %}
-            {% endif %}
+        {% if (course_user_role == 'course_team' and staff.position and staff.uuid and
+         staff.position.organization_id in organizations_ids or staff.position.organization_id is None) or is_internal_user  %}
```
schenedx: Again, can you please make sure the condition `staff.position.organization_id` is combined at the right place?
like 
```
if (course_user_role == 'course_team' and ((staff.position and staff.uuid and
         staff.position.organization_id in organizations_ids) or staff.position.organization_id is None))...
```

schenedx (changes requested):
See my comments
It makes sense we need the `org is null` check

course_discovery/static/js/publisher/instructors.js
asadazam93: Yeah you are right

asadazam93:
@schenedx I have addressed your comments.

schenedx (approved):
I still have a nitpick in the PR not addressed. But this is good to merge in my opinion. Good work!

noraiz-anwar (approved):
👍

course_discovery/static/js/publisher/instructors.js
```diff
@@ -149,21 +170,19 @@ $(document).on('click', '.selected-instructor a.delete', function (e) {
...
-        if ($.inArray(parseInt(organization_id), JSON.parse(user_organizations_ids)) > -1 && uuid) {
-            controlOptions += '<a class="edit" id="' + uuid + '"href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>';
-        }
-    }
-    else {
-        if (edit_instructor) {
-            controlOptions += '<a class="edit" id="' + uuid + '"href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>';
-        }
+    var user_is_course_team = course_user_role === "course_team";
+    var user_is_in_similar_org_as_instructor = $.inArray(parseInt(organization_id), JSON.parse(user_organizations_ids)) > -1;
```
schenedx, line 181: So, I was trying to validate this change on stage and found:
1. `user_organizations_ids` variable may contain a empty string. This means the statement `JSON.parse(user_organizations_ids)` can be throwing an exception
2. organization_id can be empty string as well. That means you don't even need to do `parseInt(organization_id)` if organization_id is empty

Please fix these two issues

srpearce:
Nit: I'm not sure if this is related to @schenedx 's comments above, but I see this behavior when I start to enter a different organization but then change my mind:

1. I create a new instructor.
2. I see this:  
screenshot 53
3. I click **Enter a different organization**. (Note: The pointer changes to a beam on hover, where I would expect it to change to a hand. The text becomes underlined on hover, though, so that's helpful.)
4. I see this:
screenshot 54
5. I decide I don't want to enter a new organization after all. I'm not sure how to get rid of the blank space so that I can see "Massachusetts Institute of Technology" again.

I played around with it and discovered that I can click **Enter a different organization** again. I can also click _near_ the **Enter a different organization** text to return to the original. I would expect to be able to click anywhere outside the **Organization** field, or to have a **Cancel** option.

PR #1219, opened 2017-11-13
Fix the edX status column sorting on the courses page.

attiyaIshaque:
### Description
This PR fixes the sorting on *edX status* colums on the publisher course page. It wouldn't load all the courses and the pagination would work correctly.

### How to Test?

**Production**
- https://prod-edx-discovery.edx.org/publisher/courses/
- Sort the column *edX status* 
- Observe it loads all the course and pagination does not work properly.

**Sandbox**
- https://discovery-escalation-discovery.sandbox.edx.org/publisher/courses/
- Sort the columns *example.com status* 
- Observe column are sorting properly and pagination works properly.

_username : Attiya
Password: attiya_

### Reviewers
- [x] Code review: @schenedx 
- [x] Code review: @asadazam93 

FYI: Tag anyone who might be interested in this PR here.

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -62,6 +62,9 @@ COURSE_ROLES.extend(DEFAULT_ROLES)
 COURSES_DEFAULT_PAGE_SIZE = 25
 COURSES_ALLOWED_PAGE_SIZES = (25, 50, 100)
 
+EDX_STATUS_COLUMN_INDEX = 6
+COURSE_RUNS_COUNT_INDEX = 3
+
 
 class CourseRunListView(mixins.LoginRequiredMixin, ListView):
     default_published_days = 30
@@ -832,10 +835,10 @@ class CourseListView(mixins.LoginRequiredMixin, ListView):
             ).values_list('organization')
             courses = courses.filter(organizations__in=organizations)
 
-        courses = self.sort_queryset(courses)
         courses = self.filter_queryset(courses)
+        courses = self.sort_queryset(courses)
 
-        return courses
+        return courses, len(courses)
 
     def get_context_data(self, **kwargs):
         context = super(CourseListView, self).get_context_data(**kwargs)
@@ -865,6 +868,7 @@ class CourseListView(mixins.LoginRequiredMixin, ListView):
             3: 'course_runs_count',
             4: 'course_state__owner_role_modified',
             5: 'course_state__owner_role_modified',
+            6: 'edx_status_column',
             7: 'course_state__owner_role_modified',
         }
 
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -870,44 +870,34 @@ def sort_queryset(self, queryset):
 
         try:
             ordering_field_index = int(self.request.GET.get('sortColumn', 0))
-            if ordering_field_index not in ordering_fields.keys():
+            if ordering_field_index not in ordering_fields.keys() and ordering_field_index != 6:
```
schenedx: why index != 6? What does the value 6 mean? Can you use a constant to specify what `6` means?

schenedx (changes requested):
Could you please create new unit tests in the CourseListViewTests tests in https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/tests/test_views.py#L1637?

course_discovery/apps/publisher/views.py
attiyaIshaque: Sure @schenedx i will change it.

attiyaIshaque:
@schenedx I have added the unit test in this PR https://github.com/edx/course-discovery/pull/1210/files#diff-7541c140671314e6639d6a0fe3f4c4dfR1917

I have just change the small implementation.

course_discovery/apps/publisher/views.py
```diff
@@ -62,6 +62,8 @@
 COURSES_DEFAULT_PAGE_SIZE = 25
 COURSES_ALLOWED_PAGE_SIZES = (25, 50, 100)
 
+INTERNAL_USER_STATUS = 6
```
schenedx: This variable should be named as `EDX_STATUS_COLUMN_INDEX` correct? Why is it `INTERNAL_USER_STATUS`?

course_discovery/apps/publisher/views.py
```diff
@@ -870,44 +872,34 @@ def sort_queryset(self, queryset):
...
         except ValueError:
             ordering_field_index = 0
 
         ordering_direction = self.request.GET.get('sortDirection', 'asc')
         ordering_field = ordering_fields.get(ordering_field_index)
 
-        if ordering_field == 'course_runs_count':
-            queryset = queryset.annotate(course_runs_count=Count('publisher_course_runs'))
+        if ordering_field_index != INTERNAL_USER_STATUS:
+            if ordering_field == 'course_runs_count':
```
schenedx: Why not use the index here as well? Why use the field name which is more likely change?

schenedx (commented):
Added more coments.

course_discovery/apps/publisher/views.py
attiyaIshaque: Oh, I don't know. But I will also change it into some constant. :)

course_discovery/apps/publisher/views.py
attiyaIshaque: Because it's calculated on the property `INTERNAL_USER_STATUS`. I will change it.

attiyaIshaque:
@schenedx I have addressed your comments.
@asadazam93 Can you please review it?

course_discovery/apps/publisher/views.py
```diff
@@ -878,36 +882,27 @@ def sort_queryset(self, queryset):
...
-        """
-         Ordering by internal user status
-        Args:
-            query_set:
-            context:
-        """
-        ordering_field_index = int(self.request.GET.get('sortColumn', 0))
-        ordering_direction = self.request.GET.get('sortDirection', 'asc')
-        if ordering_field_index == 6:
+        if ordering_field_index == EDX_STATUS_COLUMN_INDEX:
```
schenedx: Please add comment to let others know why we implemented special logic for EDX_STATUS column

course_discovery/apps/publisher/views.py
```diff
@@ -878,36 +882,27 @@ def sort_queryset(self, queryset):
...
-                context['object_list'] = sorted(query_set, key=lambda state: course_states.index(
+                queryset = sorted(queryset, key=lambda state: course_states.index(
                     str(state.course_state.internal_user_status if state.course_state.internal_user_status else '')))
             else:
-                context['object_list'] = sorted(query_set, key=lambda state: course_states.index(
+                queryset = sorted(queryset, key=lambda state: course_states.index(
                     str(state.course_state.internal_user_status if state.course_state.internal_user_status else '')),
                     reverse=True)
+        else:
+            if ordering_field_index == COURSE_RUNS_COUNT_INDEX:
```
schenedx: This if block can be moved to above line 885. Then it would make the logic easier to read

schenedx (approved):
Just minor comments. 
👍

course_discovery/apps/publisher/views.py
attiyaIshaque: If we move this line to above line 885
```
if ordering_field_index == COURSE_RUNS_COUNT_INDEX:
```
then this part not executed 
```
 if ordering_direction == 'asc':
                 queryset = queryset.order_by(Lower(ordering_field).asc())
else:
                queryset = queryset.order_by(Lower(ordering_field).desc())
```
so in my opinion it should remain in this part.

course_discovery/apps/publisher/views.py
attiyaIshaque: Okay sure i will add it.

course_discovery/apps/publisher/views.py
schenedx: I think you didn't fully understand me. 
If you move the `if ordering_field_index == COURSE_RUNS_COUNT_INDEX:` statement above line 885. 
Then let line 885 to be a single if statement again. Then the `else` where you are doing ordering directions, will be executed

attiyaIshaque: oh. Got it. Thanks i will do it.

asadazam93 (approved):

PR #1220, opened 2017-11-13
Default to populate max_effort with only 1 effort value present

McKenzieW:
When only one value  -- as opposed to a range -- is provided for estimated effort for a course run in Drupal, use this to populate max_effort for the course in Discovery instead of min_effort.

The presence of max_effort as opposed to min_effort in Discovery will allow for the calculation of Hours of Effort for course certificates having this feature enabled.

[LEARNER-3220](https://openedx.atlassian.net/browse/LEARNER-3220)


waheedahmed (approved):

waheedahmed:
@McKenzieW Zubair is out for two days you can ask someone else for review.

bderusha (approved):

clintonb (approved):
Why not update whichever client is using this data to fallback to `min_effort` if `max_effort` is not available?

McKenzieW:
@clintonb you're passing some mixed messages with PR approval and suggestion of an alternative approach

clintonb:
Acknowledged. I'm still curious as to whether that solution was explored.

McKenzieW:
It would involve caching extra data

PR #1221, opened 2017-11-13
LEARNER-2935 Boost current upgradeable runs over future runs whose start date is closer

MatthewPiatetsky:
[`LEARNER-2935` JIRA Ticket](https://openedx.atlassian.net/browse/LEARNER-2935)

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -6,33 +6,28 @@ from urllib.parse import urlencode
 import ddt
 import mock
 import pytest
-import pytz
 import responses
 from django.test import TestCase
 from django.utils.text import slugify
 from haystack.query import SearchQuerySet
 from opaque_keys.edx.keys import CourseKey
+from pytz import UTC
 from rest_framework.test import APIRequestFactory
 from waffle.models import Switch
 from waffle.testutils import override_switch
 
 from course_discovery.apps.api.fields import ImageField, StdImageSerializerField
-from course_discovery.apps.api.serializers import (AffiliateWindowSerializer, CatalogSerializer,
-                                                   ContainedCourseRunsSerializer, ContainedCoursesSerializer,
-                                                   CorporateEndorsementSerializer, CourseEntitlementSerializer,
-                                                   CourseRunSearchSerializer, CourseRunSerializer,
-                                                   CourseRunWithProgramsSerializer, CourseSearchSerializer,
-                                                   CourseSerializer, CourseWithProgramsSerializer,
-                                                   EndorsementSerializer, FAQSerializer,
-                                                   FlattenedCourseRunWithCourseSerializer, ImageSerializer,
-                                                   MinimalCourseRunSerializer, MinimalCourseSerializer,
-                                                   MinimalOrganizationSerializer, MinimalProgramCourseSerializer,
-                                                   MinimalProgramSerializer, NestedProgramSerializer,
-                                                   OrganizationSerializer, PersonSerializer, PositionSerializer,
-                                                   PrerequisiteSerializer, ProgramSearchSerializer, ProgramSerializer,
-                                                   ProgramTypeSerializer, SeatSerializer, SubjectSerializer,
-                                                   TypeaheadCourseRunSearchSerializer, TypeaheadProgramSearchSerializer,
-                                                   VideoSerializer, get_utm_source_for_user)
+from course_discovery.apps.api.serializers import (
+    AffiliateWindowSerializer, CatalogSerializer, ContainedCourseRunsSerializer, ContainedCoursesSerializer,
+    CorporateEndorsementSerializer, CourseEntitlementSerializer, CourseRunSearchSerializer, CourseRunSerializer,
+    CourseRunWithProgramsSerializer, CourseSearchSerializer, CourseSerializer, CourseWithProgramsSerializer,
+    EndorsementSerializer, FAQSerializer, FlattenedCourseRunWithCourseSerializer, ImageSerializer,
+    MinimalCourseRunSerializer, MinimalCourseSerializer, MinimalOrganizationSerializer, MinimalProgramCourseSerializer,
+    MinimalProgramSerializer, NestedProgramSerializer, OrganizationSerializer, PersonSerializer, PositionSerializer,
+    PrerequisiteSerializer, ProgramSearchSerializer, ProgramSerializer, ProgramTypeSerializer, SeatSerializer,
+    SubjectSerializer, TypeaheadCourseRunSearchSerializer, TypeaheadProgramSearchSerializer, VideoSerializer,
+    get_utm_source_for_user
+)
 from course_discovery.apps.api.tests.mixins import SiteMixin
 from course_discovery.apps.catalogs.tests.factories import CatalogFactory
 from course_discovery.apps.core.models import Partner, User
@@ -41,13 +36,11 @@ from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin, LMSAPIClientMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
-from course_discovery.apps.course_metadata.tests.factories import (CorporateEndorsementFactory, CourseFactory,
-                                                                   CourseRunFactory, EndorsementFactory,
-                                                                   ExpectedLearningItemFactory, ImageFactory,
-                                                                   JobOutlookItemFactory, OrganizationFactory,
-                                                                   PersonFactory, PositionFactory, PrerequisiteFactory,
-                                                                   ProgramFactory, ProgramTypeFactory, SeatFactory,
-                                                                   SeatTypeFactory, SubjectFactory, VideoFactory)
+from course_discovery.apps.course_metadata.tests.factories import (
+    CorporateEndorsementFactory, CourseFactory, CourseRunFactory, EndorsementFactory, ExpectedLearningItemFactory,
+    ImageFactory, JobOutlookItemFactory, OrganizationFactory, PersonFactory, PositionFactory, PrerequisiteFactory,
+    ProgramFactory, ProgramTypeFactory, SeatFactory, SeatTypeFactory, SubjectFactory, VideoFactory
+)
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
 
 
@@ -615,7 +608,7 @@ class MinimalProgramSerializerTests(TestCase):
 
         courses = CourseFactory.create_batch(3)
         for course in courses:
-            CourseRunFactory.create_batch(2, course=course, staff=[person], start=datetime.datetime.now(pytz.UTC))
+            CourseRunFactory.create_batch(2, course=course, staff=[person], start=datetime.datetime.now(UTC))
 
         return ProgramFactory(
             courses=courses,
@@ -730,21 +723,21 @@ class ProgramSerializerTests(MinimalProgramSerializerTests):
         CourseRunFactory(
             course=course_list[2],
             enrollment_start=None,
-            start=datetime.datetime(2014, 2, 1),
+            start=datetime.datetime(2014, 2, 1, tzinfo=UTC),
         )
 
         # Create a second run with matching start, but later enrollment_start.
         CourseRunFactory(
             course=course_list[1],
             enrollment_start=datetime.datetime(2014, 1, 2),
-            start=datetime.datetime(2014, 2, 1),
+            start=datetime.datetime(2014, 2, 1, tzinfo=UTC),
         )
 
         # Create a third run with later start and enrollment_start.
         CourseRunFactory(
             course=course_list[0],
-            enrollment_start=datetime.datetime(2014, 2, 1),
-            start=datetime.datetime(2014, 3, 1),
+            enrollment_start=datetime.datetime(2014, 2, 1, tzinfo=UTC),
+            start=datetime.datetime(2014, 3, 1, tzinfo=UTC),
         )
 
         program = ProgramFactory(courses=course_list)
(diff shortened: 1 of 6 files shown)


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -730,21 +723,21 @@ def test_course_ordering(self):
         CourseRunFactory(
             course=course_list[2],
             enrollment_start=None,
-            start=datetime.datetime(2014, 2, 1),
+            start=datetime.datetime(2014, 2, 1, tzinfo=UTC),
```
MatthewPiatetsky, line 726: these tests were erroring because in my property i was comparing an offset aware now = datetime.datetime.now(pytz.UTC) against an offset naive self.start.
So I'm pretty sure what we want here is to make these dates offset aware rather than comparing on the property against an offset naive date

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -171,6 +171,7 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     paid_seat_enrollment_end = indexes.DateTimeField(null=True)
     license = indexes.MultiValueField(model_attr='license', faceted=True)
     has_enrollable_seats = indexes.BooleanField(model_attr='has_enrollable_seats', null=False)
+    is_current_run = indexes.BooleanField(null=False)
```
MatthewPiatetsky: index the property

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
```diff
@@ -79,6 +79,20 @@ def get_elasticsearch_boost_config():
...
+                    'filter': {
+                        'bool': {
+                            'must': {
+                                'term': {
+                                    'is_current_run': True
+                                }
+                            }
+                        }
+                    },
+                    'weight': 20.0
```
MatthewPiatetsky: boost on presence of property, need to boost more than the start date boost to ensure this boost takes precedence

amangano-edx:
@MatthewPiatetsky can you post the JIRA ticket this is related to?

course_discovery/apps/course_metadata/models.py
```diff
@@ -484,6 +484,23 @@ def has_enrollable_paid_seats(self):
...
 
+    def is_current_run(self):
+        """
+        Return true if
+        1. Today is after the run start (or start is none) and two weeks from the run end (or end is none)
+        2. The verified upgrade deadline is in the future (or none)
+        and false otherwise
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        two_weeks = datetime.timedelta(days=14)
```
jaebradley: Nit: if you really want weeks you could do `datetime.timedelta(weeks=2)`. I don't think it matters much other than from a readability perspective (and even then, it doesn't make a _huge_ difference).

course_discovery/apps/course_metadata/models.py
```diff
@@ -484,6 +484,23 @@ def has_enrollable_paid_seats(self):
...
+        after_start = (not self.start) or (self.start and self.start < now)
+        ends_in_more_than_two_weeks = (not self.end) or (self.end.date() and now.date() <= self.end.date() - two_weeks)
```
jaebradley, line 497: Was it intentional to use date comparison instead of

```python
self.end - now > datetime.timedelta(weeks=2)
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -484,6 +484,23 @@ def has_enrollable_paid_seats(self):
...
+    def is_current_run(self):
+        """
+        Return true if
+        1. Today is after the run start (or start is none) and two weeks from the run end (or end is none)
+        2. The verified upgrade deadline is in the future (or none)
+        and false otherwise
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        two_weeks = datetime.timedelta(days=14)
+        after_start = (not self.start) or (self.start and self.start < now)
```
jaebradley, line 496: Nit: I find `has_started` implies a slightly clearer boolean condition (but again, that's just me...)

Also do we want to include the very unlikely case where the start time is exactly now?

```python
self.start <= now
```

course_discovery/apps/course_metadata/models.py
jaebradley: Also, you're using this variable in one place - you could consider just using `datetime.timedelta(weeks=2)` instead of assigning to a variable. But if you find it clearer, obviously not a huge deal.

course_discovery/apps/course_metadata/models.py
jaebradley: Won't this return `True` if `self.start` is not `None` and `self.start > now` since it would be `(True) or (False)` ?

Similar comment for `ends_in_more_than_two_weeks`.

MatthewPiatetsky:
@amangano-edx link to ticket added, thanks @jaebradley

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: if self.start is not None first condition will be false.
First condition will only be true if self.start is None (or some other falsey value like '' or false, but think None is the only one that would happen)
i don't think we need to include self.start

will change to has_started though

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: i found this easier to reason with to get my tests working an in general, but in practice i think this would give us roughly the same thing

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: i'll change it to weeks, but keep the variable

course_discovery/apps/course_metadata/models.py
jaebradley: whoops, yeah you're right, misread it

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
```diff
@@ -79,6 +79,20 @@ def get_elasticsearch_boost_config():
...
+                        'bool': {
+                            'must': {
+                                'term': {
+                                    'is_current_run': True
+                                }
+                            }
+                        }
+                    },
+                    'weight': 10.0
+                },
```
amangano-edx, line 94: I don't think we should be adding more config for every new thing that comes up. Looking at this file, we have several config entries that are similar to this one, and it is nearly impossible to determine from reading it whether or not what you've added will accomplish the goals in https://openedx.atlassian.net/wiki/spaces/LEARNER/pages/106201185/Controlling+Relevance.

Also, the weights used in this boost, and other more recent additions to this config, seem to be a lot higher than the earlier ones. The elasticsearch docs caution against boosting too much as it can throw off its own relevancy rankings (for things like keywords) and lead to irrelevant search results.

I would encourage you to take a long look at this file and determine if we really need another boosting config, or if the existing configs could be reworked to be more in line with what product actually wants.

course_discovery/apps/course_metadata/models.py
```diff
@@ -484,6 +484,23 @@ def has_enrollable_paid_seats(self):
         """
         return len(self._enrollable_paid_seats()[:1]) > 0
 
+    def is_current_run(self):
```
amangano-edx: I would prefer to use a different name for this property, as the term 'Current' is pretty loaded throughout our application.

MatthewPiatetsky: good point, i believe current refers to just now being between start and end from how i understand it.
what if i rename it to, is_current_and_still_upgradeable?
or were you suggesting not using the term current altogether
in that case i could use
has_started_not_ended_and_still_upgradeable

course_discovery/apps/api/tests/test_serializers.py
amangano-edx: Are all of the production start dates offset aware? If not, will these changes cause any issues?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -284,6 +285,32 @@ def test_get_paid_seat_enrollment_end(self, seat_config, course_end, course_enro
         expected_result = parse(expected_result) if expected_result else None
         self.assertEqual(course_run.get_paid_seat_enrollment_end(), expected_result)
 
+    now = datetime.datetime.now(pytz.timezone('utc'))
+    one_month = relativedelta(months=1)
+    two_weeks = relativedelta(days=14)
+
+    @ddt.data(
+        (None, None, None, False),
```
amangano-edx, line 293: Why is `is_current` expected to be `False` in this case? The conditions in which it should be `True`, according to the test docstring, are:
- `start < now or start is None`
- `end - two_weeks > now or end is None`
- `deadline > now or deadline is None`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -284,6 +285,32 @@ def test_get_paid_seat_enrollment_end(self, seat_config, course_end, course_enro
...
+        (now - one_month, None, None, False),
```
amangano-edx, line 294: It also seems like `is_current` should be `True` for this case.

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
```diff
@@ -79,6 +79,20 @@ def get_elasticsearch_boost_config():
                     'weight': 5.0
                 },
 
+                # Reward course runs that are currently running and still upgradeable
```
amangano-edx, line 82: This field actually uses `paid_seat_enrollment_end`, not upgrade deadline as this comment implies.

course_discovery/apps/edx_haystack_extensions/tests/test_boosting.py
```diff
@@ -167,3 +168,72 @@ def test_enrollable_course_run_boosting(self, enrollment_start, enrollment_end,
...
+    )
+    def test_current_run_boosting(self, runadates, runbdates, pacing_type, boosted):
+        """Verify that "current" CourseRuns are boosted.
+        See the is_current_run CourseRun property to understand what this means."""
+
+        (starta, enda) = runadates
+        (startb, endb) = runbdates
+
+        now = datetime.datetime.now(pytz.timezone('utc'))
+        upgrade_deadline_tomorrow = now + relativedelta(days=1)
```
amangano-edx, line 216: Nit: you could use the `one_day` variable here that you created above.

course_discovery/apps/course_metadata/tests/test_models.py
MatthewPiatetsky: the first two statements are or, but the last one is an and, rather than an or, so it is false for the last statement.
the code is right, but the docstring is wrong, I will fix that. this is because if get_paid_seat_enrollment_end returns none, that doesn't mean the run is upgradeable

course_discovery/apps/course_metadata/tests/test_models.py
MatthewPiatetsky: same reason as above

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
MatthewPiatetsky: good point, i will fix this

course_discovery/apps/edx_haystack_extensions/elasticsearch_boost_config.py
MatthewPiatetsky: i believe we discussed in hipchat how this boost differs from the others, but as for the weight, it needs to be higher than the weight for the start date boosting, and for the enrollable paid seat boosting, this boost will apply to a proper subset of the documents that receive that boost

course_discovery/apps/edx_haystack_extensions/tests/test_boosting.py
MatthewPiatetsky: will change

MatthewPiatetsky:
@amangano-edx i believe i've addressed most of your comments
@amangano-edx / @jaebradley do you have further comments?

course_discovery/apps/edx_haystack_extensions/tests/test_boosting.py
MatthewPiatetsky: oops i changed it but realized one_day variable is not in scope in this function so changed it back

jaebradley (approved):
Looks good to me

course_discovery/apps/course_metadata/models.py
```diff
@@ -484,6 +484,23 @@ def has_enrollable_paid_seats(self):
         """
         return len(self._enrollable_paid_seats()[:1]) > 0
 
+    def is_current_and_still_upgradeable(self):
+        """
+        Return true if
+        1. Today is after the run start (or start is none) and two weeks from the run end (or end is none)
+        2. The run has a seat that can still be upgraded
+        and false otherwise
```
amangano-edx: I think it would be good for this to specify that the run must still be enrollable as well.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -284,6 +285,31 @@ def test_get_paid_seat_enrollment_end(self, seat_config, course_end, course_enro
...
+        (now - one_month, now + one_month, now + two_weeks, True),
+        (now + one_month, now + one_month, now + two_weeks, False),
+    )
+    @ddt.unpack
+    def test_is_current_and_still_upgradeable(self, start, end, deadline, is_current):
+        """
+        Verify that is_current_and_still_upgradeable returns true if
+        1. Today is after the run start (or start is none) and two weeks from the run end (or end is none)
+        2. The run has a seat that can still be upgraded
+        and false otherwise
```
amangano-edx: This should clarify that it also needs to be enrollable.

course_discovery/apps/course_metadata/models.py
MatthewPiatetsky: fixed

course_discovery/apps/course_metadata/tests/test_models.py
MatthewPiatetsky: fixed

amangano-edx (approved):
The code changes here look good for implementing the new boosting logic, but the config itself is complex and I cannot tell from reading this whether or not it will do what we want. Please make sure to test thoroughly.

PR #1224, opened 2017-11-16
Transfer the course image from publisher to discovery course

schenedx:
Currently, the course image saved into discovery course object has no small variations. This is to fix that.
@clintonb Please help review

### DIFF
--- a/course_discovery/apps/publisher/api/v1/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/v1/tests/test_views.py
@@ -9,6 +9,7 @@ from rest_framework.test import APITestCase
 
 from course_discovery.apps.core.models import Currency, Partner
 from course_discovery.apps.core.tests.factories import StaffUserFactory, UserFactory
+from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.utils import serialize_datetime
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -39,13 +41,16 @@ def test_without_authorization(self):
     def _create_course_run_for_publication(self):
         organization = OrganizationFactory()
         transcript_languages = [LanguageTag.objects.first()]
-        return CourseRunFactory(
+        mock_image_file = make_image_file('test_image.jpg')
+        test_course_run = CourseRunFactory(
```
clintonb: Nit: no need for a variable; just return.

clintonb (approved):

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -8,6 +8,7 @@
 from rest_framework.test import APITestCase
 
 from course_discovery.apps.core.models import Currency, Partner
+from course_discovery.apps.core.tests.helpers import make_image_file
```
clintonb: Alphabetize.

PR #1229, opened 2017-11-20
Display verified seat if it exists over audit

asadazam93:
### Description
If a course run has more than one seat and if one of those is a paid seat, the paid seat should be displayed over the audit seat in publisher.

**Sandbox**
N/A

### Reviewers
- [x] @schenedx    
- [ ] @noraiz-anwar

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -2837,6 +2837,17 @@ class CourseRunEditViewTests(SiteMixin, TestCase):
 
         toggle_switch('enable_publisher_email_notifications', True)
 
+    def test_edit_page_with_two_seats(self):
+        """
+        Verify that if a course run has both audit and verified seats, Verified seat is displayed
+        on the course run edit page
+        """
+        factories.SeatFactory(course_run=self.course_run, type=Seat.AUDIT)
+        self.edit_page_url = reverse('publisher:publisher_course_runs_edit', kwargs={'pk': self.course_run.id})
+        response = self.client.get(self.edit_page_url)
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(response, '<div id="SeatPriceBlock" class="col col-6 hidden" style="display: block;">')
+
     def _post_data(self, data, course, course_run):
         course_dict = model_to_dict(course)
         course_dict.update(**data)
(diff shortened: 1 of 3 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -718,7 +718,11 @@ def get(self, request, *args, **kwargs):
         context['run_form'] = self.run_form(
             instance=course_run, is_project_coordinator=context.get('is_project_coordinator')
         )
-        context['seat_form'] = self.seat_form(instance=course_run.seats.first())
+        course_run_paid_seat = course_run.seats.filter(type__in=['verified', 'professional']).first()
```
schenedx: Can you move the list `['verified', 'professional']` into the Publisher Models.py at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/models.py#L34 with a constant name? This way, this list can be re-used.

schenedx (changes requested):
Please consider a unit test case for this scenario?

asadazam93:
@schenedx I have addressed your comment related to the code. Working on test case

schenedx (commented):
Please add unit tests

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2837,6 +2837,17 @@ def setUp(self):
...
+    def test_edit_page_with_two_seats(self):
+        """
+        Verify that if a course run has both audit and verified seats, Verified seat is displayed
+        on the course run edit page
+        """
+        factories.SeatFactory(course_run=self.course_run, type=Seat.AUDIT)
+        self.edit_page_url = reverse('publisher:publisher_course_runs_edit', kwargs={'pk': self.course_run.id})
+        response = self.client.get(self.edit_page_url)
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(response, '<div id="SeatPriceBlock" class="col col-6 hidden" style="display: block;">')
```
schenedx, line 2849: Shouldn't you check the seat has the name `Audit`?

schenedx: And can you point me to the test case where we have the verified seat and the publisher does show `Verified`?

schenedx (commented):
Almost there. I think we need one more test?

course_discovery/apps/publisher/tests/test_views.py
asadazam93: @schenedx this is that case. We have both verified and audit seats but the response renders verified seat. And here I am checking if the price is not hidden which would indicate that it is a verified seat thus has a price. We pass seat_form from the context so I was not able to check seat type that is why I used this to check for verified seat.

schenedx (approved):
👍

PR #1232, opened 2017-11-22
Fix server error on edit run

asadazam93:
## [EDUCATOR-1795](https://openedx.atlassian.net/browse/EDUCATOR-1795)
Fixed server error on the edit run page.

Code Review:
@awaisdar001 @noraiz-anwar


noraiz-anwar (approved):
👍

noraiz-anwar:
We need to cover these all with tests so that such errors can be caught before hand in future.

awaisdar001:
ok

awaisdar001:
@asadazam93 
Make sure you test the changes manually, given the time sensitivity, I am going to 👍 the PR.

For the future, I would like to see the `can_edit` logic at one place rather than being spread in JS and backend separately. We can introduce a model property say `can_edit` and move the backend logic there rather than in template. 
Secondly, when we create or update an instructor, we should also serialize the `can_edit` value and pass it to the frontend to only check for a variable `can_edit` rather than what is happening right now.

awaisdar001 (approved):
👍

PR #1234, opened 2017-11-22
LEARNER-3242: Implement Topics for Topic Landing Pages

jaebradley:
## Introduction
* [Subtask Ticket - `LEARNER-3242`](https://openedx.atlassian.net/browse/LEARNER-3242)
* [Main Ticket - `LEARNER-3054`](https://openedx.atlassian.net/browse/LEARNER-3054)

We are trying to implement Topic landing pages in a similar vein to how Subject landing pages are currently implemented, which is why the Topic model logic is nearly identical to the existing Subject model logic.

### Added Topics via the Admin Interface
image

### Topics `GET` Request via API
image
image


MatthewPiatetsky (approved):
In general i wouldn't add language_code since we're not using it
however, since we will likely use it in the future this is fine.
we discussed some changes you need to make in person, so probably looks good, but lets test it together before merging

mikedikan (commented):
LGTM

jaebradley:
### Tested on Stage
image

image

course_discovery/apps/course_metadata/models.py
```diff
@@ -153,6 +153,44 @@ class Meta:
         verbose_name = _('Subject model translations')
 
 
+class Topic(TranslatableModel, TimeStampedModel):
+    """ Topic model. """
+    uuid = models.UUIDField(blank=False, null=False, default=uuid4, editable=False, verbose_name=_('UUID'))
+    banner_image_url = models.URLField(blank=True, null=True)
```
BenjiLee, line 159: In case there were plans to to have different images for different languages (I doubt there is), this should be moved to TopicTranslation

PR #1235, opened 2017-11-23
pre fill organization edit instructor

asadazam93:
### Description
Edit instructor form now has pre filled organization for both organization and organization_override.
Internal users will be able to edit all existing instructors.

**Sandbox**
N/A

### Reviewers
- [x] @Rabia23    
- [x] @noraiz-anwar

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
+++ b/course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
@@ -6,7 +6,7 @@
             {% if staff.position and staff.uuid and staff.position.organization_id in organizations_ids or staff.position.organization_id is None %}
                 <a class="edit" id="{{ staff.uuid }}" href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>
             {% endif %}
-        {% elif not staff.profile_image_url %}
+        {% elif is_internal_user %}
             <a class="edit" id="{{ staff.uuid }}" href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>
         {% endif %}
         <b>{{ staff.full_name }}</b>
--- a/course_discovery/static/js/publisher/instructors.js
+++ b/course_discovery/static/js/publisher/instructors.js
@@ -271,12 +271,19 @@ $(document).on('click', '.selected-instructor a.edit', function (e) {
     $.getJSON({
         url: btnInstructor.data('url') + uuid,
         success: function (data) {
+            if (data['position']['organization_id'] == null){
+                $('#organization_override').val(data['position']['organization_override']);
+                $('#org_container').hide();
+            }
+            else {
+                $('#id_organization').val(data['position']['organization_id']);
+                $('#org_override_container').hide()
+            }
             $('.select-image').attr('src', data['profile_image_url']);
             $('#given-name').val(data['given_name']);
             $('#family-name').val(data['family_name']);
             $('#title').val(data['position']['title']);
             $('#email').val(data['email']);
-            $('#id_organization').val(data['position']['organization_id']);
             $('#bio').val(data['bio']);
             $('#majorWorks').val(data['works'].join('\n'));
             $('#facebook').val(data['urls']['facebook']);


course_discovery/static/js/publisher/instructors.js
```diff
@@ -271,12 +271,19 @@ $(document).on('click', '.selected-instructor a.edit', function (e) {
     $.getJSON({
         url: btnInstructor.data('url') + uuid,
         success: function (data) {
+            if (data['position']['organization_id'] == null){
```
noraiz-anwar, line 274: Are we sure that `data['position']['organization_id']` always exists? what if `data` has no key `position` or `organization_id` at all? 
Can you please add a docstring to clarify why we are explicitly checking for `null`?

noraiz-anwar, line 274: How about moving this logic to backend and having a single container here, lets say `org_container`? this container will now simply show any value passed.

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
```diff
@@ -6,7 +6,7 @@
             {% if staff.position and staff.uuid and staff.position.organization_id in organizations_ids or staff.position.organization_id is None %}
                 <a class="edit" id="{{ staff.uuid }}" href="#"><i class="fa fa-pencil-square-o fa-fw"></i></a>
             {% endif %}
-        {% elif not staff.profile_image_url %}
```
noraiz-anwar, line 9: I am surprised by this line.

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
```diff
@@ -6,7 +6,7 @@
...
+        {% elif is_internal_user %}
```
noraiz-anwar, line 9: Is there any check on backend to restrict/permit editing to required users?

noraiz-anwar (changes requested):

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
asadazam93: If the logged in user is an internal we want him/her to be able to edit instructors

course_discovery/static/js/publisher/instructors.js
asadazam93: Yes both position and organization_id would always exist. And we set it to null while creating an instructor so there is no doubt about that as well.

asadazam93: we want both the org_container and organization_override so that we can toggle between them if we want.

asadazam93:
@noraiz-anwar The instructor js code does need a lot of tweaks and we should create a ticket for that and maybe do it later. For this ticket we just need to pre fill organization while editing instructors.

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
noraiz-anwar: So, is there any check on edit request at backend? this check would only hide link to edit from front end, right?

course_discovery/static/js/publisher/instructors.js
noraiz-anwar: If there is not any difference at UI level, my suggestion is to move this logic to backend. We can toggle that at backend reducing a bunch of lines at front end.

asadazam93: You are right we can do that. That would require a lot of changes to be made and is out of scope for this ticket.

course_discovery/apps/publisher/templates/publisher/_personFieldLabel.html
asadazam93: The backend does exists and checks for model permissions. That is a separate thing. This check is just to display the edit button on frontend and all internal users should be able to edit instructors.

noraiz-anwar (approved):
👍 
Please rebase before merging.

Rabia23:
LGTM 👍

Rabia23 (approved):

PR #1237, opened 2017-11-27
use slug for entitlement mode serialization

McKenzieW:
So that course entitlement modes do not have to be converted to lower case everywhere that they are used programatically, the serializer should use slug rather than name for the related field.

This issue became evident in work on [LEARNER-3082](https://openedx.atlassian.net/browse/LEARNER-3082) in [this PR](https://github.com/edx/edx-platform/pull/16619)


staubina (approved):

amangano-edx (approved):
LGTM as long as we're sure this isn't currently being used anywhere.

PR #1240, opened 2017-11-28
Short Description Source Fix

schenedx:
Source the proper short description from Marketing Site course about pages into discovery service
LEARNER-3379
@edx/learner Please review

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -464,7 +464,8 @@ class CourseMarketingSiteDataLoader(AbstractMarketingSiteDataLoader):
             'mobile_available': data.get('field_course_enrollment_mobile') or False,
             'video': course.video,
             'course': course,
-            'short_description_override': self.clean_html(data['field_course_sub_title_short']) or None,
+            # We want to consume the same value for the override here to stay consistent with the marketing site
+            'short_description_override': self.clean_html(data['field_course_sub_title_long']['value']) or None,
             'min_effort': min_effort,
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -489,7 +489,7 @@ def format_course_data(self, data, key=None):
             'number': data['field_course_code'],
             'full_description': self.get_description(data),
             'video': self.get_video(data),
-            'short_description': self.clean_html(data['field_course_sub_title_short']),
+            'short_description': self.clean_html(data['field_course_sub_title_long']['value']),
```
mikedikan, line 493: why the same value for the override?  can you leave a comment explaining?  Also, why traverse the array deeper ['value']?

schenedx: @mikedikan We always consumed the same value from marketing site for both the course level short description and course_run level short description override. See `title` for example. As for why we want to traverse the array deeper, it's because drupal's json return only store the actual short description in the `value`. If we don't reference it deeper, we will not get the correct data member. See `title` for example again

mikedikan: ok, that makes sense for ['value'] - can you leave a small comment on what you explained to me for override?

mikedikan (approved):
One small request for a code comment, otherwise 👍

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
schenedx: Done.

PR #1245, opened 2017-12-01
Stop updating title of the course based on the data from Studio

schenedx:
EDUCATOR-1818
The title of the studio is not source of truth. The marketing site data is the source of truth for Course Title


schenedx:
@clintonb @mikedikan Could one of you provide a quick code review?

schenedx:
FYI @edx/learner

PR #1248, opened 2017-12-05
We should also publish the video to the discovery course run

schenedx:
It's annoying for the Discovery service users, like content managers from marketing, to try to select the video from the django admin dropdown.
We should just copy those video from publisher to the course_run level as well.
@clintonb Please review

### DIFF
--- a/course_discovery/apps/publisher/api/v1/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/v1/tests/test_views.py
@@ -139,6 +139,7 @@ class CourseRunViewSetTests(APITestCase):
         discovery_course_run = CourseRun.objects.get(key=publisher_course_run.lms_course_id)
         publisher_course = publisher_course_run.course
         discovery_course = discovery_course_run.course
+        publisher_course_video = Video.objects.get(src=publisher_course.video_link)
 
         # pylint: disable=no-member
         assert discovery_course_run.title_override == publisher_course_run.title_override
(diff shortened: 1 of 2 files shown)


clintonb:
The video is associated with the course, not course run. Why is this change necessary?

schenedx:
@clintonb The problem is that, video link exists on both the course and course run object in discovery. When we are sending the data to marketing, I chose to use the video URL from the course_run object. See https://github.com/edx/edx-mktg/blob/master/docroot/sites/all/modules/custom/edx/customizable_search/customizable_search.module#L656. However, publisher only have video stored on the Course level. https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/models.py#L89. Therefore, I figured, if we want to publish the video, we should publish the video to both course and course run.
You can argue that video should be a pure course_run level object, and I would agree. That would require the change on Publisher side. And that is too big for this PR.

clintonb:
Actually, my argument is to push more data to the course object since that is ultimately what we want to be marketing to people. This is simply further kicking the can down the road.

clintonb:
I'm tired of kicking the can down the road. The code is fine. Feel free to merge, but I'm no longer providing explicit approval for these types of changes.

schenedx:
@clintonb After talking with Marketing members, I made another PR: https://github.com/edx/course-discovery/pull/1249
Do you like the above change better?

PR #1249, opened 2017-12-05
Allow video object override on course_run level

schenedx:
This would allow the course level video object to be surfaced to the course_run level for display.


course_discovery/apps/api/serializers.py
```diff
@@ -508,6 +508,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
     )
     transcript_languages = serializers.SlugRelatedField(many=True, read_only=True, slug_field='code')
     video = VideoSerializer()
+    video_descriptor = VideoSerializer()
```
clintonb: Use the existing field instead of making a new one. Have the existing field point to your new property.

clintonb (commented):
I like this change much better!

clintonb (approved):
Add a test for the new model property.

clintonb (approved):
Remember to squash the commits.

PR #1251, opened 2017-12-06
Leave Publisher modified time stamp unchanged with import_courses management command

schenedx:
We do not want to change the modified time stamp because the "published date" in Publisher associated with course_run is based on the modified time stamp.
EDUCATOR-1923


clintonb (changes requested):
Why is the modified timestamp used to track the time the course run was published? It seems this should be an explicit field, not associated with other operations on the model.

schenedx:
@clintonb You are right. This change is not needed. I read the code at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/templates/publisher/dashboard/_published.html#L52 wrong

PR #1253, opened 2017-12-07
update entitlement product data loading from ecommerce

McKenzieW:
- Load entitlement mode by slug not by name
- Add partner as a field that is loaded
- Log errors with more detail

[LEARNER-3485](https://openedx.atlassian.net/browse/LEARNER-3485)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -311,7 +311,7 @@ def _process_entitlements(self, response):
         for body in results:
             body = self.clean_strings(body)
             skus.append(self.update_entitlement(body))
-        CourseEntitlement.objects.exclude(sku__in=skus).delete()
+        CourseEntitlement.objects.filter(partner=self.partner).exclude(sku__in=skus).delete()
```
mjfrey: Can we logs which skus we are deleting?  This will prove helpful in the future.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -366,39 +366,49 @@ def update_seat(self, course_run, product_body):
...
             return None
 
         defaults = {
+            'partner': self.partner,
             'price': price,
             'currency': currency,
             'sku': sku,
             'expires': self.parse_date(body['expires'])
         }
+        msg = 'Creating {entitlement} with sku {sku}'.format(entitlement=body['title'], sku=sku)
```
mjfrey: I would add the partner here as well.

mjfrey (changes requested):

course_discovery/apps/course_metadata/data_loaders/api.py
McKenzieW: I think we'd have to perform the query a second time to log the skus of the entitlements that are being deleted.

McKenzieW: I'm trying something out that could avoid an extra query, but I'm not positive that it will work

mjfrey: I just think it would be nice to log the entitlements we are deleting.

McKenzieW: Yeah, we can do that. I'm about to push the change up

McKenzieW: well I'm going to push the change up once the quality checks are done running locally

mjfrey (approved):

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -311,7 +311,13 @@ def _process_entitlements(self, response):
         for body in results:
             body = self.clean_strings(body)
             skus.append(self.update_entitlement(body))
-        CourseEntitlement.objects.exclude(sku__in=skus).delete()
+        entitlements_to_delete = CourseEntitlement.objects.filter(partner=self.partner).exclude(sku__in=skus)
+        for entitlement in entitlements_to_delete:
```
staubina, line 315: Nit: Is it possible for this list to be large?  Should we just log them all at once rather than looping through the list?

staubina (commented):
One nit otherwise looks good to me.

course_discovery/apps/course_metadata/data_loaders/api.py
McKenzieW: I don't imagine that it would be large, because entitlements would only likely need to be deleted if they were incorrectly created in ecommerce, then loaded into discovery before they were fixed in ecommerce, which I assume wouldn't happen in large batches or even very often.

Mjloturco: If we are assuming it wont happen in large batches or very often, would it be helpful to somehow log how many course entitlements were being deleted? It may be overkill but could help us track bugs like what we saw down. Something like 'deleted 20 entitlements for partner ____ '. thoughts?

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -402,6 +404,17 @@ def mock_products_api(self, alt_course=None, alt_currency=None, alt_mode=None):
...
 
+    def compose_warning_log(self, alt_course, alt_currency, alt_mode):
+        msg = 'Could not find '
+        if alt_course:
+            msg += 'course ' + alt_course
+        elif alt_currency:
+            msg += 'currency ' + alt_currency
+        elif alt_mode:
+            msg += 'course entitlement mode ' + alt_mode
+        msg += ' while loading Verified Entitlement to Fake Course with sku sku132'
```
Mjloturco: Nit: Course Entitlement products are named 'Course {course name}' instead of 'Verified Entitlement to {course name}'.

course_discovery/apps/course_metadata/migrations/0072_courseentitlement_partner.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ('core', '0007_auto_20171004_1133'),
+        ('course_metadata', '0071_auto_20171128_1945'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='courseentitlement',
+            name='partner',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, to='core.Partner'),
```
Mjloturco: Are we sure we want to delete any course_entitlements if the corresponding partner is deleted? I'm not sure myself, but we might want to consider.

Mjloturco (commented):
Mostly nits. A thing or two we might want to briefly consider.

course_discovery/apps/course_metadata/data_loaders/api.py
McKenzieW: I'm not quite clear on why the expectation against large batches and frequency indicates a potential usefulness of logging a count of the deleted course entitlements.

PR #1256, opened 2017-12-08
The image served from course and people object should have updated names

schenedx:
We need a new name/path for the new image updated to these objects because cache would keep the old image alive all the time
LEARNER-3455


schenedx:
@edx/learner Please review

mjfrey (approved):

PR #1258, opened 2017-12-08
WIP: Remove our custom codecov file.

feanil:
This is to see if that is causing our reporting to not work correctly.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,9 +1,0 @@
-comment: off
-coverage:
-  status:
-    patch:
-      default:
-        target: 95
-    project:
-      default:
-        target: 95


feanil:
@mjfrey @tuchfarber it looks like the problem is probably in our custom codecov file. Removing it made, codecov report again.  That's probably not the right move, since we want to keep our custom coverage limits, right?  Probably just updating the file will do?  @mjfrey are you the closest to an owner for this repo right now?

mjfrey:
@feanil are you sure this is the fix?  I am seeing it randomly work on PRs.  When it stops working we close the PR and open a new one.  That seems to work.

feanil:
Actually nevermind,  it looks like the issue was on their side and have been resolved.  https://github.com/edx/course-discovery/pull/1259  No changes on this other PR and it seems to be working correctly.

PR #1260, opened 2017-12-11
We don't have to import initial course run into publisher

schenedx:
Provide the option to the command to create or not create canonical_course_run
This will help the import of the course data into Publisher

EDCUATOR-676

@feanil Please review


course_discovery/apps/publisher/forms.py
```diff
@@ -531,6 +531,11 @@ def clean_lms_course_id(self):
 
 class AdminImportCourseForm(forms.Form):
     start_id = forms.IntegerField(min_value=1, label='This course id will import.')
+    create_canonical_course_run = forms.BooleanField(
+        label=_('Create Canonical Course Run'),
```
feanil: You already know what a `Canonical Course Run` is.  Can you add details here or link to a definition of that for someone using this form?

feanil (approved):
Seems good to me given my understanding but some more clarity around the canonical course run concept would be good.

course_discovery/apps/publisher/forms.py
schenedx: @feanil Good point. Canonical course run here don't really matter. Updated the flag to be whether we should create the "initial course run"

PR #1261, opened 2017-12-11
Changed instructor field on program to be more accurate

tuchfarber:
Changes the `instructors` field on the program model to a more accurate `instructor_ordering`. 

Not yet ready to merge. Opening PR because I'm experiencing test issues locally and want to see if they happen in travis as well.


mikedikan (commented):
This looks good - can you also leave a brief comment explaining the field - the rename is good step, but I think clarifying in more detail would help.

mikedikan:
ping @clintonb & @schenedx

course_discovery/apps/course_metadata/migrations/0074_auto_20171211_1922.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0073_program_instructors'),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
schenedx: I think doing rename field with a migration will break production system. 
The migration runs first. The old code will be running against the new db column name.
You might want to update the code to add the new `instructor_ordering` model attribute first, and mark the old attribute obsolete, then create another PR to retire the obsolete attribute.

schenedx (commented):
I am worrying about the rename migration. The generate idea here make sense to me

course_discovery/apps/course_metadata/migrations/0074_auto_20171211_1922.py
tuchfarber: Will this be an issue if no code is using this field yet? I have yet to merge the PR that references this.

schenedx: I guess, in this case, it should be OK then. Thanks for that information

tuchfarber: No problem, thanks for the confirmation.

course_discovery/apps/api/serializers.py
```diff
@@ -891,7 +891,7 @@ class ProgramSerializer(MinimalProgramSerializer):
     )
     subjects = SubjectSerializer(many=True)
     staff = PersonSerializer(many=True)
-    instructors = PersonSerializer(many=True)
+    instructor_ordering = PersonSerializer(many=True)
```
clintonb, line 894: Does this need to be a complete person, or can it simply be the UUIDs?

course_discovery/apps/course_metadata/models.py
```diff
@@ -959,7 +959,8 @@ class Program(TimeStampedModel):
     video = models.ForeignKey(Video, default=None, null=True, blank=True)
     expected_learning_items = SortedManyToManyField(ExpectedLearningItem, blank=True)
     faq = SortedManyToManyField(FAQ, blank=True)
-    instructors = SortedManyToManyField(
+    # Optional listing of instructor names used to sort instructors in LMS program about page.
+    instructor_ordering = SortedManyToManyField(
         Person,
         blank=True,
         help_text='Used to organize the instructors on the program about page.'
```
clintonb: Ditch the comment above, and update the help text to better explain what the field is used for. Also, mark it for translation.

```python
help_text=_('This field can be used by API clients to determine the order in which instructors will be displayed on program pages. Instructors in this list should appear before all others associated with this programs courses runs.')
```

clintonb (changes requested):

clintonb:
Remember to squash your commits, and update your commit message to follow our guidelines at https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/161880915/Git+Best+Practices#GitBestPractices-CommitMessages.

course_discovery/apps/api/serializers.py
tuchfarber: This needs to be a person because we are using the name fields to sort users. The instructors are coming from studio, not discovery.

clintonb: I am referring to the data that is serialized. If you only need select fields, only serialize those fields. Adding data in the future is easy. Removing unnecessary data later requires a new version of the API to be released.

Also, please investigate a method of using the UUID to match. Matching on the name is not ideal.

tuchfarber: I assumed that reusing an existing serializer would be preferred to polluting with a name-only serializer. I can see both sides of the argument though.

I agrees that UUIDs would be preferable, but instructors are currently being pulled out of studio/modulestore where they are unique for each course. If we were pulling from discovery, we could use UUIDs. A quick glance shows none of our instructors are in discovery though, only studio/modulestore.

tuchfarber: If that's something we want to fix, I think we can save that for a future iteration. I would prefer to use discovery whenever possible, but for now, I this this would be acceptable.

tuchfarber:
Also, I'm not sure how to fix the translations error? I haven't worked with this much before.

course_discovery/apps/course_metadata/models.py
```diff
@@ -959,10 +959,12 @@ class Program(TimeStampedModel):
     video = models.ForeignKey(Video, default=None, null=True, blank=True)
     expected_learning_items = SortedManyToManyField(ExpectedLearningItem, blank=True)
     faq = SortedManyToManyField(FAQ, blank=True)
-    instructors = SortedManyToManyField(
+    instructor_ordering = SortedManyToManyField(
         Person,
         blank=True,
-        help_text='Used to organize the instructors on the program about page.'
+        help_text=_('This field can be used by API clients to determine the order in which instructors will be '
```
mikedikan, line 965: This is great - thanks!

mikedikan (approved):
Thanks for the updates @tuchfarber

clintonb (approved):
Generate, and commit, the source translation files to resolve the build error.

PR #1262, opened 2017-12-12
Fixed instructor not saving on edit run

asadazam93:
### Description
Instructor was not being added if someone other then the course team tried adding it because in that case the organizations_ids was not passed at all in the context, which resulted in a json error on the front end javascript that was expecting organizations_ids to exist.

**Sandbox**
N/A

### Reviewers
- [ ] @noraiz-anwar    
- [x] @awaisdar001

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -711,9 +711,11 @@ class CourseRunEditView(mixins.LoginRequiredMixin, mixins.PublisherPermissionMix
         course_run = context.get('course_run')
         course = course_run.course
         course_user_role = course.get_user_role(user=self.request.user)
+        context['organizations_ids'] = []
         if course_user_role == PublisherUserRole.CourseTeam:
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -711,6 +711,7 @@ def get(self, request, *args, **kwargs):
         course_run = context.get('course_run')
         course = course_run.course
         course_user_role = course.get_user_role(user=self.request.user)
+        context['organizations_ids'] = []
         if course_user_role == PublisherUserRole.CourseTeam:
             context['organizations_ids'] = list(mixins.get_user_organizations(self.request.user).
```
awaisdar001: would be more readable if 
```
list(
    mixins.get_user_organizations(self.request.user).values_list('id', flat=True)
)
```

awaisdar001 (changes requested):
Any tests?

awaisdar001 (approved):

PR #1263, opened 2017-12-12
Courses having 'null' in course-end date shows incorrect course card

uzairr:
### Description
Course-discovery is returning incorrect availability status of the courses that have 'null' value in their course_end date.To avoid it,a conditional statement is modified so that it will behave as expected.


course_discovery/apps/course_metadata/models.py
```diff
@@ -724,7 +724,7 @@ def availability(self):
 
         if self.end and self.end <= now:
             return _('Archived')
-        elif self.start and self.end and (self.start <= now < self.end):
+        elif self.start and (self.start <= now):
```
ahsan-ul-haq, line 727: What status will appear for courses with null end date after this change?

ahsan-ul-haq (changes requested):

tasawernawaz (approved):
@ahsan-ul-haq I think the status will be `current`. @uzairr can confirm.

course_discovery/apps/course_metadata/models.py
uzairr: @ahsan-ul-haq if a course is started but have 'null' in its course-end date then it should return 'Current' status.

ahsan-ul-haq (approved):

PR #1264, opened 2017-12-12
Allow mulitple image resolutions to be uploaded into Publisher

schenedx:
Add a new image validator that can allow multiple image resolutions.
EDUCATOR-1954
@tasawernawaz @asadazam93 Please review


course_discovery/apps/publisher/validators.py
```diff
@@ -18,6 +19,37 @@ def compare(self, img_size, limit_size):  # pylint: disable=arguments-differ
     )
 
 
+class ImageMultiSizeValidator(ImageSizeValidator):
```
tasawernawaz, line 23: Do we still need this `ImageSizeValidator` ? If not, can we inherit this validator directly from `BaseSizeValidator`

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -111,22 +111,22 @@ def test_create_course_with_errors(self):
...
-        Verify that user can create course successfully.
-        """
-        data = {'title': 'Test2', 'number': 'testX453', 'image': make_image_file('test_banner.jpg')}
+    @ddt.data(
+        make_image_file('test_cover00.jpg', width=2120, height=1192),
+        make_image_file('test_cover01.jpg', width=1134, height=675),
+        make_image_file('test_cover02.jpg', width=378, height=225),
+    )
+    def test_create_course_valid_image(self, image):
+        data = {'title': 'Test valid', 'number': 'testX453', 'image': image}
```
tasawernawaz: sugestion: docsting will be helpful here.

course_discovery/apps/publisher/templates/publisher/course_edit_form.html
```diff
@@ -340,7 +340,7 @@ <h1 class="hd-1 emphasized">{% trans "Edit Course" %}</h1>
                                         <br/>
                                         <strong>{% trans "Course Image Guidelines:" %}</strong>
                                         <ul>
-                                            <li>{% trans "The image size must be 2120 x 1192 pixels." %}</li>
+                                            <li>{% trans "The image size must be of size 1134 X 675 pixels." %}</li>
```
tasawernawaz, line 343: should not this message look similar to what we have while creating course ? e.g  containing available sizes `(2120 X 1192), (1134 X 675), (378 X 225)`

tasawernawaz (changes requested):

course_discovery/apps/publisher/templates/publisher/course_edit_form.html
schenedx: @tasawernawaz Fair assumption. I had a discussion with @sstack22 about it and she mentioned we should message to user to hightlight the preferred resolution. But we should accept all three resolutions, therefore the difference between the instruction and the error message.

course_discovery/apps/publisher/tests/test_views.py
schenedx: Sure, I can add those

course_discovery/apps/publisher/validators.py
schenedx: It's not used. I guess clean it up is a good idea. I'll do that

course_discovery/apps/publisher/templates/publisher/course_edit_form.html
tasawernawaz: @schenedx If this is as per requirement then may be it is OK, but to me this may cause confusion while user will see error message.

course_discovery/apps/publisher/validators.py
schenedx: It turns out, I cannot clean up the `ImageSizeValidator`. This old class was referenced in model migrations. And if I remove, the unit test won't work because the migration file references to this old class would no longer be valid.
I can add a comment in code to explain this

tasawernawaz (approved):

schenedx:
@srpearce Please note the instructor change and error message change here. Could you please help me update the documentation associated with this change? We want course teams to upload 1134X675 instead of the huge 2120X1192 course image now

srpearce:
@schenedx I can update the docs. I'd also like to revise the UI text; for example, "The image **size** must be of **size** 1134 X 675 pixels" contains a redundancy, we don't use the word "please" in UI text because of translation issues, and there are stray parentheses in the error message. What is the best way to make these changes? I can update the HTML files and create a PR, but I don't know how to update tests.

For future reference, **any** user-facing UI strings should go through a doc review before they're merged.

course_discovery/apps/course_metadata/migrations/0074_auto_20171212_2016.py
```diff
@@ -0,0 +1,34 @@
...
+
+    dependencies = [
+        ('course_metadata', '0073_program_instructors'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='course',
+            name='image',
+            field=stdimage.models.StdImageField(blank=True, help_text='Please provide a course preview image', null=True, upload_to=course_discovery.apps.course_metadata.utils.UploadToFieldNamePath('uuid', path='media/course/image')),
```
srpearce, line 22: Is this the same as a course image, or is it something different?

schenedx:
@srpearce Got it. Thanks for the feedback. Can you add your text updates to the ticket EDUCATOR-1954 and I will create a second PR for the wording changes?

course_discovery/apps/course_metadata/migrations/0074_auto_20171212_2016.py
schenedx: This is the same as a course image. It's the image shows on both the dashboard and the course about page on marketing site

srpearce:
@schenedx Will do. Thanks!

schenedx:
The copy update PR is at https://github.com/edx/course-discovery/pull/1271

PR #1271, opened 2017-12-19
Update the copy for the Publisher course cover image upload

schenedx:
Both the instruction copy and the error feedback copy are changed to better help user understand the image restrictions
EDUCATOR-1954
@srpearce Please review the copy changes. See below for the Publisher look for the updated copy.
@tasawernawaz Please help review

image-copy-update


tasawernawaz (approved):
I see mostly there are text changes. @srpearce please take a look at it.

course_discovery/apps/course_metadata/migrations/0076_auto_20171219_1841.py
```diff
@@ -0,0 +1,29 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='course',
+            name='image',
+            field=stdimage.models.StdImageField(blank=True, help_text='Add the course image', null=True, upload_to=course_discovery.apps.course_metadata.utils.UploadToFieldNamePath('uuid', path='media/course/image')),
+        ),
+        migrations.AlterField(
+            model_name='program',
+            name='instructor_ordering',
+            field=sortedm2m.fields.SortedManyToManyField(blank=True, help_text='This field can be used by API clients to determine the order in which instructors will be displayed on program pages. Instructors in this list should appear before all others associated with this programs courses runs.', to='course_metadata.Person'),
```
srpearce, line 27: API clients can use this field to determine the instructor order on program pages. Instructors in this list appear before all others associated with courses and course runs in the program.

course_discovery/apps/publisher/templates/publisher/course_edit_form.html
```diff
@@ -340,7 +340,12 @@ <h1 class="hd-1 emphasized">{% trans "Edit Course" %}</h1>
                                         <br/>
                                         <strong>{% trans "Course Image Guidelines:" %}</strong>
                                         <ul>
-                                            <li>{% trans "The image size must be of size 1134 X 675 pixels." %}</li>
+                                            <li>
+                                                {% trans "New course images must be 1134 X 675 pixels in size. " %}
+                                                <a href='http://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/set_up_course/planning_course_information/image_guidelines.html#course-image-size'>
+                                                    {% trans "Learn More." %}
```
srpearce: Decap "more"

course_discovery/apps/course_metadata/migrations/0076_auto_20171219_1841.py
schenedx: @srpearce Unfortunately, I want to exclude text change on this line for the following reasons.
1. This is a change that should not be part of this PR. This is a change introduced by https://github.com/edx/course-discovery/pull/1261.
2. This is a text in a migration file. It represent a point in time in the past. The text here do not necessarily gets used by the system and are visible to the user.

course_discovery/apps/publisher/templates/publisher/course_edit_form.html
schenedx: Updated.

PR #1276, opened 2018-01-05
Making publisher code more defensive.

awaisdar001:
EDUCATOR-2061


awaisdar001:
implementing admin level validation message to prevent this.

PR #1282, opened 2018-01-10
Remove hack to force a rebuild of node-sass.

ormsbee:
The line was originally put in to address failures in the Node 8
upgrade where the Node 6 version was getting cached, but caused other
problems in Travis where the process doesn't have the permissions to
create those directories. Hopefully no longer needed.


ormsbee:
@arizzitano: Build passes.

ormsbee:
FYI @mjfrey

arizzitano (approved):
woohoo! :ship: \o/

PR #1284, opened 2018-01-12
handle stock record errors in ecommerce dataloader

McKenzieW:
Handle exceptions that occur while processing stock record information for course entitlement products being loaded into discovery from ecommerce by logging the error, then continue to load proceeding entitlement products.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -383,19 +383,39 @@ class EcommerceApiDataLoader(AbstractDataLoader):
                                           defaults=defaults)
 
     def update_entitlement(self, body):
+        """
+        Argument:
+            body (dict): entitlement product data from ecommerce
+        Returns:
+            entitlement product sku if no exceptions, else None
+        """
         attributes = {attribute['name']: attribute['value'] for attribute in body['attribute_values']}
         course_uuid = attributes.get('UUID')
+        title = body['title']
 
-        stock_record = body['stockrecords'][0]
-        currency_code = stock_record['price_currency']
-        price = Decimal(stock_record['price_excl_tax'])
-        sku = stock_record['partner_sku']
+        if body['stockrecords']:
+            stock_record = body['stockrecords'][0]
+        else:
+            msg = 'Entitlement product {entitlement} has no stockrecords'.format(entitlement=title)
+            logger.warning(msg)
+            return None
+
+        try:
+            currency_code = stock_record['price_currency']
+            price = Decimal(stock_record['price_excl_tax'])
+            sku = stock_record['partner_sku']
+        except (KeyError, ValueError):
+            msg = 'A necessary stockrecord field is missing or incorrectly set for entitlement {entitlement}'.format(
+                entitlement=title
+            )
+            logger.warning(msg)
+            return None
 
         try:
             course = Course.objects.get(uuid=course_uuid)
         except Course.DoesNotExist:
             msg = 'Could not find course {uuid} while loading entitlement {entitlement} with sku {sku}'.format(
-                uuid=course_uuid, entitlement=body['title'], sku=sku
+                uuid=course_uuid, entitlement=title, sku=sku
             )
             logger.warning(msg)
             return None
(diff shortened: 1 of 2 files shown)


McKenzieW:
@edx/helio please review

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -386,10 +386,17 @@ def update_entitlement(self, body):
...
-        stock_record = body['stockrecords'][0]
-        currency_code = stock_record['price_currency']
-        price = Decimal(stock_record['price_excl_tax'])
-        sku = stock_record['partner_sku']
+        try:
+            stock_record = body['stockrecords'][0]
+            currency_code = stock_record['price_currency']
+            price = Decimal(stock_record['price_excl_tax'])
+            sku = stock_record['partner_sku']
+        except KeyError:
```
Mjloturco: Does except KeyError automatically include that keyerror in the log somehow? if no, it be useful to include the error itself in the log msg.

Mjloturco (commented):
1 Question, but good call on adding logging 👍

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -385,17 +385,31 @@ def update_seat(self, course_run, product_body):
...
         attributes = {attribute['name']: attribute['value'] for attribute in body['attribute_values']}
         course_uuid = attributes.get('UUID')
+        title = body['title']
 
-        stock_record = body['stockrecords'][0]
-        currency_code = stock_record['price_currency']
-        price = Decimal(stock_record['price_excl_tax'])
-        sku = stock_record['partner_sku']
+        try:
+            stock_record = body['stockrecords'][0]
```
staubina: It would be more appropriate to protect against the exception here rather than relying on it. Why not just check the length of the stockrecords list and log if it is not >0?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -385,17 +385,31 @@ def update_seat(self, course_run, product_body):
...
+        try:
+            currency_code = stock_record['price_currency']
+            price = Decimal(stock_record['price_excl_tax'])
+            sku = stock_record['partner_sku']
+        except (KeyError, ValueError):
+            msg = 'A necessary stockrecord field is missing or incorrectly set for entitlement {entitlement}'.format(
+                entitlement=title
+            )
+            logger.warning(msg)
+            return None
```
staubina, line 412: Update the method Docstring and add the returns here noting what scenarios you would return None in.  For example: returns <something I am still looking for, sku?> and None when an exception occurs.

staubina (changes requested):

course_discovery/apps/course_metadata/data_loaders/api.py
McKenzieW: We could do that, but in the event that there are no stock records for the product we'd still want to log the error and skip loading the entitlement into discovery because it is an invalid representation of the product.

McKenzieW: I think an IndexError properly captures the invalid nature of an empty stockrecords list here.

staubina: The IndexError captures the fact that we are attempting to access an empty array not the true issue here which is that the array is empty.  I would prefer that we write the code to protect against exceptions like this one not use exceptions to catch errors we can avoid.

I absolutely agree that we should log the error message.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -384,7 +384,7 @@ def update_seat(self, course_run, product_body):
 
     def update_entitlement(self, body):
         """
-        Returns entitlement sku, unless an error occurs while updating or creating entitlement object.
+        Returns entitlement sku or None if an error occurs while updating or creating entitlement object.
```
staubina: Please use the format in the code style doc for the formatting of Docstrings

http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings

staubina: It is used to build code docs by automated tools.

staubina (approved):
:shipit:

PR #1286, opened 2018-01-23
LEARNER-3879: Adds in publisher entitlement model and version to course

jlajoie:


course_discovery/apps/publisher/models.py
```diff
@@ -485,6 +486,33 @@ def calculated_upgrade_deadline(self):
...
+        (VERIFIED, _('Verified')),
+        (PROFESSIONAL, _('Professional'))
+    )
+
+    PRICE_FIELD_CONFIG = {
+        'decimal_places': 2,
+        'max_digits': 10,
+        'null': False,
+        'default': 0.00,
+    }
```
amangano-edx, line 509: I'm not sure how we do it elsewhere, but aren't there issues with storing price as a decimal (due to the somewhat unreliable behavior of floats)?

course_discovery/apps/publisher/models.py
```diff
@@ -485,6 +486,33 @@ def calculated_upgrade_deadline(self):
...
+
+    course = models.ForeignKey(Course, related_name='entitlements')
+    mode = models.ForeignKey(SeatType, related_name='publisher_entitlements')
```
amangano-edx: Do we need a foreign key relationship here? The seat model uses a CharField for type.

course_discovery/apps/publisher/models.py
```diff
@@ -485,6 +486,33 @@ def calculated_upgrade_deadline(self):
...
 
 
+class CourseEntitlement(TimeStampedModel):
+    VERIFIED = 'verified'
+    PROFESSIONAL = 'professional'
+
+    SEAT_TYPE_CHOICES = (
+        (VERIFIED, _('Verified')),
+        (PROFESSIONAL, _('Professional'))
+    )
```
amangano-edx: This isn't currently used (on the Seat model, it's used for the choices argument on the 'type' field).

Also, if we need this, could we call it something like `COURSE_MODE_CHOICES` to distinguish it from `Seat`s?

course_discovery/apps/publisher/constants.py
```diff
@@ -14,3 +14,7 @@
 
 # Being used in old migration `0019_create_user_groups`.
 PARTNER_COORDINATOR_GROUP_NAME = 'Partner Coordinators'
+
+# Versions for code paths in publisher Course and Course Run Create/Edit
+COURSE_SEAT_VERSION = 0
+COURSE_ENTITLEMENT_VERSION = 1
```
amangano-edx: I think it might be better for these to live directly on the `Course` model, since they're specific to the model, not the application.

course_discovery/apps/publisher/models.py
jlajoie: This was taken directly from the Seat model, so any problems here would also occur there.

course_discovery/apps/publisher/models.py
jlajoie: We don't, but it seemed better to not use modes as strings if that was possible. Happy to change it though since it'd match the Publisher__Seat model. Opinions?

course_discovery/apps/publisher/migrations/0065_set_course_verison.py
```diff
@@ -0,0 +1,26 @@
...
+
+from course_discovery.apps.publisher.constants import COURSE_SEAT_VERSION
+
+
+def set_version(apps, schema_editor):
+    Course = apps.get_model('publisher', 'Course')
+
+    for course in Course.objects.filter(Q(version__isnull=True)):
+        course.version = COURSE_SEAT_VERSION
+        course.save()
```
amangano-edx: Do we need to fill in the historical table (https://github.com/edx/course-discovery/pull/1286/files#diff-a604c93caaa88baf79e663d0bdb94ed9R20) as well?

course_discovery/apps/publisher/migrations/0065_set_course_verison.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.13 on 2017-05-05 02:32
```
amangano-edx: Curious how this was generated using Django 1.9, but other migrations in this PR (https://github.com/edx/course-discovery/pull/1286/files#diff-a604c93caaa88baf79e663d0bdb94ed9R2) were generated using Django 1.11.

course_discovery/apps/publisher/models.py
amangano-edx: Just my opinion, but I'd prefer to use strings because:
1.) It's more consistent with how the other models work
2.) We avoid creating direct relationships between publisher models and course metadata models
3.) It's unclear what the SeatType model is meant to represent. Is it a CourseMode, or something meant to be more coupled to Seats?

course_discovery/apps/publisher/models.py
mikix: Oh yeah even my branch doesn't use this -- the type choices are just in the CourseForm (not even the model) and then we branch off that (at least in my branch).

I'm fine with this being dropped.

course_discovery/apps/publisher/models.py
mikix: Though the course_metadata.models.Seat class has this note about their type field: "TODO Replace with FK to SeatType model"

jlajoie: Updating this to just use the Char. I think I prefer the consistency inside of the Publisher tables rather so that they can be updated in parallel later. I'm going to just update my base migration so you guys that are based off of this it'll break your migrations down the road.

You'll have to drop your tables manually and then remove the entries from the django_migrations table. Reason for this is to keep it at only single migration to add the field, single migration to populate the base version, and single migration for the entitlements table, rather than having multiple per table in a single PR.

course_discovery/apps/publisher/models.py
jlajoie: Using this now that I'm going to use the CharField.

course_discovery/apps/publisher/migrations/0065_set_course_verison.py
jlajoie: Because copy + paste failure 😄 , just wrote this from hand based off of another, I'll update this properly.

amangano-edx: Ha, if I had read past the Django 1.19.13 bit I would have seen that it was also 'generated' last May.

course_discovery/apps/publisher/admin.py
```diff
@@ -116,6 +116,12 @@ class SeatAdmin(SimpleHistoryAdmin):
     search_fields = ['course_run__course__title', 'type']
 
 
+@admin.register(CourseEntitlement)
+class CourseEntitlement(SimpleHistoryAdmin):
+    list_display = ['mode']
```
mikix: You probably want course_id here too

course_discovery/apps/publisher/models.py
```diff
@@ -44,6 +44,10 @@ class Meta:
 class Course(TimeStampedModel, ChangedByMixin):
     """ Publisher Course model. It contains fields related to the course intake form."""
 
+    # Versions for code paths in publisher Course and Course Run Create/Edit
+    COURSE_SEAT_VERSION = 0
+    COURSE_ENTITLEMENT_VERSION = 1
```
mikix: Tiny ignorable nit: COURSE_ seems redundant.

course_discovery/apps/publisher/models.py
```diff
@@ -485,6 +490,33 @@ def calculated_upgrade_deadline(self):
...
+
+    PRICE_FIELD_CONFIG = {
+        'decimal_places': 2,
+        'max_digits': 10,
+        'null': False,
+        'default': 0.00,
+    }
+
+    course = models.ForeignKey(Course, related_name='entitlements')
+    mode = models.CharField(max_length=63, choices=COURSE_MODE_CHOICES, verbose_name='Course mode')
```
mikix, line 512: I'm a fan of mode instead of type, even though it does differ from Seat. Makes it easier to use variables called 'mode' rather than the reserved 'type' around in the code. :+1: 

Is it worth considering ChoiceField here?  I assume that adds a DB constraint?  Or at least a constraint in Django... I know Seat uses a CharField and we want to be compatible with that.  But maybe ChoiceField is still compatible (just strings) but with bonus features?

I dunno, I don't care enough or know enough to say it should change.  Just a thought.

mikix (changes requested):
Nice!  I left some comments, but the only obvious fix is list_display in admin.py.

course_discovery/apps/publisher/models.py
amangano-edx: Nit: It might be helpful to add a couple comments above these summarizing the different states.

course_discovery/apps/publisher/models.py
```diff
@@ -84,6 +88,7 @@ class Course(TimeStampedModel, ChangedByMixin):
     keywords = TaggableManager(blank=True, verbose_name='keywords')
     faq = models.TextField(default=None, null=True, blank=True, verbose_name=_('FAQ'))
     video_link = models.URLField(default=None, null=True, blank=True, verbose_name=_('Video Link'))
+    version = models.IntegerField(null=True, blank=True, verbose_name=_('Workflow Version'))
```
amangano-edx: Do we need to update any logic anywhere to make sure that this is always set on create?

amangano-edx: Otherwise, we risk allowing null values to be inserted during the time between when your migration runs and the rest of the publisher work is done.

course_discovery/apps/publisher/migrations/0065_set_course_verison.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.db.models import Q
+
+from course_discovery.apps.publisher.models import Course
+
+
+def set_version(apps, schema_editor):
+    for course in Course.objects.filter(Q(version__isnull=True)):
```
amangano-edx: Nit: is the Q object necessary here, or could you get by with just
`Course.objects.filter(version__isnull=True)`

course_discovery/apps/publisher/admin.py
jlajoie: Actually didn't need it, was able to add a way to search for courses, the mode as a dropdown I think is fine.

course_discovery/apps/publisher/migrations/0065_set_course_verison.py
jlajoie: Was able to do away with this file entirely by adding the default to the migration 👍

jlajoie: Also 12113 rows in the publisher_historicalcourse table, and 3151 in the publisher_course table, so this shouldn't have any real impact.

course_discovery/apps/publisher/admin.py
jlajoie: screen shot 2018-01-25 at 1 40 16 pm
What these options look like (won't load all courses in a dropdown anymore)

mikix (approved):
Looks good to me

PR #1287, opened 2018-01-24
Support creating entitlements for new courses

mikix:
Make course creation entitlement-aware by creating an entitlement along with the seats when a verified or professional mode course is made.

https://openedx.atlassian.net/browse/LEARNER-3767


course_discovery/apps/publisher/forms.py
```diff
@@ -442,6 +442,31 @@ def reset_credit_to_default(self, seat):
...
+    class Meta:
+        fields = ('mode', 'price')
+        model = CourseEntitlement
+
+    def clean(self):
+        cleaned_data = super().clean()
+        mode = cleaned_data.get('mode')
+        price = cleaned_data.get('price')
+
+        if mode and (price is None or price <= 0):
```
jlajoie, line 464: Just to double check, but an empty string registers as False in python right?

mikix: Yes. Got an improvement for this conditional?

jlajoie: Nope makes sense for me, just checking that if a mode is set and the price either isn't or is not a positive value we raise the error. Just wanted to make sure my understanding of that evaluation was right

course_discovery/static/js/publisher/seat-type-change.js
```diff
@@ -1,4 +1,4 @@
-$(document).on('change', '#id_type', function (e) {
+$(document).on('change', '#id_type, #id_mode', function (e) {
     var $seatBlock = $("#SeatPriceBlock"),
```
jlajoie, line 2: styling nit: I'd rather see the casing between $seatBlock and $creditPrice match one another

course_discovery/apps/publisher/templates/publisher/add_course_form.html
```diff
@@ -128,6 +128,25 @@ <h1 class="hd-1 emphasized">
...
+                    <div class="col col-6 help-text">
+                        {% trans "If the course offers a verified or professional education certificate, select the certificate type and enter the price for the certificate." %}
+                    </div>
+                    <div class="col col-6">
+                        <div class="row">
+                            <div class="col col-6">
+                                <label class="field-label ">{{ entitlement_form.mode.label_tag }}</label>
+                                {{ entitlement_form.mode }}
+                            </div>
+                            <div id="SeatPriceBlock" class="col col-6 {% if not entitlement_form.mode.value %}hidden{% endif %}">
```
jlajoie, line 143: styling nit: also commented below, as well I'd rather see the casing between SeatPriceBlock and other ids which look to be lowerCamelCase

course_discovery/static/js/publisher/seat-type-change.js
mikix: I didn't write those bits (SeatPriceBlock and this code existed already for run creation), but I'm *guessing* it's because someone felt like CSS ids should be capitalized (what's the typical styling there? I think I usually see all lowercases).

jlajoie: I'm accustomed to all lower cases and hyphenated, just a minor nit so doesn't really matter, was looking at the `add_course_form.html` you added the `SeatPriceBlock` so I assumed that this was also an addition, don't worry about changing this since it isn't new.

jlajoie (approved):
lgtm 👍

course_discovery/apps/publisher/templates/publisher/add_course_form.html
mikix: Next PR I'll be more aggressive about fixing casing, but I don't wanna bother changing this branch now for it.

PR #1288, opened 2018-01-24
Fix make test command

tuchfarber:
DJANGO_SETTINGS_MODULE in pytest.ini was being overriden by the environment variable version of it. This overrides the setting in the make command.


mikix (approved):
Nice. This fixes the errors on master for me.  Looks right. Thanks!

PR #1289, opened 2018-01-26
Ordered subjects in course metadata courses

asadazam93:
### Description
Ordered the subjects field of the course meta data course model (Ordering is based on which subject was inserted first, First in first out method). Publisher implements subject hierarchy by using three different fields for subjects namely primary_subject, secondary_subject and tertiary_subject. The same hierarchy should be respected in the course metadata course, but the course metadata course model has only field for subjects. So by making that field SortedManyToManyField we can make sure that the hierarchy in publisher is forwarded to course metadata.

**Sandbox**
N/A

### Reviewers
- [ ] @noraiz-anwar     
- [ ] @awaisdar001

### Post-review


schenedx (commented):
You would also need to make changes at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/api/v1/views.py#L115 to make sure subject hierarchy is transferred through.

asadazam93:
@schenedx The subjects in the set follow the hierarchy. Would it not be transferred through?

schenedx:
@asadazam93 Have you tested the "publish" button to ensure the subject order it transferred through? From what I read on https://stackoverflow.com/questions/9792664/set-changes-element-order, the `set` type do not guarantee order. The code change should be pretty simple to guarantee the ordering isn't it? Plus, I would suggest update the unit test to make sure ordering is guaranteed in the publisher code.

asadazam93:
@schenedx you are right set did not ensure ordering. I have used OrderedDict to make sure the order is transferred through.

awaisdar001:
The changes look good to me. You have failing tests tough!

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -112,11 +113,12 @@ def publish_to_discovery(self, partner, course_run):
         discovery_course.image.save(publisher_course.image.name, publisher_course.image.file)
         discovery_course.authoring_organizations.add(*publisher_course.organizations.all())
 
-        subjects = [subject for subject in set([
+        subjects = [subject for subject in [
```
schenedx, line 116: Why not just construct a ordered list directly? Rather than create an array first, then create the ordered list from the array?

schenedx (commented):
I think if you can take in one more comment from me, this would look good.

course_discovery/apps/publisher/api/v1/views.py
asadazam93: I am creating a list of non empty subjects first and then creating an ordered dict from that list.

schenedx: Why not do 
```
subjects=[subject for subject in [
    publisher_course.primary_subject,
    publisher_course.secondary_subject,
    publisher_course.tertiary_subject
] if subject]
```
According to what I understand, if you remove the `set`, the order within the list is always persistent. Why use an ordered dict at all?

asadazam93: We were using set to remove repeated subjects. Now that we are not using set anymore we need some other way to remove the repeated subjects and keep the order. OrderedDict is doing both these things for us.

schenedx: @asadazam93 Got it. Interesting. Can you please add a comment about it then?

schenedx (approved):
👍

PR #1292, opened 2018-01-29
Show mode and price for new courses

mikix:
I could have done a conditional check for the course version field, but I figured dumber was better here.  If they have an entitlement, they have a price & mode.

https://openedx.atlassian.net/browse/LEARNER-3769


jlajoie (approved):
:+1:

course_discovery/apps/publisher/templates/publisher/course_detail.html
```diff
@@ -71,6 +71,30 @@
                 <div class="show-diff"></div>
             </div>
 
+            <div class="info-item history-field-container">
+                <div class="heading">
+                    {% trans "Enrollment Track" %}
```
amangano-edx, line 76: Do these headings appear even if they don't have an entitlement?

course_discovery/apps/publisher/templates/publisher/course_detail.html
```diff
@@ -71,6 +71,30 @@
...
+                    {% with object.entitlements.first.mode as field %}
+                        {% include "publisher/_render_optional_field.html" %}
+                    {% endwith %}
+                </div>
+                <div class="show-diff"></div>
+            </div>
+
+            <div class="info-item history-field-container">
+                <div class="heading">
+                    {% trans "Certificate Price" %}
```
amangano-edx, line 88: Does this appear even without a price?

course_discovery/apps/publisher/templates/publisher/course_detail.html
amangano-edx: Also, translations should be escaped.

mikix: Yes, the headings appear without an entitlement. Per the ACs, we are supposed to treat them like optional fields that aren't filled in yet.

Django templates should be autoescaped already.

course_discovery/apps/publisher/templates/publisher/course_detail.html
mikix: Yes, the headings appear without an entitlement. Per the ACs, we are supposed to treat them like optional fields that aren't filled in yet.

course_discovery/apps/publisher/templates/publisher/course_detail.html
amangano-edx: I just tested this, and they are not autoescaped.

mikix: Right you are! As we discussed afterward in person, we'll file a ticket to address this module-wide.

PR #1295, opened 2018-01-30
Handle & show appropriate error msg when course run create fails.

awaisdar001:
When course team user exists on discovery but not in the studio, a generic error is displayed where no information what went wrong. This PR adds the first step to show some helpful error message which is easier to understand at first sight. In the next step, a more detailed error message should be displayed which might require a large amount of work.

screen shot 2018-01-30 at 2 08 35 pm


awaisdar001:
FYI: @asadazam93 / @noraiz-anwar

course_discovery/static/js/publisher/instructors.js
```diff
@@ -119,6 +119,7 @@ function getFormInstructorPosition () {
 }
 
 function loadSelectedImage (input) {
+    console.log('aj')
```
noraiz-anwar: ?

course_discovery/apps/publisher/signals.py
```diff
@@ -41,27 +44,38 @@ def create_course_run_in_studio_receiver(sender, instance, created, **kwargs):
...
-        try:
+        discovery_course_run = get_related_discovery_course_run(instance)
+
+        if discovery_course_run:
             try:
-                discovery_course_run = get_related_discovery_course_run(instance)
                 logger.info('Creating a re-run of [%s]...', discovery_course_run.key)
                 response = api.create_course_rerun_in_studio(instance, discovery_course_run)
-            except ObjectDoesNotExist:
+            except SlumberBaseException as ex:
```
noraiz-anwar, line 62: What about deciding which function to call from `create_course_rerun_in_studio` and `create_course_run_in_studio` first and then call it under single `except SlumberBaseException` . Or any other way we can remove repetition of  SlumberBaseException?

course_discovery/apps/publisher/tests/test_signals.py
```diff
@@ -213,6 +213,43 @@ def test_create_course_run_in_studio_with_api_failure(self, mock_access_token):
                     json.dumps(body).encode('utf8')
                 )
 
+    @responses.activate
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    @override_switch('enable_publisher_create_course_run_in_studio', active=True)
+    def test_create_course_run_error_with_discovery_run(self, mock_access_token):  # pylint: disable=unused-argument
```
noraiz-anwar, line 219: Why do we need `mock_access_token `?

course_discovery/apps/publisher/signals.py
awaisdar001: I thought about that but that didn't look elegant. All code within the try/except is different.
i.e. I thought something like.
```
        if discovery_course_run:
            log_info = 'Creating a re-run of [{}]...'.format(discovery_course_run.key)
            api_func = api.create_course_rerun_in_studio
            args  = (instance, discovery_course_run)
            except_log = 'Failed to create course re-run [{}] on Studio: {}'.format(discovery_course_run.key, ex.content)
            
        else:
            ... // same vars as above
```
What do you have in mind?

course_discovery/apps/publisher/tests/test_signals.py
awaisdar001: It is required down the stream. 
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/core/models.py#L127 and
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/signals.py#L54

awaisdar001:
@noraiz-anwar @asadazam93 please have a look again.

noraiz-anwar (approved):
👍

asadazam93 (approved):

PR #1296, opened 2018-01-30
Fixed incomplete course team list on edit course

asadazam93:
### Description
The api that returns users on the course edit page returns a paginated response. The page max size was 20 which was causing incomplete user lists to be displayed. I have increased the pagination limit to a 1000 for now. This should fix the problem in the short term but as a long term solution we should remove the paginated response from the api as there is no need of pagination in this case.

**Sandbox**
N/A

### Reviewers
- [ ] @noraiz-anwar     
- [ ] @awaisdar001

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/api/paginations.py
+++ b/course_discovery/apps/publisher/api/paginations.py
@@ -1,0 +1,5 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class LargeResultsSetPagination(PageNumberPagination):
+    page_size = 100
--- a/course_discovery/apps/publisher/api/views.py
+++ b/course_discovery/apps/publisher/api/views.py
@@ -11,6 +11,7 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from course_discovery.apps.core.models import User
+from course_discovery.apps.publisher.api.paginations import LargeResultsSetPagination
 from course_discovery.apps.publisher.api.permissions import (CanViewAssociatedCourse, InternalUserPermission,
                                                              PublisherUserPermission)
 from course_discovery.apps.publisher.api.serializers import (CourseRevisionSerializer, CourseRunSerializer,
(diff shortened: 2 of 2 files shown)


course_discovery/apps/publisher/api/paginations.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class LargeResultsSetPagination(PageNumberPagination):
+    page_size = 1000
```
awaisdar001: from 20 to 1000? why so much difference? is it really possible?

asadazam93: I don't think we would ever reach this limit. This is just for short term since we need this issue fixed quickly. For the proper solution we need to remove the pagination altogether.

awaisdar001: maybe we can do something 50 or even 75? I am hesitant to go this drastic increase. Even though we will not hit the limit, but defining 1000 mean we are allowing 1k records to return in page 1 which doesn't make sense (page load problems may come into action)
What do you think?

asadazam93: That can happen. I think we can make it a 100 then, that would be a safer option.

awaisdar001: Ok.

noraiz-anwar (approved):
👍

course_discovery/apps/publisher/api/paginations.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class LargeResultsSetPagination(PageNumberPagination):
+    page_size = 100
+    page_size_query_param = 'page_size'
+    max_page_size = 1000
```
awaisdar001: just a question what is `max_page_size`?

awaisdar001 (commented):
Does this needs tests?

course_discovery/apps/publisher/api/paginations.py
awaisdar001: from the docs, it looks like it is only relevant if 'page_size_query_param' has also been set.
http://www.django-rest-framework.org/api-guide/pagination/#configuration

asadazam93: Yeah we don't need it in our case. I will remove it

asadazam93:
@awaisdar001 I don't think we need tests for this one. We should create a new ticket for the proper solution of this issue which would be to remove the pagination.

awaisdar001 (approved):

PR #1297, opened 2018-01-30
Refactor CreateCourseRunView and CreateRunFromDashboardView

amangano-edx:
Refactors the publisher CreateCourseRunView CreateRunFromDashboardView(s) to be more consistent and make it easier to add logic to handle entitlement-aware courses.

Still need to add/verify tests.


mikix (approved):
Seems fine to me.  Refactor diffs are a little hard to follow, but the changes make sense.

You grouped the form UI options a bit differently?  Merged start date and end date and left type/mode separate from end date? If so, nice. :+1:

amangano-edx:
@mikix Yes, I move the form inputs around so that the inputs for each Django form-type are together.

PR #1298, opened 2018-01-30
Revert entitlement creation when making a course

mikix:
Let's guard entitlement support behind a waffle flag before releasing it into the wild.


amangano-edx (approved):

amangano-edx:
Can you change the commit message to include the message from the original commit?

PR #1299, opened 2018-01-30
Be entitlement-aware for new courses

mikix:
When creating a new course in publisher, require type and price up front rather than waiting until creating a course run. (only for verified and professional courses)

And hide this feature behind a waffle flag.

**User Flow**
* On course creation page, there is a new field for course mode.
* It starts with a 'Verified' dropdown, with two additional options of 'Audit only or credit' and 'Professional education'.
* If the user continues without choosing a mode, they get traditional flow completely (no changes). This is expected choice for audit and credit courses.
* If the user chooses a mode, they also have to set a price. And then (with follow on PRs), they don't get to set mode or price on the course run.

**Behind the scenes**
* If the user selected a mode, we create an entitlement in our publisher table and set course version to ENTITLEMENT_VERSION of 1 (verified and professional only go through this flow).
* If they don't, we don't touch anything from previous database flow except to set the course version to SEAT_VERSION of 0 (audit and credit go through this flow).

**Changes**
I made a few edits in this PR from the previous reverted one:
* Added a waffle switch and constant for its name.
* Renamed SeatPriceBlock to seatPriceBlock per a nit in the previous PR I didn't get to.
* Changed dropdown default from "Audit  only or credit" to "Verified" because that is the 80% case.

https://openedx.atlassian.net/browse/LEARNER-3767


mikix:
I'm going to make a couple changes based on a talk with Deen -- make the mode field default to 'verified'

course_discovery/apps/publisher/forms.py
```diff
@@ -442,6 +442,32 @@ def reset_credit_to_default(self, seat):
         seat.credit_price = 0.00
 
 
+class CourseEntitlementForm(BaseForm):
+    MODE_CHOICES = [
+        ('', _('Audit-only or Credit')),
+        (CourseEntitlement.VERIFIED, _('Verified')),
+        (CourseEntitlement.PROFESSIONAL, _('Professional education')),
```
amangano-edx: Nit: Does 'education' need to be capitalized?

mikix: I don't think so. I copied the string from SeatForm, which uses that same form. But I suppose if I'm using that form, I should de-capitalize "Audit-only or Credit"...  Will fix.

course_discovery/apps/publisher/forms.py
```diff
@@ -442,6 +442,32 @@ def reset_credit_to_default(self, seat):
...
+        fields = ('mode', 'price')
+        model = CourseEntitlement
+
+    def clean(self):
+        cleaned_data = super().clean()
+        mode = cleaned_data.get('mode')
+        price = cleaned_data.get('price')
+
+        if mode and (price is None or price <= 0):
+            raise ValidationError({'price': ''})  # mimics required field behavior (red border only, no error string)
```
amangano-edx: Note: if you want to attach this error directly to the price field, you can use `add_error`, as specified here: https://docs.djangoproject.com/en/1.11/ref/forms/validation

course_discovery/apps/publisher/templates/publisher/add_course_form.html
```diff
@@ -128,6 +128,25 @@ <h1 class="hd-1 emphasized">
                         {% endif %}
                     </div>
                 </div>
+
+                <div class="field-title" {% if not publisher_entitlements %}hidden{% endif %}>{% trans "CERTIFICATE TYPE AND PRICE" %}</span></div>
```
amangano-edx: Is there any reason to include this in the DOM at all if publisher_entitlements is false? If not, could we move the `if not publisher_entitlements` call outside of this div?

course_discovery/apps/publisher/templates/publisher/add_course_form.html
```diff
@@ -128,6 +128,25 @@ <h1 class="hd-1 emphasized">
...
+                                <label class="field-label ">{{ entitlement_form.mode.label_tag }}</label>
+                                {{ entitlement_form.mode }}
+                            </div>
+                            <div id="seatPriceBlock" class="col col-6 {% if not entitlement_form.mode.value %}hidden{% endif %}">
+                                <label class="field-label ">{{ entitlement_form.price.label_tag }} <span class="required">*</span></label>
+                                {{ entitlement_form.price }}
+                            </div>
+                        </div>
+                    </div>
+                </div>
```
amangano-edx, line 150: We should pass the translated text through an escape filter.

course_discovery/apps/publisher/templates/publisher/add_course_form.html
```diff
@@ -128,6 +128,25 @@ <h1 class="hd-1 emphasized">
                         {% endif %}
                     </div>
                 </div>
+
+                <div class="field-title" {% if not publisher_entitlements %}hidden{% endif %}>{% trans "CERTIFICATE TYPE AND PRICE" %}</span></div>
+                <div {% if not publisher_entitlements %}class="hidden"{% else %}class="row"{% endif %}>
```
amangano-edx: If you move the `if not publisher_entitlements` check to line 131, as suggested in the above comment, you would be able to remove it here.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -184,7 +184,7 @@ <h1 class="hd-1 emphasized">
                                 <label class="field-label ">{{ seat_form.type.label_tag }}
                                 </label> {{ seat_form.type }}
                             </div>
-                            <div id="SeatPriceBlock" class="col col-6 {% if seat_form.type.value == 'audit' or not seat_form.price.value %}hidden{% endif %}">
+                            <div id="seatPriceBlock" class="col col-6 {% if seat_form.type.value == 'audit' or not seat_form.price.value %}hidden{% endif %}">
```
amangano-edx, line 187: If we're going to change this everywhere, and if it's only being used for JS logic, it would be better to add a class like `.js-seat-block`. This is a convention we've used in several IDAs when we've need to hook into html elements using JavaScript.

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -344,6 +345,49 @@ def test_course_title_formatting(self):
...
+        if price is not None:
+            entitlement_form.cleaned_data['price'] = price
+
+        with self.assertRaises(ValidationError):
+            entitlement_form.clean()
+
+    @ddt.data(
+        (None, None),
+        (None, 0),
+        (None, 50),
```
amangano-edx: This case should probably be invalid.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -282,6 +283,70 @@ def test_create_course_with_add_run(self):
             target_status_code=200
         )
 
+    def _make_course(self, data=None):
```
amangano-edx: Nit: This name could be more descriptive. At first glance, it may seem like a simple factory function when in reality it's posting to an endpoint and validating a response.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -282,6 +283,70 @@ def test_create_course_with_add_run(self):
...
+            target_status_code=200
+        )
+        return Course.objects.get(number=course_dict['number'])
+
+    @ddt.data(CourseEntitlement.VERIFIED, CourseEntitlement.PROFESSIONAL)
+    def test_create_entitlement(self, mode):
+        """
+        Verify that we create an entitlement for appropriate types (happy path)
+        """
+        toggle_switch(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH, True)
```
amangano-edx, line 305: Sanity check: does this affect global state? I think it should probably be ok since the switches are stored in a DB table, which is theoretically rolled back between tests. Just want to be sure though.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -282,6 +283,70 @@ def test_create_course_with_add_run(self):
...
+
+        entitlement = course.entitlements.first()
+        self.assertEqual(course, entitlement.course)
+        self.assertEqual(mode, entitlement.mode)
+        self.assertEqual(50, entitlement.price)
+
+    def test_entitlement_switch(self):
+        """
+        Verify that when we create a verified seat without the entitlement waffle switch enabled, we set version right
+        """
```
amangano-edx: It might be good to explicitly toggle the switch off, in case the default ever changes.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -282,6 +283,70 @@ def test_create_course_with_add_run(self):
...
+        toggle_switch(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH, True)
+        course = self._make_course()
+        self.assertEqual(0, CourseEntitlement.objects.all().count())
+        self.assertEqual(Course.SEAT_VERSION, course.version)
+
+    @ddt.data(0, -1, None)
+    def test_invalid_entitlement_price(self, price):
+        """
+        Verify that when we check price validity when making an entitlement
```
amangano-edx: I think the word "when" can be removed after "that"

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -282,6 +283,70 @@ def test_create_course_with_add_run(self):
...
+        Verify that when we create a verified seat without the entitlement waffle switch enabled, we set version right
+        """
+        data = {'mode': CourseEntitlement.VERIFIED, 'price': 50}
+        course = self._make_course(data)
+        self.assertEqual(0, CourseEntitlement.objects.all().count())
+        self.assertEqual(Course.SEAT_VERSION, course.version)
+
+    def test_seat_version(self):
+        """
+        Verify that when we create a normal seat, we set version
```
amangano-edx: It might be good to specify what a 'normal seat' is, as it's not clear from reading the test.

course_discovery/apps/publisher/views.py
```diff
@@ -249,6 +251,8 @@ def get_success_url(self, course_id, add_new_run=None):  # pylint: disable=argum
     def get_context_data(self):
         return {
             'course_form': self.course_form(user=self.request.user),
+            'entitlement_form': self.entitlement_form(),
+            'publisher_entitlements': waffle.switch_is_active(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH),
```
amangano-edx: Alternatively, you could change the template to only render the entitlement_form if present (`{% if entitlement_form %}`), and only add it to the context when the waffle switch is present.

amangano-edx: Note: This is just a suggestion, feel free to ignore.

course_discovery/apps/publisher/views.py
```diff
@@ -268,15 +272,27 @@ def post(self, request, *args, **kwargs):
...
+                        course.version = Course.SEAT_VERSION
                     course.changed_by = user
                     course.save()
                     # commit false does not save m2m object. Keyword field is m2m.
                     course_form.save_m2m()
 
+                    # Now create entitlement if we need to
+                    if course.version == Course.ENTITLEMENT_VERSION:
+                        entitlement = entitlement_form.save(commit=False)
+                        entitlement.course = course
```
amangano-edx: The SeatForm overrides `save` and takes course as an argument. It might be good to do something similar here.

course_discovery/apps/publisher/templates/publisher/add_course_form.html
mikix: Yeah probably. I was cargo culting here, didn't consider that solution. That's better, as long as the EntitlementForm doens't need the html side to exist, which I'm guessing not.

course_discovery/apps/publisher/forms.py
mikix: True, but other of the forms do use this ValidationError strategy, seemingly because it's easier to work with / detect in the tests. (Clearing error state on a form wipes out the fake data we inject into it in the tests.)

course_discovery/apps/publisher/views.py
```diff
@@ -268,15 +272,27 @@ def post(self, request, *args, **kwargs):
         course_form = self.course_form(
             request.POST, request.FILES, user=user, organization=organization
         )
-        if course_form.is_valid():
+        entitlement_form = self.entitlement_form(request.POST)
+        if course_form.is_valid() and entitlement_form.is_valid():
```
amangano-edx, line 279: Is `entitlement_form` valid even if it's empty? It might be better to do this check somewhere between lines 291 and 294 and raise an `IntegrityError` if `False`, which I believe would rollback the database.

amangano-edx: Alternatively, you could include the waffle check in the logic here.

```python
does_not_support_entitlements = not waffle.switch_is_active(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH)
if course_form.is_valid() and (does_not_support_entitlements or entitlement_form.is_valid():
# do stuff
```

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
mikix: True. But we're starting to shave yaks now. I originally left these ids unchanged to be minimally-disruptive, before a nit in a previous review suggested the capitalization change. If you prefer a full fix to the SeatPriceBlock mess (which is a totally valid POV), I think I'll undo the changes in this PR and make a separate PR for that cleanup to keep work separated. But if this is a small nit comment, I'll leave it as-is for now to reduce PR scope creep.

course_discovery/apps/publisher/templates/publisher/add_course_form.html
mikix: This is a django template, which is autoescaped, right?

course_discovery/apps/publisher/tests/test_views.py
mikix: It just creates a Switch object in the database, which gets reset because we are using Django's TestCase class which clears the database between tests. So I believe we're safe.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
amangano-edx: I think it's fine to leave it as is.

course_discovery/apps/publisher/views.py
mikix: entitlement_form is valid when it's empty (this is tested by test_valid_price with arguments of None, None).

If you don't feel strongly about moving it, I'd prefer to keep the check where it is, for symmetry with course_form and because it falls down to a raised exception below automatically (rather than moving it and explicitly raising a new exception if it's false).

course_discovery/apps/publisher/templates/publisher/add_course_form.html
mikix: As we discussed in person, the translated text is *not* autoescaped. But we filed a separate ticket for discovery as a whole.

mikix:
OK, I've updated this PR to address your comments, @amangano-edx. Should be good for another look.

course_discovery/apps/publisher/templates/publisher/add_course_form.html
```diff
@@ -128,6 +128,27 @@ <h1 class="hd-1 emphasized">
                         {% endif %}
                     </div>
                 </div>
+
+                {% if entitlement_form %}
+                <div class="field-title">{% trans "CERTIFICATE TYPE AND PRICE" %}</span></div>
```
jlajoie, line 133: Does it make sense to have the translation be from all caps, wouldn't it make more sense to have that stylization be handled through the django `upper` filter?

jlajoie: Nevermind, just saw that the rest of the file is done this way and this will be consistent.

mikix: Great question -- you are probably right. However, this string exists already on the course run page. So doing this one the right way would introduce a new translation string.

mikix: Heh whoops, sorry. Github didn't show me your second comment.

jlajoie (approved):

course_discovery/apps/publisher/forms.py
```diff
@@ -442,6 +442,45 @@ def reset_credit_to_default(self, seat):
...
+
+    mode = forms.ChoiceField(choices=MODE_CHOICES, required=False, label=_('Enrollment Track'),
+                             initial=CourseEntitlement.VERIFIED)
+    price = forms.DecimalField(max_digits=6, decimal_places=2, required=False, initial=0.00)
+
+    class Meta:
+        fields = ('mode', 'price')
+        model = CourseEntitlement
+
+    def save(self, commit=True, course=None):  # pylint: disable=arguments-differ
```
amangano-edx, line 460: The SeatForm takes a `changed_by` keyword argument. Is that something we'll need for the `CourseEntitlement` model?

course_discovery/apps/publisher/forms.py
```diff
@@ -442,6 +442,45 @@ def reset_credit_to_default(self, seat):
...
+
+        return entitlement
+
+    def clean(self):
+        cleaned_data = super().clean()
+        mode = cleaned_data.get('mode')
+        price = cleaned_data.get('price')
+
+        has_mode = bool(mode)
+        has_price = price is not None and price > 0
```
amangano-edx, line 477: Nit: You could simplify this as: `has_price = price and price > 0`

course_discovery/apps/publisher/forms.py
```diff
@@ -442,6 +442,45 @@ def reset_credit_to_default(self, seat):
...
+        if has_mode != has_price:
```
amangano-edx, line 478: Nit: This is a little hard to parse. I think something like `if has_mode and not has_price:` is clearer.

amangano-edx: Though, I guess you would lose out on the validation that ensures price must be null when mode is null, so maybe it's better to keep it like you have it.

course_discovery/apps/publisher/templates/publisher/add_course_form.html
amangano-edx: Nit: It's easier to see the beginning/end of the if block when the enclosed code is indented.

course_discovery/static/js/publisher/seat-type-change.js
```diff
@@ -1,5 +1,5 @@
-$(document).on('change', '#id_type', function (e) {
-    var $seatBlock = $("#SeatPriceBlock"),
+$(document).on('change', '#id_type, #id_mode', function (e) {
```
amangano-edx, line 1: Sanity Check: It won't cause any issues if `#id_mode` is not in the DOM right (it won't be on the add_courserun_form)?

course_discovery/apps/publisher/views.py
```diff
@@ -327,6 +342,7 @@ def post(self, request, *args, **kwargs):
         ctx.update(
             {
                 'course_form': course_form,
+                'entitlement_form': entitlement_form,
```
amangano-edx: I think this needs to be added conditionally, otherwise we may show the entitlement form even when the waffle flag is off after an initial POST request fails with an error.

amangano-edx (changes requested):
Looks good overall, but I think we need to change `post` to ensure that the entitlement form isn't mistakenly rendered on error.

course_discovery/static/js/publisher/seat-type-change.js
mikix: No issue that I can see. For example, in the course creation page, #id_type doesn't exist either. jquery seems to handle it gracefully.

course_discovery/apps/publisher/views.py
mikix: Right you are -- I missed that when changing how the form worked, thanks. Done now (plus a nit above).

course_discovery/apps/publisher/forms.py
mikix: Actually, that can change has_price into None instead of False. I want them both to be bools for comparison's sake.

mikix:
OK, should be good for another look. Fixed the context setting. Squashed and updated translations too.

course_discovery/apps/publisher/views.py
```diff
@@ -268,15 +273,25 @@ def post(self, request, *args, **kwargs):
         course_form = self.course_form(
             request.POST, request.FILES, user=user, organization=organization
         )
-        if course_form.is_valid():
+        entitlement_form = self.entitlement_form(request.POST)
+        if course_form.is_valid() and entitlement_form.is_valid():
             try:
                 with transaction.atomic():
+                    support_entitlements = waffle.switch_is_active(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH)
```
amangano-edx: If you move this to the top of the method, you can re-use it on line 347.

mikix: Gah yes. Fixed. That's what I get for quick fixin'.

amangano-edx (approved):

PR #1300, opened 2018-01-31
Update the db limit of course short description 

schenedx:
Because we have character encoding for for foreign characters like `ó` and `ñ`, the character count on discovery be increased a little to account for that.
EDUCATOR-2244


schenedx:
FYI @edx/learner

mjfrey (approved):

bderusha (approved):

PR #1301, opened 2018-02-02
Publish entitlements along with seats

mikix:
When publishing a course run to ecommerce and discovery, also push any entitlements for its course to those same services.

This follows the same model as seats -- push serialized data to the ecommerce AtomicPublisher api and update/create models in the discovery database. (Those same models later get fleshed out with data from discovery as part of refresh_course_metadata -- things like SKU and expiration as well as price changes.)

https://openedx.atlassian.net/browse/LEARNER-4031

This would normally need the [ecommerce side](https://github.com/edx/ecommerce/pull/1642) to land first, but since we are guarding this with a waffle switch, I guess it doesn't matter.

### DIFF
--- a/course_discovery/apps/publisher/api/tests/test_utils.py
+++ b/course_discovery/apps/publisher/api/tests/test_utils.py
@@ -1,9 +1,10 @@
 import pytest
 
 from course_discovery.apps.core.utils import serialize_datetime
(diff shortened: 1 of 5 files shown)


jlajoie (approved):
Lgtm 👍

PR #1303, opened 2018-02-02
LEARNER-3771: Adds versioned view for Course Runs

jlajoie:
# Description
[LEARNER-3771](https://openedx.atlassian.net/browse/LEARNER-3771)

Adds methods to the wrappers to detect which version of the course is active. Adds those methods' results to the context for the view. Adds conditional check in view for the `SEAT_VERSION`. Updates tests with one for each `VERSION`.


course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -758,6 +758,28 @@ def _assert_dates(self, response):
                       self.course_run.enrollment_end]:
             self.assertContains(response, value.strftime(self.date_format))
 
+    def test_course_run_with_version(self):
+        """
+        Verify that a SEAT_VERSION course still shows enrollment
+        track data, and an ENTITLEMENT_VERSION course does not
+        """
+        self.client.logout()
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
```
mikix, line 833: Why this? Looks like setUp() does a login for ya.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -758,6 +758,28 @@ def _assert_dates(self, response):
...
+        response = self.client.get(self.page_url)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, 'Enrollment Track')
+
+        # For ENTITLEMENT_VERSION no Enrollment Track should be show
+        self.course.version = Course.ENTITLEMENT_VERSION
+        self.course.save()
+        response = self.client.get(self.page_url)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, 'Enrollment Track')
```
mikix: This should be negated, like NotContains or something, right?

course_discovery/apps/publisher/views.py
```diff
@@ -228,6 +228,8 @@ def get_context_data(self, **kwargs):
                                         if current_owner_role.role == PublisherUserRole.ProjectCoordinator
                                         else _('project coordinator'))
         context['is_in_preview_review'] = course_run.is_in_preview_review
+        context['is_seat_version'] = course_run.is_seat_version
+        context['is_entitlement_version'] = course_run.is_entitlement_version
```
mikix, line 233: I don't have a strong opinion on this, but you could also just call "course_run.is_seat_version" in the template. There is prior art for your way though right above your changes.

course_discovery/apps/publisher/tests/test_views.py
jlajoie: Yes it should, thanks! The bug with my log statements had this always evaluating to true so this wasn't picked up, I'll fix that.

course_discovery/apps/publisher/tests/test_views.py
jlajoie: The next method declared, `test_detail_page_with_comments` calls these login/logout functions, so I thought there was something where the session wasn't reset between tests, I'll test without it.

mikix (approved):
Works for me.

PR #1304, opened 2018-02-05
WIP LEARNER-3878 Create entitlement-aware CourseRuns in Publisher

amangano-edx:
Modifies `CreateCourseRunView`, `CreateRunFromDashboardView`, and related templates/forms to support creating Seats from CourseEntitlements associated with the parent course.

JIRA: https://openedx.atlassian.net/browse/LEARNER-3878

Need to add unit tests and determine what to do about CourseSearchForm.

### DIFF
--- a/course_discovery/apps/publisher/forms.py
+++ b/course_discovery/apps/publisher/forms.py
@@ -210,7 +210,7 @@ class CourseSearchForm(forms.Form):
     """ Course Type ahead Search Form. """
     course = forms.ModelChoiceField(
         label=_('Find Course'),
-        queryset=Course.objects.all(),
+        queryset=Course.objects.none(),
         widget=autocomplete.ModelSelect2(
             url='publisher:api:course-autocomplete',
             attrs={
@@ -219,6 +219,14 @@ class CourseSearchForm(forms.Form):
         ),
         required=True,
     )
+
+    def __init__(self, *args, **kwargs):
+        qs = kwargs.pop('queryset')
+        super(CourseSearchForm, self).__init__(*args, **kwargs)
+
+        if qs:
+            self.fields['course'].queryset = qs
+
 
 
 class CourseRunForm(BaseForm):
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -100,6 +100,12 @@ class Course(TimeStampedModel, ChangedByMixin):
     def __str__(self):
         return self.title
 
+    def uses_entitlements(self):
+        """
+        Returns a bool indicating whether or not this Course has been configured to use entitlement products.
+        """
+        return self.version == self.ENTITLEMENT_VERSION
+
     @property
     def post_back_url(self):
         return reverse('publisher:publisher_courses_edit', kwargs={'pk': self.id})
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -714,14 +714,14 @@ class CreateRunFromDashboardView(CreateCourseRunView):
     def get_context_data(self, **kwargs):
         context = {
             'cancel_url': reverse('publisher:publisher_dashboard'),
-            'course_form': self.course_form(),
+            'course_form': self.course_form(queryset=Course.objects.none()),
             'run_form': self.run_form(),
             'seat_form': self.seat_form()
         }
         return context
 
     def post(self, request, *args, **kwargs):
-        course_form = self.course_form(request.POST)
+        course_form = self.course_form(request.POST, queryset=Course.objects.all())
         run_form = self.run_form(request.POST)
         ctx_overrides = {'run_form': run_form}
 
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -668,14 +668,14 @@ class CreateCourseRunView(mixins.LoginRequiredMixin, mixins.PublisherUserRequire
     def get_context_data(self, **kwargs):
         parent_course = self._get_parent_course()
         last_run = self._get_last_run(parent_course)
-        run_form = self._initialize_run_form(last_run)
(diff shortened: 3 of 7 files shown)


course_discovery/apps/publisher/forms.py
```diff
@@ -210,7 +210,7 @@ class CourseSearchForm(forms.Form):
     """ Course Type ahead Search Form. """
     course = forms.ModelChoiceField(
         label=_('Find Course'),
-        queryset=Course.objects.all(),
+        queryset=Course.objects.none(),
```
amangano-edx: This might not be the correct setting for this, but I don't think that `all` is either, as `all` seems to mean that the form renders with all courses included as options for the select, and also that any user can choose any course when submitting the form. I need to spend some more time tomorrow figuring out what makes sense here.

course_discovery/apps/publisher/views.py
```diff
@@ -665,27 +666,42 @@ def _process_post_request(self, request, parent_course, run_form, seat_form, ctx
...
-        })
+        ctx_overrides = {'run_form': run_form}
+
+        if parent_course.uses_entitlements():
+            try:
+                seat_form = self._bind_seat_form_from_entitlement(parent_course)
+            except (CourseEntitlement.DoesNotExist, CourseEntitlement.MultipleObjectsReturned):
+                messages.error(
+                    request,
+                    _('The certificate configuration for this course is incorrect. Please fix it and try again.')
```
amangano-edx, line 697: I'm not sure what to do in this case. This code path should only be hit for courses that have `version` set to `ENTITLEMENT_VERSION` and zero or more than one entitlement.

course_discovery/apps/publisher/forms.py
```diff
@@ -220,6 +220,14 @@ class CourseSearchForm(forms.Form):
         required=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        qs = kwargs.pop('queryset')
+        super(CourseSearchForm, self).__init__(*args, **kwargs)
+
+        if qs:
+            self.fields['course'].queryset = qs
```
amangano-edx, line 228: This will probably change. I need to do some more research to determine what makes the most sense here.

course_discovery/apps/publisher/forms.py
amangano-edx, line 213: This will probably change. I need to do some more research to determine what makes the most sense here.

course_discovery/static/js/publisher/show-hide-seat-form.js
```diff
@@ -0,0 +1,30 @@
...
+            $seatForm.find('#seatPriceBlock').addClass('hidden');
+            $seatForm.find('#id_price').val('0.0');
+            $seatForm.find('#creditPrice').addClass('hidden');
+            $seatForm.find('#id_credit_price').val('0.0');
+
+            // Re-attach the form
+            $seatForm.insertAfter($courseRunForm);
+        }
+    });
+});
```
amangano-edx, line 30: I'm not sure if this is an acceptable approach. It seems to work ok, but it's pretty ugly. Curious what @AlasdairSwan thinks about this.

course_discovery/apps/publisher/views.py
```diff
@@ -698,29 +714,38 @@ def _get_parent_course(self):
...
+        ctx_overrides['course_form'] = course_form
         self.parent_course = course_form.cleaned_data.get('course')
+        if self.parent_course.uses_entitlements():
+            ctx_overrides['hide_seat_form'] = True
+            try:
+                seat_form = self._bind_seat_form_from_entitlement(self.parent_course)
+            except (CourseEntitlement.DoesNotExist, CourseEntitlement.MultipleObjectsReturned):
+                messages.error(
+                    request,
+                    _('The certificate configuration for this course is incorrect. Please fix it and try again.')
```
amangano-edx, line 743: I'm not sure what to do in this case. This code path should only be hit for courses that have `version` set to `ENTITLEMENT_VERSION` and zero or more than one entitlement.

course_discovery/static/js/publisher/show-hide-seat-form.js
AlasdairSwan: @amangano-edx it is pretty ugly, but then it's not using any JS Frameworks so you're limited in your approach I guess. Is this up in a test environment? I think I need some more details to consider alternative approaches? If not I can swing by your desk at some point today.

course_discovery/apps/publisher/models.py
```diff
@@ -100,6 +100,12 @@ class Course(TimeStampedModel, ChangedByMixin):
     def __str__(self):
         return self.title
 
+    def uses_entitlements(self):
```
mikix, line 103: Might be nicer as a @property, but fine as-is too.

course_discovery/static/js/publisher/show-hide-seat-form.js
```diff
@@ -0,0 +1,30 @@
...
+    // If the rendered form is hidden, remove it from the DOM.
+    if ($seatForm.hasClass('hidden')) {
+        $seatForm.detach();
+        $seatForm.removeClass('hidden');
+    }
+
+    $(document).on('select2:select', '#id_course', function(e) {
+        var usesEntitlements = e.params.data.uses_entitlements;
+
+        $seatForm.detach();
```
AlasdairSwan, line 14: Why detach here, if ```!usesEntitlements``` you don't need to detach do you?

amangano-edx: That's true.

amangano-edx: You need to detach before re-attaching to ensure that the form is accidentally added twice.

amangano-edx:
Replaced by https://github.com/edx/course-discovery/pull/1306

PR #1306, opened 2018-02-07
LEARNER-3878 Create entitlement-aware courseruns in publisher

amangano-edx:
Modifies CreateCourseRunView, CreateRunFromDashboardView, and related templates/forms to support creating Seats from CourseEntitlements associated with the parent course.

JIRA: https://openedx.atlassian.net/browse/LEARNER-3878

Replaces https://github.com/edx/course-discovery/pull/1304


course_discovery/apps/publisher/views.py
```diff
@@ -615,12 +616,42 @@ def _render_post_error(self, request, ctx_overrides=None, status=400):
...
+        context['run_form'] = run_form
+
+        if parent_course.uses_entitlements:
+            context['hide_seat_form'] = True
+
+            # Fail if Seat fields are present in the POST data.
+            seat_data_in_form = any([key for key in self.seat_form.declared_fields.keys() if key in request.POST])
+            if seat_data_in_form:
+                messages.error(
+                    request, _('The page could not be updated. Make sure that all values are correct, then try again.')
```
amangano-edx, line 645: Need to update this with a better error message.

course_discovery/apps/publisher/views.py
```diff
@@ -615,12 +616,42 @@ def _render_post_error(self, request, ctx_overrides=None, status=400):
...
+            try:
+                entitlement = parent_course.entitlements.get()
+                seat_form = self.seat_form({
+                    'type': CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING[entitlement.mode],
+                    'price': entitlement.price
+                })
+            except (CourseEntitlement.DoesNotExist, CourseEntitlement.MultipleObjectsReturned):
+                messages.error(
+                    request,
+                    _('The certificate configuration for this Course is incorrect. Please fix it, then try again.')
```
amangano-edx: Need to confirm with product that this error message is OK.

course_discovery/apps/publisher/tests/factories.py
```diff
@@ -28,7 +28,7 @@ class CourseFactory(factory.DjangoModelFactory):
     learner_testimonial = FuzzyText()
     level_type = factory.SubFactory(factories.LevelTypeFactory)
     image = factory.django.ImageField()
-    version = FuzzyInteger(0, 1)
+    version = 0
```
jlajoie: I know I implemented this originally with the FuzzyInteger to use 0 or 1, but would it maybe make sense here to use the SEAT_VERSION constant here instead so it's more clear what this value is?

jlajoie (approved):

course_discovery/apps/publisher/tests/factories.py
amangano-edx: Changed.

course_discovery/apps/publisher/forms.py
```diff
@@ -220,6 +220,13 @@ class CourseSearchForm(forms.Form):
         required=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        qs = kwargs.pop('queryset', None)
+        super(CourseSearchForm, self).__init__(*args, **kwargs)
+
+        if qs:
+            self.fields['course'].queryset = qs
```
mikix: This works just fine for ya? Good to hear, this does look nicer than my branch. I'll follow suit once yours lands.

course_discovery/apps/publisher/models.py
```diff
@@ -508,6 +515,11 @@ class CourseEntitlement(TimeStampedModel):
         'default': 0.00,
     }
 
+    MODE_TO_SEAT_TYPE_MAPPING = {
+        VERIFIED: Seat.VERIFIED,
+        PROFESSIONAL: Seat.PROFESSIONAL
+    }
```
mikix, line 521: This is technically the right thing to do here, but feels like overkill? I mean, we know they are going to be equal. Or we could define VERIFIED = Seat.VERIFIED at the model level. Neither of which is as safe as what you're doing in case any strings change, but both involve less complexity. I dunno, just a thought.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -349,14 +349,19 @@ def test_invalid_entitlement_price(self, price):
...
     """ Tests for the publisher `CreateCourseRunView`. """
 
     def setUp(self):
         super(CreateCourseRunViewTests, self).setUp()
         self.user = UserFactory()
         self.course_run = factories.CourseRunFactory()
+
         self.course = self.course_run.course
+        self.course.version = Course.SEAT_VERSION
+        self.course.save()
```
mikix, line 363: You don't *need* this block anymore, now that you default to Seat. But if you don't want to assume, keep it. You do this elsewhere in this PR. Dunno if that's intentional or just cleanup to make after the change of default.

course_discovery/apps/publisher/views.py
mikix: This is hopefully either a "we seriously are messed up right now" message or a "you're trying to hack me" message. I wouldn't stress about wording.

You could be more explicit and say something like "type and price information was received but was not expected" or something like that, to help narrow down why someone was seeing this unexpected flow?

course_discovery/apps/publisher/views.py
```diff
@@ -615,12 +616,42 @@ def _render_post_error(self, request, ctx_overrides=None, status=400):
...
+        context = context or {}
+
+        run_form = self.run_form(request.POST)
+        context['run_form'] = run_form
+
+        if parent_course.uses_entitlements:
+            context['hide_seat_form'] = True
+
+            # Fail if Seat fields are present in the POST data.
+            seat_data_in_form = any([key for key in self.seat_form.declared_fields.keys() if key in request.POST])
```
mikix, line 642: You don't need any() here. Non-empty lists are already True-y.

Is checking self.seat_form.declared_fields.keys() the right thing? Maybe we might add fields to SeatForm and CourseRunForm that overlap in the future? But I guess we'd catch that quickly with this check and could re-evaluate then.

course_discovery/apps/publisher/views.py
mikix: Like above error message, I wouldn't beat yourself up over perfect wording for this.

If you wanted to be more verbose, you could say that you expected 1 entitlement, but got N.

course_discovery/apps/publisher/views.py
```diff
@@ -635,16 +666,17 @@ def _process_post_request(self, request, parent_course, run_form, seat_form, ctx
...
         if not (run_form.is_valid() and seat_form.is_valid()):
             messages.error(
                 request, _('The page could not be updated. Make sure that all values are correct, then try again.')
             )
-            return self._render_post_error(request, ctx_overrides=ctx_overrides)
+            return self._render_post_error(request, ctx_overrides=context)
 
         try:
             with transaction.atomic():
+                user = request.user
```
mikix, line 696: Where does this get used?

mikix (approved):
Left comments, but nothing is a blocker. Looks nice!

course_discovery/apps/publisher/views.py
amangano-edx: I confirmed with Deen that this message is fine as is.

course_discovery/apps/publisher/views.py
amangano-edx: I confirmed with Deen that this message is ok as is.

course_discovery/apps/publisher/forms.py
amangano-edx: Yea, this works fine, but you cannot access `fields` until after the call to `super`. Credit to stackoverflow for this https://stackoverflow.com/questions/5329586/django-modelchoicefield-filtering-query-set-and-setting-default-value-as-an-obj

course_discovery/apps/publisher/tests/test_views.py
amangano-edx: Most of these tests rely on `course.version` being `SEAT_VERSION`, so I I prefer to be explicit about that here. If the factory default ever changes, these tests will still pass.

course_discovery/apps/publisher/views.py
amangano-edx: If we ever add overlapping fields to these two forms we'll have other issues, as I'm not sure how we'd be able to tell them apart in the POST data as is. The correct way to handle that is to use the `prefix` argument when instantiating the forms, so that each form will namespace it's data. We're not doing that anywhere in publisher though that I can see, and I felt it was out of scope to address that now.

course_discovery/apps/publisher/views.py
amangano-edx: In the next line 😄

course_discovery/apps/publisher/models.py
amangano-edx: Right now they're equal, but there's no guarantee that they will be in the future. 

I don't think it adds much complexity. We just key into a map instead of using `entitlement.mode` to directly assign `seat.type`

course_discovery/apps/publisher/views.py
mikix: Aha, I see, and this line was removed way up above and moved here. Got it.

course_discovery/apps/publisher/forms.py
amangano-edx: Hmm curiously this doesn't seem to be working the way I remembered it. I will switch to do it the way you have it.

amangano-edx: Actually, nvm, I will keep it as is. I didn't realize `Course.object.none()` is Falsy, so this code was not evaluating.

PR #1309, opened 2018-02-08
Fix caching of ApiAccessRequest

BenjiLee:
When a user has no ApiRequestModel results, an exception is raised and the result is never cached. This PR ensures that we cache that result.

https://openedx.atlassian.net/wiki/spaces/~clee/pages/547913828/edx-mktg+investigation

### DIFF
--- a/course_discovery/apps/core/api_client/lms.py
+++ b/course_discovery/apps/core/api_client/lms.py
@@ -12,6 +12,10 @@ from course_discovery.apps.api.utils import get_cache_key
 
 logger = logging.getLogger(__name__)
 
+SENTINEL_NO_RESULT = ()
+ONE_HOUR = 60 * 60
+ONE_MINUTE = 60
+
 
 class LMSAPIClient(object):
     """
@@ -23,13 +27,13 @@ class LMSAPIClient(object):
 
     def get_api_access_request(self, user):
         """
-        Get API Access Requests made by the given user.
+        Get ApiAccessRequests made by the given user.
 
         Arguments:
             user (User): Django User.
 
         Returns:
-            (dict): API Access requests made by the given user.
+            (dict): ApiAccessRequests for the given user.
 
         Examples:
             >> user = User.objects.get(username='staff')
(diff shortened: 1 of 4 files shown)


course_discovery/apps/core/api_client/lms.py
```diff
@@ -70,7 +74,12 @@ def get_api_access_request(self, user):
                 cache.set(cache_key, api_access_request, 60 * 60)
             except (SlumberBaseException, ConnectionError, Timeout):
                 logger.exception('Failed to fetch API Access Request from LMS for user "%s".', user.username)
-            except (IndexError, KeyError):
+            except IndexError:
                 logger.info('APIAccessRequest model not found for user [%s].', user.username)
+                cache.set(cache_key, _SENTINEL_EMPTY_RESULT, 60 * 30)
```
robrap: It seems like this could be the same error message as below, but replacing `KeyError` with `IndexError`, and replace `No` with `Empty`.

course_discovery/apps/core/api_client/lms.py
```diff
@@ -70,7 +74,12 @@ def get_api_access_request(self, user):
...
+            except KeyError:
+                logger.info('KeyError: No "results" received when fetching APIAccessRequest model for user [%s].',
+                            user.username)
+                cache.set(cache_key, _SENTINEL_EMPTY_RESULT, 60 * 30)
```
robrap: Assuming this is in seconds, maybe cache all of the for 60, or long enough to complete one long running request without multiple calls? I’m not sure. Also wouldn’t hurt to have constants like ONE_HOUR or HALF_HOUR or ONE_MINUTE.

course_discovery/apps/core/tests/test_api_clients.py
```diff
@@ -81,15 +79,15 @@ def test_get_api_access_request_with_empty_response(self, mock_access_token):  #
             self.partner.lms_url, self.user
         )
         assert self.lms.get_api_access_request(self.user) is None
-        assert 'APIAccessRequest model not found for user [%s].' % self.user.username in \
-               self.log_messages['info']
+        assert 'KeyError: No "results" received when fetching APIAccessRequest model for user [%s].' % \
```
robrap: If you want, update assertions to check if it contains the exception name to make the test more resilient to wording changes.

course_discovery/apps/core/api_client/lms.py
```diff
@@ -70,7 +74,12 @@ def get_api_access_request(self, user):
                 cache.set(cache_key, api_access_request, 60 * 60)
             except (SlumberBaseException, ConnectionError, Timeout):
                 logger.exception('Failed to fetch API Access Request from LMS for user "%s".', user.username)
-            except (IndexError, KeyError):
+            except IndexError:
```
robrap: Exception above also needs to be cached to not hammer the lms in the event it is already having problems.

robrap (commented):
Thank you.

robrap (approved):
Thanks Chris!  Not sure that the TODO is needed any more.

PR #1311, opened 2018-02-12
Publish to ecommerce after discovery

mikix:
Ecommerce wants to talk to discovery to confirm details of course like UUID. So let's publish to it after we publish to discovery.

Without this, we get errors when publishing, because ecommerce doesn't find a UUID.


jlajoie (approved):
> Question, are there other dependencies that we may be messing up if we move this?

Just pasting this from our discussion in here for completeness, but it seems like we're good to go with these changes :+1:

mikix:
Also pasting from convo: Looks like the only real discovery->ecommerce dep is for refreshing course metadata. Which as far as I can tell doesn't happen during publish, but later

PR #1313, opened 2018-02-12
Filter courses on create run POST

mikix:
When creating a course run in the Publisher dashboard, only allow the user to create runs for courses they have access to.

https://openedx.atlassian.net/browse/LEARNER-4102

**Notes**
* The big change here is the changes to CourseSearchForm to accept a user (from which we create a validation queryset) and in views.py to pass a user when creating the validation version of the form for POSTs.
* I refactored some of the code to reduce the number of times we create that queryset. We still create a very similar queryset in views.py for course runs, but it was different enough that I didn't touch it.
* I didn't have an obvious place to put the get_user_courses() utility function. Utils.py didn't work because it would have created a circular import. So I just made a static method on the PublisherUser model. I'm open to better places.

### DIFF
--- a/course_discovery/apps/publisher/forms.py
+++ b/course_discovery/apps/publisher/forms.py
@@ -210,7 +210,7 @@ class CourseSearchForm(forms.Form):
     """ Course Type ahead Search Form. """
     course = forms.ModelChoiceField(
         label=_('Find Course'),
-        queryset=Course.objects.all(),
+        queryset=Course.objects.none(),
         widget=autocomplete.ModelSelect2(
             url='publisher:api:course-autocomplete',
             attrs={
@@ -221,11 +221,11 @@ class CourseSearchForm(forms.Form):
     )
 
     def __init__(self, *args, **kwargs):
-        qs = kwargs.pop('queryset', None)
-        super(CourseSearchForm, self).__init__(*args, **kwargs)
+        user = kwargs.pop('user', None)
+        super().__init__(*args, **kwargs)
 
-        if qs is not None:
-            self.fields['course'].queryset = qs
+        if user:
+            self.fields['course'].queryset = PublisherUser.get_courses(user)
 
 
 class CourseRunForm(BaseForm):
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -12,6 +12,7 @@ from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 from django_extensions.db.models import TimeStampedModel
 from django_fsm import FSMField, transition
+from guardian.shortcuts import get_objects_for_user
 from simple_history.models import HistoricalRecords
 from sortedm2m.fields import SortedManyToManyField
 from stdimage.models import StdImageField
@@ -26,7 +27,7 @@ from course_discovery.apps.ietf_language_tags.models import LanguageTag
 from course_discovery.apps.publisher import emails
 from course_discovery.apps.publisher.choices import (CourseRunStateChoices, CourseStateChoices, InternalUserRole,
                                                      PublisherUserRole)
-from course_discovery.apps.publisher.utils import is_email_notification_enabled
+from course_discovery.apps.publisher.utils import is_email_notification_enabled, is_internal_user, is_publisher_admin
 from course_discovery.apps.publisher.validators import ImageMultiSizeValidator
 
 logger = logging.getLogger(__name__)
@@ -907,3 +908,22 @@ class PublisherUser(User):
 
     class Meta:
         proxy = True
+
+    @staticmethod
+    def get_courses(user, queryset=None):
+        if queryset is None:
+            queryset = Course.objects.all()
+
+        if is_publisher_admin(user):
+            return queryset
+        elif is_internal_user(user):
+            return queryset.filter(course_user_roles__user=user).distinct()
+        else:
+            organizations = get_objects_for_user(
+                user,
+                OrganizationExtension.VIEW_COURSE,
+                OrganizationExtension,
+                use_groups=True,
+                with_superuser=False
+            ).values_list('organization')
+            return queryset.filter(organizations__in=organizations)
--- a/course_discovery/apps/publisher/tests/test_forms.py
+++ b/course_discovery/apps/publisher/tests/test_forms.py
@@ -4,6 +4,7 @@ import ddt
 import pytest
 from django.core.exceptions import ValidationError
 from django.test import TestCase
+from guardian.shortcuts import assign_perm
 from pytz import timezone
 from waffle.testutils import override_switch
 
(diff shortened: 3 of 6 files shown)


course_discovery/apps/publisher/forms.py
```diff
@@ -220,6 +220,13 @@ class CourseSearchForm(forms.Form):
         required=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop('user', None)
+        if self.user:
+            self.declared_fields['course'].queryset = PublisherUser.get_courses(self.user)
```
amangano-edx: I think this would change the queryset globally wouldn't it? It might be safer to modify `self.fields` instead after `super()` is called.

course_discovery/apps/publisher/models.py
```diff
@@ -895,3 +896,22 @@ class PublisherUser(User):
...
+            queryset = Course.objects.all()
+
+        if is_publisher_admin(user):
+            return queryset
+        elif is_internal_user(user):
+            return queryset.filter(course_user_roles__user=user).distinct()
+        else:
+            organizations = get_objects_for_user(
+                user,
+                OrganizationExtension.VIEW_COURSE,
```
amangano-edx, line 924: Would it make more sense to use the EDIT_COURSE permission if we're using it to create course_runs for a course?

amangano-edx: Maybe the permission level could be configurable via a kwarg, so that views that only need the VIEW_COURSE permission can pass that in explicitly?

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -417,3 +419,56 @@ def test_create_audit_seat_for_credit_and_verified_course_runs(self, seat_type):
...
+        self.course_form = CourseSearchForm(user=self.user)
+
+    def test_no_user_no_courses(self):
+        course_form = CourseSearchForm()
+        self.assertEqual(Course.objects.none(), course_form.fields['course'].queryset)
+
+    def test_with_course_team(self):
+        """ Verify course autocomplete returns data for course team user. """
+        self.course_form.cleaned_data = {'course': self.course.id}
+        self.course_form.clean()
```
amangano-edx: I'm not sure I follow what's going on in this test. Could you please explain?

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -417,3 +419,56 @@ def test_create_audit_seat_for_credit_and_verified_course_runs(self, seat_type):
...
+
+    def test_with_admin(self):
+        """ Verify course autocomplete lets an admin access courses they aren't associated with. """
+        self.user.groups.remove(self.organization_extension.group)
+        self.user.groups.add(Group.objects.get(name=ADMIN_GROUP_NAME))
+        self.course_form.cleaned_data = {'course': self.course.id}
+        self.course_form.clean()
```
amangano-edx: I'm not sure I follow what's going on in this test. Could you please explain?

amangano-edx: What are you testing by calling `clean` without asserting anything?

course_discovery/apps/publisher/tests/test_forms.py
amangano-edx: What are you testing by calling `clean` without asserting anything?

amangano-edx: Does the data still get cleaned again if it's already been marked clean by being added directly to `clean_data`?

course_discovery/apps/publisher/tests/test_forms.py
amangano-edx: Does the data still get cleaned again if it's already been marked clean by being added directly to `clean_data`?

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -417,3 +419,56 @@ def test_create_audit_seat_for_credit_and_verified_course_runs(self, seat_type):
...
+            OrganizationExtension.VIEW_COURSE, self.organization_extension.group, self.organization_extension
+        )
+
+        self.course_form = CourseSearchForm(user=self.user)
+
+    def test_no_user_no_courses(self):
+        course_form = CourseSearchForm()
+        self.assertEqual(Course.objects.none(), course_form.fields['course'].queryset)
+
+    def test_with_course_team(self):
```
amangano-edx: If this is explicitly testing what happens with a course team user, it would be better to explicitly create such a user in the test, rather than relying on that being the default from the `Factories`.

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -417,3 +419,56 @@ def test_create_audit_seat_for_credit_and_verified_course_runs(self, seat_type):
...
+        """ Verify course autocomplete returns data for course team user. """
+        self.course_form.cleaned_data = {'course': self.course.id}
+        self.course_form.clean()
+
+    def test_with_admin(self):
+        """ Verify course autocomplete lets an admin access courses they aren't associated with. """
+        self.user.groups.remove(self.organization_extension.group)
```
amangano-edx: It would be good to see that the user doesn't have access after being removed from this group, but does after being added to the Admin group.

course_discovery/apps/publisher/tests/test_forms.py
amangano-edx: It would also be good to verify that a course team user can only access courses for that team.

course_discovery/apps/publisher/forms.py
mikix: I don't think it changes anything globally, no. `self.declared_fields` is set up by django's Form class in a `__new__` override, which gets called before `__init__.` So it looks like it's designed to allow this kind of usage.

And our CourseForm does this same thing (using declared_fields before super to override aspects based on `__init__` args). So I figured this was the "right" thing to do.

That said, I don't know if it's better or worse than `self.fields`. I was just aping an existing form.

course_discovery/apps/publisher/models.py
mikix: I dunno. You're not editing a course in this flow (in the sense a user would think of -- clicking the Edit button in publisher). So it's not clear that EDIT_COURSE is more correct.

The original code had VIEW_COURSE for this use case (the courses you are allowed to choose in the autocomplete form), so that's what I kept. But I could be persuaded that EDIT_COURSE is better.

amangano-edx (approved):
I left a few comments about the tests, but overall, looks good.

course_discovery/apps/publisher/models.py
amangano-edx: I think it would probably be best to check with Deen, and/or someone from the educator team.

mikix: I'd be very leery of making any change there -- teams already have a workflow and permissions set up. Changing what permissions course team members need for their flow would likely just create problems.

Unless you think using VIEW is actually busted, my preference is to leave existing behavior in place, since it doesn't strike me as obviously semantically wrong.

course_discovery/apps/publisher/tests/test_forms.py
mikix: I have fixed the tests to be a little clearer. But let me answer these questions, even though the current code tests in a different way.

Django Fields or Forms typically throw ValidationErrors when clean() is called on them and they fail validation. So if a clean() call doesn't raise an exception, it passed the test.

The cleaned_data thing is a quirk of how clean() works. A Form starts with a call to full_clean(), sets up a bunch of stuff, calls clean() on all sub-fields and then calls self.clean() which is a hook for subclasses to do further testing. But first it fills self.cleaned_data with all the cleaned values in fields and self.clean() is supposed to look at that data when it does its own validation.

A lot of our tests simply set up cleaned_data before calling clean() to simulate that flow in a simple way (it's a little awkward to iterate on the full_clean() flow because you have to clear errors, etc).

Anyway. I made the tests smaller and don't need to iterate on a form anymore. So I no longer use cleaned_data or clean() (which I don't even define anymore).

PR #1315, opened 2018-02-14
LEARNER-3770 Publisher entitlement-aware edit-courserun views

amangano-edx:
Updates the CourseRunEditView to properly handle CourseRuns that use entitlements.

JIRA: https://openedx.atlassian.net/browse/LEARNER-3770

### DIFF
--- a/course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
+++ b/course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
@@ -173,35 +173,37 @@
                     </div>
                 </div>
 
-                <div class="field-title">{% trans "CERTIFICATE TYPE AND PRICE" %} <span class="required float-right">* Required for review</span></div>
-                <div class="row">
-                    <div class="col col-6 help-text">
-                        {% trans "If the course offers a verified or professional education certificate, select the certificate type and enter the price for the certificate." %}
-                    </div>
-                    <div class="col col-6">
-                        <div class="row">
-                            <div class="col col-6">
-                                <label class="field-label ">{{ seat_form.type.label_tag }}
-                                </label> {{ seat_form.type }}
-                            </div>
-                            <div id="seatPriceBlock" class="col col-6 {% if seat_form.type.value == 'audit' or not seat_form.price.value %}hidden{% endif %}">
-                                <label class="field-label ">{{ seat_form.price.label_tag }}</label>
-                                {{ seat_form.price }}
-                                <div class="{% if seat_form.type.value != 'credit' or not seat_form.credit_price.value %}hidden{% endif %}" id="creditPrice">
-                                    <label class="field-label">{{ seat_form.credit_price.label }}</label>
-                                    {{ seat_form.credit_price }}
+                {% if seat_form %}
+                    <div class="field-title">{% trans "CERTIFICATE TYPE AND PRICE" %} <span class="required float-right">* Required for review</span></div>
+                    <div class="row">
+                        <div class="col col-6 help-text">
+                            {% trans "If the course offers a verified or professional education certificate, select the certificate type and enter the price for the certificate." %}
+                        </div>
+                        <div class="col col-6">
+                            <div class="row">
+                                <div class="col col-6">
+                                    <label class="field-label ">{{ seat_form.type.label_tag }}
+                                    </label> {{ seat_form.type }}
+                                </div>
+                                <div id="seatPriceBlock" class="col col-6 {% if seat_form.type.value == 'audit' or not seat_form.price.value %}hidden{% endif %}">
+                                    <label class="field-label ">{{ seat_form.price.label_tag }}</label>
+                                    {{ seat_form.price }}
+                                    <div class="{% if seat_form.type.value != 'credit' or not seat_form.credit_price.value %}hidden{% endif %}" id="creditPrice">
+                                        <label class="field-label">{{ seat_form.credit_price.label }}</label>
+                                        {{ seat_form.credit_price }}
+                                    </div>
                                 </div>
                             </div>
+                            {% if seat_form.price.errors %}
+                                <div class="field-message has-error">
+                                    <span class="field-message-content">
+                                        {{ seat_form.price.errors|escape }}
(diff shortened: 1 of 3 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -831,8 +833,19 @@ def post(self, request, *args, **kwargs):
         run_form = self.run_form(
             request.POST, instance=course_run, is_project_coordinator=context.get('is_project_coordinator')
         )
-        seat_form = self.seat_form(request.POST, instance=course_run.seats.first())
```
amangano-edx, line 834: This looks like it might be wrong, as the GET context explicitly tries to load paid seats first (see line 805). It seems like we should also be doing that here for the instance.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -173,35 +173,37 @@ <h1 class="hd-1 emphasized">
...
+                                    </div>
                                 </div>
                             </div>
+                            {% if seat_form.price.errors %}
+                                <div class="field-message has-error">
+                                    <span class="field-message-content">
+                                        {{ seat_form.price.errors|escape }}
+                                    </span>
+                                </div>
+                            {% endif %}
```
amangano-edx, line 203: This is all just indentation. I didn't change any of this code.

course_discovery/apps/publisher/views.py
amangano-edx: I created a bug to address this separately from this PR https://openedx.atlassian.net/browse/EDUCATOR-2337

course_discovery/apps/publisher/views.py
```diff
@@ -802,11 +802,13 @@ def get(self, request, *args, **kwargs):
...
-            context['seat_form'] = self.seat_form(instance=course_run_paid_seat)
-        else:
-            context['seat_form'] = self.seat_form(instance=course_run.seats.first())
+
+        if not course.uses_entitlements:
+            course_run_paid_seat = course_run.seats.filter(type__in=PAID_SEATS).first()
+            if course_run_paid_seat:
+                context['seat_form'] = self.seat_form(instance=course_run_paid_seat)
+            else:
+                context['seat_form'] = self.seat_form(instance=course_run.seats.first())
```
mikix, line 811: There's an opportunity for cleanup here, but no big deal, it's not your code:
```
seat = course_run.seats.filter(type__in=PAID_SEATS).first() or course_run.seats.first()
context['seat_form'] = self.seat_form(instance=seat)
```

course_discovery/apps/publisher/views.py
mikix: Yeah, that's suspicious. Leaving it alone is safe. But if you're feeling fancy, worth a fix up investigation. The risk is that we're posting with an audit seat as a base instead of verified? But all the fields are going to get set by POST data anyway right?

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -173,35 +173,37 @@ <h1 class="hd-1 emphasized">
...
-                                <span class="field-message-content">
-                                    {{ seat_form.price.errors|escape }}
-                                </span>
-                            </div>
-                        {% endif %}
                     </div>
-                </div>
+                {% endif %}
 
                 <div class="field-title">{% trans "COURSE STAFF" %} <span class="required float-right">* Required for review</span></div>
```
amangano-edx, line 208: I'm not sure what to do about the `seat-type-change.js` file that gets included in this template down below. Because it's included in the compression-enabled js block, I cannot add a condition to only include it when the seat_form is present, as the compression tool we use doesn't know how to handle that condition. Therefore, my options are:
- Leave it the way it is (it will always be included with the template, even when the seat_form doesn't render).
-- This doesn't seem to cause any issues, but it might if anyone chages `seat-type-change.js`.
- Move it to the `js-without-compression` block and allow it to be pulled into the form uncompressed.
-- I think this is my preferred solution, but I'm not sure what the implications of doing this would be.

Thoughts?

course_discovery/apps/publisher/views.py
```diff
@@ -831,8 +833,19 @@ def post(self, request, *args, **kwargs):
         run_form = self.run_form(
             request.POST, instance=course_run, is_project_coordinator=context.get('is_project_coordinator')
         )
-        seat_form = self.seat_form(request.POST, instance=course_run.seats.first())
-        if run_form.is_valid() and seat_form.is_valid():
+        context['run_form'] = run_form
+        form_data_is_valid = run_form.is_valid()
+
+        if course_run.course.uses_entitlements:
+            seat_data_in_form = any([key for key in self.seat_form.declared_fields.keys() if key in request.POST])
```
mikix, line 840: Don't need any() here. Lists are truthy.

mikix (approved):
Looks good!

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
mikix: Conditional but uncompressed makes sense in the world where we expect 90% of edits to occur on entitlement runs.

If you went with the first way (leave it in) you could always add a comment in `seat-type-change.js` to say "be careful, the form might not always be there" or similar.

course_discovery/apps/publisher/views.py
amangano-edx: I documented the issues with this in the ticket. I'm going to leave the logic as is for now and let whoever takes on that ticket address it.

course_discovery/apps/publisher/views.py
amangano-edx: I'm going to leave it as it was for now.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
amangano-edx: I talked with @arizzitano and decided that it's best to pull it in uncompressed.

PR #1319, opened 2018-02-16
Learner-3768: Adds in edit form for entitlements

jlajoie:
# Description
[LEARNER-3768](https://openedx.atlassian.net/browse/LEARNER-3768)

Adds in to the Course Edit form options for modifying the Entitlement, and allows the ability to switch from a SEAT_VERSION to an ENTITLEMENT_VERSION. Will not allow a user to switch back. Will not allow the switch to occur if any future Publisher Course Runs have Seats with a different Type or Price (excluding Audit seats).

Notes:
* Moved from form_valid method to get and post overrides to handle the multiple forms
* Still need to add in tests


course_discovery/apps/publisher/views.py
```diff
@@ -362,13 +362,40 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
...
         return reverse(self.success_url, kwargs={'pk': self.object.id})
 
+    def get_context_data(self, **kwargs):
+        context = super(CourseEditView, self).get_context_data(**kwargs)
+        history_id = self.request.GET.get('history_id', None)
+        self.object = self.get_object()
+
+        try:
+            history_object = self.object.history.get(history_id=history_id) if history_id else None
+        except:  # pylint: disable=bare-except
```
mikix: Bare excepts should be exceedingly rare. At the least, they should be "except Exception:" because otherwise you're going to catch KeyboardInterrupt and SystemExit exceptions.

But here, maybe just catch DoesNotExist? (unless you want to worry about multiple rows too)

course_discovery/apps/publisher/views.py
```diff
@@ -362,13 +362,40 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
...
+                'is_internal_user': is_internal_user(self.request.user),
+                'history_object': history_object
+            }
+        )
+
+        if waffle.switch_is_active(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH):
+            if self.object.version == Course.ENTITLEMENT_VERSION:
+                context['entitlement_form'] = self.entitlement_form(instance=self.object.entitlements.first())
+            else:
+                context['entitlement_form'] = self.entitlement_form({'mode': ''})
```
mikix, line 395: Maybe this shouldn't be waffle guarded? I dunno. Like, if we have a entitlement object, maybe we should be aware of that so we can safely edit the course.

But ::shrug:: hopefully the switch isn't around much.

course_discovery/apps/publisher/views.py
jlajoie: Not my code sadly though I can definitely update this, not positive what errors are expected here.

course_discovery/apps/publisher/views.py
jlajoie: Yeah I wasn't super certain on this, it feels like you shouldn't be able to mess with or modify anything with entitlements if the switch is disabled, though it might be a bit weird to have prices set at the Course level but not be able to do anything to them. I could always check for the course version first, and if it's the ENTITLEMENT_VERSION go ahead and assume the switch is enabled, but if it's SEAT_VERSION check the switch?

mikix: I dunno. I think it's probably "safer" to just guard everything to do with this feature behind the switch. I think what you have is fine.

course_discovery/apps/publisher/views.py
mikix: Ah, I hadn't read far enough to see that this was a copy paste job. Can you at least update it to "except Exception" then? Bare excepts hurt my soul.

jlajoie: Bare except vs Broad except, but you're right it's the smallest amount better :(

mikix: My distaste for bare excepts are more about quality than quantity. The kind of things that a bare except catches just shouldn't be caught. You don't want this try/except to intercept a Ctrl+C (KeyboardInterrupt), and you don't want it to intercept a sys.exit() (SystemExit).

mikix: Python lets you catch some surprising things. :-/

course_discovery/apps/publisher/views.py
```diff
@@ -393,33 +420,131 @@ def get_form_kwargs(self):
...
+        return misconfigured_price_runs, misconfigured_seat_type_runs
+
+    def _create_or_update_course_entitlement(self, course, entitlement_form):
+        entitlement = course.entitlements.first()
+        if entitlement:
+            entitlement.price = entitlement_form.cleaned_data.get('price')
+            entitlement.mode = entitlement_form.cleaned_data.get('mode')
+            entitlement.save()
+        else:
+            entitlement_form.save(course=course)
```
mikix, line 453: Can we hook the form up properly (maybe initial=entitlement?), so we don't need this method and form.save() does the right thing?

course_discovery/apps/publisher/views.py
```diff
@@ -393,33 +420,131 @@ def get_form_kwargs(self):
...
+                        messages.error(request, error_message)
+                    if seat_misconfigurations or type_misconfigurations:
+                        return self._render_post_error(request, ctx_overrides={
+                            'course_form': course_form,
+                            'entitlement_form': entitlement_form
+                        })
+                    else:
+                        with transaction.atomic():
+                            course = course_form.save(commit=False)
+                            course.changed_by = user
```
mikix: One thing you could do here is add changed_by as a kwarg for save (I believe SeatForm does this). It wouldn't save you much here because you also always also set version. But maybe add that as a kwarg too? And then you can avoid the commit=False/save() pairing. But no big deal if you find that grosser.

mikix: These blocks all feel very similar. If self._create_or_update_course_entitlement() was a no-op for SEAT_VERSION, you could probably have one single block at the bottom like:

```
course = course_form.save(commit=False)
course.changed_by = user
course.version = ENTITLEMENT if entitlement_mode else SEAT
self._create_or_update_course_entitlement()
course.save()
```

(Just a nit comment, ignore if you like the current layout)

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2991,6 +2991,94 @@ def test_edit_page_with_invalid_revision_id(self):
...
+        self.assertRedirects(
+            response,
+            expected_url=reverse('publisher:publisher_course_detail', kwargs={'pk': self.course.id}),
+            status_code=302,
+            target_status_code=200
+        )
+
+        # Modify the Course to try and create CourseEntitlements differing from the CourseRun and Seat type and price
+        post_data['mode'] = Seat.PROFESSIONAL
+        post_data['price'] = 1  # just a number different than what is used in the SeatFactory constructor
```
mikix: Might want to test each separately (to confirm that EITHER will fail it).

course_discovery/apps/publisher/views.py
jlajoie: Ended up moving to it's own method so it's not repeated so frequently.

course_discovery/apps/publisher/tests/test_views.py
jlajoie: Great suggestion :+1: will duplicate and test

mikix:
It would be neat if we pre-filled the price for SEAT_VERSION courses that have the same price. But that's just a nit idea.

jlajoie:
Actually love that idea but might just table that for now. Though it might be a bit weird for a user to come to this empty page and see data filled out that they didn't set?

course_discovery/apps/publisher/views.py
```diff
@@ -362,13 +362,40 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
...
         return reverse(self.success_url, kwargs={'pk': self.object.id})
 
+    def get_context_data(self, **kwargs):
+        context = super(CourseEditView, self).get_context_data(**kwargs)
+        history_id = self.request.GET.get('history_id', None)
+        self.object = self.get_object()
+
+        try:
+            history_object = self.object.history.get(history_id=history_id) if history_id else None
+        except Exception:  # pylint: disable=broad-except
```
mikix, line 380: So beautiful

mikix:
There was talk of dropping the "audit/credit" option from the drop-down, since if the user selects it, we just yell at them. Was that too hacky to implement?

course_discovery/apps/publisher/views.py
```diff
@@ -393,33 +420,134 @@ def get_form_kwargs(self):
...
+        if self.object.version == Course.SEAT_VERSION:
+            if entitlement_mode:
+                type_misconfigurations, seat_misconfigurations = self._get_misconfigured_couse_runs(
+                    self.object, entitlement_price, entitlement_mode
+                )
+                if type_misconfigurations:
+                    # pylint: disable=no-member
+                    error_message = _(
+                        'The entered price does not match the price for the following course run(s): '
+                        '{course_runs}. The price that you enter must match the price of all active '
```
mikix, line 495: This ends up printing database id numbers. Is there something more helpful we could give them to go pick the right run to edit?

jlajoie:
I looked at doing it using the `self.fields` and overriding it inside of the __init_ but it felt pretty gross to do so for both the CourseEntitlementForm as well as the CourseEditView. Though maybe this might make more sense to revisit now that we have the changing drop down requirements (blank option, audit, credit, etc.) For now it just raises the error. Would this be something you'd be okay punting on and tackling as a refactor in that followup story?

mikix:
That's fine. The user can't get in a bad state the current way, just more UX wonkiness. We can clean up with other wonkiness.

course_discovery/apps/publisher/views.py
jlajoie: I could add in the course_run start date? I wonder if maybe adding these as a hyperlink might make more sense?

mikix: Mm, yeah maybe. On the course view page, on the right hand side, we show links like "Feb. 22, 2018 - Self-paced". Maybe a list of links like that? Especially if the code to format that name is easily re-usable.

course_discovery/apps/publisher/forms.py
```diff
@@ -505,13 +505,24 @@ def clean(self):
...
             self.add_error('price', _('Only audit seat can be without price.'))
 
         return cleaned_data
 
+    def is_valid(self):
+        mode = self.cleaned_data.get('mode')
+        price = self.cleaned_data.get('price')
+        has_mode = bool(mode)
+        has_price = price is not None and price > 0
+        return has_mode == has_price
```
mikix: Do you need this in addition to the clean() checks?

jlajoie: Nope not with the modifications made to clean, good catch!

mikix (approved):
Nice, looks good. :+1:

PR #1325, opened 2018-02-20
Send UUID to ecommerce

mikix:
When publishing, we want to give ecommerce the UUID of the course so that it doesn't have to turn around and ask us for that same bit of information.

Related to https://github.com/edx/ecommerce/pull/1658

### DIFF
--- a/course_discovery/apps/publisher/api/v1/views.py
+++ b/course_discovery/apps/publisher/api/v1/views.py
@@ -33,6 +33,10 @@ class CourseRunViewSet(viewsets.GenericViewSet):
 
     PUBLICATION_SUCCESS_STATUS = 'SUCCESS'
 
+    def get_course_key(self, publisher_course):
+        return '{org}+{number}'.format(org=publisher_course.organizations.first().key,
+                                       number=publisher_course.number)
+
     @detail_route(methods=['post'])
     def publish(self, request, pk=None):
         course_run = self.get_object()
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -76,9 +76,15 @@ def publish_to_studio(self, partner, course_run):
             return 'FAILED: ' + str(ex)
 
     def publish_to_ecommerce(self, partner, course_run):
+        publisher_course = course_run.course
+        course_key = '{org}+{number}'.format(org=publisher_course.organizations.first().key,
+                                             number=publisher_course.number)
```
amangano-edx: Isn't there a utility method defined on Publisher `Course` for this? If not, would it make sense to create one and use it here?

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -76,9 +76,15 @@ def publish_to_studio(self, partner, course_run):
...
+        discovery_course = Course.objects.get(partner=partner, key=course_key)
```
amangano-edx: I think there might also be a utility method for this as well defined on the Publisher `Course`.

amangano-edx: Maybe not since it requires the partner arg.

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -76,9 +76,15 @@ def publish_to_studio(self, partner, course_run):
             return 'FAILED: ' + str(ex)
 
     def publish_to_ecommerce(self, partner, course_run):
+        publisher_course = course_run.course
+        course_key = '{org}+{number}'.format(org=publisher_course.organizations.first().key,
```
amangano-edx: Can there be more than one org?

course_discovery/apps/publisher/api/v1/views.py
mikix: Yeah I was being lazy. This code was already in publish_to_discovery and I copied it for publish_to_ecommerce. I've moved it to a utility function and called it from both places.

course_discovery/apps/publisher/api/v1/views.py
mikix: I don't know about that. My only interest in this method is getting the exact same key that we use for publish_to_discovery. That doesn't take multiple orgs into account, so I didn't. Now that we use a utility function, we're guaranteed to use same key.

jlajoie (approved):
Lgtm :+1:

PR #1331, opened 2018-02-23
Fix Course Syllabus max length issue

awaisdar001:
In the frontend, TinyMC is removing `\n`, `&nbsp` which backend wasn't, causing a mismatch between the character counts. Now updated the backend to use `get_text(strip=True)` method which will remove these from the text.

### DIFF
--- a/course_discovery/apps/publisher/tests/test_validators.py
+++ b/course_discovery/apps/publisher/tests/test_validators.py
@@ -1,0 +1,26 @@
+import ddt
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+
+from course_discovery.apps.publisher.validators import validate_text_count
+
+
+@ddt.ddt
+class ValidatorTests(TestCase):
+    """
+    Tests for form validators
+    """
+    @ddt.data(
+        ('<strong>MODULE 0:&nbsp;</strong>', 'MODULE 0:'),
+        ('   <strong>MODULE 0: </strong>  ', 'MODULE 0:'),
+        ('\n\n<strong>MODULE 0: \n\n</strong>\n\n', 'MODULE 0:')
+    )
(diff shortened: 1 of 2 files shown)


awaisdar001:
@noraiz-anwar Can you please review it as @asadazam93  is on leave

course_discovery/apps/publisher/tests/test_validators.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+    @ddt.unpack
+    def test_validate_text_count(self, text_to_validate, expected_clean_text):
+        """Tests that validate text count work as expected"""
+        max_length_allowed = len(expected_clean_text)
+        validate_text_count(max_length_allowed)(text_to_validate)
+
+        # Verify you get a Validation Error if try go below the max.
+        with self.assertRaises(ValidationError):
+            validate_text_count(max_length_allowed - 1)(text_to_validate)
```
noraiz-anwar, line 26: Can we rather check the equality of `max_length_allowed ` and length of `text_to_validate `? This way we would be more sure that we got the exact string we wanted.

awaisdar001: L#21 says they are equal.
`max_length_allowed = len(expected_clean_text)`

noraiz-anwar: Just saw that you are already doing this at L#22.

awaisdar001: Yes.

attiyaIshaque (approved):
Once rebased. 👍

awaisdar001:
@asadazam93  can you review this pr

asadazam93 (approved):
Looks good to me

PR #1336, opened 2018-02-27
EDUCATOR-2231 remove course run enrollment dates references

Rabia23:
### Description
This PR removes all the reference to the enrollment dates columns.

Related edx-platform PR: https://github.com/edx/edx-platform/pull/17559

### Reviewers
- [ ]  @schenedx 
- [ ]  @awaisdar001 

### Post-review


course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -148,8 +148,6 @@ def test_publish(self, mock_access_token):  # pylint: disable=unused-argument,to
         assert discovery_course_run.full_description_override is None
         assert discovery_course_run.start == publisher_course_run.start
         assert discovery_course_run.end == publisher_course_run.end
-        assert discovery_course_run.enrollment_start == publisher_course_run.enrollment_start
```
awaisdar001, line 148: Test to see if `discovery_course_run.enrollment_start == None`

awaisdar001 (changes requested):

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -148,8 +148,8 @@ def test_publish(self, mock_access_token):  # pylint: disable=unused-argument,to
...
-        assert discovery_course_run.enrollment_end == publisher_course_run.enrollment_end
+        assert discovery_course_run.enrollment_start is None
+        assert discovery_course_run.enrollment_end is None
         assert discovery_course_run.pacing_type == publisher_course_run.pacing_type
         assert discovery_course_run.min_effort == publisher_course_run.min_effort
         assert discovery_course_run.max_effort == publisher_course_run.max_effort
```
awaisdar001, line 152: Just curious to know, can we use `self.assertEqual` here?

awaisdar001 (commented):
👍

Rabia23:
@schenedx Can you please have a look at this one too?

schenedx (commented):
I believe you can also remove the reference of enrollment_start and end from the model.py, just don't check in the migration file generated.
Additionally, why update all the assert styles in tests? What is the value of making those changes?

Rabia23:
@schenedx just curious to know, what is the hesitation of not generating migration file after removing the reference of enrollment_start and end from the model.py file.

Rabia23:
`assert` is pure python related thing and `self. assertEqual()` comes from `unittest.TestCase.assertEqual()` so i think it's good idea to use `assertEqual` instead of `assert` in unit tests.

schenedx:
@Rabia23 
Do you know how the code deploys to production? Do you know that data base table change will be first introduced into the production system, then after a few minutes, the actual server code will change? If we introduce migration within the same PR, the database will have breaking change deployed before the servers code will be updated. Within that few minutes a lot of things will break. If you still have questions, I strongly recommend you talk to someone like @waheedahmed to understand more.
As for the assert changes, there are school of thought that python asserts are better choice than the unit test implementation of the asserts. If the changes from one thing to the other do not have any value, why make them? I would be against changing code because you thought one way is better than the other way while both ways works.

Rabia23:
@schenedx Thank you for taking deployment process into my account . I didn't know this before. I messed up this PR so i have opened new PR https://github.com/edx/course-discovery/pull/1350 in which I have removed the reference of `enrollment_start` and `enrollment_end` from the model.py file and also revert back `assert` changes. 
Please have a look at that PR.
Thanks.

PR #1339, opened 2018-02-28
LEARNER-4337: Fixes check for misconfigured courses

jlajoie:
# Description
[LEARNER-4337](https://openedx.atlassian.net/browse/LEARNER-4337)

Adds a check to make sure that course runs' `audit` seats are not compared for validating a conversion of a Course with Course Run pricing to a Course with Entitlements.

Notes:

* This method will likely need to change when tackling https://openedx.atlassian.net/browse/LEARNER-4280 as those courses will always have a mismatched type and potentially price.

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -426,21 +426,36 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
 
         return render(request, self.template_name, context)
 
-    def _get_misconfigured_couse_runs(self, course, price, mode):
+    def _get_misconfigured_course_runs(self, course, price, mode):
         misconfigured_seat_type_runs = set()
         misconfigured_price_runs = set()
         for course_run in course.publisher_course_runs.filter(end__gt=datetime.now()):
-            for seat in course_run.seats.all():
-                if seat.type != mode:
-                    misconfigured_seat_type_runs.add('{type} - {start}'.format(
-                        type=seat.course_run.get_pacing_type_display(),
-                        start=seat.course_run.start.strftime("%B %d, %Y")
-                    ))
-                if seat.price != price:
-                    misconfigured_price_runs.add('{type} - {start}'.format(
-                        type=seat.course_run.get_pacing_type_display(),
-                        start=seat.course_run.start.strftime("%B %d, %Y")
-                    ))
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -426,21 +426,39 @@ def get(self, request, *args, **kwargs):
...
+                    # There should be exactly one matching seat with the same type. Price is irrelevant.
+                    type_is_valid = len(seats) == 1 and seats[0].type == Seat.AUDIT
+                else:
+                    # There should be exactly one matching seat with the same type/price
+                    type_is_valid = len(seats) == 1 and seats[0].type == mode
+                    price_is_valid = seats[0].price == price
+
+            if not type_is_valid:
+                misconfigured_seat_type_runs.add('{type} - {start}'.format(
+                    type=course_run.get_pacing_type_display(),
```
amangano-edx, line 451: Should we include anything else to make it easier to identify the misconfigured runs? Maybe course_run.key instead of pacing_type/start_date?

amangano-edx (approved):

amangano-edx:
Sanity check: We don't need this to work for credit right? Since credit isn't an option yet at the course-level.

jlajoie:
Yeah does not need to work for credit as of now, when separating Credit out to be handled as it's own option this will, but until then this will suffice

course_discovery/apps/publisher/views.py
jlajoie: I stole this from the breadcrumbs for courses so I figured it would suffice. I originally tried to just hyperlink the Course Run but it ended up being a much larger task than simply embedding HTML. In `alert_messages.html` we could have set an `extra_tag` that the text we're passing it was `safe` and then use the `| safe` filter... but our styling is based off of extra tags which seemed like a much bigger endeavor than just using a descriptor. Breadcrumbs code below.

```
        context['breadcrumbs'] = make_bread_crumbs(
            [
                (reverse('publisher:publisher_courses'), 'Courses'),
                (reverse('publisher:publisher_course_detail', kwargs={'pk': course.id}), course.title),
                (None, '{type}: {start}'.format(
                    type=course_run.get_pacing_type_display(), start=start_date
                ))
            ]
        )
```

PR #1340, opened 2018-02-28
LEARNER-4282: QOL changes for Course Edit Form

jlajoie:
# Description
[LEARNER-4282](https://openedx.atlassian.net/browse/LEARNER-4282)

~~Adds the blank option to the dropdown by overriding the constructor.~~ (Changes were merged as part of https://github.com/edx/course-discovery/pull/1346) 

Adds a clarity warning to the Course Edit Form for type and price.

This PR is dependent on https://github.com/edx/course-discovery/pull/1346 merging first


course_discovery/apps/publisher/forms.py
```diff
@@ -485,6 +484,14 @@ class CourseEntitlementForm(BaseForm):
                              initial=CourseEntitlement.VERIFIED)
     price = forms.DecimalField(max_digits=6, decimal_places=2, required=False, initial=0.00)
 
+    def __init__(self, *args, **kwargs):
+        show_blank = kwargs.pop('show_blank', False)
+        super(CourseEntitlementForm, self).__init__(*args, **kwargs)
+        if show_blank:
+            self.fields['mode'] = forms.ChoiceField(choices= [('','')] + self.MODE_CHOICES, required=False,
+                                                    label=_('Enrollment Track'), initial=CourseEntitlement.VERIFIED)
```
amangano-edx: Do you need to instantiate an entirely new field here, or would it be possible to modify just the choices by doing something like this? https://stackoverflow.com/questions/5028731/how-to-set-choices-in-dynamic-with-django-choicefield

course_discovery/apps/publisher/views.py
```diff
@@ -390,9 +390,10 @@ def get_context_data(self, **kwargs):
 
         if waffle.switch_is_active(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH):
             if self.object.version == Course.ENTITLEMENT_VERSION:
```
amangano-edx: Nit: I added a convenience method to Course to check this. See https://github.com/edx/course-discovery/blob/8108ece481b7e09bbb252809afaeaf154dcb3670/course_discovery/apps/publisher/models.py#L105

course_discovery/apps/publisher/forms.py
```diff
@@ -476,7 +476,6 @@ def validate_and_remove_seats(self, course_run):
 
 class CourseEntitlementForm(BaseForm):
     MODE_CHOICES = [
-        ('', _('Audit only or credit')),
```
amangano-edx: This will change the `CourseCreateView` as well. If you merge this before https://openedx.atlassian.net/browse/LEARNER-4349, `CourseCreateView` may end up in a state where 'VERIFIED' and 'PROFESSIONAL' are the only two choices, which may confuse partners even more.

amangano-edx: It would also effectively prevent credit/audit-only Course creation.

course_discovery/apps/publisher/views.py
```diff
@@ -390,9 +390,10 @@ def get_context_data(self, **kwargs):
...
-                context['entitlement_form'] = self.entitlement_form(instance=self.object.entitlements.first())
+                context['entitlement_form'] = self.entitlement_form(show_blank=True,
+                                                                    instance=self.object.entitlements.first())
             else:
-                context['entitlement_form'] = self.entitlement_form({'mode': ''})
+                context['entitlement_form'] = self.entitlement_form({'mode': ''}, show_blank=True)
```
amangano-edx: I'm still new to the forms API, but is there a reason why we don't need to qualify `{'mode': ''}` by passing it via the `initial` kwargs?

course_discovery/apps/publisher/forms.py
jlajoie: Definitely can do something like that, thanks for the search!

course_discovery/apps/publisher/forms.py
jlajoie: This was originally implemented this way because of the previous changes we were going to do with Audit and Credit. But agreed, we should merge this in after the changes in 4349 go live now instead of the feature branch merge we were going to do previously.

mikix (approved):

PR #1341, opened 2018-03-02
Clee/travis master check

BenjiLee:


BenjiLee:
Note to self: Remove dependency from python requests and use urllib

PR #1343, opened 2018-03-05
Drop publisher_entitlements waffle flag

mikix:
The feature has fully landed. We just want it always on now.

https://openedx.atlassian.net/browse/LEARNER-4152

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -27,7 +27,6 @@ from course_discovery.apps.course_metadata.models import Person
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
 from course_discovery.apps.publisher import emails, mixins, serializers
 from course_discovery.apps.publisher.choices import CourseRunStateChoices, CourseStateChoices, PublisherUserRole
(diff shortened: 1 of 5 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -388,11 +382,10 @@ def get_context_data(self, **kwargs):
             }
         )
 
-        if waffle.switch_is_active(PUBLISHER_ENTITLEMENTS_WAFFLE_SWITCH):
-            if self.object.version == Course.ENTITLEMENT_VERSION:
-                context['entitlement_form'] = self.entitlement_form(instance=self.object.entitlements.first())
-            else:
-                context['entitlement_form'] = self.entitlement_form({'mode': ''})
+        if self.object.version == Course.ENTITLEMENT_VERSION:
```
amangano-edx: Nit: I added a helper to Course for testing this. See `Course.uses_entitlements`.

amangano-edx (approved):

course_discovery/apps/publisher/views.py
mikix: Alright, sure. Will use here too.

PR #1344, opened 2018-03-05
Catalog_query contains endpoint

McKenzieW:
Add an endpoint in discovery that can determine whether a given set of courses and/or course runs exist in the subset of the course catalog described by a given query.

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Discovery subtask](https://openedx.atlassian.net/browse/LEARNER-4360)

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
@@ -1,0 +1,93 @@
+import urllib
+
+from rest_framework.reverse import reverse
+
+from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase
+from course_discovery.apps.core.tests.factories import UserFactory
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseRunFactory
+
+
+class CatalogQueryViewSetTests(APITestCase):
+    def setUp(self):
+        super(CatalogQueryViewSetTests, self).setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.client.force_authenticate(self.user)
+        self.course_run = CourseRunFactory(course__partner=self.partner)
+        self.course = CourseFactory(partner=self.partner, key='simple_key')
+        self.url_base = reverse('api:v1:catalog-query_contains')
+        self.error_message = 'CatalogQueryContains endpoint requires query and identifiers list(s)'
+
+    def test_contains_single_course_run(self):
+        """ Verify that a single course_run is contained in a query. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:' + self.course_run.key,
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = '{}/?{}'.format(self.url_base, qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.data,
+            {
+                self.course_run.key: True,
+                str(self.course.uuid): False
+            }
+        )
+
+    def test_contains_single_course(self):
+        """ Verify that a single course is contained in a query. """
+        qs = urllib.parse.urlencode({
+            'query': 'key:' + self.course.key,
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = '{}/?{}'.format(self.url_base, qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.data,
+            {
+                self.course_run.key: False,
+                str(self.course.uuid): True
+            }
+        )
+
+    def test_contains_course_and_run(self):
+        """ Verify that both the course and the run are contained in the broadest query. """
+        qs = urllib.parse.urlencode({
+            'query': 'key:*',
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = '{}/?{}'.format(self.url_base, qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.data,
+            {
+                self.course_run.key: True,
+                str(self.course.uuid): True
+            }
+        )
+
+    def test_no_identifiers(self):
+        """ Verify that a 400 status is returned if request does not contain any identifier lists. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:*'
+        })
+        url = '{}/?{}'.format(self.url_base, qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, self.error_message)
+
+    def test_no_query(self):
+        """ Verify that a 400 status is returned if request does not contain a querystring. """
+        qs = urllib.parse.urlencode({
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = '{}/?{}'.format(self.url_base, qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, self.error_message)
--- a/course_discovery/apps/api/v1/urls.py
+++ b/course_discovery/apps/api/v1/urls.py
@@ -4,6 +4,7 @@ from rest_framework import routers
 
 from course_discovery.apps.api.v1.views import search as search_views
 from course_discovery.apps.api.v1.views.affiliates import AffiliateWindowViewSet
+from course_discovery.apps.api.v1.views.catalog_queries import CatalogQueryContainsViewSet
 from course_discovery.apps.api.v1.views.catalogs import CatalogViewSet
 from course_discovery.apps.api.v1.views.course_runs import CourseRunViewSet
 from course_discovery.apps.api.v1.views.courses import CourseViewSet
@@ -21,7 +22,8 @@ partners_router.register(r'affiliate_window/catalogs', AffiliateWindowViewSet, b
 urlpatterns = [
     url(r'^partners/', include(partners_router.urls, namespace='partners')),
     url(r'search/typeahead', search_views.TypeaheadSearchView.as_view(), name='search-typeahead'),
-    url(r'currency', CurrencyView.as_view(), name='currency')
+    url(r'currency', CurrencyView.as_view(), name='currency'),
+    url(r'^catalog/query-contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog-query_contains')
 ]
 
 router = routers.SimpleRouter()
--- a/course_discovery/apps/api/v1/views/catalog_queries.py
+++ b/course_discovery/apps/api/v1/views/catalog_queries.py
@@ -1,0 +1,47 @@
+from uuid import UUID
+
+from rest_framework import status
+from rest_framework.generics import GenericAPIView
+from rest_framework.permissions import DjangoModelPermissions, IsAuthenticated
+from rest_framework.response import Response
+
+from course_discovery.apps.course_metadata.models import Course, CourseRun
+
+
+class CatalogQueryContainsViewSet(GenericAPIView):
+    permission_classes = (IsAuthenticated, DjangoModelPermissions)
+    queryset = Course.objects.all()
+
+    def get(self, request):
+        """
+        Determine if a set of courses and/or course runs is found in the query results.
+
+        Returns
+            dict:  mapping of course and run indentifiers included in the request to boolean values
+                indicating whether or not the associated course or run is contained in the queryset
+                described by the query found in the request.
+        """
+        query = request.GET.get('query')
+        course_run_ids = request.GET.get('course_run_ids', None)
+        course_uuids = request.GET.get('course_uuids', None)
+        partner = self.request.site.partner
+
+        if query and (course_run_ids or course_uuids):
+            identified_course_ids = set()
+            specified_course_ids = []
+            if course_run_ids:
+                course_run_ids = course_run_ids.split(',')
+                specified_course_ids = course_run_ids
+                identified_course_ids.update(CourseRun.search(query).filter(
+                    partner=partner.short_code, key__in=course_run_ids).values_list('key', flat=True))
+            if course_uuids:
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                specified_course_ids += course_uuids
+                identified_course_ids.update(Course.search(query).filter(partner=partner, uuid__in=course_uuids).
+                                             values_list('uuid', flat=True))
+
+            contains = {str(identifier): identifier in identified_course_ids for identifier in specified_course_ids}
+            return Response(contains)
+        return Response(
+            'CatalogQueryContains endpoint requires query and identifiers list(s)', status=status.HTTP_400_BAD_REQUEST
+        )
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -118,6 +118,7 @@ class BaseCourseIndex(OrganizationsMixin, BaseIndex):
 class CourseIndex(BaseCourseIndex, indexes.Indexable):
     model = Course
 
+    uuid = indexes.CharField(model_attr='uuid')
     course_runs = indexes.MultiValueField()
     expected_learning_items = indexes.MultiValueField()
 


McKenzieW:
@edx/helio take a look

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+        described by the query found in the request.
+        ---
+        serializer: serializers.ContainedCoursesAndRunsSerializer
+        """
+        query = request.GET.get('query')
+        course_run_ids = request.GET.get('course_run_ids', None)
+        course_uuids = request.GET.get('course_uuids', None)
+        partner = self.request.site.partner
+
+        if query and course_run_ids or course_uuids:
```
Mjloturco: Python [order of precedence](http://www.mathcs.emory.edu/~valerie/courses/fall10/155/resources/op_precedence.html) with do this as 
``` python
if (query and course_run_ids) or course_uuids:
```
not sure thats what you want here.

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+            ids_in_queryset = []
+            ids_in_question = []
+            if course_run_ids:
+                course_run_ids = course_run_ids.split(',')
+                ids_in_question = course_run_ids
+                ids_in_queryset = list(CourseRun.search(query).filter(partner=partner.short_code).
```
Mjloturco: could the chained filters be condensed? 

``` python 
list(CourseRun.search(query).filter(partner=partner.short_code, key__in=course_run_ids).values_list('key', flat=True))
```

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+                                       filter(key__in=course_run_ids).values_list('key', flat=True))
+            if course_uuids:
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                ids_in_question += course_uuids
+                ids_in_queryset += list(Course.search(query).filter(partner=partner).filter(uuid__in=course_uuids).
```
Mjloturco: ditto on filter merge.

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Determine if a set of courses and/or course runs is found in the query results.
+
+        Returns a dict mapping the course and run indentifiers included in the request to boolean values
+        indicating whether or not the associated course or run is contained in the queryset
+        described by the query found in the request.
+        ---
+        serializer: serializers.ContainedCoursesAndRunsSerializer
+        """
+        query = request.GET.get('query')
```
Mjloturco: Seems like we may want to do some optimization around this query, could we somehow detect if the query is 'All' and just return an all valid response? Would cut down on a lot of DB call if we have coupons that are valid for all runs/courses.

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -118,6 +118,7 @@ def prepare_level_type(self, obj):
 class CourseIndex(BaseCourseIndex, indexes.Indexable):
     model = Course
 
+    uuid = indexes.CharField(model_attr='uuid')
```
Mjloturco, line 121: Does adding things to the index have any weird impacts on Search? Like does this expose the uuid in places it shouldn't be? (I suspect no but might be worth checking)

Mjloturco (changes requested):
Mostly looks good! I'd recommend we get someone with a good eye for performance to take a look at this, since this will get hit for most ecommerce basketCalculateView calls and for a bunch of BasketAdd calls I believe as well.

course_discovery/apps/api/v1/views/catalog_queries.py
McKenzieW: From the discussion I had with Michael he didn't seem particularly interested in adding query-aware functionality to this code, largely because it could get really mucky (unfortunately there are a lot of different variations that could essentially "all" either of courses or course runs or both) one thing I do know though is that when elasticsearch gets the `*:*` query it does just return all without performing any sort of search functionality.

course_discovery/apps/course_metadata/search_indexes.py
McKenzieW: I'm not sure, I will see what I can find about that.

Mjloturco (approved):

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+
+class CatalogQueryContainsViewSet(GenericAPIView):
+    permission_classes = (IsAuthenticated, DjangoModelPermissions)
+    queryset = Course.objects.all()
+
+    def get(self, request):
+        """
+        Determine if a set of courses and/or course runs is found in the query results.
+
+        Returns a dict mapping the course and run indentifiers included in the request to boolean values
```
staubina: Small code style things this should match the Docstrings code style defined here:

http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings

```
Returns
    dict:  mapping the course and run indentifiers included in the request to boolean values
        indicating whether or not the associated course or run is contained in the queryset
        described by the query found in the request.
```

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+        indicating whether or not the associated course or run is contained in the queryset
+        described by the query found in the request.
+        ---
+        serializer: serializers.ContainedCoursesAndRunsSerializer
```
staubina: Is this another Returned object?

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,49 @@
...
+                    partner=partner.short_code, key__in=course_run_ids).values_list('key', flat=True))
+            if course_uuids:
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                ids_in_question += course_uuids
+                ids_in_queryset += list(Course.search(query).filter(partner=partner, uuid__in=course_uuids).
+                                        values_list('uuid', flat=True))
+
+            contains = {str(indentifier): indentifier in ids_in_queryset for indentifier in ids_in_question}
+            return Response(contains)
+        return Response(status=status.HTTP_400_BAD_REQUEST)
```
staubina: It might be of value to say why this was a bad request or log the reason.

course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
```diff
@@ -0,0 +1,89 @@
...
+        """ Verify that a single course_run is contained in a query. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:' + self.course_run.key,
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = '{}/?{}'.format(reverse('api:v1:catalog_query-contains'), qs)
+        response = self.client.get(url)
+        assert response.status_code == 200
+        self.assertEqual(
```
staubina: We should stick to one form of assertion in our tests.  self.assertEqual is from the unittest package while assert is the pytest method.  I believe we are moving toward using pytest more now.  I would recommend using one or the other. I am currently using assert myself.

staubina: I am verifying this with TestEng.

staubina (changes requested):
Please let me know if you have any questions about my comments.

course_discovery/apps/api/v1/views/catalog_queries.py
McKenzieW: No it at one point was a serializer I was going to use on the dataset to construct the return object but I switched approaches and just forgot to remove it from the docstring.

staubina (approved):
Overall looks good to me.  I would take some time to take what Nimisha mentioned into consideration.

course_discovery/apps/api/v1/urls.py
```diff
@@ -21,7 +22,8 @@
 urlpatterns = [
     url(r'^partners/', include(partners_router.urls, namespace='partners')),
     url(r'search/typeahead', search_views.TypeaheadSearchView.as_view(), name='search-typeahead'),
-    url(r'currency', CurrencyView.as_view(), name='currency')
+    url(r'currency', CurrencyView.as_view(), name='currency'),
+    url(r'^catalog_query/contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog_query-contains')
```
nasthagiri: When designing APIs, a top-most consideration is consistency with other endpoints - from the perspective of the API consumer.

For example, compare the difference in usage and behavior between
`/course_runs/contains` and `/catalogs/contains`
where the former takes a query string while the latter doesn't. 

And compare the difference in naming between
`/course_runs/contains` and the proposed `/catalog_query/contains`
where their usage and behavior are consistent, but the naming isn't.

I believe to be clear to the consumer, you probably want all 4 of the permutations:

- /course_runs/contains/
- /catalogs/contains/
- /course_runs/query_contains/
- /catalogs/query_contains/

Also, as mentioned in our discussion, these "contains" endpoints are currently not RESTful.  But that opens up a larger conversation.

course_discovery/apps/course_metadata/search_indexes.py
McKenzieW: I spoke to Doug Hall and he confirmed that this change would not likely have any performance implications.

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,50 @@
...
+                described by the query found in the request.
+        """
+        query = request.GET.get('query')
+        course_run_ids = request.GET.get('course_run_ids', None)
+        course_uuids = request.GET.get('course_uuids', None)
+        partner = self.request.site.partner
+
+        if query and (course_run_ids or course_uuids):
+            ids_in_queryset = []
+            ids_in_question = []
```
jaebradley: Nit: `ids_in_queryset` and `ids_in_question` make sense - but they seem implementation-focused?

`ids_in_queryset` seems to represent all the specified course identifiers and `ids_in_question` seems to represent all the identifiable course identifiers (lol say that five times fast)?

So maybe `specified_course_ids` and `identified_course_ids` might provide more context?

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,50 @@
...
+            if course_run_ids:
+                course_run_ids = course_run_ids.split(',')
```
jaebradley: I think it's fine here, but in general, variable reassignment can sometimes lead to unexpected behavior.

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,50 @@
...
+                ids_in_question = course_run_ids
+                ids_in_queryset = list(CourseRun.search(query).filter(
+                    partner=partner.short_code, key__in=course_run_ids).values_list('key', flat=True))
+            if course_uuids:
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                ids_in_question += course_uuids
+                ids_in_queryset += list(Course.search(query).filter(partner=partner, uuid__in=course_uuids).
```
jaebradley: It's `partner=partner` here and `partner=partner.short_code` above - does it matter?

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,50 @@
...
+                                        values_list('uuid', flat=True))
+
+            contains = {str(indentifier): indentifier in ids_in_queryset for indentifier in ids_in_question}
```
jaebradley: Nit: typo - `indentifier` vs. `identifier`?

jaebradley: Probably doesn't make that big of a performance difference overall but [`set`s are probably a lot faster than `list`s for existence checking](https://stackoverflow.com/questions/2831212/python-sets-vs-lists).

Maybe consider making `ids_in_queryset` a `set`?

course_discovery/apps/api/v1/views/catalog_queries.py
Mjloturco: Courses vs CourseRuns probably store the partner relationship differently.

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,50 @@
...
+            if course_uuids:
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                specified_course_ids += course_uuids
+                identified_course_ids.union(set(Course.search(query).filter(partner=partner, uuid__in=course_uuids).
+                                                values_list('uuid', flat=True)))
+
+            contains = {str(identifier): identifier in identified_course_ids for identifier in specified_course_ids}
+            return Response(contains)
+        return Response(
+            'CatalogQueryContains endpoint requires query and indentifiers list(s)', status=status.HTTP_400_BAD_REQUEST
```
jaebradley: Nit: `indentifiers` vs. `identifiers`

course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
```diff
@@ -0,0 +1,91 @@
...
+        self.course = CourseFactory(partner=self.partner)
+
+    def test_contains_single_course_run(self):
+        """ Verify that a single course_run is contained in a query. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:' + self.course_run.key,
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = '{}/?{}'.format(reverse('api:v1:catalog-query_contains'), qs)
```
jaebradley: if you're worried about repeating yourself you could just move this to a test variable?

course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_no_identifiers(self):
+        """ Verify that a 400 status is returned if request does not contain any identifier lists. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:*'
+        })
+        url = '{}/?{}'.format(reverse('api:v1:catalog-query_contains'), qs)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, 'CatalogQueryContains endpoint requires query and indentifiers list(s)')
```
jaebradley: This `str` message seems duplicated - you could consider moving this to a test variable?

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,51 @@
...
+        course_uuids = request.GET.get('course_uuids', None)
+        partner = self.request.site.partner
+
+        if query and (course_run_ids or course_uuids):
+            identified_course_ids = set()
+            specified_course_ids = []
+            if course_run_ids:
+                course_run_ids = course_run_ids.split(',')
+                specified_course_ids = course_run_ids
+                identified_course_ids = set(CourseRun.search(query).filter(
```
jaebradley: Does `values_list` return an `iterable`?

You can append an `iterable` to a `set` by calling `update`.

image

If so, I think you can also avoid doing the `union` below and do `update` instead so you're not creating a whole new `set` just to reassign it and just update the existing one.

McKenzieW: oh yay! it does return an iterable. Thank you!

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -0,0 +1,50 @@
...
+from uuid import UUID
+
+from rest_framework import status
+from rest_framework.generics import GenericAPIView
+from rest_framework.permissions import DjangoModelPermissions, IsAuthenticated
+from rest_framework.response import Response
+
+from course_discovery.apps.course_metadata.models import Course, CourseRun
+
+logger = logging.getLogger(__name__)
```
Mjloturco: Nit: logging looks unused

Mjloturco (approved):
Just clean up that extra logger and then this looks good to go.

staubina (approved):
:shipit:

PR #1345, opened 2018-03-05
Update seat when changing entitlement

mikix:
We weren't propagating entitlement type/price changes to existing seat products.

https://openedx.atlassian.net/browse/LEARNER-4347

### DIFF
--- a/course_discovery/apps/publisher/tests/test_views.py
+++ b/course_discovery/apps/publisher/tests/test_views.py
@@ -3060,7 +3060,7 @@ class CourseEditViewTests(SiteMixin, TestCase):
         self.course.save()
         post_data = self._post_data(self.organization_extension)
         post_data['team_admin'] = self.course_team_role.user.id
-        post_data['mode'] = Seat.VERIFIED
+        post_data['mode'] = CourseEntitlement.VERIFIED
         post_data['price'] = 150
 
         response = self.client.post(self.edit_page_url, data=post_data)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -3060,7 +3060,7 @@ def test_entitlement_version_course_edit_page(self):
         self.course.save()
         post_data = self._post_data(self.organization_extension)
         post_data['team_admin'] = self.course_team_role.user.id
-        post_data['mode'] = Seat.VERIFIED
+        post_data['mode'] = CourseEntitlement.VERIFIED
```
jlajoie, line 3063: Thanks for changing these!

course_discovery/apps/publisher/views.py
```diff
@@ -419,10 +419,13 @@ def get(self, request, *args, **kwargs):
...
         return render(request, self.template_name, context)
 
+    def _get_future_course_runs(self, course):
+        return course.course_runs.filter(end__gt=datetime.now())
+
     def _get_misconfigured_course_runs(self, course, price, mode):
         misconfigured_seat_type_runs = set()
         misconfigured_price_runs = set()
-        for course_run in course.publisher_course_runs.filter(end__gt=datetime.now()):
+        for course_run in self._get_future_course_runs(course):
```
jlajoie: nit: Might be better to make this name a bit more verbose since it includes runs that are happening now as well, _get_present_and_future_course_runs() or _get_ongoing_course_runs()? or something?

mikix: Sure, good point.

mikix: Done.

jlajoie (approved):
Looks good, big fan of the refactor on the _handle method :+1:

PR #1346, opened 2018-03-05
Split 'audit' and 'credit' into two options on CourseEntitlementForm

amangano-edx:
Splits the 'Audit only or Credit' option in the CourseEntitlementForm into two distinct options. Selecting either of these options will have the same effect as selecting the combined option currently does -> we will not create an entitlement at the Course level.

JIRA https://openedx.atlassian.net/browse/LEARNER-4349


course_discovery/apps/publisher/forms.py
```diff
@@ -475,20 +475,39 @@ def validate_and_remove_seats(self, course_run):
 
 
 class CourseEntitlementForm(BaseForm):
+    AUDIT_MODE = 'audit'
+    VERIFIED_MODE = CourseEntitlement.VERIFIED
+    PROFESSIONAL_MODE = CourseEntitlement.PROFESSIONAL
+    CREDIT_MODE = 'credit'
+
+    # Modes for which we should not create entitlements.
+    NOOP_MODES = [AUDIT_MODE, CREDIT_MODE]
```
jlajoie, line 484: <3 the use of paid/noop here

course_discovery/apps/publisher/views.py
```diff
@@ -282,7 +282,9 @@ def post(self, request, *args, **kwargs):
             try:
                 with transaction.atomic():
                     course = course_form.save(commit=False)
-                    if entitlement_form['mode'].value():
+                    # Check for mode in cleaned_data, as mode will be set to None during validation for modes that
+                    # do not support entitlements.
+                    if entitlement_form.cleaned_data.get('mode', None):
```
jlajoie, line 287: tiny nit: `.get()` will return None without the explicit `None` if the key doesn't exist

course_discovery/static/js/publisher/entitlement-type-change.js
```diff
@@ -0,0 +1,9 @@
+$(document).on('change', '#id_mode', function (e) {
+    var $seatBlock = $("#seatPriceBlock"),
```
jlajoie: Since we're creating a new file specifically for entitlements, should we change the variables names where instead of being seatType they would be entitlementMode?

jlajoie (approved):
Really minor change requests, feel free to do them or not otherwise lgtm :+1:

course_discovery/static/js/publisher/entitlement-type-change.js
amangano-edx: Changed.

PR #1347, opened 2018-03-06
Updated travis validate translations

Ayub-Khan:
Create a validate_translation script and started
using the script for validation to update travis
check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ install:
 
 script:
     - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make docs'
+    - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make check_translations_up_to_date'
     - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make validate_translations'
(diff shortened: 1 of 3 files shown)


amangano-edx (changes requested):
I don't think you need a custom validate script. As of https://github.com/edx/i18n-tools/tree/0.4.0, the i18n_tool validate command will return a non-zero exit code if any translations fail validation, which will break the jenkins/travis builds. Therefore, you should be able to just use i18n_tool validate directly,.

.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
 
 script:
     - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make docs'
-    - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make validate_translations'
+    - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/course_discovery && i18n_tool validate'
```
amangano-edx: Alternatively, you could add a make target that changes the directory to `course_disovery` before running the `i18n_tool validate`, much like the `make fake_translations` command does. This would make it easier to run the validator locally.

amangano-edx:
Please validate that this works by creating a test branch, committing a bad translation to it, and confirming that the build fails.

amangano-edx:
Also, please confirm that the correct version of edx-i18n-tools is installed (>= 0.4.0) in the travis environment.

hamzamunir7300 (commented):
:+1

hamzamunir7300 (approved):
:+1

PR #1351, opened 2018-03-08
match url to what ecommerce is hitting

McKenzieW:
I was off by one character in my matching of the ecommerce API call to the new API endpoint I created.

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Discovery](https://openedx.atlassian.net/browse/LEARNER-4360)

### DIFF
--- a/course_discovery/apps/api/v1/urls.py
+++ b/course_discovery/apps/api/v1/urls.py
@@ -23,7 +23,7 @@ urlpatterns = [
     url(r'^partners/', include(partners_router.urls, namespace='partners')),
     url(r'search/typeahead', search_views.TypeaheadSearchView.as_view(), name='search-typeahead'),
     url(r'currency', CurrencyView.as_view(), name='currency'),
-    url(r'^catalog/query-contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog-query_contains')
+    url(r'^catalog/query_contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog-query_contains')
 ]
 
 router = routers.SimpleRouter()


staubina (approved):

course_discovery/apps/api/v1/urls.py
```diff
@@ -23,7 +23,7 @@
     url(r'^partners/', include(partners_router.urls, namespace='partners')),
     url(r'search/typeahead', search_views.TypeaheadSearchView.as_view(), name='search-typeahead'),
     url(r'currency', CurrencyView.as_view(), name='currency'),
-    url(r'^catalog/query-contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog-query_contains')
+    url(r'^catalog/query_contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog-query_contains')
```
Mjloturco, line 26: I know this is just a bugfix, but it seems weird that the value for the name param has both a dash and an underscore in it

McKenzieW: The dash in the name indicates a slash in the url, whereas the underscore serves as a space between the two words query and contains. The same sort of thing happens [here](https://github.com/edx/course-discovery/pull/1351/files/7e3334a121e1d0c4395fe713b879d3d804c26f7e#diff-23645e2410ba62300b82cf75f7673356L41).

McKenzieW: @Mjloturco does that assuage your concern at all?

Mjloturco: Yup, my concerns are assuaged.

PR #1356, opened 2018-03-16
LEARNER-4477: Fixes enrollment-track cascading changes at course level

jlajoie:
# Description
[LEARNER-4477](https://openedx.atlassian.net/browse/LEARNER-4477)

Adds logic to handle the cases of swapping the Course Entitlement Mode when there are published runs, as well as properly cascades that swap for when there are only unpublished active runs to the Course Run Seats.


course_discovery/apps/publisher/views.py
```diff
@@ -548,14 +571,29 @@ def _handle_entitlement_update(self, user, request, course_form):
...
+                    'entitlement_form': entitlement_form
+                })
+            published_runs = self._get_published_course_runs(self.object)
+            if published_runs:
+                # pylint: disable=no-member
+                error_message = _(
+                    'The following active course run(s) are published: {course_runs}. You cannot change the mode '
+                    'if there are published active runs.'
+                ).format(course_runs=', '.join(
+                    str(course_run_start) for course_run_start in published_runs
```
amangano-edx, line 581: Alternatively, it might be more useful to log the course run key rather than the start dates.

course_discovery/apps/publisher/views.py
```diff
@@ -427,6 +427,16 @@ def get(self, request, *args, **kwargs):
     def _get_active_course_runs(self, course):
         return course.course_runs.filter(end__gt=datetime.now())
 
+    def _get_published_course_runs(self, course):
+        published_runs = set()
+        for course_run in course.course_runs:
```
amangano-edx: Do you need to filter out the inactive runs?

course_discovery/apps/publisher/views.py
```diff
@@ -427,6 +427,16 @@ def get(self, request, *args, **kwargs):
...
+            if course_run.course_run_state.is_published:
+                published_runs.add('{type} - {start}'.format(
+                    type=course_run.get_pacing_type_display(),
+                    start=course_run.start.strftime("%B %d, %Y")
```
amangano-edx: Why pacing_type/start? It seems like it would be a lot easier to identify runs by their key, than it would be using these two fields.

course_discovery/apps/publisher/views.py
```diff
@@ -472,13 +483,25 @@ def _create_or_update_course_entitlement(self, course, entitlement_form):
...
-                # First make sure this seat is on the "entitlement side of things" (i.e. not an audit seat or similar)
-                if seat.type in CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING.values():
-                    seat.type = CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING[entitlement.mode]
-                    seat.price = entitlement.price
-                    seat.currency = entitlement.currency
-                    seat.save()
+            upgrade_deadline = None
+            paid_seat = run.seats.exclude(
+                type__in=[Seat.AUDIT, Seat.CREDIT]
+            ).first()
```
amangano-edx: What if there are multiple paid seats?

course_discovery/apps/publisher/views.py
```diff
@@ -472,13 +483,25 @@ def _create_or_update_course_entitlement(self, course, entitlement_form):
...
+            if paid_seat:
+                upgrade_deadline = paid_seat.upgrade_deadline
+            run.seats.all().delete()
```
amangano-edx: Just to confirm, we don't save any SKU info or anything like that for these seats?

jlajoie: Yup nothing there is stored for the publisher_seat with regards to SKU, just with the CourseMetadataSeat which wouldn't be created as these won't be published.

course_discovery/apps/publisher/views.py
jlajoie: That shouldn't be the case ever tough I definitely know your hesitance because of this system...I can add validation to throw an exception if > 1 are returned if you think that's sufficient?

amangano-edx: I think validation is appropriate, especially given how loosely state is enforced in this system. I would bet that there are likely several misconfigured courses out there.

course_discovery/apps/publisher/views.py
amangano-edx: Hmm the ACs don't specify whether or not it's ok to allow enrollment track changes when there are published, but inactive runs. You might was to confirm with Deen what the behavior should be in that case.

course_discovery/apps/publisher/views.py
```diff
@@ -472,13 +483,25 @@ def _create_or_update_course_entitlement(self, course, entitlement_form):
...
-            for seat in run.seats.all():
-                # First make sure this seat is on the "entitlement side of things" (i.e. not an audit seat or similar)
-                if seat.type in CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING.values():
-                    seat.type = CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING[entitlement.mode]
-                    seat.price = entitlement.price
-                    seat.currency = entitlement.currency
-                    seat.save()
+            upgrade_deadline = None
+            paid_seat = run.seats.exclude(
+                type__in=[Seat.AUDIT, Seat.CREDIT]
```
amangano-edx: Is there a `PAID_SEATS` constant that could be used instead?

course_discovery/apps/publisher/views.py
```diff
@@ -472,13 +483,25 @@ def _create_or_update_course_entitlement(self, course, entitlement_form):
...
+                run.seats.create(
+                    type=Seat.AUDIT,
+                    price=0,
+                    currency=entitlement.currency
+                )
+            run.seats.create(
+                type=CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING[entitlement.mode],
+                price=entitlement.price,
+                currency=entitlement.currency,
+                upgrade_deadline=upgrade_deadline
```
amangano-edx: Does it actually make sense to pull this in from the old seat? Do `professional` seats have upgrade deadlines (since they don't have audit seats, what would you upgrade from?)?

course_discovery/apps/publisher/views.py
```diff
@@ -472,13 +483,25 @@ def _create_or_update_course_entitlement(self, course, entitlement_form):
...
+            )
```
amangano-edx: Would it make sense to use the `SeatForm` class to create the seats here, since it already has logic to add/remove the additional 'audit' seat?

amangano-edx: One benefit of doing that is that we only need to change one place if the seat-creation logic needs to change.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -3157,10 +3159,39 @@ def refresh_from_db():
         course_data['mode'] = CourseEntitlement.PROFESSIONAL
```
amangano-edx, line 3160: Can we add a test going the other way too (professional -> verified)?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -3157,10 +3159,39 @@ def refresh_from_db():
...
-        refresh_from_db()
+        paid_seat = Seat.objects.get(course_run=course_run)  # Should be only one seat now (no Audit seat)
+        entitlement = CourseEntitlement.objects.get(course=self.course)  # Should be only one entitlement for the Course
         self.assertEqual(entitlement.mode, CourseEntitlement.PROFESSIONAL)
         self.assertEqual(paid_seat.type, Seat.PROFESSIONAL)
-        self.assertEqual(audit_seat.type, Seat.AUDIT)
+
+    def test_entitlement_published_run_failure(self):
+        """
+        Verify that a course with a published course run cannot be saved
```
amangano-edx: To clarify, you should be able to update a course with a published run, just not the enrollment-track/price fields.

course_discovery/apps/publisher/views.py
jlajoie: Yup 100% I'll use the get_active method here thanks for the catch. I know the ACs didn't specify which is why I built to it, but from a standpoint of swapping mode in between course runs it makes sense to have to be able to, but will double check with Deen to verify.

course_discovery/apps/publisher/views.py
jlajoie:
```python
type__in=list(
    map(
        lambda mode: CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING[mode], CourseEntitlementForm.PAID_MODES
    )
)
```

Lambda expression like this might be the best option? Maps the paid modes to the mode -> seat type mappings that exist from the forms.

course_discovery/apps/publisher/views.py
jlajoie: Currently in Prod there exist two `upgrade_deadline` that exist for 2 `professional` seats out of 155. Meanwhile there's 2431 `verified` seats out of 3941 that do not have an `upgrade_deadline`. I'll swap it two be cased to not include it, was just trying to preserve data but this doesn't really make sense to keep.

course_discovery/apps/publisher/views.py
jlajoie: Potentially? Though it feels pretty weird to utilize a form that isn't actually visible from the view this is presented in. I'd imagine moving that logic out to it's own method that is shared between the SeatForm and this method, but I think that's a bit too large of a refactor for this PR.

course_discovery/apps/publisher/tests/test_views.py
jlajoie: 💯 absolutely

course_discovery/apps/publisher/views.py
jlajoie: Just going off of what is currently used to identify course runs, it's the same format and used elsewhere.

screen shot 2018-03-20 at 9 28 59 am

course_discovery/apps/publisher/views.py
amangano-edx: I agree it feels a little weird, but it's not unprecedented. We use the SeatForm in the CourseRunCreateView to add seats when creating runs for ENTITLEMENT_MODE courses.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -3157,10 +3159,56 @@ def refresh_from_db():
...
-        refresh_from_db()
+        paid_seat = Seat.objects.get(course_run=course_run)  # Should be only one seat now (no Audit seat)
+        entitlement = CourseEntitlement.objects.get(course=self.course)  # Should be only one entitlement for the Course
         self.assertEqual(entitlement.mode, CourseEntitlement.PROFESSIONAL)
         self.assertEqual(paid_seat.type, Seat.PROFESSIONAL)
-        self.assertEqual(audit_seat.type, Seat.AUDIT)
+
+        # Test mode and price change again AFTER publishing
+        # course_run.course_run_state.name = CourseRunStateChoices.Published
+        # course_run.course_run_state.save()
```
amangano-edx: Can we remove this commented-out code?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -3157,10 +3159,56 @@ def refresh_from_db():
...
         response = self.client.post(self.edit_page_url, data=course_data)
         self.assertEqual(response.status_code, 302)
-        refresh_from_db()
+        paid_seat = Seat.objects.get(course_run=course_run)  # Should be only one seat now (no Audit seat)
+        entitlement = CourseEntitlement.objects.get(course=self.course)  # Should be only one entitlement for the Course
         self.assertEqual(entitlement.mode, CourseEntitlement.PROFESSIONAL)
         self.assertEqual(paid_seat.type, Seat.PROFESSIONAL)
-        self.assertEqual(audit_seat.type, Seat.AUDIT)
+
+        # Test mode and price change again AFTER publishing
```
amangano-edx: Shouldn't this not be allowed?

amangano-edx: Per the ACs here: https://openedx.atlassian.net/browse/LEARNER-4477

jlajoie: Which AC does this not fulfill? It is a professional course with only a single professional seat and no audit.

amangano-edx (approved):
Looks ok to me once the commented out code is removed.

course_discovery/apps/publisher/tests/test_views.py
jlajoie: Absolutely, good catch there was testing the failure case originally but swapped it, forgot to delete this.

course_discovery/apps/publisher/tests/test_views.py
amangano-edx: 'VERIFY that enrollment-track cannot be changed at the course-level for any ENTITLEMENT_VERSION course associated with a run that has been published.'

If you publish the course-run, it should no longer be possible to edit the mode.

jlajoie: Sorry this is a poor comment that I didn't even notice, meant it to be after saving and before publishing and skipped a few words. This test is for the success case of before publishing allowing you to swap back and forth. I'll amend since the after publishing case is handled in `test_entitlement_published_run_failure`

PR #1358, opened 2018-03-19
query course org using course's first run's org

McKenzieW:
Allow courses to be found by elasticsearch with a querystring containing "org:*" using the organization key of the course's first course run.

[LEARNER-4166](https://openedx.atlassian.net/browse/LEARNER-4166), [Discovery subtask](https://openedx.atlassian.net/browse/LEARNER-4360)

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -119,6 +119,7 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
     model = Course
 
     uuid = indexes.CharField(model_attr='uuid')
+    org = indexes.CharField()
     course_runs = indexes.MultiValueField()
     expected_learning_items = indexes.MultiValueField()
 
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -136,6 +140,13 @@ def prepare_expected_learning_items(self, obj):
     def prepare_prerequisites(self, obj):
         return [prerequisite.name for prerequisite in obj.prerequisites.all()]
 
+    def prepare_org(self, obj):
+        course_run = obj.course_runs.all().first()
```
Mjloturco, line 141: Would it be possible for the various runs in a course to have different orgs? if so this line might give some unpredictable results.

Mjloturco (approved):
One comment, and you may want to remove the log statement, otherwise looks good

course_discovery/apps/course_metadata/search_indexes.py
Mjloturco: may not be a usecase we would run into/need to support

McKenzieW: I don't believe that would be a case. The org is pulled directly out of the course run's key, all of which are just an extensions of their Course's key.
My initial attempt to allow org search on Courses included an attempt to extract the org form the course object's key; that was unfortunately unsuccessful, but I believe this should function essentially the same.

staubina (approved):
Seems like a straightforward set of changes.

PR #1359, opened 2018-03-19
Defaulted one_click_purchase_enabled to True

staubina:
[LEARNER-4061]

A small change to make `one_click_purchase_enabled` default to True.


McKenzieW (approved):

course_discovery/apps/course_metadata/migrations/0079_auto_20180319_1718.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
Mjloturco: You could give this migration a meaningful name

Mjloturco (approved):
could add a test that verifies on new program creation that 1-click is enabled, or just an assertion to an existing creation test. otherwise LGTM

staubina:
> could add a test that verifies on new program creation that 1-click is enabled

I am not sure I see the value in adding a test as it is part of the model.  

> You could give this migration a meaningful name

I will take a look at renaming the migration.

staubina:
@Mjloturco I renamed the migration per your request.

PR #1362, opened 2018-03-21
ENT-802 Include enrollment code SKU

adamstankiewicz:
**Description**
To support bulk purchase from the marketing site, we needed to gain access to enrollment code SKUs in the course run API endpoint. 

This PR updates the `refresh_course_metadata` command to retrieve and add enrollment code products from ecommerce to the appropriate seat and course run using the new `bulk_sku` field on the Seat model. If a seat for a course run does not have an associated enrollment code product, the `bulk_sku` field will be `null`. Additionally, in the API endpoint for course runs, we include the `bulk_sku` field for seats.

Jira Ticket: [ENT-802](https://openedx.atlassian.net/browse/ENT-802)

**Testing Instructions**
- [ ] Run `./manage.py refresh_course_metadata` for discovery service on business sandbox.
- [ ] Verify that enrollment codes were retrieved and created as seen in the logged output.
- [ ] Make sure `bulk_sku` field exists in each seat for all course runs: https://discovery-business.sandbox.edx.org/api/v1/course_runs/. `bulk_sku` will be `null` if an enrollment code does not exist for that seat type.
- [ ] Check that enrollment code SKUs are added to the appropriate seats and course. _Example: `bulk_sku` is correctly populated for verified seat: https://discovery-business.sandbox.edx.org/api/v1/course_runs/course-v1%3AMattx%2BTBE12%2B2018/_


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -265,11 +266,14 @@ def ingest(self):
         initial_page = 1
         course_runs = self._request_course_runs(initial_page)
         entitlements = self._request_entitlments(initial_page)
+        enrollment_codes = self._request_enrollment_codes(initial_page)
         count = course_runs['count'] + entitlements['count']
```
brittneyexline: I know you didn't write this line but I don't understand if it needs to be updated or not...

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -453,6 +474,73 @@ def update_entitlement(self, body):
...
+        Argument:
+            body (dict): enrollment code product data from ecommerce
+        Returns:
+            enrollment code product sku if no exceptions, else None
+        """
+        attributes = {attribute['code']: attribute['value'] for attribute in body['attribute_values']}
+        course_key = attributes.get('course_key')
+        title = body['title']
+
+        if body['stockrecords']:
```
brittneyexline: seems like you could refactor this validation logic out since it is the same as the entitlements

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -365,61 +366,152 @@ def mock_courses_api(self):
...
-    def assert_seats_loaded(self, body):
+    def get_product_bulk_sku(self, seat_type, course_run, products):
+        bulk_sku = None
+        products = [p for p in products if p['structure'] == 'standalone']
+        course_key = course_run.key
+        for product in products:
+            attributes = {attribute['code']: attribute['value'] for attribute in product['attribute_values']}
+            if attributes['seat_type'] == seat_type and attributes['course_key'] == course_key:
+                stock_record = product['stockrecords'][0]
+                bulk_sku = stock_record['partner_sku']
```
brittneyexline: you could return here and return None below and get rid of the bulk_sku variable.

brittneyexline (commented):
looks good! I noticed there seems to be some duplicated code between processing enrollment codes and course entitlements - I think there are some opportunities for refactoring the common code to one place.

course_discovery/apps/course_metadata/data_loaders/api.py
adamstankiewicz: Yeah, I'm still a little unsure, too but I don't think it needs to be updated because the enrollment codes ultimately end up updating the seats associated with course runs whereas entitlement are totally separate from the course runs.

That being said, I'm still a bit confused how `count` in this case ends up generating an accurate number of pages in the following line.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
adamstankiewicz: True. I usually try to avoid having multiple return statements but it would make sense here to avoid the unnecessary variable.

course_discovery/apps/course_metadata/data_loaders/api.py
adamstankiewicz: This logic is not exactly the same as the entitlements validation logic. For example, we check for `CourseRun` for enrollment codes but `Course` for entitlements. Very similar, though. It makes sense to refactor the logic into its own method and have that method take a `product_class` parameter that determines the differing logic.

brittneyexline: yea, I see that it is slightly different not just here but throughout - it's on that line where it's hard to tell how far to go with getting all the common stuff in one place. I think it's worth seeing what can be DRYer.

course_discovery/apps/course_metadata/data_loaders/api.py
brittneyexline: ok. the entitlement code is relatively new - I would see if McKenzie can explain what this is trying to do, to make sure we don't have any nasty surprises down the road.

course_discovery/apps/course_metadata/data_loaders/api.py
adamstankiewicz: Pushed up a refactor to make it DRYer. Lmk what you think :)

course_discovery/apps/course_metadata/data_loaders/api.py
adamstankiewicz: I chatted with McKenzie about this and we determined that the `enrollment_codes['count']` should be included in the total summation to calculate the pages. I've added it in.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -382,76 +403,196 @@ def update_seat(self, course_run, product_body):
...
+            body (dict): product data from ecommerce, either entitlement or enrollment code
         Returns:
-            entitlement product sku if no exceptions, else None
+            product sku if no exceptions, else None
         """
-        attributes = {attribute['name']: attribute['value'] for attribute in body['attribute_values']}
-        course_uuid = attributes.get('UUID')
+        product_classes = {
+            "entitlement": {
+                "name": "entitlement",
```
brittneyexline: style nit: i would use single quotes instead of double quotes throughout all of this.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -382,76 +403,196 @@ def update_seat(self, course_run, product_body):
...
             )
             logger.warning(msg)
             return None
 
         try:
-            course = Course.objects.get(uuid=course_uuid)
-        except Course.DoesNotExist:
-            msg = 'Could not find course {uuid} while loading entitlement {entitlement} with sku {sku}'.format(
-                uuid=course_uuid, entitlement=title, sku=sku
+            currency_code = stock_record['price_currency']
```
brittneyexline: this looks like the same code as above?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -382,76 +403,196 @@ def update_seat(self, course_run, product_body):
...
             return None
 
-        mode_name = attributes.get('certificate_type')
-        try:
-            mode = SeatType.objects.get(slug=mode_name)
-        except SeatType.DoesNotExist:
-            msg = 'Could not find mode {mode} while loading entitlement {entitlement} with sku {sku}'.format(
-                mode=mode_name, entitlement=title, sku=sku
+        # entitlement specific validation
+        if product_class["name"] == "entitlement":
```
brittneyexline: I think the validation logic above makes sense to live in one function, but I was imagining that the two separate functions would both call the truly common validation logic (which appears to all be related to the stockrecord) and then do this stuff below on their own. i imagine something like this:
```
def update_entitlement(self, body):
    if not validate_stockrecord(body['stockrecord']):
         // log error and return none
    // do the rest of the entitlement specific logic

def update_enrollment_code(self, body):
     if not validate_stockrecord(body['stockrecord']):
         // log error and return none
    // do the rest of the enrollement code specific logic

def validate_stockrecord(stockrecord):
    if not stockrecord:
        return None
    // do the stockrecord related validation logic
```

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -382,76 +403,196 @@ def update_seat(self, course_run, product_body):
...
+                    "course_run": course_run
+                },
+                "defaults": {
+                    "bulk_sku": sku
+                }
+            }
+
+        return response
+
+    def update_entitlement(self, body):
```
brittneyexline: see above - i would pull the response logic back in here since it isn't actually shared.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -265,11 +266,14 @@ def ingest(self):
         initial_page = 1
         course_runs = self._request_course_runs(initial_page)
         entitlements = self._request_entitlments(initial_page)
-        count = course_runs['count'] + entitlements['count']
+        enrollment_codes = self._request_enrollment_codes(initial_page)
+        count = course_runs['count'] + entitlements['count'] + enrollment_codes['count']
```
brittneyexline: thanks for checking with McKenzie! could you add a comment explaining why we do this? i still don't really understand this line...

brittneyexline (approved):

PR #1368, opened 2018-03-23
Journal REST API fixes

bill-filler:


course_discovery/apps/journal/api/serializers.py
```diff
@@ -3,29 +3,29 @@
...
-    #def prefetch_queryset(cls):
-    #   return Journal.objects.all().select_related('currency')
+    partner = serializers.SlugRelatedField(slug_field='name', queryset=Partner.objects.all())
+    currency = serializers.SlugRelatedField(slug_field='code', queryset=Currency.objects.all())
 
     class Meta(object):
         model = Journal
-        # TODO - add partner
-        fields = ('uuid', 'title', 'price', 'currency', 'sku', 'expires')
+        fields = ('id', 'uuid', 'partner', 'title', 'price', 'currency', 'sku', 'expires')
```
bessiesteinberg: I don't think we need 'id' exposed in the api, i think 'uuid' should be enough

bessiesteinberg (approved):

PR #1370, opened 2018-03-26
ENT-921: Create enrollment codes by default for course publication to ecommerce.

saleem-latif:
**JIRA Ticket:** [ENT-921](https://openedx.atlassian.net/browse/ENT-921)
 
__Description:__
As part of [ENT-803](https://openedx.atlassian.net/browse/ENT-803), we discovered that Publisher was making a call to ecommerce to create courses, and by default it is sending false for the parameter to create enrollment code products. This needs to be true so that enrollment code products are always created for new courses.
**Acceptance Criteria:**

1.  When the action in Publisher is taken that results in hitting ecommerce to create the course object(s) there, it should create the enrollment code products by default.

### DIFF
--- a/course_discovery/apps/publisher/api/v1/views.py
+++ b/course_discovery/apps/publisher/api/v1/views.py
@@ -88,7 +88,6 @@ class CourseRunViewSet(viewsets.GenericViewSet):
             'uuid': str(discovery_course.uuid),
             'name': course_run.title_override or course_run.course.title,
             'verification_deadline': serialize_datetime(course_run.end),
-            'create_or_activate_enrollment_code': False,
         }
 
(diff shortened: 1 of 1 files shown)


asadiqbal08 (approved):
LGTM

georgebabey:
Looks good - no tests for this though? We should probably add one to test that default value of true.

georgebabey (approved):

saleem-latif:
@georgebabey This code updates payload data of an API call and since API calls are mocked in tests, I do not think adding tests would be useful.

saleem-latif:
Verified on business sandbox, enrollment codes are now being created when course is published via publisher tool.

PR #1373, opened 2018-03-29
Update db limit of run short description override.

waheedahmed:
Increase the course run short description override db limit to count for foreign characters.

LEARNER-4673


waheedahmed:
@edx/learner-spartans Please review.

@edx/learner @schenedx fyi.

ahsan-ul-haq (approved):

PR #1374, opened 2018-03-29
update journal model with new fields

bill-filler:


bessiesteinberg (approved):

course_discovery/apps/journal/models.py
```diff
@@ -23,16 +23,24 @@ class Journal(TimeStampedModel):
         verbose_name=_('UUID'),
     )
     partner = models.ForeignKey(Partner)
+    organization = models.ForeignKey(Organization, default=1)
```
tuchfarber: Why do we want it to default to the first org?

course_discovery/apps/journal/models.py
```diff
@@ -23,16 +23,24 @@ class Journal(TimeStampedModel):
...
     )
+
+    # ecommerce related
     price = models.DecimalField(**PRICE_FIELD_CONFIG)
     currency = models.ForeignKey(Currency)
     sku = models.CharField(max_length=128, null=True, blank=True)
-    expires = models.DateTimeField(null=True, blank=True)
+
+    # marketing related fields
+    card_image_url = models.URLField(null=True, blank=True)
```
tuchfarber, line 40: Do we ant to store hero_image_url and basket_image_url? If the hero image is just being used on Journals, then I'd say no to that one. Not sure if ecomm pulls the basket_url from disc or stores it locally.

course_discovery/apps/journal/models.py
bill-filler: crap, didn't mean to leave that in. Was getting an error on migration. let me remove it

bill-filler:
@tuchfarber 
updated per comments

tuchfarber (approved):

PR #1382, opened 2018-04-03
Bfiller/tweak model and serializer

bill-filler:
JournalModel make uuid writeable as need by REST API
Add field for Organization to serializer


course_discovery/apps/journal/models.py
```diff
@@ -19,7 +19,6 @@ class Journal(TimeStampedModel):
     }
     uuid = models.UUIDField(
         default=uuid4,
```
tuchfarber: Since we're pushing UUID to discovery, do we want it to generate it's own UUID as a default?

tuchfarber (approved):
Good to merge! Just got a question on the default value for UUID

bessiesteinberg (approved):

course_discovery/apps/journal/models.py
bill-filler: @tuchfarber good point, probably not

PR #1385, opened 2018-04-04
Added test logging for resolving error

staubina:
@McKenzieW @Mjloturco 

I think we should add the following temporary logging to find where we are dropping this entitlement.

What do you think?


staubina:
@Mjloturco @McKenzieW I posted a fix for the failed test.

Mjloturco:
Looks helpful, could we include a consistent string in each of these logs that makes them more searchable? possibly (liderazgo) or (temp_debug),

McKenzieW (approved):

PR #1399, opened 2018-04-17
Fix translations

Ayub-Khan:
Description
-pull atleast 1% translated languages
-added push translations make target
-removed generate merge options
-removed i18n tools commands from extract translations
-removed existing translations to git rid of untranslated languages
-extracted translations
-new translations
-compile messages


Makefile
```diff
@@ -76,7 +76,10 @@ compile_translations: ## Compile translation files, outputting .po files for eac
 fake_translations: extract_translations dummy_translations compile_translations ## Generate and compile dummy translation files
 
 pull_translations: ## Pull translations from Transifex
-	tx pull -af --mode reviewed
+	tx pull -af --mode reviewed --minimum-perc=1
```
ahsan-ul-haq, line 79: Whats --minimum-perc=1 for?

ahsan-ul-haq (approved):

Makefile
Ayub-Khan: As the code works now it pulls all the translations languages available. So if new translation language is added it will have empty header and empty strings which will fail our test which we are adding in other pr. by specifying this we are telling transifex to only pull translations which are atleast 1% translated so they will have some translations and a proper header.

PR #1402, opened 2018-04-19
add IsAdminUser permissions to Journal REST apis

bill-filler:


tuchfarber:
Do you see a use case where a user would need to query the journals via AJAX? If so we could make the permission for read access IsAuthenticated and for write access IsAdminUser. If we're under the assumption it's all back end calls, then this is good as is.

tuchfarber (approved):

bill-filler:
@tuchfarber 
not sure about needing to query this endpoint via AJAX, can't think of a use case where we need it yet. If we do we can relax the permissions when the time comes

PR #1404, opened 2018-04-23
Update corresponding drupal node on instructor update via publisher

asadazam93:
### Description
When an instructor is updated via the publisher front end it will update the corresponding person node on the drupal site. Since we do not have the node id in discovery (which is required to update a node in drupal) a get request is first sent to drupal with the uuid to get the corresponding node id and then the node is updated via the node id.
If no node exists for a uuid in that case only discovery will be updated and the drupal update will be skipped. (This is an extreme edge case as for all person created via publisher we always fill the uuid but since there are some instructors that have uuid empty in the existing database I have covered this edge case)

**Sandbox**
N/A

### Reviewers
- [ ] @schenedx 
- [ ] @awaisdar001 

### Post-review

### DIFF
--- a/course_discovery/apps/api/v1/views/people.py
+++ b/course_discovery/apps/api/v1/views/people.py
@@ -28,7 +28,9 @@ class PersonViewSet(viewsets.ModelViewSet):
     pagination_class = PageNumberPagination
 
     def create(self, request, *args, **kwargs):
-        """ Create a new person. """
+        """
+        Create a person in discovery and also create a person node in drupal
+        """
         person_data = request.data
 
         partner = request.site.partner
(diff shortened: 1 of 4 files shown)


asadazam93:
Adding test case for update

course_discovery/apps/api/v1/views/people.py
```diff
@@ -69,7 +69,56 @@ def create(self, request, *args, **kwargs):
...
+                    'An error occurred while updating the person [%s]-[%s] on the marketing site.',
+                    serializer.validated_data['given_name'], serializer.validated_data['family_name']
+                )
+                return Response(
+                    'Failed to update person data on the marketing site.',
+                    status=status.HTTP_400_BAD_REQUEST
+                )
+
+            try:
+                self.perform_update(serializer)
```
awaisdar001: Where this update is performed after the call to `marketing_person.update_person`?

course_discovery/apps/api/v1/views/people.py
```diff
@@ -69,7 +69,56 @@ def create(self, request, *args, **kwargs):
...
+        person_data = request.data
+
+        partner = request.site.partner
+        person_data['partner'] = partner.id
+        partial = kwargs.pop('partial', False)
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=person_data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+
+        if waffle.switch_is_active('publish_person_to_marketing_site'):
```
awaisdar001: We can simplify the indentation by
```
if not waffle.switch_is_active('publish_person_to_marketing_site'):
    return

```

Even better 
```
if not publish_person_to_marketing():
    return
```

course_discovery/apps/api/v1/views/people.py
```diff
@@ -69,7 +69,56 @@ def create(self, request, *args, **kwargs):
             headers = self.get_success_headers(serializer.data)
             return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
-        return Response('publish_program_to_marketing_site is disabled.', status=status.HTTP_400_BAD_REQUEST)
+        return Response('publish_person_to_marketing_site is disabled.', status=status.HTTP_400_BAD_REQUEST)
+
+    def update(self, request, *args, **kwargs):
+        """
+        Update a person
```
awaisdar001: Please be more descriptive where it updates the person ?

course_discovery/apps/api/v1/views/people.py
asadazam93: This updates the Person instance in course metadata

awaisdar001: Can we the method name readable?

asadazam93: This method is inherited from django ModelViewSet

course_discovery/apps/api/v1/views/people.py
```diff
@@ -69,7 +69,56 @@ def create(self, request, *args, **kwargs):
...
+        serializer.is_valid(raise_exception=True)
+
+        if waffle.switch_is_active('publish_person_to_marketing_site'):
+            try:
+                marketing_person = MarketingSitePeople()
+                response = marketing_person.update_person(
+                    partner, {
+                        'given_name': serializer.validated_data['given_name'],
+                        'family_name': serializer.validated_data['family_name'],
+                        'uuid': person_data['uuid']
```
schenedx: Why do we want to update `UUID`? This value originally comes from the marketings site anyways. Let's not update it.

course_discovery/apps/api/v1/views/people.py
```diff
@@ -69,7 +69,56 @@ def create(self, request, *args, **kwargs):
...
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=person_data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+
+        if waffle.switch_is_active('publish_person_to_marketing_site'):
+            try:
+                marketing_person = MarketingSitePeople()
+                response = marketing_person.update_person(
+                    partner, {
+                        'given_name': serializer.validated_data['given_name'],
```
schenedx: I believe this is the same list of fields as the `create` method above. Can we refactor this list of fields to create/update into a single function so we can change this later with needs?

course_discovery/apps/api/v1/views/people.py
schenedx: This method should be performed whether or not the waffle flag `publish_person_to_marketing_site` is true. I think you put this inside the if statement?

schenedx: Or the waffle flag really is `enable_update_person`?

schenedx (changes requested):
Please add unit tests

course_discovery/apps/api/v1/views/people.py
asadazam93: I am just passing the uuid to the update_person method as it uses the uuid to request for the node id. This is not the data to be updated. The data to be updated is returned by _get_node_data method

course_discovery/apps/api/v1/views/people.py
asadazam93: The list is slightly different as I pass uuid to the update_person method which is not passed to create method.

course_discovery/apps/api/v1/views/people.py
asadazam93: This is how it has been implemented in the create method as well. This is just to make the data consistent. If one is updated and the other isn't that would make the data inconsistent. And the way it currently works the data would be reverted as drupal would override discovery when refresh metadata runs. I think updating the flag is what we need to do here

asadazam93:
Adding unit tests

course_discovery/apps/api/v1/views/people.py
schenedx: If you are passing in node_uuid explicitly, you will have these two objects the same.

course_discovery/apps/api/v1/views/people.py
schenedx: If you are passing the UUID into the update_person method because it needs it, please make the parameter of the method be explicit! Your `update_person` method should have this signature:
```
def update_person(self, partner, node_uuid, person):
```

schenedx (commented):
See my comments on UUID

course_discovery/apps/api/v1/views/people.py
```diff
@@ -36,40 +36,89 @@ def create(self, request, *args, **kwargs):
...
+                'An error occurred while adding the person [%s]-[%s] to the marketing site.',
+                serializer.validated_data['given_name'], serializer.validated_data['family_name']
+            )
+            return Response('Failed to add person data to the marketing site.', status=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            self.perform_create(serializer)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(
+                'An error occurred while adding the person [%s]-[%s]-[%s].',
```
awaisdar001: Can this error be explicit that where the error occurred? i.e. adding in discovery.

course_discovery/apps/api/v1/views/people.py
```diff
@@ -36,40 +36,89 @@ def create(self, request, *args, **kwargs):
...
+            )
+            marketing_person.delete_person(partner, response['id'])
+            return Response('Failed to add person data.', status=status.HTTP_400_BAD_REQUEST)
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def update(self, request, *args, **kwargs):
+        """
+        Update a person
```
awaisdar001: Doc string

course_discovery/apps/course_metadata/people.py
```diff
@@ -35,6 +36,26 @@ def _create_node(self, api_client, node_data):
...
+            raise PersonToMarketingException("Marketing site Person page update failed!")
+
+    def update_person(self, partner, person_uuid, person):
+        api_client = self._get_api_client(partner)
+        node_id = self._get_node_id_from_uuid(api_client, person_uuid)
+        if node_id:
+            node_data = self._get_node_data(person)
+            return self._update_node(api_client, node_id, node_data)
+        else:
+            logger.info('Person with UUID [%s] does not exist on the marketing site', person_uuid)
```
awaisdar001, line 57: is it intentional to use `info` here?

schenedx:
@asadazam93 We should have some unit tests on the view as well.

course_discovery/apps/course_metadata/people.py
asadazam93: yes it is intentional. If drupal node is not found it should update the course discovery instance. Just added a log here so that if any data revert issue comes up we can check if the node was found or not.

asadazam93:
@schenedx the view tests already exist. I have updated the test

awaisdar001 (approved):

course_discovery/apps/api/v1/views/people.py
```diff
@@ -28,48 +28,99 @@ class PersonViewSet(viewsets.ModelViewSet):
...
+                'An error occurred while updating the person [%s]-[%s]-[%s] in discovery.',
+                serializer.validated_data['given_name'], serializer.validated_data['family_name'],
+                response['id']
+            )
+            return Response('Failed to update person data.', status=status.HTTP_400_BAD_REQUEST)
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_200_OK, headers=headers)
+
+    def _get_person_data(self, serializer):
```
schenedx, line 118: Nice!

course_discovery/apps/course_metadata/people.py
```diff
@@ -22,6 +22,7 @@ def _get_node_data(self, person):
         return {
             'field_person_first_middle_name': person['given_name'],
             'field_person_last_name': person['family_name'],
+            'title': person['given_name'] + ' ' + person['family_name'],
```
schenedx, line 25: This is the title of the actual drupal node? It's not learner visible, correct?

schenedx (approved):
:+1:

course_discovery/apps/course_metadata/people.py
asadazam93: yes correct

PR #1410, opened 2018-05-04
Fix course run image in serializer

tuchfarber:
Serializer was using not existent field causing us to not be able to get the course card image.


tuchfarber:
This field is working correctly, ignore this PR.

PR #1416, opened 2018-05-15
Add default status unpublished for newly created runs.

waheedahmed:
Add default status unpublished for newly created course runs, also update existing course runs with null/empty status to unpublished.

LEARNER-5202


waheedahmed:
@edx/learner-spartans Please review.

course_discovery/apps/course_metadata/migrations/0083_auto_20180511_1406.py
```diff
@@ -0,0 +1,31 @@
...
+
+import djchoices.choices
+from django.db import migrations, models
+
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
+
+
+def change_runs_null_state_to_unpublished(apps, schema_editor):
+    CourseRun = apps.get_model('course_metadata', 'CourseRun')
+    for course_run in CourseRun.objects.filter(status=''):
```
staubina, line 13: Is status='' the same as NULL?

staubina (approved):
Changes and migration seem straight forward I would like a second opinion on @edx/learner-spartans though.

course_discovery/apps/course_metadata/migrations/0083_auto_20180511_1406.py
waheedahmed: @staubina No, but in database it's a empty string instead of NULL, verified on read replica.

ahsan-ul-haq (approved):

PR #1417, opened 2018-05-16
Dont start newrelic transaction on heartbeat

coryleeio:


jdmulloy (approved):
Might want to update the commit message, but otherwise looks good.

PR #1418, opened 2018-05-17
fix for update price of published course runs

tasawernawaz:
[LEARNER-5284](https://openedx.atlassian.net/browse/LEARNER-5284)

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -574,7 +574,7 @@ class CourseEditView(mixins.PublisherPermissionMixin, UpdateView):
                 })
             published_runs = self._get_published_course_runs(self.object)
             # Only check published runs if there are changes to the mode or price
-            if published_runs and (entitlement.mode != entitlement_mode or entitlement.price != entitlement_price):
+            if published_runs and entitlement.mode != entitlement_mode:
                 # pylint: disable=no-member
                 error_message = _(
                     'The following active course run(s) are published: {course_runs}. You cannot change the mode '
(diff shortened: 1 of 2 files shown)


course_discovery/apps/publisher/views.py
```diff
@@ -574,7 +574,7 @@ def _handle_entitlement_update(self, user, request, course_form):
                 })
             published_runs = self._get_published_course_runs(self.object)
             # Only check published runs if there are changes to the mode or price
-            if published_runs and (entitlement.mode != entitlement_mode or entitlement.price != entitlement_price):
+            if published_runs and entitlement.mode != entitlement_mode:
```
tasawernawaz, line 577: @staubina I have removed this check but by looking at the code it will subsequently call [_update_seats_from_entitlement](https://github.com/edx/course-discovery/blob/c01da1579c4f98a0ad9aead00cd73b843ca4119a/course_discovery/apps/publisher/views.py#L484) which will delete and recreate the seats. I am not sure if this is our desired behavior.

staubina: @jlajoie Can you answer the question above?

jlajoie: Yes that's the desired behavior. The intent is for us to cascade the changes for price and mode (though in the case just price) to all of the children course runs and seats.

jlajoie (approved):
Looks good to me 👍

course_discovery/apps/publisher/views.py
tasawernawaz: thanks

PR #1421, opened 2018-05-21
ENT-1013 Discovery model classes don't have ordering set

brittneyexline:
This resolves a bunch of noise in the error logs, as well as weird
behavior when paginating through results for things like courses.


georgebabey (approved):

adamstankiewicz (approved):
LGTM

PR #1426, opened 2018-05-28
WL-1575: Update partner if already exists

ziafazal:
This PR has changes to update partner data if partner already exists when we run `create_sites_and_partners` command.
@hasnain-naveed could you please review?


hasnain-naveed (approved):
All the configurations are good to go.

PR #1431, opened 2018-06-06
Create new instructor position if blank

asadazam93:
### Description
Blank position check in javascript while loading instructors on the course run edit page. Also if a person has no position and that person is updated with new position data it should create a new position for that person.

**Sandbox**
N/A

### Reviewers
- [ ] @attiyaIshaque
- [ ] @awaisdar001

### Post-review

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -266,10 +266,7 @@ class PersonSerializer(serializers.ModelSerializer):
         works_data = validated_data.pop('works', [])
         urls_data = validated_data.pop('urls', {})
 
-        instance.position.title = position_data['title']
-        instance.position.organization = position_data.get('organization')
-        instance.position.organization_override = position_data.get('organization_override')
-        instance.position.save()
+        Position.objects.update_or_create(person=instance, defaults=position_data)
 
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -266,10 +266,13 @@ def update(self, instance, validated_data):
...
-        instance.position.organization = position_data.get('organization')
-        instance.position.organization_override = position_data.get('organization_override')
-        instance.position.save()
+        try:
+            instance.position.title = position_data['title']
+            instance.position.organization = position_data.get('organization')
+            instance.position.organization_override = position_data.get('organization_override')
+            instance.position.save()
+        except Position.DoesNotExist:
+            Position.objects.create(person=instance, **position_data)
```
awaisdar001: Can we use `create_or_update`?

attiyaIshaque (approved):
Thanks for fixing this. 👍

awaisdar001 (changes requested):

course_discovery/apps/api/serializers.py
asadazam93: @awaisdar001 there is no unique attribute to use create_or_update. This is just for the instructors that have no position because their position was deleted. Position can no longer be deleted via the admin site. So once all the existing instructors have positions we can even remove this part of the code.

awaisdar001 (approved):

PR #1432, opened 2018-06-06
remove sortedM2M field from horizontal filter on program admin

bderusha:


bderusha:
The current code prevents these ordered fields from being orderable in the admin.  Fixed here.

FYI @mikedikan @rock345

mikedikan (approved):
Maybe a follow up PR after the hackathon can add a test?

bderusha:
@mikedikan I'm open to suggestions for how to test this.  It's letting the admin field render with it's default UI.

PR #1439, opened 2018-06-13
Fixed langauage not saving publisher

asadazam93:
### Description
The course run form was not saving many to many relationships so transcript languages which is a ManytoMany field of a course run was not being saved previously. I have explicitly saved the ManytoMany relationships for a course run.

**Sandbox**
N/A

### Reviewers
- [ ] @noraiz-anwar
- [ ] @awaisdar001

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/views.py
+++ b/course_discovery/apps/publisher/views.py
@@ -1044,6 +1044,7 @@ class CourseRunEditView(mixins.LoginRequiredMixin, mixins.PublisherPermissionMix
             try:
                 with transaction.atomic():
                     course_run = run_form.save(changed_by=user)
+                    run_form.save_m2m()
                     course_run.staff.clear()
                     course_run.staff.add(*staff)
 


course_discovery/apps/publisher/views.py
```diff
@@ -1044,6 +1044,7 @@ def post(self, request, *args, **kwargs):
             try:
                 with transaction.atomic():
                     course_run = run_form.save(changed_by=user)
+                    run_form.save_m2m()
```
awaisdar001, line 1047: Can we add tests for this?

awaisdar001 (changes requested):

course_discovery/apps/publisher/views.py
asadazam93: @awaisdar001 This is an urgent fix and needs to be deployed today to unblock course teams. I will create another PR and add the test.

awaisdar001 (approved):
Given your comment on urgency. 👍

PR #1440, opened 2018-06-14
Add wait time to courses api for refresh_course_metadata

BenjiLee:
This is the quickest "handling" of 429s I can think/code up today.

https://openedx.atlassian.net/browse/LEARNER-5560

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -1,6 +1,7 @@
 import concurrent.futures
 import logging
 import math
+import time
 from decimal import Decimal
 from io import BytesIO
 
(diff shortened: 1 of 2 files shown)


feanil (approved):

BenjiLee:
The only issue is that I need to make sure that it gets to that specific logic. May have to look at this again tomorrow.

Update: I added the sleeps to both conditions.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -84,6 +86,7 @@ def ingest(self):
         with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:  # pragma: no cover
             if self.is_threadsafe:
                 for page in pagerange:
+                    time.sleep(30)
```
robrap: Do we have any control over page size?
Can you provide a calculation for page size and expected calls before waiting 30 seconds?  Do we need to wait 60+ seconds to avoid future rate limiting, or is that not a concern for now?
 
Please add a comment with the following:
1. Calculations of potential call rate and what we need/want.
2. A JIRA ticket linking to details about a deeper follow-up fix.

BenjiLee: >Do we have any control over page size?

Yes.

Let's talk offline about this.

BenjiLee: > Can you provide a calculation for page size and expected calls before waiting 30 seconds? Do we need to wait 60+ seconds to avoid future rate limiting, or is that not a concern for now?
 
Our crude estimate was to span this over 30 mins. There are ~3000 course and if we are using a page_size of 50, it would be 60 calls. With a time.sleep(30), that should be about 30 mins. I don't think we need to worry about being rate limited for this management command as this number leaned more towards the conservative side and intended to avoid a 429 altogether rather than trying to address a 429.

BenjiLee: > 1. Calculations of potential call rate and what we need/want.
> 2. A JIRA ticket linking to details about a deeper follow-up fix.

Added

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -84,9 +85,21 @@ def ingest(self):
         with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:  # pragma: no cover
             if self.is_threadsafe:
                 for page in pagerange:
+                    # This time.sleep is to make it very unlikely that this method does not encounter a 429 status
```
BenjiLee: doubel nagative

PR #1447, opened 2018-06-27
Add staff permission on discovery people API.

attiyaIshaque:
### Description
This PR fixes that add staff permission on the discovery people API. The list of all people only accessible by staff members.

### After Fix
screen shot 2018-06-27 at 2 57 56 pm

### Test
- [x] unit tests

### Reviewers
- [x] @schenedx  
- [x] @asadazam93 

### Post-review

### DIFF
--- a/course_discovery/apps/api/permissions.py
+++ b/course_discovery/apps/api/permissions.py
@@ -1,0 +1,11 @@
+from rest_framework.permissions import BasePermission
+
+
+class ReadByStaffOnly(BasePermission):
+    """
+    Custom permission to only allow owners of the object.
+    """
+    def has_permission(self, request, view):
+        if request.method == 'GET':
+                return request.user.is_staff
+        return True
--- a/course_discovery/apps/api/v1/views/people.py
+++ b/course_discovery/apps/api/v1/views/people.py
@@ -8,6 +8,7 @@ from rest_framework.response import Response
 
 from course_discovery.apps.api import filters, serializers
 from course_discovery.apps.api.pagination import PageNumberPagination
+from course_discovery.apps.api.permissions import ReadByStaffOnly
 from course_discovery.apps.course_metadata.exceptions import MarketingSiteAPIClientException, PersonToMarketingException
 from course_discovery.apps.course_metadata.people import MarketingSitePeople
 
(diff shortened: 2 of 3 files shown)


course_discovery/apps/api/v1/views/people.py
```diff
@@ -14,6 +14,16 @@
 logger = logging.getLogger(__name__)
 
 
+class IsStaff(permissions.BasePermission):
```
asadazam93: Can we move this to a permissions file?

attiyaIshaque: Sure @asadazam93. There is no permission file. I will create a new one and add it to the permissions file.

course_discovery/apps/api/permissions.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework.permissions import BasePermission
+
+
+class IsStaff(BasePermission):
+    """
+    Custom permission to only allow owners of the object.
+    """
+    def has_permission(self, request, view):
+        if request.method == 'GET':
```
schenedx, line 9: Why only check on GET requests? Why not all request methods?

schenedx (approved):
Just a minor question. Rest LGTM

course_discovery/apps/api/permissions.py
attiyaIshaque: If we add permissions on other requsets method then the course teams can't add or update the instructors because all course team are not having the staff rights.

schenedx: Interesting. Do you know if this change affects publisher viewing instructors? Could you please test that case?
If you are limiting the permission only to "GET" method, could you please update the name of the permission class to be "ReadByStaffOnly"?

schenedx (changes requested):

course_discovery/apps/api/permissions.py
attiyaIshaque: @schenedx yeah i have test that it does not affects the publisher viewing instructors. Course runs does not use this GET method.  CC:@asadazam93

And i will change the name of permission class.

attiyaIshaque:
@schenedx I have changed the name of permission class.

schenedx (approved):
👍

asadazam93 (approved):

PR #1454, opened 2018-07-05
Credit Pathway model

ColeLRogers:
Added a model for credit pathways, and a test for it.
Currently, rather than using a foreign key to an organization, a charfield is being used to represent an organization name until further details are fleshed out.


course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1274,20 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             super(Program, self).save(*args, **kwargs)
 
 
+class CreditPathway(TimeStampedModel):
+    """ Credit Pathway model """
+    name = models.CharField(max_length=255)
+    org_name = models.CharField(max_length=255, verbose_name=_("Organization name"))
+    email = models.EmailField(max_length=255)
+    programs = SortedManyToManyField(Program)
```
mikix: Does this need to be a sorted field? I'm not sure there's a concept of ordering here.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1274,20 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             super(Program, self).save(*args, **kwargs)
 
 
+class CreditPathway(TimeStampedModel):
+    """ Credit Pathway model """
+    name = models.CharField(max_length=255)
+    org_name = models.CharField(max_length=255, verbose_name=_("Organization name"))
+    email = models.EmailField(max_length=255)
```
mikix: The max length of an email is 254. And that's already the default for an EmailField.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1274,20 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             super(Program, self).save(*args, **kwargs)
 
 
+class CreditPathway(TimeStampedModel):
+    """ Credit Pathway model """
+    name = models.CharField(max_length=255)
+    org_name = models.CharField(max_length=255, verbose_name=_("Organization name"))
```
mikix: I might add a comment above this to just say "This field doesn't necessarily map to our normal org models, it's just a convenience field for pathways while we figure them out" or something like that.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1274,20 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+    email = models.EmailField(max_length=255)
+    programs = SortedManyToManyField(Program)
+
+    def __str__(self):
+        return "{org_name}: {name}".format(org_name=self.org_name, name=self.name)
+
+    class Meta(object):
+        unique_together = ('name', 'org_name')
```
mikix: We might want name to be unique all by itself. It should be able to be displayed in a list of these without needing the org_name is my brief understanding.

course_discovery/apps/course_metadata/models.py
ColeLRogers: I noticed all of the other many to many fields in this file were sorted, assumed it was the standard. I can switch if there's no need.

course_discovery/apps/course_metadata/models.py
ColeLRogers: I was assuming name could just be something like "Blended", and then multiple orgs would have that pathway type (name). I will update to make just name unique.

course_discovery/apps/course_metadata/models.py
mikix: I don't think it hurts anything to leave it sorted, except I think maybe certain operations you can do on a normal manytomany you can't do on a sorted? But if everything else here is sorted, it's fine to be cautious here and leave it sorted.

course_discovery/apps/course_metadata/models.py
mikix: Well, we originally were going to have just name. So I think the intent was for name to be unique enough to stand in a list by itself. The org_name addition I had suggested just as a backup safety in case we wanted to start using that somewhere. But I'm guessing 'name' will often include the org name too?

I'm leery of encoding assumptions here. But we have to one way or the other.

If we keep unique_together(name, org_name), we are committing ourselves to always using both fields in the UI and requiring an org_name. If we start with just name unique, we give ourselves some flexibility in how these fields are used. And can always switch to unique_together later.

course_discovery/apps/course_metadata/models.py
ColeLRogers: Fixed

course_discovery/apps/course_metadata/models.py
ColeLRogers: Done

course_discovery/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: edx-platform\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-07-05 12:03+0000\n"
+"POT-Creation-Date: 2018-07-06 12:57+0000\n"
```
dianakhuang, line 11: Were these translation files added intentional? It might make sense to leave them out and let the automated process for pulling these in do all the work.

dianakhuang (approved):
Just one question, but otherwise, 👍

mikix (approved):
:+1:

PR #1458, opened 2018-07-10
add filter fields for content_type course

muhammad-ammar:
[ENT-1047](https://openedx.atlassian.net/browse/ENT-1047)

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -141,6 +141,14 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
     uuid = indexes.CharField(model_attr='uuid')
     card_image_url = indexes.CharField(model_attr='card_image_url', null=True)
     org = indexes.CharField()
+
+    status = indexes.CharField(model_attr='course_runs__status')
+    start = indexes.DateTimeField(model_attr='course_runs__start', null=True)
+    end = indexes.DateTimeField(model_attr='course_runs__end', null=True)
+    enrollment_start = indexes.DateTimeField(model_attr='course_runs__enrollment_start', null=True)
+    enrollment_end = indexes.DateTimeField(model_attr='course_runs__enrollment_end', null=True)
+    availability = indexes.CharField(model_attr='course_runs__availability')
+
     course_runs = indexes.MultiValueField()
     expected_learning_items = indexes.MultiValueField()
 


muhammad-ammar:
@georgebabey I have added these fields to filter courses by courseruns fields. I have also tested these filters on Enterprise Customer Catalog admin interface and they are working as expected. Can you please take a quick look at the changes and see if things are in the correct direction.

muhammad-ammar:
@georgebabey sandbox is not updated yet. I will update it as soon as possible.

georgebabey (approved):

muhammad-ammar:
@mattdrayer I have added new filter fields to filter courses by courseruns. [Sandbox](https://business.sandbox.edx.org/admin/enterprise/enterprisecustomercatalog/81fba3bb-65f8-400d-871d-821d94608809/change/) is updated. Now we can use the below courserun fields/attributes to filter courses

- status  --- valid values are (`'published'`, `'unpublished'`)
- start
- end
- enrollment_start
- enrollment_end
- availability --- valid values are (`'Archived'`, `'Current'`, `'Starting Soon'`, `'Upcoming'`)

A sample query to get all courses having published courseruns looks like below
```python
{
  "content_type": "course",
  "status": "published"
}
```

Please test the new filters and let me know if these are what we needed?

zubair-arbi (approved):
Good solution. Once @mattdrayer is ok 👍

mattdrayer:
This is great!  Happy to see it was such a straightforward solution -- I tested on the sandbox by adding `"start__lte":"2018-04-30"` to the content filter, which dropped the total number of courses from 53 to 42.  Nice work!  👍

muhammad-ammar:
@georgebabey @mattdrayer @zubair-arbi I was thinking that filter fields should be prefixed with `courserun` to avoid confusion.

- courserun_status
- courserun_start
- courserun_end
- courserun_enrollment_start
- courserun_enrollment_end
- courserun_availability

Please share your thoughts? Should the fields be prefixed or it is clear that a filter field is referring to a courserun?

mattdrayer:
I'm fine with doing this -- agreed it would help to clarify which filter fields apply to which objects

zubair-arbi (approved):
After we have prefixed the new course runs filter fields with `courserun_` to avoid confusion 👍

muhammad-ammar:
@georgebabey @mattdrayer @zubair-arbi I did some more testing and it turns out that courses without courseruns are not filtered. As per my understanding a course should not be included in the catalog if it has not courserun. Looking into how to fix this.

muhammad-ammar:
@georgebabey @mattdrayer @zubair-arbi I am getting the courserun filtering issue on course [edX+123](https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs/81fba3bb-65f8-400d-871d-821d94608809/courses/edX+123/). This course has no course runs.

zubair-arbi:
@muhammad-ammar Can you please share the query? and also can you try to reproduce it with a course which have only one archived course run?

muhammad-ammar:
@zubair-arbi The below query will produce the `edX+123` course without courserun

```json
{
  "content_type":"course",
  "availability__not":[
    "Archived",
    "Current",
    "Starting Soon",
    "Upcoming"
  ]
}
```

muhammad-ammar:
@zubair-arbi This course has no existance in studio and ecommerce. Looks like that its course id was manually [here](https://discovery-business.sandbox.edx.org/admin/course_metadata/course/48/change/?_changelist_filters=q%3DedX%252B123).

muhammad-ammar:
@georgebabey @mattdrayer @zubair-arbi It turns out that for courserun prefixed fields we need to do more changes like adding the names of new fields [here](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/serializers.py#L1290) plus doing the testing again. So I am going with the unprefixed fields and merge this PR.

PR #1459, opened 2018-07-10
LEARNER-5665: Add credit pathway endpoint for the api.

dianakhuang:
Adding in an endpoint for pulling in a list of `CreditPathway` objects. Currently, this is a very bare bones endpoint only meant to be used by the `credentials` repo.

### DIFF
--- a/course_discovery/apps/api/v1/views/pathways.py
+++ b/course_discovery/apps/api/v1/views/pathways.py
@@ -1,0 +1,14 @@
+""" Views for accessing CreditPathway data """
+from rest_framework import viewsets
+from rest_framework_extensions.cache.mixins import CacheResponseMixin
+
+from course_discovery.apps.api import serializers
+from course_discovery.apps.api.permissions import ReadOnlyByPublisherUser
+
+from course_discovery.apps.course_metadata.models import CreditPathway
+
+
+class CreditPathwayViewSet(CacheResponseMixin, viewsets.ReadOnlyModelViewSet):
+    permission_classes = (ReadOnlyByPublisherUser,)
+    serializer_class = serializers.CreditPathwaySerializer
+    queryset = CreditPathway.objects.all()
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/v1/views/pathways.py
```diff
@@ -0,0 +1,14 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework_extensions.cache.mixins import CacheResponseMixin
+
+from course_discovery.apps.api import serializers
+
+from course_discovery.apps.course_metadata.models import CreditPathway
+
+
+class CreditPathwayViewSet(CacheResponseMixin, viewsets.ReadOnlyModelViewSet):
+    permission_classes = (IsAuthenticated,)
```
mikix: Just to confirm - we feel like this is enough protection? I'm still thinking that because we're spitting out emails, we might want to be more cautious.

Like, what protections does the LMS's /api/user/v1/accounts/{username} endpoint have?

mikix (commented):
Looks good. Nice and simpler than I expected. One question about access.

course_discovery/apps/api/v1/views/pathways.py
dianakhuang: I think the question here is what sort of permissions are needed to be able to authenticate against course discovery. I can lock things down to publisher users to be on the safe side here.

ColeLRogers (approved):
👍

course_discovery/apps/api/v1/views/pathways.py
dianakhuang: I also just tested this on devstack, and by default, the credentials_worker user doesn't have access to the API, so we'll have to make modifications on prod if we want to use these protections.

mikix: Probably worth it? Like, this is PII-adjacent stuff, right? We don't want randos to query it from our API.

PR #1460, opened 2018-07-11
New Course Run must not be added to associated retired programs.

uzairr:
### Description
Project Coordinators need to be able to exclude all future runs of courses in a program, so that they can convert an XSeries program into a Professional Certificate (or into a MicroMasters), without all future course runs applying to both the XSeries program and the Professional Certificate.

- [x] Unit Tests

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -770,6 +770,7 @@ class CourseRun(TimeStampedModel):
         return '{key}: {title}'.format(key=self.key, title=self.title)
 
     def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        is_new_course_run = not self.pk
         suppress_publication = kwargs.pop('suppress_publication', False)
         is_publishable = (
             self.course.partner.has_marketing_site and
@@ -793,6 +794,11 @@ class CourseRun(TimeStampedModel):
         else:
             logger.info('Course run [%s] is not publishable.', self.key)
             super(CourseRun, self).save(*args, **kwargs)
+
+        if is_new_course_run:
+            retired_programs = self.programs.filter(status=ProgramStatus.Retired)
+            for program in retired_programs:
+                program.excluded_course_runs.add(self)
 
 
 class SeatType(TimeStampedModel):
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -794,6 +794,11 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             logger.info('Course run [%s] is not publishable.', self.key)
             super(CourseRun, self).save(*args, **kwargs)
 
+        retired_programs = self.course.programs.filter(status=ProgramStatus.Retired)
```
waheedahmed: You should make sure that this code only executes on new course run creation, otherwise it will exclude old course runs as well upon updating.

waheedahmed (changes requested):

course_discovery/apps/course_metadata/models.py
staubina: With the method being named save I think this is a valid concern here.

uzairr: good catch @waheedahmed .Now, I have added a check for pk which will ensure that newly added code will be executed only for newly created courseruns

course_discovery/apps/course_metadata/models.py
```diff
@@ -778,6 +778,7 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             not suppress_publication
         )
 
+        is_new_course_run = not self.pk
```
waheedahmed: Move this to top of the function.

course_discovery/apps/course_metadata/models.py
```diff
@@ -794,6 +795,11 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             logger.info('Course run [%s] is not publishable.', self.key)
             super(CourseRun, self).save(*args, **kwargs)
 
+        if is_new_course_run:
+            retired_programs = self.course.programs.filter(status=ProgramStatus.Retired)
```
waheedahmed: CourseRun model have a property `programs`, use that `self.programs.filter(status=ProgramStatus.Retired)`.

waheedahmed (approved):
Some minor changes otherwise :+1:

staubina (approved):
Please make @waheedahmed's recommended changes then it should be ready to merge.

waheedahmed (approved):

PR #1461, opened 2018-07-12
Yro add degree submodels

yro:
### Description
Implement simple `Degree` / `DegreeMarketing` model for future masters / masters marketing site API connectivity.

### Testing
- [ ] Unit tests

### Reviewers
- [ ] Code review: @schenedx @iloveagent57 

FYI: @edx/educator-neem

### Post-review


course_discovery/apps/course_metadata/models.py
```diff
@@ -1129,6 +1128,24 @@ def entitlements(self):
...
+
+    @property
+    def degree_marketing_data(self):
+        """
+        Return Degree Marketing Information
+        """
+        try:
+            degree_marketing = DegreeMarketing.objects.get(degree=self.associated_degree)
+            return degree_marketing.uuid
+        except DegreeMarketing.DoesNotExist:
```
yro: This is meant as kind of a validation -- ultimately either this will be its own endpoint or call the various and sundry `DegreeMarketing` fields as an addition to the `Program` model for rendering by the future react template.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1291,36 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+    """
+    Marketing model for degree landing pages
+    """
+    degree = models.OneToOneField(
+        Degree,
+        on_delete=models.CASCADE,
+        primary_key=True,
+    )
+    uuid = models.UUIDField(blank=True, default=uuid4, editable=False, unique=True, verbose_name=_('UUID'))
+    title = models.CharField(
```
schenedx: Both `title` and `subtitle` are common fields between `Program` and `DegreeMarketing`. Are you thinking Degree fields like these should serve as overrides? Or should they be limited to unique fields only related to Degree?
If it's the latter, maybe find a unique field like `application_deadline` or `apply_url` to be the initial field?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1129,6 +1128,24 @@ def entitlements(self):
...
+            return
+
+    @property
+    def degree_marketing_data(self):
+        """
+        Return Degree Marketing Information
+        """
+        try:
+            degree_marketing = DegreeMarketing.objects.get(degree=self.associated_degree)
+            return degree_marketing.uuid
```
schenedx: Instead just returning the `uuid` field here, can you create a new serializer for `DegreeMarketing` model and use that serializer in the MinimalProgramsSerializer?
This is the pattern DRF promotes. Keep the selection of fields to be visible to API in serializer.

schenedx (changes requested):
Most of the work is there done.
I do feel this can be further improved.

course_discovery/apps/course_metadata/models.py
yro: @schenedx I don't disagree, but it's still unclear whether or not this is an override. I'll change these fields so that there's not any confusion.

course_discovery/apps/course_metadata/models.py
yro: @schenedx agree. Will do.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1273,43 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+class DegreeMarketing(TimeStampedModel):
+    """
+    Marketing model for degree landing pages
+    """
+    degree = models.OneToOneField(
+        Degree,
+        on_delete=models.CASCADE,
+        primary_key=True,
+    )
+    uuid = models.UUIDField(blank=True, default=uuid4, editable=False, unique=True, verbose_name=_('UUID'))
```
schenedx: I don't yet see how we can use this field anywhere.
Let's take this out, and will add it back in when there is a use?

schenedx (approved):
This is much better.
I think we can work to merge it asap.

schenedx:
@mikedikan @bderusha FYI
Here is our prototype for the degree model to host Master data in discovery.
If you have strong opinions on this PR, please come and talk to us. We are faced with tight deadline and would like to iterate as quickly and frequently as possible.

course_discovery/apps/course_metadata/models.py
yro: 👍

course_discovery/apps/course_metadata/models.py
```diff
@@ -1274,6 +1273,43 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
             super(Program, self).save(*args, **kwargs)
 
 
+class Degree(TimeStampedModel):
+    """
+    Degree 'master' model linking requirements and templates
+    """
+    name = models.CharField(max_length=255, unique=True)
+    program = models.OneToOneField(
```
iloveagent57, line 1281: I realize this is already merged, and we can get to it later: this should really be a many-to-many relationship: https://docs.djangoproject.com/en/2.0/topics/db/examples/many_to_many/ with an intermediate table between Degree and Program like `ProgramDegreeMembership`.  This is what enables a Degree to consist of many programs, and for a program to be a part of many different Degrees.

yro: @iloveagent57 Caught up IRL, but absolutely. This model is just a placeholder, and is wide open; I agree -- do your thing.

PR #1468, opened 2018-07-20
Improving catalog typeahead search results

adeelkhan:
This patch would enable search results to
work for complete string too. Now it
would show same results for search string
'chemistry' as 'chemistr'

LEARNER-3051

### DIFF
--- a/course_discovery/settings/synonyms.py
+++ b/course_discovery/settings/synonyms.py
@@ -63,4 +63,5 @@ SYNONYMS = [
     ['vba', 'excel'],
     ['usa', 'united states of america', 'america', 'murika'],
     ['virtual reality', 'VR'],
+    ['chemistri', 'chemistry']
 ]


MatthewPiatetsky:
👍

course_discovery/settings/synonyms.py
```diff
@@ -63,4 +63,5 @@
     ['vba', 'excel'],
     ['usa', 'united states of america', 'america', 'murika'],
     ['virtual reality', 'VR'],
+    ['chemistri', 'chemistry']
```
robrap, line 66: I'm not clear:
1. Why we needed/wanted this synonym, and
2. How it has to do with fixing the issue described, where `chemistr` (no 'i' or 'y'), does not bring up results.
FYI: @marcotuts: I agree with your note on the ticket.

PR #1471, opened 2018-07-23
Add the degree curriculum tables and fields.

iloveagent57:
This PR adds the many-to-many relationship between degrees and programs/courses via the `DegreeProgramCurriculum` and `DegreeCourseCurriculum` intermediate tables.  Also adds some django admin classes for these tables.

degree-admin-screen


schenedx (approved):
👍

PR #1475, opened 2018-07-25
Tie pathways to a partner

mikix:
I chose to go with tying pathways to partners over sites because that's how the rest of the course_metadata models did it. Partners and Sites are a 1-to-1 relationship anyway.

It's a *little* confusing because we also have org_name (which is for the pathway org) but that same confusion exists on, for example, Programs where authoring_organizations co-exists with partner.
 
https://openedx.atlassian.net/browse/LEARNER-5936


spencerhance (commented):
Looks good!

PR #1477, opened 2018-07-27
Update the decision record for the Design of degrees.

iloveagent57:
I updated this ADR to reflect the design that we're about to merge.  I'd like to get a diagram in here at some point, but didn't have time to make a nice one yesterday.


schenedx (approved):
👍

PR #1478, opened 2018-07-27
EDUCATOR-3227: Discovery feeds "Details" to Master's Product Page

sandroroux:
[EDUCATOR-3227](https://openedx.atlassian.net/browse/EDUCATOR-3227)


course_discovery/apps/api/serializers.py
```diff
@@ -733,19 +732,40 @@ class Meta:
...
+
+
 class DegreeSerializer(serializers.ModelSerializer):
     """ Degree model serializer """
     campus_image_mobile = serializers.ImageField()
     campus_image_tablet = serializers.ImageField()
     campus_image_desktop = serializers.ImageField()
+    deadlines = DegreeDeadlineSerializer(many=True)
     rankings = RankingSerializer(many=True)
+    tuition_costs = DegreeCostSerializer(many=True)
```
schenedx: #NIT maybe the property name should just be called "costs"? That's simple and easy to understand.

schenedx (changes requested):
You need migration file in this PR.

schenedx (changes requested):
You need the changes to the `DegreeFactory` in factories.py. Those new fields needs to be there and generated for tests to work.
The model in general looks good. Can you put screenshots of how Django Admin interface look on this PR?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1291,15 +1291,29 @@ def __str__(self):
...
 class Degree(Program):
     """
     This model captures information about a Degree (e.g. a Master's Degree).
     It mostly stores information relevant to the marketing site's product page for this degree.
     """
-    application_deadline = models.CharField(
-        help_text=_('String-based deadline field (e.g. FALL 2020)'),
-        max_length=255,
-    )
+    deadlines = SortedManyToManyField(DegreeDeadline, blank=True)
```
iloveagent57: I'm not sure these should be many-to-many relationships.  It could lead to cases where an existing deadline, used by two different degrees, is modified for the sake of one degree, which means the other degree will now show that updated deadline, too.  This isn't that far-fetched a scenario, since many schools use very similar or identical deadlines for their programs (e.g. January 1st).  It might be better to have `degree` be a FK field of the DegreeDeadline model.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1291,15 +1291,29 @@ def __str__(self):
...
+    costs = SortedManyToManyField(DegreeCost, blank=True)
```
iloveagent57: I think the same argument as for `DegreeDeadline` above holds here.

iloveagent57 (dismissed):
I think using a M2M relationship is not the right thing for these models.  Also, I'm a little wary of us over-using it in general (e.g. for Rankings, which are less likely to run into the issue I've described here, but are still susceptible to it).

course_discovery/apps/api/serializers.py
```diff
@@ -748,13 +768,17 @@ class DegreeSerializer(serializers.ModelSerializer):
     campus_image_desktop = serializers.ImageField()
     rankings = RankingSerializer(many=True)
     curriculum = CurriculumSerializer()
+    prerequisite_coursework = serializers.CharField()
```
schenedx: Why do we want to serialize this to `CharField`? Why not use the default `TextField`, the same as with the model?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -675,23 +677,30 @@ def test_degree_marketing_data(self):
...
             'campus_image_mobile': degree.campus_image_mobile,
             'campus_image_tablet': degree.campus_image_tablet,
             'campus_image_desktop': degree.campus_image_desktop,
             'curriculum': expected_curriculum,
-            'rankings': expected_rankings
+            'rankings': expected_rankings,
+            'deadlines': expected_degree_deadlines,
+            'costs': expected_degree_costs,
         }
+        self.maxDiff = None
```
schenedx: You would want these two lines to be removed

schenedx (changes requested):
Please squash all 3 of your migrations into 1

schenedx (approved):
👍

course_discovery/apps/course_metadata/migrations/0098_degree_lead_capture_image.py
```diff
@@ -10,7 +10,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('course_metadata', '0096_degree_lead_capture_list_name'),
```
schenedx, line 13: This is a super bad change!
Because of this update, our deployment to stage for this commit fails. The django admin for Degrees model is now throwing 500 errors. I need to think of a way to revert this change cleanly

PR #1481, opened 2018-07-30
Add pytest-xdist, use multiple procs to run tests.

iloveagent57:
This should speed up tests locally (with docker configured to use multiple CPUs) and on travis (where the workers have 2 cores available).  It reduces the total runtime of unit tests in Travis from around 25 minutes to about 20 minutes (20% improvement).  Changes include:

- Installing `pytest-xdist` and setting the options `-n2` and `--dist=loadscope`.  This will make the tests always use 2 CPUs and will distribute tests in the same class to the same worker.
- Makes the `django_cache` pytest fixture not call `cache.clear()` after every test.  Instead, we clear the cache (for the purposes of testing) once, at the beginning of the test run.
- Adds a hard-coded cache key for currency data (accessed via a function, because the `@cache_response` decorator wants a callable to define the cache key).
- Adds a serparate, in-memory cache, when running tests, for testing user rate throttling.
- Adds a `FuzzyAssertNumQueriesMixin` class for use in the tests of views (everything under `course_discovery/apps/api/v1/tests/test_views/`).  This makes it so that `assertNumQueries()` returns True as long as the expected count is within 2 of the actual count (the threshold of 2 is a default value to the overridden method).

Reasoning for that last point:
I worked with Jeremy about getting distributed tests to jibe with the fact that a couple of waffle switches are looked up and then cached in two different processes (not to mention the other places where we clear the cache willy-nilly).  My original attempt to work around this issue was a decorator that would prefetch the switch values.  I then wrapped any test functions that made query count assertions and also used those switches with that decorator.  However, this never worked 100% consistently, for reasons that are unclear to me.  This change helps comply with the spirit of `assertNumQueries()` without ripping it out entirely.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_currency.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_currency.py
@@ -1,23 +1,26 @@
 import mock
 import pytest
+from django.core.cache import cache
 from django.urls import reverse
 
-from course_discovery.apps.api.v1.views.currency import CurrencyView
+from course_discovery.apps.api.v1.views.currency import CurrencyView, exchange_rate_cache_key
 
 
-# NOTE: All of the tests that make authenticated requests result in the response being cached. We
-# force the use of a fixture here to ensure the cache is cleared after each test.
 @pytest.mark.usefixtures('django_cache')
 @pytest.mark.django_db
 class TestCurrencyCurrencyView:
     list_path = reverse('api:v1:currency')
 
+    def setup_method(self, method):  # pylint: disable=unused-argument
+        cache.delete(exchange_rate_cache_key())
+
     def test_authentication_required(self, client):
         response = client.get(self.list_path)
         assert response.status_code == 403
 
-    def test_get(self, admin_client, django_cache, responses, settings):
+    def test_get(self, admin_client, django_cache, responses, settings):  # pylint: disable=unused-argument
         settings.OPENEXCHANGERATES_API_KEY = 'test'
+
         rates = {
             'GBP': 0.766609,
             'CAD': 1.222252,
@@ -34,6 +37,7 @@ class TestCurrencyCurrencyView:
         responses.add(responses.GET, CurrencyView.EXTERNAL_API_URL, json={'rates': rates})
 
         response = admin_client.get(self.list_path)
+
         assert all(item in response.data.items() for item in expected.items())
         assert len(responses.calls) == 1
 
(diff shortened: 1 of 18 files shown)


course_discovery/apps/api/v1/tests/test_views/test_currency.py
```diff
@@ -42,12 +46,6 @@ def test_get(self, admin_client, django_cache, responses, settings):
         assert all(item in response.data.items() for item in expected.items())
         assert len(responses.calls) == 1
 
-        # Clearing the cache should result in the external service being called again
-        django_cache.clear()
-        response = admin_client.get(self.list_path)
-        assert all(item in response.data.items() for item in expected.items())
-        assert len(responses.calls) == 2
```
schenedx: Is this an assert we have to remove? Or this assert gets moved somewhere else?

schenedx (approved):
One minor question. Rest LGTM. This is some skilled python ninjutsu.

course_discovery/apps/api/v1/tests/test_views/test_currency.py
iloveagent57: That deletion is a vestige from before I created a constant cache key for currencies.  I've added this section back in.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -89,16 +87,13 @@ def test_authentication(self):
...
+
+        with django_assert_num_queries(FuzzyInt(57, 2)):
             response = self.assert_retrieve_success(program)
         # property does not have the right values while being indexed
         del program._course_run_weeks_to_complete
         assert response.data == self.serialize_program(program)
 
-        # Verify that repeated retrieve requests use the cache.
-        with django_assert_num_queries(4):
-            self.assert_retrieve_success(program)
```
jmbowman, line 100: Why were the checks here and below regarding cache utilization removed?

jmbowman (approved):
Just one question, otherwise 👍

course_discovery/apps/api/v1/tests/test_views/test_programs.py
iloveagent57: We incorporate the timestamp of the last change to a model in our cache keys: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/cache.py

This is fine when the tests are run in series, but when they are distributed, a model in a different test class could change, which makes the key, value pair cached from the first call (on line 92) useless to the next call.  I tried building a locking mechanism using memcached, but it never worked quite as I intended, and I didn't want to introduce that level of logic for the sake of a testing change.  We might be able to work around it by introducing a different cache key class that's scoped to test settings.

iloveagent57: ...although that could end up breaking other tests in a different way.

PR #1482, opened 2018-07-30
Add RANKING model to the Degrees

schenedx:
Add ranking model so we can serve the brag bar on master about page.
@iloveagent57 @sandroroux Please review
rankingsbox


course_discovery/apps/course_metadata/models.py
```diff
@@ -1315,6 +1327,8 @@ class Degree(Program):
         help_text=_('Provide a campus image to display on desktop displays'),
     )
 
+    rankings = SortedManyToManyField(Ranking, blank=True)
```
iloveagent57, line 1330: Maybe consider adding a `related_name=degrees` to this field?

iloveagent57 (approved):
LGTM

course_discovery/apps/course_metadata/models.py
schenedx: I think that's redundant. There is no need for us to refer to the degree model from the ranking model at the moment. If we do have that need, we can add that later.

PR #1483, opened 2018-07-31
Added search option in CourseEntitlementAdmin view

Ayub-Khan:
Added search option in CourseEntitlementAdmin view.
CourseEntitlementAdmin can not be used with out a search option as there are 706 entries and no way to find the one you are looking for.

### DIFF
--- a/course_discovery/apps/publisher/admin.py
+++ b/course_discovery/apps/publisher/admin.py
@@ -120,8 +120,9 @@ class SeatAdmin(SimpleHistoryAdmin):
 
 @admin.register(CourseEntitlement)
 class CourseEntitlementAdmin(SimpleHistoryAdmin):
-    list_display = ['course', 'mode']
+    list_display = ['course', 'course__number', 'mode']
     raw_id_fields = ['course']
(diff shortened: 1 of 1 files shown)


hamzamunir7300 (approved):
LGTM :+1:

schenedx:
@Ayub-Khan Did you test this? See https://stackoverflow.com/questions/163823/can-list-display-in-a-django-modeladmin-display-attributes-of-foreignkey-field
The way you have don't work according to the knowledge article above

course_discovery/apps/publisher/admin.py
```diff
@@ -120,8 +120,9 @@ class SeatAdmin(SimpleHistoryAdmin):
 
 @admin.register(CourseEntitlement)
 class CourseEntitlementAdmin(SimpleHistoryAdmin):
-    list_display = ['course', 'mode']
+    list_display = ['course', 'course__number', 'mode']
```
schenedx, line 123: list_display here does follow foreign key reference at all.

Ayub-Khan:
ah my bad. i was under the impression it does, because of search_fields.

PR #1485, opened 2018-08-01
Added search option for course entitlements

Ayub-Khan:
Added search option for course entitlements using course name and course number.
CourseEntitlementAdmin can not be used with out a search option as there are 706 entries and no way to find the one you are looking for.

all changes have been tested locally.


hamzamunir7300 (approved):
LGTM :+1:

PR #1488, opened 2018-08-03
Allow email field to be blank or null

dianakhuang:
We want to allow the email field to be null or blank so that we can configure 'disabled' pathways without email addresses. 

https://openedx.atlassian.net/browse/LEARNER-6048


course_discovery/apps/course_metadata/models.py
```diff
@@ -1378,7 +1378,7 @@ class CreditPathway(TimeStampedModel):
     # this field doesn't necessarily map to our normal org models, it's just a convenience field for pathways
     # while we figure them out
     org_name = models.CharField(max_length=255, verbose_name=_("Organization name"))
-    email = models.EmailField()
+    email = models.EmailField(null=True, blank=True)
```
bbaker6225: I think we just want blank=True, because otherwise we will have 2 different values for an empty email which might cause issues down the line.  See [Django null ref](https://docs.djangoproject.com/en/2.0/ref/models/fields/#null).

dianakhuang: Yeah, I wasn't sure about that. It's been updated.

bbaker6225 (approved):

PR #1489, opened 2018-08-03
Yro/EDUCATOR-3226

yro:
### Description
Implement 'Quick Facts' model for marketing site

### Testing
- [ ] Unit tests

### Reviewers
- [ ] Code review: @iloveagent57 

FYI: @edx/educator-neem

### Post-review


course_discovery/apps/course_metadata/models.py
```diff
@@ -1328,6 +1336,7 @@ class Degree(Program):
     )
 
     rankings = SortedManyToManyField(Ranking, blank=True)
+    quick_facts = SortedManyToManyField(IconTextPairing, blank=True)
```
iloveagent57, line 1339: Having a M2M relationship here means the following can happen: degree A and degree B both use IconTextPairing 1.  Someone updates the pairing with a new icon at the request of degree B's institution.  Now that icon is used for both degree A and degree B.  It's probably better to have `degree` by a FK field on the `IconTextPairing` model.
(I just left a comment to this effect on Alessandro's PR, too)

iloveagent57 (changes requested):

yro:
Closing PR.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1291,6 +1291,14 @@ def __str__(self):
         return self.description
 
 
+class IconTextPairing(TimeStampedModel):
+    """
+    Represents an icon:text model
+    """
+    icon = models.CharField(max_length=100, verbose_name=_('Icon URL or fa class'))
```
schenedx, line 1298: Having `CharField` here means the data entry person from Marketing needs to know what to enter as the Icon URL. That may not be information they have.
I think a way to provide convenience for marketing user is to create a new table that contains all the FA class options. Then we in engineering fill out the urls for each FA icon or class in the new table. 
This means, on the `IconTextPairing`, it will be just a drop down for Marketing user to select.

schenedx: That said, if you think we should consider my suggestion as an enhancement to this, please create a ticket for our project, and we can work on this later.

yro:
@schenedx @iloveagent57 Thank you for your suggestions, since there are so many requests with downstream effects, I'm going to start a fresh branch.

PR #1492, opened 2018-08-06
Add a new column to Degree to capture Sailthru list name

schenedx:
@yro @sandroroux @iloveagent57 Please review
This is to add the Sailthru list name into the Django Admin for the lead capture form.
All the other value of the sailthru lead capture form can be constructed or hard coded from the existing set of data

leadcaptureapi

leadcapture


course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -348,6 +348,7 @@ class Meta(object):
     application_deadline = FuzzyText()
     apply_url = FuzzyURL()
     overall_ranking = FuzzyText()
+    lead_capture_list_name = FuzzyText()
```
iloveagent57, line 351: You may want to add a line [here](https://github.com/edx/course-discovery/blob/b4311d61ea33a2d9e426e1f61503c47ed2708684/course_discovery/apps/course_metadata/tests/test_models.py#L1345) to test that this field isn't null on a newly created Degree.

iloveagent57 (commented):
Looks good, one minor point.

course_discovery/apps/course_metadata/tests/factories.py
schenedx: @iloveagent57 Done. Thank you!

iloveagent57 (approved):

PR #1493, opened 2018-08-06
yro/EDUCATOR-3226

yro:
### Description
Add `quick_facts` section to degree model as child model.

### Testing
- [x] Unit tests

### Reviewers
- [x] Code review: @schenedx 

FYI: @edx/educator-neem

### Post-review


course_discovery/apps/course_metadata/admin.py
```diff
@@ -320,6 +320,13 @@ class CurriculumAdminInline(admin.StackedInline):
     extra = 1
 
 
+class IconTextPairingInline(admin.StackedInline):
+    model = IconTextPairing
+    extra = 3
```
schenedx, line 325: Why extra 3 here? Is it just to show 3 different empty items on admin UI?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1336,6 +1336,21 @@ def __str__(self):
         return str('Degree: {}'.format(self.title))
 
 
+class IconTextPairing(TimeStampedModel):
+    """
+    Represents an icon:text model
+    """
+    degree = models.ForeignKey(Degree, related_name='quick_facts', on_delete=models.CASCADE)
+    icon = models.CharField(max_length=100, verbose_name=_('Icon URL or fa class'))
```
schenedx: My comment on the new table for all hard coded path for fa icons are not in this PR. I am OK with that. If we want, we can work on that in a new ticket.

schenedx (approved):
Looks good in general. A minor question only.

course_discovery/apps/course_metadata/admin.py
yro: @schenedx Yes, since that's sort of the aesthetic ideal here (less or more is fine, of course, but three seems to be the ideal).

course_discovery/apps/course_metadata/models.py
yro: @schenedx I had a couple of questions about that -- did you mean *all* of the fa icons? There are quite a few, and they change from time to time
https://fontawesome.com/cheatsheet
Or:
Did you mean the ones that have been requested so far -- the latter plan involves a code change for new icons. 
I could do a subset of about 15-20 of useful ones to start, and we could build from there if you want. It's all pretty easy to do, I'm happy to get this in this PR.

schenedx: Yeah, what you said about a subset of useful ones is a good start.
If you think it's going to be easy, by all means. But I think it will also be easy if we make that change later. It's not as important as giving our users fields to input data into course-discovery for a new master about page.

yro:
@schenedx This should be more what you're talking about, right?

schenedx (commented):
It's partly there. I was more thinking that the different `fa` icon choices will be stored in their own new table, so anyone who wants to insert a new fa icon can just update the table, and have that be a new drop down item in the `IconTextParing`. This works tho. We should merge 👍

PR #1495, opened 2018-08-08
EDUCATOR-3238 | Add a lead capture image field.

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-3238

This adds a `lead_capture_image` standardized image field to the `Degree` model. It updates the Degree admin page with the field and exposes it via the programs API.


schenedx (approved):
👍

course_discovery/apps/course_metadata/models.py
```diff
@@ -1335,6 +1335,18 @@ class Degree(Program):
...
     )
+    lead_capture_image = StdImageField(
+        upload_to=UploadToAutoSlug(populate_from='uuid', path='media/degree_marketing/lead_capture_images/'),
+        blank=True,
+        null=True,
+        variations={
+            'large': (1440, 480),
+            'medium': (726, 242),
+            'small': (435, 145),
+            'x-small': (348, 116),
```
schenedx, line 1346: Do we use the `x-small` size at all? If we don't do that, should we remove this variation?

PR #1496, opened 2018-08-09
Add micromasters descriptive fields

yro:
### Description
Add micromasters descriptive fields for template rendering

### Testing
- [ ] Unit tests

### Reviewers
- [ ] Code review:

FYI: @edx/educator-neem

### Post-review


course_discovery/apps/course_metadata/models.py
```diff
@@ -1336,6 +1336,25 @@ class Degree(Program):
...
+    )
+    micromasters_long_title = models.CharField(
+        help_text=_('Micromasters verbose title'),
+        max_length=255,
+        blank=True,
+        null=True
+    )
+    micromasters_long_description = models.CharField(
+        help_text=_('Micromasters descriptive paragraph'),
+        max_length=255,
```
schenedx: Are you sure max_length 255 is going to be long enough for the descriptive paragraph? What if we just updated this to be `TextField`?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1336,6 +1336,25 @@ class Degree(Program):
         null=True,
     )
 
+    micromasters_url = models.CharField(
```
schenedx: Do you think there is value to use `UrlField` instead of `CharField` here? I don't have a strong opinion on this one.

schenedx (commented):
I think you would also need to run `make fake_translations` with this PR to get the translation text in.

course_discovery/apps/course_metadata/models.py
yro: Yeah makes sense. I kind of got the feeling that whatever I chose would be wrong. I'll update it.

course_discovery/apps/course_metadata/models.py
yro: Smart, wil do.

schenedx (approved):
👍

yro:
@schenedx the BG image is not templatized.

schenedx:
@yro Sounds good. Thank you

PR #1497, opened 2018-08-10
Add max length validation

awaisdar001:
I want to fix and sync course metadata model `Course` and publisher model `Course`. The issue is, when we `publish` a course from publisher, it copies the data to course metadata course model and it is very important that both models should be in sync in terms of `max_length` validations. 

I have removed the old `validate_text_count` validator b/c the approch it was using doesn't make any sense to count text only when you are saving the html in the model. Instead, I used Djangos' own `MaxLengthValidator`. 

- Course Metadata Model ([Course](https://github.com/edx/course-discovery/blob/b4e864145902f4b21ce54a96aca074a8451c4e42/course_discovery/apps/course_metadata/models.py#L337)): 
- Publisher Model ([Course](https://github.com/edx/course-discovery/blob/2c7ebc6aafa92907e30e6544ecdc5e98bbb6e372/course_discovery/apps/publisher/models.py#L44)):

[EDUCATOR-3315](https://openedx.atlassian.net/browse/EDUCATOR-3315)


course_discovery/apps/publisher/forms.py
```diff
@@ -75,22 +76,24 @@ class CourseForm(BaseForm):
         label=_('Organization Name'),
         required=True
     )
-    title = forms.CharField(label=_('Course Title'), required=True)
+    title = forms.CharField(label=_('Course Title'), required=True, validators=[MaxLengthValidator(255)])
```
awaisdar001: Note for reviewers: I add `255` length constraint to match with [course metadata course model title](https://github.com/edx/course-discovery/blob/b4e864145902f4b21ce54a96aca074a8451c4e42/course_discovery/apps/course_metadata/models.py#L345) constraint.

course_discovery/apps/publisher/forms.py
```diff
@@ -75,22 +76,24 @@ class CourseForm(BaseForm):
...
         label=_('Course Number'), required=True,
-        validators=[validate_text_count(max_length=50)]
+        validators=[MaxLengthValidator(50)]
     )
     short_description = forms.CharField(
         label=_('Short Description'),
-        widget=forms.Textarea, required=False, validators=[validate_text_count(max_length=255)]
+        widget=forms.Textarea,
+        required=False,
+        validators=[MaxLengthValidator(350)]
```
awaisdar001, line 90: Same for this one. Matching with  [course metadata course model short_description](https://github.com/edx/course-discovery/blob/b4e864145902f4b21ce54a96aca074a8451c4e42/course_discovery/apps/course_metadata/models.py#L346) constraint.

course_discovery/static/js/publisher/tinymce-charactercount.plugin.js
```diff
@@ -30,16 +30,7 @@ tinymce.PluginManager.add('charactercount', function (editor) {
   });
 
   _self.getCount = function () {
-    var tx = editor.getContent({ format: 'raw' });
-    var decoded = decodeHtml(tx);
-    var decodedStripped = decoded.replace(/(<([^>]+)>)/ig, "").trim();
```
awaisdar001, line 35: Similarly, Do not need to remove `html` elements to show counts, be explicit to the user about how much the data length is.

course_discovery/static/js/publisher/tinymce-init.js
```diff
@@ -77,6 +77,9 @@ $(document).ready(function () {
                     }
                 }
             });
+            editor.on('GetContent', function (e) {
```
awaisdar001, line 80: Basically, Just to match the Django `max_length` constraint I removed `\n` and `\r` from the content.

noraiz-anwar (approved):
Nice cleanup! 
👍

asadazam93 (approved):
Looks good

PR #1498, opened 2018-08-10
Add uuid to CreditPathway model

ColeLRogers:
[LEARNER-6100](https://openedx.atlassian.net/browse/LEARNER-6100)

4 migrations for adding the uuid field:
1) Adding the uuid field with null = True
2) Assigning a uuid to each existing pathway
3) Changing null = True to unique = True
4) Removing the previous unique_together constraint 

Still keeping the id field in the serializer since the lms copy_catalog command uses it, we would have to change that first before we remove it here.

I also made pathways searchable in Django admin while I was at it.


course_discovery/apps/course_metadata/migrations/0099_auto_20180809_1821.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.15 on 2018-08-09 18:21
+from __future__ import unicode_literals
+
+import uuid
+from django.db import migrations
+
+from course_discovery.apps.course_metadata.models import CreditPathway
```
bbaker6225: Because this uses CreditPathway in  a seeding migration, we'll have to squash migrations before renaming the CreditPathway model as part of [LEARNER-6088](https://openedx.atlassian.net/browse/LEARNER-6088).

What's wrong with just starting off with uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True, verbose_name='UUID')?  If you do this, I believe you can cut it down to 1 migration to handle everything.

ColeLRogers: If you start off with that, it attempts to use the same uuid for every object (it only gets generated once), which of course breaks the uniqueness constraint.

I followed this as a guide: https://docs.djangoproject.com/en/1.11/howto/writing-migrations/#migrations-that-add-unique-fields

bbaker6225: Ok good to know.  

Anyway, using CreditPathway in a seeding migration still causes issues down the line for LEARNER-6088.  It's possible we might want to wait for the CreditPathway model to be renamed.  Alternatively, we can try to handle squashing migrations as part of LEARNER-6088, but last time we tried that @mikix and I were still seeing issues.

ColeLRogers: DevOps recommends sticking with completing this story first, they can assist with doing a hard squash when the time comes in LEARNER-6088.

bbaker6225: If that's what devops thinks is best, I'm fine with that.  To potentially make our lives easier, can we first ask about using `apps.get_model` for the historical context instead of a direct import , which I believe could actually remove this issue entirely.
You can find more info here: https://docs.djangoproject.com/en/2.1/ref/migration-operations/#runpython

ColeLRogers:
Latest commit updates to using apps.get_model(), which will make renaming this model in the future much easier.
I also added a no-op value to RunPython's reverse_code argument, such that the migration can be reversed successfully.

bbaker6225 (approved):
👍

dianakhuang (approved):

PR #1507, opened 2018-08-15
mroytman/remove publisher pacing type

michaelroytman:


michaelroytman:
@schenedx @bderusha Could either of you take a look at this PR since you have the domain knowledge about Publisher and course-discovery? Please let me know if you have any questions. Thank you!

CC: @fysheets

course_discovery/apps/core/tests/test_api_clients.py
```diff
@@ -198,3 +198,20 @@ def test_get_api_access_request_with_multiple_records(self, mock_access_token):
...
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    def test_get_course_details(self, mock_access_token):  # pylint: disable=unused-argument
+        """
+        Verify that `get_course_details` returns correct value.
+        """
+        self.response = {
+            'pacing_type': 'self_paced'
+        }
+
+        fake_lms_url = "test"
```
schenedx: The variable name is confusing. This is the `fake_course_key`. Can we please update the variable name?

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -144,17 +146,31 @@ def publish_to_discovery(self, partner, course_run):
...
-            'pacing_type': course_run.pacing_type,
-            'title_override': course_run.title_override,
-            'min_effort': course_run.min_effort,
-            'max_effort': course_run.max_effort,
-            'language': course_run.language,
-            'weeks_to_complete': course_run.length,
-            'learner_testimonials': publisher_course.learner_testimonial,
-        }
+        if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING):
+            defaults = {
```
schenedx: Why define it twice? Why not switch on the key that actually mattered? i.e. `pacing_type`?

course_discovery/apps/publisher/studio_api_utils.py
```diff
@@ -57,18 +59,32 @@ def generate_data_for_studio_api(cls, publisher_course_run):
...
+                'title': publisher_course_run.title_override or course.title,
+                'org': course.organizations.first().key,
+                'number': course.number,
+                'run': cls.calculate_course_run_key_run_value(publisher_course_run),
+                'schedule': {
+                    'start': serialize_datetime(publisher_course_run.start),
+                    'end': serialize_datetime(publisher_course_run.end),
+                },
+                'team': team,
+                'pacing_type': publisher_course_run.lms_pacing,
```
schenedx: Again, please make this dry

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -68,6 +68,19 @@ def test_studio_url(self):
                                             id=self.course_run.lms_course_id)
         assert actual == self.course_run.studio_url
 
+    def test_studio_schedule_and_details_url(self):
+        assert self.course_run.studio_schedule_and_details_url is None
```
schenedx, line 77: This can be in it's own test case.

schenedx (commented):
@MichaelRoytman I have all minor comments. I can only imagine all the bumps a long the way throughout this PR. I want to give you a big pad on the back to getting this PR in the mergeable state!

course_discovery/apps/core/tests/test_api_clients.py
bderusha: also, fake doesn't imply whether or not it should work.  We create fake data in tests all the time.  If it's meant to be a failure case something like `bad_course_key` or `non_existant_course_key` might be even better

course_discovery/apps/publisher/views.py
```diff
@@ -196,6 +197,10 @@ def get_context_data(self, **kwargs):
                 context['publish_date'] = history_object.modified
 
         start_date = course_run.start.strftime("%B %d, %Y") if course_run.start else None
+        pacing_type = (
+            course_run.lms_pacing_type_display if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING)
```
bderusha: NIT:  if you can't fit the ternary on a single line, use standard `if` block structure

course_discovery/apps/publisher/models.py
```diff
@@ -435,6 +442,25 @@ def paid_seats(self):
     def get_absolute_url(self):
         return reverse('publisher:publisher_course_run_detail', kwargs={'pk': self.id})
 
+    @property
+    def lms_pacing(self):
```
bderusha: Should this be a database backed field rather than making an API call each time the value is needed?  I'm a bit worried about performance here. 

Is it possible to use the metadata refresh command to populate this rather than rely on an API call each time?

course_discovery/apps/publisher/templates/publisher/course_detail/_widgets.html
```diff
@@ -23,7 +23,11 @@ <h5 class="hd-5 emphasized course-runs-heading">{% trans "COURSE RUNS" %}</h5>
                 <div class="layout-col layout-col-b course-run-details">
                     <div class="courserun-details">
                         <a class="courserun-detail-link" href="{% url 'publisher:publisher_course_run_detail' pk=course_run.id %}">
-                            {{ course_run.start.date }} - {{ course_run.get_pacing_type_display }}
+                            {% if publisher_remove_pacing_type_editing %}
+                                {{ course_run.start.date }} - {{ course_run.lms_pacing_type_display }}
+                            {% else %}
+                                {{ course_run.start.date }} - {{ course_run.get_pacing_type_display }}
```
bderusha: I think a lot of this would be easier to reason about if we overrode the `pacing_type` and `get_pacing_type_display` methods and checked the waffle flag there to know where to get the value from.

course_discovery/apps/publisher/templates/publisher/course_run_detail/_drupal.html
```diff
@@ -99,7 +99,15 @@
             {% trans "Self Paced" %}
             {% include "publisher/course_run_detail/_clipboard.html" %}
         </div>
-        <div class="copy">{{ course_run.pacing_type }}</div>
+        <div class="copy">
+            {{
+                {% if publisher_remove_pacing_type_editing %}
```
bderusha: see above comment about pulling this logic into an overriding method

course_discovery/apps/publisher/templates/publisher/course_run_detail/_studio.html
```diff
@@ -43,7 +43,13 @@
                 {% trans "Pacing Type" %}
                 {% include "publisher/course_run_detail/_clipboard.html" %}
             </div>
-            <div class="copy">{{ course_run.pacing_type }}</div>
+            <div class="copy">
+                {% if publisher_remove_pacing_type_editing %}
```
bderusha: see above comment about pulling this logic into an overriding method

course_discovery/apps/publisher/views.py
```diff
@@ -432,8 +437,13 @@ def _get_published_course_runs(self, course):
         for course_run in self._get_active_course_runs(course):
             if course_run.course_run_state.is_published:
                 start_date = course_run.start.strftime("%B %d, %Y") if course_run.start else None
+                pacing_type = (
+                    course_run.lms_pacing_type_display if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING)
```
bderusha: same comment as above:  move to block `if` format

course_discovery/apps/publisher/views.py
```diff
@@ -445,6 +455,10 @@ def _get_misconfigured_course_runs(self, course, price, mode):
             seats = course_run.seats.all()
             type_is_valid = True
             price_is_valid = True
+            pacing_type = (
+                course_run.lms_pacing_type_display if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING)
```
bderusha: same comment as above:  move to block `if` format

course_discovery/apps/publisher/views.py
```diff
@@ -1006,12 +1023,16 @@ def get(self, request, *args, **kwargs):
             context['seat_form'] = self.seat_form(instance=course_run_seat)
 
         start_date = course_run.start.strftime("%B %d, %Y") if course_run.start else None
+        pacing_type = (
+            course_run.lms_pacing_type_display if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING)
```
bderusha: same comment as above:  move to block `if` format

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -39,15 +41,19 @@ def send_email_for_comment(comment, created=False):
             # Translators: subject_desc will be choice from ('New comment added', 'Comment updated'),
             # 'pacing_type' will be choice from ('instructor-paced', 'self-paced'),
             # 'title' and 'start' will be the value of course title & start date fields.
+            pacing_type = (
+                publisher_obj.lms_pacing_type_display if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING)
```
bderusha: same comment as above:  move to block `if` format

bderusha (dismissed):
A handful of style comments.  But a bigger architecture question I'd love to chat about: Specifically the use of API calls vs storing the value locally

course_discovery/apps/publisher/tests/test_models.py
michaelroytman: What do you mean by that? The line `assert self.course_run.studio_schedule_and_details_url is None` should be a separate test function? I think this assert statement checking that the URL is None before any of the necessary changes are made to the CourseRun object that allows for the URL to be generated. I'm not sure that the assert is actually necessary, but I was following the pattern of the `test_studio_url function` above it. 

I'm not sure it'll serve the same purpose if it's a separate test case, since I think the intention is to make sure that the test is starting with a "blank slate", so to speak.

michaelroytman:
@schenedx Could you take another look at this now that we have made significant changes? Thank you!

course_discovery/apps/publisher/models.py
```diff
@@ -448,16 +459,50 @@ def pacing_type_temporary(self):
...
 
     def get_pacing_type_temporary_display(self):
-        return self.get_pacing_type_display()
+        if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING):
+            discovery_counterpart = self.discovery_counterpart
+
+            if discovery_counterpart and discovery_counterpart.pacing_type:
+                return discovery_counterpart.get_pacing_type_display()
+
+            return 'Instructor-paced'
```
dsjen: Should this be translated?

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -153,29 +169,241 @@ def test_get_absolute_url(self):
...
+        # create a fresh course run object to avoid issues with caching of discovery_counterpart property
         course_run = factories.CourseRunFactory()
 
         assert course_run.get_pacing_type_temporary_display() == course_run.get_pacing_type_display()
 
+    @override_switch(PUBLISHER_REMOVE_PACING_TYPE_EDITING, active=True)
+    def test_pacing_type_temporary_display_success_with_switch(self):
+        """
+        Verify that pacing_type_temporary display function returns the
+        value of the corresponding Discovery course runs's pacing_type display function when waffle switch is enabled.
```
dsjen: Nit: spelling "runs's"

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -153,29 +169,241 @@ def test_get_absolute_url(self):
...
+        self.create_discovery_course_run_with_course_and_pacing_type(discovery_course, pacing_type_test_value)
+
+        self.add_organization_to_course(course_run.course, organization)
+
+        # make sure Publisher course key and Course Metadata course key match
+        course_run.course.key = discovery_course.key
+
+        assert course_run.get_pacing_type_temporary_display() == 'Instructor-paced'
+
+    def create_discovery_course_with_partner(self, partner):
```
dsjen: Nit: This method and the below are utility methods. Maybe use [staticmethod ](https://docs.python.org/3/library/functions.html#staticmethod)?

course_discovery/apps/publisher/tests/test_models.py
schenedx: Now I see what you are saying. That make sense. Ignore my comment here.

schenedx (approved):
👍

course_discovery/apps/publisher/models.py
```diff
@@ -448,16 +459,50 @@ def pacing_type_temporary(self):
             The progress of the above work will be tracked in the following ticket:
             https://openedx.atlassian.net/browse/EDUCATOR-3488.
         """
-        return self.pacing_type
+        if waffle.switch_is_active(PUBLISHER_REMOVE_PACING_TYPE_EDITING):
+            discovery_counterpart = self.discovery_counterpart
```
dsjen, line 463: Nit: I think the lines in the if statement could be reduced to:

```
return getattr(self.discovery_counterpart, 'pacing_type', self.DEFAULT_PACING_TYPE)
```

Maybe that's less readable though?

dsjen (approved):
LGTM! I left a few nits (truly feel free to ignore them as this PR looks good).

course_discovery/apps/publisher/models.py
michaelroytman: Yes, thank you!

course_discovery/apps/publisher/tests/test_models.py
michaelroytman: Thank you! I didn't know that was an available decorator.

course_discovery/apps/publisher/models.py
michaelroytman: I like that :).

PR #1511, opened 2018-08-17
Add middleware for collecting request metrics.

douglashall:


requirements/base.txt
```diff
@@ -35,9 +35,10 @@ edx-auth-backends==1.1.3
 edx-ccx-keys==0.2.0
 edx-django-release-util==0.3.1
 edx-django-sites-extensions==2.3.0
-edx-drf-extensions==1.2.3
+edx-django-utils==0.4.1
```
robrap, line 38: Just for fun we should probably update to 0.5.1 when we rebase, which it looks like we need to do anyway.  I can take care of this and update on Monday, although I have worse luck than you. ;)

robrap (approved):

requirements/base.txt
robrap: FYI: I rebased and updated to 0.5.1 in the old PR.  We'll see if it ever passes. https://github.com/edx/course-discovery/pull/1500

robrap:
Merged in this PR: https://github.com/edx/course-discovery/pull/1500

PR #1515, opened 2018-08-20
Update the master about page header to use two images

schenedx:
@reillz10 This is the PR to do the header image for about page.


reillz10:
I took care of the changes in this PR with the following commit: 02f226a

PR #1517, opened 2018-08-21
Update for refresh course metadata to continue if an exception is

staubina:
reached on Connection errors.

[LEARNER-6211]

Continue trying the other partners when the connection to one partner fails.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
@@ -94,9 +94,12 @@ class Command(BaseCommand):
                     partner.oidc_secret,
                     token_type=token_type
                 )
-            except Exception:
-                logger.exception('No access token acquired through client_credential flow.')
-                raise
+            except Exception:  # pylint: disable=broad-except
+                message = '[{name}] Failed to retrieve access token through client_credential flow.'.format(
+                    name=partner.name
+                )
+                logger.exception(message)
+                continue
             username = jwt.decode(access_token, verify=False)['preferred_username']
             kwargs = {'username': username} if username else {}
 
--- a/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
@@ -95,7 +95,7 @@ class Command(BaseCommand):
                     token_type=token_type
                 )
             except Exception:  # pylint: disable=broad-except
-                message = '[{name}] Failed to retrieve access token through client_credential flow.'.format(
+                message = 'Failed to retrieve access token through client_credential flow for Partner [{name}]'.format(
                     name=partner.name
                 )
                 logger.exception(message)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -94,9 +94,12 @@ def handle(self, *args, **options):
                     partner.oidc_secret,
                     token_type=token_type
                 )
-            except Exception:
-                logger.exception('No access token acquired through client_credential flow.')
-                raise
+            except Exception:  # pylint: disable=broad-except
```
BenjiLee, line 97: Do we know what exception is being raised?

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -94,9 +94,12 @@ def handle(self, *args, **options):
...
+                message = '[{name}] Failed to retrieve access token through client_credential flow.'.format(
```
BenjiLee: I think we need some sort of descriptor of what the `{name}` variable is.
In the case where name = null, it will not be clear what `{name}` is supposed to be.

For example:
`[] Failed to retrieve access token through client_credential flow.`
vs
`partner.name=[ ] Failed to retrieve access token through client_credential flow.`
or
`Failed to retrieve access token through client_credential flow for partner.name=[ ]`

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -94,9 +94,12 @@ def handle(self, *args, **options):
...
+                    name=partner.name
+                )
+                logger.exception(message)
+                continue
```
calistenson: Can you clarify why we're safe continuing? Is there a ticket to trigger an alert somewhere else to follow up on this exception?

calistenson: This fix seems okay as a temporary solution but could lead to a larger issue of these exceptions being ignored. Looks okay to me if there's a plan to address the exceptions with a different alerting process.

calistenson (approved):

BenjiLee (approved):
Approval when the log comment is addressed.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
staubina: ConnectionError, I am looking at how to add it so I do not need the broad except.  This initial fix was to get the service working again.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
staubina: There is no Alert right now.  In the ticket I added that we need to add one, but the fix here will continue to process the connections to our other partners when a connection fails to work.  This way we will still collect data when one connection is not working correctly.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -94,9 +94,12 @@ def handle(self, *args, **options):
...
                 )
-            except Exception:
-                logger.exception('No access token acquired through client_credential flow.')
-                raise
+            except Exception:  # pylint: disable=broad-except
+                message = 'Failed to retrieve access token through client_credential flow for Partner [{name}]'.format(
+                    name=partner.name
+                )
+                logger.exception(message)
+                continue
```
bderusha, line 102: This will mark the job as succeeded even when there are failures. 

We should have an alert for this or capture the error and exit with the proper code so jenkins knows it failed even if we allow some of it to go through.

Either way, not sure this is necessary with the new on-call approach to errors

PR #1525, opened 2018-08-22
Add a test_local.py settings file to speed up local test setup.

iloveagent57:
Before this change, the first and every subsequent run of 
```
pytest course_discovery/apps/course_metadata/tests/test_utils.py
```
runs in `13 passed in 41.72 seconds` time.  Most of that time is database setup.

With this PR you can now run:
```
pytest course_discovery/apps/course_metadata/tests/test_utils.py --reuse-db --ds=course_discovery.settings.test_local
```
and the first run will still take about 40 seconds, but the second run finishes much faster: `13 passed in 7.27 seconds`

FYI: @edx/educator-neem


iloveagent57:
...this is very nice when you only want to run one or two test functions and don't want to wait 30 seconds.

schenedx (approved):
👍

PR #1527, opened 2018-08-24
Take a different approach to the unit test code structure

schenedx:
Update the tests to create a different utility function for Degrees. This way, we can make degree specific fields somewhat separated from rest of the Program code.


yro (approved):
👍

PR #1534, opened 2018-08-27
Add pathway_type field to pathway model

bbaker6225:
pathway_type is used to distinguish between credit and professional
pathways, with industry being used in the code to separate the concept
from professional certificate programs.  Accompanies a similar
change in credentials.

https://openedx.atlassian.net/browse/LEARNER-6149


course_discovery/apps/course_metadata/models.py
```diff
@@ -1560,6 +1561,11 @@ class Pathway(TimeStampedModel):
     programs = SortedManyToManyField(Program)
     description = models.TextField(null=True, blank=True)
     destination_url = models.URLField(null=True, blank=True)
+    pathway_type = models.CharField(
+        max_length=255,
+        choices=[(tag.value, tag.value) for tag in PathwayType],
```
bbaker6225: Trying to set the pathway in admin with anything other than tag.value as the first variable caused django to make a fuss.  This is still compatible with the choices in credentials though

course_discovery/apps/course_metadata/models.py
```diff
@@ -1560,6 +1561,11 @@ class Pathway(TimeStampedModel):
     programs = SortedManyToManyField(Program)
     description = models.TextField(null=True, blank=True)
     destination_url = models.URLField(null=True, blank=True)
+    pathway_type = models.CharField(
+        max_length=255,
```
mikix: 255 is pretty large for an enum field. I've seen us use numbers like 24 and 32 before. Not a problem though. Just saying.

course_discovery/apps/api/serializers.py
```diff
@@ -1006,10 +1006,21 @@ class PathwaySerializer(serializers.ModelSerializer):
...
+        fields = (
+            'id',
+            'uuid',
+            'name',
+            'org_name',
+            'email',
+            'programs',
+            'description',
+            'destination_url',
+            'pathway_type',
```
mikix, line 1022: I assume we didn't go with just "type" partly because it's a keyword and is annoying to use as a variable name. I agree with the choice, just noticing.

mikix (approved):
:+1:

course_discovery/apps/course_metadata/models.py
bbaker6225: Yeah, I wasn't really sure what to put for the length, and I just knew that 255 is the standard go to.  Happy to change it if you want

course_discovery/apps/api/serializers.py
bbaker6225: Yeah, even though python does let you use it as a variable I find it pretty annoying.  Especially when it's a different color from all the other variables in my editor

course_discovery/apps/course_metadata/models.py
mikix: I don't feel strongly. It feels too large, but like, I don't know if that's worth optimizing for really.

bbaker6225: I'll change it to 32 here and in credentials

PR #1541, opened 2018-09-04
EDUCATOR-3290: Adds short and long marketing text. Supports Curriculums "See More" button.

sandroroux:
[EDUCATOR-3290](https://openedx.atlassian.net/browse/EDUCATOR-3290)


iloveagent57 (commented):
You're gonna need:
* A migration to drop the old field and add the new ones.
* Probably to update the fields in the degree factory model, so that tests all work seamlessly.

schenedx (changes requested):
@sandroroux The way you structured your migration here will break prod. You need to take this into multiple PR and releases so you don't break prod. 
Step 1: Add the new fields (marketing_text_brief', 'marketing_text_long'). Release this out
Step 2: Fill in the existing data into the new fields
Step 3: A PR to remove all reference to model field `marketing_text`, but not including the migration
Step 4: a PR with just the migration to remove `marketing_text` column.

course_discovery/apps/course_metadata/migrations/0113_brief_text_curriculum.py
```diff
@@ -0,0 +1,30 @@
...
+            model_name='curriculum',
+            name='marketing_text_brief',
+            field=models.TextField(blank=True, help_text='A high-level overview of the degree\'s courseware. The "brief" text is the first 300 characters of "marketing_text" and must be valid HTML.', max_length=300, null=True),
+        ),
+        migrations.AlterField(
+            model_name='curriculum',
+            name='marketing_text',
+            field=models.TextField(blank=True, help_text="A high-level overview of the degree's courseware.", null=True),
+        ),
+        migrations.AlterField(
```
iloveagent57: I don't understand why this field is included in this migration.  You didn't change anything about it in the model definition.

iloveagent57 (commented):
I think this makes sense, I just don't get why there are changes involving `banner_border_color` in the migration.

course_discovery/apps/course_metadata/migrations/0113_brief_text_curriculum.py
schenedx: It could be someone forgot to include this particular migration.
I am fine having this migration including this change.

schenedx (approved):
👍

schenedx:
You need to do `make fake_translations` in your discovery container.

PR #1542, opened 2018-09-04
Remove no longer used model columns

schenedx:
Remove the no longer used campus_image_x fields on the about page header.
Intentionally left out the migration file, because that needs to be created and deployed in a following PR
@reillz10 and @iloveagent57 Please review


iloveagent57 (approved):
LGTM

PR #1547, opened 2018-09-06
Add pyinotify to improve dev server code reloading

davestgermain:
On devstack, each IDA consumes ~2% CPU while idle, because of Django's dev server code reloading. Adding the pyinotify library decreases CPU usage to <0.1%

@nedbat

### DIFF
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -10,3 +10,6 @@ edx-i18n-tools==0.4.4
 
 # docker devstack
 docker-compose>=1.7.1,<2.0.0
+
+# For devserver code reloading
+pyinotify==0.9.6


nedbat:
@edx/platform-team @jmbowman Do we know about this for other services?

@davestgermain Seems fine to me.

nedbat (approved):

davestgermain:
@nedbat I was going to open PRs for the other services after this one -- ecommerce, edx_notes_api, and credentials. LMS and Studio already use pyinotify.

nedbat (approved):

jmbowman:
Hi, catching up on notifications after getting back from vacation.  I could have sworn I'd created JIRA tickets to do this, but so far I've only found a text note I left for myself to follow up on this (and I think it came up verbally in a meeting).  Anyway, we definitely want pyinotify installed for each of the Django services using runserver in devstack, preferably only as part of the dev environment dependencies.  @davestgermain , could you go ahead and create tickets for this so we don't lose track of it again?

jmbowman:
@davestgermain What's the status on this?  Do you need any help with test failures, or does it just need to be rebased?

davestgermain:
@jmbowman It just needed to be rebased. I hadn't merged this yet because I hadn't gotten around to creating those tickets, as you suggested.

davestgermain:
@jmbowman Looks like there are some flaky tests. My change should have no effect on any tests.

PR #1549, opened 2018-09-07
Increases marketing_text_brief character limit to 750 characters.

sandroroux:


schenedx (approved):
You would need run `make fake_translations`. Other than that, you are good to merge and deploy

PR #1551, opened 2018-09-08
Upgrade edx-django-utils.

robrap:
ARCH-223


nasthagiri:
@robrap Looks like a query count failure.  Does Discovery use the new RequestCache yet?

robrap:
@nasthagiri: Discovery has a lot of nasty flaky tests that I think the team just ignores and reruns.  I may start the flaky test process on some of them.  We'll see.

PR #1554, opened 2018-09-11
Fix the character count both in publisher and discovery.

attiyaIshaque:
I have reverted this PR #1497 to keep the character count same on backend and frontend.
Also i have remove [max_lenght](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L348)  from **short_description** to keep the sync between [course metadata model Course](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L348) and [publisher model Course](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/models.py#L55).


course_discovery/apps/publisher/forms.py
```diff
@@ -75,27 +75,22 @@ class CourseForm(BaseForm):
         label=_('Organization Name'),
         required=True
     )
-    title = forms.CharField(
-        label=_('Course Title'), required=True,
-        validators=[MaxLengthValidator(255)]
-    )
+    title = forms.CharField(label=_('Course Title'), required=True)
```
awaisdar001: I think the title doesn't have tinymce widget and hence styling is not available, I would suggest keeping the validator for `title` makes sense.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2804,7 +2821,15 @@ def test_update_course_organization(self):
 
         self.assertEqual(self.course.organizations.first(), self.organization_extension.organization)
 
-        self.AssertEditCourseSuccess(post_data=self._post_data(organization_extension))
```
awaisdar001: I would suggest keeping these methods `AssertEditCourseSuccess` in as they reduce significant LOCs

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2927,60 +2984,80 @@ def test_video_link_field(self):
         response = self.client.get(self.edit_page_url)
         self.assertNotContains(response, 'VIDEO LINK')
 
-    def test_update_course_with_empty_title(self):
```
awaisdar001: `test_update_course_with_empty_title` name makese sense. do not change it.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2927,60 +2984,80 @@ def test_video_link_field(self):
...
-        ('short_description', 350),
-        ('full_description', 2500),
-        ('prerequisites', 1000),
-        ('expected_learnings', 2500),
-        ('learner_testimonial', 500),
-        ('faq', 2500),
-        ('syllabus', 2500)
-    )
-    @ddt.unpack
-    def test_max_length_exceed_error(self, field, max_length):
```
awaisdar001: Add `ddt.data` to test it for all fields like it was testing before.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -2927,60 +2984,80 @@ def test_video_link_field(self):
...
-        ('expected_learnings', 2500),
-        ('learner_testimonial', 500),
-        ('faq', 2500),
-        ('syllabus', 2500)
+        'short_description', 'full_description', 'prerequisites', 'expected_learnings',
+        'learner_testimonial', 'faq', 'syllabus'
     )
-    @ddt.unpack
-    def test_max_length_validation_pass(self, field, max_length):
+    def test_text_area_post_with_html(self, field):
```
awaisdar001: Similarly for this one.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -3251,36 +3376,6 @@ def test_course_with_published_course_run(self):
 
         self.assertEqual(str(mail.outbox[0].subject), expected_subject)
 
-    def AssertEditCourseSuccess(self, post_data):
-        """Helper method to assert the course edit was successful"""
-        response = self.client.post(self.edit_page_url, data=post_data)
-        self.assertRedirects(response, expected_url=self.course_detail_url, status_code=302, target_status_code=200)
-        self.assertNotContains(response, text=self.course_error_message, status_code=302)
-
-    def AssertEditFailedWithError(self, post_data, error):
```
awaisdar001: These are good helper methods which we should not be removing.

awaisdar001 (changes requested):
I have made few comments also @schenedx to review the migration of the PR as migrations in discovery are often very tricky.

schenedx:
@attiyaIshaque Do we have a jira ticket documenting the requirement associated with this work?
also, I don't like how you packaged two different sets of efforts into a single PR. This makes this PR really difficult to review. Can you please separate out your work on converting the short description field from charfield to textfield a separate PR from the revert?

attiyaIshaque:
Okay, @schenedx I will work in two separate PR's.

PR #1557, opened 2018-09-12
Updates to the Ecommerce Data loader to detect network errors

staubina:
[LEARNER-6317]

This PR adds the following to the EcommerceDataLoader in the refresh_course_metadata command. This was added to protect the command from flaky network connections.

- Network request error detection 
- Basic retry to remove the likelihood of intervention being required.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -1,12 +1,14 @@
 import concurrent.futures
 import logging
 import math
+import threading
 import time
 from decimal import Decimal
 from io import BytesIO
 
 import requests
 from django.core.files import File
+from django.core.management import CommandError
 from opaque_keys.edx.keys import CourseKey
 
 from course_discovery.apps.core.models import Currency
@@ -265,6 +267,8 @@ class CoursesApiDataLoader(AbstractDataLoader):
 class EcommerceApiDataLoader(AbstractDataLoader):
     """ Loads course seats, entitlements, and enrollment codes from the E-Commerce API. """
 
+    LOADER_MAX_RETRY = 2
+
     def __init__(self, partner, api_url, access_token=None, token_type=None, max_workers=None,
                  is_threadsafe=False, **kwargs):
         super(EcommerceApiDataLoader, self).__init__(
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -273,74 +278,151 @@ def __init__(self, partner, api_url, access_token=None, token_type=None, max_wor
...
-                for page in pageranges['course_runs']:
-                    executor.submit(self._load_course_runs_data, page)
-                for page in pageranges['entitlements']:
-                    executor.submit(self._load_entitlements_data, page)
-                for page in pageranges['enrollment_codes']:
-                    executor.submit(self._load_enrollment_codes_data, page)
+    def ingest(self):
+        attempt_count = 0
+
+        while (attempt_count == 0 or
```
staubina: Note: When adding in a retry loop the code was shifted in so it looks like more was changed.

bderusha: For readability can we put the code that does the thing in its own method so the ingest method looks more like

```
succeeded = do_the_thing()
attempt_count=1
while(not succeeded and attempt_count < MAX):
    succeeded = do_the_thing()
    attempt_count += 1

if not succeeded:
    error
```

This will also make reviewing this code easier because the indentation will be clear as to what you actually changed.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -273,74 +278,151 @@ def __init__(self, partner, api_url, access_token=None, token_type=None, max_wor
...
+            with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:  # pragma: no cover
+                # Create pageranges to iterate over all existing pages for each product type
+                pageranges = {
+                    'course_runs': self._pagerange(course_runs['count']),
+                    'entitlements': self._pagerange(entitlements['count']),
+                    'enrollment_codes': self._pagerange(enrollment_codes['count'])
+                }
+
+                if self.is_threadsafe:
+                    for page in pageranges['course_runs']:
```
bderusha: Rather than wrapping the body of these methods with the `try` block can we just wrap this whole section?

bderusha (changes requested):
The gist of this code looks great.  Left a few comments of things we might want to restructure

course_discovery/apps/course_metadata/data_loaders/api.py
staubina: No because the exception will not bubble up out of the thread.  That is why we are not seeing the exceptions at process level.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -301,46 +344,87 @@ def ingest(self):
...
                 for future in [executor.submit(self._request_course_runs, page) for page in pagerange]:
-                    response = future.result()
-                    self._process_course_runs(response)
+                    check_exception = future.exception()
+                    if check_exception is None:
+                        response = future.result()
+                        self._process_course_runs(response)
+                    logger.exception(check_exception)
+                    # Protect against deletes if exceptions occurred
+                    self.processing_failure_occurred = True
```
bderusha: I don't doubt that this works... but I don't understand how.  How does this not get recorded as a failure every time?

bderusha (approved):
One sanity check question, more for me than anything else.  Otherwise :shipit: 


Also squash before merging

course_discovery/apps/course_metadata/data_loaders/api.py
staubina: Looking at this again I think I need to flip the logic or add a continue.

Good catch.

staubina: @bderusha Updated with a new commit.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -273,14 +277,53 @@ def __init__(self, partner, api_url, access_token=None, token_type=None, max_wor
...
+    def _load_ecommerce_data(self):
         course_runs = self._request_course_runs(self.initial_page)
         entitlements = self._request_entitlments(self.initial_page)
         enrollment_codes = self._request_enrollment_codes(self.initial_page)
+
         self.entitlement_skus = []
         self.enrollment_skus = []
+
+        self.processing_failure_occurred = False
+        self.course_run_count = 0
```
jlajoie, line 324: I feel for best practice you'll want the locks acquired/released here on course_run_count since this method will be callable from within the threads. Better to just future proof this.

staubina: 👍 on this catch.

staubina: Actually looking at this again I am not sure it is needed because the threads are not currently active at this point in the code.

The threading starts just below this initialization after the with call.

jlajoie (approved):

PR #1561, opened 2018-09-18
add pacing_type_temporary property and change uses of course_run.paci…

michaelroytman:
…ng_type to use it instead, where appropriate

### DIFF
--- a/course_discovery/apps/publisher/models.py
+++ b/course_discovery/apps/publisher/models.py
@@ -281,7 +281,7 @@ class CourseRun(TimeStampedModel, ChangedByMixin):
     certificate_generation = models.DateTimeField(null=True, blank=True)
     pacing_type = models.CharField(
         max_length=255, db_index=True, null=True, blank=True, choices=CourseRunPacing.choices,
-        validators=[CourseRunPacing.validator]
+        validators=[CourseRunPacing.validator],
     )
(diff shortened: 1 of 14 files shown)


michaelroytman:
@edx/educator-dahlia Please review. Thanks!

course_discovery/apps/publisher/models.py
```diff
@@ -435,6 +435,27 @@ def paid_seats(self):
...
+    @property
+    def pacing_type_temporary(self):
+        """
+            This property serves as a temporary intemediary in order to support a waffle
+            switch that will toggle between the original database backed pacing_type value
+            and a new read-only value that is pulled from course_discovery.
+
+            Once the switch is no longer needed, the pacing_type field will be removed,
+            and this property will be renamed appropriately.
+        """
```
fysheets: 👍 Thank you for the detailed comment

michaelroytman:
Just as an FYI, @bderusha.

fysheets (commented):
This looks good, but looking at the project structure - I noticed there were references to `pacing_type` in the `publisher_comments` section as well. Wanted to sanity check if we should update those as well?

michaelroytman:
@fysheets Yes, you're right. I'll fix that now.

michaelroytman:
@fysheets This should be all set for another look. Thanks!

fysheets (approved):
Thanks!

PR #1562, opened 2018-09-20
Optimize catalogs endpoint a bit

mikix:
If the catalog query is simply '*', which is very common, don't bother with an elasticsearch query at all.

About 2/3rds of the catalogs are just '*'. On my devstack, it changes the query time from 3-4s to under 1s.

I also optimized a separate place where we were sometimes doing the elasticsearch query twice for no reason.


inventhouse:
This looks good to me, nice find!

PR #1563, opened 2018-09-20
EDUCATOR-3360 | Add a mm bg image field to the degree model.

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-3360


schenedx (approved):
👍

PR #1564, opened 2018-09-20
Fix the missing award icon to make it trophy icon

schenedx:
Apparently, fa-award is not a valid icon class name.
fa-trophy is a valid one tho. Switching the class name to trophy.
Also add fa-star because that's what will be used.

[EDUCATOR-3491](https://openedx.atlassian.net/browse/EDUCATOR-3491)

@iloveagent57 and @yro Please review


iloveagent57 (approved):

schenedx:
@iloveagent57 I added both Star and Trophy icons. Care to review again?

PR #1567, opened 2018-09-25
Make lead_capture_list_name a required field with default = 'Master_d…

reillz10:
Make lead_capture_list_name a required field with default = 'Master_default'


course_discovery/apps/course_metadata/models.py
```diff
@@ -1373,8 +1373,7 @@ class Degree(Program):
     lead_capture_list_name = models.CharField(
         help_text=_('The sailthru email list name to capture leads'),
         max_length=255,
-        blank=True,
-        null=True,
+        default='Master_default',
```
schenedx, line 1376: Is there a way to update the field without specifying a default? I am curious. This do not block merge and deploy.

schenedx (approved):
👍

course_discovery/apps/course_metadata/models.py
reillz10: I just checked on my local machine and we DO NOT need to specify a default value. That said, I thought you mentioned that we wanted to.

schenedx: Let's specify one to be safe. What you have here is good to ship

iloveagent57 (approved):

PR #1570, opened 2018-09-26
EDUCATOR-3520 | Use the UploadToFieldNamePath class to populate stdimage field names

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-3520

We already had a custom class to do this in `utils.py`, making use of it for Degree fields now.

The `UploadToFieldNamePath` class updates the field name with the last 12 digits of a uuid4 (these uuids are random, they don't correspond to time, MAC addresses, etc.)


schenedx (approved):
👍

PR #1571, opened 2018-09-27
Add extra_description field to course object in course_metadata

schenedx:
For the course metadata course object, we have another field called "Additional Promo". We would like to ingest the data in this field in drupal into course metadata.

I have tested the marketing site ingestion and ensured the data are ingested.

Please review @mikedikan @MichaelRoytman


course_discovery/apps/api/serializers.py
```diff
@@ -165,6 +165,18 @@ class Meta(object):
         fields = ('name',)
 
 
+class TitleDescriptionSerializer(serializers.ModelSerializer):
```
mikedikan, line 168: why did you add this base class?

course_discovery/apps/course_metadata/migrations/0122_auto_20180928_1902.py
```diff
@@ -0,0 +1,35 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0121_degree_image_naming'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='AdditionalPromoArea',
```
mikedikan: why create a new model instead of adding the fields onto existing model?

course_discovery/apps/api/serializers.py
schenedx: Because I believe this pattern may be used for future cases. It's strictly not necessary.  But I thought why not.

course_discovery/apps/course_metadata/migrations/0122_auto_20180928_1902.py
schenedx: Because there are two related fields we are ingesting from drupal. In Drupal, the additional Promo Area consist of a Title and a Description. Both fields have value.
So I thought let's just introduce a model to capture the relationship.

mikedikan (approved):
My personal choice would be to not have the base class (if we need it later, we'll add it when we have 2 examples of the thing, not 1) and to just have the fields added directly and not have a separate model.  However, I'm not that invested in this perspective and don't feel like it's worth arguing over!

schenedx:
@mikedikan I messed up the commit here. I squashed my commit during master rebase 🤦‍♂️ .  I'll create a new PR with the proper commit.

schenedx:
This change is now transferred to https://github.com/edx/course-discovery/pull/1579

PR #1573, opened 2018-10-01
Try to limit the output of make static to make travis happy

schenedx:
@iloveagent57 @mikedikan 
I think it make sense to just merge this into course-discovery to get travis happy.

@MichaelRoytman, @staubina  FYI


iloveagent57 (approved):
Even if this doesn't work, I'm happy to cut down some of this log output.

PR #1574, opened 2018-10-01
Create start_date_temporary property in Publisher and Leverage Instead of start Model Value

fysheets:
### [EDUCATOR-3524](https://openedx.atlassian.net/browse/EDUCATOR-3524)

Update Publisher to leverage a new property for start date that will help enable a waffle flag for testing start date as a read only field when editing. The plan is to make it so that start date is only editable from Studio.


fysheets:
FYI @edx/educator-dahlia - A manual run through / visual check is probably needed here, and I want to do a sanity check that I both updated all necessary locations & did not update places that were unnecessary.

course_discovery/apps/publisher/models.py
```diff
@@ -459,6 +459,30 @@ def pacing_type_temporary(self, value):
...
 
+    @property
+    def start_date_temporary(self):
+        """
+            This property serves as a temporary intemediary in order to support a waffle
+            switch that will toggle between the original database backed pacing_type value
+            and a new read-only value that is pulled from course_discovery.
+
+            The start date will need to continue to exist for the write on create,
+            until that funcitonality is officially moved to Studio creation.
```
matthugs: nit: typo "functionality"

course_discovery/apps/publisher_comments/emails.py
```diff
@@ -35,6 +35,7 @@ def send_email_for_comment(comment, created=False):
         if comment_class == CourseRun:
             course = publisher_obj.course
             object_path = reverse('publisher:publisher_course_run_detail', args=[publisher_obj.id])
+            publisher_start_date = publisher_obj.start_date_temporary
```
matthugs: For my edification: can you help me understand better why this was worth pulling out of the `publisher_obj` in this context? Is it purely for the sake of line lengths?

I'm concerned that pulling it out may obfuscate the source of the information. I.e. that the indirection is unnecessary and counterproductive? It would be a nit either way, but it made me wonder whether this was something out of Robert Martin's book?

matthugs (commented):
This looks good and I would probably thumb it if I weren't brand new at looking at django models. Thanks for modeling for me/telling me how you did this so I can replicate for end dates!

course_discovery/apps/publisher_comments/emails.py
fysheets: This change was strictly for quality/line lengths.

matthugs: /shrug ok!

course_discovery/apps/publisher/models.py
```diff
@@ -459,6 +459,30 @@ def pacing_type_temporary(self, value):
     def get_pacing_type_temporary_display(self):
         return self.get_pacing_type_display()
 
+    @property
+    def start_date_temporary(self):
+        """
+            This property serves as a temporary intemediary in order to support a waffle
```
MichaelRoytman: Nit: This typo and the one Matt found are totally my fault, but "intemediary" => "intermediary".

michaelroytman: Nit: This typo and the one Matt found are totally my fault, but "intemediary" => "intermediary".

course_discovery/apps/publisher/models.py
```diff
@@ -459,6 +459,30 @@ def pacing_type_temporary(self, value):
...
+
+            The progress of the above work will be tracked in the following ticket:
+            https://openedx.atlassian.net/browse/EDUCATOR-3524.
+        """
+        return self.start
+
+    @start_date_temporary.setter
+    def start_date_temporary(self, value):
+        # Treat empty strings as NULL
+        value = value or None
```
michaelroytman: I'm a little unsure about this setter, because the start model field is a DateTimeField, and I'm not sure how that works as far as a setter goes. Do we still need this line, since we won't really be setting start to a string? This might be my lack of familiarity with Python and its dynamic typing; just something that I thought of.

course_discovery/apps/publisher/models.py
```diff
@@ -459,6 +459,30 @@ def pacing_type_temporary(self, value):
...
+        self.start = value
+
+    def get_start_date_temporary_display(self):
```
michaelroytman: I think I made a mistake in my original JIRA write-up by suggesting that we add a `get_FOO_display_ method as part of this work. I think `get_FOO_display` methods are only available for model fields that use `choices` (see [here](https://docs.djangoproject.com/en/2.1/ref/models/instances/#django.db.models.Model.get_FOO_display)). Therefore, I believe that you don't need this method.

course_discovery/apps/publisher/management/commands/tests/test_import_courses.py
```diff
@@ -347,13 +347,13 @@ def _assert_course(self, publisher_course):
...
 
-        self.assertEqual(publisher_course_run.start, metadata_course_run.start)
+        self.assertEqual(publisher_course_run.start_date_temporary, metadata_course_run.start)
         self.assertEqual(publisher_course_run.end, metadata_course_run.end)
         self.assertEqual(publisher_course_run.min_effort, metadata_course_run.min_effort)
         self.assertEqual(publisher_course_run.max_effort, metadata_course_run.max_effort)
         self.assertEqual(publisher_course_run.length, metadata_course_run.weeks_to_complete)
         self.assertEqual(publisher_course_run.language, metadata_course_run.language)
-        self.assertEqual(publisher_course_run.pacing_type, metadata_course_run.pacing_type)
+        self.assertEqual(publisher_course_run.pacing_type_temporary, metadata_course_run.pacing_type)
```
michaelroytman, line 356: Oopsies; thanks for catching this. Just as a sanity check, will course import/export work as expected? I don't recall ever changing the command to use `pacing_type_temporary` versus `pacing_type`, so I'm wondering how we can be sure that this case will be true? I can't find the command.

michaelroytman (commented):
I haven't had a chance to visually verify these changes yet, so I'll hold off on approval until I can do that, but here are some initial comments I have.

michaelroytman (approved):
Looks good to me - just needs a final squash!

PR #1579, opened 2018-10-04
Add extra_description field to course_metadata to host drupal data in AdditionalPromoArea field group

schenedx:
For the course metadata course object, we have another field called "Additional Promo". We would like to ingest the data in this field in drupal into course metadata.

I have tested the marketing site ingestion and ensured the data are ingested.

Please review @mikedikan @MichaelRoytman


mikedikan (approved):
My personal choice would be to not have the base class (if we need it later, we'll add it when we have 2 examples of the thing, not 1) and to just have the fields added directly and not have a separate model. However, I'm not that invested in this perspective and don't feel like it's worth arguing over!

PR #1583, opened 2018-10-09
Revert the person search api so we can green the haystack update_index

schenedx:
Reverting the change at https://github.com/edx/course-discovery/pull/1566

@fysheets FYI, This is the PersonSearch API revert PR.
@MichaelRoytman Please help review.


michaelroytman (approved):
I'm not sure how migrations behave with reverts, but I'm assuming that they'll be reapplied - just putting it out there as a place I was a little hesitant due to lack of experience.

Otherwise, looks good to me (i.e. it reverts all the code added in the original PR). Let me know if you need a more thorough look.

schenedx:
This is grossly out of date. If we ever need this again, we will have a new PR.

PR #1584, opened 2018-10-09
Instead of error on index count limit surprise, just log about it

schenedx:
Remove the command error when the index limit check was tripped.  Turn that into an info
@mikedikan @MatthewPiatetsky @fysheets  Please review

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -50,7 +50,7 @@ class Command(HaystackCommand):
             if not options.get('disable_change_limit', False):
                 record_count_is_sane, index_info_string = self.sanity_check_new_index(backend.conn, index, record_count)
                 if not record_count_is_sane:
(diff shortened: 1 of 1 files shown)


MatthewPiatetsky:
I imagine if we make this an info no one will look at it and it might as well be removed.  
It seems like we introduced a new type of object which caused the index to change by more than 10% for an expected reason.
To deal with that scenario, perhaps we should just add an optional flag to the command/job to disable the check, so we can update the index by manually running the job with the parameter when such changes are made?

This check was added in the first place after we lost more than half of our production data and didn't notice right away, so I'm not sure that we want to remove it altogether.

schenedx:
@MatthewPiatetsky That's great historical context. I think it's fair to question this check to see whether it's valuable to have it or not. I knew I don't have all the answers to why this was here in the first place. Thank you for letting me know about it.

mikedikan:
I'm with @MatthewPiatetsky on this one - I remember the issues this check was addressing and they were painful.

schenedx:
@mikedikan @MatthewPiatetsky Thank you for your comments. I am closing this. We will leave that check in.

PR #1586, opened 2018-10-10
add waffle switch to toggle between reading start date from the Publi…

michaelroytman:
…sher model and pulling it in from the Course Metadata model; replace references to start field with start_date_temporary property


course_discovery/apps/publisher/forms.py
```diff
@@ -365,9 +367,12 @@ def clean(self):
 
     def __init__(self, *args, **kwargs):
         self.is_project_coordinator = kwargs.pop('is_project_coordinator', None)
+        self.hide_start_date_field = kwargs.pop('hide_start_date_field', None)
         super(CourseRunForm, self).__init__(*args, **kwargs)
         if not self.is_project_coordinator:
             self.fields['lms_course_id'].widget = forms.HiddenInput()
+        if waffle.switch_is_active(PUBLISHER_REMOVE_START_DATE_EDITING) and self.hide_start_date_field:
+            self.fields['start'].widget = forms.HiddenInput()
```
michaelroytman, line 375: If waffle switch is enabled and we should hide the start date field (i.e. on the edit course run page and NOT on the course run create page), hide the start date input.

course_discovery/apps/publisher/models.py
```diff
@@ -503,6 +505,19 @@ def discovery_counterpart(self):
             )
             return None
 
+    @cached_property
+    def discovery_counterpart(self):
+        try:
+            discovery_course = self.course.discovery_counterpart
+            return discovery_course.course_runs.get(key=self.lms_course_id)
```
michaelroytman, line 512: Compared to `discovery_counterpart_latest`, this function attempts to get the corresponding course run from course_metadata by matching on the key.

course_discovery/apps/publisher/views.py
```diff
@@ -803,7 +805,7 @@ def _initialize_run_form(self, last_run=None):
         run_initial_data = {}
         if last_run:
             run_initial_data = {'pacing_type': last_run.pacing_type_temporary}
-        return self.run_form(initial=run_initial_data)
+        return self.run_form(initial=run_initial_data, hide_start_date_field=False)
```
michaelroytman, line 808: This is in the `CreateCourseRunView` class. Since this view and the `CourseRunEditView` use the same Django form, we have to communicate to the form when to hide the start input. When on the `CreateCourseRunView`, we want to show the start input, so we pass `hide_start_date_field=False`.

course_discovery/apps/publisher/views.py
```diff
@@ -1004,7 +1007,10 @@ def get(self, request, *args, **kwargs):
 
         context['course_user_role'] = course_user_role
         context['run_form'] = self.run_form(
-            instance=course_run, is_project_coordinator=context.get('is_project_coordinator')
+            instance=course_run,
+            is_project_coordinator=context.get('is_project_coordinator'),
+            hide_start_date_field=True,
```
michaelroytman, line 1012: This is in the `CourseRunEditView` class. Since this view and the `CreateCourseRunView` use the same Django form, we have to communicate to the form when to hide the start input. When on the `CourseRunEditView`, we want to hide the start input, so we pass`hide_start_date_field=True`.

course_discovery/apps/publisher/views.py
```diff
@@ -1004,7 +1007,10 @@ def get(self, request, *args, **kwargs):
...
+            initial={'start': course_run.start_date_temporary}
```
michaelroytman, line 1013: Although the start input is hidden on the course run edit form, it's still part of the form and is subject to Django's form validation. Therefore, it cannot be empty. We pre-populate the hidden input field with the value of the `start_date_temporary` property on the Publisher CourseRun model to satisfy the validation.

course_discovery/apps/publisher/models.py
```diff
@@ -491,7 +493,7 @@ def get_pacing_type_temporary_display(self):
             return self.get_pacing_type_display()
 
     @cached_property
-    def discovery_counterpart(self):
+    def discovery_counterpart_latest(self):
```
schenedx: How about `discovery_counterpart_latest_by_start_date`

schenedx (approved):
Just a #NIT suggestion on the name of the function. Rest LGTM

course_discovery/apps/publisher/forms.py
fysheets: How does the hide_start_date_field get set?

course_discovery/apps/publisher/models.py
```diff
@@ -516,11 +531,23 @@ def start_date_temporary(self):
...
+        if waffle.switch_is_active(PUBLISHER_REMOVE_START_DATE_EDITING):
+            discovery_counterpart = self.discovery_counterpart
+
+            if discovery_counterpart and discovery_counterpart.start:
+                return discovery_counterpart.start
+            else:
+                # fall back to Publisher model start value, since there is no default
+                return self.start
+        else:
+            return self.start
```
fysheets: Can we refactor this to have a single return at the end? I.e something like  `start_date = self.start ... (in if) start_date = discovery_counterpart.start ... (end of function) return start_date`

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -181,14 +185,17 @@ def test_discovery_counterpart_success(self):
         organization = OrganizationFactory()
 
         discovery_course = self.create_discovery_course_with_partner(organization.partner)
-        discovery_course_run = self.create_discovery_course_run_with_course_and_pacing_type(
-            discovery_course, pacing_type_test_value
+        discovery_course_run = self.create_discovery_course_run_with_metadata(
+            discovery_course,
+            {'pacing_type': pacing_type_test_value}
```
fysheets, line 190: Question around why this change is necessary for general understanding

course_discovery/apps/publisher/views.py
fysheets: Ah-ha this answers my first question from this review

PR #1595, opened 2018-10-29
Fix elasticsearch endpoint for Persons

mikix:
The new "people" elasticsearch query endpoint doesn't work.
 
1. All people are returned for any query: https://prod-edx-discovery.edx.org/api/v1/search/people/?q=complicatedquery
1. 500 errors for an "all" query that hits a person: https://prod-edx-discovery.edx.org/api/v1/search/all/?q=Arcia
1. The profile image field is serialized as "get_profile_image_url" instead of just "profile_image_url"
1. When people are included in an "all" query, there is no way to tell what sort of object it is (wasn't setting the aggregation_key or content_type fields).

This PR should fix all of those. The endpoint still requires exact matches on a field. But that's not new with our elasticsearch endpoints.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1405,24 +1405,29 @@ class CourseRunFacetSerializer(BaseHaystackFacetSerializer):
 
 
 class PersonSearchSerializer(HaystackSerializer):
+    profile_image_url = serializers.SerializerMethodField()
+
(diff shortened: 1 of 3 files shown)


staubina (approved):
The structure of this makes sense to me, but I think we should have someone with a little more context review this.

mikix:
Talked to Farhanah, she said this should be fine. I'll land it.

PR #1601, opened 2018-11-06
DISCO-256: Adds in OFAC serialization and views

jlajoie:


staubina (approved):
This seems like a straightforward change that will enable to simple workflow for this text.

PR #1604, opened 2018-11-09
Disco-224: Improving warning messages and removing length validation …

Dillon-Dumesnil:
…from most fields on course edit page


course_discovery/apps/publisher/templates/publisher/course_edit_form.html
```diff
@@ -198,7 +198,7 @@ <h1 class="hd-1 emphasized">{% trans "Edit Course" %}</h1>
                                         <input id="id_short_description_revision" type="hidden" value="{{ history_object.short_description }}" class="history">
                                         <label class="field-label ">{{ form.short_description.label_tag }} <span class="required">*</span></label>
                                         {{ form.short_description }}
-                                        <p>{% trans "255 character limit, including spaces." %}</p>
+                                        <p class="recommended-text" id="id_short_description_recommended_character_limit" data-character-limit=255><span class="icon fa fa-warning"></span>{% trans "Recommended 255 character limit, including spaces." %}</p>
                                             {% if form.short_description.errors %}
```
Dillon-Dumesnil: Do you know if I'm able to remove these if statements? I think the only time this part would execute is if there was a ValidationError caused by the validate_text_count function that we now don't use on any of these fields.

jlajoie: You likely can, because with no validators and no required=True there's nothing that can throw an error from these fields now.

Dillon-Dumesnil: Okay, sounds good. I wasn't sure about the "nothing that can throw an error". I'll update the PR

jlajoie (approved):
Nice 👍

PR #1609, opened 2018-11-15
Disco-314: Pull in social links from drupal, add detailed urls that handle titles, temporarily remove social links from publisher UI to make transfer and maintaining data easier, and switch to preferring drupal's image url

Dillon-Dumesnil:


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1276,13 +1276,14 @@ def setUp(self):
...
-            'profile_image_url': self.person.profile_image.url,
+            'profile_image_url': self.person.profile_image_url,
             'position': PositionSerializer(position).data,
             'works': [work.value for work in self.person.person_works.all()],
             'urls': {
                 'facebook': None,
                 'twitter': None,
-                'blog': None
+                'blog': None,
+                'others': None,
```
mikix: Yeah boiii, fix them trailing commas.

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -188,7 +188,8 @@ def _person_data(self):
             'urls': {
                 'facebook': 'http://www.facebook.com/hopkins',
                 'twitter': 'http://www.twitter.com/hopkins',
-                'blog': 'http://www.blog.com/hopkins'
+                'blog': 'http://www.blog.com/hopkins',
+                'others': 'http://www.others.com/hopkins',
```
mikix: Hmm, others.com doesn't seem to be a thing. Blog.com is, but is private? Prompted for a username/password.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -337,6 +338,26 @@ def set_position(self, person, data):
         except:  # pylint: disable=bare-except
             logger.exception('Failed to set position for person with UUID [%s]!', uuid)
 
+    def set_social_network(self, person, data):
+        # Used for error messages
+        uuid = data['uuid']
+        try:
+            data = data.get('field_person_social_links', [])
```
mikix: Super nit: you don't need the default [] if all you are checking is "if data:" since get() will return None normally.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -337,6 +338,26 @@ def set_position(self, person, data):
...
+            if data:
+                for social_link in data:
+                    if social_link['field_person_social_link']:
+                        url = social_link['field_person_social_link']['url']
+                        link_type = social_link["field_person_social_link_type"]
```
mikix: Probably want single quotes here and below (linter might complain otherwise?)

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -337,6 +338,26 @@ def set_position(self, person, data):
...
+                        # remapping from Drupal terminology to discovery
+                        link_type = "others" if link_type == 'generic' else link_type
```
mikix: Might want to defensively prepare for unknown types. Like:

```
if link_type not in SOCIAL_NETWORK_CHOICES:
    link_type = 'others'
```

(Although you can't do that so easily, since CHOICES is a nested tuple.)

This is all temporary code, so maybe such defensiveness is unwarranted. But just saying.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -337,6 +338,26 @@ def set_position(self, person, data):
...
+                        defaults = {'value': url}
+                        PersonSocialNetwork.objects.update_or_create(person=person, type=link_type, defaults=defaults)
```
mikix: One thing that this code (and positions code) do not do is delete old social links that aren't present in Drupal anymore. I don't know how valuable that is to cover here. But it would technically be part of a full solution for this. Since we're doing an initial import and then removing this code, I don't think we need to do it here. But positions should have been doing it.

course_discovery/apps/course_metadata/models.py
```diff
@@ -327,10 +327,10 @@ def full_name(self):
 
     @property
     def get_profile_image_url(self):
-        if self.profile_image and hasattr(self.profile_image, 'url'):
-            return self.profile_image.url
-        else:
+        if self.profile_image_url:
             return self.profile_image_url
+        elif self.profile_image and hasattr(self.profile_image, 'url'):
+            return self.profile_image.url
```
mikix, line 316: I know I said "aw yeah, brevity" but I think relying on default None return behavior is actually too brief here. I think it's better to be explicit that we return None in the else case.

If we weren't about to remove this code too... I would say it would be good to comment what the heck is going on with this method. But I guess we needn't bother.

mikix (approved):
I did not test, only reviewed code. But looks good. Got a few comments, but nothing crazy.

course_discovery/apps/api/v1/tests/test_views/test_people.py
Dillon-Dumesnil: I don't think they are using real urls in any of them. The /hopkins is because the person writing this was in to Westworld Season 1 and Anthony Hopkins was a main character. Just for testing the url is present and agrees with what is saved and that if there is an update, they go away.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
Dillon-Dumesnil: I've verified that the only three types Drupal returns is facebook, twitter, and generic. Happy to change this, but I feel comfortable with the way it is due to the temporary nature of this code. Let me know.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
Dillon-Dumesnil: /shrug I'm not planning on changing it unless you push for it.

course_discovery/apps/api/v1/tests/test_views/test_people.py
mikix: Naw I was just observing. If others.com was like a porn site or something, we may want a different one in our code.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
mikix: Yeah, I think leaving alone is good.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
mikix: Yeah, leaving alone is good here.

course_discovery/apps/api/v1/tests/test_views/test_people.py
Dillon-Dumesnil: 👍

course_discovery/apps/api/serializers.py
```diff
@@ -305,21 +318,23 @@ def validate(self, data):
         validated_data = super(PersonSerializer, self).validate(data)
         validated_data['works'] = self.initial_data.get('works', [])
         validated_data['urls'] = self.initial_data.get('urls')
+        validated_data['urls_detailed'] = self.initial_data.get('urls_detailed')
         return validated_data
 
     def create(self, validated_data):
         position_data = validated_data.pop('position')
         works_data = validated_data.pop('works', [])
         urls_data = validated_data.pop('urls', {})
```
mikix: Do we need urls_data anymore?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1652,30 +1633,35 @@ def validate_partner_programs(cls, partner, programs):
...
+    SOCIAL_NETWORK_CHOICES = (
+        (FACEBOOK, _('Facebook')),
+        (TWITTER, _('Twitter')),
+        (BLOG, _('Blog')),
+        (OTHERS, _('Others')),
+    )
+
+    type = models.CharField(max_length=15, choices=SOCIAL_NETWORK_CHOICES, db_index=True)
+    url = models.CharField(max_length=500)
+    title = models.CharField(max_length=255, null=True)
```
mikix: You generally don't want null=True on Charfields. Should be safe to drop it and just have empty strings.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1661,30 +1642,46 @@ def validate_partner_programs(cls, partner, programs):
...
         verbose_name_plural = 'Person SocialNetwork'
 
         unique_together = (
-            ('person', 'type'),
+            ('person', 'type', 'title'),
         )
         ordering = ['created']
 
+    def __str__(self):
+        return '{type}: {url}'.format(type=self.type, url=self.url)
```
Dillon-Dumesnil: @mikix Any thoughts on how to represent the string now that we have the inclusion of titles?

mikix: Maybe just "{title}: {url}"? This is just for quick debug situations where you're trying to print(row) or something yeah?

course_discovery/apps/api/serializers.py
```diff
@@ -279,15 +280,27 @@ def get_social_network_url(self, url_type, obj):
...
         if social_networks:
-            return social_networks[0].value
+            return social_networks[0].url
 
     def get_urls(self, obj):
         return {
             PersonSocialNetwork.FACEBOOK: self.get_social_network_url(PersonSocialNetwork.FACEBOOK, obj),
             PersonSocialNetwork.TWITTER: self.get_social_network_url(PersonSocialNetwork.TWITTER, obj),
             PersonSocialNetwork.BLOG: self.get_social_network_url(PersonSocialNetwork.BLOG, obj),
+            PersonSocialNetwork.OTHERS: self.get_social_network_url(PersonSocialNetwork.OTHERS, obj),
```
mikix: Maybe don't add "others" here. Let's just leave old crappy urls list alone. (and the name is confusing as a plural in this specific list)

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: I don't know what all the use cases are, but yeah, probably. Just not sure how much detail it should contain. Like could be something like `type: ___, title: ____, url: ____` :shrug:

course_discovery/apps/api/serializers.py
```diff
@@ -279,15 +280,27 @@ def get_social_network_url(self, url_type, obj):
...
         }
 
+    def get_urls_detailed(self, obj):
+        urls_detailed = []
+        for network in obj.person_networks.all():
```
mikix: Default id-ordering is fine here I guess. If we have UI editing, I don't think that will get messed up? The UI will always show them in id-order yeah?

mikix: Hmm.. Though we don't want id-ordering do we?  Like if I have "facebook1", "twitter", "facebook2" - we probably want to show the two facebooks next to each other? What does Drupal do? I guess we should just do what it does in this exact situation.

course_discovery/apps/api/serializers.py
```diff
@@ -288,6 +289,14 @@ def get_urls(self, obj):
             PersonSocialNetwork.BLOG: self.get_social_network_url(PersonSocialNetwork.BLOG, obj),
         }
 
+    def get_urls_detailed(self, obj):
+        urls_detailed = [{
+            'type': network.type,
+            'title': network.display_title,
+            'url': network.url,
+        } for network in obj.person_networks.all()]
+        return urls_detailed
```
mikix: I mean, you *could* do this with one statement by saying "return [{..." but I'm already happy enough with the compehension.  :)

course_discovery/apps/api/serializers.py
```diff
@@ -307,39 +316,41 @@ class Meta(MinimalPersonSerializer.Meta):
...
-            if value:
-                person_social_networks.append(PersonSocialNetwork(person=person, type=url_type, value=value))
-        PersonSocialNetwork.objects.bulk_create(person_social_networks)
+        # person_social_networks = []
+        # for url_detailed in urls_detailed_data:
+        #     if url_detailed['url']:
+        #         person_social_networks.append(PersonSocialNetwork(
+        #             person=person, type=url_detailed['type'], title=url_detailed['title'], url=url_detailed['url'],
+        #         ))
+        # PersonSocialNetwork.objects.bulk_create(person_social_networks)
```
mikix, line 336: Might be worth a comment here for why we've got this giant commented-out block. But we'll be in there soon enough, that this is just a minor nit comment.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1277,15 +1277,16 @@ def setUp(self):
             'family_name': self.person.family_name,
             'bio': self.person.bio,
             'profile_image': image_field.to_representation(self.person.profile_image),
-            'profile_image_url': self.person.profile_image.url,
+            'profile_image_url': self.person.profile_image_url,
```
mikix: You could change these to "self.person.get_profile_image_url" to be a little future-proof for this kind of thing.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1546,15 +1547,16 @@ def get_expected_data(cls, person, request):
...
             'position': PositionSerializer(person.position).data,
             'works': [],
             'major_works': person.major_works,
             'urls': {
                 'facebook': None,
                 'twitter': None,
-                'blog': None
+                'blog': None,
             },
+            'urls_detailed': [],
```
mikix, line 1596: We don't have any urls_detailed tests for the serializer specifically? I guess the other tests end up testing the serializer too. That's fine.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -345,6 +346,30 @@ def set_position(self, person, data):
...
+                    if social_link['field_person_social_link']:
+                        url = social_link['field_person_social_link']['url']
+                        title = social_link['field_person_social_link']['title']
+                        link_type = social_link['field_person_social_link_type']
+                        # remapping from Drupal terminology to discovery
+                        link_type = 'others' if link_type == 'generic' else link_type
+                        defaults = {'url': url}
+                        PersonSocialNetwork.objects.update_or_create(
+                            person=person, type=link_type, title=title, defaults=defaults
+                        )
```
mikix, line 366: This is going to have problems. If you don't delete all PersonSocialNetwork objects for this person first, you're going to get weird behavior when a drupal user changes titles or link types on existing social links.

But if you do delete all objects first, the manually entered stuff in Publisher will be lost forever. I think that's fine? Same deal as major_works, when Julia said it was alright to overwrite what was in publisher, since the drupal stuff was what was being used.

With one caveat: only delete all objects first if there are any social links defined at all. In cases where publisher has content but drupal doesn't, let's preserve publisher's data (like we do for bio and major_works).

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -1235,6 +1236,34 @@
...
+                'revision_id': '7761',
+                'field_name': 'field_person_social_links',
+                'archived': '0',
+                'uuid': 'da41b1e9-da0b-49e3-b207-f619f166b0e8',
+                'url': 'https://www.edx.org/field-collection/field-person-social-links/5711',
+            },
+            {
+                'field_person_social_link': {
+                    'title': '@Mikix',
+                    'url': 'http://twitter.com/mikix'
```
mikix: I don't love pointing at a random person's live twitter page, even in tests. What if that account becomes like a white nationalist account or something. Better to do like twitter.com/edxonline or something, even though it's more boring.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1661,30 +1642,45 @@ def validate_partner_programs(cls, partner, programs):
...
+    url = models.CharField(max_length=500)
+    title = models.CharField(max_length=255, blank=True)
     person = models.ForeignKey(Person, related_name='person_networks')
 
     class Meta(object):
         verbose_name_plural = 'Person SocialNetwork'
 
         unique_together = (
-            ('person', 'type'),
+            ('person', 'type', 'title'),
```
mikix, line 1668: So I think this is right for now. But we'll have interesting problems once we want to i18n-ize Persons.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1661,30 +1642,45 @@ def validate_partner_programs(cls, partner, programs):
...
 
-class PersonSocialNetwork(AbstractSocialNetworkModel):
+class PersonSocialNetwork(TimeStampedModel):
     """ Person Social Network model. """
+    FACEBOOK = 'facebook'
+    TWITTER = 'twitter'
+    BLOG = 'blog'
+    OTHERS = 'others'
+
+    choices = {
```
mikix: You should probably all-caps that. And differentiate the two fields here (they're both "choices" -- how do I know why we have two of these same named things in 20 years?).

Maybe just remove the second one alltogether and just inline it to the CharField constructor?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1661,30 +1642,45 @@ def validate_partner_programs(cls, partner, programs):
...
-        )
-        ordering = ['created']
+    @property
+    def display_title(self):
+        if self.title:
+            return self.title
+        elif self.type == self.OTHERS:
+            return self.url
+        else:
+            return self.choices[self.type]
```
mikix: :+1: nice @property.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -461,12 +461,10 @@ class Meta:
...
 
-class AbstractSocialNetworkModelFactory(factory.DjangoModelFactory):
-    type = FuzzyChoice([name for name, __ in AbstractSocialNetworkModel.SOCIAL_NETWORK_CHOICES])
-    value = FuzzyText()
-
-
-class PersonSocialNetworkFactory(AbstractSocialNetworkModelFactory):
+class PersonSocialNetworkFactory(factory.DjangoModelFactory):
+    type = FuzzyChoice([name for name, __ in PersonSocialNetwork.SOCIAL_NETWORK_CHOICES])
+    url = FuzzyText()
```
mikix, line 466: You don't need to change anything here since it was like this before, but FuzzyText is clearly not the right type. I bet there's a FuzzyUrl somewhere. It doesn't matter, we never validate this. Just saying.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1210,17 +1210,9 @@ def setUp(self):
     def test_str(self):
         """Verify that a person-social-network is properly converted to a str."""
         self.assertEqual(
-            str(self.network), '{type}: {value}'.format(type=self.network.type, value=self.network.value)
+            str(self.network), '{title}: {url}'.format(title=self.network.display_title, url=self.network.url)
         )
 
-    def test_unique_constraint(self):
```
mikix: This probably still has (some minor) value, just needs to be updated to include title?

course_discovery/apps/publisher/templates/publisher/course_run_detail/_instructor_profile.html
```diff
@@ -55,26 +55,6 @@
             <div class="copy bio pre"></div>
         </div>
 
-        <div class="social-links">
```
mikix, line 58: This section still has value, yeah? I guess it will need a redesign too. I'm fine with hiding for now.

course_discovery/apps/publisher/wrappers.py
```diff
@@ -224,10 +224,14 @@ def course_staff(self):
                 'profile_url': staff.profile_url,
                 'bio': staff.bio,
                 'email': staff.email,
-                'social_networks': {
-                    staff.type: staff.value
+                'social_networks': [
```
mikix, line 227: This is a change in type. The only consumer was the instructor_profile.html page?

course_discovery/apps/api/tests/test_serializers.py
Dillon-Dumesnil: Just writing a function for this? Or is this a nit comment? I feel like ideally once we transition out of Drupal, this should clear up.

course_discovery/apps/api/tests/test_serializers.py
Dillon-Dumesnil: I decided to keep parity with the existing urls. They had them all set to None so I replicated by just adding in urls_detailed as an empty list. Let me know if you want me to add them.

course_discovery/apps/course_metadata/tests/factories.py
Dillon-Dumesnil: https://factoryboy.readthedocs.io/en/latest/fuzzy.html Nothing out of the box and I'm not convinced it's worthwhile to create a custom fuzzy for this since we don't validate anyway.

course_discovery/apps/publisher/templates/publisher/course_run_detail/_instructor_profile.html
Dillon-Dumesnil: It has _some_ value, but I'm not really sure how it would work for people with more than one thing. I feel like it is better to just completely hide everything for now.

course_discovery/apps/api/tests/test_serializers.py
mikix: No, using the existing (badly-named) @property get_profile_image_url here. That is designed to hide the details of where the url comes from that this test is (imho) needlessly concerning itself with.

course_discovery/apps/api/tests/test_serializers.py
mikix: Yeah, I added the comment to that line specifically, but I really just meant adding new tests in that file. But I think the serializer gets a workout elsewhere too.

PR #1612, opened 2018-11-19
Removing all references to CourseRunSocialNetwork

Dillon-Dumesnil:
To kill it with fire in #1609

### DIFF
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -472,13 +472,6 @@ class PersonSocialNetworkFactory(AbstractSocialNetworkModelFactory):
         model = PersonSocialNetwork
 
 
-class CourseRunSocialNetworkFactory(AbstractSocialNetworkModelFactory):
-    course_run = factory.SubFactory(CourseRunFactory)
-
-    class Meta:
-        model = CourseRunSocialNetwork
-
-
 class SeatTypeFactory(factory.django.DjangoModelFactory):
     class Meta(object):
         model = SeatType
--- a/course_discovery/apps/course_metadata/tests/test_signals.py
+++ b/course_discovery/apps/course_metadata/tests/test_signals.py
@@ -3,7 +3,9 @@ import pytest
 from django.apps import apps
 from factory import DjangoModelFactory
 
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -25,7 +25,7 @@ def test_model_change(self, mock_set_api_timestamp):
         # connecting to. We want to test each of them.
         for model in apps.get_app_config('course_metadata').get_models():
             # Ignore models that aren't exposed by the API or are only used for testing.
-            if model in [DataLoaderConfig, SubjectTranslation, TopicTranslation] \
+            if model in [DataLoaderConfig, SubjectTranslation, TopicTranslation, CourseRunSocialNetwork] \
```
Dillon-Dumesnil, line 30: This is so tests don't fail. Will be removed when the model is deleted.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -472,13 +472,6 @@ class Meta:
         model = PersonSocialNetwork
 
 
-class CourseRunSocialNetworkFactory(AbstractSocialNetworkModelFactory):
```
mikix, line 475: You didn't want to drop Abstract version in this PR? (I assume that's in a future PR?)

course_discovery/apps/course_metadata/tests/test_signals.py
mikix: I don't understand this. How is that a symbol that will resolve? It's not imported.

mikix: OK, I understand what's happening here now. But yeah should be imported.

course_discovery/apps/course_metadata/tests/factories.py
Dillon-Dumesnil: It would have been messier to drop it in this PR because the CourseRunSocialNetwork model still exists and it inherits from the Abstract. So once I can delete the CourseRun, I can move the appropriate code to PersonSocialNetwork and also delete the Abstract

mikix: Yeah, this is the factory version though, not the real model. You could probably clean up the abstract factory, if there's only one consumer. But yeah maybe a future PR is better.

mikix (approved):

PR #1613, opened 2018-11-20
Aj/disco 312

staubina:
Moving the OFAC Restriction text selection to Course Run instead of Course.


course_discovery/apps/course_metadata/migrations/0128_courserun_has_ofac_restrictions.py
```diff
@@ -0,0 +1,51 @@
...
+    course_model = apps.get_model('course_metadata', 'Course')
+    course_run_model = apps.get_model('course_metadata', 'CourseRun')
+    for course in course_model.objects.all():
+        course_runs = course_run_model.objects.filter(course=course)
+        for course_run in course_runs:
+            course_run.has_ofac_restrictions = course.has_ofac_restrictions
+            course_run.save()
+
+
+def reverse_migrate_has_ofac_restrictions_data(apps, schema_editor):
```
Dillon-Dumesnil: Can you explain the reason for this function? I don't think I understand it. It seems like since this runs after `migrate_has_ofac_restriction_data`, it won't actually have any effect since the course_run you use to set the course level has_ofac_restrictions would have just been set by that same course.

Dillon-Dumesnil (commented):
Looks good. Just have one question about the migration files.

course_discovery/apps/course_metadata/migrations/0128_courserun_has_ofac_restrictions.py
staubina: @Dillon-Dumesnil 
Take a look at https://docs.djangoproject.com/en/2.1/ref/migration-operations/#django.db.migrations.operations.RunPython

This is the reverse migration function.  If someone needs to revert the migration this method will be run.

Dillon-Dumesnil: @staubina I see. Didn't know about that. That's a cool function. But I don't think it's being properly used here. It is meant to reverse the migration, but you're changing different models. The original migration isn't removing has_ofac_restrictions from Courses so there isn't a need to re-save all of the Courses. I don't think the reverse method is applicable here or in the publisher migration.

staubina: It is reversing the migration of data from Course_run back to Course, in case the data has drifted.  Without a reverse migration method provided you will be unable to reverse a RunPython migration. So if you attempt to reverse the migration using django-admin tools it will fail.

You make a good point though that we are not removing the column, so we might be able to get away with a no op reverse.  Let me look into how to do that with a RunPython migration like this.

staubina: Or if we need to rollback the release it will fail, without a method there.  There may be a way to do this without having the method there.

Dillon-Dumesnil: Oh I think I see what you mean. The point of this method is if we rollback this migration in like a couple weeks and we have saved new ofac restrictions in course runs and need to make sure we don't lose that information. Probably fine as it is now. I was thinking in a short time frame which would kind of render this less useful.

staubina: > The reverse_code argument is called when unapplying migrations. This callable should undo what is done in the code callable so that the migration is reversible.

staubina: There is a `RunPython.noop` that can be added in its stead, but I think you should be cautious when using it.

staubina: @Dillon-Dumesnil Without a method or a RunPython.noop in there when you unapply the migration Django will throw an exception.

Dillon-Dumesnil: I think our use of this reverse is slightly different. Because if we unapply the migration, the whole column is going to be dropped so we wouldn't need to worry about unsaving all of that information and only resaving the potentially unsynched information. I think you did it correctly here. And yeah, maybe but /shrug

staubina: @Dillon-Dumesnil If you are good with the changes please Approve on the PR.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -336,6 +336,19 @@ <h1 class="hd-1 emphasized">
                     </div>
                 </div>
 
+                {% if is_internal_user %}
+                    <div class="field-title">{% trans "OFAC RESTRICTION TEXT" %} <span class="required float-right">Entered after legal review</span></div>
```
jlajoie: nit: I think other parts of this form use the .label call, don't really care but it might be nice to be consistent with it? https://github.com/edx/course-discovery/blob/87b8439f241328da7da97645c56988c795699e1c/course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html#L71

jlajoie (approved):
Single minor nit, still looks :+1:

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
staubina: I do not see .label in this file?  Do you have an example?

I believe the other course edit form has different names.

Dillon-Dumesnil (approved):

jlajoie (approved):

PR #1614, opened 2018-11-21
Add major_works to Person

mikix:
- Adds a new text field 'major_works' to the Person model.
- Migrates data from PersonWork to this new field.
- Pulls in major works data from the marketing site into this new field. (I got Julia Song's blessing for overwriting Publisher data with marketing data if both exist.)
- Connects the Publisher UI to this new field.
- Changes serialization from a 'works' list to a 'major_works' text field (keeping an empty 'works' in there to avoid breaking parsers) 
- Removes all code references to the PersonWork model (in prep for a future PR that removes the model completely).


course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -228,14 +228,12 @@ def _person_data(self):
...
             },
-            'works': ["Delores", "Teddy", "Maive"],
+            'major_works': 'Delores\nTeddy\nMaive',
             'urls': {
                 'facebook': 'http://www.facebook.com/hopkins',
                 'twitter': 'http://www.twitter.com/hopkins',
                 'blog': 'http://www.blog.com/hopkins'
             },
-            'course_runs_staffed': [],
-            'publisher_course_runs_staffed': [],
```
mikix, line 238: I realized that these fields aren't used in this test, whoops. So taking em out in this PR (here and below).

course_discovery/apps/course_metadata/migrations/0130_person_works_to_major_works.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0129_person_major_works'),
+    ]
+
+    operations = [
+        migrations.RunPython(move_works),
+    ]
```
mikix: I *could* do a reversal here that empties out the field... But I figured who cares.

course_discovery/apps/api/v1/tests/test_views/test_people.py
Dillon-Dumesnil: 🙏

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -88,7 +88,7 @@ def test_retrieve(self, django_assert_num_queries):
         """ Verify the endpoint returns the details for a single program. """
         program = self.create_program()
 
-        with django_assert_num_queries(FuzzyInt(57, 2)):
+        with django_assert_num_queries(FuzzyInt(53, 2)):
```
Dillon-Dumesnil, line 91: These are the worst.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -1249,7 +1252,10 @@
             'url': 'https://www.edx.org/sites/default/files/person/image/orgless.jpg',
         },
         'field_person_areas_of_expertise': [],
-        'field_person_major_works': [],
+        'field_person_major_works': {
+            'value': 'Published books:<br/><ul><li>Book1</li><li>Book2</li></ul>',
+            'format': 'standard_html',
```
Dillon-Dumesnil, line 1257: Nice!

Dillon-Dumesnil (commented):
I think everything is good here. My only question is about the way the major_works are designed to be written in the UI and how we view them.

I noticed that when you're looking at the values in test_people, you use newlines. In the mock_data and migration, you add breaks **and** newlines. Then in test_marketing_site, there are only breaks.

I feel like it is probably fine, but it might be easiest for you to explain it to me so I can understand it more clearly and am confident we are always representing the data in the correct way.

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -87,7 +87,7 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
                 <input class="field-input input-text" type="text" id="blog" name="blog"/>
 
                 <label class="field-label" for="majorWorks">{% trans "Major Works" %}
-                    <span class="optional">({% trans "optional - one per line" %} )</span>
+                    <span class="optional">({% trans "optional" %} )</span>
```
Dillon-Dumesnil: Add in ```Must be valid HTML (ex. <br/> to signal a newline)``` to this field and the bio field or something like that.

Nit: Get rid of the parentheses

Dillon-Dumesnil (approved):
Spoke in person regarding earlier concern. After fixes to help text on add / edit instructor page, LGTM

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
mikix: Done.

PR #1619, opened 2018-11-26
Add db_index to key in Course model

BenjiLee:
https://openedx.atlassian.net/browse/DISCO-320


ormsbee:
Don't you need to generate the migration for this?

BenjiLee:
yes it does >.<

ormsbee (approved):

PR #1622, opened 2018-11-27
DISCO-237: Adds in drupal data management command

jlajoie:


course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,158 @@
...
+        self.course_ids = course_ids
+
+        super(CourseMarketingSiteDataLoader, self).__init__(
+            partner, api_url, access_token, token_type, max_workers, is_threadsafe, **kwargs
+        )
+
+    def process_node(self, data):
+        field_course_end_date = data.get('field_course_end_date')
+        end_date = datetime.datetime.fromtimestamp(int(field_course_end_date)) \
+            if field_course_end_date else datetime.datetime.min
```
staubina: nit: use parenthesis rather than the \ at the eol.

staubina: Am I understanding this block of code correctly, If there is no end date, use the earliest datetime?

I would expect no end date to mean it is a course that is not ending? Is that an incorrect assumption?

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,158 @@
...
+                                'Processed course with key [%s] based on the data from courserun [%s]',
+                                course.key,
+                                course_run.key
+                            )
+
+                            # Ingest the course and course runs into Publisher tables via the dataloader methods
+                            logger.info('Processing Course [%s] for Publisher tables', course_run_key)
+                            process_course(course, True)
+                        except AttributeError:
+                            pass
```
staubina, line 65: Should we log this exception?

course_discovery/apps/publisher/migrations/0070_auto_20181128_2021.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('publisher', '0069_drupalloaderconfig'),
+    ]
+
+    operations = [
```
staubina: Please squash the migrations

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+
+        token_type = 'JWT'
+        for partner in Partner.objects.filter(short_code=config.partner_code):
+            try:
+                access_token, __ = EdxRestApiClient.get_oauth_access_token(
+                    '{root}/access_token'.format(root=partner.oidc_url_root.strip('/')),
+                    partner.oidc_key,
+                    partner.oidc_secret,
+                    token_type=token_type
+                )
```
mikix: You can replace this with "partner.access_token"

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+        # and post_delete signals prevents model changes during data loading from
+        # repeatedly invalidating the cache.
+        for model in apps.get_app_config('course_metadata').get_models():
+            for signal in (post_save, post_delete):
+                signal.disconnect(receiver=api_change_receiver, sender=model)
+
+        config = DrupalLoaderConfig.get_solo()
+
+        token_type = 'JWT'
+        for partner in Partner.objects.filter(short_code=config.partner_code):
```
mikix: If I typo the partner_code, might be nice to give an error rather than silent success. Maybe a get works better here?

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+from course_discovery.apps.core.models import Partner
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
+from course_discovery.apps.course_metadata.data_loaders.marketing_site import CourseMarketingSiteDataLoader
+from course_discovery.apps.publisher.dataloader.create_courses import process_course
+from course_discovery.apps.publisher.models import DrupalLoaderConfig
+
+logger = logging.getLogger(__name__)
+
+
+# Override to only be used in this management command
```
mikix: Sorry, I don't follow this comment - can it be more verbose?

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+        self.course_ids = course_ids
+
+        super(DrupalCourseMarketingSiteDataLoader, self).__init__(
+            partner, api_url, access_token, token_type, max_workers, is_threadsafe, **kwargs
+        )
+
+    def process_node(self, data):
+        field_course_end_date = data.get('field_course_end_date')
+        end_date = (datetime.datetime.fromtimestamp(int(field_course_end_date))
+                    if field_course_end_date else datetime.datetime.min)
```
mikix: We're treating courses without end dates as archived here. It seems intentional. I just want to confirm - I'm not sure which courses would be in that state.

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+        is_unarchived = datetime.datetime.now() < end_date
```
mikix, line 39: Super minor nit - I find negative bool variables harder to reason about. Maybe is_active? Or flip the logic and use is_archived. But who cares.

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+        course_run_key = data.get('field_course_id')
+
+        # Make sure the course run key is one we intend to process and it is unarchived
+        if course_run_key in self.course_ids and is_unarchived:
+            if not data.get('field_course_uuid'):
```
mikix, line 44: This whole block could just be replaced with a call to super().process_node(). With a call to process_course() afterward, if process_node succeeded. Which it doesn't tell you right now. But it might be easier to augment process_node with a return value (or move it's logic into a separate function that does) than duplicate all this subtle code.

course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -0,0 +1,142 @@
...
+                    except InvalidKeyError:
+                        logger.error('Invalid course key [%s].', course_run_key)
+
+                    if created:
+                        course.canonical_course_run = course_run
+                        course.save()
+
+                    # Ingest the course and course runs into Publisher tables via the dataloader methods
+                    logger.info('Processing Course [%s] for Publisher tables', course_run_key)
+                    process_course(course, True)
```
mikix, line 88: Process_course just logs a notice if it fails or there are no orgs. Do we want something harder here?

mikix (approved):
Seems good! Got some optional questions/suggestions, but seems like it's all in order.

course_discovery/apps/publisher/management/commands/load_drupal_data.py
jlajoie: There are none where this isn't set, but swapped it to max to process it as "has no end"

PR #1626, opened 2018-11-30
Adding org_marketing_url to Position serializer

Dillon-Dumesnil:


course_discovery/apps/api/serializers.py
```diff
@@ -232,14 +232,22 @@ class Meta(object):
 
 class PositionSerializer(serializers.ModelSerializer):
     """Serializer for the ``Position`` model."""
+    organization_marketing_url = serializers.SerializerMethodField()
```
bderusha, line 235: [SlugRelatedField](https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield) might be a better fit here

PR #1631, opened 2018-12-03
Removing email field from add and edit instructors. Adding small fix …

Dillon-Dumesnil:
…to profile images for instructors to aid in discovery being source of truth


course_discovery/apps/api/serializers.py
```diff
@@ -304,15 +303,9 @@ def get_urls_detailed(self, obj):
             'url': network.url,
         } for network in obj.person_networks.all()]
 
-    def get_email(self, _obj):
-        # For historical reasons, we provide this field. But for privacy reasons, we don't provide a value in this
-        # minimal serializer. It is provided in the full serializer.
-        return None
-
```
mikix: Naw, we want to keep all this stuff - you'd break any current parsers expecting email to be there. Just leave it as null (which means you just have to drop the EmailField line in PersonSerializer).

course_discovery/apps/api/serializers.py
```diff
@@ -365,6 +358,12 @@ def update(self, instance, validated_data):
 
         for attr, value in validated_data.items():
             setattr(instance, attr, value)
+
+        # Overwrite the Drupal profile_image_url (if one exists) with the publisher image
+        # as we move to publisher being source of truth
+        if instance.get('profile_image_url'):
+            setattr(instance, 'profile_image_url', instance.profile_image.url)
```
mikix: Why the get() and setattr() instead of direct references?

course_discovery/apps/core/tests/factories.py
```diff
@@ -27,7 +27,6 @@ class UserFactory(factory.DjangoModelFactory):
     is_active = True
     is_superuser = False
     is_staff = False
-    email = factory.Faker('email')
```
mikix: That's on the core.models.User model factory - might want to leave that.

course_discovery/apps/api/serializers.py
```diff
@@ -273,7 +272,7 @@ class Meta(object):
         model = Person
         fields = (
             'uuid', 'salutation', 'given_name', 'family_name', 'bio', 'slug', 'position',
-            'profile_image', 'partner', 'works', 'urls', 'urls_detailed', 'email', 'profile_image_url', 'major_works',
+            'profile_image', 'partner', 'works', 'urls', 'urls_detailed', 'profile_image_url', 'major_works',
```
mikix: Don't delete it (as above comments mention) from the fields list. But you might want to add it to Meta.read_only_fields, so it can't be updated from the API at all, for safety?

mikix (approved):
Lovely! Nice test.

PR #1634, opened 2018-12-04
Restrict profile image to an exact 110x110 size

staubina:


course_discovery/static/js/publisher/instructors.js
```diff
@@ -126,8 +126,10 @@ function loadSelectedImage (input) {
                 reader.addEventListener("load", function (e) {
                     var image = new Image();
                     image.addEventListener("load", function () {
-                        if (image.width > imageDimension && image.height > imageDimension) {
```
Dillon-Dumesnil, line 129: That's weird that this used to be &&. Would that mean you could upload a 256x110 and it's valid? We are bananas

Dillon-Dumesnil (approved):
Woot!

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -13,7 +13,7 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
                     <input name="staff-image" id="staffImageSelect" type="file" accept="image/*" onchange="loadSelectedImage(this);">
                 </div>
                 <ul>
-                    <li><span class="required"> {% trans "The image dimensions must be less than 110 x 110" %}</span>
+                    <li><span class="required"> {% trans "The image dimensions must be equal to 110x110" %}</span>
```
mikix: While we're here.... we could replace the x letter with a multiplication sign: ×

Does "must be exactly" sound any less engineer-y than "must be equal to"? Maybe not.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -126,8 +126,10 @@ function loadSelectedImage (input) {
...
-                            addModalError(gettext("The image dimensions must be less than 110 x 110"));
+                        if (image.width !== imageDimension || image.height !== imageDimension) {
+                            addModalError(
+                                gettext("The image dimensions must be equal to " + imageDimension + "x" + imageDimension)
```
mikix: Same multiplication sign note, if you're feeling typographically fancy.

mikix (approved):
Looks great! I've got some typography notes, but they are super ignorable.

course_discovery/static/js/publisher/instructors.js
mikix: Yeah... That old code, not so great.

PR #1635, opened 2018-12-04
Updating social link UI and making discovery the source of truth for social links

Dillon-Dumesnil:
DISCO-304


course_discovery/apps/api/serializers.py
```diff
@@ -321,45 +323,49 @@ class Meta(MinimalPersonSerializer.Meta):
...
-        #     if url:
-        #         network, __ = PersonSocialNetwork.objects.get_or_create(
-        #             person=instance, type=url_detailed['type'], title=url_detailed['title'],
-        #         )
-        #         network.url = url
-        #         network.save()
+        active_social_network_ids = [url_detailed['id'] for url_detailed in urls_detailed_data]
+        for network in PersonSocialNetwork.objects.filter(person=instance):
+            if network.id not in active_social_network_ids:
+                PersonSocialNetwork.objects.filter(id=network.id).delete()
```
mikix: I think you can just do "network.delete()" to save on a query here.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -283,36 +283,9 @@ def process_node(self, data):
             person.slug = slug
             person.save()
 
-        self.set_social_network(person, data)
```
mikix, line 286: :fireworks:

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,91 @@ $(document).ready(function () {
     });
 });
 
+$(document).on('click', '#add-social-link-btn', function (e) {
+    // We are prepending new social link ids with the string banana to distinguish between newly
+    // created social links and existing ones. When sending these to the backend, we will be able
+    // to indicate whether an old link should be updated or a new link created.
+    addNewSocialLink('banana' + Math.floor(Math.random()*1000000));
```
mikix: You might want the prefix "new" or something? The topical "banana" reference is going to just be a confusing bit of code a year from now.

Also, we have a one in a million chance of a hitting a duplicate id here. You could do something like:

```
function generateNewSocialLinkName() {
    if (typeof generateNewSocialLinkName.counter == 'undefined' ) {
        generateNewSocialLinkName.counter = 0;
    }
    return 'new' + generateNewSocialLinkName.counter++;
}
```

As a way to get a static variable in javascript (so says the internet).

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,91 @@ $(document).ready(function () {
...
+        url = url || '',
+        socialLinksWrapper = $('#social-links-wrapper'),
+        linkHtml = '<div class="social-link" data-id="' + id + '">\
+                        <label style="display: inline-block" for="social-link-type-' + id + '">' + gettext("Type") +
+                            '<select name="link-type" id="social-link-type-' + id + '">\
+                                <option disabled selected></option>\
+                                <option value="facebook">Facebook</option>\
+                                <option value="twitter">Twitter</option>\
+                                <option value="blog">Blog</option>\
+                                <option value="others">Other</option>\
```
mikix: Blog and other should be gettext-translated. And probably also Facebook and Twitter? I dunno if they have any localized versions, but they might. Maybe so in languages with totally different scripts, like Arabic, they have some translated version.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,91 @@ $(document).ready(function () {
...
+                            </select>\
+                        </label>\
+                        <label style="display: inline-block" for="social-link-title-' + id + '">' + gettext("Title - optional") +
+                            '<input class="field-input input-text" type="text" id="social-link-title-' + id + '"/>\
+                        </label>\
+                        <label style="display: inline-block" for="social-link-url-' + id + '">' + gettext("URL") +
+                            '<input class="field-input input-text" type="text" id="social-link-url-' + id + '"/>\
+                        </label>\
+                        <button class="remove-social-link-btn" type="button"> X </button>\
```
mikix: I hate to be this guy, but you might consider using ✕ or ✖ (multiplication or heavy multipication) here as more symetrical close glyphs. And more resistant to odd font faces on the client side.

mikix (commented):
Haven't yet reviewed the UI yet - will do that today. But some quick code comments below:

course_discovery/static/js/publisher/instructors.js
mikix: Or the fontawesome [close icon](https://fontawesomeicons.com/close), if we use font awesome in discovery? (which looks identical to heavy multiplication to me)

mikix:
In terms of how it looks - it'd be nice if the type and close buttons were as tall as the text boxes. And if we can make the "optional" tag for the title look as grey as the optional tag elsewhere in the form.

But looks great!

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,95 @@ $(document).ready(function () {
     });
 });
 
+var newLinkCount = 1;
```
Dillon-Dumesnil, line 106: Probably a dirty-ish way to do it, but the count will never get particularly high, guarantees we won't see duplicates, and will reset whenever the page is loaded.

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -63,6 +63,15 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
                 <textarea class="field-textarea input-textarea textarea-medium textarea-block" id="bio"
                           name="bio"></textarea>
 
+                <div id="social-links-wrapper">
+                    <label class="field-label">{% trans "Social Links" %}
```
AlasdairSwan, line 67: What form element is this label for?

mikix (approved):
All my concerns have been addressed! Thanks and nice work.

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
```diff
@@ -63,6 +63,15 @@ <h2 class="hd-2 emphasized new-instructor-heading">{% trans "New Instructor" %}<
...
+                        <span class="optional"> {% trans "optional" %}</span>
+                    </label>
+                </div>
+                <button class="btn-base" type="button" id="add-social-link-btn" style="margin-bottom: 20px;">
```
AlasdairSwan: CSS should be used instead of inline styles. You shouldn't need to add ```type="button"``` when using the ```button``` element. Is the ```id``` being  used somewhere (i.e. I'm not sure you need an id here) ?

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,95 @@ $(document).ready(function () {
...
+        socialLinksWrapper = $('#social-links-wrapper'),
+        linkHtml = '<div class="social-link" data-id="' + id + '">\
+                        <label class="social-link-field" for="social-link-type-' + id + '">' + gettext('Type') +
+                            '<select class="social-link-input social-link-select-type" name="link-type" \
+                            id="social-link-type-' + id + '">\
+                                <option disabled selected></option>\
+                                <option value="facebook">' + gettext('Facebook') + '</option>\
+                                <option value="twitter">' + gettext('Twitter') + '</option>\
+                                <option value="blog">' + gettext('Blog') + '</option>\
+                                <option value="others">' + gettext('Other') + '</option>\
```
AlasdairSwan, line 134: Nitpick: ```value="others"``` or ```value="other```.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,95 @@ $(document).ready(function () {
...
+                            </select>\
+                        </label>\
+                        <label class="social-link-field" for="social-link-title-' + id + '">' + gettext('Title') +
+                            '<span class="optional"> - ' + gettext('optional') + '</span>\
+                            <input class="social-link-input field-input input-text" type="text" id="social-link-title-' + id + '"/>\
+                        </label>\
+                        <label class="social-link-field" for="social-link-url-' + id + '">' + gettext('URL') +
+                            '<input class="social-link-input field-input input-text" type="text" id="social-link-url-' + id + '"/>\
+                        </label>\
+                        <button class="remove-social-link-btn fa fa-close" type="button"></button>\
```
AlasdairSwan: No need to set ```type="button"```.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -91,6 +103,95 @@ $(document).ready(function () {
...
+        if (id.includes('new')) id = '';
+
+        socialLinksArray.push({
+            'id': id,
+            'type': type,
+            'title': title,
+            'url': url,
+        });
+    }
+    return [socialLinksArray, error];
```
AlasdairSwan: Why not just return an object;
```
return {
  socialLinks: [],
  error: ''
};
```

course_discovery/static/js/publisher/instructors.js
```diff
@@ -285,6 +386,14 @@ $(document).on('click', '.selected-instructor a.edit', function (e) {
             $('#family-name').val(data['family_name']);
             $('#bio').val(data['bio']);
             $('#majorWorks').val(data['major_works']);
+            for (var i = 0; i < data['urls_detailed'].length; i++) {
+                addNewSocialLink(
+                    data['urls_detailed'][i]['id'],
+                    data['urls_detailed'][i]['type'],
+                    data['urls_detailed'][i]['title'],
+                    data['urls_detailed'][i]['url'],
```
AlasdairSwan: Nitpick: dot notation is more standard when possible (e.g. ```data.urls_detailed.[i].url```).

course_discovery/static/sass/publisher/course_form.scss
```diff
@@ -159,6 +159,29 @@
...
         }
+
+        .social-link-field {
+          display: inline-block;
+        }
+
+        .social-link-input {
+          height: 44px;
+          margin-top: 5px;
+          @include margin-right(5px);
```
AlasdairSwan: Nitpick: ```@include```s should be declared at the top of the selector.

course_discovery/static/js/publisher/instructors.js
```diff
@@ -41,13 +44,22 @@ $(document).ready(function () {
             addModalError(gettext("Please upload a instructor image."));
             return;
         }
+
+        socialLinks = getSocialLinks();
+        urlsDetailed = socialLinks[0];
+        socialLinkError = socialLinks[1];
```
AlasdairSwan: It's weird that you need to know the position in an array to get various types of data. I would be easier to read/understand if ```getSocialLinks``` returned an object.

AlasdairSwan (changes requested):
Nice work in an ```interesting``` code base, just a few suggestions.

course_discovery/apps/publisher/templates/publisher/_add_instructor_popup.html
Dillon-Dumesnil: Talked about in person. Despite this label not corresponding to a form element, it was decided it was not worth switching the element and then fixing the styling.

course_discovery/static/js/publisher/instructors.js
Dillon-Dumesnil: Talked about in person. It is value 'others' because that is what the database expects.

Dillon-Dumesnil:
image

AlasdairSwan (approved):
Why not :shipit:

PR #1636, opened 2018-12-04
Add delete_person_dups management command

mikix:
This command lets us programmatically delete people somewhat safely. It moves any endorsements, programs, and courses they're attached with to another person, specified on the command line.

Use like `./manage.py delete_person_dups --partner=edx uuid1:uuid2` where you want to delete uuid1 and move all of their courses to uuid2.

Multiple partners can be specified, and you can pass --args-from-database to instead grab all the command line args from a database config model (useful for running the command as a Jenkins job or similar).

The only safeguard against deleting the wrong person is that we currently don't let you delete people with a profile_url (i.e. people that are on a marketing site). This will eventually change with v2 of this script. But for now, I'm targetting people that aren't on a marketing site.


course_discovery/apps/course_metadata/management/commands/delete_person_dups.py
```diff
@@ -0,0 +1,153 @@
...
+            ' {}: {}\n'.format(_('Publisher Course Runs'), pinfo.person.publisher_course_runs_staffed.count()) +
+            ' {}: {} ({})\n'.format(_('Target'), pinfo.target.full_name, pinfo.target.uuid)
+        )
+        if not commit:
+            return
+
+        # Move endorsements
+        Endorsement.objects.filter(endorser=pinfo.person).update(endorser=pinfo.target)
+
+        def filter_person(person):
```
Dillon-Dumesnil, line 100: Ooooh. Nested function. Nice

Dillon-Dumesnil (approved):
Nothing bad stands out to me. Reminder to look through transaction.atomic to determine if it should be used at all. If yes, remove one out of the two uses of it.

course_discovery/apps/course_metadata/management/commands/delete_person_dups.py
```diff
@@ -0,0 +1,149 @@
...
+logger = logging.getLogger(__name__)
+
+
+class PersonInfo(object):
+    def __init__(self, partner, uuid, target_uuid):
+        self.person = Person.objects.get(partner=partner, uuid=uuid)
+        self.target = Person.objects.get(partner=partner, uuid=target_uuid)
+
+        if self.person.profile_url:
+            raise CommandError('Cannot delete person {} - they have a marketing profile url'.format(uuid))
```
mikix, line 18: This is not translated on purpose - it's a temporary string, that will be removed once this tool gains the ability to delete Drupal nodes.

PR #1645, opened 2018-12-06
Displaying social links and major works on view instructor on publish…

Dillon-Dumesnil:
…er course run page


Dillon-Dumesnil:
image

Dillon-Dumesnil:
One weird thing is that the `Straight HTML?` is actually `<h1>Straight HTML? </h1>`, and if you hit the copy, it will copy the HTML tags. Not sure what desired behaviour there is.

mikix (approved):
Nice CSS cleanup. I might personally have a little less space between the social links (use a br instead of p maybe?) but that's just a matter of taste. Love it!

Dillon-Dumesnil:
image

PR #1686, opened 2018-12-10
Publishing instructors to Drupal on creation in publisher

Dillon-Dumesnil:


mikix (approved):
You'll need to update translations. But yeah, :shipit:

Dillon-Dumesnil:
That's annoying. It told me it was fine locally. Also, someone merged in between your comment and when I force pushed. Curses

mikix:
Oh I might be wrong then. I had assumed "New" wasn't elsewhere in the source code.

Dillon-Dumesnil:
It's real weird. /shrug The detect_changed_source_translations passed locally, but then fake_translations still added new ones so I dunno. They're there. It should pass?

PR #1693, opened 2018-12-12
Adding areas of expertise to admin and removing profile_url from the

Dillon-Dumesnil:
model in favour of using a property


course_discovery/apps/course_metadata/admin.py
```diff
@@ -59,6 +59,11 @@ class PersonSocialNetworkInline(admin.TabularInline):
     extra = 0
 
 
+class PersonAreaOfExpertiseInline(admin.TabularInline):
+    model = PersonAreaOfExpertise
+    extra = 0
```
mikix, line 64: What does extra do?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -863,7 +863,7 @@ def setUp(self):
         )
         self.course_state.name = CourseStateChoices.Approved
         self.course_state.save()
-        self.course_run.staff.add(PersonFactory(profile_url='http://test-profile'))
+        self.course_run.staff.add(PersonFactory())
```
mikix, line 867: Odd to have specified a previous URL - it wasn't using that in any tests?

mikix (approved):
:+1:

course_discovery/apps/course_metadata/admin.py
Dillon-Dumesnil: I'm not sure. I based this off the PersonSocialNetworkInline found above this. I'll look into it.

course_discovery/apps/publisher/tests/test_views.py
Dillon-Dumesnil: It's not even part of the factory so I'm not really sure what was happening here. I tried to run the tests locally, but they crapped out due to some compression message so I'm using jenkins to test. Based on my understanding though, I think it should be fine without.

mikix: I think the kwargs here go to the model eventually, so the factory doesn't need to override each field. So not being part of the factory is fine, I think. But yeah, maybe the tests just never used that.

course_discovery/apps/course_metadata/admin.py
Dillon-Dumesnil: > def get_extra(self, request, obj=None, **kwargs):
        """Hook for customizing the number of extra inline forms."""
        return self.extra

So I'm still not really sure, but it seems like we would want 0 extra inline forms so I think we did the right thing.

course_discovery/apps/publisher/tests/test_views.py
Dillon-Dumesnil: The test for it is
>      # assert person name appearing with url.
        person = self.course_run.staff.all()[0]
        self.assertContains(
            response, '<a href="{url}" target="_blank">{name}</a>'.format(
                url=person.profile_url,
                name=person.full_name
            )
        )

So I think it should be fine since the one that is in the response should be person.profile_url.

PR #1695, opened 2018-12-13
Stop pulling in Person slugs from marketing site

mikix:
This also means we can drop the entire PersonMarketingSiteDataLoader class.


Dillon-Dumesnil (approved):
🔥

PR #1696, opened 2018-12-13
Remove some extra fields I added to people API that aren't needed anymore

mikix:


course_discovery/apps/api/serializers.py
```diff
@@ -253,8 +253,6 @@ def get_organization_marketing_url(self, obj):
 class MinimalPersonSerializer(serializers.ModelSerializer):
```
Dillon-Dumesnil, line 253: Can we move this all into just a `PersonSerializer`? Or do we think we might have need for the split again in the future?

mikix: I prefer keeping the split just because we might want to differentiate more later and it keeps the writable methods (create/update) on the appropriate version.

Dillon-Dumesnil: :shrug: okay

Dillon-Dumesnil (approved):

mikix:
Hey @bderusha - any objection to this PR's dropping of the requirement that the logged in user belong to a publisher group to view the people API? (email and attached courses are no longer included)

bderusha:
@mikix :shipit:

PR #1698, opened 2018-12-14
Disables post_save signal for load_drupal_data command

jlajoie:


course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -118,6 +119,8 @@ def handle(self, *args, **options):
         for model in apps.get_app_config('course_metadata').get_models():
             for signal in (post_save, post_delete):
                 signal.disconnect(receiver=api_change_receiver, sender=model)
+        # Disable the post save as we do NOT want to publish these courses to Studio
+        post_save.disconnect(receiver=create_course_run_in_studio_receiver, sender=CourseRun)
```
staubina, line 123: Do we need to add it again or does this just disconnect in the scope of the command?

jlajoie: My understanding is it's just the scope, otherwise RCM would permanently kill off  the above lines 119:121 for caching

staubina: Sadly I do not see anything in the Docs about the scope of the disconnect.

staubina (approved):

PR #1699, opened 2018-12-14
Updating delete_person_dups to support only deletions

Dillon-Dumesnil:


mikix (approved):
Looks good!

I might have flipped the meaning of the argument. I think it's generally clearer if arguments to a command add functionality rather than remove it (both for person running it, and code - you could flip most of the "not delete_only" conditions). But it's good as-is.

PR #1702, opened 2018-12-17
Adding analytics api integration models

mikedikan:
DISCO-341

Adding models for integration of analytics api, and a WIP version of the updates to the metadata refresh to take in this data.  The metadata refresh updates are disabled until models are updated.


course_discovery/apps/core/models.py
```diff
@@ -85,6 +85,8 @@ class Partner(TimeStampedModel):
     studio_url = models.URLField(max_length=255, null=True, blank=True, verbose_name=_('Studio URL'))
     site = models.OneToOneField(Site, on_delete=models.PROTECT)
     lms_url = models.URLField(max_length=255, null=True, blank=True, verbose_name=_('LMS URL'))
+    analytics_url = models.URLField(max_length=255, null=True, blank=True, verbose_name=_('Analytics API URL'))
```
bbaker6225: I know the other fields are already doing it, but usually you don't want `null=True` and `blank=True` for char fields, since then you can get both `null` and `''` in the db

course_discovery/apps/core/models.py
```diff
@@ -85,6 +85,8 @@ class Partner(TimeStampedModel):
...
+    analytics_token = models.CharField(max_length=255, null=True, blank=True, verbose_name=_('Analytics Access Token'))
```
bbaker6225: Same thing about `null/blank`

bbaker6225 (commented):
Holding off on reviewing the data loader changes until they're more finalized

course_discovery/apps/core/models.py
mikedikan: so, if I understand - pick one, right?

bbaker6225: I would go with just blank https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.Field.null

mikedikan: Updated to blank

course_discovery/apps/core/models.py
mikedikan: Updated to blank

course_discovery/apps/core/models.py
bbaker6225: 👍 Looks good, just need to regenerate the migration I think

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -161,6 +162,7 @@ def handle(self, *args, **options):
                     with concurrent.futures.ProcessPoolExecutor() as executor:
                         for loader_class, api_url, max_workers in stage:
                             if api_url:
+                                logger.info('Executing Loader [{}]'.format(api_url))
```
mikedikan: These were helpful during my development and seem worthy of leaving in permanently and are not too noisy, but I can leave a comment to remove them later if we think that they should be removed.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -176,6 +178,7 @@ def handle(self, *args, **options):
                 # Flatten pipeline and run serially.
                 for loader_class, api_url, max_workers in itertools.chain(*(stage for stage in pipeline)):
                     if api_url:
+                        logger.info('Executing Loader [{}]'.format(api_url))
```
mikedikan: see above comment

course_discovery/apps/course_metadata/data_loaders/analytics_api.py
```diff
@@ -0,0 +1,109 @@
+import logging
```
mikedikan: Open to the idea of just removing this file and the changes to refresh metadata from this PR and putting them in a separate branch and if it's worth the effort.

dianakhuang (approved):
Looks good to me.

PR #1703, opened 2018-12-17
Order program subjects by how common they are among its courses

inventhouse:
WIP, ~not tested~ _now_ tested! https://openedx.atlassian.net/browse/DISCO-399

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1,7 +1,7 @@
 import datetime
 import itertools
 import logging
-from collections import defaultdict
+from collections import Counter, defaultdict
 from urllib.parse import urljoin
 from uuid import uuid4
 
@@ -1160,9 +1160,19 @@ class Program(TimeStampedModel):
 
     @property
     def subjects(self):
-        subjects = [course.subjects.all() for course in self.courses.all()]
-        subjects = itertools.chain.from_iterable(subjects)
-        return set(subjects)
+        """
+        :return: The list of subjects; the first subject should be the most common primary subjects of its courses,
+        other subjects should be collected and ranked by frequency among the courses.
+        """
+        primary_subjects = []
+        course_subjects = []
+        for course in self.courses.all():
+            subjects = course.subjects.all()
+            primary_subjects.extend(subjects[:1])  # "Primary" subject is the first one
+            course_subjects.extend(subjects)
+        common_primary = [s for s, _ in Counter(primary_subjects).most_common()][:1]
+        common_others = [s for s, _ in Counter(course_subjects).most_common() if s not in common_primary]
+        return common_primary + common_others
 
     @property
     def seats(self):
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -316,7 +316,7 @@ class ProgramIndex(BaseIndex, indexes.Indexable, OrganizationsMixin):
         return self.prepare_authoring_organizations(obj) + self.prepare_credit_backing_organizations(obj)
 
     def prepare_subject_uuids(self, obj):
-        return set([str(subject.uuid) for course in obj.courses.all() for subject in course.subjects.all()])
+        return [str(subject.uuid) for subject in obj.subjects]
 
     def prepare_staff_uuids(self, obj):
         return set([str(staff.uuid) for course_run in obj.course_runs for staff in course_run.staff.all()])
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -626,9 +626,9 @@ class ProgramTests(TestCase):
     def setUp(self):
         super(ProgramTests, self).setUp()
         transcript_languages = LanguageTag.objects.all()[:2]
-        subjects = factories.SubjectFactory.create_batch(2)
+        self.subjects = factories.SubjectFactory.create_batch(3)
         self.course_runs = factories.CourseRunFactory.create_batch(
-            3, transcript_languages=transcript_languages, course__subjects=subjects,
+            3, transcript_languages=transcript_languages, course__subjects=self.subjects,
             weeks_to_complete=2)
         self.courses = [course_run.course for course_run in self.course_runs]
         self.excluded_course_run = factories.CourseRunFactory(course=self.courses[0])
@@ -999,13 +999,21 @@ class ProgramTests(TestCase):
         self.assertGreater(len(actual_transcript_languages), 0)
         self.assertEqual(actual_transcript_languages, expected_transcript_languages)
 
-    def test_subjects(self):
-        expected_subjects = itertools.chain.from_iterable([list(course.subjects.all()) for course in self.courses])
-        expected_subjects = set(expected_subjects)
-        actual_subjects = self.program.subjects
+    def test_subject_order(self):
+        """
+        Verify the program's subjects are in order of frequency among courses, with primary subjects coming first
+        """
+        course1 = factories.CourseFactory(subjects=self.subjects[:1])  # A
+        course2 = factories.CourseFactory(subjects=self.subjects[:2])  # A, B
+        course3 = factories.CourseFactory(subjects=self.subjects[::-1])  # C, B, A
 
-        self.assertGreater(len(actual_subjects), 0)
-        self.assertEqual(actual_subjects, expected_subjects)
+        program1 = factories.ProgramFactory(courses=[course1, course2, course3])
+        self.assertEqual(program1.subjects, self.subjects)
+
+        course4 = factories.CourseFactory(subjects=self.subjects[::-2])  # C, A; these make C the most common primary
+        course5 = factories.CourseFactory(subjects=self.subjects[::-2])  # C, A; and A the most common overall in p2
+        program2 = factories.ProgramFactory(courses=[course1, course2, course3, course4, course5])
+        self.assertEqual(program2.subjects, [self.subjects[2], self.subjects[0], self.subjects[1]])
 
     def test_start(self):
         """ Verify the property returns the minimum start date for the course runs associated with the


dianakhuang:
This looks good to me, but I do think we want a test, maybe in https://github.com/edx/course-discovery/blob/a3eb88def3e3fa93cf1ab54bb9f215b68d529906/course_discovery/apps/api/tests/test_serializers.py ?

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -316,7 +318,17 @@ def prepare_organizations(self, obj):
         return self.prepare_authoring_organizations(obj) + self.prepare_credit_backing_organizations(obj)
 
     def prepare_subject_uuids(self, obj):
-        return set([str(subject.uuid) for course in obj.courses.all() for subject in course.subjects.all()])
+        # Program primary subject should be the most common primary subject of its courses; other subjects should be
```
mikedikan: Ben, do we have any idea how much time this will add to search index creation?

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -316,7 +318,17 @@ def prepare_organizations(self, obj):
...
+        # collected and ranked by frequency among the courses
+        primary_subjects = []
+        course_subjects = []
+        for course in obj.courses.all():
+            uuids = [str(subject.uuid) for subject in course.subjects.all()]
+            primary_subjects.extend(uuids[:1])  # "Primary" subject is the first one
+            course_subjects.extend(uuids)
+        common_primary = [uuid for uuid, _ in Counter(primary_subjects).most_common()][:1]
```
mikedikan: If I understand this correctly, if a program has 6 courses, and 2 courses have a primary subject in common, it would outrank a subject that they all share in common, right?  I think that may be the right answer anyway, but want to make sure i follow the logic.

bbaker6225:
@inventhouse 
I might be missing something with how this connects, but don't we need to update https://github.com/edx/course-discovery/blob/dd707f8453f39dad75e8a437e12a1cd7c2e19ed8/course_discovery/apps/course_metadata/models.py#L1160-L1164 to see the subject ordering reflected in the endpoint that prospectus hits?

The change to search_indexes doesn't do anything to /api/v1/programs for me locally, but changing the subjects property on the program model does

course_discovery/apps/course_metadata/search_indexes.py
inventhouse: Should be negligible; it was already iterating the program --> courses --> subjects relationships, we're just doing slightly more with the results.

course_discovery/apps/course_metadata/search_indexes.py
inventhouse: Yes, if the other four courses all have different primary subjects; perhaps worse, if all the courses have different primary subjects, this will just arbitrarily pick one of those and _not_ a secondary subject most or all of them share.  

However, I'm guessing it will be more common for most of the courses in a program to have a particular primary subject and that subject will be usually be the best one for the program, even if they also share a more general secondary subject.

I could work out a hybrid that uses the most dominant of the two weightings, but I felt that was getting too complicated.

inventhouse: _But_, the algorithm implemented above does guarantee that the program's primary subject is the primary subject of at least one of its courses; it won't "come from nowhere"

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -1,11 +1,13 @@
 import json
+from collections import Counter
 
 from haystack import indexes
 from opaque_keys.edx.keys import CourseKey
 
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import Course, CourseRun, Degree, Person, Position, Program
 
+
```
mikedikan: unintentional space added?

course_discovery/apps/course_metadata/search_indexes.py
mikedikan: I think this is fine - the important thing is that the remedy to the above scenario is (relatively) easy to communicate and implement

inventhouse:
@bbaker6225 good catch, we were having a hard time finding where this should be done; sounds like we need to port it here.  Any idea if we still need it in the search index method? (ie does the search index care about or even preserve the order?)

course_discovery/apps/course_metadata/search_indexes.py
inventhouse: Intentional; two blanks around top-level functions/classes/etc (pep8 doesn’t take it far enough, I prefer two blanks around all functions)

bbaker6225:
I don't think you need it in the search index method. It's probably fine in just the program model property. If we want to sort the subject uuids in the search endpoint, we can add that at another time

inventhouse:
> I don't think you need it in the search index method. It's probably fine in just the program model property. If we want to sort the subject uuids in the search endpoint, we can add that at another time

@bbaker6225 Also, if this logic is in the programs model, why would the search index be iterating the course subjects itself?  Should I change that to just be `[str(subject.uuid) for subject in obj.subjects]`?

bbaker6225:
That's a good question, I don't know why it's doing that. I don't see any reason why it wouldn't just be what you said

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -999,14 +999,27 @@ def test_transcript_languages(self):
         self.assertGreater(len(actual_transcript_languages), 0)
         self.assertEqual(actual_transcript_languages, expected_transcript_languages)
 
-    def test_subjects(self):
-        expected_subjects = itertools.chain.from_iterable([list(course.subjects.all()) for course in self.courses])
-        expected_subjects = set(expected_subjects)
-        actual_subjects = self.program.subjects
+    def test_subject_sets(self):
```
dianakhuang: Part of me is questioning the utility of this test, since the next one tests a superset of this functionality.

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -316,7 +316,7 @@ def prepare_organizations(self, obj):
         return self.prepare_authoring_organizations(obj) + self.prepare_credit_backing_organizations(obj)
 
     def prepare_subject_uuids(self, obj):
-        return set([str(subject.uuid) for course in obj.courses.all() for subject in course.subjects.all()])
+        return [str(subject.uuid) for subject in obj.subjects]
```
dianakhuang, line 319: 👍 👍 👍

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -999,14 +999,27 @@ def test_transcript_languages(self):
...
+        # expected_subjects = itertools.chain.from_iterable([list(course.subjects.all()) for course in self.courses])
```
dianakhuang: If we do keep this test around, don't forget to remove this comment.

dianakhuang (commented):
Overall, it looks good!

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -999,14 +999,27 @@ def test_transcript_languages(self):
...
+        expected_subjects = set(self.subjects)
+        actual_subjects = set(self.program.subjects)
```
bbaker6225: This method seems like it could just be
```python
self.assertEqual(set(self.actual_subjects), set(self.program.subjects))
```
I don't think testing the length of the subjects field gives us much

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -999,14 +999,27 @@ def test_transcript_languages(self):
...
 
         self.assertGreater(len(actual_subjects), 0)
         self.assertEqual(actual_subjects, expected_subjects)
 
+    def test_subject_order(self):
```
bbaker6225, line 1002: It would be great to get a short docstring on these tests to more clearly state what they're each testing

course_discovery/apps/course_metadata/search_indexes.py
bbaker6225, line 319: Nice

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -999,14 +999,27 @@ def test_transcript_languages(self):
...
+        c1 = factories.CourseFactory(subjects=self.subjects[:1])  # A
```
bbaker6225: Nit: I'd prefer we called these variables course2, program2, etc. to be a little clearer, but not a big deal

course_discovery/apps/course_metadata/tests/test_models.py
inventhouse: Yup, I had updated the original test, but with the new one, it's not needed; I'll just drop it.

course_discovery/apps/course_metadata/tests/test_models.py
inventhouse: I had forgotten it, but dropping the whole thing.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1160,9 +1160,17 @@ def transcript_languages(self):
 
     @property
     def subjects(self):
-        subjects = [course.subjects.all() for course in self.courses.all()]
-        subjects = itertools.chain.from_iterable(subjects)
-        return set(subjects)
+        # Program primary subject should be the most common primary subject of its courses; other subjects should be
```
bbaker6225: Could you also add a short docstring for this property/method? Something like `returns a list of the program's subjects, in order of frequency among courses, with primary subjects coming first`

course_discovery/apps/course_metadata/tests/test_models.py
inventhouse: I'm going to drop this test, since it doesn't confirm anything that the other one doesn't also prove

course_discovery/apps/course_metadata/tests/test_models.py
inventhouse: Will do

course_discovery/apps/course_metadata/tests/test_models.py
inventhouse: Pep8 would agree with you (if-and-only-if a variable is super-local like this I prefer the super-abbreviated version, but we're a pep8 shop)

course_discovery/apps/course_metadata/models.py
inventhouse: Will do

dianakhuang (approved):

bbaker6225 (approved):

PR #1705, opened 2018-12-18
Use a unicode-capable slugify

mikix:
商汤科技研究 -> shang-tang-ke-ji-yan-jiu

This adds a new dependency on [unicode-slugify](https://pypi.org/project/unicode-slugify/). This is a project of Mozilla's (so it seems like a safe dependency) which further depends on [unidecode](https://pypi.org/project/Unidecode/) for the bulk of its work.


Dillon-Dumesnil (approved):
Can you add in a test to demonstrate behaviour? Good fix!

mikix:
Tests! Yes, a reasonable request. On it.

PR #1708, opened 2018-12-18
Updated the serializer to save the correct profile_image url

staubina:
[DISCO-400]

This change is required to ensure we have the updated file name and path for the image after saving it in the data store and model.


mikix (approved):
Seems fine to me!

Maybe worth a test? But maybe that's too much.

PR #1709, opened 2018-12-19
Aj/disco 283 url priority

staubina:


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -518,22 +518,23 @@ def test_unicode_slug(self):
 
     def test_get_profile_image_url(self):
         """
-        Verify that property returns profile_image_url if profile_image_url
-        exists other wise it returns uploaded image url.
+        Verify that property returns profile_image_url, which should always be the
+        profile_image.url and never profile_image_url from the DB.
         """
-        self.assertEqual(self.person.get_profile_image_url, self.person.profile_image_url)
+        print(self.person.profile_image.url + ' == ' + self.person.get_profile_image_url)
```
mikix: That looks like a leftover debug print?

mikix (approved):
Looks good!

course_discovery/apps/course_metadata/models.py
```diff
@@ -311,12 +311,9 @@ def profile_url(self):
...
     @property
     def get_profile_image_url(self):
-        # self.profile_image_url comes from Drupal which is currently the source
-        # of truth so we want to prefer that over self.profile_image.url which
-        # comes from discovery
-        if self.profile_image_url:
-            return self.profile_image_url
-        elif self.profile_image and hasattr(self.profile_image, 'url'):
+        # Discovery and Publisher are now the source of truth for images so we should
+        # use the image from Publisher over the one stored in Discovery.
```
mikix: "the image from Discovery over the one stored in Drupal" maybe?

course_discovery/apps/course_metadata/tests/test_models.py
Dillon-Dumesnil: Did I do that? Probably. Oops

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1298,7 +1298,6 @@ def test_data(self):
         self.assertDictEqual(serializer.data, self.expected)
 
     def test_profile_image_url_override(self):
```
Dillon-Dumesnil: Can we just delete this test? We don't support profile_image_url anymore so don't need to worry about overrides.

course_discovery/apps/course_metadata/models.py
```diff
@@ -311,12 +311,9 @@ def profile_url(self):
 
     @property
     def get_profile_image_url(self):
```
Dillon-Dumesnil, line 313: Is this property still necessary? If we are only ever returning the publisher image or none, it seems like we could remove this property. If not, the new comment seems misleading. We aren't giving the option of using the image from discovery _over_ the image from Drupal, we just use the image from discovery or nothing.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -518,22 +518,22 @@ def test_unicode_slug(self):
 
     def test_get_profile_image_url(self):
```
Dillon-Dumesnil, line 519: Similar to above, if we can remove the property, let's remove this test. I believe profile_image.url is tested elsewhere.

Dillon-Dumesnil (commented):
Couple comments about if it is necessary to continue having any references to profile_image_url.

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -484,7 +484,9 @@ def test_update_without_position(self):
         self.assertEqual(updated_person.position.title, data['position']['title'])
 
     def test_profile_image_url_not_set(self):
```
Dillon-Dumesnil: Look into deleting

staubina:
@Dillon-Dumesnil I have removed the use of profile_image_url in other areas of the code as requested.

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -78,7 +78,7 @@ def get_result_label(self, result):
                 'full_name': result.full_name,
                 'position': result.position if hasattr(result, 'position') else None,
                 'organization_id': result.position.organization_id if hasattr(result, 'position') else None,
-                'can_edit_instructor': not result.profile_image_url,
+                'can_edit_instructor': not result.get_profile_image_url,
```
Dillon-Dumesnil, line 81: Without looking into this, that's a weird way to determine a field called `can_edit_instructor`

course_discovery/apps/api/tests/test_serializers.py
Dillon-Dumesnil: Bump. If we are removing uses of profile_image_url, it feels weird to keep a test where we are updating self.expected['profile_image_url']

Dillon-Dumesnil (approved):
One question about a test. Otherwise all good.

course_discovery/apps/course_metadata/lookups.py
mikix: That feels like a throwback to when the _url field determed `is_new` by being a proxy for whether the person had been edited in drupal.

mikix (approved):
Looks good.

course_discovery/apps/course_metadata/lookups.py
mikix: Maybe can_edit_instructor is always True now? I don't know context for this variable.

staubina: So I did some research.

This will only be true if the user does not have a profile_image.

It also does not seem to do anything.  So I think it is safe to leave it and chalk it up to being an oddity of the existing codebase.  To be deleted when we deprecate this version of Publisher.

PR #1710, opened 2018-12-19
Add hint to course teams about instructor ordering.

mikix:
https://openedx.atlassian.net/browse/DISCO-288


mikix:
@Dillon-Dumesnil is that a clear sentence?

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -256,6 +256,7 @@ <h1 class="hd-1 emphasized">
                 <div class="row">
                     <div class="col col-6 help-text">
                         <div>{% trans "The primary instructor or instructors for the course." %}</div>
+                        <div>{% trans "The order that instructors are entered here is the same order used everywhere else." %}</div>
```
Dillon-Dumesnil: @mikix I'm not in love with it.

> The order that instructors are entered here is the same order they will be displayed on the marketing site.
> The order that instructors are entered here is the same order they will be displayed on the course about pages.
> The order that instructors are entered here is the same order they will be displayed within the courseware.

Dillon-Dumesnil (approved):
:bravo:

PR #1711, opened 2018-12-19
Add status field in django admin list for CourseRun

mikix:
Easier for marketing folks to see status at a glance.


Dillon-Dumesnil (approved):
👏

PR #1713, opened 2018-12-20
Adding published to Person model and pulling in from data loader

Dillon-Dumesnil:


course_discovery/apps/api/v1/views/people.py
```diff
@@ -42,23 +41,18 @@ def create(self, request, *args, **kwargs):
...
                 'An error occurred while adding the person [%s]-[%s] to the marketing site.',
                 serializer.validated_data['given_name'], serializer.validated_data['family_name']
             )
             return Response('Failed to add person data to the marketing site.', status=status.HTTP_400_BAD_REQUEST)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(
+                'An error occurred while adding the person [%s]-[%s] in discovery.',
+                serializer.validated_data['given_name'], serializer.validated_data['family_name'],
+            )
+            return Response('Failed to add person data.', status=status.HTTP_400_BAD_REQUEST)
```
mikix, line 54: This is fine if you prefer it, but I can see an argument for not having two exception blocks. logger.exception will print the exception, so they'll know whether the fault was marketing or discovery from that. The only difference in these two blocks is a slight difference in log message.

course_discovery/apps/api/v1/views/people.py
```diff
@@ -106,6 +81,12 @@ def update(self, request, *args, **kwargs):  # pylint: disable=unused-argument
                 'Failed to update person data on the marketing site.',
                 status=status.HTTP_400_BAD_REQUEST
             )
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(
+                'An error occurred while updating the person [%s]-[%s] in discovery.',
+                serializer.validated_data['given_name'], serializer.validated_data['family_name']
+            )
+            return Response('Failed to update person data.', status=status.HTTP_400_BAD_REQUEST)
```
mikix, line 88: Same exception comment here.

course_discovery/apps/course_metadata/migrations/0141_person_published.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('course_metadata', '0140_remove_profile_url_and_fix_admin'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='person',
+            name='published',
+            field=models.BooleanField(default=True),
```
mikix: Can't do that - all existing people will be marked as published. You might be able to use preserve_default=False here to do some trickery with the default value? I've never used that flag though.

But I might say it's better to have a model default of False and just set it to True when creating 'em.

mikix:
Refresh Course Metadata is going to slow down a lot too here - we're going to pull status, call save(), then push status for each person. Might want to add a flag to save() to disable the push?

course_discovery/apps/api/v1/views/people.py
Dillon-Dumesnil: > An error occurred while adding the person [%s]-[%s] to the marketing site or in discovery.
?

I liked it because it clearly stated whether the issue was marketing or discovery related and I believe the log it prints to will be the same for either.

mikix (approved):
Still seems good!

PR #1716, opened 2018-12-21
Adding enrollment model fields

mikedikan:
DISCO-342

Adding course run, course, and program model fields for enrollment data - total count and last 6 months.


bbaker6225 (approved):
LGTM 👍

course_discovery/apps/course_metadata/migrations/0141_auto_20181221_1339.py
```diff
@@ -0,0 +1,45 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='course',
+            name='enrollment_count',
+            field=models.IntegerField(default=0, help_text='Total number of learners who have enrolled in this course'),
+        ),
+        migrations.AddField(
+            model_name='course',
+            name='enrollment_count_6_months',
```
inventhouse: Do we want to have _6_months in the name?  If we change the timeframe, we'll have to drop the fields and create new ones; maybe that's easy and better than keeping a more generic field if we change the code around it, I don't know.

mikedikan: I'm open to changing it - perhaps recent_enrollment_count?

inventhouse: I'm asking for my own edification as much as anything; I honestly don't know how hard/costly/etc it is to drop and replace a column like that if we decide to change it.  

If that's easy, I'd say be specific in the naming, otherwise be a bit more generic (like `recent_enrollment_count`) to be more future-proof.

mikedikan: done!

mikedikan:
Ok, got tests fixed - this needs another look @bbaker6225 & @inventhouse

inventhouse (approved):

PR #1717, opened 2018-12-21
Remove profile_image_url from the person model, it is no longer neeed

staubina:
[DISCO-283]

As part of the work to make course_metadata the source of truth,
we have moved all profile_images to the Discovery data store and
are now using the profile_image field exclusively to store the
URL and information.


course_discovery/apps/api/serializers.py
```diff
@@ -290,6 +290,9 @@ def get_social_network_url(self, url_type, obj):
         if social_networks:
             return social_networks[0].url
 
+    def get_profile_image_url(self, obj):
+        return obj.get_profile_image_url
+
```
mikix, line 295: Why this change? Same result right?

mikix (approved):

course_discovery/apps/api/serializers.py
staubina: It was having issues returning the correct url with the old serializer. This update also matches the approach in another serializer below this one.

PR #1718, opened 2018-12-26
Add tags to the Course model.

dianakhuang:
https://openedx.atlassian.net/browse/DISCO-407


mikedikan (approved):

inventhouse (commented):
Looks like Mike's migration changes got scooped up in this diff too?

dianakhuang:
@inventhouse we discussed it, and it seems like it's just adding `blank=True`. It's a little annoying to have these changes split over two migrations, but I don't think it's a huge deal.

inventhouse:
> @inventhouse we discussed it, and it seems like it's just adding `blank=True`. It's a little annoying to have these changes split over two migrations, but I don't think it's a huge deal.

Huh, ok.

inventhouse (approved):

PR #1722, opened 2019-01-03
Courses for a subject [DISCO-445]

inventhouse:
~No unit tests yet,~ only accepts one subject and doesn't limit query to the primary (first) subject of each course - now with test!

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -52,9 +52,13 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
     def get_queryset(self):
         partner = self.request.site.partner
         q = self.request.query_params.get('q')
+        subject = self.request.query_params.get('subject')
 
         if q:
             queryset = Course.search(q)
+            queryset = self.get_serializer_class().prefetch_queryset(queryset=queryset, partner=partner)
+        elif subject:
+            queryset = Course.objects.filter(subjects__uuid__exact=subject)
             queryset = self.get_serializer_class().prefetch_queryset(queryset=queryset, partner=partner)
         else:
             if get_query_param(self.request, 'include_hidden_course_runs'):
@@ -138,6 +142,12 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
               mulitple: false
             - name: q
               description: Elasticsearch querystring query. This filter takes precedence over other filters.
+              required: false
+              type: string
+              paramType: query
+              multiple: false
+            - name: subject
+              description: A subject UUID. This filter takes precedence over other filters.
               required: false
               type: string
               paramType: query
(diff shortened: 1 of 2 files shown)


dianakhuang (approved):
Once tests pass.

bbaker6225 (commented):
This looks good to me, I'm just curious why filtering on multiple subjects and then getting the distinct courses didn't work?

inventhouse:
@bbaker6225 
>  I'm just curious why filtering on multiple subjects and then getting the distinct courses didn't work?

Distinct is applied at the SQL level; when a course row is joined with its subject rows, the "course A + subject X" combination is distinct from the "course A + subject Y" combination

bbaker6225:
Ah I was hoping that you could do a `SELECT DISTINCT ON` but looks like that's only in Postgres

Oh well

bbaker6225 (approved):

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -52,10 +52,14 @@ def get_object(self):
     def get_queryset(self):
         partner = self.request.site.partner
         q = self.request.query_params.get('q')
```
bbaker6225, line 54: Wait, sorry one more question. Can't we use this q param already to do what we want with a request like:
```
https://prod-edx-discovery.edx.org/api/v1/courses/?q=subjects__uuid__in="["8aaac548-1930-4614-aeb4-a089dae7ae26"]"
```
?

Or just query off the exact subject uuid if that gives duplicates

inventhouse: Maybe, does that work?  If that works, is that a better way?  The documentation says that `q` should be an Elasticsearch query string so I thought it was just for things exposed through that, like here: https://prod-edx-discovery.edx.org

bbaker6225: That seems to work for me, and if you enter `subjects__uuid="8aaac548-1930-4614-aeb4-a089dae7ae26"` on that documentation page, you should get results even though the documentation doesn't explicitly say you can query off subject. I don't know if it's actually better, but if we can verify it does what we want, I would say we might as well use it since it's already there

inventhouse: Huh.  Perhaps we'll hold this PR for now, run with `q=subjects__uuid__exact` (or `__in`) and see how that goes

bbaker6225: @dianakhuang @mikedikan  

Do either of you have any more information about this? I had no idea we could do this query like this until just now, but it seems like it would work for us

dianakhuang: Unfortunately, I don't know much about this parameter, but it seems both useful and somewhat dangerous. It looks like it's run under `Course.search`, but maybe that also lets you write Django ORM formats, too.

inventhouse: Upon further investigation, we're going to run with the `q=` technique and not add yet more query options if we don't need to.

PR #1723, opened 2019-01-03
Adds in flag for ingesting unpublished course runs to Publisher

jlajoie:


course_discovery/apps/publisher/management/commands/load_drupal_data.py
```diff
@@ -68,6 +72,14 @@ def process_node(self, data):
                             process_course(course, True, course_run=course_run)
                         except AttributeError:
                             pass
+                    elif self.load_unpublished_course_runs:
```
staubina, line 75: Are there any states other than `published` and `unpublished`?

staubina (commented):
The code changes seem straightforward.  

The need for this is that the Course Run is not published right? 

So for example we might have a Course with 1 published and 1 unpublished run.  The Published run was already brought in meaning that the Course was already created. This change will bring in the unpublished run and create it now?

course_discovery/apps/publisher/management/commands/load_drupal_data.py
jlajoie: Nope! Only 0 and 1. From prod:

```SELECT count(1)
FROM 
  node
WHERE 
  status != 0 
AND
  status != 1;
``` returned 0.

jlajoie:
Yes, the goal for this is if there are course runs which exist in Drupal that aren't published and are part of a course that exists in Publisher, we'd still like to not manage that content through Drupal, but not modify the course settings. If it exists in strictly Drupal and it is unpublished, it can be published and we run without the boolean enabled, or we leave it as it's unused content.

PR #1724, opened 2019-01-03
Push course slug to Drupal rather than pulling it in

mikix:
A few notable things:
- We stop pulling in `slug` from Drupal here.
- We make `preview_url` in Publisher.CourseRun an `@property` field, deleting the database column. This won't break green-blue deploys because the attribute still exists.
- We convert the slug field to an AutoSlugField (keeping its very large max_length for historical reasons). But notably, we use AutoSlugField's default behavior of not allowing nulls, whereas the old slug field did allow nulls. Just a little cleanup. But it has some ripple effects in tests and such.
- There are some database rows that are null (no longer allowed) or blank. I let save() keep setting the slug manually in this case rather than trying to solve it in a database migration because migrations don't have the full model methods available (they use a fake model). So I couldn't directly use the correct slugify logic if I did it in-migration.
- The UI in publisher for overriding the slug is the same. It could be improved, but I figured why bother when we're replacing publisher.
- The workflow in publisher though might have changed a bit. Now that we set the slug automatically rather than any manual steps being involved. Hopefully it will still make sense (nothing depends on this manual step...?) - will have to have marketing folks try it on stage.


course_discovery/apps/publisher/tests/test_emails.py
```diff
@@ -536,7 +538,6 @@ def test_course_published_email(self):
         self.assertEqual(str(mail.outbox[0].subject), subject)
         body = mail.outbox[0].body.strip()
         self.assertIn(self.course_run.preview_url, body)
-        self.assertIn(self.course_run.preview_url, body)
```
Dillon-Dumesnil, line 539: Weird...

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1800,6 +1805,8 @@ def assert_dashboard_response(self, studio_count=0, published_count=0, progress_
         self.assertContains(response, 'Course About Pages')
         self.assertContains(response, 'EdX Publisher is used to create course About pages.')
 
+        print('MIKE:', len(response.context['studio_request_courses']), len(response.context['published_course_runs']),
```
Dillon-Dumesnil: 😬

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1848,6 +1855,7 @@ def test_filters(self):
 
         response = self.client.get(self.page_url)
 
+        print("MIKE: ", response.content)
```
Dillon-Dumesnil: 😬 x2

Dillon-Dumesnil (approved):
DiscoveryCourseRun is a little weird to understand, but /shrug. Minor things to remove, but otherwise all good from me.

course_discovery/apps/publisher/tests/test_views.py
mikix: Haha, whoops.

course_discovery/apps/publisher/tests/test_emails.py
mikix: Yeah, I don't know what the intention was (like this should be a different assert statement and they forgot to edit it after pasting?). But figure just getting rid of it makes sense now.

PR #1731, opened 2019-01-07
Increasing prod timeout to value that matches prod likely times

mikedikan:


mikedikan:
Spoke with Ben in person

PR #1732, opened 2019-01-08
Program topic aggregation [DISCO-489]

inventhouse:
Program topics from course topic tags


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1072,6 +1072,22 @@ def test_subject_order(self):
...
+
+        course1 = factories.CourseFactory()
+        course1.topics.set(*topics[:1])  # A
+        course2 = factories.CourseFactory()
+        course2.topics.set(*topics[:2])  # A, B
+        course3 = factories.CourseFactory()
+        course3.topics.set(*topics[1:])  # B, C
+
+        program1 = factories.ProgramFactory(courses=[course1, course2, course3])
+        self.assertEqual({t.name for t in program1.topics}, set(topics))
```
bbaker6225: I'm seeing the topics as a set of Tag objects like `{<Tag: stardust-2019>, <Tag: finance>}` when I add the topics through the tag manager in django admin, so I'm not sure if we want to update this test to reflect that

inventhouse: The test confirms that the program's topic tags' names match the names used to add topic tags to the courses; should I instead create tag objects from the list and compare those?

bbaker6225: I'd probably prefer that, since people are going to be adding tags to courses through the admin which means the course topics tags will be objects

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1072,6 +1072,22 @@ def test_subject_order(self):
         program2 = factories.ProgramFactory(courses=[course1, course2, course3, course4, course5])
         self.assertEqual(program2.subjects, [self.subjects[2], self.subjects[0], self.subjects[1]])
 
+    def test_program_topics(self):
+        """
+        Verify the program's subjects are in order of frequency among courses, with primary subjects coming first
```
bbaker6225: Looks like this is a little out of date

inventhouse: Whoops, copy-paste!

course_discovery/apps/course_metadata/tests/test_models.py
bbaker6225: https://github.com/alex/django-taggit/blob/master/tests/tests.py#L54 looks like something like this will add topics as Tag objects

inventhouse: That link was very helpful, I've pushed changes to incorporate this feedback

bbaker6225 (approved):
Nice

dianakhuang (approved):

PR #1735, opened 2019-01-09
When PC approves a course run, move direct to course team

mikix:
This cuts out one of the publisher-owned (marketing-owned) steps where they craft the preview URL. Now that it's done automatically, we don't need to block on the course publisher.

I left in the bit where when a publisher *does* change the URL, it goes back to CourseTeam and sends email again. That should keep doing that. But I just also made the "send for review" button that the PC presses also do it.


Dillon-Dumesnil (approved):
Sure?

PR #1736, opened 2019-01-09
Show length of a program on the courses endpoint

bbaker6225:
DISCO-502

Example result for programs on a specific course:
screen shot 2019-01-09 at 11 34 56 am


inventhouse (approved):
Looks good

PR #1739, opened 2019-01-10
Hide preview URL until Publish is clicked

mikix:


course_discovery/apps/publisher/templates/publisher/_approval_widget.html
```diff
@@ -42,8 +42,6 @@ <h5 class="hd-5 emphasized approvals-heading">{% trans "REVIEWS" %}</h5>
                             <span class="preview-status">{{course_run.course_run_state.preview_status_for_publisher}}</span>
                                 <button data-url="{% url 'publisher:api:update_course_run' course_run.id %}" class="btn btn-neutral btn-edit-preview-url">{% trans "Edit" %}</button>
                         {% endif %}
-                    {% elif course_run.course.publisher == request.user %}
-                        <button data-url="{% url 'publisher:api:update_course_run' course_run.id %}" class="btn btn-neutral btn-save-preview-url">{% trans "Save" %}</button>
```
mikix, line 46: This change and the one below are removing the "if not course_run.preview_url" code paths - now that we secretly have a preview_url but don't show it, we don't want to let marketing do anything in this situation - they shouldn't be able to set a preview_url they can't see.

They'll just have to wait like normal until a PC left-publishes the course run.

course_discovery/apps/publisher/api/serializers.py
```diff
@@ -99,6 +99,9 @@ def validate_lms_course_id(self, value):
         return value
 
     def validate_preview_url(self, value):
+        if value is None:
+            return value
```
Dillon-Dumesnil, line 103: Nit: return None is more clear to read

Dillon-Dumesnil (approved):
Sure

PR #1741, opened 2019-01-11
Adding in wrapper for Drupal node redirect and business logic to support

Dillon-Dumesnil:
when to add a redirect.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_publishers.py
+++ b/course_discovery/apps/course_metadata/tests/test_publishers.py
@@ -1,5 +1,6 @@
 import json
 
+import ddt
 import mock
 import pytest
 import responses
@@ -8,7 +9,7 @@ from course_discovery.apps.core.tests.factories import PartnerFactory
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.exceptions import (
     AliasCreateError, AliasDeleteError, FormRetrievalError, NodeCreateError, NodeDeleteError, NodeEditError,
-    NodeLookupError
+    NodeLookupError, RedirectCreateError
 )
 from course_discovery.apps.course_metadata.publishers import (
     BaseMarketingSitePublisher, CourseRunMarketingSitePublisher, ProgramMarketingSitePublisher
@@ -19,7 +20,7 @@ from course_discovery.apps.course_metadata.tests.mixins import MarketingSitePubl
 
 
 class DummyObject:
-    dummy = 2
+    dummy = '2'
 
 
 class BaseMarketingSitePublisherTests(MarketingSitePublisherTestMixin):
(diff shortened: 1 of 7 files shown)


course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -349,6 +349,46 @@ def test_alias(self):
...
+        self.mock_node_retrieval(self.publisher.node_lookup_field, lookup_value)
+        lookup_value = getattr(previous_obj, self.publisher.unique_field)
+        self.mock_node_retrieval(self.publisher.node_lookup_field, lookup_value)
+
+        self.mock_get_redirect_form()
+        self.mock_add_redirect(self.obj, previous_obj)
+
+        self.publisher.add_url_redirect(self.obj, previous_obj)
+
+        assert responses.calls[-1].request.url == '{}/add'.format(self.publisher.redirect_api_base)
```
mikix: Good line. I wish `responses` had a method you could call like `responses.assertAllRegisteredCallsActivated()` or something.  But in absence of that, good to check manually. It *does* check that itself when used as a context manager. So an alternative would be to restructure the test that way. But this is fine.

I find it so odd that some of our test classes don't inherit from `TestCase` and thus don't have `self.assertEqual`.

course_discovery/apps/course_metadata/tests/test_publishers.py
```diff
@@ -349,6 +349,46 @@ def test_alias(self):
...
+
+        responses.reset()
+
+        # Same scenario, but this time a non-200 status code is returned during
+        # redirect creation. An exception should be raised.
+        self.mock_api_client()
```
mikix: optional nit: This second block is probably more logically a second test method. Or use ddt.

course_discovery/apps/publisher/models.py
```diff
@@ -995,7 +1001,31 @@ def approved(self):
...
+        # These will be unpublished in favor of the new course.
+        for run in discovery_course.course_runs.all():
+            if run != discovery_run and run.status == CourseRunStatus.Published and run.end and run.end < now:
+                CourseRunMarketingSitePublisher(site.partner).add_url_redirect(discovery_run, run)
+                run.status = CourseRunStatus.Unpublished
+                run.save()
+
+        # Notify course team
+        if waffle.switch_is_active('enable_publisher_email_notifications'):
+            emails.send_course_run_published_email(self.course_run, site)
```
mikix, line 1028: Guh, this whole method sucks, man. #bebetter

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -1106,7 +1110,7 @@ def test_internal_user_status(self):
 
 
 @ddt.ddt
-class CourseRunStateTests(TestCase):
+class CourseRunStateTests(MarketingSitePublisherTestMixin):
```
mikix, line 1113: Humph. I think this chain of mixins is abusing the word. Shouldn't take place of base class like that. Not your doing, and not worth changing. Just harrumphing.

mikix (approved):
Code looks good to me! We tested this together. Let's get this on stage and test further.

course_discovery/apps/publisher/models.py
Dillon-Dumesnil: +1

PR #1742, opened 2019-01-11
Add topics to program endpoint

bbaker6225:
DISCO-490

Example output:
screen shot 2019-01-11 at 3 32 09 pm

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -13,6 +13,7 @@ from haystack.query import SearchQuerySet
 from opaque_keys.edx.keys import CourseKey
 from pytz import UTC
 from rest_framework.test import APIRequestFactory
+from taggit.models import Tag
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -655,9 +656,11 @@ def create_program(self):
         organizations = OrganizationFactory.create_batch(2)
         person = PersonFactory()
 
+        topicA = Tag.objects.create(name="topic A")
```
mikedikan: why topic A?

bbaker6225: I just picked an arbitrary tag name for the test

mikedikan: nit: seems like it's weird to have a topic a without a topic b, i would change to just 'topic' but again, this is a nit and can be safely ignored.

mikedikan (approved):

PR #1746, opened 2019-01-16
added CSRF url route to the discovery endpoint

staubina:


douglashall (commented):
You also need to add the `csrf` app to INSTALLED_APPS. See https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2276 for an example.

douglashall (approved):

PR #1747, opened 2019-01-16
Enable elasticsearch queries on programs endpoint

bbaker6225:
DISCO-532

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -937,8 +937,11 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     degree = DegreeSerializer()
 
     @classmethod
-    def prefetch_queryset(cls, partner):
-        return Program.objects.filter(partner=partner).select_related('type', 'partner').prefetch_related(
+    def prefetch_queryset(cls, partner, queryset=None):
+        # Explicitly check if the queryset is None before selecting related
+        queryset = queryset if queryset is not None else Program.objects.filter(partner=partner)
+
+        return queryset.select_related('type', 'partner').prefetch_related(
             'excluded_course_runs',
             # `type` is serialized by a third-party serializer. Providing this field name allows us to
             # prefetch `applicable_seat_types`, a m2m on `ProgramType`, through `type`, a foreign key to
@@ -1065,7 +1068,7 @@ class ProgramSerializer(MinimalProgramSerializer):
     topics = serializers.SerializerMethodField()
 
     @classmethod
-    def prefetch_queryset(cls, partner):
+    def prefetch_queryset(cls, partner, queryset=None):
         """
         Prefetch the related objects that will be serialized with a `Program`.
 
@@ -1073,7 +1076,9 @@ class ProgramSerializer(MinimalProgramSerializer):
         chain of related fields from programs to course runs (i.e., we want control over
         the querysets that we're prefetching).
         """
-        return Program.objects.filter(partner=partner).select_related('type', 'video', 'partner').prefetch_related(
+        queryset = queryset if queryset is not None else Program.objects.filter(partner=partner)
+
+        return queryset.select_related('type', 'video', 'partner').prefetch_related(
             'excluded_course_runs',
             'expected_learning_items',
             'faq',
--- a/course_discovery/apps/api/v1/views/programs.py
+++ b/course_discovery/apps/api/v1/views/programs.py
@@ -32,7 +32,10 @@ class ProgramViewSet(CacheResponseMixin, viewsets.ReadOnlyModelViewSet):
         # This method prevents prefetches on the program queryset from "stacking,"
         # which happens when the queryset is stored in a class property.
         partner = self.request.site.partner
-        return self.get_serializer_class().prefetch_queryset(partner)
+        q = self.request.query_params.get('q')
+        queryset = q and Program.search(q)
+
+        return self.get_serializer_class().prefetch_queryset(queryset=queryset, partner=partner)
 
     def get_serializer_context(self, *args, **kwargs):
         context = super().get_serializer_context(*args, **kwargs)
@@ -81,6 +84,12 @@ class ProgramViewSet(CacheResponseMixin, viewsets.ReadOnlyModelViewSet):
               multiple: false
             - name: types
               description: Filter by comma-separated list of program type slugs
+              required: false
+              type: string
+              paramType: query
+              multiple: false
+            - name: q
+              description: Elasticsearch querystring query. This filter takes precedence over other filters
               required: false
               type: string
               paramType: query
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -344,7 +344,35 @@ class Position(TimeStampedModel):
         return name
 
 
-class Course(TimeStampedModel):
+class PkSearchableMixin:
+    """
+    Represents objects that have a search method to query elasticsearch and load by primary key.
+    """
+
+    @classmethod
+    def search(cls, query):
+        """ Queries the search index.
+
+        Args:
+            query (str) -- Elasticsearch querystring (e.g. `title:intro*`)
+
+        Returns:
+            QuerySet
+        """
+        query = clean_query(query)
+
+        if query == '(*)':
+            # Early-exit optimization. Wildcard searching is very expensive in elasticsearch. And since we just
+            # want everything, we don't need to actually query elasticsearch at all.
+            return cls.objects.all()
+
+        results = SearchQuerySet().models(cls).raw_search(query)
+        ids = {result.pk for result in results}
+
+        return cls.objects.filter(pk__in=ids)
+
+
+class Course(PkSearchableMixin, TimeStampedModel):
     """ Course model. """
     partner = models.ForeignKey(Partner)
     uuid = models.UUIDField(default=uuid4, editable=False, verbose_name=_('UUID'))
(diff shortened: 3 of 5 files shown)


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -660,6 +660,27 @@ def create_program_with_seats(self):
...
+    def test_search(self):
+        """
+        Verify that the program endpoint correctly handles basic elasticsearch queries
+        """
+        title = 'Some random title'
+        expected = set(factories.ProgramFactory.create_batch(1, title=title))
+        query = 'title:' + title
+        assert set(Program.search(query)) == expected
+
+    def test_subject_search(self):
```
bbaker6225, line 674: This test isn't really necessary, but figured it would be nice to test the specific functionality we're looking for

course_discovery/apps/api/serializers.py
```diff
@@ -937,8 +937,11 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     degree = DegreeSerializer()
 
     @classmethod
-    def prefetch_queryset(cls, partner):
-        return Program.objects.filter(partner=partner).select_related('type', 'partner').prefetch_related(
+    def prefetch_queryset(cls, partner, queryset=None):
+        # Explicitly check if the queryset is None before selecting related
+        queryset = queryset if queryset is not None else Program.objects.filter(partner=partner)
```
mikix, line 942: If empty querysets weren't falsy, we could do `queryset = queryset or Programs.objects...` Ah well.

It might be clearer to split this into two lines like so? Matter of taste.
```
if queryset is None:
    queryset = Program.objects.filter(partner=partner)
```

course_discovery/apps/api/serializers.py
```diff
@@ -1065,15 +1068,17 @@ class ProgramSerializer(MinimalProgramSerializer):
...
+    def prefetch_queryset(cls, partner, queryset=None):
         """
         Prefetch the related objects that will be serialized with a `Program`.
 
         We use Prefetch objects so that we can prefetch and select all the way down the
         chain of related fields from programs to course runs (i.e., we want control over
         the querysets that we're prefetching).
         """
-        return Program.objects.filter(partner=partner).select_related('type', 'video', 'partner').prefetch_related(
+        queryset = queryset if queryset is not None else Program.objects.filter(partner=partner)
```
mikix, line 1079: Same deal here.

course_discovery/apps/api/serializers.py
```diff
@@ -937,8 +937,11 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     degree = DegreeSerializer()
 
     @classmethod
-    def prefetch_queryset(cls, partner):
-        return Program.objects.filter(partner=partner).select_related('type', 'partner').prefetch_related(
+    def prefetch_queryset(cls, partner, queryset=None):
```
mikix, line 940: It's a little odd that we aren't using partner at all if queryset is provided? Maybe partner=None so it doesn't have to be specified in that case?

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -32,7 +32,13 @@ def get_queryset(self):
         # This method prevents prefetches on the program queryset from "stacking,"
         # which happens when the queryset is stored in a class property.
         partner = self.request.site.partner
-        return self.get_serializer_class().prefetch_queryset(partner)
+        q = self.request.query_params.get('q')
+        queryset = None
+
+        if q:
+            queryset = Program.search(q)
```
mikix: You could replace the last bits with: `queryset = q and Program.search(q)`

course_discovery/apps/course_metadata/models.py
```diff
@@ -344,7 +344,35 @@ def organization_name(self):
         return name
 
 
-class Course(TimeStampedModel):
+class PkSearchableObject:
```
mikix: nit: I'd call this PkSearchableMixin.

course_discovery/apps/course_metadata/models.py
```diff
@@ -344,7 +344,35 @@ def organization_name(self):
...
+            # want everything, we don't need to actually query elasticsearch at all.
+            return cls.objects.all()
+
+        results = SearchQuerySet().models(cls).raw_search(query)
+        ids = {result.pk for result in results}
+
+        return cls.objects.filter(pk__in=ids)
+
+
+class Course(TimeStampedModel, PkSearchableObject):
```
mikix: Mixins go first: https://www.ianlewis.org/en/mixins-and-python

course_discovery/apps/course_metadata/models.py
```diff
@@ -477,35 +505,6 @@ def first_enrollable_paid_seat_price(self):
...
-        if query == '(*)':
-            # Early-exit optimization. Wildcard searching is very expensive in elasticsearch. And since we just
-            # want everything, we don't need to actually query elasticsearch at all.
-            # No point logging if log_course_search_queries is enabled for this wildcard, so skip that too.
-            return cls.objects.all()
-
-        results = SearchQuerySet().models(cls).raw_search(query)
-        ids = {result.pk for result in results}
-
-        if waffle.switch_is_active('log_course_search_queries'):
```
mikix, line 501: You don't have a replacement for this block. Do we not care about this log message anymore? (was it just a debug thing?)

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -660,6 +660,27 @@ def create_program_with_seats(self):
...
         return factories.ProgramFactory(type=program_type, courses=[course_run.course])
 
+    def test_search(self):
+        """
+        Verify that the program endpoint correctly handles basic elasticsearch queries
+        """
+        title = 'Some random title'
+        expected = set(factories.ProgramFactory.create_batch(1, title=title))
+        query = 'title:' + title
+        assert set(Program.search(query)) == expected
```
mikix: You might want to add a second (not matching) program to confirm that your code isn't simply returning everything and is actually using the query provided.

Also, self.assertSetEqual probably gives a nicer message when they aren't, but who cares.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -660,6 +660,27 @@ def create_program_with_seats(self):
 
         return factories.ProgramFactory(type=program_type, courses=[course_run.course])
 
+    def test_search(self):
+        """
+        Verify that the program endpoint correctly handles basic elasticsearch queries
+        """
+        title = 'Some random title'
+        expected = set(factories.ProgramFactory.create_batch(1, title=title))
```
mikix, line 668: This could just be `set(factories.ProgramFactory(title=title))` eh?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -660,6 +660,27 @@ def create_program_with_seats(self):
...
+        """
+        Verify that the program endpoint correctly handles elasticsearch queries on the subject uuid
+        """
+        subject = factories.SubjectFactory()
+        course = factories.CourseFactory(subjects=[subject])
+        expected = set(factories.ProgramFactory.create_batch(1, courses=[course]))
+        # Create an extra program that should not show up
+        factories.ProgramFactory()
+        query = str(subject.uuid)
+        assert set(Program.search(query)) == expected
```
mikix: Similar comments here.

Also - it's odd that a search for a subject uuid will hit programs. I guess by definition, there's unlikely to be overlap between the two uuid sets. But struck me as odd. Though it sounds like this is the whole point of this PR?

mikix (approved):
I don't see a test that hits the endpoint with a q= arg - might be nice to add. But seems good! Wrote a few comments, nothing blocking.

course_discovery/apps/course_metadata/models.py
bbaker6225: I got the ok to remove it from @douglashall who added it 2 years ago, since it doesn't seem to be used anymore. https://openedx.atlassian.net/browse/DISCO-538 is the ticket for making a migration to remove the waffle switch once this code is gone

course_discovery/apps/api/serializers.py
bbaker6225: For this I was trying to keep the same style that a few other serializers have. For example:
https://github.com/edx/course-discovery/blob/ddfe529e209d3a1d1c1622cb2ffdbf74b66e5166/course_discovery/apps/api/serializers.py#L562

course_discovery/apps/api/serializers.py
bbaker6225: I think it does use partner further down in the `select_related` chain
https://github.com/edx/course-discovery/blob/ddfe529e209d3a1d1c1622cb2ffdbf74b66e5166/course_discovery/apps/api/serializers.py#L1093

mikix: Ah yes of course. I was either looking at the wrong method or just missed it. nm me! :)

course_discovery/apps/course_metadata/tests/test_models.py
bbaker6225: The final result is in the form of a set of lists, so I could switch it to `set([factories.ProgramFactory(title=title)])`, but I didn't feel like that was much nicer

course_discovery/apps/course_metadata/tests/test_models.py
bbaker6225: Yeah, it does seem a little strange, but this PR is just implementing the same functionality for programs that you get for courses where you can filter them by subject uuids by hitting a url like:
https://prod-edx-discovery.edx.org/api/v1/courses/?q=a168a80a-4b6c-4d92-9f1d-4c235206feaf

mikedikan (approved):

course_discovery/apps/api/v1/views/programs.py
bbaker6225: Do you mind clarifying what you mean here? I'm not following the replacement right now

mikix: So python does this nice thing when you've got a series of `and` or `or` - the series will return the first object that stops the resolution (or the last object if none do).

So `q and Program.search(q)` will return `q` if it's falsy or `Program.search(q)` if not. Which is exactly the state of the three lines above starting with `queryset = None` through the `if`

bbaker6225: Ah right right. Nice

mikix: (All that said, if you think it makes it less readable, ignore me.)

bbaker6225: Nah, I like it, I just didn't think of it

PR #1750, opened 2019-01-17
Add ADR proposal about changes to program-curricula relationship.

iloveagent57:


docs/decisions/0003-program-curricula-changes.rst
```diff
@@ -0,0 +1,200 @@
...
+distinct specializations, and each specialization may require the completion
+of different courses or MicroMasters (i.e. other, smaller programs).
+
+Decision
+========
+
+To accurately model data as described above, the following changes to models
+in the course_metadata application are proposed:
+
+A Program may have multiple Curricula
```
schenedx: This is cool. But this also means a change to the Master Program page curriculum section. If we have a Master program with multiple Curricula, then the Marketing Site about page would render the first one returned or be broken. To accommodate this change, we need to update that section of the marketing site about page.

docs/decisions/0003-program-curricula-changes.rst
```diff
@@ -0,0 +1,200 @@
...
+  objects instead of the ``courses`` field.
+- Remove the ``Program.courses`` field.
+- Optionally, add a ``Program.courses`` property that transforms course
+  curriculum data into a list of course objects as currently expected of the
+  Program ``courses`` interface.
+
+Action B.2
+^^^^^^^^^^
+
+We can keep the ``Program.excluded_course_runs`` field (I think), but the
```
schenedx: Why not also version the Program.excluded_course_runs field? If we don't version, that would be a bit inconsistent for linkage between programs and courses, right? I argue that membership of courses change less frequently than the exclusion of course runs within a program.

docs/decisions/0003-program-curricula-changes.rst
```diff
@@ -0,0 +1,200 @@
...
+Action 1
+^^^^^^^^
+
+- Add ``program`` as a FK field in the ``Curriculum`` model, and point
+  that FK at the ``Program`` model.
+- Add the following fields to the ``Curriculum`` model to help
+  support versioning and specialization:
+
+  - ``name: str`` The name of this curriculum.
+  - ``current: boolean`` Whether this curriculum is currently offered/available.
```
schenedx: I think `is_active` is a better name to the field.

docs/decisions/0003-program-curricula-changes.rst
```diff
@@ -0,0 +1,200 @@
...
+
+To support the addition/removal and versioning thereof of programs and courses
+within curricula, we should add the following fields to both the
+``CurriculumProgramMembership`` and ``CurriculumCourseMembership`` models:
+
+- ``effective_at: datetime`` The time at which this program or course is
+  available within the curriculum.
+- ``retired_at: datetime`` The time at which this program or course is
+  becomes unavailable within the curriculum.
+- ``current: boolean`` Whether this program or course is currently available
```
schenedx: I think `is_active` is a better name to the field.

docs/decisions/0003-program-curricula-changes.rst
```diff
@@ -0,0 +1,200 @@
...
+^^^^^^^^
+
+Support the versioning of degree costs and deadlines.  Add the following fields
+to each of the ``DegreeDeadline`` and ``DegreeCost`` models.
+
+- ``effective_at: datetime`` The time at which the deadline/cost was effective
+  w.r.t. the degree.
+- ``retired_at: datetime`` The time at which this deadline/cost was no longer
+  effective w.r.t. the degree.
+- ``current: boolean`` Whether this deadline/cost is currently effective
```
schenedx: I think `is_active` is a better name to the field

schenedx (commented):
How much would Django admin interface to create Degree change? Are there work related to ensure Degree creation is still intuitive that we should capture here?

The biggest worry about this design is the complexity and engineering workload. Updating `ProgramCourse` relationship involves updating Serializers, Views, Unit tests, management commands, refresh_course_metadata jobs, Publisher APIs and maybe LMS cache structure (For example: https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/programs/utils.py#L129). It's a good thing we don't need to worry about ElasticSearch and the Search API. 
Do you think we can make this update **without making changes to the contract of Program API**? If we can do that, it would be a much smaller change with less disruption.

docs/decisions/0003-program-curricula-changes.rst
schenedx: Also, a bigger issue with multiple active versions of Curriculum associated with the program is the assumption that each program would only have ONE active set of courses associated. This assumption is coded in a way scattered throughout the course discovery code base. For example: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/search_indexes.py#L322. We maybe able to keep this assumption by introducing the "Default Curriculum" concept.

docs/decisions/0003-program-curricula-changes.rst
```diff
@@ -0,0 +1,200 @@
...
+Action B.1
+^^^^^^^^^^
+
+- Migrate existing data from the ``Program.courses`` many-to-many relationship
+  into a new ``Curriculum`` model related to the program (for each program
+  that we have).
+- Update whatever interface currently creates new programs to use ``Curriculum``
+  objects instead of the ``courses`` field.
+- Remove the ``Program.courses`` field.
+- Optionally, add a ``Program.courses`` property that transforms course
```
schenedx: This is a great way to make sure our change won't break too many things. To implement this we have to introduce the "Default Curriculum" concept? Or otherwise how would you implement this property?

iloveagent57:
Thanks for your feedback, Simon.  As we discussed in person on Friday, yes, we can definitely scope this to just the Degree model for now, and leave the changes that pertain to the Program model for a second phase of future work.

I'd like to make as few changes as possible to the Django admin interface for the Degree models.  I think the biggest challenge will be to clearly articulate when a user should create a new version of a Curriculum and when they should only update the current structure of the curriculum.

Dillon-Dumesnil (approved):
Not going to proofread as that is better left to your team, but will approve so you can merge when ready

docs/decisions/0011-program-curricula-changes.rst
```diff
@@ -0,0 +1,126 @@
+Refactor the Curriculum model to relate to the Program model
+============================================================
+
+Status
+======
+
+Proposed (January 2019)
```
crice100: I'm a little fuzzy on how _status_ is used in these ADRs, but should this be "proposed" if the changes have already been made?

crice100 (approved):

PR #1752, opened 2019-01-18
Add publisher permissions decision doc

mikix:


staubina (approved):
This is clear to me.

docs/decisions/0003-publisher-permissions.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Legal staff users will have an extra bit of control to toggle some settings
+like whether a course is OFAC restricted, for example. There may be some
+checkpoints waiting on them before a course team can publish. These users will
+be marked by belonging to a particular group. That group should be configurable.
+
+Membership in an org group will be managed by staff users. But once you're in
+that group, you can create courses and course runs for that org. Once you do,
+you are the first and only editor of that course. Any current editor can add
+any other editor from that same org. Or can remove any editor.
```
Dillon-Dumesnil, line 43: Nit: "... can add or remove any other editor..."

docs/decisions/0003-publisher-permissions.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Consequences
+------------
+
+Removing a user from an org group should remove editor privileges to all
+courses in that org. Likewise deleting a user.
+
+If a course has no editors left, rather than orphaning that course, anyone
+in that org should be able to edit it. Once an editor is added back to it,
```
Dillon-Dumesnil, line 52: Implementation detail, but how is an editor going to be added back to a course? Will that be an edX staff move or just the first one to edit it is now the sole editor.

Dillon-Dumesnil (approved):

docs/decisions/0003-publisher-permissions.rst
mikix: I didn't want to say that because of the "from that same org" clause. You can remove anyone, even if not in same org. (Shouldn't get in that situation, I don't think? But just being abundantly cautious.)

docs/decisions/0003-publisher-permissions.rst
mikix: Since anyone can edit the course, any course team member should be able to add a new editor.

docs/decisions/0003-publisher-permissions.rst
Dillon-Dumesnil: How would anyone ever get added from a different org? Like if we had a course being co-taught by MIT and Harvard and someone at edX adds an MIT editor after the Harvard person makes the course?

docs/decisions/0003-publisher-permissions.rst
Dillon-Dumesnil: Ah okay, and then once a single editor is specified, everyone else's access is removed. Sounds good. We should add functionality for batch add editors

docs/decisions/0003-publisher-permissions.rst
mikix: For joint courses, we make a new org representing the combo. So really, this "not in same org" problem would likely come up from a person being removed from an org but still being attached to a course. I try to cover that case elsewhere in this doc, so it's not a practical concern. Just being defensive here.

mikix: Really, I probably don't need to get in weeds about who can be added either in this ADR. Should probably just say "editors manage other editors" or something. But eh, good enough.

PR #1753, opened 2019-01-18
Support CORS access

mikix:
Using django-cors-headers. Only current devstack allowance is for the new publisher-frontend on localhost port 18400.


Dillon-Dumesnil (approved):
Should we be whitelisting publisher stage somewhere as well?

mikix:
Yes, you are right to ask that. But that's apparently all done in the nginx config (not through django) via the configuration repo:

https://github.com/edx/configuration/pull/4959 
https://github.com/edx/edx-internal/pull/698

Dillon-Dumesnil:
👍

PR #1761, opened 2019-01-28
Aj/disco 479 publisher orgs

staubina:
Updating the organizations endpoint to respect a User's permissions.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_organizations.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_organizations.py
@@ -1,10 +1,13 @@
 import uuid
 
 from django.urls import reverse
+from guardian.shortcuts import assign_perm
 
 from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase, SerializationMixin
 from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
 from course_discovery.apps.course_metadata.tests.factories import Organization, OrganizationFactory
+from course_discovery.apps.publisher.models import OrganizationExtension
+from course_discovery.apps.publisher.tests import factories as publisher_factories
 
 
 class OrganizationViewSetTests(SerializationMixin, APITestCase):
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -65,6 +69,35 @@ def test_list(self):
...
+        OrganizationFactory.create(partner=self.partner)
+
+        extension1 = publisher_factories.OrganizationExtensionFactory(organization=org1)
+        publisher_factories.OrganizationExtensionFactory(organization=org2)
+        assign_perm(OrganizationExtension.VIEW_COURSE, extension1.group, extension1)
+
+        self.non_staff_user.groups.add(extension1.group)
+
+        # Check Staff user get all groups
+        with self.assertNumQueries(7):
```
jlajoie: nit: I hate numQuery checks, what is it we actually care about from these queries?

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
```diff
@@ -65,6 +69,35 @@ def test_list(self):
...
+            response = self.client.get(self.list_path)
+
+        self.assertEqual(response.status_code, 200)
+        self.assert_response_data_valid(response, Organization.objects.all())
+
+        # Check non staff user gets 1 group
+        self.client.logout()
+        self.client.login(username=self.non_staff_user.username, password=USER_PASSWORD)
+
+        with self.assertNumQueries(11):
```
jlajoie: nit: same as above

jlajoie (approved):
Just nits about the numQueries tests, would rather we test for results rather than the means to get them (more/less queries don't feel like they should break tests imo)

course_discovery/apps/api/v1/tests/test_views/test_organizations.py
staubina: I almost removed them, they are remnants of the previous tests.
 
They did rise quickly when I stopped using the prefetch.

PR #1762, opened 2019-01-28
DISCO-471: Adds in Dynamic Fields Serializer to explicitly query fields

jlajoie:


staubina (approved):
This seems almost too simple.  Nice find.

PR #1763, opened 2019-01-28
EDUCATOR-3939: Add Program FK to Curriculum model

kdmccormick:
**Team:** @edx/educator-neem 
**Ticket:** https://openedx.atlassian.net/browse/EDUCATOR-3939


course_discovery/apps/api/serializers.py
```diff
@@ -935,6 +936,7 @@ class MinimalProgramSerializer(serializers.ModelSerializer):
     courses = serializers.SerializerMethodField()
     type = serializers.SlugRelatedField(slug_field='name', queryset=ProgramType.objects.all())
     degree = DegreeSerializer()
+    curricula = CurriculumSerializer(many=True)
```
kdmccormick, line 939: @iloveagent57 Is it appropriate to add the `curricula` field here as well?

iloveagent57: I think it's going to break a lot of tests and possibly violate some assumptions that clients of the Programs APIs have, too.  So I would add it to the Degree serializer for now and we can tackle adding it to the Program serializer when we get to phase 2 of this work (which is not for a while).

iloveagent57: Sorry, I told you the completely wrong thing.  You _have_ to add `curriculum` to the program serializer, not the degree serializer.  That's what's going to make it so that the JSON you get back from a call to the programs API has curriculum as a direct child of the program (and not a child of the degree, which is a child of the program).

iloveagent57 (approved):
Nicely done!

PR #1766, opened 2019-02-01
Update the degreeDeadlineModel Time field to be not required

schenedx:
https://openedx.atlassian.net/browse/MC-4217

Set the time field to be blankable (not required)

@edx/educator-neem FYI


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1450,14 +1450,17 @@ def setUp(self):
...
-        deadline_name = "A test deadline"
-        deadline_date = "January 1, 2019"
+        deadline_name = 'A test deadline'
+        deadline_date = 'January 1, 2019'
+        deadline_time = ''
         degree_deadline = DegreeDeadline.objects.create(
             degree=self.degree,
             name=deadline_name,
             date=deadline_date,
+            time=deadline_time,
```
iloveagent57: Don't you want to test that the deadline object's `time` field is the empty string even if you don't specify a value during creation?

schenedx: Good point. I will do a new test case.

schenedx: Test added. Please review again @iloveagent57

iloveagent57 (approved):

PR #1773, opened 2019-02-06
Rename curriculum models to a better name

reillz10:


iloveagent57:
That test failure is because you're missing test factories for these new models.  Add factories for each new model in `course_discovery/apps/course_metadata/tests/factories.py`, you can copypasta the existing models.

PR #1774, opened 2019-02-06
Switch the references from the old names for curriculum models to the…

reillz10:
… new names


course_discovery/apps/course_metadata/models.py
```diff
@@ -1674,10 +1674,10 @@ class Curriculum(TimeStampedModel):
         help_text=_('A high-level overview of the degree\'s courseware.'),
     )
     program_curriculum = models.ManyToManyField(
-        Program, through='course_metadata.DegreeProgramCurriculum', related_name='degree_program_curricula'
```
schenedx, line 1677: Are you sure we don't need new migration files for changes here?

reillz10: We do need one! Sorry about that, that file didn't get added.

schenedx (approved):
👍

PR #1775, opened 2019-02-06
Remove the old named curriculum models

reillz10:


schenedx (changes requested):
Please remove the migration file out of this PR and deployment. You need to do another PR and deployment with just the migration file.

schenedx (approved):
👍

PR #1777, opened 2019-02-07
WIP - adding tags from refresh metadata

mikedikan:
DISCO-419

Updating refresh_metadata to add course tags when ingesting data from the marketing site.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -363,7 +363,7 @@ class CourseMarketingSiteDataLoader(AbstractMarketingSiteDataLoader):
             )
 
         # Once we've ingested the course and course run, we update the topic tags with any marketing site tags found.
-        if course_run:
+        if waffle.switch_is_active('migrate_drupal_tags') and course_run:
             self.set_topic_tags(course_run.course, data)
 
     def get_course_run(self, data):
--- a/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
+++ b/course_discovery/apps/course_metadata/data_loaders/marketing_site.py
@@ -7,6 +7,7 @@ from urllib.parse import parse_qs, urlencode, urlparse
 from uuid import UUID
 
 import pytz
+import waffle
 from dateutil import rrule
 from django.utils.functional import cached_property
 from opaque_keys import InvalidKeyError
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -333,6 +343,10 @@ def process_node(self, data):
                 data['field_course_uuid']
             )
 
+        # Once we've ingested the course and course run, we update the topic tags with any marketing site tags found.
```
mikedikan, line 365: @staubina - the topic tag updating happens separately from course/course run ingestion

mikedikan:
This is now ready for review for merge

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -518,6 +551,23 @@ def get_extra_description(self, raw_objects_data):
...
 
+    def set_topic_tags(self, course, raw_objects_data):
+        updated = False
+
+        course_tag_map = self.STAGE_COURSE_TAG_ID_MAP
+        if waffle.switch_is_active('production_data_for_tag_migration'):
+            course_tag_map = self.PROD_COURSE_TAG_ID_MAP
+
+        for term_object in raw_objects_data.get('field_course_tags'):
+            if term_object.get('uuid') in course_tag_map.keys():
```
bbaker6225: I think you're using `term_object.get('uuid')` enough here to warrant a variable for it. Also, past the first if statement, you should know that `course_tag_map` has the `term_object` uuid in it, so that could be another variable as well

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -1315,7 +1315,20 @@
         'field_course_has_prerequisites': True,
         'field_course_enrollment_credit': None,
         'field_course_is_disabled': None,
-        'field_course_tags': [],
+        'field_course_tags': [
```
bbaker6225: Does this file have a good place where you could declare the mock `field_course_tags` once and then reuse it?

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -289,8 +317,9 @@ def get_query_kwargs(self):
 
     def process_node(self, data):
 
+        course_run = self.get_course_run(data)
```
bbaker6225, line 320: Should this remain after the `if` statement?

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -477,6 +477,11 @@ def assert_course_loaded(self, data):
 
         self.validate_relationships(data, data_map)
 
+    def assert_topics_loaded(self, data):
+        # Verify the topics were loaded from course tags
+        course = self._get_course(data)
+        self.assertEqual(list(course.topics.slugs()), [u'professional-programs', u'communications'])
```
bbaker6225: It would be great if we could also check the names (`course.topics.names()`), since I believe the name field is what we actually get from the API

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
mikedikan: 👍

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
mikedikan: I'll try to find a way

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
mikedikan: i'll check

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
mikedikan: 👍

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
mikedikan: No, the logic for this has slightly changed - it will return none if the course is not found (which will happen for newly imported courses) but this should still be sound.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
mikedikan: Tried something here

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
mikedikan: updated.

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
mikedikan: updated.

mikedikan:
@bbaker6225 This is ready for you to take another look

mikedikan:
@bbaker6225 - comfortable with you merging on tuesday if you are up for it as well!

bbaker6225 (approved):
Looks good to me, but I'd like it if @dianakhuang could also take a quick look since I'm not very familiar with RCM

dianakhuang (approved):
This seems good to me. I'm not super familiar with the inner workings of RCM either.

bbaker6225:
Just realizing, shouldn't we also tag them with stardust? Additionally, I think we will need to tag all the courses in a program that gets updated, so we don't break the prospectus build

dianakhuang:
I believe these changes should be reviewable now @bbaker6225

bbaker6225 (approved):
Seems good to me, assuming we have decided that we can get rid of stardust

PR #1780, opened 2019-02-08
EDUCATOR-3944 Add is_active field to curriculum membership models

kdmccormick:
**Ticket:** https://openedx.atlassian.net/browse/EDUCATOR-3944
**Team:** @edx/educator-neem


course_discovery/apps/course_metadata/models.py
```diff
@@ -1700,6 +1701,7 @@ class CurriculumCourseMembership(TimeStampedModel):
     """
     curriculum = models.ForeignKey(Curriculum, on_delete=models.CASCADE)
     course = models.ForeignKey(Course, on_delete=models.CASCADE)
+    is_active = models.BooleanField(default=False)
```
schenedx: I think we should make `default=True` for both models. The rational is, during the first setup, the intention is definitely to make the membership count. But then we should leave the option for anyway to toggle it off later.

schenedx (changes requested):
Please have the default to be True

course_discovery/apps/course_metadata/models.py
kdmccormick: Sounds good :+1:

schenedx (approved):
👍

PR #1782, opened 2019-02-13
Fixed create course run 400

asadazam93:
### Description
Fixed course run create error. When creating a new run all previous runs with the same course number should be checked for duplication. Previously it only checked related runs for duplicate key which was causing an issue since the course number can be updated but the lms_course_id of previous runs does not change even if the course number is updated.

**Sandbox**
N/A

### Reviewers
- [ ] @awaisdar001 
- [ ] @fysheets 

### Post-review

### DIFF
--- a/course_discovery/apps/publisher/studio_api_utils.py
+++ b/course_discovery/apps/publisher/studio_api_utils.py
@@ -4,6 +4,7 @@ import math
 from opaque_keys.edx.keys import CourseKey
 
 from course_discovery.apps.core.utils import serialize_datetime
+from course_discovery.apps.course_metadata.models import CourseRun as DiscoveryCourseRun
 
 logger = logging.getLogger(__name__)
 
@@ -30,13 +31,10 @@ class StudioAPI:
         trimester = math.ceil(start.month / 4.)
         run = '{trimester}T{year}'.format(trimester=trimester, year=start.year)
 
-        try:
-            discovery_course = course_run.course.discovery_counterpart
-        except:  # pylint:disable=bare-except
-            logger.exception('Failed to get Discovery counterpart for Publisher course [%d]', course_run.course.id)
-            return run
+        related_course_runs = DiscoveryCourseRun.objects.filter(
+            key__contains=course_run.course.number
+        ).values_list('key', flat=True)
 
-        related_course_runs = discovery_course.course_runs.values_list('key', flat=True)
         related_course_runs = [CourseKey.from_string(key).run for key in related_course_runs]
         return cls._get_next_run(run, '', related_course_runs)
 


course_discovery/apps/publisher/studio_api_utils.py
```diff
@@ -4,6 +4,7 @@
 from opaque_keys.edx.keys import CourseKey
 
 from course_discovery.apps.core.utils import serialize_datetime
+from course_discovery.apps.course_metadata.models import CourseRun
```
awaisdar001: `as DiscoveryCourseRun`

course_discovery/apps/publisher/studio_api_utils.py
```diff
@@ -30,13 +31,10 @@ def calculate_course_run_key_run_value(cls, course_run):
         trimester = math.ceil(start.month / 4.)
         run = '{trimester}T{year}'.format(trimester=trimester, year=start.year)
 
-        try:
-            discovery_course = course_run.course.discovery_counterpart
-        except:  # pylint:disable=bare-except
-            logger.exception('Failed to get Discovery counterpart for Publisher course [%d]', course_run.course.id)
-            return run
+        related_course_runs = DiscoveryCourseRun.objects.filter(
```
awaisdar001, line 34: @asadazam93 Have you tested this following scenrio. 

1. Create Course Run
2. Publish Course Run
2. Change the Course Number
3. Publish the Course Run Again <== Does this creates a new course nun in the studio?

awaisdar001 (commented):
I am basically want to confirm if this change is not affecting other existing course runs in the system. 
Specifically, _update_course_run_details_in_studio_ method, which is using the method you are modifying here. 

Can you take a look and see we are covering all use cases here? Also, Please point me out the existing tests. 
Thank you.

asadazam93:
@awaisdar001 this change does not effect any existing course runs. I have just changed the way **related_discovery_runs** are fetched to check for duplication. It now fetches all runs that have the same key (which makes more sense since the key should be unique for all runs not just unique within the related course runs of a course) unlike the previous implementation which only fetched course runs that are related to the course. [Here](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/tests/test_studio_api_utils.py#L32-L32) are the related tests.

course_discovery/apps/publisher/studio_api_utils.py
asadazam93: @awaisdar001 this change does not effect this scenario. Moreover a new run is created in studio only when a new run is created in publisher. Publishing the course only updates the studio instance it does not create a new instance.

awaisdar001:
Sweet, Thank you.

The tests you mentioned does not follow the scenario of changing the Course Number and simulating the Pubilsh action.

awaisdar001 (approved):

PR #1784, opened 2019-02-13
add history to several models

jansenk:
add history to

- Curriculum
- CurriculumProgramMembership
- CurriculumCourseMembership
- DegreeDeadline
- DegreeCost


course_discovery/apps/course_metadata/migrations/0159_historicalcurriculum_historicalcurriculumcoursemembership_historicalcurriculumprogrammembership_hist.py
```diff
@@ -0,0 +1,136 @@
+# -*- coding: utf-8 -*-
```
iloveagent57: Could you make this filename a wee bit shorter? :)

course_discovery/apps/publisher/migrations/0075_auto_20190213_2015.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
```
iloveagent57, line 1: I'm guessing that this was generated when upgrading the `simple-history` version, correct?  Any idea what exactly the purpose of this migration is?

iloveagent57 (commented):
The changes in `course_metadata` generally look good, I just have a question about the publisher migration.

course_discovery/apps/course_metadata/migrations/0159_historicalcurriculum_historicalcurriculumcoursemembership_historicalcurriculumprogrammembership_hist.py
jansenk: Can I just rename it?

course_discovery/apps/publisher/migrations/0075_auto_20190213_2015.py
jansenk: You're correct, it seems to be due to the upgrade. I compared it tothe current definition of these columns and it looks like it is adding default=None to the historical fields to match the fields on the base models

jansenk:
@iloveagent57 any other comments?

schenedx (approved):
👍

PR #1789, opened 2019-02-19
Add CourseEditor model

mikix:
This will tie a course and a user together to determine if they can edit a course.

Also modify the course API view to check for editors or staff state before letting a user modify a course (or organization membership before letting a user create a course).

I left the course run API alone for now - it doesn't yet support editing.

https://openedx.atlassian.net/browse/DISCO-555


course_discovery/apps/api/permissions.py
```diff
@@ -11,11 +11,24 @@ def has_permission(self, request, view):
...
+            # You must be a member of the organization within which you are creating a course
+            org = request.data.get('org')
+            return org and request.user.groups.filter(organization_extension__organization__key=org).exists()
+        else:
+            return True  # other write access attempts will be caught by object permissions below
+
+    def has_object_permission(self, request, view, obj):
+        if request.user.is_staff:  # staff users are always editors
+            return True
+        elif request.method in SAFE_METHODS:
```
jlajoie: Is SAFE_METHODS just like GET requests?

course_discovery/apps/api/permissions.py
```diff
@@ -11,11 +11,24 @@ def has_permission(self, request, view):
...
+            return True
+        else:
+            return request.user.courses_edited.filter(course=obj).exists()
```
jlajoie: So this is "if they've already edited it, they can still edit this". Does this impact a scenario where we remove some one from an Organization (fired/quit/etc.) they could alway still come in and do malicious things? That's kind of a stretch right, so this is probably reasonable?

course_discovery/apps/api/permissions.py
mikix: Yeah, exactly: ('GET', 'HEAD', 'OPTIONS')

jlajoie (approved):

course_discovery/apps/api/permissions.py
mikix: `courses_edited` is maybe a bad name I gave it? I meant "courses for which this user is an editor" -- is there a less confusing way of saying that?

You are right that I didn't add any logic around handling when someone is removed from an org, to then remove all their editor access. I think that would be something good to do... Do you know a non-painful way of adding that logic?

mikix:
Oh, I also forgot one aspect of that ADR I wrote - if a course has no editors for whatever reason, everyone in its authoring orgs is an editor... will fix

mikix:
OK, updated permission logic and added new tests -- now we allow access if no other valid editors and we check group membership every access, rather than assuming if you're an editor, you're always allowed in.

PR #1795, opened 2019-02-21
Allow filtering courses by editable

mikix:
If you pass editable=1 to the courses list endpoint, you'll now only see courses that you will be able to edit.

I refactored the logic into classmethods on CourseEditor. I felt it made sense to have it all in a row.

The ticket mentions also allowing all internal users to "see the course" but I believe that item was mostly about not messing up the current list view for internal users. Bots (internal users) don't need to list editable courses and see everything, to my knowledge, eh? 

https://openedx.atlassian.net/browse/DISCO-539


Dillon-Dumesnil (approved):
Code looks fine. I'm not sure about the answer to your question re: bots.

staubina (approved):
It looks like the internal user's things was a remnant from when the story was originally written.

The code looks good to me.

PR #1796, opened 2019-02-22
Add username replacement API

tuchfarber:
New API replaces all copies of username in this service with a new
username. Requires user be in a username_replacement_admin group. Should
only be ran as a larger job to update usernames across all services,
otherwise the system will be left in a broken state for those users.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_user_management.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_user_management.py
@@ -1,0 +1,94 @@
+import json
+
+import ddt
+import mock
+from django.urls import reverse
+
+from course_discovery.apps.api.tests.jwt_utils import generate_jwt_header_for_user
+from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase
+from course_discovery.apps.core.tests.factories import UserFactory
+
+
+JSON_CONTENT_TYPE = 'application/json'
+
+
+@ddt.ddt
+@mock.patch('django.conf.settings.USERNAME_REPLACEMENT_WORKER', 'test_replace_username_service_worker')
+class UsernameReplacementViewTests(APITestCase):
+    """ Tests UsernameReplacementView """
+    SERVICE_USERNAME = 'test_replace_username_service_worker'
+
+    def setUp(self):
+        super(UsernameReplacementViewTests, self).setUp()
+        self.service_user = UserFactory(username=self.SERVICE_USERNAME)
+        self.url = reverse("api:v1:replace_usernames")
+
+    def build_jwt_headers(self, user):
+        """
+        Helper function for creating headers for the JWT authentication.
+        """
+        token = generate_jwt_header_for_user(user)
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/v1/views/user_management.py
```diff
@@ -0,0 +1,142 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class UsernameReplacementView(APIView):
+    """
+    WARNING: This API is only meant to be used as part of a larger job that
+    updates usernames across all services. DO NOT run this alone or users will
+    not match across the system and things will be broken.
+
```
bessiesteinberg, line 23: Also add note that this can only be called AFTER calling the edx-platform api b/c it does not do error checking on name uniqueness

course_discovery/apps/api/v1/tests/test_views/test_user_management.py
```diff
@@ -0,0 +1,90 @@
...
+        fake_usernames = ["myname_" + str(x) for x in range(5)]
+        existing_users = [{user.username: user.username + '_new'} for user in random_users]
+        non_existing_users = [{username: username + '_new'} for username in fake_usernames]
+        data = {
+            "username_mappings": existing_users + non_existing_users
+        }
+        expected_response = {
+            'failed_replacements': [],
+            'successful_replacements': existing_users + non_existing_users
+        }
```
bessiesteinberg, line 91: Is the expected behavior that there will be successful replacements for non-existing users?  Are new user records created for users if the old username doesn't exist? It's fine if that is the case just add that to the docstring for the api and maybe add a comment about that in this test

bessiesteinberg (approved):
Looks good, just a few comments I think need clarification after that should be good to merge (assuming any dependencies are also merged)

Also does pylint have a like "dangerous" tag like React's Dangerously Set Inner HTML (https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html) ?  B/c if there is an easy way to tag this api as dangerous that would be great. I do have concerns about this getting productized more broadly.  This shouldn't block any merging, just a thing to think about.

course_discovery/apps/api/v1/tests/test_views/test_user_management.py
tuchfarber: Copied from credentials:
> I'll definitely add some more comments around this in the services as this was something I realized later and changed. If a user never views a credential, they aren't going to exist on the credentials service at all. So we should treat the lack of a user a success as we don't need to do any work.

course_discovery/apps/api/v1/views/user_management.py
tuchfarber: Fixed

course_discovery/apps/api/v1/tests/test_views/test_user_management.py
bessiesteinberg: great i saw that you clarified it in the comments on one of the other PR's and it makes sense to me now - thanks!

georgebabey (approved):

PR #1800, opened 2019-02-27
DISCO-583: Adds in decision doc for Publisher Draft state

jlajoie:


docs/decisions/0004-publisher-draft.rst
```diff
@@ -0,0 +1,64 @@
+Publisher Draft Model Decisions
+===============================
+
+Status
+------
+
+Accepted
```
Dillon-Dumesnil, line 7: 👍

docs/decisions/0004-publisher-draft.rst
```diff
@@ -0,0 +1,64 @@
...
+history for. This will prevent us from increasing our table size, as well as
+prevent us from having to modify our APIs outside of a specific query param
+for unpublished data. The default manager will need to query against the
+"published" states. Our schema will also be up to date via default migrations,
+and any consumer of the API will be able to act directly on reading from either
+publisher or draft rows.
+
+On a successful "publish", the data of the "draft" row will be written to the
+"published" row. The draft row does not need to be updated outside of this.
+Consequences
```
Dillon-Dumesnil, line 61: You need a line break here

docs/decisions/0004-publisher-draft.rst
```diff
@@ -0,0 +1,64 @@
...
+Add an additional column representing the differing state for the forms we need
+history for. This will prevent us from increasing our table size, as well as
+prevent us from having to modify our APIs outside of a specific query param
+for unpublished data. The default manager will need to query against the
+"published" states. Our schema will also be up to date via default migrations,
+and any consumer of the API will be able to act directly on reading from either
+publisher or draft rows.
+
+On a successful "publish", the data of the "draft" row will be written to the
+"published" row. The draft row does not need to be updated outside of this.
```
Dillon-Dumesnil: Should this be `The published row does not need to be updated outside of this.` The draft row is the one we are always writing to

Dillon-Dumesnil (commented):
Decent first pass

docs/decisions/0004-publisher-draft.rst
```diff
@@ -0,0 +1,65 @@
...
+
+- Base object manager classes will need to be overridden
+
+- Primary/Composite Primary keys will need to consider the draft/publish state
+
+- Historical changes will not exist by default, it will be difficult to rollback
+  and difficult to restore revisions
+
+- Relations for many to many, or one to many may not be the easiest to propagate
+  to the live "published" versions (update vs drop/create)
```
staubina: We should add a section with the other options we considered.  I think that it will provide historical value.

staubina: Would it make sense to add a section highlighting the benefits of this choice?

staubina (changes requested):
Overall looks good.  I left a couple of comments about sections we might want to add for clarity and historical context.

docs/decisions/0004-publisher-draft.rst
jlajoie: Ahh sure, I sort of covered that in the decision section if a bit informally but I can break it down into bullets which might be easier to consume

jlajoie: Yeah can do, this along with the lower part for benefits will at least make it look like we did our due diligence so down the road people won't be questioning how we decided on doing it this way

docs/decisions/0004-publisher-draft.rst
jlajoie: Sure 👍

staubina (approved):

PR #1801, opened 2019-02-27
API query improvements

jlajoie:


course_discovery/apps/api/serializers.py
```diff
@@ -304,7 +305,11 @@ class MinimalPersonSerializer(serializers.ModelSerializer):
     def prefetch_queryset(cls):
         return Person.objects.all().select_related(
             'position__organization'
-        ).prefetch_related('person_networks')
+        ).prefetch_related(
+            'person_networks',
+            'areas_of_expertise',
+            'position__organization__partner'
```
bbaker6225, line 311: Potential area for `Prefetch` object, although that's a lot of nesting

course_discovery/apps/api/serializers.py
```diff
@@ -812,16 +815,17 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None):
...
             'prerequisites',
-            'subjects',
+            Prefetch('subjects', queryset=SubjectSerializer.prefetch_queryset()),
+            Prefetch('entitlements', queryset=CourseEntitlementSerializer.prefetch_queryset()),
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
             Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
             Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
         )
 
+    # Executes as an n+1 query because of dependence on context
```
bbaker6225, line 843: It seems like we should be able to pass context through https://github.com/edx/course-discovery/blob/91f691ff809d5f48ff4e28831d4b29f3ecce8bbd/course_discovery/apps/api/v1/views/courses.py#L126-L133

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -28,7 +28,7 @@ class CourseRunViewSet(viewsets.ModelViewSet):
     lookup_value_regex = COURSE_RUN_ID_REGEX
     ordering_fields = ('start',)
     permission_classes = (IsAuthenticated, DjangoModelPermissions)
-    queryset = CourseRun.objects.all().order_by(Lower('key'))
+    queryset = CourseRun.objects.all().prefetch_related('seats').order_by(Lower('key'))
```
bbaker6225: `Prefetch('seats', queryset=SeatSerializer.prefetch_queryset())` ?

course_discovery/apps/course_metadata/models.py
```diff
@@ -649,12 +648,27 @@ class CourseRun(TimeStampedModel):
...
+        return datetime.datetime.max.replace(tzinfo=pytz.UTC)
+
     def _enrollable_paid_seats(self):
         """
         Return a QuerySet that may be used to fetch the enrollable paid Seats (Seats with price > 0 and no
         prerequisites) associated with this CourseRun.
         """
-        return self.seats.exclude(type__in=Seat.SEATS_WITH_PREREQUISITES).filter(price__gt=0.0)
+        # If we already have seats fetched, do the filtering in python to save DB calls
+        if hasattr(self, '_prefetched_objects_cache') and 'seats' in self._prefetched_objects_cache:
```
bbaker6225: The docstring on this method should be updated, and I think the second return should also be cast to a list

course_discovery/apps/api/serializers.py
```diff
@@ -719,7 +724,7 @@ def prefetch_queryset(cls, queryset=None, course_runs=None):
 
         return queryset.select_related('partner').prefetch_related(
             'authoring_organizations',
```
jmyatt, line 734: to what extent could there be additional prefetch optimizations we're not handling in this PR, such as using a Prefetch here with OrganizationSerializer.prefetch_queryset?  Is the approach that only egregiously high numbers of queries get attention, so that we avoid premature optimization?

jmyatt:
can you add a note about some of the improvements this PR achieves?  # of sql queries pre/post on a given endpoint, etc?  Nothing super detailed, just a sense of the overall impact

course_discovery/apps/api/serializers.py
bbaker6225: There definitely could be optimizations that we missed. Jeff has an old copy of prod discovery on his devstack, so our approach was to work on reducing/eliminating queries that were duplicated an egregious amount for each endpoint using his large set of data

bbaker6225:
@jmyatt I believe @jlajoie has screenshots of the local changes in query count for each endpoint touched.  I don't have enough local data to demonstrate a significant change

jmyatt:
One thing Cale did was to add additional staff in the test factories to see what that breaks, and it did affect some things:

https://github.com/edx/course-discovery/pull/1797/files#diff-6e8ba091821c649daba08336e4f1ab60R161

can we consider doing that?

jmyatt:
About the new test failures:
- search sanity check just needs a higher expected result count
- `staff_uuids ` issue might simply need the assertion to account for a different ordering
- in some cases there is now a slightly higher query count than expected - also might be unimportant and the expected query count should be adjusted, but we should check that

course_discovery/apps/api/serializers.py
```diff
@@ -345,13 +350,13 @@ def get_urls_detailed(self, obj):
             'title': network.title,
             'display_title': network.display_title,
             'url': network.url,
-        } for network in obj.person_networks.all().order_by('id')]
+        } for network in sorted(obj.person_networks.all(), key=attrgetter('id'))]
```
mikix, line 357: attrgetter is fun, haven't used that before!

This sort of thing might benefit from a comment? I can imagine someone coming by in a year and being like "don't these fools know about order_by?" and switching it.

course_discovery/apps/api/serializers.py
```diff
@@ -600,7 +605,7 @@ def prefetch_queryset(cls, queryset=None):
 
         return queryset.select_related('course').prefetch_related(
             'course__partner',
-            Prefetch('seats', queryset=SeatSerializer.prefetch_queryset()),
+            Prefetch('seats', queryset=SeatSerializer.prefetch_queryset())
```
mikix: Aw, don't be removing trailing commas, I love those.

course_discovery/apps/api/serializers.py
```diff
@@ -719,7 +724,7 @@ def prefetch_queryset(cls, queryset=None, course_runs=None):
 
         return queryset.select_related('partner').prefetch_related(
             'authoring_organizations',
-            'entitlements',
+            Prefetch('entitlements', CourseEntitlementSerializer.prefetch_queryset()),
```
mikix: I usually see this passed as a kwarg, not an arg. Both work because of the source definition, but probably better to stick to kwarg.

course_discovery/apps/api/serializers.py
```diff
@@ -750,11 +755,17 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None):
...
+            'video',
+            'video__image',
+            'partner',
+            'extra_description'
+        ).prefetch_related(
             'expected_learning_items',
             'prerequisites',
             'subjects',
-            'entitlements',
+            'topics',
```
mikix, line 776: entitlements is the only prefetch you remove - you found it wasn't helping at all?

course_discovery/apps/course_metadata/models.py
```diff
@@ -504,7 +503,7 @@ def active_course_runs(self):
 
     @property
     def first_enrollable_paid_seat_price(self):
-        for course_run in self.active_course_runs.order_by(Lower('key')):
+        for course_run in sorted(list(self.active_course_runs), key=lambda active_course_run: active_course_run.key):
```
mikix: No attrgetter here?! Same note about a comment though.

I wonder what the lower was about. That does seem unnecessary here.

course_discovery/apps/course_metadata/models.py
```diff
@@ -649,12 +648,24 @@ class CourseRun(TimeStampedModel):
...
-        Return a QuerySet that may be used to fetch the enrollable paid Seats (Seats with price > 0 and no
+        Return a list that may be used to fetch the enrollable paid Seats (Seats with price > 0 and no
         prerequisites) associated with this CourseRun.
         """
-        return self.seats.exclude(type__in=Seat.SEATS_WITH_PREREQUISITES).filter(price__gt=0.0)
+        seats = []
+        for seat in self.seats.all():
+            if seat.type not in Seat.SEATS_WITH_PREREQUISITES and seat.price > 0.0:
+                seats.append(seat)
+        return seats
```
mikix, line 678: 1) Can you explain the performance logic her to me, just for my own knowledge? (And maybe add a comment with a quick note about same) You're still hitting the database with .all(), so same number of queries, right? Is mysql that bad with set or number comparisons?

2) You also switch the return type to a list, but keep calling list() on the result elsewhere.

3) has_enrollable_paid_seats() had an optimization to slice the queryset being returned by this method. I don't know if that optimization is worth it. But at the least, you might want to rewrite has_enrollable_paid_seats to no longer do the slice thing.

course_discovery/apps/course_metadata/models.py
```diff
@@ -665,7 +676,8 @@ def clean(self):
 
     @property
     def first_enrollable_paid_seat_price(self):
-        seats = list(self._enrollable_paid_seats().order_by('upgrade_deadline'))
+        # Just do the sort in python to keep parity, but don't make an additional request for order_by
+        seats = sorted(list(self._enrollable_paid_seats()), key=self._upgrade_deadline_sort)
```
mikix: Nice comment! Though it mentions parity, and a year from now, I'm not going to know parity with what. Might just want to say "Do the sort in python to avoid wiping out any queryset preselects" or something like that.

mikix (approved):
I have no comments on the whole preselect stuff - I assume your changes are the best. But I left some non-blocking comments here and there about smaller things.

bbaker6225:
@jmyatt 
- Updated the sanity check
- Not sure how to handle the ordering of staff at the moment, but I'll keep looking into that
- The query count jumps seem pretty big for some of these test failures, but I'm not sure how many of them are unavoidable

course_discovery/apps/api/serializers.py
bbaker6225: 'twas an accident, I would never remove them

course_discovery/apps/course_metadata/models.py
bbaker6225: I think we wanted the sorting key so we could call `lower()` in python on the course run key

jmyatt (approved):

course_discovery/apps/course_metadata/models.py
bbaker6225: 1. From the django docs: `We can reduce to just two queries using prefetch_related:

>>> Pizza.objects.all().prefetch_related('toppings')
This implies a self.toppings.all() for each Pizza; now each time self.toppings.all() is called, instead of having to go to the database for the items, it will find them in a prefetched QuerySet cache that was populated in a single query.`

So we should have seats.all() prefetched (for the API from the serializer) and ready to go, but exclude/filter are separate queries and so the prefetching wouldn't help us. I'll add a comment on this

2. Good point

3. I can change the slice. It would be nice if we could keep the slice with a generator or something but not sure if that will work here

course_discovery/apps/api/serializers.py
bbaker6225: I can't remember, I'll add it back since I doubt it hurts anything

course_discovery/apps/course_metadata/models.py
mikix: But you drop the lower call.

course_discovery/apps/course_metadata/models.py
```diff
@@ -504,7 +504,14 @@ def active_course_runs(self):
...
     @property
     def first_enrollable_paid_seat_price(self):
-        for course_run in self.active_course_runs.order_by(Lower('key')):
+        """
+        Sort the course runs with sorted rather than order_by to avoid
+        additional calls to the database
+        """
+        for course_run in sorted(
+            list(self.active_course_runs),
+            key=lambda active_course_run: lower(active_course_run.key)
```
cpennington: Why did you make a list out of `self.active_course_runs`? `sorted` accepts an iterable as the first argument.

course_discovery/apps/api/serializers.py
cpennington: The intent of `prefetch_queryset` was that it prefetch all of the fields that this serializer is going to serialize out. So, I would fetch `entitlements` only if this serializer is going to render them, which it looks like it won't based on my cursory reading of `Meta.fields`.

course_discovery/apps/course_metadata/models.py
bbaker6225: Added back in

course_discovery/apps/course_metadata/models.py
bbaker6225: 👍 good catch

bbaker6225:
After talking with @jmyatt , we've decided to remove the change to create extra staff in `factories.py` and handle the work around that down the road. Please let me know if there are any other changes you'd like to see in this PR

jlajoie:
Local update for 2353 Courses, 4687 Course Runs, 11 Programs.

/courses
screen shot 2019-03-04 at 8 53 08 am
screen shot 2019-03-04 at 8 56 19 am

/courses/:uuid
screen shot 2019-03-04 at 8 53 32 am
screen shot 2019-03-04 at 8 56 38 am

/programs
screen shot 2019-03-04 at 8 54 45 am
screen shot 2019-03-04 at 8 57 07 am

/programs/:uuid (unchanged)

cpennington:
@bbaker6225: When are you going to add the extra data to the factories? Until you add those staff members (and really, add instances of all foreign-key data types), the tests are at high risk of missing badly performing query patterns due to missing `select_related` or `prefetch_related` calls.

bbaker6225:
@cpennington At the moment our team is occupied with trying to launch the new prospectus topic pages this week, and last week was the amount of time we were given to work on this performance stuff. I agree that it's a good thing to add, but I'm not the right person to prioritize that work

cpennington:
@rock345 course-discovery performance is going to continue to be unexpectedly bad until we land something like my suggestion above.

jlajoie:
@cpennington agreed that we should add those changes so our tests more accurately reflect what we're expecting to serve. I'll create a follow-up ticket when the Disco squad has cycles we can address those factories, but I think getting these few tweaks in now is a big enough win that we can move on.

Is there anything I missed in the comment chain above that you still think is actionable outside of updating the factories for the relationships?

cpennington:
No, my main concern is just getting enough test data around the various related objects to validate that query performance isn't going to go off a cliff at some point.

cpennington:
Although, looking at the query counts in your local instance, I suspect we might be missing some joins... Have you checked in django-debug-toolbar to see if any queries are listed as "Duplicate query detected" (I forget the exact wording it uses in the query panel, but it tells you if the same query was run w/ different arguments).

jmyatt:
@jlajoie let me know when you've created the follow-on api performance ticket and I'll make sure that we capture all notes there on the outstanding pieces of this, and that the work gets prioritized so that we avoid the surprise performance hit down the road that @cpennington is concerned about

PR #1802, opened 2019-02-28
Add drf_dynamic_fields to programs

bbaker6225:
screen shot 2019-02-28 at 11 34 57 am

### DIFF
--- a/course_discovery/apps/api/v1/views/programs.py
+++ b/course_discovery/apps/api/v1/views/programs.py
@@ -1,4 +1,5 @@
 from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import filters as rest_framework_filters
 from rest_framework import viewsets
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/v1/views/programs.py
```diff
@@ -16,7 +17,8 @@ class ProgramViewSet(CacheResponseMixin, viewsets.ReadOnlyModelViewSet):
     lookup_field = 'uuid'
     lookup_value_regex = '[0-9a-f-]+'
     permission_classes = (IsAuthenticated,)
-    filter_backends = (DjangoFilterBackend,)
+    # Check if there's available syntax for ordering by join children elements
```
mikix: Is that for future folks or yourself?

bbaker6225: I think the original intent from courses (https://github.com/edx/course-discovery/blob/0a906a2702784406ab20d844e51103bdd30046cb/course_discovery/apps/api/v1/views/courses.py#L61) was that it would be for future folks, but I don't know if it adds much value, so I'll just remove it from both spots

mikix (approved):
But yeah regardless of my comment question, seems great!

PR #1803, opened 2019-02-28
Add scaffolding for CurriculumCourseMembership to automatically add c…

reillz10:
…ourse mode to courses

### DIFF
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -1,11 +1,16 @@
+import logging
+
 import waffle
 from django.apps import apps
 from django.db.models.signals import post_delete, post_save, pre_delete
 from django.dispatch import receiver
 
 from course_discovery.apps.api.cache import api_change_receiver
-from course_discovery.apps.course_metadata.models import Program
+from course_discovery.apps.course_metadata.models import CurriculumCourseMembership, Program
 from course_discovery.apps.course_metadata.publishers import ProgramMarketingSitePublisher
+from course_discovery.apps.course_metadata.waffle import masters_course_mode_enabled
+
+logger = logging.getLogger(__name__)
 
 
 @receiver(pre_delete, sender=Program)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1785,6 +1786,14 @@ class CurriculumCourseRunExclusion(TimeStampedModel):
 
     history = HistoricalRecords()
 
+    def save(self, *args, **kwargs):
```
schenedx: Why is this hook added to the CourseRunExclusion model, but not the CurriculumCourseMembership?

schenedx (commented):
See inline question

course_discovery/apps/course_metadata/models.py
```diff
@@ -1785,6 +1786,14 @@ class CurriculumCourseRunExclusion(TimeStampedModel):
...
+        if masters_course_mode_enabled():
```
iloveagent57: I think when we actually implement this, it would be good to utilize the `post_save` Django signal.  We'd define a receiver function, connect it to the `post_save` signal from this class, and then the receiver function can be in charge of adding the masters course mode.
https://docs.djangoproject.com/en/1.11/topics/signals/

iloveagent57 (commented):
This looks good.  If you're up for it, we can convert to the signals-style implementation now (and have your receiver just log the same message that you're currently logging in `save()) - I could walk through it quickly with you this afternoon.

course_discovery/apps/course_metadata/models.py
iloveagent57: Oh, haha, yeah, same question - how did we even see these tests pass?

reillz10: I'm not sure at all. It was on CurriculumCourseMembership. I must have accidentally done something weird in my editor right before pushing. Will update accordingly

reillz10: It was a rebase issue! I fixed it now.

course_discovery/apps/course_metadata/models.py
reillz10: Gave it my best shot! If you get a chance, please take a look

reillz10: Okay, things are now moved to their proper homes.

iloveagent57 (approved):

PR #1805, opened 2019-03-04
add autocomplete to curriculum course membership/exclusion

jansenk:
## Curriculum admin page with autofill selectors for course and program, with displays of included and excluded course runs
screen shot 2019-03-04 at 11 21 54 am

## Curriculum Course membership admin page with autofill selector, filtered to only show runs of the selected course
screen shot 2019-03-04 at 11 27 05 am


course_discovery/apps/course_metadata/admin.py
```diff
@@ -334,18 +337,39 @@ class LevelTypeAdmin(SortableAdminMixin, admin.ModelAdmin):
...
+    form = CurriculumCourseMembershipInlineAdminForm
+    readonly_fields = ("custom_course_runs_display", "course_run_exclusions", "get_edit_link",)
+
+    def custom_course_runs_display(self, obj):
+        return mark_safe('<br>'.join([str(run) for run in obj.course_runs]))
+
+    custom_course_runs_display.short_description = _('Included course runs')
+
+    def get_edit_link(self, obj=None):
+        if obj.pk:
```
schenedx: Do we need to handle the case when passed in `obj` is `None`?

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -24,6 +24,11 @@ class CourseRunAutocomplete(autocomplete.Select2QuerySetView):
     def get_queryset(self):
         if self.request.user.is_authenticated() and self.request.user.is_staff:
             qs = CourseRun.objects.all().select_related('course')
+
+            course = self.forwarded.get('course', None)
```
schenedx: Can we have a better variable? Like `filter_by_course` or something more descriptive?

schenedx (commented):
Nice work in all. 
Do we have ways to write unit tests for Forms and Admins classes? I see we have admin class tests. It will be great if you can write some tests for the Forms classes as well.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1803,6 +1803,16 @@ class CurriculumCourseMembership(TimeStampedModel):
...
 
+    @property
+    def course_runs(self):
+        excluded_course_run_ids = [course_run.id for course_run in self.course_run_exclusions.all()]
+        for run in self.course.course_runs.all():
+            if run.id not in excluded_course_run_ids:
+                yield run
+
+    def __str__(self):
+        return str(self.curriculum) + " : " + str(self.course)
```
iloveagent57, line 1811: Nit: consider updating the `Curriculum` class's `__str__()` method to use the name instead of the uuid (maybe only use the uuid if name is empty).

course_discovery/apps/course_metadata/models.py
```diff
@@ -1803,6 +1803,16 @@ class CurriculumCourseMembership(TimeStampedModel):
 
     history = HistoricalRecords()
 
+    @property
+    def course_runs(self):
+        excluded_course_run_ids = [course_run.id for course_run in self.course_run_exclusions.all()]
```
iloveagent57: Nit: consider using set arithmetic:
```python
return set(self.course.course_runs.all()) - set(self.course_run_exclusions.all())
```
That should work as long as `CourseRun` is hashable.

iloveagent57 (commented):
Just a couple of minor suggestions that are probably worth doing if you're fixing up tests.

iloveagent57 (approved):
Looks good, nice work!

schenedx (approved):
👍

PR #1806, opened 2019-03-04
Support create and patch in course run API

mikix:
Allow POST, PATCH, and PUT on the main course run API. And push to studio when changes are made too.

https://openedx.atlassian.net/browse/DISCO-622


course_discovery/apps/api/permissions.py
```diff
@@ -29,3 +29,29 @@ def has_object_permission(self, request, view, obj):
...
+        self.django_perms = DjangoModelPermissions()
+
+    def has_permission(self, request, view):
+        if self.django_perms.has_permission(request, view):
+            return True
+        elif request.user.is_staff:
+            return True
+        elif request.method == 'POST':
+            course = request.data.get('course')
+            org = course and course.split('+')[0].split('/')[0]  # handle old and new style keys
```
staubina: Are the Opaque keys available in Discovery?  If they are we should use CourseKey here.

staubina: Example here: https://github.com/edx/course-discovery/blob/dd8f65abc75dc9a424fe5423f1905399f8a990c9/course_discovery/apps/course_metadata/data_loaders/marketing_site.py

course_discovery/apps/api/permissions.py
```diff
@@ -29,3 +29,29 @@ def has_object_permission(self, request, view, obj):
             return True
         else:
             return CourseEditor.is_course_editable(request.user, obj)
+
+
+class IsCourseRunEditorOrDjangoOrReadOnly(BasePermission):
```
staubina, line 35: Can we combine this with the Course permissions method that we created?

course_discovery/apps/api/serializers.py
```diff
@@ -636,20 +645,67 @@ def get_marketing_url(self, obj):
...
+                    break
+            except ObjectDoesNotExist:
+                pass  # no organization extension
+
+        if allow_key_override and 'key' in data:
+            return
+
+        # Now, override whatever value for key was provided by looking at the start date
+        start = data['start']  # required
+        key_sep = '+' if '+' in course.key else '/'
```
staubina: We should use CourseKey here. It will save us a lot of work and understands CourseKey formatting

course_discovery/apps/api/utils.py
```diff
@@ -1,8 +1,13 @@
 import hashlib
 import logging
-
+import math
 import six
 
+from opaque_keys.edx.keys import CourseKey
```
staubina, line 6: Just wanted to point to the use of CourseKey here as a possible example.

course_discovery/apps/api/utils.py
```diff
@@ -65,3 +70,132 @@ def get_cache_key(**kwargs):
     key = '__'.join(['{}:{}'.format(item, value) for item, value in six.iteritems(kwargs)])
 
     return hashlib.md5(key.encode('utf-8')).hexdigest()
+
+
+class StudioAPI:
+    """ A convenience class for talking to the Studio API - designed to allow subclassing for publisher models. """
```
staubina: Is there any need to update this comment?

staubina: Also is this all new code or a move from the old Publisher code?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -1,24 +1,43 @@
...
 
 
+def writable_request_wrapper(method):
+    def inner(*args, **kwargs):
+        try:
+            with transaction.atomic():
+                return method(*args, **kwargs)
+        except (PermissionDenied, ValidationError):
+            raise  # just pass along
+        except Exception as e:  # pylint: disable=broad-except
```
staubina, line 34: Do we need a broad exception here?  Is there something that we are trying to catch?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -133,9 +154,42 @@ def list(self, request, *args, **kwargs):
         """
         return super(CourseRunViewSet, self).list(request, *args, **kwargs)
 
-    def partial_update(self, request, *args, **kwargs):
+    @classmethod
+    def push_to_studio(cls, request, course_run, create=False, old_course_run=None):
+        if course_run.course.partner.studio_url:
```
staubina, line 159: What do we do id the URL is not available or set?  Should we log or anything?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -133,9 +154,42 @@ def list(self, request, *args, **kwargs):
...
+        serializer.is_valid(raise_exception=True)
+
+        # Grab any existing course run for this course (we'll use it when talking to studio to form basis of rerun)
+        course_key = request.data['course']  # required field
+        course = Course.objects.get(key=course_key)
+        old_course_run = course.canonical_course_run
+
+        # And finally, save to database and push to studio
+        course_run = serializer.save()
+        self.push_to_studio(request, course_run, create=True, old_course_run=old_course_run)
```
staubina, line 182: Do we care if this fails or succeeds?

course_discovery/apps/publisher/signals.py
```diff
@@ -44,40 +44,20 @@ def create_course_run_in_studio_receiver(sender, instance, created, **kwargs):
...
-                    ex.content
-                )
-                raise
-        else:
-            try:
-                logger.info('Creating a new run of [%s]...', instance.course.key)
-                response = api.create_course_run_in_studio(instance)
-            except SlumberBaseException as ex:
-                logger.exception('Failed to create course run [%s] on Studio: %s', instance.id, ex.content)
-                raise
```
staubina, line 67: Just looking at these changes at a high level, will they impact old publisher in any way?

course_discovery/apps/publisher/signals.py
```diff
@@ -44,40 +44,20 @@ def create_course_run_in_studio_receiver(sender, instance, created, **kwargs):
...
+        try:
+            response = api.push_to_studio(instance, create=True, old_course_run=discovery_course_run)
+        except SlumberBaseException as ex:
+            logger.exception('Failed to create course run [%s] on Studio: %s', course.key, ex.content)
+            raise
+        except Exception:  # pylint: disable=broad-except
```
staubina, line 54: I think we should be cautious with broad exception catches, are there specific exceptions we are looking for here?

staubina (changes requested):
First run through with comments.  I will try to look again after standup.

course_discovery/apps/api/permissions.py
mikix: As discussed in IRC, OpaqueKeys is not sufficient here because the parsing you are commenting on is parsing of course key fragments, not full course keys.

I'm adding a utility function to make this clear and collect the two instances of me parsing the course key in one place.

Good call out.

course_discovery/apps/api/permissions.py
mikix: I think two classes fits more naturally for this case:

(1) The course run endpoint also looks at django permissions. And combining "or" permissions kind of has to be done all together in one permission class. But the course endpoint doesn't need that django permission check.

(2) Looking up the course changes a lot based on whether we are at a course or course run endpoint. In one former, the course / org is provided in the incoming data. In the latter, we are looking up a course run / course from the incoming data. So while they call the same permission checking functions, the request parsing is all different.

course_discovery/apps/api/utils.py
mikix: This is a move from old publisher with additional support added for subclassing and overriding all the utility bits like "how do I get the course run key from a course run" - that changes between CourseMetadata and Publisher tables.

The comment is a new comment - doesn't need to be updated. Although maybe it needs to be updated to be better, if you're asking the question? :)

course_discovery/apps/api/v1/views/course_runs.py
mikix: If the partner doesn't have a studio url set (which is an optional field), I figured we don't want to fill their logs with warnings about it.

course_discovery/apps/api/v1/views/course_runs.py
mikix: If it fails, it will raise an exception that is caught by writable_request_wrapper and returns a 400 to the caller.

course_discovery/apps/publisher/signals.py
mikix: The hope is not. Old publisher (in this code block) was always doing "1) check if re-run, then create re-run or run, 2) update icon" -- the new endpoint wanted to do the same thing. So I moved the logic into the shared StudioAPI class. So this is hopefully just a cleanup, not a change in behavior.

course_discovery/apps/api/v1/views/course_runs.py
mikix: This is a "did anything go wrong handling this API call" exception catch. Whether that should be broad or not is possibly a philisophical issue. I lean towards broad catches being acceptable for cases like this, because we are on a "boundary" where regardless of what goes wrong, we want to present a uniform answer to the world outside the boundary.

course_discovery/apps/publisher/signals.py
mikix: This specific instance is actually a tightening of a bare-except in the old code that I was refactoring. I'm not sure what all the exceptions it expects or cares about are. Probably just SlumberBaseException in practice.

I *suspect* the original bare except was added simply so they can log a message that tests can pick up. They have one test that adds a side effect exception and checks that it was caught. :man_shrugging: I'd be fine with dropping this except (which simply logs and raises anyway) and reworking that test.

course_discovery/apps/api/utils.py
mikix: Expanded the comment slightly.

course_discovery/apps/api/v1/views/course_runs.py
staubina: When would we care to know about this though?

course_discovery/apps/publisher/signals.py
staubina: Have you been able to do any manual testing with old publisher to verify we did not break it?

staubina (approved):
Overall the changes look good to me.  Should we plan some manual testing if we think it is necessary for old publisher though?  As a result of moving some of the code to shared functionality.

course_discovery/apps/api/v1/views/course_runs.py
mikix: Devstack didn't have it set, but I think that would be an expected situation. If it weren't set in prod, I imagine various little integration points with studio wouldn't work.

I don't know when we would care to know. I'm happy to throw an info log in there - I imagine most prod instances have that field set and wouldn't be spammed so much?

course_discovery/apps/publisher/signals.py
mikix: Yes, I can manually create and update the studio course info and image from old publisher.

course_discovery/apps/api/v1/views/course_runs.py
mikix: Added an info output here and a test to confirm it.

PR #1807, opened 2019-03-07
Updates to edit course api to handle video and image saves. Only crea…

Dillon-Dumesnil:
…te entitlements for paid seats

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -378,7 +378,7 @@ class CourseViewSetTests(SerializationMixin, APITestCase):
 
     @oauth_login
     @responses.activate
-    def test_create_with_authentication(self):
+    def test_create_with_authentication_verified_mode(self):
         url = reverse('api:v1:course-list')
         course_data = {
             'title': 'Course title',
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -998,6 +998,7 @@ class Seat(TimeStampedModel):
     CREDIT = 'credit'
 
     SEAT_TYPES = [HONOR, AUDIT, VERIFIED, PROFESSIONAL, CREDIT]
+    PAID_SEATS = [VERIFIED, PROFESSIONAL]
```
mikix: This is a potentially confusing term, especially since CREDIT does involve payment. common/djangoapps/course_modes/models.py in platform does not have this term.

Closest it has is VERIFIED_MODES (which these days includes MASTERS, but we can ignore that for now). 

Maybe the best term is just ENTITLEMENT_MODES?

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -489,7 +513,7 @@ def test_create_with_api_exception(self):
             'mode': 'audit',
         }
         with mock.patch(
-            'course_discovery.apps.api.v1.views.courses.CourseViewSet.perform_create',
+            'course_discovery.apps.api.v1.views.courses.CourseViewSet.get_course_key',
```
mikix, line 518: Such a suspicious line, but we talked about this in person.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -248,6 +255,19 @@ def update_course(self, data, partial=False):
         for entitlement_data in entitlements_data:
             entitlements.append(self.update_entitlement(course, entitlement_data, partial=partial))
 
+        # Save video if a new video source is provided
+        if video_src:
+            video, __ = Video.objects.get_or_create(src=video_src)
+            course.video = video
```
mikix, line 261: So in my PR, I expect the same input as output for video (that is, I expect to be given `video: {src: xxx}`). Only on the theory that maybe we should/would eventually support setting the other fields of the video model.

I don't have a strong preference, but we should be in alignment at least. Do you think my way is unnecessary, since we don't set those fields right now anyway?

Image has the same question, but a less clear analog between input and output. What do other similar APIs in our system do for situations like this?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -248,6 +255,19 @@ def update_course(self, data, partial=False):
...
+
+        # Save image and convert to the correct format
+        if image_data and isinstance(image_data, str) and image_data.startswith('data:image'):
+            # base64 encoded image - decode
+            file_format, imgstr = image_data.split(';base64,')  # format ~= data:image/X;base64,/xxxyyyzzz/
+            ext = file_format.split('/')[-1]  # guess file extension
+            image_data = ContentFile(base64.b64decode(imgstr), name='tmp.' + ext)
+            course.image.save("Image name", image_data)
```
mikix: "image name"?

course_discovery/apps/api/v1/tests/test_views/test_courses.py
Dillon-Dumesnil: I added a comment.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: This has been addressed by changing the video to being a dictionary with the key `src`. Image has remained the same.

mikix (approved):
:joy_cat:

PR #1808, opened 2019-03-08
Rir/auto course mode

reillz10:


schenedx:
Looks like you need to re-gen translations again

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -232,6 +232,7 @@ class Meta(object):
 
     name = FuzzyText()
     logo_image = FuzzyText(prefix='https://example.com/program/logo')
+    slug = FuzzyText()
```
schenedx, line 235: Don't believe this is needed anymore

schenedx (approved):
👍

course_discovery/apps/course_metadata/signals.py
```diff
@@ -26,13 +27,75 @@ def delete_program(sender, instance, **kwargs):  # pylint: disable=unused-argume
...
-        make an api call here to ensure that the 'masters' course_mode is added to
-        that course""")
-    # otherwise we will still save but we won't do any other side effects
+        program = instance.curriculum.program
+
+        if program.type.slug != 'masters':
+            return None
+
+        if not program.partner.lms_commerce_api_url:
+            return None
```
iloveagent57, line 41: I would consider adding a helpful logging message here, since hitting this path indicates that we've forgotten to apply a necessary manual step.  Something to the effect of "Cannot add masters mode to course, lms_commerce_api_url not set on partner."

iloveagent57: ...you could do this before you rebase the updates from the master branch into your branch.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -26,13 +27,75 @@ def delete_program(sender, instance, **kwargs):  # pylint: disable=unused-argume
...
+        for course_run in instance.course_runs:
+            _add_masters_track_on_commerce_course_run(course_run, partner.lms_commerce_api_url, commerce_api_client)
+
+
+def _add_masters_track_on_commerce_course_run(course_run, api_url, commerce_api_client):
+    """ Given a course_run, api_url, and a commerce_api_client, this helper method will call the commerce api to
+    add the masters track to a course run. """
+
+    course_run_key = course_run.key
+    url = api_url + 'courses/{}/'.format(course_run_key)
```
iloveagent57, line 55: Nit: It might be better to strip the trailing slash off of `api_url` and explicitly add a lead slash to '/courses/', that way you can safely handle cases where we forgot to add a trailing slash to the base URL via Django admin:
```python
url = api_url.rstrip('/') + '/courses/{}/'.format(course_run_key)
```

iloveagent57 (approved):
I have very minor suggestions that it might be worth doing if you're going to rebase anyway.  Looks good!

PR #1816, opened 2019-03-15
remove django compressor support

DawoudSheraz:
### Description
This PR removes the django-compressor and adds the minified version of some standard libraries that were being compressed. This short term fix unblocks the course teams that were receiving 500 response error on the publisher. The initial solution was to disable OFFLINE_COMPRESSION but that change spawned other errors that related to generating new static files.

The offline compression working comprises compressing the files once and doesn't compress on the runtime. However, the compressed files were for one context only(the output of _python manage.py compress_ pinpointed this). So, if the context were to change, the information would not be in the compressed files and this would throw **compressor.exceptions.OfflineGenerationError:** So, it seemed that disabling it would fix the problem.

Surprisingly, disabling the offline compression initiated something else. Now, if the context changed, new files would be generated.The new files required saving, which caused **PermissionError: [Errno 13] Permission denied:**. This error signified why Offline compression was necessary(to avoid creating new files on the run), but since all the contexts weren't covered in the compressed files, the last resort was to remove the compression entirely.

To avoid the large loading time, which was being done by compressor, minified versions of various libraries & css files were added. This ensured that removing the django-compressor will not increase the response time. There will be a follow up PR that would use the webpack to generate bundles of the files that were previously being compressed. This PR would unblock course teams that were receiving 500 error because of the compression.

### Reviewers
 - [x] @awaisdar001 
 - [x] @asadazam93 

### Sandbox
 - https://discovery-dawoudprod65.sandbox.edx.org/publisher/course_runs/1/
 - username: staff

### Post Review
 - [x] Squash & Rebase Commits


course_discovery/apps/publisher/templates/publisher/base.html
```diff
@@ -59,13 +59,13 @@
     {{ block.super }}
 
     {% compress js %}
-        <script src="{% static 'bower_components/jquery/dist/jquery.js' %}"></script>
+        <script src="{% static 'bower_components/jquery/dist/jquery.min.js' %}"></script>
         <script src="{% static 'admin/js/jquery.init.js' %}"></script>
         <script src="{% static 'bower_components/js-cookie/src/js.cookie.js' %}"></script>
-        <script src="{% static 'bower_components/underscore/underscore.js' %}"></script>
+        <script src="{% static 'bower_components/underscore/underscore-min.js' %}"></script>
         <script src="{% static 'bower_components/moment/moment.js' %}"></script>
```
awaisdar001: Can we check if `moment.js` has a min version available if not, can we have that added here instead of non minified file?

course_discovery/apps/publisher/templates/publisher/base.html
```diff
@@ -59,13 +59,13 @@
     {{ block.super }}
 
     {% compress js %}
-        <script src="{% static 'bower_components/jquery/dist/jquery.js' %}"></script>
+        <script src="{% static 'bower_components/jquery/dist/jquery.min.js' %}"></script>
         <script src="{% static 'admin/js/jquery.init.js' %}"></script>
         <script src="{% static 'bower_components/js-cookie/src/js.cookie.js' %}"></script>
```
awaisdar001, line 61: is this a jquery plugin? can we treat it the same as `underscore-min.js`?

course_discovery/apps/publisher/templates/publisher/base.html
```diff
@@ -59,13 +59,13 @@
...
-        <script src="{% static 'bower_components/underscore/underscore.js' %}"></script>
+        <script src="{% static 'bower_components/underscore/underscore-min.js' %}"></script>
         <script src="{% static 'bower_components/moment/moment.js' %}"></script>
         <script src="{% static 'bower_components/pikaday/pikaday.js' %}"></script>
```
awaisdar001, line 64: Same here.

awaisdar001 (commented):
Just have one concern to add min version of the files which are jquery plugins and we do not change them ever.

DawoudSheraz:
There are two js files, pikaday, and js-cookie, whose minified versions can't be provided. The reason being that [bower](https://bower.io/) doesn't perform minification. It just downloads the packages and saves them. Both [pikaday](https://github.com/owenmead/Pikaday) and [js-cookie ](https://github.com/js-cookie/js-cookie/tree/master/src) don't have min file on their repo (Other libs, like [moment](https://github.com/moment/moment/tree/develop/min) have min file).

awaisdar001:
> There are two js files, pikaday, and js-cookie, whose minified versions can't be provided. The reason being that [bower](https://bower.io/) doesn't perform minification. It just downloads the packages and saves them. Both [pikaday](https://github.com/owenmead/Pikaday) and [js-cookie ](https://github.com/js-cookie/js-cookie/tree/master/src) don't have min file on their repo (Other libs, like [moment](https://github.com/moment/moment/tree/develop/min) have min file).

Ok, Use `.min` file where is it is possible.

awaisdar001:
The PR looks good, but lets check if we can test it on the sandbox.

awaisdar001 (approved):
Looks good to me.

asadazam93 (approved):
Looks good to me. Please rebase

PR #1817, opened 2019-03-15
Update Person querysets to take optional org argument.

dianakhuang:
https://openedx.atlassian.net/browse/DISCO-683

Technically, the search updates aren't necessary, but I don't think they're hurting anyone to have them in there either.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -296,6 +296,26 @@ class PositionSerializer(serializers.ModelSerializer):
             return obj.organization.marketing_url
 
 
+class MinimalOrganizationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Organization
+        fields = ('uuid', 'key', 'name',)
+
+
+class OrganizationSerializer(TaggitSerializer, MinimalOrganizationSerializer):
+    """Serializer for the ``Organization`` model."""
+    tags = TagListSerializerField()
+
+    @classmethod
+    def prefetch_queryset(cls, partner):
+        return Organization.objects.filter(partner=partner).select_related('partner').prefetch_related('tags')
+
+    class Meta(MinimalOrganizationSerializer.Meta):
+        fields = MinimalOrganizationSerializer.Meta.fields + (
+            'certificate_logo_image_url', 'description', 'homepage_url', 'tags', 'logo_image_url', 'marketing_url',
+        )
+
+
 class MinimalPersonSerializer(serializers.ModelSerializer):
     """
     Minimal serializer for the ``Person`` model.
@@ -316,7 +336,7 @@ class MinimalPersonSerializer(serializers.ModelSerializer):
         ).prefetch_related(
             'person_networks',
             'areas_of_expertise',
-            'position__organization__partner'
+            'position__organization__partner',
         )
 
     class Meta(object):
@@ -530,26 +550,6 @@ class CourseEntitlementSerializer(serializers.ModelSerializer):
     class Meta(object):
         model = CourseEntitlement
         fields = ('mode', 'price', 'currency', 'sku', 'expires')
-
-
-class MinimalOrganizationSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Organization
-        fields = ('uuid', 'key', 'name',)
-
-
-class OrganizationSerializer(TaggitSerializer, MinimalOrganizationSerializer):
-    """Serializer for the ``Organization`` model."""
-    tags = TagListSerializerField()
-
-    @classmethod
-    def prefetch_queryset(cls, partner):
-        return Organization.objects.filter(partner=partner).select_related('partner').prefetch_related('tags')
-
-    class Meta(MinimalOrganizationSerializer.Meta):
-        fields = MinimalOrganizationSerializer.Meta.fields + (
-            'certificate_logo_image_url', 'description', 'homepage_url', 'tags', 'logo_image_url', 'marketing_url',
-        )
 
 
 class CatalogSerializer(serializers.ModelSerializer):
@@ -1721,6 +1721,7 @@ class PersonSearchSerializer(HaystackSerializer):
             'bio_language',
             'profile_image_url',
             'position',
+            'organizations',
         )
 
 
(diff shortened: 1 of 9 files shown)


course_discovery/apps/api/v1/views/people.py
```diff
@@ -103,7 +104,18 @@ def retrieve(self, request, *args, **kwargs):
...
         # Only include people from the current request's site
-        return self.queryset.filter(partner=self.request.site.partner)
+        queryset = self.queryset.filter(partner=self.request.site.partner)
+        course_key = self.request.query_params.get('course', None)
+
+        if course_key:
+            try:
+                course = Course.objects.get(key=course_key)
+                orgs = [org.key for org in course.authoring_organizations.all()]
+                queryset = queryset.filter(courses_staffed__course__authoring_organizations__key__in=orgs)
```
dianakhuang: @bbaker6225 any suggestions on improving the performance of these queries is much appreciated.

course_discovery/apps/api/v1/views/search.py
```diff
@@ -80,8 +81,9 @@ def filter_facet_queryset(self, queryset):
         facet_serializer_cls = self.get_facet_serializer_class()
         field_queries = getattr(facet_serializer_cls.Meta, 'field_queries', {})
 
-        # Ensure we only return published, non-hidden items
-        queryset = queryset.filter(published=True).exclude(hidden=True)
+        if self.ensure_published:
```
dianakhuang, line 84: We don't have these fields on `Person` objects, so this endpoint was not returning any data when using these filters.

course_discovery/apps/api/serializers.py
```diff
@@ -316,15 +337,16 @@ def prefetch_queryset(cls):
         ).prefetch_related(
             'person_networks',
             'areas_of_expertise',
-            'position__organization__partner'
+            'position__organization__partner',
+            'person__authoring_organizations',
```
bbaker6225: If you have partner and can pass it, using `Prefetch('authoring_organizations', queryset=MinimalOrganizationSerializer.prefetch_queryset(partner=partner))` might help, but not sure what causes the issues with prefetch

course_discovery/apps/api/v1/views/people.py
bbaker6225: I think you'll want to start from the person serializer and prefetch the course_runs, then get the related down from there going to course then org. I'm not sure what the most performant way to do that considering @ormsbee and I think we have seen some issues with django prefetch.

It might be better to do `Prefetch` and try and limit the queryset like several of our serializers, or maybe it's better to specifically prefetch `courses_staffed__course__authoring_organizations` straight from your person serializer.

I could be wrong though, so I would get a 2nd opinion and ultimately the most accurate thing will probably be to monitor it in new relic when it goes out.

bbaker6225: Although, if you point RCM at your local devstack to get a large local data set, you can get a good idea of the sql query performance by searching for duplicated queries with ddt

course_discovery/apps/api/serializers.py
dianakhuang: FYI: I removed this because it was leftovers from when we were adding an `authoring_organizations` field to the model. Now that we're not, this is not relevant anymore, and I'm pulling it.

course_discovery/apps/api/v1/views/people.py
```diff
@@ -103,7 +104,13 @@ def retrieve(self, request, *args, **kwargs):
 
     def get_queryset(self):
         # Only include people from the current request's site
-        return self.queryset.filter(partner=self.request.site.partner)
+        queryset = self.queryset.filter(partner=self.request.site.partner)
+        org_keys = self.request.query_params.getlist('org', None)
+
+        if org_keys:
+            queryset = queryset.filter(courses_staffed__course__authoring_organizations__key__in=org_keys).order_by()
```
ormsbee, line 113: BTW, you might want to comment exactly why it's like this. I can very easily see someone stopping by later and getting rid of this "unnecessary" `order_by()`, or otherwise modifying the sort order without understanding the consequences.

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -198,6 +199,41 @@ def test_list_filter_by_slug(self):
         self.assertEqual(response.status_code, 200)
         self.assertListEqual(response.data['results'], self.serialize_person([person], many=True))
```
mikedikan, line 201: Should we also have a test_list_with_no_org?

mikedikan (approved):

mikedikan:
One comment that could be a follow-up story if valid

course_discovery/apps/api/serializers.py
```diff
@@ -296,6 +296,26 @@ def get_organization_marketing_url(self, obj):
             return obj.organization.marketing_url
 
 
+class MinimalOrganizationSerializer(serializers.ModelSerializer):
```
mikix, line 299: Looks like no changes, just moving them around?

mikix (approved):
Clearing the order_by is done for performance or to avoid some bug?

Having the org-attachment be implicit rather than explicit is going to make trying to figure out what duplicates exist much harder in the future. But that's future-us's problem. I agree that it's usually nice to avoid fields that can just be derived.

This PR would have been much smaller and less likey to introduce a regression without all the search index stuff. But I assume you were already just far enough with that that you threw them in. I don't understand the ramifications of the changes there as well as I wish. But seems like it was busted before. So can't be much worse.

(I believe instructor support for searches was added rather recently by the Dahlia team(?) because they were going to use elasticsearch to fix instructor searching - but when we got handed the work, we just tweaked the existing endpoint to work correctly - so I don't think anyone has ever intentionally used the search endpoint for instructors. Might be a kindness to remove it, rather than augment it.)

course_discovery/apps/api/serializers.py
dianakhuang: I was originally using them in the `PersonSerializer` when I realized that we didn't need it. I didn't make any changes to it.

dianakhuang:
The `order_by` changes are to improve performance. @ormsbee and I were looking at the queries and realized that the default sort on `Person` was slowing things down.

The changes to the instructor search endpoints only impact the `facets` functionality. Nothing about instructors was using facets before (facet-support is partially bundled into some of the common base-classes we use), so I don't think that has much in the way of regression impact.

course_discovery/apps/api/v1/tests/test_views/test_people.py
dianakhuang: Added this test.

PR #1820, opened 2019-03-18
Utility function to create just in time draft world for Courses or Course runs.

Dillon-Dumesnil:
It can take a Course or Course run as input and will create the entire draft world surrounding that object. It handles creation of a draft course, all draft course runs associated with that course, all draft seats associated with all course runs, and all draft entitlements associated with the course.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_utils.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_utils.py
@@ -1,0 +1,161 @@
+import ddt
+from django.test import TestCase
+
+from course_discovery.apps.api.tests.mixins import SiteMixin
+from course_discovery.apps.api.v1.views import utils
+from course_discovery.apps.course_metadata.models import Course, CourseEntitlement, CourseRun, Seat
+from course_discovery.apps.course_metadata.tests.factories import (
+    CourseEntitlementFactory, CourseFactory, CourseRunFactory, SeatFactory
+)
+
+
+@ddt.ddt
+class TestEnsureDraftWorld(SiteMixin, TestCase):
+    def setUp(self):
+        super().setUp()
+
+    @ddt.data(
+        None,
+        {'weeks_to_complete': 7},
+        {'weeks_to_complete': 7, 'title_override': 'New Title'},
+    )
+    def test_set_draft_state(self, attrs):
+        course_run = CourseRunFactory()
+        draft_course_run, original_course_run = utils.set_draft_state(course_run, CourseRun, attrs)
+
+        self.assertEqual(1, len(CourseRun.objects.all()))
+        self.assertEqual(2, len(CourseRun.everything.all()))
+
+        self.assertTrue(draft_course_run.draft)
+        self.assertFalse(original_course_run.draft)
+
+        if attrs:
+            model_fields = [field.name for field in CourseRun._meta.get_fields()]
+            diff_of_fields = [field for field in filter(
+                lambda f: getattr(original_course_run, f, None) != getattr(draft_course_run, f, None),
+                model_fields
+            )]
+            for key, value in attrs.items():
+                # Make sure that any attributes we changed are different in the draft course run from the original
+                self.assertIn(key, diff_of_fields)
+                self.assertEqual(getattr(draft_course_run, key), value)
+
+    def test_set_draft_state_with_foreign_key(self):
+        course = CourseFactory()
+        course_run = CourseRunFactory(course=course)
+        draft_course, original_course = utils.set_draft_state(course, Course)
+        draft_course_run, original_course_run = utils.set_draft_state(course_run, CourseRun, {'course': draft_course})
+
+        self.assertEqual(1, len(CourseRun.objects.all()))
+        self.assertEqual(2, len(CourseRun.everything.all()))
+        self.assertEqual(1, len(Course.objects.all()))
+        self.assertEqual(2, len(Course.everything.all()))
+
+        self.assertTrue(draft_course_run.draft)
+        self.assertFalse(original_course_run.draft)
+
+        self.assertTrue(draft_course.draft)
+        self.assertFalse(original_course.draft)
+
+        self.assertNotEqual(draft_course_run.course, original_course_run.course)
+        self.assertEqual(draft_course_run.course, draft_course)
+        self.assertEqual(original_course_run.course, original_course)
+
+    def test_ensure_draft_world_draft_obj_given(self):
+        course_run = CourseRunFactory(draft=True)
+        ensured_draft_course_run = utils.ensure_draft_world(course_run)
+
+        self.assertEqual(ensured_draft_course_run, course_run)
+        self.assertEqual(ensured_draft_course_run.id, course_run.id)
+        self.assertEqual(ensured_draft_course_run.uuid, course_run.uuid)
+        self.assertEqual(ensured_draft_course_run.draft, course_run.draft)
+
+    def test_ensure_draft_world_not_draft_course_run_given(self):
+        course = CourseFactory()
+        course_run = CourseRunFactory(course=course)
+        verified_seat = SeatFactory(type='verified', course_run=course_run)
+        audit_seat = SeatFactory(type='audit', course_run=course_run)
+        course_run.seats.add(verified_seat, audit_seat)
+
+        ensured_draft_course_run = utils.ensure_draft_world(course_run)
+        not_draft_course_run = CourseRun.objects.get(uuid=course_run.uuid)
+
+        self.assertNotEqual(ensured_draft_course_run, not_draft_course_run)
+        self.assertEqual(ensured_draft_course_run.uuid, not_draft_course_run.uuid)
+        self.assertTrue(ensured_draft_course_run.draft)
+        self.assertNotEqual(ensured_draft_course_run.course, not_draft_course_run.course)
+        self.assertEqual(ensured_draft_course_run.course.uuid, not_draft_course_run.course.uuid)
+
+        # Check slugs are equal
+        self.assertEqual(ensured_draft_course_run.slug, not_draft_course_run.slug)
+
+        # Seat checks
+        draft_seats = ensured_draft_course_run.seats.all()
+        not_draft_seats = Seat.objects.filter(course_run=not_draft_course_run)
+        self.assertNotEqual(draft_seats, not_draft_seats)
+        for i, __ in enumerate(draft_seats):
+            self.assertEqual(draft_seats[i].price, not_draft_seats[i].price)
+            self.assertEqual(draft_seats[i].sku, not_draft_seats[i].sku)
+            self.assertNotEqual(draft_seats[i].course_run, not_draft_seats[i].course_run)
+            self.assertEqual(draft_seats[i].course_run.uuid, not_draft_seats[i].course_run.uuid)
+            self.assertEqual(draft_seats[i].official_version, not_draft_seats[i])
+            self.assertEqual(not_draft_seats[i].draft_version, draft_seats[i])
+
+        # Check draft course is also created
+        not_draft_course = Course.objects.get(uuid=course.uuid)
+        draft_course = ensured_draft_course_run.course
+        self.assertNotEqual(draft_course, not_draft_course)
+        self.assertEqual(draft_course.uuid, not_draft_course.uuid)
+        self.assertTrue(draft_course.draft)
+
+        # Check official and draft versions match up
+        self.assertEqual(ensured_draft_course_run.official_version, not_draft_course_run)
+        self.assertEqual(not_draft_course_run.draft_version, ensured_draft_course_run)
+
+    def test_ensure_draft_world_not_draft_course_given(self):
+        course = CourseFactory()
+        entitlement = CourseEntitlementFactory(course=course)
+        course.entitlements.add(entitlement)
+        course_runs = CourseRunFactory.create_batch(3, course=course)
+        for run in course_runs:
+            course.course_runs.add(run)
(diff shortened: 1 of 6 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1959,3 +1959,26 @@ class ProfileImageDownloadConfig(SingletonModel):
     Configuration for management command to Download Profile Images from Drupal.
     """
     person_uuids = models.TextField(default=None, null=False, blank=False, verbose_name=_('Profile Image UUIDs'))
+
+
+class AddDraftRowConfig(SingletonModel):
```
mikix: We have two different models for how to do this sort of config. The approach taken by `DrupalPublishUuidConfig` or the approach taken by `DeletePersonDupsConfig` -- either is fine, but just wanted to make you aware that there are two.

I think it depends on how user-friendly you want the django admin page to be, how user-friendly you want the console management command to be, how often you expect to make changes in how it could be run (and thus how often a schema might change), and how many options you support.

Subtext to this is that I kind of prefer the `DeletePersonDupsConfig` approach, but I recognize that both are good for different things.

course_discovery/apps/course_metadata/management/commands/add_draft_row.py
```diff
@@ -0,0 +1,48 @@
...
+        elif config.table == AddDraftRowConfig.PERSON:
+            model = Person
+        elif config.table == AddDraftRowConfig.SEAT:
+            model = Seat
+        elif config.table == AddDraftRowConfig.VIDEO:
+            model = Video
+
+        if config.ids:
+            # Create draft rows for only these ids in the given table
+            ids = config.ids.split(',')
```
mikix: Might want to clear whitespace too?

course_discovery/apps/course_metadata/management/commands/add_draft_row.py
```diff
@@ -0,0 +1,48 @@
...
+        elif config.table == AddDraftRowConfig.COURSE_ENTITLEMENT:
+            model = CourseEntitlement
+        elif config.table == AddDraftRowConfig.COURSE_RUN:
+            model = CourseRun
+        elif config.table == AddDraftRowConfig.PERSON:
+            model = Person
+        elif config.table == AddDraftRowConfig.SEAT:
+            model = Seat
+        elif config.table == AddDraftRowConfig.VIDEO:
+            model = Video
```
mikix: Maybe `getattr(models, DRAFT_ROW_TABLE_CHOICES[config.table])`?

(that won't work, because choices is not a dict, but you get the idea)

course_discovery/apps/course_metadata/management/commands/add_draft_row.py
```diff
@@ -0,0 +1,48 @@
...
+            ids = config.ids.split(',')
+        else:
+            # Create draft rows for all rows with draft column set to False
+            ids = model.objects.values_list('id', flat=True)
+
+        need_draft_objects = model.objects.filter(id__in=ids)
+        for obj in need_draft_objects:
+            if obj.draft:
+                logger.warn('{obj} is already a draft object. Not creating a new draft.'.format(obj=obj))
+                continue
```
mikix: In this filter, with my changes to the object manager, you won't find any draft objects with the query you use. (you'd have to query agains `model._base_manager` which is a django thing or however else I expose the full set) But that seems fine.

So either remove this bit or query differently or check for ids that weren't found by your query?

Dillon-Dumesnil: I guess the question would be: do we find it important to notify the user that an id they provided is for a draft row or not? Either way, we will not be creating a new draft row.

mikix: I think it would make sense to error out or provide a warning if the user asked you to do something with an id but you can't because it doesn't make any sense.

course_discovery/apps/course_metadata/management/commands/add_draft_row.py
Dillon-Dumesnil: The filter actually doesn't care. Good ol black magic!

course_discovery/apps/course_metadata/management/commands/add_draft_row.py
Dillon-Dumesnil: Yeah, I'm thinking this scenario might actually be pretty common (not providing a draft id, but providing an id that I can't use). For example, if we need to create a draft Seat, but the associated draft Course Run does not exist. 

I'm afraid this management command is about to get MESSY

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: Switching from management command to utility function.

course_discovery/apps/course_metadata/management/commands/add_draft_row.py
Dillon-Dumesnil: Not a management command anymore so not relevant.

That being said, I did come up with `model = getattr(models, config.table)` and just made the config.table map to the class names

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -519,6 +522,7 @@ class CourseEntitlementFactory(factory.DjangoModelFactory):
     sku = FuzzyText(length=8)
     expires = FuzzyDateTime(datetime.datetime(2014, 1, 1, tzinfo=UTC))
     course = factory.SubFactory(CourseFactory)
+    draft = False
```
mikix: Do you need this? False is the default yah?

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        obj (Model object): The object to create a draft state for. *Must have draft as an attribute.*
+
+    Returns:
+        obj (Model object): The returned object will be the draft version on the input object.
+    """
+    if obj.draft:
+        return obj
+
+    if isinstance(obj, CourseRun):
+        ensure_draft_world(obj.course)
```
mikix, line 56: You need to look up the draft version of obj after this call to return it, right?

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,62 @@
...
+                if not run.draft:
+                    set_draft_state(run, ['course'], [obj])
+                    for seat in run.seats:
+                        set_draft_state(seat, ['course_run'], [run])
+                    if run.canonical_for_course:
+                        obj.canonical_course_run = run
+                        obj.save()
+        if obj.entitlements:
+            entitlement = obj.entitlements[0]
+            set_draft_state(entitlement, ['course'], [obj])
```
mikix: Why not just iterate the entitlements? Would even be less lines:
```
for e in obj.entitlements:
    set_draft_state(e, ['course'], [obj])
```

course_discovery/apps/api/v1/tests/test_views/test_utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@ddt.ddt
+class TestEnsureDraftWorld(SiteMixin, TestCase):
+    def setUp(self):
+        super().setUp()
+        self.course = CourseFactory(partner=self.partner)
+        self.course_run = CourseRunFactory(course__partner=self.partner)
+
+
```
mikix, line 25: TWO SPACES!? PR BLOCKED.

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if isinstance(obj, CourseRun):
+        ensure_draft_world(obj.course)
+
+    elif isinstance(obj, Course):
+        set_draft_state(obj)
+        # Create draft course runs, the corresponding draft seats, and the draft entitlement
+        if obj.course_runs:
+            for run in obj.course_runs:
+                if not run.draft:
```
mikix: I don't think you need to check None there with the `if obj.course_runs`. And the check on draft is making a different set of assumptions than the early exit at the top of the function. Either the course-world is all draft or no draft says the early exit. But this one here is trying to handle a partial world.

course_discovery/apps/course_metadata/tests/factories.py
Dillon-Dumesnil: Apparently not. It wasn't showing up when I was looking at the factory instance so that worried me, but upon more testing, seems like I don't need to explicitly define it. Nice

course_discovery/apps/api/v1/views/utils.py
Dillon-Dumesnil: Good catch!

mikix (approved):
Assuming your manual testing works out, looks good to me

mikix (approved):
Re-approving!

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
+from course_discovery.apps.course_metadata.models import Course, CourseEntitlement, CourseRun, Seat
+
+
+def set_draft_state(obj, attr_keys=None, attr_values=None):
+    """
+    Sets the draft state for an object by giving it a new primary key. Also sets any given
+    attributes (primarily used for setting foreign keys that also point to draft rows). This will
+    make any additional operations on the object to be done to the new draft state object.
+
+    Note: attr_keys and attr_values must be the same length and the key and value indices must match.
```
mikix: Wouldn't life be simpler if it was a dict?

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
...
+        for i, __ in enumerate(attr_keys):
+            setattr(obj, attr_keys[i], attr_values[i])
+    # Will throw an integrity error if the draft row already exists, but this
+    # should be caught as part of a try catch in the API calling ensure_draft_world
+    obj.save()
+
+
+def ensure_draft_world(obj):
+    """
+    Ensures the draft world exists for an object. Assumes that if the given object is already
```
mikix: Might be worth explaining what a "draft world" is? That a course and all its course runs and all the seat/entitlement products are all drafts together.

Also, it looks like this method already assumes no draft=True row exists - that the caller already checked for that... Might want to make that assumption explicit in the method docs.

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Returns:
+        obj (Model object): The returned object will be the draft version on the input object.
+    """
+    if obj.draft:
+        return obj
+
+    if isinstance(obj, CourseRun):
+        ensure_draft_world(obj.course)
+        obj = CourseRun.everything.get(uuid=obj.uuid, draft=True)
```
mikix: Curiously enough, there is no unique_together for uuid & draft. No uniqueness promises made about uuid at all. But there is a unique_together for key & draft, so might be better to use that.

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    elif isinstance(obj, Course):
+        original_course = Course.objects.get(uuid=obj.uuid)
```
mikix: deepcopy had some problems?

This lookup should consider Course's unique_togethers - it should be uuid & partner. Or just switch it to id=obj.id to be safe and futureproof.

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
...
+            original_run.draft_version = run
+            original_run.save()
+
+            for seat in original_run.seats.all():
+                set_draft_state(seat, ['course_run'], [run])
+                original_seat = Seat.objects.get(course_run=original_run, type=seat.type)
+                original_seat.draft_version = seat
+                original_seat.save()
+            if original_course.canonical_course_run:
+                if run.uuid == original_course.canonical_course_run.uuid:
```
mikix: Tiny nit: might be nicer if you had just one `if` here. Maybe save the canonical course run or its uuid above.

```
canonical_uuid = original_course.canonical_course_run and original_course.canonical_course_run.uuid
...
if run.uuid = canonical_uuid:
```

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
...
+            original_run = CourseRun.objects.get(uuid=run.uuid)
+            set_draft_state(run, ['course'], [obj])
+            run.slug = original_run.slug
+            run.save()
+            original_run.draft_version = run
+            original_run.save()
+
+            for seat in original_run.seats.all():
+                set_draft_state(seat, ['course_run'], [run])
+                original_seat = Seat.objects.get(course_run=original_run, type=seat.type)
```
mikix: The unique_together for Seat is more complicated. Have to pass in currency & credit_provider too. But better to just save this before setting draft state and you can use id=seat.id.

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -0,0 +1,82 @@
...
+                original_seat.draft_version = seat
+                original_seat.save()
+            if original_course.canonical_course_run:
+                if run.uuid == original_course.canonical_course_run.uuid:
+                    obj.canonical_course_run = run
+        for entitlement in original_course.entitlements.all():
+            set_draft_state(entitlement, ['course'], [obj])
+            original_entitlement = CourseEntitlement.objects.get(course=original_course)
```
mikix: Same comment here about lookups. Should be course & mode. But better to just do id= before you set draft state.

mikix (approved):
:confetti_ball:

PR #1821, opened 2019-03-19
Add DraftManager to hide draft rows by default

mikix:
This acts as a convenience object manager for models with draft rows, so that only non-draft rows appear by default.

https://openedx.atlassian.net/browse/DISCO-690

### DIFF
--- a/course_discovery/apps/course_metadata/managers.py
+++ b/course_discovery/apps/course_metadata/managers.py
@@ -1,0 +1,11 @@
+from django.db import models
+
+
+class DraftManager(models.Manager):
+    """ Model manager that hides draft rows unless you ask for them. """
+
+    def get_queryset(self):
+        return super().get_queryset().filter(draft=False)
+
+    def with_drafts(self):
+        return super().get_queryset()
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -27,6 +27,7 @@ from taggit_autosuggest.managers import TaggableManager
 from course_discovery.apps.core.models import Currency, Partner
 from course_discovery.apps.course_metadata.choices import CourseRunPacing, CourseRunStatus, ProgramStatus, ReportingType
 from course_discovery.apps.course_metadata.constants import PathwayType
+from course_discovery.apps.course_metadata.managers import DraftManager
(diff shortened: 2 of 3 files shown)


course_discovery/apps/course_metadata/managers.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+
+class DraftManager(models.Manager):
```
Dillon-Dumesnil, line 4: Do we ever have the use case of only_drafts? I'm honestly not sure, but seems like something we might want to support.

course_discovery/apps/course_metadata/models.py
```diff
@@ -41,6 +42,29 @@
...
+    self.objects.with_drafts().
+
+    Remember to add 'draft' to your unique_together clauses.
+
+    Django doesn't allow real model mixins, but since everything has to inherit from models.Model, we shouldn't be
+    stepping on anyone's toes. This is the best advice I could find (at time of writing for Django 1.11).
+
+    .. no_pii:
+    """
+    draft = models.BooleanField(default=False, help_text='Is this a draft version?')
```
Dillon-Dumesnil, line 60: Should this include null=False? I noticed the original definitions of these included that. Might be redundant with the default, but I don't know.

Dillon-Dumesnil (approved):
Looks good. Couple of questions, but could be no-ops.

course_discovery/apps/course_metadata/models.py
mikix: BooleanField cannot be null by definition. It will throw an exception if you try to set it. You would use NullBooleanField to have a tri-state. So I just dropped the unnecessary arg.

course_discovery/apps/course_metadata/managers.py
mikix: You can do `with_drafts().filter(draft=True)`. Do you think it's common enough I should have syntactic sugar?

mikix: Easy enough to add.

Dillon-Dumesnil: I'm not sure it's common. AJ thinks we could just implement it once needed. I think I agree with that. No reason to over-engineer now and have a never used function lying around.

PR #1822, opened 2019-03-19
Add curriculum details to program endpoint

zacharis278:
### [EDUCATOR-4163](https://openedx.atlassian.net/browse/EDUCATOR-4163)
This will allow master's degree programs setup under the new curriculum structure to return detailed course information.
### [EDUCATOR-4214](https://openedx.atlassian.net/browse/EDUCATOR-4214)
Expose curriculum uuid

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -28,10 +28,10 @@ from course_discovery.apps.core.api_client.lms import LMSAPIClient
 from course_discovery.apps.course_metadata import search_indexes
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
-    FAQ, AdditionalPromoArea, CorporateEndorsement, Course, CourseEntitlement, CourseRun, Curriculum, Degree,
-    DegreeCost, DegreeDeadline, Endorsement, IconTextPairing, Image, LevelType, Organization, Pathway, Person,
-    PersonAreaOfExpertise, PersonSocialNetwork, Position, Prerequisite, Program, ProgramType, Ranking, Seat, SeatType,
-    Subject, Topic, Video
+    FAQ, AdditionalPromoArea, CorporateEndorsement, Course, CourseEntitlement, CourseRun, Curriculum,
+    CurriculumCourseMembership, CurriculumProgramMembership, Degree, DegreeCost, DegreeDeadline, Endorsement,
+    IconTextPairing, Image, LevelType, Organization, Pathway, Person, PersonAreaOfExpertise, PersonSocialNetwork,
+    Position, Prerequisite, Program, ProgramType, Ranking, Seat, SeatType, Subject, Topic, Video
 )
 from course_discovery.apps.course_metadata.utils import parse_course_key_fragment
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1077,9 +1077,73 @@ class Meta:
...
+                'published_course_runs_only': self.context.get('published_course_runs_only'),
+                'exclude_utm': self.context.get('exclude_utm'),
+                'course_runs': self.prefetched_course_runs(curriculum),
+                'use_full_course_serializer': self.context.get('use_full_course_serializer', False),
+            }
+        )
+
+        return course_serializer.data
+
+    def get_programs(self, curriculum):
```
iloveagent57, line 1103: Should we do the same null check here that we do for `get_courses()` above?

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -98,6 +115,37 @@ def test_retrieve(self, django_assert_num_queries):
...
+        CurriculumProgramMembershipFactory(
+            program=child_program1,
+            curriculum=curriculum
+        )
+        CurriculumProgramMembershipFactory(
+            program=child_program2,
+            curriculum=curriculum
+        )
+
+        with django_assert_num_queries(FuzzyInt(105, 2)):
```
iloveagent57: It would be good to leave a comment with:
* What the number of queries is without any curricula
* How many queries we expect to have for each course in the curriculum
* How many queries we expect to have for each program in the curriculum

That way, our future selves won't look at "105" and kind of stare blankly into the void, questioning all existence.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -98,6 +115,37 @@ def test_retrieve(self, django_assert_num_queries):
         response = self.assert_retrieve_success(program, querystring={'use_full_course_serializer': 1})
         assert response.data == self.serialize_program(program, extra_context={'use_full_course_serializer': 1})
 
+    def test_retrieve_basic_curriculum(self, django_assert_num_queries):
+        program = self.create_program()
+        self.create_curriculum(program)
+
+        with django_assert_num_queries(FuzzyInt(71, 2)):
```
iloveagent57: Similarly, why 71 (-ish) queries?

iloveagent57 (commented):
Looks good, just a couple questions.

course_discovery/apps/api/serializers.py
```diff
@@ -1077,9 +1077,73 @@ class Meta:
...
+                'exclude_utm': self.context.get('exclude_utm'),
+                'use_full_course_serializer': self.context.get('use_full_course_serializer', False),
+            }
+        )
+
+        return program_serializer.data
+
+    def prefetch_memberships(self, curriculum):
+        """
+        Prefetch all member courses and related objects for this curriculm
```
schenedx: NIT `s/curriculm/curriculum`

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -98,6 +115,37 @@ def test_retrieve(self, django_assert_num_queries):
...
+            response = self.assert_retrieve_success(program)
+        # property does not have the right values while being indexed
+        del program._course_run_weeks_to_complete
```
schenedx: Why this value in programs is not correct?

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -98,6 +115,37 @@ def test_retrieve(self, django_assert_num_queries):
...
+        )
+        CurriculumProgramMembershipFactory(
+            program=child_program2,
+            curriculum=curriculum
+        )
+
+        with django_assert_num_queries(FuzzyInt(105, 2)):
+            response = self.assert_retrieve_success(parent_program)
+        # property does not have the right values while being indexed
+        del parent_program._course_run_weeks_to_complete
```
schenedx: Again, why this attribute have incorrect values?

schenedx (commented):
Overall, this is good work. I got some questions.

course_discovery/apps/api/serializers.py
zacharis278: I didn't notice any improvement when doing that so I just left it off.  I actually just tested this with get_courses and it actually doesn't appear to ever hit that case so I should probably just nix it entirely.

iloveagent57: That probably makes sense.  I imagine DRF doesn't steer into the serializer for a model that doesn't exist.

kdmccormick (commented):
My only concerns:
* The query counts are surprisingly high -- like Alex said, a comment explaining them would be good 
* Once this merges it'd be good to prioritize [EDUCATOR-4203](https://openedx.atlassian.net/browse/EDUCATOR-4203), which would prevent cyclic program structures

@Zacharis278 thanks for addressing the first bullet point. @schenedx do you agree with aiming to get 4203 done next sprint?

course_discovery/apps/api/v1/tests/test_views/test_programs.py
zacharis278: Honestly I'm not entirely sure.  This is repeated anywhere we assert against a serialized program so I just assumed it to be the source of some flakiness

schenedx: @Zacharis278 You wanna test out what happens if you remove this line and see how your unit test fails? This way, you can write better comments than other tests here?

schenedx (approved):
👍

zacharis278:
I've put in more prefetching on a curriculums child programs to knock down the query count a bit more.  This case is still by no means perfect since we still end up with similar queries on program__courses/curricula/etc with different parent nodes.

In addition to documenting a bit on where the query counts are coming from I altered the tests to be a bit more realistic.  A test program with a curriculum is no longer also populated with courses which was making the results a bit more difficult to understand.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
zacharis278: It seems to have no effect locally, I could run it through travis and see what happens tho

kdmccormick:
@Zacharis278 @schenedx Removing `del program._course_run_weeks_to_complete` on all tests in that class doesn't cause any failures for me locally.

zacharis278:
@schenedx I ran the tests in travis a few times without that `# property does not have the right values..` code block and it never failed so I've just removed it

kdmccormick (approved):
LGTM :+1:

PR #1829, opened 2019-03-27
Have editable=1 only deliver drafts

mikix:
For the course endpoint, move editable=1 handling into the endpoint itself and have it surface draft versions, if available.

https://openedx.atlassian.net/browse/DISCO-715

There's a bit of test-noise because I moved tests from the filter file to the model file.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -333,6 +333,26 @@ class CourseViewSetTests(SerializationMixin, APITestCase):
             # We'll probably fail because we didn't include the right data - but at least we'll have gotten in
             self.assertNotEqual(response.status_code, 403)
 
+    def test_editable_list_gives_drafts(self):
+        draft = CourseFactory(partner=self.partner, uuid=self.course.uuid, key=self.course.key, draft=True)
+        extra = CourseFactory(partner=self.partner, key=self.course.key + 'Z')  # set key so it sorts later
+
+        response = self.client.get(reverse('api:v1:course-list') + '?editable=1')
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data['results'], self.serialize_course([draft, extra], many=True))
+
+    def test_editable_get_gives_drafts(self):
+        draft = CourseFactory(partner=self.partner, uuid=self.course.uuid, key=self.course.key, draft=True)
+        extra = CourseFactory(partner=self.partner)
+
+        response = self.client.get(reverse('api:v1:course-detail', kwargs={'key': self.course.uuid}) + '?editable=1')
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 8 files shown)


Dillon-Dumesnil (approved):
LGTM

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -333,6 +333,26 @@ def test_editor_access_detail_endpoint(self, method, is_staff, in_org, is_editor
             # We'll probably fail because we didn't include the right data - but at least we'll have gotten in
             self.assertNotEqual(response.status_code, 403)
 
+    def test_editable_list_gives_drafts(self):
+        draft = CourseFactory(partner=self.partner, uuid=self.course.uuid, key=self.course.key, draft=True)
+        extra = CourseFactory(partner=self.partner, key=self.course.key + 'Z')  # set key so it sorts later
```
jlajoie, line 338: 👍 nice comment because I was all sorts of confused

mikix: Heh. I probably could have found a more elegant solution, but :man_shrugging:

course_discovery/apps/course_metadata/managers.py
```diff
@@ -9,3 +9,52 @@ def get_queryset(self):
...
+        return list(answers.values())
+
+    def filter_drafts(self, **kwargs):
+        """
+        Acts like filter(), but prefers draft versions.
+        If a draft is not available, we give back the non-draft version.
+        """
+        drafts = self.find_drafts(**kwargs)
+        ids = (x.id for x in drafts)
+        return self.with_drafts().filter(id__in=ids)
```
jlajoie, line 47: Just a comment don't need to change this, looks like this will run a second query, I wonder if there's a better way for us to say "give me this row if draft exists, else give me the other". Like if this is better to do in memory rather than a second query out to the DB.

jlajoie (approved):
Just the one comment, but I think this is still good to merge as is 👍

course_discovery/apps/course_metadata/managers.py
mikix: Yeah so I struggled with this - remember when you left and you commented on me doing some deep thinking?

So I want a way to return a queryset - that's how the rest of the DRF scaffolding expects to be able to handle a set of rows. But I could not come up with a query that would give us the drafts-but-only-sometimes without being able to iterate and inspect the rows.

We could do it if we added a new column like draft_id that connected draft and non-draft rows. Then it would be something like `draft=True | draft_id=None` - but then we have to keep that column in sync. But could be worth the performance... Thoughts?

jlajoie: Hmm that might be the better approach, I guess that or we handle this all in memory and don't treat it as a queryset?  If DRF makes things easier then we should probably just stick with the queryset approach.

PR #1831, opened 2019-04-01
Add an optional query parameter for serializing person autocomplete endpoint

dianakhuang:
To support https://openedx.atlassian.net/browse/DISCO-681

Also add more testing around org limiting filters.

### DIFF
--- a/course_discovery/apps/course_metadata/lookups.py
+++ b/course_discovery/apps/course_metadata/lookups.py
@@ -5,6 +5,7 @@ from django.contrib.auth.mixins import LoginRequiredMixin
 from django.db.models import Q
 from django.template.loader import render_to_string
 
+from course_discovery.apps.api.serializers import PersonSerializer
 from .models import Course, CourseRun, Organization, Person, Program
 
 
@@ -95,8 +96,12 @@ class PersonAutocomplete(LoginRequiredMixin, autocomplete.Select2QuerySetView):
 
     def get_result_label(self, result):
         http_referer = self.request.META.get('HTTP_REFERER')
+        serialize = self.request.GET.get('serialize')
         if http_referer and '/admin/' in http_referer:
             return super(PersonAutocomplete, self).get_result_label(result)
+        elif serialize:
+            context = {'request': self.request}
+            return PersonSerializer(result, context=context).data
         else:
             context = {
                 'uuid': result.uuid,
--- a/course_discovery/apps/course_metadata/tests/test_lookups.py
+++ b/course_discovery/apps/course_metadata/tests/test_lookups.py
@@ -130,15 +130,14 @@ class AutoCompletePersonTests(SiteMixin, TestCase):
             self.organization_extensions.append(factories.OrganizationExtensionFactory(organization=organization))
 
         disco_course = CourseFactory(authoring_organizations=[self.organizations[0]])
+        disco_course2 = CourseFactory(authoring_organizations=[self.organizations[1]])
         CourseRunFactory(course=disco_course, staff=[first_instructor])
+        CourseRunFactory(course=disco_course2, staff=[second_instructor])
 
         self.user.groups.add(self.organization_extensions[0].group)
 
-    def query(self, q, org=None):
+    def query(self, q):
         query_params = '?q={q}'.format(q=q)
-
-        if org:
-            query_params = query_params + '&org={org}'.format(org=org)
 
         return self.client.get(
             reverse('admin_metadata:person-autocomplete') + query_params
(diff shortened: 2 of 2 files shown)


course_discovery/apps/course_metadata/lookups.py
```diff
@@ -5,6 +5,7 @@
 from django.db.models import Q
 from django.template.loader import render_to_string
 
+from ..api.serializers import PersonSerializer
```
mikedikan: Sorry, this may be a dumb question, but why two dots? (..api)

mikedikan (approved):

course_discovery/apps/course_metadata/lookups.py
dianakhuang: The serializers are defined in a different djangoapp. That was one of the reasons why I was worried about bringing it in.

Dillon-Dumesnil: Nit: Switch these to use the full path. Most files I've seen in course_discovery use the full path

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -95,8 +96,12 @@ def get_queryset(self):
 
     def get_result_label(self, result):
         http_referer = self.request.META.get('HTTP_REFERER')
-        if http_referer and '/admin/' in http_referer:
+        serialize = self.request.GET.get('serialize')
+        if (http_referer and '/admin/' in http_referer):
             return super(PersonAutocomplete, self).get_result_label(result)
+        if serialize:
```
Dillon-Dumesnil: Nit: Switch to `elif` since it seems like we want to explicitly only allow these three conditions.

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -130,16 +130,15 @@ def setUp(self):
             self.organization_extensions.append(factories.OrganizationExtensionFactory(organization=organization))
 
         disco_course = CourseFactory(authoring_organizations=[self.organizations[0]])
+        disco_course2 = CourseFactory(authoring_organizations=[self.organizations[1]])
         CourseRunFactory(course=disco_course, staff=[first_instructor])
+        CourseRunFactory(course=disco_course2, staff=[second_instructor])
 
         self.user.groups.add(self.organization_extensions[0].group)
 
-    def query(self, q, org=None):
```
Dillon-Dumesnil, line 137: I'm not sure I understand the reason for this change.

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -252,11 +251,30 @@ def test_instructor_autocomplete_from_django_admin(self):
         expected_results = [{'id': instructor.id, 'text': str(instructor)} for instructor in self.instructors]
         assert data.get('results') == expected_results
 
+    def test_instructor_autocomplete_without_rendering(self):
+        person_autocomplete_url = reverse(
+            'admin_metadata:person-autocomplete'
+        ) + '?q={q}&serialize=1'.format(q=self.instructors[0].uuid)
+
+        response = self.client.get(person_autocomplete_url)
```
Dillon-Dumesnil, line 259: I think this can be rewritten using the query helper function.
`response = self.query(self.instructors[0].uuid + '&serialize=1')`

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -252,11 +251,30 @@ def test_instructor_autocomplete_from_django_admin(self):
...
+        assert data['results'][0]['text']['bio'] == self.instructors[0].bio
+
     def test_autocomplete_limit_by_org(self):
         org = self.organizations[0]
-        response = self.query('ins', org=org.key)
+        person_autocomplete_url = reverse(
+            'admin_metadata:person-autocomplete'
+        ) + '?q=ins'
+        single_autocomplete_url = person_autocomplete_url + '&org={key}'.format(key=org.key)
+        response = self.client.get(single_autocomplete_url)
```
Dillon-Dumesnil, line 270: Similar to above, I would advocate to switching these two responses to
`response = self.query('ins&org={key}'.format(key=org.key))`
and then
`response = self.query('ins&org={key}&org={key_2}'.format(key=org.key, key_2=org2.key))`

course_discovery/apps/course_metadata/tests/test_lookups.py
Dillon-Dumesnil: Upon further investigation and looking at the below code changes, I think a preferred change for this would be to have `self.query` just take in url parameters instead of only the query arg. This way it would look less awkward if you want to add in any additional arguments like orgs.

Example:
`self.query('q=blah&org=key')` instead of `self.query('blah&org=key')`.

Dillon-Dumesnil (approved):
A couple comments, but I don't think any of them are blockers, just things for consideration.

PR #1834, opened 2019-04-02
DISCO-777: Allows nullable for fields on course runs

jlajoie:


course_discovery/apps/api/serializers.py
```diff
@@ -614,7 +614,7 @@ class MinimalCourseRunSerializer(DynamicFieldsMixin, TimestampModelSerializer):
     seats = SeatSerializer(required=False, many=True)
     key = serializers.CharField(required=False)
     title = serializers.CharField(required=False)
-    short_description = serializers.CharField(required=False)
+    short_description = serializers.CharField(required=False, allow_null=True)
```
mikix: I think it would be slightly more correct to pass allow_blank=True here and for the other text fields (which is also accepted by DRF).

mikix (approved):

PR #1835, opened 2019-04-03
Add new status and put in place new workflow restrictions on API endpoints

dianakhuang:
https://openedx.atlassian.net/browse/DISCO-755


course_discovery/apps/course_metadata/choices.py
```diff
@@ -5,6 +5,9 @@
 class CourseRunStatus(DjangoChoices):
     Published = ChoiceItem('published', _('Published'))
     Unpublished = ChoiceItem('unpublished', _('Unpublished'))
+    Reviewed = ChoiceItem('reviewed', _('Reviewed'))
+    AwaitingLegalReview = ChoiceItem('review_by_legal', _('Awaiting Review from Legal'))
+    AwaitingEdServicesReview = ChoiceItem('review_by_ed_services', _('Awaiting Review from Educational Services'))
```
dianakhuang: @staubina Could I get a review of these new status names? I took a stab at trying to make something reasonable, but feedback is welcome.

mikix: Is EdServices too edx-the-company specific?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -185,9 +186,23 @@ def create(self, request, *args, **kwargs):
     @writable_request_wrapper
     def update(self, request, *args, **kwargs):
         """ Update one, or more, fields for a course run. """
-        response = super().update(request, *args, **kwargs)
 
         course_run = self.get_object()
+
+        patch_put = request.method in ['PATCH', 'PUT']
```
mikix: Maybe `not in SAFE_METHODS`? 

(Found like `from rest_framework.permissions import SAFE_METHODS`)

Athough... how do you get into update() without doing a patch or a put? Can we just assume patch_put is True?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -185,9 +186,23 @@ def create(self, request, *args, **kwargs):
...
+
+        # Disallow patch or put if the course run is in review.
+        if patch_put and course_run.in_review:
+            return Response(status=status.HTTP_403_FORBIDDEN)
```
mikix: Might be kind to provide a message like "Course is in review, please wait" or something.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -185,9 +186,23 @@ def create(self, request, *args, **kwargs):
...
+
+        response = super().update(request, *args, **kwargs)
+
+        # If changes are made after review and before publish,
+        # revert status to unpublished.
+        if patch_put and course_run.status == CourseRunStatus.Reviewed:
+            course_run.status = CourseRunStatus.Unpublished
+            course_run.save()
```
mikix: A minor performance point, but if you checked status beforehand, you could add the status downgrade to request.data before calling super().update() and avoid a second save.

Does it make sense for this logic to be inside the CourseRun.save() method? I think... maybe not? We want to let admins poke and prod without resetting state. OK, convinced myself.

course_discovery/apps/course_metadata/models.py
```diff
@@ -893,6 +893,10 @@ def full_description(self, value):
     def outcome(self):
         return self.outcome_override or self.course.outcome
 
+    @property
+    def in_review(self):
+        return self.status in [CourseRunStatus.AwaitingLegalReview, CourseRunStatus.AwaitingEdServicesReview]
```
mikix: Might be nice if we could add this "set of review states" to the CourseRunStatus class - so it's more obvious that it should be updated if a new review status is added.

Like a list that says REVIEW_STATES = [....] on the Status class and then this class property just references that.

course_discovery/apps/course_metadata/choices.py
dianakhuang: Do we need two review states? For simple OFAC, no, but in the future?

mikix: So far, the thinking is yeah, we'll want to be able to distinguish between the two reviews, to point legal or the PCs at their respective queues and email them when it's their turn.

mikix: (Even in simple OFAC, we still have both teams involved in the review, it's just handed off out of band.)

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -185,9 +186,22 @@ def create(self, request, *args, **kwargs):
...
+        if course_run.in_review:
+            return Response(
+                _('Course run is in review. Editing disabled.'),
+                status=status.HTTP_403_FORBIDDEN
+            )
+
+        # If changes are made after review and before publish,
+        # revert status to unpublished.
+        if course_run.status == CourseRunStatus.Reviewed:
+            request.data['status'] = 'unpublished'
```
mikix: `CourseRunStatus.Unpublished.value`?

mikix (approved):
Looks good, thank you! Assuming that someone helps give us better names.

course_discovery/apps/course_metadata/choices.py
staubina: I am going to second Mike that I think EdServices is probably too much of an internal name.

Could it just be "internal review"?

Also, do we need the Awaiting portion of the name?

course_discovery/apps/course_metadata/choices.py
```diff
@@ -5,6 +5,13 @@
 class CourseRunStatus(DjangoChoices):
     Published = ChoiceItem('published', _('Published'))
     Unpublished = ChoiceItem('unpublished', _('Unpublished'))
+    Reviewed = ChoiceItem('reviewed', _('Reviewed'))
+    LegalReview = ChoiceItem('review_by_legal', _('Awaiting Review from Legal'))
+    InternalReview = ChoiceItem('review_by_internal', _('Awaiting Internal Review'))
```
dianakhuang, line 10: Names have been updated. Let me know what you think, @staubina .

PR #1837, opened 2019-04-03
Delete the file where migrations are already squashed

schenedx:
This is to clean up the migration file list in the **Core** app because migrations are squashed already from 1 to 11. This is also following the instructions on https://docs.djangoproject.com/en/2.1/topics/migrations/#squashing-migrations
@edx/masters-neem Please review


reillz10 (approved):
This will make things far less confusing in the future. I know that I stumbled on this the last time I tried to add migrations to core. 👍

kdmccormick (approved):
LGTM :+1:

PR #1838, opened 2019-04-03
Updating course run endpoint for draft models

Dillon-Dumesnil:


course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -33,7 +33,7 @@ def get(self, request):
                 course_run_ids = course_run_ids.split(',')
                 specified_course_ids = course_run_ids
                 identified_course_ids.update(CourseRun.search(query).filter(
-                    partner=partner.short_code, key__in=course_run_ids).values_list('key', flat=True))
+                    course__partner=partner, key__in=course_run_ids).values_list('key', flat=True))
```
mikix, line 36: How did this work before?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -70,14 +71,19 @@ def get_queryset(self):
...
         if q:
-            qs = SearchQuerySetWrapper(CourseRun.search(q).filter(partner=partner.short_code))
-            # This is necessary to avoid issues with the filter backend.
-            qs.model = self.queryset.model
-            return qs
+            queryset = CourseRun.search(q, queryset=queryset).filter(course__partner=partner)
         else:
-            queryset = super(CourseRunViewSet, self).get_queryset().filter(course__partner=partner)
-            return self.get_serializer_class().prefetch_queryset(queryset=queryset)
+            queryset = queryset.filter(course__partner=partner)
```
mikix: You're filtering in both branches on partner. Maybe could pull that last line out of the `else:` and drop it from the `if q:` branch.

mikix (approved):
Love it. Don't understand all the SearchQuery stuff that got dropped. But trust that it works.

course_discovery/apps/api/v1/views/catalog_queries.py
Dillon-Dumesnil: Because search used to return a Haystack SearchQuerySet and now it returns a Django queryset. I have no idea what the difference is or how it was able to work before, but that's the change that happened.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -103,6 +103,9 @@ def get_queryset(self):
         q = self.request.query_params.get('q')
         edit_mode = get_query_param(self.request, 'editable') or self.request.method not in SAFE_METHODS
 
+        if edit_mode and q:
+            raise Exception(_('Specifying both editable=1 and a q parameter is not supported.'))
```
mikix: When duplicating a string like this, it's easy to get string drift if someone changes one but not the other. Then translators are translating two slightly different strings, which kinda sucks, effort wise.

Maybe throw a tiny exception class in the djangoapp, like:
```
class EditableAndQUnsupported(Exception):
    def __init__(self):
        super().__init__(_('my string'))
```

That you can then reference both places? That would also give a single place to add a comment about why it's unsupported - just that we saw odd results and rather than investigate, we just disallowed, since we didn't have a use case yet. And that future archeologists are welcome to investigate and fix the root issue and remove the exception if they are feeling energetic.

Otherwise, folks in a couple years are going to have no idea why we disallow this.

PR #1839, opened 2019-04-04
fix 500 error when course related to a course run is changed manually

DawoudSheraz:
### Description
Whenever a course run is published from Publisher, its equivalent is created in the course metadata(and other locations). The course metadata, along with other places, is supposed to be in sync with the publisher. It is possible to cause the data inconsistency in the course metadata from the discovery Django admin. One such case is the changing of the parent course of a course run in the metadata. This inconsistency leads to the inability of the user to publish the course run again as the unexpected behavior is not what is to be expected. Course teams do come across a situation where they have to change the parent course for a course run in Discovery.

This PR creates a patch to allow for publish to work even after causing the data inconsistency. This PR assumes that Discovery course has precedence over publisher course and if a discover course run already has an associated course, use it instead of publisher course run's course

### DIFF
--- a/course_discovery/apps/publisher/api/v1/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/v1/tests/test_views.py
@@ -17,7 +17,7 @@ from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.models import CourseEntitlement as DiscoveryCourseEntitlement
 from course_discovery.apps.course_metadata.models import Seat as DiscoverySeat
 from course_discovery.apps.course_metadata.models import CourseRun, SeatType, Video
-from course_discovery.apps.course_metadata.tests.factories import OrganizationFactory, PersonFactory
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, OrganizationFactory, PersonFactory
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
 from course_discovery.apps.publisher.api.utils import (
     serialize_entitlement_for_ecommerce_api, serialize_seat_for_ecommerce_api
@@ -224,6 +224,63 @@ class CourseRunViewSetTests(APITestCase):
             price=verified_seat.price,
             **common_seat_kwargs
         )
+
+    @responses.activate
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    def test_publish_after_discovery_courserun_course_change(self, mock_access_token):  # pylint: disable=unused-argument,too-many-statements
+        """
+        Verify publish works fine when Parent course of a discovery course run
+        is changed manually.
+
+        Scenario: Given that a course run is created in Publisher
+        After pressing Publish, it is published to Discovery
+        Change the newly created Discovery course run's parent course
+        Publish the discovery course run again
+        The publish should work without any error
+        and course run's parent course will not change
+        """
+        publisher_course_run = self._create_course_run_for_publication()
+        partner = publisher_course_run.course.organizations.first().partner
+        self._set_test_client_domain_and_login(partner)
+        self._mock_studio_api_success(publisher_course_run)
+        self._mock_ecommerce_api(publisher_course_run)
+
+        with LogCapture(LOGGER_NAME) as log:
+            url = reverse('publisher:api:v1:course_run-publish', kwargs={'pk': publisher_course_run.pk})
+            response = self.client.post(url, {})
+            assert response.status_code == 200
+            log.check((LOGGER_NAME, 'INFO',
+                       'Published course run with id: [{}] lms_course_id: [{}], user: [{}], date: [{}]'.format(
+                           publisher_course_run.id, publisher_course_run.lms_course_id, self.user, date.today())))
+
+        expected = {
+            'discovery': CourseRunViewSet.PUBLICATION_SUCCESS_STATUS,
+            'ecommerce': CourseRunViewSet.PUBLICATION_SUCCESS_STATUS,
+            'studio': CourseRunViewSet.PUBLICATION_SUCCESS_STATUS,
+        }
+        assert response.data == expected
+
+        discovery_course_run = CourseRun.objects.get(key=publisher_course_run.lms_course_id)
+        course_run_prev_course = discovery_course_run.course
+
+        new_discovery_course = CourseFactory(title="My Test Course")
+        discovery_course_run.course = new_discovery_course
+        discovery_course_run.save()
+
+        # Verify new course has been associated to the course run
+        assert discovery_course_run.course.title == "My Test Course"
+        assert discovery_course_run.course != course_run_prev_course
+
+        # Publish the course run
+        url = reverse('publisher:api:v1:course_run-publish', kwargs={'pk': publisher_course_run.pk})
+        response = self.client.post(url, {})
+        assert response.status_code == 200
+        assert response.data == expected
+
+        # Verify that Discovery Course Run is still associated with the manually associated course
+        discovery_course_run = CourseRun.objects.get(key=publisher_course_run.lms_course_id)
+        assert discovery_course_run.course.title == new_discovery_course.title
+        assert discovery_course_run.course.uuid == new_discovery_course.uuid
 
     # pylint: disable=unused-argument
     @responses.activate
--- a/course_discovery/apps/publisher/api/v1/views.py
+++ b/course_discovery/apps/publisher/api/v1/views.py
@@ -154,6 +154,13 @@ class CourseRunViewSet(viewsets.GenericViewSet):
             'weeks_to_complete': course_run.length,
             'has_ofac_restrictions': course_run.has_ofac_restrictions
         }
+
+        # Give priority to Discovery course run's parent course rather than
+        # Publisher course run's parent course.
+        discovery_course_run = course_run.discovery_course_run
+        if discovery_course_run and course_run.discovery_course:
+            discovery_course = course_run.discovery_course
+
         discovery_course_run, __ = DiscoveryCourseRun.objects.update_or_create(
             course=discovery_course,
             key=course_run.lms_course_id,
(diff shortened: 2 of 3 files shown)


DawoudSheraz:
This fix reverts the course change done in the course metadata as all the course creation in the metadata is done using the publisher course run and the changes are not 2-way i.e. only publisher changes are supposed to reflect in both publisher & Metadata.

awaisdar001:
Are going to add a test?

staubina:
I am going to review this.  I have also asked @mikix to take a look.

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -155,8 +155,13 @@ def publish_to_discovery(self, partner, course_run):
             'max_effort': course_run.max_effort,
             'language': course_run.language,
             'weeks_to_complete': course_run.length,
-            'has_ofac_restrictions': course_run.has_ofac_restrictions
+            'has_ofac_restrictions': course_run.has_ofac_restrictions,
         }
+
+        discovery_course_run = DiscoveryCourseRun.objects.filter(key=course_run.lms_course_id).first()
```
staubina: This seems like a sensible approach to me.

This should no longer be an issue as we move to the new version of Publisher we are working on that will not rely on the Publisher data model.

staubina (commented):
Please wait on one more review from someone on the Publisher Pod before merging.  Good idea with this small change.

course_discovery/apps/publisher/api/v1/views.py
mikix: You can rewrite this as `discovery_course_run = course_run.discovery_course_run` - there's a dynamic property that does that query for you, with an early exit for None.

Actually, it may be worth adding a new dynamic property called `discovery_course` on the Publisher.CourseRun model, that would do this whole lookup you're doing - to encourage others to avoid this bug? Maybe even with some nice comments about why the property exists.

(And regardless if you do above - some nice comments here about why you're doing these lines might be nice - would be confusing if you don't have context.)

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -155,8 +155,13 @@ def publish_to_discovery(self, partner, course_run):
...
+        if discovery_course_run and discovery_course_run.course:
+            discovery_course = discovery_course_run.course
```
mikix: So you're overriding the `discovery_course` object from earlier in this method (where one was updated_or_created). Why not do this check earlier and update the existing CourseMetadata.Course object if it exists? Rather than update the wrong one, then finally grabbing the right one here?

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -225,6 +225,64 @@ def test_publish(self, mock_access_token):  # pylint: disable=unused-argument,to
             **common_seat_kwargs
         )
 
+    @responses.activate
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
```
mikix: We are actively planning to remove this switch. It would make our future lives easier if we didn't add more tests that rely on it. Could maybe something else be mocked to avoid whatever you're trying to avoid here?

course_discovery/apps/publisher/api/v1/views.py
DawoudSheraz: I don't know about the dynamic property, that's why I did it like this. I will take a look at it. and I will add a comment later. It was just a quick PR update to see if the fix would work.

course_discovery/apps/publisher/api/v1/views.py
DawoudSheraz: That is a twilight zone as if course team changed the course, that doesn't necessarily mean that the previous course was entirely wrong. I have recently investigated the case where the course team changed course because there were enrollments in the old course and new course didn't have any. If I were to do so as you are trying to ask, it might cause more data inconsistencies. But I can try investigating and analyze the results.

course_discovery/apps/publisher/api/v1/tests/test_views.py
DawoudSheraz: Sure.

course_discovery/apps/publisher/api/v1/views.py
DawoudSheraz: I looked at the mentioned property, and it is doing the same as I am doing at this location. Thanks for pointing that out. and I will try adding `discovery_course` property as it does look worth an effort.

DawoudSheraz:
@staubina @mikix There are some questions that I would like you guys to help in answering. First, the place where I have introduced the fix, after that place, the `discovery_course` is used for two purposes:
1. Course [Entitlements](https://github.com/edx/course-discovery/blob/b6a535826ac928a740e2e9f0dbb0d34c0834c7c7/course_discovery/apps/publisher/api/v1/views.py#L175)
2. If created is true, then [associate](https://github.com/edx/course-discovery/blob/b6a535826ac928a740e2e9f0dbb0d34c0834c7c7/course_discovery/apps/publisher/api/v1/views.py#L205) the newly created `discovery_course_run` with `discovery_course`.

Since we have changed the parent course ourselves, are the entitlements supposed to be created with the changed course or with publisher course? Also, Is there any effect when the `discovery_course_run` is made canoncial_course_run of `discovery_course`?

I would also like to address again the question that @mikix asked:

> So you're overriding the discovery_course object from earlier in this method (where one was updated_or_created). Why not do this check earlier and update the existing CourseMetadata.Course object if it exists? Rather than update the wrong one, then finally grabbing the right one here?

The problem is that If we try to make the `discovery_course` consistent across the function, it causes more problems. I will explain it with an example. Say that for discovery run, we originally had the course 001. We changed it to 002 manually. In Publisher, it is still 001. In terms of the data, 001 has subjects s1,s2 and s3 and 002 has None. If I make `discovery_course` consistent by moving it to top of the function and avoid [this](https://github.com/edx/course-discovery/blob/b6a535826ac928a740e2e9f0dbb0d34c0834c7c7/course_discovery/apps/publisher/api/v1/views.py#L133) call, after publish, now 002 will also contain s1, s2 & s3. I am sure this is not something we would want to do as the data inconsistency will grow tremendously. If we want to create the Entitlements with the changed course, then my fix will not cause any issues. But, yes, the `discovery_course` object will be inconsistent in first half of the function, and I can't think of any other way at the moment to avoid that situation.

mikix:
I see what you're saying about the inconsistent courses. I hope/assume that people in this situation will know the score and will not expect edits to the wrong course info in publisher to show up anywhere useful.

And I would say that if we take that stance - that course-level edits on this abandoned course are ignored - we don't have to worry about setting entitlement info either, since that's a course-level thing. (So continue to set entitlements before looking up new discovery_course.) Does that seem correct?

staubina:
The Entitlement should be created with the Course that the Course Run is pointing to in Discovery Course metadata.  This is important because when the entitlement is used it will use the Course UUID to get the available Course Runs.

The Course UUID is also sent to ecommerce.

DawoudSheraz:
@staubina @mikix I discussed this problem within the team and we have reached the conclusion that the fix will cause problems for E-commerce. When publishing to e-commerce, the Discovery course is being fetched using partner and course_key([reference](https://github.com/edx/course-discovery/blob/6f9e323a7a3d7e5a36c8123f693947162a4ffab7/course_discovery/apps/publisher/api/v1/views.py#L87)). The course_key is being made from the publisher course, and when the Discovery course is queried, it returns the one that corresponds to publisher course. I tested this locally, and here is how the flow went:

#### Note: I initally had to disable publish_to_studio and publish_to_ecommerce as I haven't set them locally
1. I had a course_run cr_1 in publisher, that corresponded to c_1 in publisher.
2. I published the course and same were created in Metadata.
3. I changed the metadata course run's parent course to c_2.
4. Press publish. The publish happened.
5. I put a breakpoint in publish_to_ecommerce and pressed publish. In ecommerce, it would make the key with publisher course, so it would be like key_c_1. When it tries to get the Discovery course for that key, it is able to find and returns the course with key_c_1, but we would want key_c_2. Since it gets c_1 course, it will pass its UUID and ecommerce data will go out of sync.
6. Also, the [canonical course run](https://github.com/edx/course-discovery/blob/6f9e323a7a3d7e5a36c8123f693947162a4ffab7/course_discovery/apps/publisher/api/v1/views.py#L205) bit doesn't update as we are not making the new course run, so the course run is the canonical course run of c_1, not c_2.

The problem that have arisen is that wherever we are getting Discovery course based on publisher course run, we will have to check and get the discovery course from publisher run. I have added that property, but I am not sure of the scope of the places where the fix will be required.

staubina:
@DawoudSheraz Are you planning to continue to investigate and work on this issue?

FYI @fysheets

staubina:
Thank you for the detailed status.

DawoudSheraz:
@staubina Yes, I would like to continue the work, if possible. Do you have any suggestions for any further action?

staubina:
It seems like we should take a step back and assess any other changes that would be required to make this work. 

Then we should assess the value in fixing this here or fixing it in the new version of Publisher which relies only on the Course Metadata database.

> I put a breakpoint in publish_to_ecommerce and pressed publish. In ecommerce, it would make the key with publisher course, so it would be like key_c_1. When it tries to get the Discovery course for that key, it is able to find and returns the course with key_c_1, but we would want key_c_2. Since it gets c_1 course, it will pass its UUID and ecommerce data will go out of sync.

At some point, we send the Course Key and potentially the Course UUID for entitlements to create the products.  We need to make sure we sent the course key or UUID for c_2 in this case.

I would assume this would be wherever we create the Products.

DawoudSheraz:
Given all the conversation we had on the PR, this patch is not enough to allow publishing in case of data inconsistencies. Closing this PR.

PR #1840, opened 2019-04-04
Fix publisher change role bug

awaisdar001:
This fixes a bug in publisher where when a user has opted out of emails, and the system tries to remove it from the list of course team users which causes a list ValueError.
I have made the code more robust so that it only removes an email from the list if it is present in there. 

This PR also includes other minor refactoring changes which were necessary and couldn't be missed. 

[https://openedx.atlassian.net/browse/PROD-139](https://openedx.atlassian.net/browse/PROD-139)

Reviewers: 
@asadazam93 
@noraiz-anwar

### DIFF
--- a/course_discovery/apps/publisher/tests/test_emails.py
+++ b/course_discovery/apps/publisher/tests/test_emails.py
@@ -564,15 +564,22 @@ class CourseChangeRoleAssignmentEmailTests(SiteMixin, TestCase):
 
         self.marketing_role = factories.CourseUserRoleFactory(role=PublisherUserRole.MarketingReviewer, user=self.user)
         self.course = self.marketing_role.course
+        self.course_team_role = factories.CourseUserRoleFactory(course=self.course, role=PublisherUserRole.CourseTeam,
+                                                                user=self.user)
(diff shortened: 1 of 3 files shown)


course_discovery/apps/publisher/tests/test_emails.py
```diff
@@ -564,34 +564,64 @@ def setUp(self):
...
+                                                                user=self.user)
         factories.CourseUserRoleFactory(course=self.course, role=PublisherUserRole.Publisher)
         factories.CourseUserRoleFactory(course=self.course, role=PublisherUserRole.ProjectCoordinator)
-        factories.CourseUserRoleFactory(course=self.course, role=PublisherUserRole.CourseTeam)
 
         toggle_switch('enable_publisher_email_notifications', True)
 
+    def update_user_email_notification_status(self, user, enable=True):
+        """Update user attribute which indicate if the user should receive emails or not."""
+        user_attribute, __ = UserAttributes.objects.get_or_create(user=user)
```
noraiz-anwar, line 576: Maybe a single underscore would be enough?

noraiz-anwar (approved):
Just a nit. Seems good overall.

asadazam93 (approved):

PR #1841, opened 2019-04-04
Add migrations for new statuses.

dianakhuang:
This seems to have picked up some of the new managers for the drafts as well.


schenedx (approved):
👍

mikix (approved):
A 255 long enum CharField feels generous. :) Not your doing, I know.

I like to see human readable file names for migrations, but it's no big deal. Thank you for catching the manager migration too - I did not realize those went into migration files.

We should write a test that ran makemigrations and failed if any were found, eh?

PR #1843, opened 2019-04-04
Support creating and editing course drafts

mikix:
Now every time we hit the course API endpoint, we'll create any drafts that are necessary and then edit those drafts.


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -362,28 +362,30 @@ def test_course_without_editors(self):
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.uuid})
         self.user.is_staff = False
         self.user.save()
+        self.course.draft = True
+        self.course.save()
 
         # Try without being in the organization nor an editor
-        self.assertEqual(self.client.patch(url).status_code, 403)
+        self.assertEqual(self.client.patch(url).status_code, 404)
```
mikix, line 377: These 404/403 changes are because now that we filter based on editable, the object just isn't found at all and does a 404. I don't super care about the distinction, so I didn't add code to explicitly check for this case. But am willing to discuss if this 403 feels more correct to folks.

staubina: Do we have any concern that users of the endpoint might have issues with the new status code?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -42,8 +44,8 @@ def inner(*args, **kwargs):
         except ValidationError as exc:
             return Response(exc.message if hasattr(exc, 'message') else str(exc),
                             status=status.HTTP_400_BAD_REQUEST)
-        except PermissionDenied:
-            raise  # just pass a 403 along
+        except (PermissionDenied, Http404):
+            raise  # just pass some exceptions along
```
staubina: Do we care to log out here that we denied permission on a request?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -101,7 +103,7 @@ def get_queryset(self):
         q = self.request.query_params.get('q')
 
         # Start with either draft versions or real versions of the courses
-        if get_query_param(self.request, 'editable'):
+        if get_query_param(self.request, 'editable') or self.request.method not in SAFE_METHODS:
```
staubina: `self.request.method not in SAFE_METHODS`

I am interested in this addition here.  Why would we want to set the editable courses on methods that are not "SAFE"?

course_discovery/apps/api/v1/views/utils.py
```diff
@@ -20,14 +20,25 @@ def set_draft_state(obj, model, attrs=None):
...
     # Will throw an integrity error if the draft row already exists, but this
     # should be caught as part of a try catch in the API calling ensure_draft_world
     obj.save()
+
     original_obj.draft_version = obj
     original_obj.save()
+
+    # Copy many-to-many fields manually (they are not copied by the pk=None trick above).
+    # This must be done after the save() because we need an id.
+    for f in model._meta.get_fields():
```
staubina: nit: can we use a more descriptive name for `f`?

staubina (changes requested):
Some questions and small comments.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
mikix: Naw - the only user of this flow is PFE.

course_discovery/apps/api/v1/views/courses.py
mikix: SAFE methods are get, options, head, etc. Non-SAFE methods are post, put, and patch.

So basically, this is "if the user asked for draft mode or they are trying to write, give 'em drafts"

course_discovery/apps/api/v1/views/courses.py
mikix: Sure... Makes some sense

course_discovery/apps/api/v1/views/utils.py
mikix: Done.

course_discovery/apps/api/v1/views/courses.py
mikix: Upon testing, we do log the status codes like 403 and 404 naturally. That's probably enough?

mikix:
@staubina I added the course_runs-being-empty fix to this PR. I've tested manually with my created-runs, but if you also want to merge-and-test, that would be welcome. Else we'll see.

course_discovery/apps/api/v1/views/courses.py
staubina: I recommend a comment here then to make that a little more clear.

Also team note, should we write an ADR about the API changes we have made to go with the Draft ADR?

staubina:
I will pull down the changes and test locally

staubina (approved):
Looks good to me.

course_discovery/apps/api/v1/views/courses.py
mikix: I believe our draft ADR covered the fact that when writing / reading, we'll deal with drafts:

```
Add a query parameter to list endpoints that will return draft versions of
rows if available, and if not available, will return the official versions.
We can just re-use the editable=1 flag for that.

All form and API updates will be applied to the draft row. And only after a
successful review, the data of the draft row will be written to the official
row. (Although that "review" might be a no-op depending on business logic -
it may be skipped if the course run is already live for example.)
```

These API changes are additions, not contract breakages - they used to just not allow writes at all.

PR #1845, opened 2019-04-09
Don't push draft data to ecommerce

mikix:
We'll add pushing to ecommerce back once we have a review flow. Then, once successfully reviewed, a course's ecommerce data will be updated or created.


Dillon-Dumesnil (changes requested):
Now that we won't be storing the sku, the `update_entitlements` function should be updated to not check for that. Might need to verify that the CourseEntitlement serializer doesn't freak out about that, but if it does, calling update directly would be a fix.

Secondary question, should we update the ecommerce api updates we made or leave those for when we want to use them?

mikix:
Ugh great point abut sku. Removed. I think we can keep the ecommerce API changes in, yeah? We'll need em soon.

Dillon-Dumesnil:
I'm happy to leave them, just want to make sure it's a conscious decision and not because we forgot about them.

Dillon-Dumesnil (approved):
Beautiful. Just wish there was one more line to delete so you could hit that nice 200 deletions

PR #1846, opened 2019-04-09
add checkbox for master's track creation on new course run publisher form

jansenk:
Add checkbox to publisher Seat form and corresponding boolean field to publisher Seat model

When publisher course_run is published to discovery, if there is a Publisher.Seat that has this flag, also create a Masters Course_Metadata.Seat (which in turn should create the course mode) 

### Creating a run for a course that maps to a course in course_discovery but is not in a masters program
no_masters_hidden

### Refresh after adding the course_metadata course to a masters program curriculum
masters_program_show

### Refresh after adding a canonical_course_run to the course and adding a masters seat to the run
master_seat_checked


course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -359,3 +359,26 @@ def test_publish_with_ecommerce_api_error(self, mock_access_token):  # pylint: d
...
+
+        self._mock_studio_api_success(publisher_course_run)
+        self._mock_ecommerce_api(publisher_course_run)
+
+        publish_url = reverse('publisher:api:v1:course_run-publish', kwargs={'pk': publisher_course_run.pk})
+        response = self.client.post(publish_url, {})
+        assert response.status_code == 200
+        discovery_course_run = CourseRun.objects.get(key=publisher_course_run.lms_course_id)
+        assert discovery_course_run.seats.all().count() == 2
+        assert discovery_course_run.seats.filter(type=DiscoverySeat.MASTERS).count() == 1
```
iloveagent57: Nit: end the file with a newline

course_discovery/apps/publisher/forms.py
```diff
@@ -470,6 +475,7 @@ def clean(self):
         if seat_type == Seat.CREDIT and not credit_price:
             self.add_error('credit_price', _('Only audit seat can be without price.'))
 
+        print(self.errors)
```
iloveagent57: please delete

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -529,3 +529,23 @@ def test_type_is_required(self):
 
         seat_form_with_type = SeatForm(data={'type': Seat.AUDIT})
         self.assertTrue(seat_form_with_type.is_valid())
+
+        seat_form_masters = SeatForm(data={'masters_track': True})
+        self.assertTrue(seat_form_masters.is_valid())
+    @ddt.data(
```
iloveagent57, line 537: You need a newline here.

course_discovery/apps/publisher/tests/test_forms.py
```diff
@@ -529,3 +529,23 @@ def test_type_is_required(self):
...
+        {'type': 'audit'},
+        {'type': 'audit', 'masters_track': True},
+        {'masters_track': True},
+    )
+    def test_create_seat_masters_track(self, form_data):
+        course_run = CourseRunFactory()
+        form = SeatForm(data=form_data)
+        seat = form.save(course_run=course_run)
+        self.assertEqual('masters_track' in form_data, seat.masters_track)
+
```
iloveagent57: ...and fewer newlines here.

course_discovery/apps/publisher/views.py
```diff
@@ -918,17 +919,44 @@ def _process_post_request(self, request, parent_course, context=None):
             logger.exception('Unable to create course run and seat for course [%s].', parent_course.id)
             return self._render_post_error(request, ctx_overrides=context)
 
+    def get_course_key(self, publisher_course):
+        return '{org}+{number}'.format(org=publisher_course.organizations.first().key,
+                                            number=publisher_course.number)
+
```
iloveagent57: 1 too many newlines

course_discovery/apps/publisher/views.py
```diff
@@ -918,17 +919,44 @@ def _process_post_request(self, request, parent_course, context=None):
...
+
+    def lookup_masters_info(self, publisher_course):
```
iloveagent57: This could use a doc string.

iloveagent57 (commented):
This looks good in general, mostly linting errors.

course_discovery/apps/publisher/views.py
```diff
@@ -743,6 +745,41 @@ def _get_last_run(self, parent_course):
 
         return self.last_run
 
+    def _get_course_key(self, publisher_course):
```
iloveagent57: This already exists here: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/publisher/api/v1/views.py#L36 . It's probably worth extracting this into a `utils.py` and using that from both the API views and this view

course_discovery/apps/publisher/views.py
```diff
@@ -743,6 +745,41 @@ def _get_last_run(self, parent_course):
...
+        course_in_masters = False
+        last_run_masters_seat = False
+        course_key = self._get_course_key(publisher_course)
+        try:
+            discovery_course = DiscoveryCourse.objects.get(key=course_key)
+        except DiscoveryCourse.DoesNotExist:
+            return False, False
+        curriculum_memberships = discovery_course.curriculum_course_membership.all()
+        for membership in curriculum_memberships:
+            if membership.curriculum.program.type.slug == 'masters':
```
iloveagent57: We have another place that's checking for this "magic string" https://github.com/edx/course-discovery/blob/2c0819c411e1af4113e814ed718256e3fad14d32/course_discovery/apps/course_metadata/signals.py#L30 . It would be good to expose a module-level constant somewhere like `MASTERS_PROGRAM_TYPE_SLUG = 'masters'` that could replace this string literal.

course_discovery/apps/publisher/views.py
```diff
@@ -743,6 +745,41 @@ def _get_last_run(self, parent_course):
...
         return self.last_run
 
+    def _get_course_key(self, publisher_course):
+        """ Construct the course key to look up a course_metadata course from a publisher course """
+        return '{org}+{number}'.format(
+            org=publisher_course.organizations.first().key,
+            number=publisher_course.number
+        )
+
+    def _lookup_masters_info(self, publisher_course):
```
iloveagent57: I think it'd be better to break this into two functions: `_course_in_masters_program()` and `_canonical_run_has_masters_seat()`.  If the result of the first function isn't `True`, you don't have to call the second one.  And I guess one helper function to get a discovery course given a publisher course.  So something like:
```python
def _discovery_course(pubisher_course):
    course_key = self._get_course_key(publisher_course)
    try:
       return DiscoveryCourse.objects.get(key=course_key)
    except DiscoveryCourse.DoesNotExist:
        return False
```

And then down in `get_context_data`:
```python
course_in_masters, last_run_masters_seat = False, False
discovery_course = _discovery_course(parent_course)
if discovery_course:
    course_in_masters = _course_in_masters_program(discovery_course)
    last_run_masters_seat = course_in_masters and _canonical_run_has_masters_seat(discovery_course)
...
```
This adds a few lines to `get_context_data()`, but it makes what and how you're looking up a lot more explicit.

course_discovery/apps/publisher/views.py
```diff
@@ -1425,7 +1463,8 @@ def get(self, request, *args, **kwargs):
     def post(self, request, *args, **kwargs):
         """Post method for import page."""
 
-        #  inline import to avoid any circular issues.
+        # inline import to avoid any circular issues.
+        # pylint: disable=reimported
```
iloveagent57: This is a bad smell (I know you didn't write the code below).  Does doing a `s/CourseMetaData/DiscoveryCourse` clean this up?  Like, you're already importing the same thing at the module-level, so clearly there's no circular import.

iloveagent57 (commented):
More clean-code/refactoring suggestions, lemme know if you want help looking at the reimport nonsense.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -632,13 +635,65 @@ def test_existing_run_and_seat_data_auto_populated(self):
...
+            course=discovery_course,
+            curriculum=discovery_curriculum
+        )
+        if last_run_masters_seat:
+            discovery_course.canonical_course_run = CourseRunFactory(course=discovery_course)
+            discovery_course.save()
+            discovery_course.canonical_course_run.seats.add(SeatFactory(type=DiscoverySeat.MASTERS))
+            discovery_course.canonical_course_run.save()
+
+        print(discovery_course.key)
```
iloveagent57: nit: stray print

course_discovery/apps/publisher/views.py
```diff
@@ -743,6 +746,37 @@ def _get_last_run(self, parent_course):
 
         return self.last_run
 
+    def _get_discovery_course(self, publisher_course):
+        """
+        Given a pubisher course, lookup the course_metadata course
+        """
+        course_key = get_course_key(publisher_course)
+        print(course_key)
```
iloveagent57: nit: stray print

course_discovery/apps/publisher/views.py
```diff
@@ -920,15 +954,23 @@ def _process_post_request(self, request, parent_course, context=None):
 
     def get_context_data(self, **kwargs):
         parent_course = self._get_parent_course()
+        course_in_masters, last_run_masters_seat = False, False
+        discovery_course = self._get_discovery_course(parent_course)
+        print(discovery_course)
```
iloveagent57: stray print

course_discovery/apps/publisher/views.py
```diff
@@ -920,15 +954,23 @@ def _process_post_request(self, request, parent_course, context=None):
...
+        if discovery_course:
+            course_in_masters = self._course_in_masters_program(discovery_course)
+            last_run_masters_seat = course_in_masters and self._canonical_run_has_masters_seat(discovery_course)
+            print(course_in_masters)
```
iloveagent57: two stray prints

iloveagent57 (approved):
You've got a couple stray prints to kill, but it looks good, otherwise.

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -359,3 +359,26 @@ def test_publish_with_ecommerce_api_error(self, mock_access_token):  # pylint: d
...
         assert response.data == expected
+
+    @responses.activate
+    @mock.patch.object(Partner, 'access_token', return_value='JWT fake')
+    def test_publish_masters_track_seat(self, mock_access_token):
+        """
+        Test that publishing a seat with masters_track creates a masters seat and masters track
+        """
+        publisher_course_run = self._create_course_run_for_publication()
+        masters_seat = SeatFactory(type="Audit", course_run=publisher_course_run, masters_track=True)
```
schenedx: #NIT name of this variable is more accurate with `audit_seat_with_masters_track`

course_discovery/apps/publisher/forms.py
```diff
@@ -405,12 +405,13 @@ def __init__(self, *args, **kwargs):
         (Seat.CREDIT, _('Credit')),
     ]
 
-    type = forms.ChoiceField(choices=TYPE_CHOICES, required=True, label=_('Enrollment Track'))
+    type = forms.ChoiceField(choices=TYPE_CHOICES, required=False, label=_('Enrollment Track'))
```
schenedx: Why the type here is no longer required? Is this for the case of `Masters Only` track creation? How does Publish button handle seats with type being empty?

course_discovery/apps/publisher/views.py
```diff
@@ -743,6 +746,36 @@ def _get_last_run(self, parent_course):
...
 
+    def _get_discovery_course(self, publisher_course):
+        """
+        Given a pubisher course, lookup the course_metadata course
+        """
+        course_key = get_course_key(publisher_course)
+        try:
+            return DiscoveryCourse.objects.get(key=course_key)
+        except DiscoveryCourse.DoesNotExist:
+            return False
```
schenedx: #NIT please return `None` here since the function returns objects

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -187,6 +184,16 @@ def publish_to_discovery(self, partner, course_run):
                     'upgrade_deadline': seat.calculated_upgrade_deadline,
                 }
             )
+            if seat.masters_track:
```
schenedx, line 187: I am worried for the case where the seat have no valid `type` value. Is that a possible case? If that is the case, we would create a seat with no `type` on the course_discovery/course_metadata, right?

schenedx (commented):
I got a few questions.

schenedx (approved):
LGTM 👍

course_discovery/apps/publisher/models.py
```diff
@@ -638,6 +638,7 @@ class Seat(TimeStampedModel, ChangedByMixin):
     credit_provider = models.CharField(max_length=255, null=True, blank=True)
     credit_hours = models.IntegerField(null=True, blank=True)
     credit_price = models.DecimalField(**PRICE_FIELD_CONFIG)
+    masters_track = models.BooleanField(default=False)
```
Dillon-Dumesnil, line 641: Why make this a boolean field instead of a seat type to replicate the way it is in course_metadata? This would also eliminate the need for the special `if seat.master_track:` code above.

Dillon-Dumesnil (approved):
Just one question, but I don't see anything troubling. Looks good.

course_discovery/apps/publisher/models.py
jansenk: We could have added masters to the list of possible seat types, but we wanted to add this as an option in addition to the seat type selection that was already being made. We wanted them to be able to select any other seat type and also whether or not to add a masters seat.

Dillon-Dumesnil: Ah so it's a boolean in publisher that tells course_metadata to add it as an actual seat. That sounds fine then. I would just make sure you all are keeping in mind that we are trying to transition to only using course_metadata and how to handle master's seat creation there as well. I don't think it should be a major difference and likely just a boolean field that will be passed to the api along with the rest of the course run data, but want to make sure you were aware of this transition.

PR #1848, opened 2019-04-10
Fixing the course run endpoint to not only look for official versions…

Dillon-Dumesnil:
… on creation since any new courses will only have the draft version


mikix (approved):
Seems good, nice catch

PR #1849, opened 2019-04-11
DISCO-781: Forces queries for draft version

jlajoie:


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -610,10 +637,11 @@ def test_update_operates_on_drafts(self):
         self.assertTrue(course.entitlements.first().draft)
         self.assertEqual(course.title, 'Title')
 
-        self.course.refresh_from_db()
-        self.assertFalse(self.course.draft)
-        self.assertFalse(self.course.entitlements.first().draft)
-        self.assertEqual(self.course.title, 'Fake Test')
+        # TODO: Uncomment block below as part of DISCO-818 when we care about non-drafts again
```
staubina, line 641: 👍 For marking what we need to reverse

staubina (approved):

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -97,7 +97,8 @@ def get_queryset(self):
 
         # Start with either draft versions or real versions of the courses
         if edit_mode:
-            queryset = Course.objects.filter_drafts()
+            # TODO: For now hardcode in draft=True until we choose to roll this out live, DISCO-818
+            queryset = Course.objects.filter_drafts(draft=True)
```
mikix, line 101: It might be worth also touching the CourseRun endpoint? But it doesn't matter in practice, this is all temporary anyway, and it would be extra test-change work. So ignore me.

mikix (approved):
I think that's all we need?!

PR #1850, opened 2019-04-11
Allow incoming nulls for some course serializer fields

mikix:


course_discovery/apps/api/serializers.py
```diff
@@ -540,7 +540,7 @@ class CourseEntitlementSerializer(serializers.ModelSerializer):
         max_digits=CourseEntitlement.PRICE_FIELD_CONFIG['max_digits']
     )
     currency = serializers.SlugRelatedField(read_only=True, slug_field='code')
-    sku = serializers.CharField()
+    sku = serializers.CharField(allow_blank=True, allow_null=True)
```
mikix, line 543: We send null from UI, despite it being a charfield. And the model itself has blank=True and null=True... So I did both here.

jlajoie (approved):

PR #1851, opened 2019-04-11
Bjh/disco 805 disallow non draft post

inventhouse:
This is for the first AC on the parent ticket

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -148,6 +148,28 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         self.assertEqual(str(new_course_run.end), '2001-01-01 00:00:00+00:00')  # spot check that input made it
         self.assertTrue(new_course_run.draft)
 
+    @ddt.data(True, False, "bogus")
+    @responses.activate
+    def test_create_draft_ignored(self, draft):
+        """ Verify the endpoint supports creating a course_run, but always as a draft. """
+        course = self.draft_course_run.course
+        new_key = 'course-v1:{}+1T2000'.format(course.key.replace('/', '+'))
+        self.mock_post_to_studio(new_key)
+        url = reverse('api:v1:course_run-list')
+
+        # Send minimum + draft: True/False/bogus
+        response = self.client.post(url, {
+            'course': course.key,
+            'start': '2000-01-01T00:00:00Z',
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -178,6 +178,10 @@ def create(self, request, *args, **kwargs):
         # Set a pacing default when creating (studio requires this to be set, even though discovery does not)
         request.data.setdefault('pacing_type', 'instructor_paced')
 
+        # Guard against externally setting the draft state
+        if 'draft' in request.data:
+            del request.data['draft']
```
mikix: Probably want to change update() too.

This would visually look a little cleaner to me as the one-liner `request.data.pop('draft', None)`, but it's a matter of style.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -148,6 +148,29 @@ def test_create_minimum(self):
         self.assertEqual(str(new_course_run.end), '2001-01-01 00:00:00+00:00')  # spot check that input made it
         self.assertTrue(new_course_run.draft)
 
+    @responses.activate
+    def test_create_non_draft(self):
```
Dillon-Dumesnil: Nit: we are trying to switch to the language of draft and official. Switch test to `test_create_official`

Dillon-Dumesnil: I think this test and the last test (test_create_minimum) could just be combined using ddt and passing in True, False for the two different draft states. Then the bottom assert can just be `self.assertEqual(isDraft, new_course_run.draft)`

course_discovery/apps/api/v1/views/course_runs.py
inventhouse: ✅

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
inventhouse: Renamed to better reflect purpose of the test

Dillon-Dumesnil: This was discussed in person and we believe now that the test is ensuring a draft parameter is always ignored for a POST, it is sufficiently different from the create_minimum draft.

Dillon-Dumesnil (approved):
Rebase, squash, commit, :shipit: 

Nice job!

mikix (approved):
:+1:

PR #1854, opened 2019-04-12
Remove the lms_commerce_api_url field from partners object

schenedx:
@edx/masters-neem Please review

This is to remove the lms_commerce_api_url from the partner object because the field is no longer used and referenced.


jansenk (approved):
lgtm, just gotta rebase

PR #1855, opened 2019-04-12
WIP-Add history to Course and CourseRun

hunytalk:
This PR adds History for Course and CourseRun models.


bmedx (commented):
You will also need to locally run a Django `makemigrations` command to create the migrations files that will create the new tables.

PR #1857, opened 2019-04-16
Move course runs into Review when draft parameter passed in [DISCO-806]

inventhouse:


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -212,15 +212,14 @@ def update(self, request, *args, **kwargs):
...
-        # (For now, soon that should trigger status change to review)
-        request.data.pop('draft', None)
-
         # If changes are made after review and before publish,
         # revert status to unpublished.
         if course_run.status == CourseRunStatus.Reviewed:
             request.data['status'] = CourseRunStatus.Unpublished
 
+        if not request.data.get('draft', True) and not course_run.status == CourseRunStatus.Published:
+            request.data['status'] = CourseRunStatus.LegalReview
```
mikix, line 223: A comment above this line might be useful. Nothing crazy, just mention that a review was requested, so we're givin' it to 'em.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -212,15 +212,14 @@ def update(self, request, *args, **kwargs):
                 status=status.HTTP_403_FORBIDDEN
             )
 
-        # Guard against externally setting the draft state
-        # (For now, soon that should trigger status change to review)
-        request.data.pop('draft', None)
```
mikix, line 217: We still don't want the `draft` field to be in `request.data` - so the pop still makes sense. Maybe just keep the value instead of throwing it away and use `True` instead of `None` as the default.

course_discovery/apps/api/v1/views/course_runs.py
inventhouse: ✅

course_discovery/apps/api/v1/views/course_runs.py
inventhouse: ✅

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -212,15 +212,16 @@ def update(self, request, *args, **kwargs):
...
-
         # If changes are made after review and before publish,
         # revert status to unpublished.
         if course_run.status == CourseRunStatus.Reviewed:
             request.data['status'] = CourseRunStatus.Unpublished
 
+        # Sending draft=False triggers the review process
+        draft = request.data.pop('draft', True)  # Don't let draft parameter trickle down
+        if not draft and not course_run.status == CourseRunStatus.Published:
```
mikix: Oh I just noticed the construction here - `!=` feels more natural to me than `not ... ==`

mikix (approved):
Seems good to me, just one nit comment.

course_discovery/apps/api/v1/views/course_runs.py
inventhouse: ✅

dianakhuang (approved):
LGTM.

PR #1858, opened 2019-04-16
Create official version when moving to Reviewed state.

dianakhuang:
https://openedx.atlassian.net/browse/DISCO-759

When a CourseRun is moved into `Reviewed`, we create the official version (if necessary) and update all references.

FYI: @inventhouse


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -514,6 +514,62 @@ def test_get_video(self):
...
+        factories.CourseEntitlementFactory(course=self.course_run.course, draft=True)
+        self.course_run.course.save()
+        self.course_run.save()
+        assert CourseRun.everything.all().count() == 2
+        official_run = CourseRun.everything.get(key=self.course_run.key, draft=False)
+        draft_run = CourseRun.everything.get(key=self.course_run.key, draft=True)
+
+        assert official_run.draft_version == draft_run
+        assert official_run != draft_run
+        assert official_run.slug == draft_run.slug
```
mikix, line 531: You don't like self.assertEqual?

dianakhuang: I have no strong opinions on it. This is just the style of the file I was writing in. From what I understand, `pytest` does all the fancy handling of those cases with bare `assert`s now. 🤷‍♀️

mikix: Oh really!? Fascinating. Does it catch them and fancify them? Or maybe it redefines the `assert` keyword, because it probably can, because Python is nuts.

course_discovery/apps/course_metadata/models.py
```diff
@@ -42,6 +42,38 @@
...
+            obj.pk = None
+            obj.draft = False
+            obj.draft_version = draft_version
+            obj.save()
+            official_obj = obj
+            # Copy many-to-many fields manually (they are not copied by the pk=None trick above).
+            # This must be done after the save() because we need an id.
+            for field in model._meta.get_fields():
+                if field.many_to_many and not field.auto_created:
+                    getattr(draft_version, field.name).add(*list(getattr(official_obj, field.name).all()))
```
Dillon-Dumesnil: I believe you want to switch `draft_version` and `official_obj` here so we are adding the fields from the draft into the official

course_discovery/apps/course_metadata/models.py
```diff
@@ -42,6 +42,38 @@
...
+    else:
+        official_obj = obj
+
+    # Now set fields we were told to
+    if attrs:
+        for key, value in attrs.items():
+            setattr(official_obj, key, value)
+
+    # Will throw an integrity error if the draft row already exists, but this
```
Dillon-Dumesnil: Update the comment for official and set_official_state. 
Especially consider the part about if this will throw an error or not. I think it might not since you have an explicit check of if the official version exists or not.

course_discovery/apps/course_metadata/models.py
```diff
@@ -996,6 +1046,21 @@ def get_video(self):
     def __str__(self):
         return '{key}: {title}'.format(key=self.key, title=self.title)
 
+    def ensure_official_version(self):
+        # If there isn't an official version yet, create one.
+        if not self.official_version:
+            official_version = set_official_state(self, CourseRun)
+            draft_version = official_version.draft_version
+            official_version.slug = draft_version.slug
+            for seat in draft_version.seats.all():
```
Dillon-Dumesnil, line 1026: Nit: Stay consistent in how you do these loops. In the course ensure_official_version, you use `official_course.draft_version.entitlements` and here you pull out the draft_version. I don't really have a preference, just consistency.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -514,6 +514,62 @@ def test_get_video(self):
...
+
+        assert official_run.course.draft is False
+        assert official_run.course.draft_version == draft_run.course
+        assert official_run.course != draft_run.course
+        assert official_run.slug == draft_run.slug
```
Dillon-Dumesnil: I think you might mean `assert official_run.course.slug == draft_run.course.slug`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -514,6 +514,62 @@ def test_get_video(self):
...
+        assert official_entitlement.draft_version == draft_entitlement
+        assert official_entitlement.draft is False
+        assert draft_entitlement.draft is True
+        assert official_entitlement != draft_entitlement
+
+        official_seat = official_run.seats.first()
+        draft_seat = draft_run.seats.first()
+        assert official_seat.draft_version == draft_seat
+        assert official_seat != draft_seat
+        assert official_seat.draft is False
```
Dillon-Dumesnil, line 548: Super nit: Since the entitlement and seat checks are identical, I'd like if the order the asserts are called in was mirrored.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -514,6 +514,62 @@ def test_get_video(self):
...
+        self.course_run.draft = True
+        self.course_run.status = CourseRunStatus.Reviewed
+        self.course_run.course.canonical_course_run = self.course_run
+        self.course_run.course.draft = True
+        self.course_run.course.save()
+        self.course_run.save()
+
+        official_run = CourseRun.everything.get(key=self.course_run.key, draft=False)
+        assert official_run.course.canonical_course_run == official_run
+
```
Dillon-Dumesnil, line 562: Probably not necessary, but maybe add in an additional check to verify the draft course has the draft course run as its canonical run. We want to make sure they don't both point to the same thing

course_discovery/apps/course_metadata/models.py
```diff
@@ -562,6 +594,20 @@ def first_enrollable_paid_seat_price(self):
 
         return None
 
+    def ensure_official_version(self, course_run):
+        if self.official_version:
+            return self.official_version
+
+        self.canonical_course_run = None
+        official_course = set_official_state(self, Course, {'canonical_course_run': course_run})
```
Dillon-Dumesnil, line 570: So I think this code will overwrite the official version's canonical_course_run which is something we don't actually want to do. The canonical course run should only be set to the first ever run for a course (because of weird stuff in studio I believe).

Dillon-Dumesnil (commented):
Couple comments, but overall looks like a great start.

One thing that I want to call out that needs to happen at some point and is related to this, we still need to create the ecommerce products as part of going from `In Review` to `Reviewed`. I'm happy with it not being a part of this PR, but want to make sure we don't forget that part either.

course_discovery/apps/course_metadata/models.py
dianakhuang: We had a conversation with @mikix about this, and his current recommendation is to assume that the first time an official course is created, that course run is the canonical. I am happy to add a comment to this effect.

mikix: @Dillon-Dumesnil you weren't around at the time - but your take would be valuable here. In the case when the draft course points at run A as the canonical course run, but run B is the first to be reviewed and made official - what should we do?

My takeaway was that it wouldn't be bad to switch to run B as the canonical run. But I wasn't 100% sure about side effects of doing that? Or maybe there's a better way.

Dillon-Dumesnil: My original comment was because I saw this canonical_course_run being passed in as an attr and that once you hit set_official_state, regardless of an official version existing, it overwrites the attrs. Diana pointed out that there is the check before this line is run that handles this scenario.

Re your point: We discussed this more in person and believe it is safe and recommended to switch the canonical_course_run to run B since that will be the first official version. That being said, for syncing reasons, we should also update the draft course to point to run B as its canonical_course_run in this case.

Dillon-Dumesnil (approved):
After responding to the feedback on the canonical course run 👍

Dillon-Dumesnil (approved):
LGTM

course_discovery/apps/course_metadata/models.py
```diff
@@ -42,6 +42,35 @@
 logger = logging.getLogger(__name__)
 
 
+def set_official_state(obj, model, attrs=None):
```
mikix: This kinda function living in models.py feels off to me. It would probably make sense to live in a utils.py file. And it's so closely related to the api/views/utils.py file that has the reverse direction, maybe it makes sense for both to live in the same file (though one is needed for api, one is needed for course_metadata - not sure where is best for both).

course_discovery/apps/course_metadata/models.py
```diff
@@ -562,6 +591,23 @@ def first_enrollable_paid_seat_price(self):
 
         return None
 
+    def ensure_official_version(self, course_run):
```
mikix, line 565: These two `ensure_official_version` methods feel a little "heavy" for models.py. I might throw them in the same utility file as the above comment suggested.

That is an extremely subjective call though. 

There is a useful distinction between `world` and `version` here - world does *everything*, version does just it and the course. Might be worth talking about that verbiage somewhere in a comment.

mikix (approved):

course_discovery/apps/course_metadata/models.py
dianakhuang: There is a `course_metadata/utils.py` file, which might be the best place for both this and the `api/views/utils.py` methods.

PR #1861, opened 2019-04-16
Be more permissive when receving OPTIONS requests

mikix:
The handling of an OPTIONS request goes through a fake PUT/POST
request to determine permission levels. So we should allow that
in POST situations where the incoming request may not have all
the data we expect.

This should help fix publisher-frontend with non-staff users.

### DIFF
--- a/course_discovery/apps/api/permissions.py
+++ b/course_discovery/apps/api/permissions.py
@@ -24,7 +24,11 @@ class IsCourseEditorOrReadOnly(BasePermission):
     def has_permission(self, request, view):
         if request.method == 'POST':
             org = request.data.get('org')
-            return org and CourseEditor.can_create_course(request.user, org)
+            if not org:
+                # Fail happily because OPTIONS goes down this path too with a fake POST.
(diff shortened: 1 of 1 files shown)


jlajoie (approved):
Seems like a reasonable approach :+1:

course_discovery/apps/api/permissions.py
```diff
@@ -24,7 +24,11 @@ class IsCourseEditorOrReadOnly(BasePermission):
     def has_permission(self, request, view):
         if request.method == 'POST':
             org = request.data.get('org')
-            return org and CourseEditor.can_create_course(request.user, org)
+            if not org:
```
staubina, line 27: Should we look for the OPTIONS request method explicitly?

mikix: We can't - at this point, it's faking a POST. And doesn't set a flag that it's a fake POST.  :(

It actually goes and clones it's request then re-runs the permissions with that new cloned request.

staubina: Oh 🤦‍♂️

PR #1864, opened 2019-04-16
Allow re-publishing already published course-runs

inventhouse:


course_discovery/apps/course_metadata/utils.py
```diff
@@ -53,20 +53,19 @@ def set_official_state(obj, model, attrs=None):
     """
```
dianakhuang, line 53: I think we should update the docstring with this new behavior.

dianakhuang (approved):
Just one minor comment, but this looks good otherwise.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -289,7 +289,7 @@ def test_partial_update_no_studio_url(self):
         with mock.patch('course_discovery.apps.api.v1.views.course_runs.log.info') as mock_logger:
             response = self.client.patch(url, {}, format='json')
 
-        self.assertEqual(response.status_code, 200)
+        assert response.status_code == 200, "Status {}: {}".format(response.status_code, response.content)
```
mikix: assertEqual can also take a third argument as the message.

mikix (approved):
Seems reasonable! Thanks!

PR #1866, opened 2019-04-22
Adding go live date to course run

Dillon-Dumesnil:


course_discovery/apps/api/serializers.py
```diff
@@ -684,10 +684,14 @@ def ensure_key(self, data):
     def validate(self, data):
         start = data.get('start', self.instance.start if self.instance else None)
         end = data.get('end', self.instance.end if self.instance else None)
+        go_live_date = data.get('go_live_date')
```
mikix: You should do the same thing as above - `self.instance.go_live_date if self.instance else None` - that will ensure that you have access to current value, even if it isn't being changed.

mikix (approved):
Didn't manually test, but seems reasonable. Please fix the thing I pointed out, but besides that, approved.

PR #1868, opened 2019-04-22
If the draft course is in Reviewed, both draft and official should be kicked back to Unpublished

inventhouse:


Dillon-Dumesnil (approved):
LGTM

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -216,6 +216,9 @@ def update(self, request, *args, **kwargs):
         # revert status to unpublished.
         if course_run.status == CourseRunStatus.Reviewed:
             request.data['status'] = CourseRunStatus.Unpublished
+            official_run = course_run.official_version
+            official_run.status = CourseRunStatus.Unpublished
```
dianakhuang, line 221: Ww should be assured that an official version exists at this point, but should we be defensive about it anyway?

dianakhuang (approved):
Minor nitpick.

course_discovery/apps/api/v1/views/course_runs.py
inventhouse: Something like `if not official_run: official_run = ensure_official_version()`?

dianakhuang: Yeah, I think that would be a good idea, just to be defensive about it.

PR #1869, opened 2019-04-22
Updating person-autocomplete to use DRF

mikedikan:
DISCO-840

Rebuilding person-autocomplete to use DRF due to authentication issues in stage.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 import urllib.parse
 
 import ddt
@@ -8,13 +9,14 @@ from django.urls import reverse
 from course_discovery.apps.api import serializers
 from course_discovery.apps.api.v1.tests.test_views import mixins
 from course_discovery.apps.api.v1.views.search import TypeaheadSearchView
-from course_discovery.apps.core.tests.factories import PartnerFactory
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, PartnerFactory, UserFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import CourseRun
 from course_discovery.apps.course_metadata.tests.factories import (
-    CourseFactory, CourseRunFactory, OrganizationFactory, PersonFactory, ProgramFactory
+    CourseFactory, CourseRunFactory, OrganizationFactory, PersonFactory, PositionFactory, ProgramFactory
 )
+from course_discovery.apps.publisher.tests import factories as publisher_factories
 
 
 @ddt.ddt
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -1,20 +1,24 @@
 import datetime
+import json
 import urllib.parse
 
 import ddt
 import pytz
+from django.test import TestCase
```
dianakhuang: Nit: Why are we using this instead of the `APITestCase` imported already?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -1,20 +1,24 @@
...
-from course_discovery.apps.core.tests.factories import PartnerFactory
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, PartnerFactory, UserFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import CourseRun
 from course_discovery.apps.course_metadata.tests.factories import (
-    CourseFactory, CourseRunFactory, OrganizationFactory, PersonFactory, ProgramFactory
+    CourseFactory, CourseRunFactory, OrganizationFactory, PersonFactory, PositionFactory, ProgramFactory
 )
+from course_discovery.apps.publisher.tests import factories
```
dianakhuang: I know this is imported elsewhere like this, but I would rename this to be `publisher_factories`, since it's less confusing that way.

course_discovery/apps/api/v1/tests/test_views/test_search.py
inventhouse: most of the tests were copied from the old version; can I just drop-in replace APITestCase?  I'll give that a try at least...

course_discovery/apps/api/v1/urls.py
```diff
@@ -27,6 +27,7 @@
     url(r'currency', CurrencyView.as_view(), name='currency'),
     url(r'^catalog/query_contains/?', CatalogQueryContainsViewSet.as_view(), name='catalog-query_contains'),
     url(r'^replace_usernames/$', UsernameReplacementView.as_view(), name="replace_usernames"),
+    url(r'^search/person_typeahead', search_views.PersonTypeaheadSearchView.as_view(), name='person-search-typeahead'),
```
mikix: I might have put this right under the search/typeahead line.

course_discovery/apps/api/v1/tests/test_views/test_search.py
inventhouse: Rename in code or import-as?  I'll see if the former is reasonably small

mikix (approved):
Nice. In testing, does this avoid the login redirect issue?

mikix (commented):
Also, we can remove the bulked up support in the old endpoint that we added for org arguments.

course_discovery/apps/api/v1/tests/test_views/test_search.py
dianakhuang: I just meant in import-as. Renaming in code is probably not worth the effort.

dianakhuang:
@mikix I am working on that separately: https://github.com/edx/course-discovery/pull/1871

course_discovery/apps/api/v1/tests/test_views/test_search.py
inventhouse: Nearly drop-in, done.

dianakhuang (approved):

PR #1876, opened 2019-04-26
Add history table to Course

pwnage101:
This adds a migration which creates a course_metadata_historicalcourse
table.

DE-1422


bmedx (approved):
LGTM. An interesting tidbit in the PR you link to in the comments seems to indicate that removal of the history tables greatly sped up metadata refreshing. I'm not sure if that's a concern in re-adding them, but something we may need to be aware of if it causes issues.

PR #1877, opened 2019-04-26
add django-compressor support

noraiz-anwar:
https://openedx.atlassian.net/browse/PROD-135

Adding compressor support.
We need to make sure that any conditional addition of CSS/js resources is not inside `compress` tag as offline compression doesn't work that way. 
Offline compression is unable to find a compressed file if the content of the block is not the same as it was on compression time.


asadazam93 (approved):
Looks good to me

DawoudSheraz (approved):
Nice work on this one :+1:

PR #1879, opened 2019-04-26
Fix publishing to drupal by suppressing publication on initial save

Dillon-Dumesnil:
since that creates a new slug which we then copy over. Also switching
the marketing published condition to status=Reviewed or
status=Published so course teams are able to see the marketing page as
soon as the course is reviewed.

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -191,9 +191,7 @@ class CourseRunViewSet(viewsets.ModelViewSet):
         old_course_run = course.canonical_course_run
 
         # And finally, save to database and push to studio
-        course_run = serializer.save()
-        course_run.draft = True
-        course_run.save()
+        course_run = serializer.save(draft=True)
         self.push_to_studio(request, course_run, create=True, old_course_run=old_course_run)
 
         headers = self.get_success_headers(serializer.data)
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -191,9 +191,7 @@ def create(self, request, *args, **kwargs):
         old_course_run = course.canonical_course_run
 
         # And finally, save to database and push to studio
-        course_run = serializer.save()
-        course_run.draft = True
-        course_run.save()
+        course_run = serializer.save(draft=True)
```
mikix, line 194: That's legal?!

course_discovery/apps/course_metadata/models.py
```diff
@@ -583,6 +583,13 @@ def ensure_official_version(self, course_run):
 
         return official_course
 
+    def save(self, *args, **kwargs):
+        # We have to pop off suppress_publication because the base models.Model will treat it as an
+        # unexpected keyword arg. We actually use suppress_publication in the CourseRun model, but it
+        # is now passed in as a kwarg when saving in set_official_state to all models that are affected by that method.
+        kwargs.pop('suppress_publication', False)
```
mikix: Ah bummer. But good comment.

Anther option would be checking isinstance before passing `supress_publication`? I don't have a strong preference.

Or another option - is this something that could go into the DraftModelMixin? So if the model doesn't strip it, the mixin will?

mikix (approved):

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: Yessir! It's pretty cool. https://www.django-rest-framework.org/api-guide/serializers/#passing-additional-attributes-to-save

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: I think I like adding in an `isinstance` check because then it will at least keep all of the logic in one file and not having to reference multiple spots to see why things are happening. Will update now

PR #1881, opened 2019-04-26
ENT-1469 - Exposing course and course_run api fields in api

christopappas:
Exposing a few fields for course and course_run. Includes the following:

Course
- organizations

Course_run
- availability
- pacing type
- type (aka, seat type)

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1702,6 +1702,9 @@ class CourseSearchSerializer(HaystackSerializer):
                 'go_live_date': course_run.go_live_date,
                 'start': course_run.start,
                 'end': course_run.end,
+                'availability': course_run.availability,
+                'pacing_type': course_run.pacing_type,
+                'enrollment_mode': course_run.type,
             }
             for course_run in result.object.course_runs.all()
         ]
@@ -1720,6 +1723,7 @@ class CourseSearchSerializer(HaystackSerializer):
             'uuid',
             'subjects',
             'languages',
+            'organizations',
         )
 
 
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -1576,7 +1576,14 @@ class CourseSearchSerializerTests(TestCase, CourseSearchSerializerMixin):
 
     def test_data(self):
         request = make_request()
-        course = CourseFactory(subjects=SubjectFactory.create_batch(3))
+        organization = OrganizationFactory()
+        # 'organizations' in serialzed data should not return duplicate organization names
+        # Add the same organization twice to the course and make sure only one is in the serialized data
+        course = CourseFactory(
+            subjects=SubjectFactory.create_batch(3),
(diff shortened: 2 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1720,6 +1723,8 @@ class Meta:
             'uuid',
             'subjects',
             'languages',
+            'organizations',
+
```
Dillon-Dumesnil: Nit: Remove this empty line

Dillon-Dumesnil (approved):
Looks good. One small fixup request and then rebase

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -182,9 +182,9 @@ def prepare_subject_uuids(self, obj):
         return [str(subject.uuid) for subject in obj.subjects.all()]
 
     def prepare_languages(self, obj):
-        return [
+        return {
```
Dillon-Dumesnil, line 185: Are there actually duplicates for a course run's language? That feels odd, but I guess this keeps it from being surfaced. Feels like a bandaid on a bigger problem though, but not necessarily yours to fix.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1576,7 +1576,12 @@ class CourseSearchSerializerTests(TestCase, CourseSearchSerializerMixin):
 
     def test_data(self):
         request = make_request()
-        course = CourseFactory(subjects=SubjectFactory.create_batch(3))
+        organization = OrganizationFactory()
+        course = CourseFactory(
+            subjects=SubjectFactory.create_batch(3),
+            authoring_organizations=[organization],
```
Dillon-Dumesnil, line 1584: I might leave a comment somewhere around here explaining that you expect to only see one of these in the expected data because it forces a set.

Dillon-Dumesnil (approved):
So I don't have great context on everything happening here, but it looks mostly good. Seems a little odd that the fields you touched in the search_indexes both are being returned as sets when everything else is a list, but you know better than I do for the reasons for that.

course_discovery/apps/course_metadata/search_indexes.py
christopappas: While I was doing this, another ticket came in (ENT-1872) about multiple languages being returned for the course[run] metadata, so this change intends to fix that.

The prepare_languages logic touches every course_run, checking for its language, so my understanding is that there will be duplicates when there are handful of english course runs, but my assumption is there is a possibility of there being at least 1 non-english course_run (which is why the original implementation looped over all course runs in the first place)

course_discovery/apps/api/tests/test_serializers.py
christopappas: Can do

course_discovery/apps/course_metadata/search_indexes.py
Dillon-Dumesnil: Sounds good

PR #1883, opened 2019-05-01
DISCO-867: Compares original and new values before saving runs

jlajoie:


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -209,29 +209,47 @@ def update(self, request, *args, **kwargs):
...
             official_run.status = CourseRunStatus.Unpublished
             official_run.save()
 
-        # Sending draft=False triggers the review process for unpublished courses
-        draft = request.data.pop('draft', True)  # Don't let draft parameter trickle down
-        if not draft and course_run.status != CourseRunStatus.Published:
-            request.data['status'] = CourseRunStatus.LegalReview
+        if new_status:
+            course_run = serializer.save(status=new_status)
+        else:
```
staubina: Nit: you could probably not have an else here.  In no way a blocker.

```
if new_status:
  course_run = serializer.save(status=new_status)
course_run = serializer.save()
```

course_discovery/apps/course_metadata/data_loaders/analytics_api.py
```diff
@@ -2,10 +2,9 @@
 import logging
 
 import pytz
-
+from analyticsclient.client import Client
```
staubina: Was this just iSort?

staubina (approved):
Overall the changes look good.

I was trying to think of a way to quickly simplify the changes in course_runs.py, but we can iterate on it if we need to in the future.

course_discovery/apps/api/v1/views/course_runs.py
jlajoie: Don't want to run two saves if we don't need it!

course_discovery/apps/course_metadata/data_loaders/analytics_api.py
jlajoie: Yes :( not sure why my local complained about it, I might undo since I think it failed me now

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -209,29 +209,47 @@ def update(self, request, *args, **kwargs):
...
+            course_run = serializer.save()
```
mikix: This construction feels a little odd to me. I see why you don't want to pass status=None though.

What about having a `save_kwargs = {}` field that you then do `save_kwargs['status'] = CourseRunStatus.Unpublished` and `course_run = serializer.save(**save_kwargs)`?

Just a nit, this is fine regardless.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -209,29 +209,47 @@ def update(self, request, *args, **kwargs):
...
+        serializer.is_valid(raise_exception=True)
 
-        # If changes are made after review and before publish,
-        # revert status to unpublished.
-        if course_run.status == CourseRunStatus.Reviewed:
-            request.data['status'] = CourseRunStatus.Unpublished
+        # Only do the following if the object has been modified
+        new_status = None
+        # Sending draft=False triggers the review process for unpublished courses
+        draft = request.data.pop('draft', True)  # Don't let draft parameter trickle down
```
mikix: If we want to avoid `draft` trickling down, pop it before handing the data to the serializer.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -209,29 +209,47 @@ def update(self, request, *args, **kwargs):
...
+        if not draft and course_run.status != CourseRunStatus.Published:
+            new_status = CourseRunStatus.LegalReview
```
mikix: You moved this up. If you leave it below the other status changing block, you dont need the `not new_status` check below.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -209,29 +209,47 @@ def update(self, request, *args, **kwargs):
...
+
+        changed = False
+        for key, new_value in serializer.validated_data.items():
+            original_value = getattr(course_run, key, None)
+            # If the values aren't equal (base compare) and they both aren't None (additional compare for [] and None)
+            if new_value != original_value and not (
+                    isinstance(new_value, list) and new_value is None and original_value is None
```
mikix: This logic is hard to follow. Could an intermediate variable help reading clarity?

jlajoie: Yeah probably, can do so 👍

course_discovery/apps/api/v1/views/course_runs.py
jlajoie: It doesn't get serialized but this is a good call. (also I didn't write this :( )

course_discovery/apps/api/v1/views/course_runs.py
jlajoie: Big fan, happy to do so 👍

course_discovery/apps/api/v1/views/course_runs.py
mikix: I wrote the pop - and it was just because it is not serialized today. And we don't want to accidentally serialize it later, if someone adds draft to the serializer for reading purposes (we don't want to write it - just intercept it)

course_discovery/apps/api/v1/views/course_runs.py
staubina: Good catch, in my head there was a return.

mikix (approved):
:+1:

PR #1884, opened 2019-05-01
We want to allow updating entitlement mode and price prior to the cou…

Dillon-Dumesnil:
…rse being reviewed. This will mean we will have draft Audit entitlements, but we will ensure these are not created as official versions or in ecommerce.


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -249,7 +249,6 @@ def update(self, request, **kwargs):
             save_kwargs['status'] = CourseRunStatus.LegalReview
 
         course_run = serializer.save(**save_kwargs)
-
```
staubina: nit: You could save one file edit by removing this edit. /shrug

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -208,20 +206,21 @@ def update_entitlement(self, course, data, partial=False):
...
-            raise ValidationError(_('Existing entitlement not found for mode {0} in course {1}.')
-                                  .format(data['mode'], course.key))
+            raise ValidationError(_('Existing entitlement not found for course {0}.')
+                                  .format(course.key))
 
         # We have an entitlement object, now let's deserialize the incoming data and update it
         serializer = CourseEntitlementSerializer(entitlement, data=data, partial=partial)
         serializer.is_valid(raise_exception=True)
         serializer.save()
-        return entitlement
```
staubina, line 220: Why are we removing the return?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -234,9 +233,10 @@ def update_course(self, data, partial=False):
         serializer.is_valid(raise_exception=True)
 
         # First, update nested entitlements
-        entitlements = []
         for entitlement_data in entitlements_data:
-            entitlements.append(self.update_entitlement(course, entitlement_data, partial=partial))
+            self.update_entitlement(course, entitlement_data, partial=partial)
+        # We need to grab the entitlement in case we need to update the official version later
+        entitlement = CourseEntitlement.everything.filter(course=course, draft=True).first()
```
staubina: Why are we doing the get here again?

Can we avoid the extra get here?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -255,7 +255,14 @@ def update_course(self, data, partial=False):
             course.image.save('', image_data)
 
         # Then the course itself
-        serializer.save()
+        course = serializer.save()
+        if not draft:
+            official_course = set_official_state(course, Course)
+            if entitlement.mode != SeatType.objects.get(slug='audit'):
```
staubina: Do we have the slug as a constant?

I feel like it might be a good constant.

course_discovery/apps/course_metadata/migrations/0173_course_entitlement_unique_update.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.15 on 2019-05-01 17:07
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
staubina, line 8: Can we merge these migrations into one?

staubina (changes requested):
Take a look at my questions and comments and please let me know if you have questions.

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: Yeah, it was accidentally adding during a merge conflict. But it was also late and I didn't want to change and wait for tests again. Will update.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: We used to be adding it into a list that we did nothing with so I removed the return.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: So I tried to be able to just grab the result from the `update_entitlement` method, but that code is only run if `entitlements` are passed in as part of `data` and we allow not passing them in. We also have to do the get early and not within the `if not draft` because it doesn't seem to like the course that is returned from `serializer.save` when trying to do a filter.

I guess I could do a if we have entitlements grab that, else do the get. Does that sound better? @staubina

course_discovery/apps/course_metadata/migrations/0173_course_entitlement_unique_update.py
Dillon-Dumesnil: We could? I would argue keeping them separate since they aren't doing anything similar to one another, but if you have strong feelings, I'm willing to hear them out.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: Will update to Seat.AUDIT

course_discovery/apps/api/v1/views/courses.py
staubina: Ok, as long as no one else was expecting the return.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: >        if entitlements:
>            entitlement = entitlements[0]
>        else:
>            entitlement = CourseEntitlement.everything.filter(course=course, draft=True).first()

staubina: That seems sensible to me if anything to avoid hitting the database twice for the same data. I would like a second opinion on this one though.

Thoughts @jlajoie

course_discovery/apps/api/v1/views/courses.py
staubina: 👍

course_discovery/apps/course_metadata/migrations/0173_course_entitlement_unique_update.py
staubina: My only reason for merging was to keep the number of migrations down.  I also do not have a strong conviction on this one.  I see the logical split you mention now also.

Dillon-Dumesnil: Are there concerns about number of migrations? I figured it was a small load to apply them in general

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: I updated the code. It's objectively better where we still maintain not making any assumptions, and for the 80% case, we won't do another database hit.

course_discovery/apps/course_metadata/migrations/0173_course_entitlement_unique_update.py
staubina: The main concern is that the migrations over time become difficult to follow.  Neither of these migrations will have a severe performance hit, so feel free to keep them split.

staubina (approved):

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -185,16 +184,15 @@ def create(self, request, *args, **kwargs):
...
-            mode = SeatType.objects.get(slug=course_creation_fields['mode'])
-            CourseEntitlement.objects.create(
-                course=course,
-                mode=mode,
-                partner=partner,
-                price=price,
-                draft=True,
-            )
+        price = request.data.get('price', 0.00)
+        mode = SeatType.objects.get(slug=course_creation_fields['mode'])
```
jlajoie, line 188: Is the swap from using the conditional here just to fail this more loudly?

jlajoie: Nevermind saw the change down below to allow for "audit entitlements" on draft states

Dillon-Dumesnil: No. It's because we now allow creating Draft Audit entitlements so we don't need to filter on ENTITLEMENT_MODES anymore. We still prevent creation of Official Audit entitlements.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -67,12 +67,15 @@ def set_official_state(obj, model, attrs=None):
         obj.pk = official_obj.pk if official_obj else None  # pk=None will create it if it didn't exist.
         obj.draft = False
         obj.draft_version = draft_version
+        if isinstance(obj, Course):
+            obj.canonical_course_run = official_obj.canonical_course_run if official_obj else None
```
jlajoie, line 71: Should we fail this if there's no canonical course run?

course_discovery/apps/api/v1/views/courses.py
jlajoie: Yeah I'm fine with this, not sure there's a better option

jlajoie (approved):
Just the one question about failing loudly for missing canonical course runs, but I'm guessing that's fine since it'll only be for the non official case and wouldn't need it. LGTM 👍

course_discovery/apps/course_metadata/utils.py
Dillon-Dumesnil: We don't want that because when we are originally creating the official course, we pass in the canonical course run which is set later. It looks like this is also when we set the draft canonical course run, but we are planning on fixing that up.
Reference point: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L574

PR #1890, opened 2019-05-02
Updates to Course Create to validate Course number before creating

staubina:


course_discovery/apps/course_metadata/utils.py
```diff
@@ -235,6 +236,20 @@ def parse_course_key_fragment(fragment):
...
+    """
+    Verifies that the Course Number and Org do not contain invalid characters.
+    Args:
+        course_key_fragment: Partial CourseKey string without term
+    Returns:
+        bool: True if there are no invalid characters and raises a ValueError if there are invalid characters
+    """
+    org, course = parse_course_key_fragment(course_key_fragment)
+    if not CourseLocator.ALLOWED_ID_RE.match(org) or not CourseLocator.ALLOWED_ID_RE.match(course):
+        raise ValueError('Special characters not allowed in org or name.')
```
mikix: The common case is that the org will be fine (is that validated elsewhere? - shit will be wack if org has weird characters), but that the course number will be busted.

The endpoint is crafting the key from the org and number parts separately. Could we have a flow that checked just the number and gave a more targetted error message?

Also, we use the word "number" to the user for the course number, rather than "name"

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -148,7 +148,7 @@ def get_course_key(self, data):
     @writable_request_wrapper
     def create(self, request, *args, **kwargs):
         """
-        Create a Course, Course Entitlement, and Entitlement Product in E-commerce.
+        Create a Course, Course Entitlement, and Entitlement.
```
mikix, line 151: :+1:

course_discovery/apps/course_metadata/utils.py
staubina: 👍

course_discovery/apps/course_metadata/utils.py
```diff
@@ -242,6 +243,20 @@ def parse_course_key_fragment(fragment):
     return split[0], split[1]
 
 
+def validate_course_key_fragment(course_key_fragment):
+    """
+    Verifies that the Course Number and Org do not contain invalid characters.
```
mikix: `Verifies that a fragment of a course key does not contain invalid characters.` maybe?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -170,6 +170,9 @@ def create(self, request, *args, **kwargs):
         partner = request.site.partner
         course_creation_fields['partner'] = partner.id
         course_creation_fields['key'] = self.get_course_key(course_creation_fields)
+
+        validate_course_key_fragment(course_creation_fields['key'])
```
mikix: But why not just pass in `course_creation_fields['number']` and then you don't have to be creating/parsing the key just to end up with the number again.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -242,6 +243,20 @@ def parse_course_key_fragment(fragment):
...
+    Args:
+        course_key_fragment: Partial CourseKey string without term
+    Returns:
+        bool: True if there are no invalid characters and raises a ValueError if there are invalid characters
+    """
+    org, course = parse_course_key_fragment(course_key_fragment)
+    if not CourseLocator.ALLOWED_ID_RE.match(course):
+        raise ValueError('Special characters not allowed in Course Number.')
+    return True
```
mikix: Harmless, but feels odd having an unconditional return.

staubina: Would you prefer an else? 

I am fine either way.  Depends on the developer really.

In this case, seems safer.

mikix: No, not an else. I was suggesting just dropping the return altogether and have it be a `void` function, such as Python allows. No one that calls this will be checking the return value, they'll be catching or not.

course_discovery/apps/course_metadata/utils.py
staubina: Was just updating this.

mikix (approved):
Beautiful

PR #1895, opened 2019-05-07
DISCO-904: Swaps to set comparison for elements on transcript languages

jlajoie:


course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -481,6 +481,31 @@ def test_patch_put_reset_status(self):
...
+        response = self.client.put(url, {
+            'course': self.draft_course_run.course.key,  # required, so we need for a put
+            'start': self.draft_course_run.start,  # required, so we need for a put
+            'end': self.draft_course_run.end,  # required, so we need for a put
+            # Have the same number of elements, as the original array
+            'transcript_languages': ['en-us'],  # just any language tag to force a difference
+        }, format='json')
+        assert response.status_code == 200, "Status {}: {}".format(response.status_code, response.content)
+        self.draft_course_run.refresh_from_db()
+        draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
```
Dillon-Dumesnil, line 501: Why refresh_from_db and then to a get? Both are getting the updated version. I think you should just need the `self.draft_course_run.refresh_from_db()` and then switch the lower checks to self.draft_course_run

staubina (approved):

PR #1896, opened 2019-05-08
reorder default auth to favor JWT tokens

staubina:


jlajoie (approved):
15 files for one line change :(

Dillon-Dumesnil (approved):

douglashall:
Do we know why the response changed from a 403 to a 401?

staubina:
I believe 401 is the right response as all the tests were supposed to be testing users who were not authenticated.  I believe the JWT Auth is now appropriately catching that issue.

I am testing this more now locally and on Stage.

If you have a moment to stop by 9, I would be happy to discuss more.

PR #1897, opened 2019-05-08
Fixing issue where RCM will create an official version of a course ru…

Dillon-Dumesnil:
…n (and potentially course) since it exists in Studio, but not in discovery. It exists in Studio because we add it there in PublisherFE when we create the draft version of a course run.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -138,6 +138,12 @@ class CoursesApiDataLoader(AbstractDataLoader):
                         # Therefore, we don't need to do the statements below
                         course = self.update_course(course, body)
                 else:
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -138,6 +138,8 @@ def _process_response(self, response):
                         # Therefore, we don't need to do the statements below
                         course = self.update_course(course, body)
                 else:
+                    if CourseRun.everything.filter(key=course_run_id, draft=True).first():
```
staubina, line 145: I think a comment would be a good idea here. Something short maybe?

Dillon-Dumesnil: Added

staubina (approved):

PR #1900, opened 2019-05-13
Add dynamic field mixin to fields on course

bbaker6225:
Add mixin to nested program, subject serializer so we can drop fields from them on the course endpoint


jmyatt (approved):
🚢

PR #1901, opened 2019-05-14
Adding course run creation into the course endpoint so they can be wr…

Dillon-Dumesnil:
…apped in an atomic transaction


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -194,6 +196,13 @@ def create(self, request, *args, **kwargs):
             draft=True,
         )
 
+        # We want to create the course run here so it is captured as part of the atomic transaction
+        # and we can roll all of these changes back if the course run creation fails.
```
staubina: Nit: I think you could remove the part of the comment `and we can roll all of these changes back if the course run creation fails.`, AtomicTransaction implies it.  Not a blocker.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -194,6 +196,13 @@ def create(self, request, *args, **kwargs):
...
+        # Note: We have to send the request object as well because it is used for its metadata
+        # (like request.user and is set as part of the serializer context)
+        if course_run_creation_fields:
+            CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
```
staubina, line 204: I misunderstood your description, this looks fine to me.

staubina: I like that it follows the same path as the course run endpoint

👍

staubina (changes requested):
The changes seem straight forward, but I think we should add/update some testing as it will be a critical issue if this workflow is altered in a bad way.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: That was the goal! Just a bit annoying making it work.

staubina (approved):

PR #1905, opened 2019-05-15
Fix duplicate queries on courses endpoint

bbaker6225:
I'll probably have to change some tests, just want to put this up to get feedback that it makes sense.

Changes for 1 page  of results on the courses endpoint
Before:
Screen Shot 2019-05-15 at 3 52 07 PM
After:
Screen Shot 2019-05-16 at 11 28 58 AM

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -333,7 +333,7 @@ class MinimalPersonSerializer(serializers.ModelSerializer):
     @classmethod
     def prefetch_queryset(cls):
         return Person.objects.all().select_related(
-            'position__organization'
+            'position__organization',
         ).prefetch_related(
             'person_networks',
             'areas_of_expertise',
@@ -584,6 +584,12 @@ class NestedProgramSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
     """
     type = serializers.SlugRelatedField(slug_field='name', queryset=ProgramType.objects.all())
     number_of_courses = serializers.SerializerMethodField()
+
+    @classmethod
+    def prefetch_queryset(cls, queryset=None):
+        # Explicitly check for None to avoid returning all Programs when the
+        # queryset passed in happens to be empty.
+        return queryset if queryset is not None else Program.objects.all()
 
     class Meta:
         model = Program
@@ -892,6 +898,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
             'partner',
             'extra_description',
             '_official_version',
+            'canonical_course_run',
         ).prefetch_related(
             'expected_learning_items',
             'prerequisites',
@@ -947,10 +954,10 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
     course_runs = serializers.SerializerMethodField()
-    programs = serializers.SerializerMethodField()
+    programs = NestedProgramSerializer(read_only=True, many=True)
 
     @classmethod
-    def prefetch_queryset(cls, partner, queryset=None, course_runs=None):
+    def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=None):
         """
         Similar to the CourseSerializer's prefetch_queryset, but prefetches a
         filtered CourseRun queryset.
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -125,10 +125,16 @@ def get_queryset(self):
             if get_query_param(self.request, 'published_course_runs_only'):
                 course_runs = course_runs.filter(status=CourseRunStatus.Published)
 
+            if get_query_param(self.request, 'include_deleted_programs'):
```
jlajoie, line 128: Not a you thing but it feels so weird that we need to create a new querystring param every time we want to search for something that excludes a constant.

jlajoie (approved):

jlajoie (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -982,15 +992,6 @@ def get_course_runs(self, course):
             }
         ).data
 
-    # Executes as an n+1 query because of dependence on context
-    def get_programs(self, obj):
-        if self.context.get('include_deleted_programs'):
-            eligible_programs = obj.programs.all()
-        else:
-            eligible_programs = obj.programs.exclude(status=ProgramStatus.Deleted)
```
bbaker6225, line 990: Don't want to do an exclude after we've already prefetched

course_discovery/apps/api/serializers.py
```diff
@@ -333,7 +333,7 @@ class MinimalPersonSerializer(serializers.ModelSerializer):
     @classmethod
     def prefetch_queryset(cls):
         return Person.objects.all().select_related(
-            'position__organization'
+            'position__organization',
```
mikix, line 336: :fire:

course_discovery/apps/api/serializers.py
```diff
@@ -585,6 +585,12 @@ class NestedProgramSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
     type = serializers.SlugRelatedField(slug_field='name', queryset=ProgramType.objects.all())
     number_of_courses = serializers.SerializerMethodField()
 
+    @classmethod
+    def prefetch_queryset(cls, queryset=None):
+        # Explicitly check for None to avoid returning all Programs when the
+        # queryset passed in happens to be empty.
+        return queryset if queryset is not None else Program.objects.all()
```
mikix, line 592: nit: This construction feels slightly hard to read to me. Maybe it's the `not`. Maybe it's because what you're really doing here is "providing a default queryset" and then separately, you are "doing no prefetch". Maybe this would be clearer?

```
        if queryset is None:
            queryset = Program.objects.all()
        return queryset  # no prefetch
```

This sorta separates out those two ideas. And avoids `not`. I know you just copied this code from other methods, so not really your doing. Just a thought.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -248,6 +254,7 @@ def update_course(self, data, partial=False):
         # We need to grab the entitlement in case we need to update the official version later
         if entitlements:
             entitlement = entitlements[0]
+            course.entitlements = entitlements
```
mikix, line 257: Is this change related to the duplicate queries or fixing some bug?

mikix (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -961,17 +968,20 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None):
...
+            'topics',
             Prefetch('subjects', queryset=SubjectSerializer.prefetch_queryset()),
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
             Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
             Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('programs', queryset=NestedProgramSerializer.prefetch_queryset(queryset=programs)),
+            Prefetch('entitlements', queryset=CourseEntitlementSerializer.prefetch_queryset()),
         )
 
-    # Executes as an n+1 query because of dependence on context
```
bbaker6225, line 974: The course run duplicate queries we saw were from `canonical_course_run_key`, this is fine

course_discovery/apps/api/v1/views/courses.py
bbaker6225: The course entitlements weren't actually updating for me anymore because of something being cached in the serializer with the new prefetch on `entitlements`. I don't really understand why 🤷‍♂

course_discovery/apps/api/serializers.py
bbaker6225: I agree with you, and I like yours. I don't want to wait another 30 minutes for tests though, so I'll try and remember to put that in on the next discovery PR

PR #1906, opened 2019-05-15
Updating travisyml and openedx for compliance with py3.6

christopappas:


christopappas:
Python 3.5 comes with the ubuntu we're using to build the docker container for this service. Holding off on trying to upgrade now, as it might probably makes sense to wait until we can upgrade ubuntu

PR #1907, opened 2019-05-16
WL-1936 | Added estimated hours for course run in catalog list endpoint.

hasnain-naveed:


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1610,6 +1610,7 @@ def get_expected_data(cls, course, course_run, request):  # pylint: disable=unus
                 'availability': course_run.availability,
                 'pacing_type': course_run.pacing_type,
                 'enrollment_mode': course_run.type,
+                'estimated_hours': ((course_run.min_effort + course_run.max_effort) / 2) * course_run.weeks_to_complete,
```
ziafazal: we should use the `get_course_run_estimated_hours` method here too

ziafazal (changes requested):

ziafazal (approved):
👍

course_discovery/apps/course_metadata/utils.py
```diff
@@ -259,6 +259,20 @@ def validate_course_number(course_number):
         raise ValueError(_('Special characters not allowed in Course Number.'))
 
 
+def get_course_run_estimated_hours(course_run):
```
georgebabey, line 262: I assume this logic is pulled from some place else - is there any chance we can consolidate it or, if not, at least link to it. Just thinking if we ever want to update the algorithm we'd want to make sure we do so once (or at least in all of the places)

hasnain-naveed: This logic is in [story](https://openedx.atlassian.net/browse/WL-1936) description. I have asked a question on [ENT-1598](https://openedx.atlassian.net/browse/ENT-1598) about this whether we need estimated hours or all three fields (min_effort, max_effort, and weeks_to_complete).
@georgebabey

georgebabey (approved):

PR #1908, opened 2019-05-16
DISCO-923: Re-implements search without PkSearchableMixin for CourseRun

jlajoie:
Re-implementing previous state before changes were made in https://github.com/edx/course-discovery/pull/1838 as it broke performance of the query_contains endpoint.

Additional followup work is scheduled at https://openedx.atlassian.net/browse/DISCO-954 for `.search` performance.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -641,7 +641,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         query = 'title:Some random title'
         url = '{root}?q={query}'.format(root=reverse('api:v1:course_run-list'), query=query)
 
-        with self.assertNumQueries(14):
+        with self.assertNumQueries(44):
             response = self.client.get(url)
 
         actual_sorted = sorted(response.data['results'], key=lambda course_run: course_run['key'])
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -20,6 +20,7 @@ from course_discovery.apps.api.permissions import IsCourseRunEditorOrDjangoOrRea
 from course_discovery.apps.api.serializers import MetadataWithRelatedChoices
 from course_discovery.apps.api.utils import StudioAPI, get_query_param
 from course_discovery.apps.api.v1.exceptions import EditableAndQUnsupported
+from course_discovery.apps.core.utils import SearchQuerySetWrapper
 from course_discovery.apps.course_metadata.choices import CourseRunStatus
 from course_discovery.apps.course_metadata.constants import COURSE_RUN_ID_REGEX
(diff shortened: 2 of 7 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -86,7 +87,10 @@ def get_queryset(self):
             queryset = self.queryset
 
         if q:
-            queryset = CourseRun.search(q, queryset=queryset)
+            qs = SearchQuerySetWrapper(CourseRun.search(q).filter(partner=partner.short_code))
+            # This is necessary to avoid issues with the filter backend.
```
mikix, line 91: Can you add some specifics here? I barely know what this is talking about now, let alone a year from now.

Dillon-Dumesnil (approved):
LGTM. Maybe leave a comment with the PR this is partially reverting.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1026,6 +1026,24 @@ def availability(self):
...
+        """
+        query = clean_query(query)
+        queryset = SearchQuerySet().models(cls)
+
+        if query == '(*)':
+            # Early-exit optimization. Wildcard searching is very expensive in elasticsearch. And since we just
+            # want everything, we don't need to actually query elasticsearch at all.
+            return queryset.load_all()
+
+        return queryset.raw_search(query).load_all()
```
mikix, line 1045: Why not update PkSearchableMixin with this new version? Because it would require updating all callers?

course_discovery/apps/api/v1/views/course_runs.py
jlajoie: Ugh I _also_ don't know what this is talking about, but it's a manual copy + re-paste from the previous changes

course_discovery/apps/course_metadata/models.py
jlajoie: PkSearchableMixin is used already on Course.search and Program.search yeah, so I guess I could make a new mixin, but it's only used here... so doesn't seem much point if we're not hoping to re-use this

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -641,7 +641,7 @@ def test_list_query(self):
         query = 'title:Some random title'
         url = '{root}?q={query}'.format(root=reverse('api:v1:course_run-list'), query=query)
 
-        with self.assertNumQueries(14):
+        with self.assertNumQueries(44):
```
jlajoie, line 644: This is in agreement with the decrease we implemented as a part of https://github.com/edx/course-discovery/pull/1838/files#diff-79bab5eece6e39cdef00bd3f8f030461R464

Local queries against this endpoint in DDT produce minimal duplicates
Screen Shot 2019-05-20 at 11 25 30 AM

mikix (approved):
Alright, I wish we were fixing Programs and Courses too here, but I get that we want a fix for the pain point now.

PR #1911, opened 2019-05-20
Remove duplicate queries on programs list

bbaker6225:
Local results for 11 programs (all I have)

Before:
Screen Shot 2019-05-20 at 9 21 42 AM
After:
Screen Shot 2019-05-20 at 11 28 04 AM

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -105,7 +105,7 @@ class TestProgramViewSet(SerializationMixin):
         """ Verify the endpoint returns the details for a single program. """
         program = self.create_program()
 
-        with django_assert_num_queries(FuzzyInt(64, 2)):
+        with django_assert_num_queries(FuzzyInt(55, 2)):
             response = self.assert_retrieve_success(program)
         # property does not have the right values while being indexed
         del program._course_run_weeks_to_complete
@@ -142,13 +142,7 @@ class TestProgramViewSet(SerializationMixin):
             curriculum=curriculum
         )
 
-        # Notes on query count:
-        # 46 queries to get program with single curriculum (test_retrieve_basic_curriculum)
-        # +6 for first child program
-        # +1 for additional child program
-        # +8 for 1 or more first child program courses
-        # +3 for 1 or more additional child program courses
-        with django_assert_num_queries(FuzzyInt(69, 2)):
+        with django_assert_num_queries(FuzzyInt(61, 2)):
             response = self.assert_retrieve_success(parent_program)
         assert response.data == self.serialize_program(parent_program)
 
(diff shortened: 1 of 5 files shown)


bbaker6225:
Removed the `expires` logic since it's `NULL` on every entitlement in prod, after consulting with @staubina

jlajoie (approved):

staubina (approved):

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -142,13 +142,7 @@ def test_retrieve_curriculum_with_child_programs(self, django_assert_num_queries
             curriculum=curriculum
         )
 
-        # Notes on query count:
-        # 46 queries to get program with single curriculum (test_retrieve_basic_curriculum)
-        # +6 for first child program
-        # +1 for additional child program
-        # +8 for 1 or more first child program courses
-        # +3 for 1 or more additional child program courses
```
mikix, line 150: That was clearly written by someone that never had to update a test's query number before.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1350,12 +1350,17 @@ def is_program_eligible_for_one_click_purchase(self):
         applicable_seat_types = [seat_type.name.lower() for seat_type in self.type.applicable_seat_types.all()]
 
         for course in self.courses.all():
-            entitlement_products = set(course.entitlements.filter(mode__name__in=applicable_seat_types).exclude(
-                expires__lte=datetime.datetime.now(pytz.UTC)))
+            # Filter the entitlements in python, to avoid duplicate queries for entitlements after prefetching
+            all_entitlements = course.entitlements.all()
+            entitlement_products = {entitlement for entitlement in all_entitlements
+                                    if entitlement.mode.name.lower() in applicable_seat_types}
```
mikix: Heck yes set comprehension.

Lower seems harmless, but I'd be surprised if it's needed. Old code didn't do that.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1350,12 +1350,17 @@ def is_program_eligible_for_one_click_purchase(self):
...
             if len(entitlement_products) == 1:
                 continue
 
-            course_runs = set(course.course_runs.filter(status=CourseRunStatus.Published)) - excluded_course_runs
+            # Filter the course_runs in python, to avoid duplicate queries for course_runs after prefetching
+            all_course_runs = course.course_runs.all()
+            course_runs = {course_run for course_run in all_course_runs
+                           if course_run.status == CourseRunStatus.Published} - excluded_course_runs
```
mikix, line 1363: This construction (comprehension then a set operation on the same line) is a little hard to parse. I might have used a temporary variable like `published_runs` or something. But no biggie. Just a minor style nit.

mikix (approved):
Nice job commenting why it's in Python.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
bbaker6225: In theory it's nice, in practice everyone will bump the number and ignore the comment

course_discovery/apps/course_metadata/models.py
bbaker6225: I was surprised too, but apparently `mode__name__in=applicable_seat_types` seems to be doing case insensitive comparison under the hood since tests failed before this 🤷‍♂

mikix: Huh... maybe we should be using slug instead of name? That feels more "machine readable" than the name...

bbaker6225: Agreed, done

PR #1912, opened 2019-05-20
Schen/add org name override

schenedx:
@edx/masters-neem This is the PR to add the micromasters org name override to the degree data table.  With this change, we can override some rare case where, the linked micromasters program does not come from the same masters program org.

Please help review


kdmccormick (approved):
One change that may simplify logic downstream would be adding a property to the Degree model like:

```python
@property 
def micromasters_org_name(self):
   if self.micromasters_org_name_override:
        return self.micromasters_org_name_override
   if self.authoring_organizations:
        return self.authoring_organizations.first().name
   return None
```

and then just putting `micromasters_org_name` on the serializer, and always using that as the micromasters org name instead of doing if-logic on the marketing site.

But whether or not you decide to do that, this LGTM :+1:

schenedx:
@kdmccormick Interesting suggestion. I don't see much value of doing this here. But I captured your suggestion in https://openedx.atlassian.net/browse/EDUCATOR-4349. If we get to do that work, we should do what you suggested together with that ticket.

PR #1913, opened 2019-05-22
Validate database html fields

mikix:
Only allow simple html in the database, disallowing anything complicated or dangerous.

I have tried to look in edx.org's database for any weird tags or attributes that would be disallowed by this approach. I think the only thing we'll hit is https://www.edx.org/bio/water-faculty which has an embedded pdf. Which I think we want to disallow in future (and we'll only bother that bio if someone tries to edit it).

https://openedx.atlassian.net/browse/DISCO-612


course_discovery/apps/course_metadata/validators.py
```diff
@@ -0,0 +1,63 @@
...
+    }
+    ALLOWED_ATTRS = {  # attrs allowed for any tag
+        'align',  # rtl, formatting
+        'dir',
+        'lang',
+        'style',  # allow for now, but we might want to reconsider this
+        'typeof',  # a harmless semantic-web RDF thing, seen in the wild in our database
+    }
+    ALLOWED_TAG_ATTRS = {
+        'a': {'href', 'rel', 'target', 'title'},
```
staubina, line 20: We should reach out the PCs to see if we want to allow links.  This is not related directly to this PR, just something to think about again soon.

course_discovery/apps/course_metadata/validators.py
```diff
@@ -0,0 +1,63 @@
...
+        allowed_attrs = self.ALLOWED_ATTRS | self.ALLOWED_TAG_ATTRS.get(tag, frozenset())
+        unknown_attrs = attr_keys - allowed_attrs
+        if unknown_attrs:
+            self.error(_('{0} attribute is not allowed on the {1} tag').format(unknown_attrs.pop(), tag))
+
+    def handle_endtag(self, tag):
+        if tag not in self.ALLOWED_TAGS:
+            self.error(_('{0} tag is not allowed').format(tag))
+
+    def handle_comment(self, data):
```
staubina, line 48: Are these just helper methods we do not think we need included in the Parser, we are overriding?

staubina (commented):
The changes seem straightforward, Just a couple of questions

* Should we reach out to the PCs about the embeded tags before this work?
* What will the impact of these model changes be on Old Publisher?
* Should we remove links from the Rich editors and strip them out?

course_discovery/apps/course_metadata/tests/test_validators.py
```diff
@@ -0,0 +1,27 @@
+import ddt
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+
+from course_discovery.apps.course_metadata.validators import validate_html
+
+
+@ddt.ddt
+class TestHtmlValidator(TestCase):
+    @ddt.data('', None, '<p>Hello World</p>', '<p>')
```
Dillon-Dumesnil: Doesn't a non-closed tag result in broken HTML for the page? I feel like we might want to prevent that if possible.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -41,12 +41,14 @@
 # pylint: disable=no-member
 
 @pytest.mark.django_db
-class TestCourse:
+@pytest.mark.usefixtures('haystack_default_connection')
+@ddt.ddt
+class TestCourse(TestCase):
```
Dillon-Dumesnil, line 46: Can we add in a test somewhere in this file for the happy case and make sure it doesn't complain? I don't think the factory created fields will ever include HTML tags.

Dillon-Dumesnil (approved):
Nothing that's blocking, but couple of small comments. Looks good!

course_discovery/apps/course_metadata/validators.py
mikix: These are overridden HTMLParser methods. The default versions do nothing. But I figured I'd block anything that wasn't just basic HTML. Comments aren't likely to be an attack vector, but still. Things that aren't visible seem odd to include.

mikix:
@staubina 

> Should we reach out to the PCs about the embeded tags before this work?

I hadn't planned to, but I have just now dropped a note to Julia. They only used it once for a single pdf. I'm guessing it's not a common workflow for them.

> What will the impact of these model changes be on Old Publisher?

Hopefully nothing if I was correctly careful about what tags are being used today.

For course descriptions, Old Publisher uses rich text editors. So nothing weird should be coming in from there.

For person bios, Old Publisher just has a text field that you enter HTML into. Which is probably how we got into the embed situation. So the user will see an error if they try to enter anything weird there from now on, even on Old Publisher.

But I'd say that's a feature not a bug.

> Should we remove links from the Rich editors and strip them out?

This was feedback from the partner tests right? But Old Publisher has links in the rich text editors...  See https://discovery.stage.edx.org/publisher/courses/3837/edit/

So is this a new request by PCs?

course_discovery/apps/course_metadata/tests/test_validators.py
mikix: HTML5 is a pretty generous standard. Any modern browser would just implicitly close that p tag next time a block level tag appeared.

The parser I'm using doesn't try to validate the HTML against a spec at all (like matching opening and closing tags). It just parses.

Since my interest was mostly around security, and odd HTML was likely not coming in from our rich text editors, I was planning on leaving spec-validation stuff alone.

course_discovery/apps/course_metadata/tests/test_models.py
mikix: Fair, can do.

course_discovery/apps/course_metadata/validators.py
mikix: We can and should do that. But I also want to draw a distinction between what the database allows and what we allow as part of the UI.

It's possibly reasonable for the database to allow links, even if we don't expose that as part of the publisher rich text editors.

staubina: 👍

staubina (approved):

PR #1917, opened 2019-05-23
Try removing cacheresponsemixin from pathways

bbaker6225:
Because why not

This endpoint seems to rarely be used, and when it is used it doesn't get very high throughput, so it seems reasonable to test dropping this.


bbaker6225:
Reviewed in person by Jeff

PR #1921, opened 2019-05-24
Removing unnecessary validation error now that we allow in progress s…

Dillon-Dumesnil:
…aves


mikix (approved):

Dillon-Dumesnil:
On second thought, should I leave this so we continue to support this functionality in current publisher since it is just in the validator for the course run?

mikix:
Was this was code we added? If so, I'd say it's fine to drop.

Dillon-Dumesnil:
Oh hey, turns out you wrote it. Guess we are good to drop. Sweet! 👍

PR #1924, opened 2019-05-29
Move course_discovery defaults from configration repo

syedimranhassan:


course_discovery/settings/base.py
```diff
@@ -573,3 +573,180 @@
...
+
+# See edx_django_service_automated_users for an example of what this should be
+DISCOVERY_AUTOMATED_USERS = {}
+
+DISCOVERY_CSRF_COOKIE_SECURE = False
+DISCOVERY_CORS_ORIGIN_WHITELIST = []
+
+DISCOVERY_ENABLE_NEWRELIC_DISTRIBUTED_TRACING = False
+
+discovery_service_config_overrides = {
```
coryleeio: The values under discovery_service_config_overrides are the only ones that get injected to the application environment. aka discovery.yml

the intent of this ticket is that the 
values that are in discovery.yml have their default values set in the application, not in ansible.

I think you would want to only keep the contents of the 
discovery_service_config_overrides

section as top level keys

For example say that DISCOVERY_EMAIL_USE_TLS is set to true.

instead of 
discovery_service_config_overrides = {
   EMAIL_USE_TLS: DISCOVERY_EMAIL_USE_TLS
...
}

you would just put:
EMAIL_USE_TLS: true

We are trying to kill the 'config prefixing'

so there shouldn't be any DISCOVERY_MY_VAR type variables, only MY_VAR

(The values that get read in from the yml AFTER the discovery.yml gets rendered)

coryleeio (changes requested):

course_discovery/settings/base.py
coryleeio: You will also want to de-dupe those top level variables

syedimranhassan: Hi,
Thanks @corylee. I have set default values of variables in base.py that are in discovery.yml using default.yml. But there are some variable in discovery.yml that are not in default.yml or base.py. I think they are coming from internal or secure. So, should they also be moved to base.py?

coryleeio: No, just the defaults - the idea is that we are going to remove the defaults from configuration soon, so we want to copy them into the application repo so that developers can own the defaults for their application.

Ideally this PR does not change the behavior of the application, even for upstream users, so we don't want to set any defaults that are not already set, or set any defaults that were not set in configuration.

course_discovery/settings/base.py
```diff
@@ -163,6 +163,7 @@
 
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#media-url
 MEDIA_URL = '/media/'
+DEFAULT_FILE_STORAGE = "django.core.files.storage.FileSystemStorage"
```
coryleeio: This value was under 
DISCOVERY_MEDIA_STORAGE_BACKEND

which eventually (if you trace it through edx_django_service and the like resolves to:
MEDIA_STORAGE_BACKEND

which already has a default of {}, so I think this can be deleted


I dont see it being used anywhere

course_discovery/settings/base.py
```diff
@@ -271,6 +272,7 @@
 SOCIAL_AUTH_EDX_OIDC_URL_ROOT = 'replace-me'
 SOCIAL_AUTH_EDX_OIDC_LOGOUT_URL = 'replace-me'
 SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY = SOCIAL_AUTH_EDX_OIDC_SECRET
+SOCIAL_AUTH_REDIRECT_IS_HTTPS = False
```
coryleeio, line 275: I think what you did here is correct, we need to get all the values from under 

edx_django_service_config_default
LANGUAGE_CODE
SECRET_KEY
TIME_ZONE
STATIC_ROOT
etc


and default them to whatever they are currently defaulted too for discovery

coryleeio (changes requested):

course_discovery/settings/base.py
coryleeio: You need to include all these values as well:
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml#L182-L222

as they get merged into the edx_django_service_config here:
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml#L226

course_discovery/settings/base.py
```diff
@@ -168,7 +168,7 @@
 
 # STATIC FILE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#static-root
-STATIC_ROOT = root('assets')
+STATIC_ROOT = "/edx/app/NOT-SET/staticfiles"
```
coryleeio: Two more little things:

Why was the STATIC_ROOT changed?
Can you move MEDIA_STORAGE_BACKEND into here and remove the comment above it so that all the defaults live here that used to live in configuration.

coryleeio (changes requested):

course_discovery/settings/base.py
syedimranhassan: I have added all these varaiables and also I have moved databases to local.py and copy caches from local.py to base.py

course_discovery/settings/base.py
```diff
@@ -109,20 +109,6 @@
 # Python dotted path to the WSGI application used by Django's runserver.
 WSGI_APPLICATION = 'course_discovery.wsgi.application'
 
-# Database
-# https://docs.djangoproject.com/en/1.8/ref/settings/#databases
-# Set this value in the environment-specific files (e.g. local.py, production.py, test.py)
```
coryleeio: I want to have both the DATABASES entry and the CACHES to both bein base.py or both not be, i'd say put DATABASES back and add CACHES.

course_discovery/settings/local.py
```diff
@@ -17,12 +17,12 @@
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': root('default.db'),
+        'ENGINE': 'django.db.backends.',
```
coryleeio: I would revert this part, we dont want to change their local dev settings.

coryleeio (changes requested):
2 more changes - then I think this is ready to go.

coryleeio (approved):
This looks good to me

coryleeio:
:shipit:

PR #1926, opened 2019-05-30
DISCO-956: Adds CourseEditor create to Course create

ciduarte:


mikix (approved):
:1st_place_medal: First edX PR!

PR #1927, opened 2019-05-30
DISCO-983: Adding Seats to Discovery on Course Run creation and updat…

Dillon-Dumesnil:
…ing Seats on saves and publishes


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -1,5 +1,7 @@
+import datetime
```
bbaker6225: Do we need to use this library?

.coveragerc
```diff
@@ -4,6 +4,7 @@ data_file = .coverage
 source=course_discovery
 concurrency=multiprocessing
 omit =
+    *
```
jlajoie: This one seems aggressive?

Dillon-Dumesnil: Oops. I add that in during testing so I don't have the coverage report taking up my whole console. Removed

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -151,6 +162,12 @@ def test_create_minimum(self):
         self.assertEqual(str(new_course_run.end), '2001-01-01 00:00:00+00:00')  # spot check that input made it
         self.assertTrue(new_course_run.draft)
 
+        new_seat = Seat.everything.get(course_run=new_course_run)
+        self.assertEqual(Seat.everything.filter(course_run=new_course_run).count(), 1)
```
mikix: This is redundant with a `get()` call, but is also harmless to have.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -206,6 +211,23 @@ def create_run_helper(self, run_data, request=None):
...
+            )
+            if mode in Seat.REQUIRES_AUDIT_SEAT:
+                Seat.objects.create(
+                    course_run=course_run,
+                    type='audit',
+                    price=0.00,
+                    draft=True,
+                )
+        else:
+            raise ValidationError(_('Seat mode [{}] is not valid.').format(mode))
```
mikix: nit: I generally prefer to have early exits like this on their own. So that you don't have to indent the main path. Like so:

```
         if mode not in Seat.SEAT_TYPES:
            raise ValidationError(_('Seat mode [{}] is not valid.').format(mode))

         <NORMIE CODE>
```

Also, might be worth moving this into a `create_seats` helper method, like you did for `update_seats`? Although actually, `update_seats` does an update_or_create... could you just reuse it somehow?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,12 +247,47 @@ def create(self, request, *args, **kwargs):
...
+                    course_run=course_run,
+                    type='audit',
+                    draft=True,
+                    defaults={
+                        'price': 0.00,
+                        'upgrade_deadline': None,
+                    },
+                )
+        else:
+            raise ValidationError(_('Seat mode [{}] is not valid.').format(mode))
```
mikix: Same error path / happy path comment.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -244,6 +301,10 @@ def update(self, request, **kwargs):
         serializer = self.get_serializer(course_run, data=request.data, partial=partial)
         serializer.is_valid(raise_exception=True)
 
+        # First update the Seat(s) if the course run isn't already published
+        if course_run.status != CourseRunStatus.Published:
+            self.update_seats(mode, price, course_run)
```
mikix: Do we want to do anything if the price & mode don't match current values, but we're already published? Current code looks like it will silently ignore that request, which might be misleading. Maybe it doesn't super matter, if the UI doesn't let us get in that situation.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1235,7 +1242,7 @@ class Seat(DraftModelMixin, TimeStampedModel):
     # TODO Replace with FK to SeatType model
     type = models.CharField(max_length=63, choices=SEAT_TYPE_CHOICES)
     price = models.DecimalField(**PRICE_FIELD_CONFIG)
-    currency = models.ForeignKey(Currency)
+    currency = models.ForeignKey(Currency, default='USD')
```
mikix, line 1311: Whoa, you can specify a default primary key lookup like that? Makes sense, but neat. I guess we don't usually have non-ID primary keys.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1096,19 +1096,24 @@ def __str__(self):
...
             # If there's an official course, associate the new one with it.
             # Otherwise, create it.
             course = self.course.ensure_official_version(official_version)
             official_version.course = course
             official_version.save()
 
+        # Seats should be updated every time a course moves into Reviewed and this
+        # function is called everytime that happens.
+        for seat in draft_version.seats.all():
+            set_official_state(seat, Seat, {'course_run': official_version})
```
mikix, line 1170: It feels like you're taking advantage of a side effect here. The previous promise of this `ensure_official_version` is that it's a no-op if a draft exists. That's no longer true. You're using a `please make sure the DB is sane by migrating old data` check to become a `also move data over from draft`

HOWEVER - is our code broken right now? Do we only copy data over the first time we become reviewed? It looks like we only ever call `ensure_official_version` to do the copy... Hmph. I think someone looked at the `ensure_draft_version` which *was* a `make the DB sane for migration purposes` check and assumed that `ensure_official_version` would be same vibe. But it's not, it's should just have a `always copy draft to official` vibe.

We should... fix that, yeah? Don't know if it makes sense in your PR or not. But looks like we want to take your blurred line and just destroy the line...

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: I was thinking no because right now, we send a patch request for every course run on every save and we don't want to update the Seat prices because of a change in the entitlement (which is where we are currently getting our mode/price from).

I wrote this code thinking that once a course run is published, we shouldn't really be changing the Seat prices for it, but we can still leave the ability to change the entitlement cost and have that apply to future runs without changing the older ones.

Dillon-Dumesnil: @transaction.atomic
    def _update_seats_from_entitlement(self, course, entitlement, changed_by):
        for run in self._get_active_course_runs(course):
            run.seats.all().delete()
            seat = Seat(
                type=CourseEntitlement.MODE_TO_SEAT_TYPE_MAPPING[entitlement.mode],
                price=entitlement.price,
                currency=entitlement.currency
            )

            # Use the SeatForm here to not duplicate logic for creating seats
            seat_form = SeatForm(instance=seat)
            seat_form.save(commit=True, course_run=run, changed_by=changed_by)

Dillon-Dumesnil: Based on this function in current publisher, we should continue to update seats even after publishing. I will update.

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: Pretty cool stuff :sunglasses:

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -205,6 +223,36 @@ def test_create_draft_ignored(self, draft):
         self.assertDictEqual(response.data, self.serialize_course_run(new_course_run))
         self.assertTrue(new_course_run.draft)
 
+    @responses.activate
+    def test_create_with_verified_seat(self):
+        """ Verify the endpoint supports creating a course_run, but always as a draft. """
+        course = self.draft_course_run.course
+        new_key = 'course-v1:{}+1T2000'.format(course.key.replace('/', '+'))
```
mikix: Ugh, I hate this stuff. I kinda wish we just fixed our factories to generate new style formats and added a test somewhere that we still handle old. But that's never something I want to take on when I'm doing something else. And this is not me asking you to, I'm just griping.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -205,6 +223,36 @@ def test_create_draft_ignored(self, draft):
...
+            'price': 77,
+        }, format='json')
+
+        self.assertEqual(response.status_code, 201)
+        new_course_run = CourseRun.everything.get(key=new_key)
+        self.assertDictEqual(response.data, self.serialize_course_run(new_course_run))
+        self.assertTrue(new_course_run.draft)
+
+        # When creating a verified Seat, an audit Seat should also be created
+        self.assertEqual(Seat.everything.filter(course_run=new_course_run).count(), 2)
```
mikix: Super-nit: You could probably drop the course_run= bit, since these are only seats in database (here and in other tests).

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,12 +232,47 @@ def create(self, request, *args, **kwargs):
         headers = self.get_success_headers(serializer_data)
         return Response(serializer_data, status=status.HTTP_201_CREATED, headers=headers)
 
+    def update_or_create_seats(self, mode, price, course_run):
+        if mode not in Seat.SEAT_TYPES:
+            raise ValidationError(_('Seat mode [{}] is not valid.').format(mode))
```
mikix: Do we have validation somewhere around a negative price? I feel like we did for entitlements, but maybe not seats. Not a high priority, but this made me think of that.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,12 +232,47 @@ def create(self, request, *args, **kwargs):
...
+
+        deadline = None
+        if mode == Seat.AUDIT or mode == Seat.PROFESSIONAL:
+            Seat.everything.filter(draft=True, course_run=course_run).exclude(type=mode).delete()
+        elif mode == Seat.VERIFIED:
+            Seat.everything.filter(draft=True, course_run=course_run, type='professional').delete()
```
mikix: This might be more futureproof as `.exclude(mode).exclude(audit)` - should also probably be:

```
if mode in Seat.REQUIRES_AUDIT_SEAT:
 ...
else:
 ...
```

Rather than so explicit about naming the types and leaving a hanging elif without an else.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,12 +232,47 @@ def create(self, request, *args, **kwargs):
         headers = self.get_success_headers(serializer_data)
         return Response(serializer_data, status=status.HTTP_201_CREATED, headers=headers)
 
+    def update_or_create_seats(self, mode, price, course_run):
+        if mode not in Seat.SEAT_TYPES:
+            raise ValidationError(_('Seat mode [{}] is not valid.').format(mode))
+
+        deadline = None
+        if mode == Seat.AUDIT or mode == Seat.PROFESSIONAL:
+            Seat.everything.filter(draft=True, course_run=course_run).exclude(type=mode).delete()
```
mikix: Note to self - do we handle the draft->official well for mode changes?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,12 +232,47 @@ def create(self, request, *args, **kwargs):
...
+        elif mode == Seat.VERIFIED:
+            Seat.everything.filter(draft=True, course_run=course_run, type='professional').delete()
+            deadline = course_run.end - datetime.timedelta(days=settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
+            deadline = deadline.replace(hour=23, minute=59, second=59, microsecond=99999)
```
mikix: Can we just compare < rather than <= and avoid this trick? Or is that comparison buried in ecommerce logic?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -244,6 +286,9 @@ def update(self, request, **kwargs):
         serializer = self.get_serializer(course_run, data=request.data, partial=partial)
         serializer.is_valid(raise_exception=True)
 
+        if mode or price:
```
mikix: This is a little tricky. Price can be validly falsy at 0.00. Plus, the `or` isn't quite right with the pops above. Like, if I PATCH mode='verified' but don't pass in price, you'll reset my price on me, even though I didn't ask for that to be patched.

Is this better?
```
        update_seats = 'mode' in request.data or 'price' in request.data
        mode = request.data.pop('mode', course_run.mode)
        price = request.data.pop('price', course_run.price)
        if update_seats:
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -609,25 +609,31 @@ def first_enrollable_paid_seat_price(self):
...
+    def update_or_create_official_version(self, course_run=None):
+        draft_version = Course.everything.get(pk=self.pk)
+        if draft_version.official_version:
+            official_version = set_official_state(draft_version, Course)
+            for entitlement in draft_version.entitlements.all():
+                # The draft version could have audit entitlements, but we only
+                # want to create official entitlements for the valid entitlement modes.
+                if entitlement.mode.slug in Seat.ENTITLEMENT_MODES:
+                    set_official_state(entitlement, CourseEntitlement, {'course': official_version})
+            return official_version
```
mikix: You already commented on and had a plan for the duplication of code in this function. Just mentioning it here for recordkeeping.

course_discovery/apps/course_metadata/models.py
```diff
@@ -609,25 +609,31 @@ def first_enrollable_paid_seat_price(self):
...
-        official_course.save()
+    def update_or_create_official_version(self, course_run=None):
+        draft_version = Course.everything.get(pk=self.pk)
+        if draft_version.official_version:
+            official_version = set_official_state(draft_version, Course)
+            for entitlement in draft_version.entitlements.all():
+                # The draft version could have audit entitlements, but we only
+                # want to create official entitlements for the valid entitlement modes.
+                if entitlement.mode.slug in Seat.ENTITLEMENT_MODES:
+                    set_official_state(entitlement, CourseEntitlement, {'course': official_version})
```
mikix: What about existing official entitlements with a different mode? I guess we should modify/delete those? Same Q for seats.

mikix (commented):
Looks good overall! Thanks for taking on the expanded update_official work!

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: Will write a ticket so this PR doesn't go crazy

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: This was copied over from the Publisher Seat model and I'm not sure where the comparison is actually happening. I agree about the code smell, but without having the time to invest in looking where it's used, I'm just going to leave it for now.

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: I have added in code for raising a ValidationError on non allowed mode changes to both Seats and Entitlements. Essentially this covers changing from anything but audit -> verified after creation of products.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1094,21 +1098,28 @@ def search(cls, query):
...
-            official_version.slug = draft_version.slug
-            for seat in draft_version.seats.all():
-                set_official_state(seat, Seat, {'course_run': official_version})
+    def update_or_create_official_version(self):
+        draft_version = CourseRun.everything.get(pk=self.pk)
+        creating = False
+        # If there isn't already an official_version linked to the draft_version, then
+        # this is our first time.
+        if not draft_version.official_version:
+            creating = True
```
mikix: `creating  = not draft_version.official_version`

bbaker6225: 🔥

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,12 +232,69 @@ def create(self, request, *args, **kwargs):
...
+            Seat.everything.filter(draft=True, course_run=course_run).exclude(type=mode).delete()
+            Seat.everything.update_or_create(
+                course_run=course_run,
+                type=mode,
+                draft=True,
+                defaults={
+                    'price': price,
+                    'upgrade_deadline': deadline,
+                },
+            )
```
mikix: We talked about this in person, but for reference: we should un-indent this block or move it out or something.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -232,6 +236,15 @@ def update_entitlement(self, course, data, partial=False):
...
                                   .format(course.key))
 
+        # We want to allow upgrading an entitlement from Audit -> Verified, but allow no other
+        # mode changes. We use the official version existing as an indicator for ecom products
+        # having already been created.
+        mode_switch_whitelist = {'audit': 'verified'}
+        if course.official_version and entitlement.mode != mode and \
+                mode_switch_whitelist.get(entitlement.mode.slug) != mode.slug:
+            raise ValidationError(_('Switching entitlement modes after being reviewed is not supported. Please reach '
+                                    'out to your coordinator for additional help if necessary.'))
```
mikix: "project coordinator"

course_discovery/apps/course_metadata/models.py
```diff
@@ -609,25 +609,29 @@ def first_enrollable_paid_seat_price(self):
...
-    def ensure_official_version(self, course_run):
-        if self.official_version:
-            return self.official_version
+    def update_or_create_official_version(self, course_run=None):
+        draft_version = Course.everything.get(pk=self.pk)
+        creating = False
+        # If there isn't an official_version set yet, then this is a create instead of update
+        # and we will want to set additional attributes.
+        if not draft_version.official_version:
+            creating = True
```
mikix: `creating = not draft_version.official_version` again

mikix (approved):
I think all my comments have been addressed. Thanks so much for working through the twists and turns of this PR!

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,6 +229,68 @@ def create(self, request, *args, **kwargs):
...
+            type=mode,
+            draft=True,
+            defaults={
+                'price': price,
+                'upgrade_deadline': deadline,
+            },
+        )
+        seats = [seat]
+        if mode in Seat.REQUIRES_AUDIT_SEAT or mode == Seat.AUDIT:
+            Seat.everything.filter(draft=True, course_run=course_run)\
```
staubina: nit: Please avoid using the \ for new lines.

PEP8 https://www.python.org/dev/peps/pep-0008/#maximum-line-length recommends using the \ only when there is no other option.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,6 +229,68 @@ def create(self, request, *args, **kwargs):
...
+                type=Seat.AUDIT,
+                draft=True,
+                defaults={
+                    'price': 0.00,
+                    'upgrade_deadline': None,
+                },
+            )
+            seats.append(audit_seat)
+        else:
+            Seat.everything.filter(draft=True, course_run=course_run).exclude(type=mode).delete()
```
staubina: What is the goal of the deletes?

It looks like it is clean up for when the mode changes, am I interpreting it correctly?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -232,6 +237,15 @@ def update_entitlement(self, course, data, partial=False):
             raise ValidationError(_('Existing entitlement not found for course {0}.')
                                   .format(course.key))
 
+        # We want to allow upgrading an entitlement from Audit -> Verified, but allow no other
+        # mode changes. We use the official version existing as an indicator for ecom products
+        # having already been created.
+        mode_switch_whitelist = {'audit': 'verified'}
+        if course.official_version and entitlement.mode != mode and \
```
staubina: Same as above, use a parenthesis here not a slash please. (PEP-8)

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -232,6 +237,15 @@ def update_entitlement(self, course, data, partial=False):
             raise ValidationError(_('Existing entitlement not found for course {0}.')
                                   .format(course.key))
 
+        # We want to allow upgrading an entitlement from Audit -> Verified, but allow no other
+        # mode changes. We use the official version existing as an indicator for ecom products
+        # having already been created.
+        mode_switch_whitelist = {'audit': 'verified'}
```
staubina: I recommend using the constants here for AUDIT and VERIFIED

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -257,12 +271,11 @@ def update_course(self, data, partial=False):
...
             entitlements.append(self.update_entitlement(course, entitlement_data, partial=partial))
-        # We need to grab the entitlement in case we need to update the official version later
         if entitlements:
-            entitlement = entitlements[0]
             course.entitlements = entitlements
-        else:
-            entitlement = CourseEntitlement.everything.filter(course=course, draft=True).first()
+            # If entitlements were updated, we also want to update seats
+            for course_run in course.course_runs.filter(end__gt=datetime.now()):
+                CourseRunViewSet().update_or_create_seats(course_run)
```
staubina: I feel like the right home for` update_or_create_seats()` might be in the model or in a utility file, not as part of the View code.  I am going to look closer at the method though.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -225,6 +229,68 @@ def create(self, request, *args, **kwargs):
         headers = self.get_success_headers(serializer_data)
         return Response(serializer_data, status=status.HTTP_201_CREATED, headers=headers)
 
+    def update_or_create_seats(self, course_run):
```
staubina: I think a Docstring might be useful for this method as you seem to be adding comments where you use it a lot.  You could condense those comments into a clear, concise method docstring.

staubina: Also, I believe this method might be better housed in the model.  All of the action in this method seemed to be Database related.  Is there another reason to housed it in the ViewSet code?

staubina (changes requested):
The first pass is done.  I need to a deeper look at the logic to review any further, let me know if you think that is necessary with all the reviews you are getting.

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: Yeah, basically. Just want to make sure we are only keeping track of the Seats that should exist. Like Professional and Audit should both exist alone and Verified has both so if you're switching from Professional to Audit or Verified, we just need to delete the professional seat for you.

staubina (approved):

PR #1928, opened 2019-05-31
Update tests to account for flaky query counts

bbaker6225:


jlajoie (approved):
Not my favorite pattern that we have, but I'm fine with adding this

PR #1929, opened 2019-05-31
Send emails when a course run state changes

mikix:
For the new publisher workflow, we want to send out state-change notifications like old publisher did. So whenever a CourseRun is saved, we do so using new email templates.

https://openedx.atlassian.net/browse/DISCO-958


course_discovery/apps/course_metadata/emails.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def get_project_coordinator(org):
+    # Model imports here to avoid a circular import
+    from course_discovery.apps.publisher.models import OrganizationUserRole
+
+    if not org:
+        return None
+
+    role = OrganizationUserRole.objects.filter(organization=org,
```
staubina, line 60: Do we run the risk here of only getting 1 role of many? How common is it to have multiple roles?

course_discovery/apps/course_metadata/emails.py
```diff
@@ -0,0 +1,162 @@
...
+    course = course_run.course
+    partner = course.partner
+    org = course.authoring_organizations.first()
+    project_coordinator = project_coordinator or get_project_coordinator(org)
+    if not project_coordinator:
+        log_missing_project_coordinator(course_run, org, template_name)
+        return
+
+    publisher_url = partner.publisher_url
+    if not publisher_url:
```
staubina, line 86: I wonder if in the future we should send alerts on this issue?

course_discovery/apps/course_metadata/emails.py
```diff
@@ -0,0 +1,162 @@
...
+        'recipient_name': recipient_name,
+        'platform_name': settings.PLATFORM_NAME,
+        'org_name': org.name,
+        'contact_us_email': project_coordinator.email,
+        'course_page_url': review_url,
+        'course_approval_url': approve_url,
+        'studio_url': run_studio_url,
+    })
+
+    txt_template = template_name + '.txt'
```
staubina, line 119: Nit: I think it might be wise to make .txt and .html constants.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -0,0 +1,162 @@
...
+
+    editors = CourseEditor.course_editors(course_run.course)
+    send_email(course_run, template_name, subject, editors, _('course team'), context=context)
+
+
+def send_email_for_legal_review(course_run):
+    """ Send email when a course run is submitted for legal review.
+
+        Arguments:
+            course_run (Object): CourseRun object
```
staubina, line 195: For consistency could we add Docstrings to the other methods?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1101,7 +1109,19 @@ def ensure_official_version(self):
             official_version.course = course
             official_version.save()
 
-    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+    def handle_status_change(self):
```
staubina, line 1112: Nit: A docstring might be helpful on this method.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1101,7 +1109,19 @@ def ensure_official_version(self):
...
+        if not self.draft or self._old_status == self.status:
+            return
+
+        if self.status == CourseRunStatus.LegalReview:
+            emails.send_email_for_legal_review(self)
+        if self.status == CourseRunStatus.InternalReview:
```
staubina: Should this be an `elif`?

staubina (commented):
Overall the changes look great.  I like the clean method of separation, which will make future work pretty straightforward. 

I just had a couple of questions and small change requests.

course_discovery/apps/course_metadata/emails.py
mikix: According to the database model, there can be multiple users per org-role. But in practice, we only assign one PC.

course_discovery/apps/course_metadata/emails.py
mikix: For edx.org, that would be an alertable situation, yeah. But also hopefully unlikely that our config is messed up.

course_discovery/apps/course_metadata/emails.py
mikix: Sure. In my head, these three were the "public methods" and the rest were implementation details. But doesn't mean they don't also deserve docstrings.

course_discovery/apps/course_metadata/models.py
mikix: Yes.

course_discovery/apps/course_metadata/emails.py
mikix: Where would we reuse these file extension suffixes?

course_discovery/apps/course_metadata/emails.py
mikix: I now call that situation out in the docstring

PR #1934, opened 2019-06-04
Use cache_response decorator for courses list

bbaker6225:
Supposedly the cacheresponsemixin does not work for list if you define
your own list for an API view. Let's try this instead.


mikix (approved):
Let's try it!

PR #1939, opened 2019-06-06
add external course key

jansenk:
-For discovery CourseRun and the publisher CourseRun, add CharField called external_key.
-Generated migrations.
-Update django admin for both models.
-add field to serializers


course_discovery/apps/publisher/api/serializers.py
```diff
@@ -98,6 +98,18 @@ def validate_lms_course_id(self, value):
 
         return value
 
+    def validate_external_key(self, value):
+        if value is None:
+            return value
+
+        if not re.match(r'[A-Za-z0-9-_:]+', value):
```
iloveagent57: NIT: could you extract this pattern into a constant?

iloveagent57 (approved):
Looks great!  Just the one nit.

PR #1944, opened 2019-06-11
adding expected program info to the Course run to support review

staubina:
Backend changes to support adding expected program type and name


staubina:
Front end: https://github.com/edx/publisher-frontend/pull/196

course_discovery/apps/course_metadata/models.py
```diff
@@ -151,6 +151,44 @@ class Meta:
         ordering = ('order',)
 
 
+class SeatType(TimeStampedModel):
```
staubina, line 156: Note: I moved these two classes so they were defined before use.

course_discovery/apps/course_metadata/models.py
```diff
@@ -784,6 +822,8 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
         verbose_name=_('Add OFAC restriction text to the FAQ section of the Marketing site')
     )
 
+    expected_program_type = models.ForeignKey(ProgramType, default=None, null=True, blank=True)
+    expected_program_name = models.CharField(max_length=255, default=None, null=True)
```
mikix: The Django recommendation is to not set `null=True` for CharFields and just use the empty string as the "empty" value. And you probably want `blank` too.., so something like:
`     expected_program_name = models.CharField(max_length=255, default='', blank=True)`?

Do we ever envision wanting to set two program types? (old publisher allows it, right?)

This whole thing feels like boundary leakage a bit. It's putting "not yet true" publisher hints into the CourseMetadata tables. It's enshrining some kind of specific-to-current-edx.org-internal-workflow details in the models...

All of that, plus the fact that there are two closely related fields here make me wonder if this shouldn't be a separate table, as we kind of talked about at standup today. Just for ease of modification-of-workflow/removal in future. It would mean more work, serializing them and such. :(

course_discovery/apps/course_metadata/migrations/0178_auto_20190610_1547.py
```diff
@@ -0,0 +1,36 @@
...
+            model_name='historicalcourserun',
+            name='expected_program_name',
+            field=models.CharField(default=None, max_length=255, null=True),
+        ),
+        migrations.AddField(
+            model_name='historicalcourserun',
+            name='expected_program_type',
+            field=models.ForeignKey(blank=True, db_constraint=False, default=None, null=True, on_delete=django.db.models.deletion.DO_NOTHING, related_name='+', to='course_metadata.ProgramType'),
+        ),
+    ]
```
mikix: As discussed, you probably want to port the publisher table info here.

And update the `publish_to_discovery` method in `course_discovery/apps/publisher/api/v1/views.py` to also copy over the program info.

mikix:
Also, do we even need the `expected_program_type`? I thought we didn't really use that?

course_discovery/apps/course_metadata/models.py
staubina: I talked with PC and Product and determined there should only ever be one Program type assigned.  Also confirmed that in the Publisher data right now there is only 1 set of 7 runs that have more than one type set.  I am going to to work to get this data corrected before running the migration.

I wondered the same thing about whether or not this should be its own table.  It complicates things a little around getting the options and setting the data but should be reasonable.  I will reach out today to discuss this new approach.

course_discovery/apps/course_metadata/models.py
```diff
@@ -152,6 +152,66 @@ class Meta:
...
+
+    # Do not record the slug field in the history table because AutoSlugField is not compatible with
+    # django-simple-history.  Background: https://github.com/edx/course-discovery/pull/332
+    history = HistoricalRecords(excluded_fields=['slug'])
+
+    def __str__(self):
+        return self.name
+
+    @staticmethod
+    def get_program_type_data(pub_course_run):
```
staubina: Note: This method is new

mikix: This feels pretty publisher specific for the CM models, but there's already a lot of bleed there.

Could have this method return an object, not a slug, since the only times you use it, you immediately do a `get` anyway. Ah... but during the migration, you want to control which model class is used. Could pass in a `model=` argument?

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -152,7 +156,9 @@ def publish_to_discovery(self, partner, course_run):
             'max_effort': course_run.max_effort,
             'language': course_run.language,
             'weeks_to_complete': course_run.length,
-            'has_ofac_restrictions': course_run.has_ofac_restrictions
+            'has_ofac_restrictions': course_run.has_ofac_restrictions,
+            'expected_program_name': program_name if program_name else '',
```
mikix: `program_name or ''` would do same thing but feel a little more obvious to me

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.21 on 2019-06-11 19:44
+from __future__ import unicode_literals
+from django.db import migrations, models
+import django.db.models.deletion
+
+from course_discovery.apps.course_metadata.models import ProgramType
```
mikix: You don't want this here, ya? You want to only use the result of `apps.get_model`? Ah... you need it for the class utilty method. Still, tempting to accidentally use it.

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def migrate_expected_programs_from_pub_cm(apps, schema_editor):
+    CourseRun = apps.get_model('course_metadata', 'CourseRun')
+    pub_course_run_model = apps.get_model('publisher', 'CourseRun')
+    program_type = apps.get_model('course_metadata', 'ProgramType')
+
+    for course_run in pub_course_run_model.objects.all():
+        slug, name = ProgramType.get_program_type_data(pub_course_run=course_run)
```
mikix: It's an arg, not a kwarg. Don't *need* the `pub_course_run=` - but maybe you like it for clarity.

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
```diff
@@ -0,0 +1,63 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+
+from course_discovery.apps.course_metadata.models import ProgramType
+
+
+def migrate_expected_programs_from_pub_cm(apps, schema_editor):
+    CourseRun = apps.get_model('course_metadata', 'CourseRun')
+    pub_course_run_model = apps.get_model('publisher', 'CourseRun')
+    program_type = apps.get_model('course_metadata', 'ProgramType')
```
mikix: Three different naming schemes here. :)

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
```diff
@@ -0,0 +1,63 @@
...
+        expected_program_type = None
+        if slug:
+            expected_program_type = program_type.objects.get(slug=slug)
+        defaults = {
+            'expected_program_name': name if name else '',
+            'expected_program_type': expected_program_type,
+        }
+
+        CourseRun.everything.update_or_create(
+            key=course_run.lms_course_id,
```
mikix: Not every CourseRun object in publisher hsa an lms_course_id set I don't think? I think that gets set... first left-publish maybe? (You'll see a lot of `if self.lms_course_id` in the publisher models.py file.) Might want to guard this with an `if`

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -101,7 +101,7 @@ def get_queryset(self):
         # Start with either draft versions or real versions of the courses
         if edit_mode:
             # TODO: For now hardcode in draft=True until we choose to roll this out live, DISCO-818
```
mikix: Should remove the comment too if you're doing this (and all the versions of this comment in tests).

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
staubina: I will look to see if there is another id to use then.  Thanks for the heads up on this.

course_discovery/apps/api/v1/views/courses.py
staubina: 👍

staubina: I might actually remove this change for now.  I only needed it for one test.  Then do it in a separate clean PR.

staubina:
> This feels pretty publisher specific for the CM models, but there's already a lot of bleed there.

> Could have this method return an object, not a slug, since the only times you use it, you immediately do a get anyway. Ah... but during the migration, you want to control which model class is used. Could pass in a model= argument?

I am tempted to move the helper method again 🤦‍♂.  You are right that this is a specific Publisher task and really only needed for Publisher and this one migration. 

Thoughts @mikix ?

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
staubina: To be able to call `get_program_type_data` right now I am importing `ProgramType`, if we move it I could remove/update the import.

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
staubina: A remnant of testing, removing

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
staubina: Another remnant of debugging, cleaning up

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
staubina: Can we talk about this one when you get in?

staubina: This is currently the key used to determine the CourseRun that we push to in old publisher.  I did a quick query and all but I think 3 runs have the lms_course_id set in Prod.

staubina: https://github.com/edx/course-discovery/blob/e25f9e2334301d36046632d7b4bad24aeab674aa/course_discovery/apps/publisher/api/v1/views.py#L163

I will look into adding a conditional though to be safe.

course_discovery/apps/api/v1/views/course_runs.py.orig
```diff
@@ -0,0 +1,347 @@
+import logging
```
mikix: Whoops.

course_discovery/apps/course_metadata/migrations/0182_courserun_expected_program.py
```diff
@@ -0,0 +1,63 @@
...
+def migrate_expected_programs_from_pub_cm(apps, schema_editor):
+    cm_course_run = apps.get_model('course_metadata', 'CourseRun')
+    pub_course_run_model = apps.get_model('publisher', 'CourseRun')
+    cm_program_type = apps.get_model('course_metadata', 'ProgramType')
+
+    for course_run in pub_course_run_model.objects.all():
+        expected_program_type, name = ProgramType.get_program_type_data(course_run, cm_program_type)
+
+        defaults = {
+            'expected_program_name': name if name else '',
```
mikix: Could do same `name or ''` trick here, but who care.

mikix (approved):
But yeah, all good if you fix the .orig file inclusion,

course_discovery/apps/api/v1/views/course_runs.py.orig
staubina: 🤦‍♂

PR #1947, opened 2019-06-12
Push products to ecom in new publisher

mikix:
Consolidate the 'publish to ecommerce' logic between old and new publishers into a common code flow. Then use it in both places.

I chose **not** to publish to ecommerce when products are changed via django admin (i.e. didn't push to ecommerce in the save() method for product models). It would add to the complexity, lose the benefits of pushing all products at once, and isn't a supported flow. We can revisit that if needed.

The places we **do** now  push to ecommerce are: (1) old publisher left-publish button, (2) when updating an already-published course run via the endpoint, (3) when updating an already-published course via the endpoint, we do it for all future & published sub runs in that course, and (4) when saving a draft course run into the Reviewed state (from anywhere, including django admin).

I also did a little refactoring around mock access tokens and such. Using the non-deprecated oauth client for this work caused a little mess around our tests.

https://openedx.atlassian.net/browse/DISCO-984


course_discovery/apps/course_metadata/models.py
```diff
@@ -1169,13 +1175,17 @@ def update_or_create_official_version(self):
         for seat in self.seats.all():
             set_official_state(seat, Seat, {'course_run': official_version})
 
-        official_course = self.course.update_or_create_official_version(official_version)
+        official_course = self.course._update_or_create_official_version(official_version)  # pylint: disable=protected-access
```
jlajoie, line 1178: Should this maybe just be a public method, or is it because of the ```only be called from CourseRun.update_or_create_official_version``` comment above that we're keeping it private to discourage use elsewhere?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1285,6 +1295,9 @@ class Seat(DraftModelMixin, TimeStampedModel):
 
     REQUIRES_AUDIT_SEAT = [VERIFIED]
 
+    # Seat types that we don't push to ecommerce -- they are either manually or never created as products
+    NON_PRODUCT_MODES = [CREDIT, MASTERS]
```
jlajoie, line 1299: Just for clarity for me, what is Product in this context, like a Product model in oscar?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -284,6 +289,104 @@ def get_course_run_estimated_hours(course_run):
...
+        'price': str(seat.price),
+        'product_class': 'Seat',
+        'attribute_values': [
+            {
+                'name': 'certificate_type',
+                'value': '' if seat.type == Seat.AUDIT else seat.type,
+            },
+            {
+                'name': 'id_verification_required',
+                'value': seat.type in (Seat.VERIFIED, Seat.PROFESSIONAL),
```
jlajoie, line 324: Do we still support professional no id verification, and/or do we need to worry about that?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -284,6 +289,104 @@ def get_course_run_estimated_hours(course_run):
...
+            {
+                'name': 'certificate_type',
+                'value': entitlement.mode if isinstance(entitlement.mode, str) else entitlement.mode.slug,
+            },
+        ],
+    }
+
+
+def push_to_ecommerce_for_course_run(course_run):
+    from course_discovery.apps.course_metadata.models import Seat
```
jlajoie, line 345: Might be worth having the "avoid circular dependencies" comment here too?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -284,6 +289,104 @@ def get_course_run_estimated_hours(course_run):
...
+
+    course = course_run.course
+    api = course.partner.lms_api_client
+    if not api or not course.partner.ecommerce_api_url:
+        return False
+
+    # NOTE: We only order here to aid testing. The E-Commerce API does NOT care about ordering.
+    seats = course_run.seats.exclude(type__in=Seat.NON_PRODUCT_MODES).order_by('created')
```
jlajoie: Might be worth us omitting these order bys as they'll run additional queries instead of using the cached version if there is one? If it's just for testing we should probably not be dependent on arrays matching exactly anyways? Know this probably isn't something you did but figured I'd point it out.

jlajoie (approved):
Minor comments/questions but overall great refactor 👍

course_discovery/apps/course_metadata/models.py
mikix: Yeah just to discourage use elsewhere. Do you think the docstring is enough for that?

course_discovery/apps/course_metadata/models.py
mikix: Yes. Maybe clearer as NON_ECOMMERCE_MODES?

course_discovery/apps/course_metadata/utils.py
mikix: This is copied code anyway, but this specific check is for the value of `id_verification_required` - where we wouldn't want to include professional-no-id anyway.

course_discovery/apps/course_metadata/utils.py
mikix: GP, will do

course_discovery/apps/course_metadata/utils.py
mikix: GP, will do. I'll figure the test fallout out

jlajoie: Bit late response, but alternatively we can sort in-memory?

course_discovery/apps/course_metadata/models.py
jlajoie: Naw I'm fine with this was just double checking!

course_discovery/apps/course_metadata/models.py
jlajoie: I think the private is a great call actually, just wanted to double check

PR #1949, opened 2019-06-13
Add regular courses endpoint to api yaml files

bbaker6225:
Starting with a barebones api gateway integration. We'll add more
endpoints and integrations once we know everything is working.


jlajoie (approved):
Not familiar enough with best practices on where these .yaml files should live at or how they should be represented, but seems sane/reasonable. LGTM 👍

PR #1952, opened 2019-06-13
DISCO-1048: Adds in Salesforce Configuration model tied into Partner

jlajoie:
Config model for SF configs:

Screen Shot 2019-06-13 at 3 40 20 PM


course_discovery/apps/core/migrations/0014_salesforceconfiguration.py
```diff
@@ -0,0 +1,28 @@
...
+            name='SalesforceConfiguration',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('salesforce_username', models.CharField(max_length=255, verbose_name='Salesforce Username')),
+                ('salesforce_password', models.CharField(max_length=255, verbose_name='Salesforce Password')),
+                ('salesforce_organization_id', models.CharField(blank=True, default='', max_length=255, verbose_name='Salesforce Organization Id')),
+                ('salesforce_security_token', models.CharField(blank=True, default='', max_length=255, verbose_name='Salesforce Security Token')),
+                ('salesforce_is_sandbox', models.BooleanField(default=True, verbose_name='Is a Salesforce Sandbox?')),
+                ('partner', models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, to='core.Partner')),
+            ],
```
mikix, line 26: Do we want to add a historical model to this too? They added one to Partner. So this might want it as well.

course_discovery/apps/core/models.py
```diff
@@ -144,3 +144,35 @@ def lms_api_client(self):
             return None
 
         return OAuthAPIClient(self.lms_url.strip('/'), self.oidc_key, self.oidc_secret)
+
+
+class SalesforceConfiguration(models.Model):
+    partner = models.OneToOneField(Partner)
```
mikix, line 150: Wanna specify a nice `related_name`? I think the default would be `salesforceconfiguration`.

And passing `on_delete=` is going to become mandatory in 2.x, might as well start doing it now with `on_delete=models.CASCADE`

course_discovery/apps/core/models.py
```diff
@@ -144,3 +144,35 @@ def lms_api_client(self):
...
+    salesforce_username = models.CharField(
```
mikix, line 151: Could we get away with dropping the `salesforce_` prefix on these? If we have a `related_name` of `salesforce` for this model, references could be `partner.salesforce.username` which looks nice.

Or if you like the clarity of a namespace prefix, maybe because the context is more obvious, we could get away with `sf_username` or something? I dunno. I just hate long line lengths. :)

course_discovery/apps/core/models.py
```diff
@@ -144,3 +144,35 @@ def lms_api_client(self):
...
+        max_length=255,
+        verbose_name=_('Salesforce Username'),
+        blank=False,
+        null=False,
```
mikix, line 155: Blank and null are `False` by default.

mikix (approved):
Maybe provide a factory too? And optionally hook it into PartnerFactory as a subfactory. Or maybe you'd prefer to wait until you actually use this and write tests for that.

course_discovery/apps/core/models.py
mikix: Although.... Django bizarrely raises a `DoesNotExist` if you access the related name of a OneToOne and it doesn't exist, instead of an (arguably) kinder `None`. So *maybe* an awkward related name is fine, and we use an `@property` to catch the exception and return `None` -- if we like that usage pattern more. Depends how/how-often you envision this used.

course_discovery/apps/core/models.py
jlajoie: Just double checking, CASCADE would be impacting the Partner -> Config flow, and wouldn't impact partner when deleting the config right?

course_discovery/apps/core/models.py
jlajoie: Related name sounds good to me 👍

course_discovery/apps/core/migrations/0014_salesforceconfiguration.py
jlajoie: Yeah probably isn't necessary but doesn't really hurt to add for this

jlajoie:
> Maybe provide a factory too? And optionally hook it into PartnerFactory as a subfactory. Or maybe you'd prefer to wait until you actually use this and write tests for that.

Gonna hold off til we're using that (subsequent PR), everything else seems like a GP and I'll fix up 👍

course_discovery/apps/core/models.py
mikix: Right. Cascade means, when a Partner is deleted, delete me too. (It's also the default if you don't provide `on_delete`)

PR #1958, opened 2019-06-17
Unpin as many dependencies as possible

mikix:
This will need some extra heavy testing in stage.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_lookups.py
+++ b/course_discovery/apps/course_metadata/tests/test_lookups.py
@@ -248,7 +248,8 @@ class AutoCompletePersonTests(SiteMixin, TestCase):
         )
         assert response.status_code == 200
         data = json.loads(response.content.decode('utf-8'))
-        expected_results = [{'id': instructor.id, 'text': str(instructor)} for instructor in self.instructors]
+        expected_results = [{'id': str(instructor.id), 'text': str(instructor), 'selected_text': str(instructor)}
+                            for instructor in self.instructors]
         assert data.get('results') == expected_results
 
     def _make_user_non_staff(self):
--- a/course_discovery/apps/publisher/api/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/tests/test_views.py
@@ -968,7 +968,7 @@ class CoursesAutoCompleteTests(SiteMixin, TestCase):
         data = json.loads(response.content.decode('utf-8'))
         self.assertEqual(len(data['results']), 2)
 
-        results_by_id = {record['id']: record for record in data['results']}
+        results_by_id = {int(record['id']): record for record in data['results']}
         self.assertFalse(results_by_id[self.course.id]['uses_entitlements'])
         self.assertTrue(results_by_id[self.course2.id]['uses_entitlements'])
 
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -1,7 +1,7 @@
 beautifulsoup4
 django
 django-admin-sortable2
-django-autocomplete-light==3.1.8
+django-autocomplete-light
 django-choices
 django-compressor
 django-contrib-comments
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -24,7 +24,7 @@ ddt==1.2.1
 defusedxml==0.6.0         # via djangorestframework-xml, python3-openid, social-auth-core
 django-admin-sortable2==0.7.2
 django-appconf==1.0.3     # via django-compressor
-django-autocomplete-light==3.1.8
+django-autocomplete-light==3.4.1
 django-choices==1.7.0
 django-compressor==2.3
 django-contrib-comments==1.9.1
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -15,7 +15,7 @@ cryptography==2.7         # via pyopenssl
 defusedxml==0.6.0         # via djangorestframework-xml, python3-openid, social-auth-core
 django-admin-sortable2==0.7.2
 django-appconf==1.0.3     # via django-compressor
-django-autocomplete-light==3.1.8
+django-autocomplete-light==3.4.1
 django-choices==1.7.0
 django-compressor==2.3
 django-contrib-comments==1.9.1


course_discovery/apps/api/filters.py
```diff
@@ -111,9 +111,9 @@ def filter(self, qs, value):  # pylint: disable=method-hidden
 class UUIDListFilter(CharListFilter):
     """ Filters a field via a comma-delimited list of UUIDs. """
 
-    def __init__(self, name='uuid', label=None, widget=None, method=None, lookup_expr='in', required=False,
+    def __init__(self, field_name='uuid', label=None, widget=None, method=None, lookup_expr='in', required=False,
```
mikix: django-filters renamed this field

course_discovery/apps/api/renderers.py
```diff
@@ -62,4 +62,5 @@ class CourseRunCSVRenderer(CSVStreamingRenderer):
         'seats.credit.credit_provider',
         'seats.credit.credit_hours',
         'modified',
+        'course_key',
```
mikix: I'm honestly not sure why, but the old CSV output included `course_key` but did not request it to be included? This will need confirmation in stage probably -- I don't grok this fully.

course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
```diff
@@ -326,7 +327,7 @@ def test_csv(self):
         content = list(reader)
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(set(expected), set(content[1]))
+        self.assertEqual(expected, content[1])
```
mikix: This wasn't necessary, but I figured `set` makes no sense? Order matters in csv, plus we don't want to miss someone adding a duplicate column.

course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
```diff
@@ -172,6 +172,7 @@ def test_courses(self, state):
         for function in state:
             function(course_run)
 
+        course_run.course.save()
```
mikix: State functions can now change course (because I changed it below to do so sometimes).

course_discovery/conftest.py
```diff
@@ -55,6 +55,8 @@ def end_future(course_run):
         course_run.end = future
 
     def slug_blank(course_run):
+        course_run.course.title = ''
+        course_run.title_override = ''
```
mikix: AutoSlugField is more aggressive about resetting slug (it now does it if slug is blank, even if it's not row-creation time. So we zero out the source of the slug too.

course_discovery/apps/core/migrations/0001_squashed_0011_auto_20161101_2207.py
```diff
@@ -20,8 +20,8 @@ def add_currencies(apps, schema_editor):
     """
     Currency = apps.get_model('core', 'Currency')
     Currency.objects.bulk_create(
-        [Currency(code=currency.letter, name=currency.name) for currency in pycountry.currencies if
-            not currency.letter.startswith('X')]
+        [Currency(code=currency.alpha_3, name=currency.name) for currency in pycountry.currencies if
+            not currency.alpha_3.startswith('X')]
```
mikix: currency model field got renamed upstream

course_discovery/apps/course_metadata/tests/__init__.py
```diff
@@ -17,5 +19,6 @@ def toggle_switch(name, active=True):
     switch, __ = Switch.objects.get_or_create(name=name, defaults={'active': active})
     switch.active = active
     switch.save()
+    switch.flush()
```
mikix: waffle is now more aggressive about caching.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1915,14 +1915,9 @@ class TestGetUTMSourceForUser(LMSAPIClientMixin, TestCase):
 
     def setUp(self):
         super(TestGetUTMSourceForUser, self).setUp()
-
-        self.switch, __ = Switch.objects.update_or_create(
-            name='use_company_name_as_utm_source_value', defaults={'active': True}
-        )
```
mikix: Waffle got a little more black-box with caching and all that. Direct model access doesn't work as well. Switched this class to only using the `override_switch` decorator.

course_discovery/apps/api/v1/tests/test_views/test_currency.py
```diff
@@ -38,18 +38,18 @@ def test_get(self, admin_client, django_cache, responses, settings):  # pylint:
 
         response = admin_client.get(self.list_path)
 
-        assert all(item in response.data.items() for item in expected.items())
+        assert all(item in response.json().items() for item in expected.items())
```
mikix: HttpResponse seems to now not always provide `data`, but does always provides `json()`

.pycodestyle
```diff
@@ -1,4 +1,4 @@
-[pep8]
-ignore=E501
+[pycodestyle]
+ignore=E501,W504
```
mikix: W504 is the pep8 rule that wants `or` and `and` to be at the beginning of lines instead of the end. Like:
```
if (one
    or two):
```
instead of:
```
if (one or
    two):
```

This warning is normally disabled by default if you don't specify an ignore line.

Makefile
```diff
@@ -42,15 +42,23 @@ production-requirements: ## Install Python and JS requirements for production
...
 	## The node_modules .bin directory is added to ensure we have access to Geckodriver.
 	PATH="$(NODE_BIN):$(PATH)" pytest --ds=course_discovery.settings.test --durations=25
 	coverage combine
 	coverage report
 
-quality: ## Run pep8 and Pylint
+quality: ## Run pycodestyle and Pylint
 	isort --check-only --diff --recursive acceptance_tests/ course_discovery/
-	pep8 --config=.pep8 acceptance_tests course_discovery *.py
+	pycodestyle --config=.pycodestyle acceptance_tests course_discovery *.py
```
mikix: pep8 tool is deprecated in favor of pycodestyle.

course_discovery/conftest.py
```diff
@@ -54,12 +54,6 @@ def end_past(course_run):
     def end_future(course_run):
         course_run.end = future
 
-    def slug_blank(course_run):
-        course_run.slug = ''
-
-    def slug_valid(course_run):
-        course_run.slug = 'foo'
-
```
mikix: So AutoSlugField is now more aggressive and does not let you save an empty slug. So this can't be tested anymore.

course_discovery/apps/api/v1/tests/test_views/test_people.py
```diff
@@ -20,6 +20,7 @@
 User = get_user_model()
 
 
+@override_switch('publish_person_to_marketing_site', True)
```
mikix: I transitioned from our own toggle_switch to the official override_switch, because waffle got more aggressive about caching its switches. Because of that, we now need both an enter and exit (to wipe cache) when overriding.

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -248,7 +248,8 @@ def test_instructor_autocomplete_from_django_admin(self):
         )
         assert response.status_code == 200
         data = json.loads(response.content.decode('utf-8'))
-        expected_results = [{'id': instructor.id, 'text': str(instructor)} for instructor in self.instructors]
+        expected_results = [{'id': str(instructor.id), 'text': str(instructor), 'selected_text': str(instructor)}
+                            for instructor in self.instructors]
```
mikix, line 252: The autocomplete code started returning a new parameter `selected_text`. :man_shrugging:

course_discovery/apps/publisher/tests/test_emails.py
```diff
@@ -681,50 +715,3 @@ def test_seo_review_email(self):
         page_url = 'https://{host}{path}'.format(host=self.site.domain.strip('/'), path=page_path)
         self.assertIn(page_url, body)
         self.assertIn('determine OFAC status', body)
-
-
-class CourseRunPublishedEditEmailTests(CourseRunPublishedEmailTests):
```
mikix: This was caught by one of pylint/pycodestyle - this test class inherits another and dups all its classes (but without any setUp() code that would make that an interesting thing to do). So I just moved its two tests into its parent class.

openedx.yaml
```diff
@@ -1,14 +1,13 @@
 # This file describes this Open edX repo, as described in OEP-2:
-# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+# https://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html#specification
 
 nick: ctlg
 openedx-release: {ref: master}
 owner: rock345
-track-pulls: true
```
mikix: OEP-2 says this field can be removed, it's not used anymore.

Makefile
```diff
@@ -42,15 +42,23 @@ production-requirements: ## Install Python and JS requirements for production
...
+	pip-compile --upgrade -o requirements/test.txt requirements/test.in
+	pip-compile --upgrade -o requirements/local.txt requirements/local.in
+	pip-compile --upgrade -o requirements/production.txt requirements/production.in
+	chmod a+rw requirements/*.txt
+
 test: clean ## Run tests and generate coverage report
 	## The node_modules .bin directory is added to ensure we have access to Geckodriver.
 	PATH="$(NODE_BIN):$(PATH)" pytest --ds=course_discovery.settings.test --durations=25
-	coverage combine
+	coverage combine || true  # will fail if nothing to do, but don't abort if that happens
```
mikix: coverage 4.3 made "no data to combine" an error case. I don't know when we need to combine coverage (multiple parallel test processes?), but didn't aggressive enough to drop the combine line altogether. Instead, I just made the no-op case not an error.

mikix:
@jmbowman I'd love a look here. I know in another OEP-18 PR, you asked that the `.txt` files be included with `-r`, not the `.in` files. But then you lose the the nice `# via pytest, coverage` comments. What's the benefit of including the `.txt`?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1932,6 +1926,7 @@ def test_with_waffle_switch_turned_off(self, mock_access_token):  # pylint: disa
 
         assert get_utm_source_for_user(self.partner, self.user) == self.user.username
 
+    @override_switch('use_company_name_as_utm_source_value', active=True)
```
Dillon-Dumesnil: Nit: Is it possible to put this override switch on the class of tests to save lines? And then only override false when that is desired?

course_discovery/apps/api/v1/tests/test_views/test_mixins.py
```diff
@@ -4,7 +4,7 @@
 def test_fuzzy_int_equality():
     fuzzy_int = FuzzyInt(10, 4)
 
-    for i in range(0, 6):
+    for i in range(6):
```
Dillon-Dumesnil: 🙏

course_discovery/apps/course_metadata/tests/test_query.py
```diff
@@ -105,16 +105,6 @@ def test_marketable(self):
 
         self.assertEqual(list(CourseRun.objects.marketable()), [course_run])
 
-    def test_marketable_exclusions(self):
```
Dillon-Dumesnil: ..?

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -16,8 +16,9 @@
 from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.models import CourseEntitlement as DiscoveryCourseEntitlement
+from course_discovery.apps.course_metadata.models import CourseRun, ProgramType
 from course_discovery.apps.course_metadata.models import Seat as DiscoverySeat
-from course_discovery.apps.course_metadata.models import CourseRun, ProgramType, SeatType, Video
+from course_discovery.apps.course_metadata.models import SeatType, Video
```
Dillon-Dumesnil: Not sure I like this one... ¯\_(ツ)_/¯

course_discovery/apps/publisher/forms.py
```diff
@@ -681,7 +682,7 @@ def clean(self):
         course = cleaned_data.get('course')
         if not CourseUserRole.objects.filter(course=course, role=owner_role):
             raise forms.ValidationError(
-                _('Please create {} course user role before assigning it owner role'.format(owner_role))
+                format_lazy(_('Please create {} course user role before assigning it owner role'), owner_role)
```
Dillon-Dumesnil: Nit: Instead of using format_lazy, switch the `{}` to `{owner_role}` and then `.format(owner_role=owner_role)`.

This just seems to be the only spot we are using this format_lazy instead of .format. Unless there are other reasons, I just haven't seen format_lazy before

course_discovery/apps/publisher/tests/test_emails.py
```diff
@@ -552,7 +546,49 @@ def test_course_published_email_with_error(self):
...
+        self.assertEqual([self.course.publisher.email], mail.outbox[0].to)
+        self.assertEqual(str(mail.outbox[0].subject), subject)
+        body = mail.outbox[0].body.strip()
+        self.assertIn('has made changes to the following published course run.', body)
+        page_path = reverse('publisher:publisher_course_run_detail', kwargs={'pk': self.run_state.course_run.id})
+        self.assertIn(page_path, body)
+
+    def test_email_with_error(self):
+        """ Verify that email failure logs error message."""
+
```
Dillon-Dumesnil: Remove empty line here

Dillon-Dumesnil (approved):
Definitely let Jeremy's comments strongly override anything I say, but looks good! Couple nits, but it's beautiful. Thanks for doing this @mikix !

course_discovery/apps/course_metadata/tests/test_query.py
mikix: Oh this was because of the can't-set-slug-as-blank anymore, and this test was testing for a blank slug.

course_discovery/apps/publisher/api/v1/tests/test_views.py
mikix: Yeah... I hear ya.

course_discovery/apps/publisher/forms.py
mikix: Ah, it's because in that file, we are importing `ugettext_lazy as _` rather than `ugettext as _`. Which (I think) you do if you use `_()` in situations where simply importing the file would process the `_()` - like class level definitions like in arguments to class fields in `models.py`.

And the linter is complaining that the object returned by `ugettext_lazy` doesn't have a `format` call - which I *think* is a red herring? But there's also `format_lazy` for this situtation.

mikix:
I'm abandoning this PR for now, in favor of https://github.com/edx/course-discovery/pull/2005 which keeps a lot more pinned, to avoid any code changes and avoid the flaky tests I didn't want to spend more time fixing here.

But keeping this PR for historical context for now, next time I try to remove some pins.

PR #1959, opened 2019-06-17
Add a command to publish runs once live date hits

mikix:
A new management command publish_live_course_runs has now been added.
Any course runs which are reviewed and whose dates have passed are
now able to be published by running this command.

https://openedx.atlassian.net/browse/DISCO-815


course_discovery/apps/course_metadata/models.py
```diff
@@ -1252,60 +1252,97 @@ def update_or_create_official_version(self):
...
-        is_new_course_run = not self.pk
-        suppress_publication = kwargs.pop('suppress_publication', False)
-        is_publishable = (
-            self.course.partner.has_marketing_site and
-            waffle.switch_is_active('publish_course_runs_to_marketing_site') and
-            # Pop to clean the kwargs for the base class save call below
-            not suppress_publication and
-            not self.draft
-        )
+    def save(self, suppress_publication=False, **kwargs):
```
mikix, line 1289: I refactored `save()` a bit here, but it wasn't really necessary for my work. Just saw an opportunity. I moved the marketing site stuff to its own method. I think it's a bit more readable now.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1331,7 +1368,7 @@ def is_marketable(self):
         A course run is considered marketable if it's published, has seats, and
         a non-empty marketing url.
         """
-        is_published = self.status.lower() == 'published'
+        is_published = self.status == CourseRunStatus.Published
```
mikix, line 1371: A test in travis started failing on ChoiceItem not having a lower method?...  Not sure why I triggered that. But this code wasn't great anyway. That caution around case is not done anywhere else (and our DB has only lowercase...). So I just figured I'd clean it up.

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Based on course run go_live dates, publish any runs that are due'
+
+    def handle(self, *args, **options):
+        now = datetime.datetime.now(pytz.UTC)
+        failed = False
+
+        for course_run in CourseRun.objects.all():
```
staubina: Why don't we filter the Course run by datetime rather than looking at every course run?

staubina: We should be able to use the ORMs filter query to greatly reduce the set of course runs which should by default be faster than looping through all course runs.

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
```diff
@@ -0,0 +1,38 @@
...
+            reviewed = course_run.status == CourseRunStatus.Reviewed
+            past_due = course_run.go_live_date and course_run.go_live_date <= now
+            if not reviewed or not past_due:
+                continue
+
+            logger.info(_('Publishing course run {key}').format(key=course_run.key))
+
+            try:
+                course_run.publish()
+            except Exception:  # pylint: disable=broad-except
```
staubina, line 27: nit: I am still not a fan of Bare exceptions. Any chance we can be more specific here?

staubina (changes requested):
I am curious why we did not just filter rather than going through all course runs. 

Also, are we applying the redirects here too?  I might just be missing the method that we apply the redirects in.

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
staubina: These two checks seem like the ones that could easily be made into a filter 
```
            reviewed = course_run.status == CourseRunStatus.Reviewed
            past_due = course_run.go_live_date and course_run.go_live_date <= now
```

mikix: Yeah, you're right. I just brain farted on that.

mikix:
> Also, are we applying the redirects here too?

No. That was taken out of this ticket and moved to https://openedx.atlassian.net/browse/DISCO-898 (which asks for a redirect cron and to update this management command to include redirects).

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
mikix: Minor nitpick - "bare" exceptions and "broad" exceptions are different. "Bare" is without the word "Exception" in there, and I agree. Those are 99% of the time a bad idea.

But I'd push back on broad exceptions being always bad. I believe they are useful in situations like this where you are treating a code block like a black box and don't care why something fails. You just want to ensure that any failure doesn't stop the loop.

Like... even if we had a list of all possible errors that could be raised, what would be the deciding factor between continuing the loop and stopping it? I guess an error that would suggest the rest of the publish attempts would fail anyway, so why bother. But I'm not sure what those are.

mikix: I guess a common complaint of broad exceptions is also that it makes it hard to debug, because they hide errors. Which is fair if the code is trying to "handle" an error silently. But for cases where you catch-and-log, it seems a little less bad to me.

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
mikix: Fixed. Thank you!

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
```diff
@@ -0,0 +1,34 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Based on course run go_live dates, publish any runs that are due'
+
+    def handle(self, *args, **options):
+        failed = False
+        now = datetime.datetime.now(pytz.UTC)
+        course_runs = CourseRun.objects.filter(status=CourseRunStatus.Reviewed, go_live_date__lte=now)
```
staubina, line 20: 👍

course_discovery/apps/course_metadata/management/commands/publish_live_course_runs.py
staubina: I think my distaste comes from what you mentioned.  Catch and log seems reasonable.

staubina (approved):

PR #1963, opened 2019-06-18
Removed temporary draft=true block for Publisher endpoint

staubina:
[DISCO-818]

[DISCO-818]: https://openedx.atlassian.net/browse/DISCO-818


mikix (approved):
Sure

PR #1965, opened 2019-06-19
DISCO-1079: Adding external key to publisher edit form

Dillon-Dumesnil:
The way I implemented this will first try and check the Publisher Course Run Seats to see if any of them are marked as masters_track. If that fails, it will see if the course run we are looking at is the canonical course run for a course and thus only has it in course_metadata and then will check discovery to see if there is a Masters seat. 

I believe this is the best way to go about this since we have to wait for the canonical course run, but now we are able to expose the external_key field in the edit form immediately after course run creation in Publisher.


course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -250,6 +250,18 @@ <h1 class="hd-1 emphasized">
                             {% endif %}
                         </div>
                     </div>
+                    {% if seat_form.masters_track.value %}
```
staubina: @schenedx This is the check that I mentioned we were going to use for the Master's track.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
```diff
@@ -250,6 +250,18 @@ <h1 class="hd-1 emphasized">
...
+                        <div class="field-title">{% trans "INSTITUTION COURSE ID" %}<span class="optional float-right">Optional</span></div>
+                        <div class="row">
+                            <div class="col col-6 help-text">
+                                {% trans "Only fill in this field if you are planning to integrate with your Institution's systems through edX registrar service." %}
```
staubina: Should we mention contacting edX support here is you have questions?

staubina (approved):
The changes seem straightforward. I recommend we get a review from the Master's team though before merging.

course_discovery/apps/publisher/templates/publisher/course_run/edit_run_form.html
schenedx: @staubina, yesterday, I met with @Dillon-Dumesnil and we talked about this check. The problem I see with this check is, for the first course_run introduced into the masters program, this check will be false even tho the course_run is properly linked with Master program. This is because the first run have the Masters seat created by linking to Masters program on course_metadata model. This Masters seat would not populate back to the Publisher seat model. Since this check is on the Publisher seat model, this check will show false while the expected is to show true.
This check will be correct in the re-run case, however, since re-run case means both the Publisher seat model and course_metadata seat model will all have Masters seat.
The new check I suggested to @Dillon-Dumesnil is to check the course_metadata seat associated with this course_run instead. That is a better indicator the course run is associated with Masters program.

course_discovery/apps/publisher/templates/publisher/course_run_detail/_all.html
```diff
@@ -37,27 +38,36 @@
...
+          {% trans "Credit Price" %}
         </div>
-      {% endif %}
+        <div>
+          ${{ course_run.credit_seat_price }}
+        </div>
+      </div>
+    {% endif %}
+
+    {% if course_run.has_masters_seat %}
```
schenedx: @Dillon-Dumesnil It's interesting that this condition is different from the condition you had on the edit_run_form. Can those two share the same property or function?

Dillon-Dumesnil:
image
image

schenedx (approved):
Out of curiosity, we no longer write unit tests in publisher code?

PR #1966, opened 2019-06-20
reduce atomic transaction window for course_run patch

staubina:
[DISCO-1093]

[DISCO-1093]: https://openedx.atlassian.net/browse/DISCO-1093

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -230,6 +230,29 @@ class CourseRunViewSet(viewsets.ModelViewSet):
         return Response(serializer_data, status=status.HTTP_201_CREATED, headers=headers)
 
     @writable_request_wrapper
+    def _update_course_run(self, course_run, draft, changed, serializer, request, save_kwargs):
+
+        # If changes are made after review and before publish, revert status to unpublished.
+        # Unless we're just switching the status
+        if changed and course_run.status == CourseRunStatus.Reviewed:
+            save_kwargs['status'] = CourseRunStatus.Unpublished
+            official_run = course_run.official_version
+            official_run.status = CourseRunStatus.Unpublished
+            official_run.save()
+
+        if not draft and course_run.status != CourseRunStatus.Published:
+            save_kwargs['status'] = CourseRunStatus.LegalReview
+
+        course_run = serializer.save(**save_kwargs)
+
+        self.push_to_studio(request, course_run, create=False)
+
+        # Published course runs can be re-published directly
+        if not draft and course_run.status == CourseRunStatus.Published:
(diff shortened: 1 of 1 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -269,22 +288,7 @@ def update(self, request, **kwargs):
...
-            save_kwargs['status'] = CourseRunStatus.LegalReview
-
-        course_run = serializer.save(**save_kwargs)
-
-        self.push_to_studio(request, course_run, create=False)
-
-        # Published course runs can be re-published directly
-        if not draft and course_run.status == CourseRunStatus.Published:
-            course_run.update_or_create_official_version()
+        self._update_course_run(course_run, draft, changed, serializer, request, save_kwargs)
```
mikix: Do a `return self._update_course_run` because the wrapper decorator might return a Response. Then move the `return Response` below into this utility func.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -230,6 +230,25 @@ def create(self, request, *args, **kwargs):
         return Response(serializer_data, status=status.HTTP_201_CREATED, headers=headers)
 
     @writable_request_wrapper
+    def _update_course_run(self, course_run, draft, changed, serializer, request, save_kwargs):
+        if changed and course_run.status == CourseRunStatus.Reviewed:
+            save_kwargs['status'] = CourseRunStatus.Unpublished
+            # An official version should already exist, but just make sure
```
mikix: Drop this comment, that's no longer relevant. (and no longer in the code you are moving)

mikix (approved):
Looks great, with the small changes I suggest.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -269,24 +290,7 @@ def update(self, request, **kwargs):
 
         # If changes are made after review and before publish, revert status to unpublished.
         # Unless we're just switching the status
```
mikix: Also, move these comments with the code block they belong to above.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -230,6 +230,30 @@ def create(self, request, *args, **kwargs):
         return Response(serializer_data, status=status.HTTP_201_CREATED, headers=headers)
 
     @writable_request_wrapper
+    def _update_course_run(self, course_run, draft, changed, serializer, request, save_kwargs):
+        """
+        If changes are made after review and before publish, revert status to unpublished.
+        Unless we're just switching the status
```
mikix: These comments were actually just for the code block at the top here (the one that sets status to unpublished), not the whole function.

PR #1967, opened 2019-06-20
Add query params and program endpoints to api yaml

bbaker6225:
This should allow us to pass these query parameters through API gateway

The status checks on https://github.com/edx/api-manager/pull/86
shows that this yaml is at the very least valid


jlajoie (approved):
sure 👍

PR #1971, opened 2019-06-24
Link to embargo admin in review email

mikix:
When a course run needs to be internally reviewed, let's also provide a link to the restricted course admin page on the LMS side, which is often needed during that review.

This required adding a new optional Partner field: lms_admin_url


course_discovery/apps/course_metadata/tests/test_emails.py
```diff
@@ -77,6 +78,21 @@ def assertEmailContains(self, subject=None, to_users=None, both_regexes=None, te
         for regex in html_regexes or []:
             self.assertRegex(html, regex)  # pylint: disable=deprecated-method
 
+    def assertEmailDoesNotContain(self, both_regexes=None, text_regexes=None, html_regexes=None, index=0):
+        email = mail.outbox[index]
+        text = email.body
+        html = email.alternatives[0][0]
+
+        for regex in both_regexes or []:
```
Dillon-Dumesnil, line 86: Rather than have to use this `or []`, would it be helpful to switch the default values to being `[]` instead of `None`? I wasn't sure if there was context for using None instead.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -205,8 +205,13 @@ def send_email_for_internal_review(course_run):
         Arguments:
             course_run (Object): CourseRun object
     """
+    lms_admin_url = course_run.course.partner.lms_admin_url
+    restricted_url = lms_admin_url and (lms_admin_url.rstrip('/') + '/embargo/restrictedcourse/')
```
Dillon-Dumesnil, line 209: Nice use of rstrip

Dillon-Dumesnil (approved):
One comment, but otherwise LGTM

course_discovery/apps/course_metadata/tests/test_emails.py
mikix: Sadly, that's an anti-pattern. Python has this (to me) weird feature, that kwarg defaults are like, persistent objects? So the following code:

```
def test(input=[]):
     input.append('a')
     print(input)

test()
test()
test()
```

Will print:

```
['a']
['a', 'a']
['a', 'a', 'a']
```

So ideally, you never do `input=[]` or `input={}`

mikix: I guess it wouldn't have mattered here, where I was not modifying the argument. But I just have learned to always avoid doing that pattern, in case I added a subtle bug.

PR #1972, opened 2019-06-24
DISCO-1041: Adds in Salesforce utility methods

jlajoie:


course_discovery/apps/course_metadata/utils.py
```diff
@@ -18,7 +19,7 @@
 from course_discovery.apps.course_metadata.exceptions import MarketingSiteAPIClientException
 
 RESERVED_ELASTICSEARCH_QUERY_OPERATORS = ('AND', 'OR', 'NOT', 'TO',)
-
+##
```
mikix: Whoops

requirements/base.txt
```diff
@@ -53,4 +53,5 @@ python-dateutil==2.5.3
 pytz==2016.6.1
 # Pinning `requests` at 2.9.1 because the latest version of the package causes test failures with mocked requests
 requests==2.9.1
+simple-salesforce==0.74.2
```
mikix, line 56: They don't have official Python bindings? :(

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -370,6 +371,8 @@ def list(self, request, *args, **kwargs):
               paramType: query
               multiple: false
         """
+        sf_util = SalesforceUtil(request.site.partner)
+        sf_util.get_accounts()
```
mikix: What's happening here?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -473,3 +474,165 @@ def headers(self):
...
+            }
+            return Salesforce(**sf_kwargs)
+
+    instances = {}
+
+    def __init__(self, partner):
+        self.partner = partner
+        instance = SalesforceUtil.instances.get(partner)
+        if not instance:
+            SalesforceUtil.instances[partner] = SalesforceUtil.__SalesforceUtil(partner)
```
mikix: Nit: Another strategy that we use elsewhere is a method on the Salesforce django model like:
```
    @cached_property
    def client(self):
        return SalesforceUtil.__SalesforceUtil(partner)
```

Which would have the caching happen be managed in django's cache and access via `partner.salesforce.client`. Not clear if that's better or not, just throwing out there.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -473,3 +474,165 @@ def headers(self):
...
+                'domain': 'test' if salesforce_config.is_sandbox else ''
+            }
+            return Salesforce(**sf_kwargs)
+
+    instances = {}
+
+    def __init__(self, partner):
+        self.partner = partner
+        instance = SalesforceUtil.instances.get(partner)
+        if not instance:
```
mikix: You don't use the instance variable, so you could just do `if partner not in SalesforceUtil.instances:`

course_discovery/apps/course_metadata/utils.py
```diff
@@ -473,3 +474,165 @@ def headers(self):
...
+            return method(self, *args, **kwargs)
+    return inner
+
+
+class SalesforceUtil:
+    class __SalesforceUtil:
+        def __init__(self, partner):
+            self.partner = partner
+            if self.salesforce_is_enabled():
+                self.client = self.login()
```
mikix: I don't see any guarding elsewhere around an undefined `client`

course_discovery/apps/course_metadata/utils.py
```diff
@@ -473,3 +474,165 @@ def headers(self):
...
+        return getattr(SalesforceUtil.instances.get(self.partner), name)
+
+    def soql_escape(self, soql):
+        """
+        Escapes a soql string against injection
+
+        The single quote (') and backlash (\) characters are reserved in SOQL
+        queries and must be preceded by a backslash to be properly interpreted.
+        """
+        return soql.replace('\\', '\\\\').replace("'", "\\'")
```
mikix: And those are the only characters we care about? Does simple-salesforce not provide this sort of thing for us? :(

course_discovery/apps/course_metadata/utils.py
```diff
@@ -473,3 +474,165 @@ def headers(self):
...
+    def inner(self, *args, **kwargs):
+        try:
+            return method(self, *args, **kwargs)
+        except SalesforceExpiredSession:
+            self.login()
+            return method(self, *args, **kwargs)
+    return inner
+
+
+class SalesforceUtil:
```
mikix: Maybe a big ol' docstring here? How to use it, mentioning that attrs fall through to an underlying simple-salesforce client object, etc?

requirements/base.txt
jlajoie: Using this since it's just a nice API wrapper for SF. The two popular libraries are https://github.com/django-salesforce/django-salesforce and https://github.com/simple-salesforce/simple-salesforce. I didn't like the configuration options for django-sf so decided for the other approach

course_discovery/apps/api/v1/views/courses.py
jlajoie: Not meant to be here! Was messing around and can drop these. Thanks!

course_discovery/apps/course_metadata/utils.py
jlajoie: Yeah nothing for it internally, they're big on the "simple" I think. https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_reservedcharacters.htm

course_discovery/apps/course_metadata/utils.py
jlajoie: Hmm so I was thinking we'd use the salesforce_is_enabled() method externally to check if we should be sending queries/creating things around blocks of code. Do you think I should raise exceptions if client is None and we're trying to call creates? Or silently pass though?

course_discovery/apps/course_metadata/utils.py
mikix: Also... is this class big enough / distinct enough that we should throw it in a `salesforce.py` file? No real opinion here, just our models.py and utils.py files are so large.

jlajoie: I'm cool with it, do you have preferences on where this lives? Just same base directory as utils.py?

mikix: Yeah seems good?

course_discovery/apps/course_metadata/utils.py
mikix: Ah ok. So the usage would be like:

```
sf_util = SalesforceUtil(request.site.partner)
if sf_util.salesforce_is_enabled():
    ...
```

That feels reasonable. (Though maybe make salesforce_is_enabled into an `@property` method called like `enabled` or something?) 

The only oddity would be the exception that gets raised if users failed to guard their code would be AttributeError or something, which might be confusing for a second. To dodge that, you'd like... make your own exception and throw that intentionally? Not clear if that's worth the bother.

jlajoie: Yeah that's what I'm leaning toward, I think I'll make my own exception class, ideally never gets hit but good to be defensive for other people that might use the SF code

course_discovery/apps/course_metadata/tests/test_salesforce.py
```diff
@@ -0,0 +1,121 @@
...
+        self.salesforce_config = SalesforceConfigurationFactory()
+
+    def tearDown(self):
+        # Zero out the instances that are created during testing
+        SalesforceUtil.instances = {}
+
+    def test_login(self):
+        # Update the config to reflect what we'll run locally
+        self.salesforce_config.security_token = ''
+        self.salesforce_config.save()
```
mikix, line 22: Maybe just change the factory then?

course_discovery/apps/course_metadata/tests/test_salesforce.py
```diff
@@ -0,0 +1,121 @@
...
+
+        with mock.patch('course_discovery.apps.course_metadata.salesforce.Salesforce') as MockSalesforce:
+            SalesforceUtil(self.salesforce_config.partner)
+            MockSalesforce.assert_called_with(**{
+                'username': self.salesforce_config.username,
+                'password': self.salesforce_config.password,
+                'organizationId': self.salesforce_config.organization_id,
+                'security_token': '',
+                'domain': 'test',
```
mikix, line 31: You don't have tests that alter these two values (i.e. no tests with `is_sandbox = False` or an empty `organization_id` and thus a provided security token).

course_discovery/apps/course_metadata/tests/test_salesforce.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_get_account_by_name(self):
+        return_value = 'A record'
+        with mock.patch('course_discovery.apps.course_metadata.salesforce.Salesforce'):
+            with mock.patch(
+                    'course_discovery.apps.course_metadata.salesforce.SalesforceUtil._query',
+                    return_value={'records': [return_value]}
+            ) as mock_query:
+                util = SalesforceUtil(self.salesforce_config.partner)
+                accounts = util.get_account_by_name('Test')
+                mock_query.assert_called_with('SELECT Id,Name FROM Account WHERE Name=Test')
```
mikix, line 80: Small nit, but I am always nervous around the Mock assert methods. If a typo exists, you'll never get a failure. Like, 'mock_query.asert_called_with` would always just be happy.

I tend to use `self.assertEqual` along with `mock_query.call_count` or `mock_query.call_args` or some such. But it isn't as succinct. :-/  Carry on, just mentioning.

mikix (approved):
Sure! Looks good. I didn't really test or dive into the details of the queries you were making. I assume they're great. :)

jlajoie:
> Sure! Looks good. I didn't really test or dive into the details of the queries you were making. I assume they're great. :)

They'll change when IT finalizes the Records in SF anyways so didn't seem worth getting hung up on the content bodies 😄

course_discovery/apps/course_metadata/tests/test_salesforce.py
jlajoie: I also don't have a Salesforce environment that I could test those against to verify they work even in practice ☹️

course_discovery/apps/course_metadata/tests/test_salesforce.py
jlajoie: Good feedback, unsure if those other two methods would help the case of we're firing off to an external service. Maybe doing the call_args on something internal to the simple-salesforce package right before the request goes out might be the best approach?

PR #1974, opened 2019-06-25
Adds in revised CompressedCacheResponseMixin

bbaker6225:
The mixin will eventually compress response date before going into the
cache, in an effort to try and fix the memcached issues we are seeing
with large items. This first commit does not have compression, but
only attempts to safely decompress items from the cache, to avoid
issues with blue/green deployments. A follow up will be to actually add
compression once the safe decompression is deployed.

Also, only caches JSON responses to prevent users from seeing other
usernames in the cached browsable api responses

### DIFF
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -4,8 +4,9 @@ import zlib
 
 from django.conf import settings
 from django.core.cache import cache
+from rest_framework.renderers import JSONRenderer
 from rest_framework_extensions.cache.decorators import CacheResponse
-from rest_framework_extensions.key_constructor.bits import KeyBitBase, QueryParamsKeyBit, UserKeyBit
+from rest_framework_extensions.key_constructor.bits import KeyBitBase, QueryParamsKeyBit
 from rest_framework_extensions.key_constructor.constructors import (
     DefaultListKeyConstructor, DefaultObjectKeyConstructor
 )
@@ -28,10 +29,6 @@ class TimestampedListKeyConstructor(DefaultListKeyConstructor):
     # cache collisions when other query params are involved. For more, see:
     # https://github.com/chibisov/drf-extensions/blob/master/rest_framework_extensions/key_constructor/bits.py#L48-L49
     querystring = QueryParamsKeyBit()
-    # The UserKeyBit ensures that responses are only cached on a per-user basis,
-    # to avoid the issue where another user's username will appear in the api view
-    # rendered response, even though you are not actually the other user.
-    user = UserKeyBit()
 
 
 class TimestampedObjectKeyConstructor(DefaultObjectKeyConstructor):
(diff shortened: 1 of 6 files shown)


course_discovery/apps/api/v1/tests/test_cache.py
```diff
@@ -0,0 +1,68 @@
...
+from rest_framework import permissions, views
+from rest_framework.response import Response
+from rest_framework_extensions.test import APIRequestFactory
+
+from api.cache import compressed_cache_response
+
+factory = APIRequestFactory()
+
+
+class CompressedCacheResponseTest(TestCase):
```
bbaker6225, line 15: Basic test structure copied from https://github.com/chibisov/drf-extensions/blob/0.3.1/tests_app/tests/unit/cache/decorators/tests.py#L134

course_discovery/apps/api/cache.py
```diff
@@ -25,13 +28,21 @@ class TimestampedListKeyConstructor(DefaultListKeyConstructor):
...
 
 class TimestampedObjectKeyConstructor(DefaultObjectKeyConstructor):
     timestamp = ApiTimestampKeyBit()
     # The DefaultObjectKeyConstructor doesn't include querystring parameters
     # in its cache key.
     querystring = QueryParamsKeyBit()
+    # The UserKeyBit ensures that responses are only cached on a per-user basis,
+    # to avoid the issue where another user's username will appear in the api view
+    # rendered response, even though you are not actually the other user.
+    user = UserKeyBit()
```
mikedikan: this will create new cache entries, are we worried about overall cache size?

mikedikan (approved):

course_discovery/apps/api/cache.py
bbaker6225: Not particularly worried, but I swapped the approach to only cache json responses. I can change it back if people prefer the `UserKeyBit` approach, but I think only caching json (but for all authenticated users) is more useful

jmyatt: yes I think caching only json responses makes the most sense - simpler and handles the case for which we're trying to improve caching

course_discovery/apps/api/cache.py
```diff
@@ -61,3 +65,65 @@ def api_change_receiver(sender, **kwargs):  # pylint: disable=unused-argument
...
+        )
+        response = self.cache.get(key)
+
+        if not response:
+            response = view_method(view_instance, request, *args, **kwargs)
+            response = view_instance.finalize_response(request, response, *args, **kwargs)
+            response.render()  # should be rendered, before pickling while storing to cache
+
+            if (not (response.status_code >= 400 or self.cache_errors)
+                and type(response.accepted_renderer) == JSONRenderer):
```
jmyatt: indent

jmyatt (approved):

PR #1975, opened 2019-06-25
DISCO-1027: Management command for migrating course editors

Dillon-Dumesnil:


course_discovery/apps/course_metadata/utils.py
```diff
@@ -168,6 +168,12 @@ def ensure_draft_world(obj):
         draft_course, original_course = set_draft_state(obj, Course)
         draft_course.slug = original_course.slug
 
+        # Move editors from the original course to the draft course since we only care about CourseEditors
+        # in the context of draft courses.
```
mikix: Maybe a mention in this comment that this code is support a transition from old publisher to new, so that someone down the line can feel comfortable removing it.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2319,6 +2319,13 @@ class DrupalPublishUuidConfig(SingletonModel):
     push_people = models.BooleanField(default=False)
 
 
+class MigrateCourseEditorsConfig(SingletonModel):
+    """
+    Configuration for the migrate_course_editors command.
+    """
+    org_keys = models.TextField(default=None, null=False, blank=True, verbose_name=_('Organization Keys'))
```
mikix: `null=False` is a default anyway, I think.

`default=None` with `null=False` is confusing. Maybe `default=''`?

Maybe add `help_text` explaining format of this field?

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
```diff
@@ -0,0 +1,51 @@
...
+    help = _('Based on the configuration object, migrates all of the course editors for an organization '
+             'from publisher tables to course metadata tables for use in Publisher Frontend.')
+
+    def handle(self, *args, **options):
+        config = MigrateCourseEditorsConfig.get_solo()
+
+        if config.org_keys:
+            COURSE_TEAM_ROLE = 'course_team'
+            UserModel = get_user_model()
+            org_keys = config.org_keys.split(',')
```
mikix: That's pretty restrictive? Can't use any whitespace this way. (Though you could fix that by calling `strip` on any org keys before using them.)

You could also split on commas AND whitespace, so that input keys could just be space separated (or line separated) by doing something fancy with the `re` module:

```
>>> re.findall(r'[^,\s]+', '1 , 2\n\n3, 4 ')
['1', '2', '3', '4']
```

This doesn't super matter, as long as whatever format we use is mentioned in the django admin. People can massage input as needed.

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Command(BaseCommand):
+    help = _('Based on the configuration object, migrates all of the course editors for an organization '
+             'from publisher tables to course metadata tables for use in Publisher Frontend.')
+
+    def handle(self, *args, **options):
+        config = MigrateCourseEditorsConfig.get_solo()
+
+        if config.org_keys:
+            COURSE_TEAM_ROLE = 'course_team'
```
mikix: `PublisherUserRole.CourseTeam` yeah?

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
```diff
@@ -0,0 +1,51 @@
...
+                    org = Organization.objects.get(key=org_key)
+                except Organization.DoesNotExist:
+                    logger.exception(_(
+                        'Organization key [{org_key}] is not a valid key for any existing organization.'.format(
+                            org_key=org_key
+                        )
+                    ))
+                    continue
+                for discovery_course in Course.objects.filter(authoring_organizations__contains=org):
+                    course_number = discovery_course.key.split('+')[-1]
```
mikix: No old format keys to worry about?

Do all courses in prod have `key` filled in? I think that's an optional field? But I guess if not, nothing we could do here anyway. Probably all publisher courses have it filled...

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
```diff
@@ -0,0 +1,51 @@
...
+            for org_key in org_keys:
+                try:
+                    org = Organization.objects.get(key=org_key)
+                except Organization.DoesNotExist:
+                    logger.exception(_(
+                        'Organization key [{org_key}] is not a valid key for any existing organization.'.format(
+                            org_key=org_key
+                        )
+                    ))
+                    continue
```
mikix, line 53: Might be worth failing the job (at the end) if there were any exceptions? Makes it more likely they'd be noticed.

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class Command(BaseCommand):
+    help = _('Based on the configuration object, migrates all of the course editors for an organization '
+             'from publisher tables to course metadata tables for use in Publisher Frontend.')
+
+    def handle(self, *args, **options):
+        config = MigrateCourseEditorsConfig.get_solo()
+
+        if config.org_keys:
```
mikix: You could unindent this whole file one level by just early exiting (or actually, warning and failing?) if no org keys are defined.

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
Dillon-Dumesnil: Shockingly no. Some old Course runs still have the "/", but all courses in course_metadata use "+"

Yes, all courses in prod have it filled in. I think it is required.
There are 4101 courses in course_metadata. There are 3776 courses in Publisher and 3381 of those have number defined.
Looks like there are 395 courses in publisher that do not have a `number` defined. I think it should all be fine though because anything that only exists in course_metadata can't have any course team members so there isn't anybody to move to CourseEditor

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
Dillon-Dumesnil: 👍

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
Dillon-Dumesnil: Oooh nice.

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
Dillon-Dumesnil: Although, in light of the number of course differences between course_metadata and publisher, I hesitate to add in a condition for failure when for some organizations, it will always fail since they will run into PublisherCourse.DoesNotExist. Thoughts?

mikix: Maybe that one shouldn't be fatal then.

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
```diff
@@ -0,0 +1,71 @@
...
+                    course_id=publisher_course, role=PublisherUserRole.CourseTeam
+                ):
+                    user_id = user_role.user_id
+                    user = UserModel.objects.get(pk=user_id)
+                    # Using update_or_create in case a course has multiple authoring organizations and the course
+                    # shows up more than once.
+                    if discovery_course.draft_version:
+                        CourseEditor.objects.update_or_create(course=discovery_course.draft_version, user=user)
+                    else:
+                        CourseEditor.objects.update_or_create(course=discovery_course, user=user)
```
mikix: nit: this could be less repetitive by doing:
```
editor_course = discovery_course.draft_version or discovery_course
CourseEditor.objects.update_or_create(course=editor_course, user=user)
```

This section could also use a comment along the lines of "we choose to link to an official version rather than create a draft, even though editors don't make sense on official versions. we move them when creating drafts in ensure_draft_world" or something like that.

mikix (approved):
Nice

course_discovery/apps/course_metadata/management/commands/migrate_course_editors.py
Dillon-Dumesnil: I meant to switch that up, but couldn't think of the correct syntax. Thanks!

PR #1979, opened 2019-06-25
Add management command to redirect old course runs to new ones

mikix:
Add a new Course.update_marketing_redirects method and a new management command that calls it for all courses.

This code will keep a drupal marketing site's course redirects up to date, as course runs become archived over time.

Ideally, this would be run as part of a regularly scheduled job.
    
https://openedx.atlassian.net/browse/DISCO-898


course_discovery/apps/course_metadata/models.py
```diff
@@ -623,9 +624,10 @@ def marketing_course_runs(self):
         """
         marketable_runs = []
         other_marketable_runs = []
+        now = datetime.datetime.now(pytz.UTC)
 
         for course_run in self.course_runs.all():
-            if course_run.is_marketable and course_run.availability.lower() != 'archived':
+            if course_run.is_marketable and not course_run.has_ended(now):
```
mikix, line 630: This was a straight up bug ('Archived' is a translated string that we are comparing with an English one).

course_discovery/apps/course_metadata/models.py
```diff
@@ -966,12 +1007,8 @@ def get_paid_seat_enrollment_end(self):
...
             return None
 
-        # An unenrolled user may not enroll and purchase paid seats after the course has ended.
-        deadline = self.end
-
-        # An unenrolled user may not enroll and purchase paid seats after enrollment has ended.
-        if self.enrollment_end and (deadline is None or self.enrollment_end < deadline):
-            deadline = self.enrollment_end
+        # An unenrolled user may not enroll and purchase paid seats after the course or enrollment has ended.
+        deadline = self.enrollment_deadline
```
mikix, line 1011: You'll see a lot of this style of deadline cleanup in this file. I did this to try to consolidate the logic that was duplicated a lot around the place. That logic isn't hard, but still, a little easy to mess up.

Specifically, I started duplicating the "is end or enrollment_end past" logic that was in the Course.active_course_runs method. That logic is in Queryset terms, but I wanted a Python version. And to help avoid drift, I figured all the Python versions should be in one place.

PLEASE look carefully at each of these tiny cleanups. It'd be easy for me to mess one of them up.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1354,11 +1418,8 @@ def is_enrollable(self):
         fields.
         """
         now = datetime.datetime.now(pytz.UTC)
-        if ((not self.enrollment_end or self.enrollment_end >= now) and
-                (not self.enrollment_start or self.enrollment_start <= now)):
-            return True
-
-        return False
+        return (not self.has_enrollment_ended(now) and
+                (not self.enrollment_start or self.enrollment_start <= now))
```
mikix, line 1420: This is a behavior change -- we didn't use to check `end` here, which `has_enrollment_ended` does do. But this `is_enrollable` property is only used in one non-test place, in this same file, after a separate check against has_ended. So it's a change that doesn't affect any actual behavior.

And I think it's a sensible semantic change. So I left it in.

course_discovery/apps/course_metadata/management/commands/update_marketing_redirects.py
```diff
@@ -0,0 +1,52 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Redirect marketing site URLs from any old inactive course runs to newer active runs'
+
+    def handle(self, *args, **options):
+        success = True
+
+        # Since we know we will call update_marketing_redirects for every single course in our catalog, let's try to
```
staubina: Nit: this is not entirely true because you are filtering out only `Published` course runs.

course_discovery/apps/course_metadata/models.py
```diff
@@ -674,6 +678,43 @@ def first_enrollable_paid_seat_price(self):
...
+
+        if published_runs is None:
+            published_runs = self.course_runs.filter(status=CourseRunStatus.Published)
+        published_runs = frozenset(published_runs)
+
+        # Now separate out the active ones from the inactive
+        # (done in Python rather than hitting self.active_course_runs to avoid a second db query)
+        now = datetime.datetime.now(pytz.UTC)
+        inactive_runs = {run for run in published_runs if run.has_enrollment_ended(now)}
+        active_runs = published_runs - inactive_runs
```
staubina, line 703: I am still pulling this apart, but since this is just more filtering could we add to the DB filter above and complete this all in 1 DB query?

course_discovery/apps/course_metadata/models.py
```diff
@@ -674,6 +678,43 @@ def first_enrollable_paid_seat_price(self):
...
+        if not active_runs or not inactive_runs:
+            return False  # if there are no runs to redirect from or to, there's no point in continuing
```
staubina, line 705: It seems like this might be worth logging.

course_discovery/apps/course_metadata/models.py
```diff
@@ -674,6 +678,43 @@ def first_enrollable_paid_seat_price(self):
...
+
+        # Find the earliest active run, to which we will redirect any inactive runs
+        earliest_active_run = min(active_runs, key=attrgetter('start'))
+
+        publisher = CourseRunMarketingSitePublisher(self.partner)
+        for run in inactive_runs:
```
staubina, line 711: This seems like it would be valuable to log as we are calling an external service. It might be useful to know when debugging.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1331,6 +1366,9 @@ def publish(self):
                 run.status = CourseRunStatus.Published
                 run.save()
 
+        # It is likely that we are sunsetting an old run in favor of this new run, so update redirects just in case
+        self.course.update_marketing_redirects()
```
staubina, line 1368: 👍

course_discovery/apps/course_metadata/models.py
staubina: Looking at this again I think I see now why you are doing it in Python.

staubina (commented):
Would it be worthwhile to add a little logging to make the management command easier to debug in Production?

Overall the PR looks good to me.

course_discovery/apps/course_metadata/models.py
mikix: The redirect call does its own logging

course_discovery/apps/course_metadata/models.py
mikix: Eh, that's a pretty common case, at least in the management command case - it's how we sort out which courses need redirection and which don't. Feels noisy to log each thing we don't do?

course_discovery/apps/course_metadata/models.py
mikix: Just to follow up for other eyeballs - we can't do this filtering in the same DB query, because we want both sides of the coin - we want to split into active and inactive, but we do look at both sets.

staubina (approved):

course_discovery/apps/course_metadata/management/commands/update_marketing_redirects.py
```diff
@@ -0,0 +1,52 @@
...
+            success = self.update_course(current_course, current_runs) and success
+
+        if not success:
+            raise CommandError(_('One or more courses failed to redirect.'))
+
+    @staticmethod
+    def update_course(course, runs):
+        try:
+            if course.update_marketing_redirects(published_runs=runs):
+                logger.info(_('Successfully redirected runs in course {key}').format(key=course.key))
```
Dillon-Dumesnil, line 48: Should the `{key}` be `[{key}]`? I remember you saying something once about that helping out translators or something.

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: 🙏

course_discovery/apps/course_metadata/models.py
```diff
@@ -996,15 +1033,12 @@ def enrollable_seats(self, types=None):
         now = datetime.datetime.now(pytz.UTC)
         enrollable_seats = []
 
-        if self.end and now > self.end:
+        if self.has_enrollment_ended(now):
```
Dillon-Dumesnil: Nit: Combine this with the lower check.

Nit: just return [] instead of enrollable_seats. It feels more explicit to me that way and I had a moment of confusion on why you're returning enrollable seats if there shouldn't be any and had to remember enrollable_seats is []

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -221,6 +222,125 @@ def test_marketing_course_runs_no_open_enrollment(self):
...
+        self.inactive = factories.CourseRunFactory(**self.base_args, end=self.past)
+        self.api_root = self.partner.marketing_site_url_root.rstrip('/')  # overwrite the mixin's version
+
+    def assertRedirect(self, published_runs=None, succeed=True, to_run=None, from_runs=None):
+        """
+        Args:
+            published_runs: Runs to pass to the redirect call
+            succeed: Whether the redirect should return True or False
+            to_run: A CourseRun that all from_runs are redirected to
+            from_run: An iterable of CourseRuns you expect to be redirected from
```
Dillon-Dumesnil: Nit and only fix if something else needs fixing: `from_run` -> `from_runs`

Dillon-Dumesnil (approved):
Couple nits, but nothing major. Looks awesome. Make sure you tell Marketing! I'm sure their going to be thrilled

course_discovery/apps/course_metadata/management/commands/update_marketing_redirects.py
mikix: Hmm, I might have said something about marking it off with quotes or something because like in some languages where like, the starting letter of `key` might affect the preceding word or something, it will read more naturally. But actually, I suppose in those languages, the translator can add quotes as necessary.

course_discovery/apps/course_metadata/models.py
mikix: Agreed. Will do.

course_discovery/apps/course_metadata/tests/test_models.py
mikix: Thanks!

PR #1980, opened 2019-06-26
Change CompressedCacheMixin strategy

bbaker6225:
Changes the compression/caching strategy to separate the response into
a tuple, and compress the rendered content, before reassembling the
response when it comes out of the cache. This is done because rendered
content can't actually be set.

The CompressedCacheResponseMixin is removed from all endpoints where
it is used so that we can remove all keys from the cache, then add the
mixin back in, hopefully preventing any issues with different data
formats in the cache.

### DIFF
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -4,6 +4,7 @@ import zlib
 
 from django.conf import settings
 from django.core.cache import cache
+from django.http.response import HttpResponse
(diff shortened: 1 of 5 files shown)


jmyatt (approved):
discussed the approach with Brandon - looks good.  We should be fine while in the state where the cache is not being used, and the followup PR to add the new mixin back to these endpoints will be straightforward after memcached is cleared.

PR #1983, opened 2019-06-27
Remove publish date validations

ciduarte:
Publish date no longer *has* to be prior to
start date, this happens with self-paced courses

DISCO-1097


Dillon-Dumesnil (approved):
LGTM

PR #1985, opened 2019-07-01
DISCO-1082: Ensure audit entitlement is created

Dillon-Dumesnil:
When an official course has no entitlement, it might be an audit course
and so we will set a default of creating a draft audit entitlement. This
happens as part of ensure_draft_world and now we ensure the update code
does not error if no entitlement data is passed in.


course_discovery/apps/course_metadata/utils.py
```diff
@@ -190,7 +190,7 @@ def ensure_draft_world(obj):
             for entitlement in original_course.entitlements.all():
                 set_draft_state(entitlement, CourseEntitlement, {'course': draft_course})
         else:
-            mode = SeatType.objects.get(slug='audit')
+            mode = SeatType.objects.get(slug=Seat.AUDIT)
```
mikix, line 193: Nice catch

mikix (approved):

PR #1986, opened 2019-07-01
Avoid duplicate editor courses

mikix:
When querying against the list of course editors, it can be easy for Django to return multiple objects across the many joins that query performs.

This will make sure that CourseEditor class methods always return distinct object sets.

https://openedx.atlassian.net/browse/DISCO-1134

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -778,12 +778,14 @@ class CourseEditor(TimeStampedModel):
         # for it. This handles cases of being dropped from an org... But might be too restrictive in case we want
         # to allow outside guest editors on a course? Let's try this for now and see how it goes.
         valid_editors = course.editors.filter(user__groups__organization_extension__organization__in=authoring_orgs)
+        valid_editors = valid_editors.select_related('user')
 
         if valid_editors:
             return {editor.user for editor in valid_editors}
 
         # No valid editors - this is an edge case where we just grant anyone in an authoring org access
-        return get_user_model().objects.filter(groups__organization_extension__organization__in=authoring_orgs)
+        user_model = get_user_model()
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -778,12 +778,14 @@ def course_editors(cls, course):
         # for it. This handles cases of being dropped from an org... But might be too restrictive in case we want
         # to allow outside guest editors on a course? Let's try this for now and see how it goes.
         valid_editors = course.editors.filter(user__groups__organization_extension__organization__in=authoring_orgs)
+        valid_editors = valid_editors.select_related('user')
```
mikix, line 781: This was an unrelated performance improvement, once I started measuring queries for this method in the tests.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -354,53 +356,88 @@ def setUp(self):
...
+        self.assertResultsEqual(self.filter_editable_course_runs, {self.run_bad_editor_in_group, self.run_editor},
+                                queries=1)
+
+    def test_course_editors_when_valid_editors(self):
+        self.assertResultsEqual(partial(CourseEditor.course_editors, self.course_editor), {self.user}, queries=1)
+
+    def test_course_editors_when_no_editors(self):
+        # two queries: one to check for valid editors, one for everybody in group
+        self.assertResultsEqual(partial(CourseEditor.course_editors, self.course_bad_editor_in_group),
+                                {self.user, self.good_editor}, queries=2)
```
mikix, line 440: We didn't test the `editable_course_runs` or `course_editors` methods in this test suite before, so I added some more (unrelated to my purpose) methods.

But those same methods now also check for duplicates via the `assertResultsEqual` helper method.

staubina (approved):

Dillon-Dumesnil (approved):
LGTM

PR #1989, opened 2019-07-02
Save image separate from course run creation

Dillon-Dumesnil:
We don't want to rollback discovery on failure of uploading an image to
studio and instead only rollback on course run creation failing.

DISCO-1137 and DISCO-1138


course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -350,11 +350,16 @@ def test_partial_update_no_studio_url(self):
             response = self.client.patch(url, {'start': '2019-01-01T00:00:00Z'}, format='json')
 
         self.assertEqual(response.status_code, 200, "Status {}: {}".format(response.status_code, response.content))
-        mock_logger.assert_called_with(
+        mock_logger.assert_any_call(
```
mikix: You've heard my dislike for the mock assert calls before, yah? I won't repeat. :) Matter of taste.

course_discovery/apps/publisher/api/v1/views.py
```diff
@@ -62,7 +62,8 @@ def publish_to_studio(self, partner, course_run):
         api = StudioAPI(partner.studio_api_client)
 
         try:
-            api.push_to_studio(course_run)
+            course_run_response = api.push_to_studio(course_run)
+            api.update_course_run_image_in_studio(course_run, course_run_response)
```
mikix: We don't want to fix the bug here in old publisher too?

course_discovery/apps/publisher/signals.py
```diff
@@ -48,6 +48,7 @@ def create_course_run_in_studio_receiver(sender, instance, created, **kwargs):
 
         try:
             response = api.push_to_studio(instance, create=True, old_course_run=discovery_course_run)
+            api.update_course_run_image_in_studio(instance, response)
```
mikix: Or here?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -180,6 +180,16 @@ def push_to_studio(cls, request, course_run, create=False, old_course_run=None):
...
                      course_run.key, course_run.course.partner.short_code)
 
+    @classmethod
+    def update_course_run_image_in_studio(cls, course_run):
+        if course_run.course.partner.studio_url:
+            api = StudioAPI(course_run.course.partner.studio_api_client)
+            return api.update_course_run_image_in_studio(course_run)
+        else:
+            log.info('Not updating course run image for %s to Studio as partner %s has no studio_url set.',
+                     course_run.key, course_run.course.partner.short_code)
```
mikix, line 190: Not translated?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -217,7 +217,9 @@ def create(self, request, *args, **kwargs):
                 'price': price,
                 'mode': course_creation_fields['mode'],
             })
-            CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
+            course_run_viewset = CourseRunViewSet()
+            __, course_run = course_run_viewset.create_run_helper(course_run_creation_fields, request)
+            course_run_viewset.push_image_to_studio(course_run)
```
mikix: Same problem here too? Heck though, in new publisher we don't send an image on create, so I guess this never really does anything useful?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
Dillon-Dumesnil: I don't think I have... What's the issue?

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: That's actually a big help knowing that. I'll just not call it here.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
mikix: Ah, just that typos are problematic. Since mocks take any unknown attribute and return a new mock, if you do `mock_logger.asert_any_call` accidentally, you'd never know, and you wouldn't be asserting anything.

So as useful as the mock assert utiities are, I prefer to do `self.assertEqual(mock_logger.call_count, 1)` or similar with `call_args`. Less pretty, but I am so fearful of my fat fingers, I feel better off.

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: It wasn't translated in the push_to_studio method above so I wasn't planning on translating this one.. Maybe they should just both be translated?

course_discovery/apps/publisher/signals.py
```diff
@@ -58,6 +58,15 @@ def create_course_run_in_studio_receiver(sender, instance, created, **kwargs):
...
         instance.save()
 
+        try:
+            api.update_course_run_image_in_studio(instance, response)
+        except SlumberBaseException as ex:
+            logger.exception('Failed to update image for course run [%s] on Studio: %s. Course run creation was still '
+                             'successful. Please try resaving the image.', course.key, ex.content)
+        except Exception:  # pylint: disable=broad-except
+            logger.exception('Failed to update image for course run [%s] on Studio, but course run creation was still '
+                             'successful. Please try resaving the image.', course.key)
```
mikix: Should we just move that logging into the call? When wouldn't we want it?

mikix (approved):

mikix (approved):
Nice

PR #1990, opened 2019-07-02
Unpublish reviewed course runs upon course edit

jmyatt:
DISCO-1080

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -704,6 +704,33 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         self.assertEqual(self.course.title, 'Fake Test')
 
     @responses.activate
+    def test_patch_resets_run_status(self):
+        self.mock_access_token()
+        self.mock_ecommerce_publication()
+        self.create_course_and_course_run()
+
+        draft_course = Course.everything.last()
+        draft_course_run = CourseRun.everything.last()
+        draft_course_run.status = CourseRunStatus.Reviewed  # Triggers creation of official versions
+        draft_course_run.save()
+        official_course_run = draft_course_run.official_version
+        self.assertEqual(official_course_run.status, CourseRunStatus.Reviewed)
+
+        url = reverse('api:v1:course-detail', kwargs={'key': draft_course.uuid})
+        patch_data = {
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -300,6 +302,15 @@ def update_course(self, data, partial=False):
                     # This will also update the course
                     course_run.update_or_create_official_version()
 
+        # Revert any Reviewed course runs back to Unpublished
+        if changed:
+            for course_run in course.course_runs.filter(status=CourseRunStatus.Reviewed):
+                course_run.status = CourseRunStatus.Unpublished
+                course_run.save()
+                if course_run.official_version and course_run.official_version.status == CourseRunStatus.Reviewed:
```
Dillon-Dumesnil: Should be unnecessary to do `course_run.official_version` since we create the official version when a course run goes to Reviewed. Not a big deal.

On second thought, I might just get rid of this if statement altogether. There isn't really a reason to check the official versions status. They should have parity so if we are moving the draft, let's move the official too.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -703,6 +703,29 @@ def test_update_operates_on_drafts(self):
         self.assertFalse(self.course.entitlements.first().draft)
         self.assertEqual(self.course.title, 'Fake Test')
 
+    @responses.activate
+    def test_patch_resets_run_status(self):
+        self.mock_access_token()
+        self.mock_ecommerce_publication()
```
Dillon-Dumesnil, line 709: This is potentially concerning. I don't think there should be any writing to ecommerce for this kind of change.

Dillon-Dumesnil (approved):
Couple small comments. Looks good! Nice job Jason :)

course_discovery/apps/api/v1/tests/test_views/test_courses.py
jmyatt: the call to ecommerce (/api/v2/publication/) is coming from here: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L1320

Dillon-Dumesnil: Ah I see. It's there because of the creation of official objects by saving a course run in Reviewed. Sounds good!

PR #1993, opened 2019-07-03
Update upgrade deadline in publisher seats

uzairr:
Currently, update in deadline date in course metadata seats has no
effect on the seats in publisher.To make the publisher seats coherent
with the course metadata save method is overriden which will update
the deadline date in publisher seats.

PROD-423

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1450,6 +1450,10 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
 
 class Seat(DraftModelMixin, TimeStampedModel):
     """ Seat model. """
+    def __init__(self, *args, **kwargs):
+        super(Seat, self).__init__(*args, **kwargs)
+        self.__original_upgrade_deadline = self.upgrade_deadline
+
     HONOR = 'honor'
     AUDIT = 'audit'
     VERIFIED = 'verified'
@@ -1502,6 +1506,21 @@ class Seat(DraftModelMixin, TimeStampedModel):
             ('course_run', 'type', 'currency', 'credit_provider', 'draft')
         )
         ordering = ['created']
+
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        super(Seat, self).save(*args, **kwargs)
+        if self.upgrade_deadline and self.upgrade_deadline != self.__original_upgrade_deadline:
+            # Avoid circular imports
+            from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun
+            try:
+                run = PublisherCourseRun.objects.get(lms_course_id=self.course_run.key)
+                seat = run.seats.get(type=self.type)
+                seat.upgrade_deadline = self.upgrade_deadline
+                seat.save()
+            except ObjectDoesNotExist as exc:
+                logger.warning(
+                    'An Error occurred while updating upgrade deadline in publisher\'s seat: {}'.format(str(exc))
+                )
 
 
 class CourseEntitlement(DraftModelMixin, TimeStampedModel):
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -1002,6 +1002,26 @@ class CourseRunTests(OAuth2Mixin, TestCase):
 
         self.assertEqual(course_run.is_marketable, expected)
 
+    def test_sync_upgrade_deadline_with_publisher_seat(self):
+        key = 'course-run-id/nKWVadtfUwfX/fake'
+        CourseRun.objects.create(course=self.course_run.course, key=key)
+        verified_seat = Seat.objects.create(type='verified', course_run=self.course_run, price=100)
+
+        # Avoid circular import
+        from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun, \
+            Seat as PublisherSeat, Course as PublisherCourse
+
+        publisher_run = PublisherCourseRun.objects.create(
+            lms_course_id=self.course_run.key, course=PublisherCourse.objects.create()
+        )
+        PublisherSeat.objects.create(type='verified', course_run=publisher_run, price=10)
+        verified_seat.upgrade_deadline = datetime.datetime.utcnow()
+        verified_seat.save()
+
+        verified_seat = Seat.objects.first()
+        publisher_seat = PublisherSeat.objects.first()
+        self.assertEqual(verified_seat.upgrade_deadline, publisher_seat.upgrade_deadline)
+
 
 @ddt.ddt
 class OrganizationTests(TestCase):


course_discovery/apps/course_metadata/models.py
```diff
@@ -1497,6 +1497,20 @@ class Meta(object):
...
+        try:
+            # Avoid circular imports
+            from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun
+            run = PublisherCourseRun.objects.get(lms_course_id=self.course_run.key)
+        except PublisherCourseRun.DoesNotExist:
+            return
+        seats = run.seats.all()
+        for seat in seats:
+            if seat.type != 'audit':
+                seat.upgrade_deadline = self.upgrade_deadline
```
waheedahmed, line 1518: What if `self.upgrade_deadline` is `None` or someone is changing something else is seat object, we shouldn't make these DB calls if not needed.

Add a check after super call if upgrade_deadline is changed then do the rest.

waheedahmed (changes requested):
Also add some tests.

course_discovery/apps/course_metadata/models.py
waheedahmed: Also, is this required on other seat types as well e.g. credit, no-id-professional etc?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1497,6 +1497,19 @@ class Meta(object):
         )
         ordering = ['created']
 
+    def save(self, *args, **kwargs):
+        super(Seat, self).save(*args, **kwargs)
+        if self.upgrade_deadline:
```
waheedahmed: Is there any way to somehow check here if the value is changed for `upgrade_deadline`, because this check will always gets `True` if `upgrade_deadline` is not `NULL`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1497,6 +1497,25 @@ class Meta(object):
         )
         ordering = ['created']
 
+    def save(self, *args, **kwargs):
+        try:
+            old_seat = Seat.objects.get(id=self.id)
```
waheedahmed: I believe you can do following to identify if the value is changed or not instead of querying again:
```
    super(Seat, self).__init__(*args, **kwargs)
    self.__original_upgrade_deadline = self. upgrade_deadline
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -1497,6 +1497,25 @@ class Meta(object):
...
+            super(Seat, self).save(*args, **kwargs)
+        except ObjectDoesNotExist:
+            super(Seat, self).save(*args, **kwargs)
+            return
+
+        if self.upgrade_deadline and self.upgrade_deadline != old_seat.upgrade_deadline:
```
waheedahmed: And do this here:
```
if self.upgrade_deadline and self.upgrade_deadline != self.__original_upgrade_deadline:

File: course_discovery/apps/course_metadata/models.py [hunk 5]
```diff
@@ -1497,6 +1497,25 @@ class Meta(object):
...
+            # Avoid circular imports
+            from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun
+            try:
+                run = PublisherCourseRun.objects.get(lms_course_id=self.course_run.key)
+                seat = run.seats.get(type=self.type)
```
waheedahmed, line 1517: What if somehow there is no seat object for this in publisher? Do we need try except here?

waheedahmed (changes requested):

course_discovery/apps/course_metadata/models.py
```diff
@@ -1498,6 +1502,19 @@ class Meta(object):
...
+        if self.upgrade_deadline and self.upgrade_deadline != self.__original_upgrade_deadline:
+            # Avoid circular imports
+            from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun
+            try:
+                run = PublisherCourseRun.objects.get(lms_course_id=self.course_run.key)
+                seat = run.seats.get(type=self.type)
+                seat.upgrade_deadline = self.upgrade_deadline
+                seat.save()
+            except ObjectDoesNotExist:
+                return
```
waheedahmed: Shouldn't we log some kind of warning message here for debuging?

waheedahmed: Also why we need return statement here?

waheedahmed (approved):
Couple of questions, 👍 once addressed.

Also get it reviewed by someone from discovery-squad as well.

course_discovery/apps/course_metadata/models.py
uzairr: Thanks for pointing out! I added a warning log message instead of returning.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1450,6 +1450,10 @@ def is_marketable(self):
 
 class Seat(DraftModelMixin, TimeStampedModel):
     """ Seat model. """
+    def __init__(self, *args, **kwargs):
+        super(Seat, self).__init__(*args, **kwargs)
```
mikix, line 1454: extremely minor nit, but Disco is py3 only, so you can just do `super().__init__(*args, **kwargs)` (no need for `Seat, self`). Same comment below in `save`.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1002,6 +1002,26 @@ def test_is_marketable(self, status, create_seats, create_marketing_url, expecte
...
         self.assertEqual(course_run.is_marketable, expected)
 
+    def test_sync_upgrade_deadline_with_publisher_seat(self):
+        key = 'course-run-id/nKWVadtfUwfX/fake'
+        CourseRun.objects.create(course=self.course_run.course, key=key)
+        verified_seat = Seat.objects.create(type='verified', course_run=self.course_run, price=100)
+
+        # Avoid circular import
+        from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun, \
+            Seat as PublisherSeat, Course as PublisherCourse
```
mikix, line 1012: Circular import? The publisher model code ends up calling into test_models?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1503,6 +1507,21 @@ class Meta(object):
         )
         ordering = ['created']
 
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
```
mikix, line 1510: You can drop `*args` and then the `arguments-differ` exception. `save` takes no arguments.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1503,6 +1507,21 @@ class Meta(object):
...
+            try:
+                run = PublisherCourseRun.objects.get(lms_course_id=self.course_run.key)
+                seat = run.seats.get(type=self.type)
+                seat.upgrade_deadline = self.upgrade_deadline
+                seat.save()
+            except ObjectDoesNotExist as exc:
+                logger.warning(
+                    'An Error occurred while updating upgrade deadline in publisher\'s seat: {}'.format(str(exc))
+                )
+
```
mikix, line 1524: You probably want to do `self.__original_upgrade_deadline = self.upgrade_deadline` at the bottom of this method. And maybe just call it `_old_upgrade_deadline` since you just care about the most recent value, not the original one. (And one underscore instead of two - one is for private stuff, two is for python stuff).

course_discovery/apps/course_metadata/models.py
```diff
@@ -1503,6 +1507,21 @@ class Meta(object):
...
+        if self.upgrade_deadline and self.upgrade_deadline != self.__original_upgrade_deadline:
+            # Avoid circular imports
+            from course_discovery.apps.publisher.models import CourseRun as PublisherCourseRun
+            try:
+                run = PublisherCourseRun.objects.get(lms_course_id=self.course_run.key)
+                seat = run.seats.get(type=self.type)
+                seat.upgrade_deadline = self.upgrade_deadline
+                seat.save()
+            except ObjectDoesNotExist as exc:
+                logger.warning(
```
mikix, line 1521: `logger.exception` would cause the stacktrace to be printed, which is usually helpful for figuring out where in the code a message is coming from / why it's happening.

But more importantly, why is it an error if the publisher model doesn't exist? The seat could be from before publisher existed. The seat could be from the new publisher frontend, which doesn't keep its data in the old publisher models. I think you should just silently ignore it if no publisher model exists for this seat/run.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1503,6 +1507,21 @@ class Meta(object):
...
+                    'An Error occurred while updating upgrade deadline in publisher\'s seat: {}'.format(str(exc))
```
mikix, line 1522: I myself prefer to also translate error messages, although I know our code base is inconsistent in that regard.

mikix (commented):
This is not usually the direction in which we sync data. We're trying to build up publisher wherever we find instances like this of people needing to manually tweak data in django admin or whatever. We're trying to make the line be "make any course edits in publisher directly".

My preference would be to add an upgrade_deadline UI field if we need it, to publisher. But I understand if we don't want to invest in that.

Is the problem because someone's runbook for updating upgrade deadlines is to manually tweak the course in django admin? We probably need to update it to also mention the draft rows that new publisher creates in the Course Metadata tables too...

mikix:
Oh I see from the ticket - not really a runbook issue per se, but an RCM import issue. We'll still need to do something there for new-publisher. Like update draft row too if the seat changes. This code might make more sense outside of the `save` method and more in the ecom RCM code as a special case for that flow especially. Easier to reason about and safer to avoid updating both publishers when we shouldn't.

mikix:
Though I still stand by "we should just add this to publisher"

jmyatt:
+1 for moving this from `save` to the ecom RCM code

jmyatt:
let's hold off on this PR for now until we hear back on the latest questions in [the ticket](https://openedx.atlassian.net/browse/PROD-423)

PR #1995, opened 2019-07-03
Add endpoint for program data fixture

kdmccormick:
@edx/masters-neem 
https://openedx.atlassian.net/browse/EDUCATOR-4472

This is a staff-only endpoint that allows fetching of a fixture for up to 10 programs and all their related model instances. For use by Master's integration sandboxes to load production catalog data.

### DIFF
--- a/course_discovery/apps/edx_catalog_extensions/api/v1/views.py
+++ b/course_discovery/apps/edx_catalog_extensions/api/v1/views.py
@@ -1,6 +1,24 @@
+import logging
+from collections import defaultdict
+from uuid import UUID
+
+from django.contrib.sites.models import Site
+from django.core.serializers import json
+from django.http.response import HttpResponse
+from rest_framework.permissions import IsAdminUser
+from rest_framework.views import APIView
+
 from course_discovery.apps.api.v1.views.search import AggregateSearchViewSet
+from course_discovery.apps.core.models import Partner, User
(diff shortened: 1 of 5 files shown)


course_discovery/apps/edx_catalog_extensions/api/v1/views.py
```diff
@@ -13,3 +30,177 @@ def get_queryset(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+                pk for pk, instance in results_by_model[model].items()
+                if not instance
+            }
+            if pks_to_load:
+                break
+        else:
+            # All models fully loaded; return results flattened into a list.
+            return [
+                instance
+                for _mdl, results in results_by_model.items()
```
iloveagent57: if you don't need the keys, can you use `.values()` instead of `.items()`?  Same thing on the line directly below.

iloveagent57 (approved):
LGTM, only the most minor of suggestions.

course_discovery/apps/edx_catalog_extensions/api/v1/views.py
kdmccormick: Good call.

schenedx (approved):
👍

schenedx (approved):
Good bug catch

PR #1998, opened 2019-07-08
Create an entitlement for a course JIT

mikix:
When a GET request is made for a course, if it doesn't have an entitlement, but it could - we upgrade that course to use entitlements. This is new behavior we are testing for publisher-frontend.

https://openedx.atlassian.net/browse/DISCO-1018

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -46,7 +46,7 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         """ Verify the endpoint returns the details for a single course. """
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
 
-        with self.assertNumQueries(31):
+        with self.assertNumQueries(45):
             response = self.client.get(url)
             self.assertEqual(response.status_code, 200)
             self.assertEqual(response.data, self.serialize_course(self.course))
@@ -55,7 +55,7 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         """ Verify the endpoint returns the details for a single course with UUID. """
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.uuid})
 
-        with self.assertNumQueries(FuzzyInt(32, 2)):
+        with self.assertNumQueries(FuzzyInt(46, 2)):
             response = self.client.get(url)
             self.assertEqual(response.status_code, 200)
             self.assertEqual(response.data, self.serialize_course(self.course))
@@ -64,7 +64,7 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         """ Verify the endpoint returns no deleted associated programs """
         ProgramFactory(courses=[self.course], status=ProgramStatus.Deleted)
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
-        with self.assertNumQueries(18):
+        with self.assertNumQueries(31):
             response = self.client.get(url)
             self.assertEqual(response.status_code, 200)
             self.assertEqual(response.data.get('programs'), [])
@@ -77,7 +77,7 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         ProgramFactory(courses=[self.course], status=ProgramStatus.Deleted)
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
         url += '?include_deleted_programs=1'
-        with self.assertNumQueries(37):
+        with self.assertNumQueries(51):
             response = self.client.get(url)
             self.assertEqual(response.status_code, 200)
             self.assertEqual(
@@ -1003,3 +1003,25 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         self.assertEqual(data['subjects']['child']['choices'],
                          [{'display_name': 'Subject1', 'value': 'subject1'}])
         self.assertFalse('choices' in data['partner'])  # we don't whitelist partner to show its choices
+
+    @responses.activate
+    @ddt.data(True, False)
+    def test_retrieve_will_create_entitlement(self, has_entitlement):
+        """ When retrieving a course, test that an entitlement gets created if needed """
+        self.mock_access_token()
+
+        self.assertFalse(self.course.entitlements.exists())  # sanity check
+
+        run = CourseRunFactory(course=self.course)
+        SeatFactory(type=Seat.VERIFIED, course_run=run, price=40)
+        if has_entitlement:
+            CourseEntitlementFactory(course=self.course, price=40, mode=SeatType.objects.get(slug=Seat.VERIFIED))
+
+        response = self.client.get(reverse('api:v1:course-detail', kwargs={'key': self.course.uuid}))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn('entitlements', response.data)
+        self.assertEqual(len(response.data['entitlements']), 1)
+        self.assertTrue(self.course.entitlements.exists())
+        self.assertEqual(self.course.entitlements.first().mode.slug, Seat.VERIFIED)
+        self.assertEqual(self.course.entitlements.first().price, 40)
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -1,7 +1,6 @@
 import base64
 import logging
 import re
-from datetime import datetime
 
 from django.core.exceptions import ValidationError
 from django.core.files.base import ContentFile
@@ -30,7 +29,9 @@ from course_discovery.apps.course_metadata.constants import COURSE_ID_REGEX, COU
 from course_discovery.apps.course_metadata.models import (
     Course, CourseEditor, CourseEntitlement, CourseRun, Organization, Program, Seat, SeatType, Video
 )
-from course_discovery.apps.course_metadata.utils import ensure_draft_world, validate_course_number
+from course_discovery.apps.course_metadata.utils import (
+    create_missing_entitlement, ensure_draft_world, validate_course_number
+)
 
 logger = logging.getLogger(__name__)
 
@@ -275,7 +276,7 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         if entitlements:
             course.entitlements = entitlements
             # If entitlements were updated, we also want to update seats
-            for course_run in course.course_runs.filter(end__gt=datetime.now()):
+            for course_run in course.active_course_runs:
                 course_run.update_or_create_seats()
 
         # Save video if a new video source is provided
@@ -297,7 +298,7 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         # Then the course itself
         course = serializer.save()
         if not draft:
-            for course_run in course.course_runs.filter(end__gt=datetime.now()):
+            for course_run in course.active_course_runs:
                 if course_run.status == CourseRunStatus.Published:
                     # This will also update the course
                     course_run.update_or_create_official_version()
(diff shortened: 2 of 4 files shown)


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -976,3 +976,23 @@ def test_options(self):
...
+        SeatFactory(type=Seat.VERIFIED, course_run=run)
+        if has_entitlement:
+            CourseEntitlementFactory(course=self.course)
+
+        response = self.client.get(reverse('api:v1:course-detail', kwargs={'key': self.course.uuid}))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn('entitlements', response.data)
+        self.assertEqual(len(response.data['entitlements']), 1)
+        self.assertTrue(self.course.entitlements.exists())
```
Dillon-Dumesnil, line 1025: Nit: Maybe it's worth checking the price and mode are equal as part of this test just to be able to catch those if anything changes?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -275,7 +276,7 @@ def update_course(self, data, partial=False):
         if entitlements:
             course.entitlements = entitlements
             # If entitlements were updated, we also want to update seats
-            for course_run in course.course_runs.filter(end__gt=datetime.now()):
```
Dillon-Dumesnil, line 278: My bad. Pretty sure I wrote this line. Nice catch

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
         self.assertEqual(draft_entitlement.official_version, not_draft_entitlement)
         self.assertEqual(not_draft_entitlement.draft_version, draft_entitlement)
 
-    def test_ensure_draft_world_no_course_entitlement(self):
+    def test_ensure_draft_world_creates_course_entitlement_from_seats(self):
         """
-        If the official course has no entitlement, a draft audit entitlement should be created.
+        If the official course has no entitlement, an entitlement is created from the seat data.
```
Dillon-Dumesnil: Nit: `an entitlement is created from the seat data from active course runs.`

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+        """
+        If the official course has no entitlement and odd seats, a draft audit entitlement should be created.
         """
         course = CourseFactory()
         ensured_draft_course = utils.ensure_draft_world(course)
 
         draft_entitlement = ensured_draft_course.entitlements.first()
         self.assertEqual(draft_entitlement.price, 0.00)
-        self.assertEqual(draft_entitlement.mode.slug, 'audit')
+        self.assertEqual(draft_entitlement.mode.slug, Seat.AUDIT)
```
Dillon-Dumesnil, line 480: 👍

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+
+
+@ddt.ddt
+class TestCreateMissingEntitlement(TestCase):
+    @ddt.data(
+        # single verified seat makes entitlement
+        ([[{'type': Seat.VERIFIED, 'price': 50, 'currency': 'EUR'}]], (Seat.VERIFIED, 50, 'EUR')),
+        # single professional seat makes entitlement too
+        ([[{'type': Seat.PROFESSIONAL, 'price': 70, 'currency': 'USD'}]], (Seat.PROFESSIONAL, 70, 'USD')),
+        ([[{'type': Seat.VERIFIED}, {'type': Seat.PROFESSIONAL}]], None),  # multiple valid seats makes nothing
```
Dillon-Dumesnil, line 490: This should hopefully be an impossible case, but probably good to verify

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+        ([[{'type': Seat.AUDIT}]], None),  # no valid seats make nothing
+        ([[]], None),  # no seats at all make nothing
+        ([], None),  # no runs at all make nothing
+        # runs that disagree about valid seats make nothing
+        ([[{'type': Seat.VERIFIED}], [{'type': Seat.PROFESSIONAL}]], None),
```
Dillon-Dumesnil, line 495: Duplicate test. See line 489

Dillon-Dumesnil: Maybe this should be switched to something like 
```
[[{'type': Seat.VERIFIED, 'price': 20}], [{'type': Seat.VERIFIED, 'price': 40}]]

File: course_discovery/apps/course_metadata/tests/test_utils.py [hunk 7]
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+        # multiple agreeing runs makes entitlement
+        ([[{'type': Seat.VERIFIED, 'price': 20, 'currency': 'EUR'}],
+          [{'type': Seat.VERIFIED, 'price': 20, 'currency': 'EUR'}]], (Seat.VERIFIED, 20, 'EUR')),
+        ([[{'type': Seat.MASTERS}, {'type': Seat.CREDIT}, {'type': Seat.PROFESSIONAL, 'price': 10, 'currency': 'USD'}]],
+         (Seat.PROFESSIONAL, 10, 'USD')),  # non-relevant seats don't stop us from making an entitlement
```
Dillon-Dumesnil, line 505: `non-relevant seats`
Straight Savage

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+        seat = SeatFactory(course_run=run, type=Seat.VERIFIED, draft=True)
+
+        self.assertFalse(course.entitlements.exists())  # sanity check
+        create_missing_entitlement(course)
+
+        self.assertEqual(course.entitlements.count(), 1)
+        entitlement = course.entitlements.first()
+        self.assertEqual(entitlement.mode.slug, Seat.VERIFIED)
+        self.assertEqual(entitlement.price, seat.price)
+        self.assertEqual(entitlement.currency, seat.currency)
```
Dillon-Dumesnil, line 545: Maybe add in a check here of `self.assertTrue(entitlement.draft)` just to be safe. Definitely want to avoid having an official entitlement point at a draft course

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+        ((10,), 10),
+        ((-10,), -10),
+        ((-10, -20), -20),
+    )
+    @ddt.unpack
+    def test_active_runs(self, dates, expected):
+        """ Verifies that we only consider active runs (or last inactive) """
+        course = CourseFactory()
+        now = datetime.datetime.now(pytz.UTC)
+        usd = Currency.objects.get(code='USD')
```
Dillon-Dumesnil, line 569: Not sure this is necessary, right? Doesn't it do USD by default? Also, even if it didn't, the currency should match between seat and entitlement. This might apply for Currency everywhere in this file, but I figured you wanted to check above that using euros would be fine.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -449,13 +452,111 @@ def test_ensure_draft_world_not_draft_course_given(self):
...
+        for date in dates:
+            run = CourseRunFactory(course=course, end=now + datetime.timedelta(days=date), enrollment_end=None)
+            SeatFactory(course_run=run, type=Seat.VERIFIED, price=date, currency=usd)
+
+        self.assertFalse(course.entitlements.exists())  # sanity check
+        create_missing_entitlement(course)
+
+        entitlement = course.entitlements.first()
+        self.assertEqual(entitlement.price, expected)
```
Dillon-Dumesnil: I think this will fail once we add in checks about negative prices so I'd be cautious about that. Not sure what the best quick fix for this would be to maintain the idea of the test.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -136,6 +136,56 @@ def set_draft_state(obj, model, attrs=None):
...
+
+def _calculate_entitlement_for_course(course):
+    # Get all active runs or latest inactive run
+    runs = course.active_course_runs
+    if not runs and course.course_runs.exists():
+        runs = [course.course_runs.last()]
+    if not runs:
+        return None
+
+    entitlement_data = {_calculate_entitlement_for_run(run) for run in runs}
```
Dillon-Dumesnil, line 158: Nice use of a set to do your duplicate check for ya.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -136,6 +136,56 @@ def set_draft_state(obj, model, attrs=None):
...
+    if len(entitlement_data) > 1:
+        return None  # some runs disagree - we can't form an entitlement from this
```
Dillon-Dumesnil, line 160: Do we have a plan for what we are doing in this case?

Dillon-Dumesnil: It looks like we would end up creating a draft Audit entitlement when ensure_draft_world is run. Is that the desired behavior? We are likely disagreeing because it is verified with multiple prices, and creating a draft Audit entitlement kind of goes against this. Although audit entitlements are the only ones that can be upgraded so at least if we do that, they can correct it at a later date. Hmmm. Yeah, this might be good.

Dillon-Dumesnil (commented):
Couple of comments here and there, but I don't think anything is too serious. Just some thoughts to consider.

Also just writing it here for documentation even though we talked offline, but taking a look at ecommerce data loader to see if that has any effects on creating official entitlements for courses. I think we agreed that just pushing it to ecommerce is the right solution since when entitlements are usually created in course_metadata from publisher, we are pushing to ecommerce at the same time.

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: Ah, it's not actually a duplicate, but I understand the confusion. Line 489 is about differing seats in one run. This is about differing runs.

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: The factories pick a random currency. So I must have been specifying it so that I could compare it later. But I see now that I never compare it. I'll drop the line.

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: Yeah, depends on whether we validate at model or serializer level. I can rework to be future proof.

course_discovery/apps/course_metadata/utils.py
mikix: https://openedx.atlassian.net/browse/DISCO-1145 is the ticket to deal with courses where runs disagree. So i'm not trying to solve too much here. We'll validate whether a partner will need that support before migrating them.

My secret hope is that we won't ever need it.

But you're right that for now, we'll return a serialization without an entitlement on GETs and end up creating an Audit entitlement in create_draft_world. Which isn't the greatest. But it's still an "unsupported work flow" for new publisher.

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: Ah. It was to keep all the seats at the same currency. So when I have the same price, they will match as equal.

course_discovery/apps/course_metadata/tests/test_utils.py
Dillon-Dumesnil: Ah my bad. Might still be worth having the above test too though, right?

mikix:
OK, I've pushed an update that addresses all your test comments. Will now go back and add a call to push to ecommerce.

course_discovery/apps/course_metadata/tests/test_utils.py
Dillon-Dumesnil: Ah, good catch. Nice

course_discovery/apps/course_metadata/tests/test_utils.py
Dillon-Dumesnil: That's true. Up to you. Can also leave as an exercise to the reader.

course_discovery/apps/course_metadata/utils.py
Dillon-Dumesnil: Yeah, I think it'll be pretty rare that we will end up in that scenario. Just wanted to make sure we had some sort of plan for it.

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: Yeah, true. And one for currency.

PR #2001, opened 2019-07-09
Allow read-only access to courses

mikix:
If a user is not an editor, but is still in an authoring organization for that course, let them view the course.

Also add an 'editable' field to the serialization for a course, so a frontend will know whether to allow edits in the UI.

This means we have some overloading of `editable` - but I'm not sure how much a problem it is:
* GET with editable=0 -> can read, `editable=true/false` in serialization
* GET with editable=1 and not in org -> 403
* GET with editable=1 and in org but not editor -> read, `editable=false` in serialization
* GET with editable=1 and in org and editor -> read, `editable=true` in serialization

https://openedx.atlassian.net/browse/DISCO-1143


Dillon-Dumesnil (approved):
LGTM. I'm sure your tests will be perfect.

PR #2002, opened 2019-07-09
Only create entitlements if editable

mikix:
We only want to auto-upgrade courses to entitlements if we're using the new publisher style editable=1 API calls.


Dillon-Dumesnil (approved):
Good catch. Sorry for missing it

PR #2004, opened 2019-07-09
add org uuid to person response

rgraber:


course_discovery/apps/api/serializers.py
```diff
@@ -296,6 +297,10 @@ def get_organization_marketing_url(self, obj):
         if obj.organization:
             return obj.organization.marketing_url
 
+    def get_organization_uuid(self, obj):
+        if obj.organization:
+            return obj.organization.uuid
```
bbaker6225, line 302: This mostly looks good, you'll just want to add a `select_related` for organization on the position serializer to avoid `n+1` queries

bbaker6225: https://github.com/edx/course-discovery/blob/f33e80f903aa10c845647b6d1abffca417e4fb56/course_discovery/apps/api/serializers.py#L310-L312 gives an example of how to write a `prefetch_queryset` for the serializer, and then you'll just want to make sure you call `prefetch_queryset` wherever we're using the position serializer

rgraber: There's already a select_related(position__organization) on the MinimalPersonSerializer, which is the only place the PositionSerializer is called so I don't know if we need to add another to the PositionSerializer proper

bbaker6225: 👍

bbaker6225 (approved):

PR #2005, opened 2019-07-10
OEP-18 Take 2

mikix:
Moved to OEP-18, removed some pins, but kept most.


Dillon-Dumesnil (approved):
So many pins 😢 Looks better though. Ship it

PR #2009, opened 2019-07-10
DISCO-1081 add min_value validation for entitlements and seats on the…

xnick421:
… serializer level


Dillon-Dumesnil (approved):
Test is probably flaky. Rebase and repush and you should be all good. Nice job.

PR #2012, opened 2019-07-10
DISCO-1081 add min_value validation for entitlements and seats on the…

xnick421:
… serializer level

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -41,7 +41,7 @@ from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import Course, CourseRun, Person, Program
 from course_discovery.apps.course_metadata.tests.factories import (
-    AdditionalPromoAreaFactory, CorporateEndorsementFactory, CourseFactory, CourseRunFactory,
+    AdditionalPromoAreaFactory, CorporateEndorsementFactory, CourseEntitlementFactory, CourseFactory, CourseRunFactory,
     CurriculumCourseMembershipFactory, CurriculumFactory, CurriculumProgramMembershipFactory, DegreeCostFactory,
     DegreeDeadlineFactory, DegreeFactory, EndorsementFactory, ExpectedLearningItemFactory, IconTextPairingFactory,
     ImageFactory, JobOutlookItemFactory, OrganizationFactory, PathwayFactory, PersonAreaOfExpertiseFactory,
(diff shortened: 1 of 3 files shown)


Dillon-Dumesnil (approved):
Looks good. Check and see if there are any tests for the serializer about max_digits or decimal_places and if so, maybe throw one in for min_value.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1329,24 +1329,55 @@ def get_expected_data(cls, organization):
...
+            'bulk_sku': self.seat.bulk_sku
         }
 
         self.assertDictEqual(serializer.data, expected)
 
+    def test_price_decimal_validation(self):
+        self.seat.price = "0.000"
+        serializer = SeatSerializer(data={
+            **self.seat.__dict__,
+            **self.mock_credit_info
```
Dillon-Dumesnil: Is this necessary? It doesn't look like it was used in the `test_data` test. Seems like you should be able to just call `serializer = SeatSerializer(self.seat)` after changing what you want for your test.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1329,24 +1329,55 @@ def get_expected_data(cls, organization):
...
+            'upgrade_deadline': json_date_format(self.seat.upgrade_deadline),
+            'credit_provider': self.seat.credit_provider,
+            'credit_hours': self.seat.credit_hours,
+            'sku': self.seat.sku,
+            'bulk_sku': self.seat.bulk_sku
         }
 
         self.assertDictEqual(serializer.data, expected)
 
+    def test_price_decimal_validation(self):
```
Dillon-Dumesnil: Maybe a quick docstring on all these methods about what is not allowed. I.e. no more than 2 decimal points, < 10 digits, no negative prices (0 should probably also not be allowed if you want to test that too)

Dillon-Dumesnil: Additionally, these tests are all very similar. Maybe you can combine them to some test_validation_failures that uses ddt and you can pass in the value. Something like:

```
@ddt.data('0.000', '100000000.00', '-1')
def test_validation_errors(self, price)
    self.seat.price = price
    serializer = SeatSerializer(self.seat)
    self.assertFalse(serializer.is_valid())

File: course_discovery/apps/api/tests/test_serializers.py [hunk 3]
```diff
@@ -1329,24 +1329,55 @@ def get_expected_data(cls, organization):
...
+        self.seat.price = "0.000"
+        serializer = SeatSerializer(data={
+            **self.seat.__dict__,
+            **self.mock_credit_info
+        })
+
+        self.assertEqual(serializer.is_valid(), False)
```
Dillon-Dumesnil: Use `self.assertFalse` in all of these spots.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1329,24 +1329,55 @@ def get_expected_data(cls, organization):
...
+
+    def test_price_length_validation(self):
+        self.seat.price = "100000000.00"
+        serializer = SeatSerializer(data={
+            **self.seat.__dict__,
+            **self.mock_credit_info
+        })
+
+        self.assertEqual(serializer.is_valid(), False)
```
Dillon-Dumesnil: `self.assertFalse`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1329,24 +1329,55 @@ def get_expected_data(cls, organization):
...
+
+    def test_price_min_validation(self):
+        self.seat.price = "-1"
+        serializer = SeatSerializer(data={
+            **self.seat.__dict__,
+            **self.mock_credit_info
+        })
+
+        self.assertEqual(serializer.is_valid(), False)
```
Dillon-Dumesnil: `self.assertFalse`

Dillon-Dumesnil (commented):
Couple things to fix up in the tests, but nice job!

course_discovery/apps/api/tests/test_serializers.py
xnick421: in order to call `is_valid()` on a serializer, the `data` keyword argument is required to be passed to it.  The SeatSerializer has `credit_provider` and `credit_hour` fields it is expecting, neither of which exist on the `SeatFactory`, so `is_valid()` will always return false for these missing missing fields which is why i passed in the mock credit info.  We could add these to the`SeatFactory` but I think a better/simpler approach would be just to test for the presence of a error specific to the `price`

Dillon-Dumesnil: Hmmm, yeah. I'm fine keeping them out of SeatFactory. Sounds good.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1328,25 +1328,38 @@ def get_expected_data(cls, organization):
...
+            'upgrade_deadline': json_date_format(self.seat.upgrade_deadline),
+            'credit_provider': self.seat.credit_provider,
+            'credit_hours': self.seat.credit_hours,
+            'sku': self.seat.sku,
+            'bulk_sku': self.seat.bulk_sku
         }
 
         self.assertDictEqual(serializer.data, expected)
 
+    """Test Cases: More than two decimals, More than 8 digits, Less than 0"""
```
Dillon-Dumesnil: This should be inside the function definition.

```
def test_price_validation_errors(self, price):
    """ Test Cases: More than two decimals, More than 8 digits, Less than 0.01 """

File: course_discovery/apps/api/tests/test_serializers.py [hunk 7]
```diff
@@ -1328,25 +1328,38 @@ def get_expected_data(cls, organization):
...
+    @ddt.data("0.000", "100000000.00", "-1")
```
Dillon-Dumesnil: Switch to using single quotes to follow the convention in the file.

Dillon-Dumesnil (commented):
Couple of small comments. Copy the test format for Entitlements and should be good to go!

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1328,6 +1328,35 @@ def get_expected_data(cls, organization):
...
+            'price': str(self.entitlement.price),
+            'currency': self.entitlement.currency.code,
+            'sku': self.entitlement.sku,
+            'mode': str(self.entitlement.mode).lower(),
+            'expires': json_date_format(self.entitlement.expires)
+        }
+
+        self.assertDictEqual(serializer.data, expected)
+
+    '''Test Cases: More than two decimals, More than 8 digits, Less than 0'''
```
Dillon-Dumesnil: Move this inside the function. Also annoyingly enough, docstrings typically use double quotes so you can switch this to `""" Test Cases: More than two decimals, More than 8 digits, Less than 0 """`

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1351,14 +1380,14 @@ def test_data(self):
 
         self.assertDictEqual(serializer.data, expected)
 
-    """Test Cases: More than two decimals, More than 8 digits, Less than 0"""
-    @ddt.data("0.000", "100000000.00", "-1")
+    '''Test Cases: More than two decimals, More than 8 digits, Less than 0'''
```
Dillon-Dumesnil: Move this inside of the function. Also annoyingly enough, docstrings typically use double quotes so you can switch this to """ Test Cases: More than two decimals, More than 8 digits, Less than 0 """

Dillon-Dumesnil (approved):
Move the doc strings and I'd say you're all good! Don't forget to squash your commits down to one commit before merging to master. Also rebasing on master. It won't let you merge unless your branch is up to date with master.

mikix:
I think there's an argument that this would have fit better on the model level (we don't want negative prices no matter how they come in) - maybe even a new PriceField class that hides the complexity of our price fields (they both have the same arguments but just do `**PRICE_FIELD_CONFIG`, which is needlessly repetitive) and add on a new `validators` init argument. (See our HtmlField as an example).

That said, this work is already done on the serializer level, so good enough.

PR #2013, opened 2019-07-10
Management command fetch and store fixture data

zacharis278:
### [EDUCATOR-4456](https://openedx.atlassian.net/browse/EDUCATOR-4456)


course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,80 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Command to populate catalog database with programs from another environment
+    using the /program-fixtures endpoint
+
+    Usage:
+        ./manage.py load_program_fixtures 707acbed-0dae-4e69-a629-1fa20b87ccf1
```
kdmccormick: `load_program_fixture` (no s)

A single fixture refers to the whole shebang of programs, courses, etc. that you can load in.

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,153 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Command to populate catalog database with programs from another environment
+    using the /program-fixtures endpoint
+
+    Usage:
+        ./manage.py load_program_fixture 707acbed-0dae-4e69-a629-1fa20b87ccf1
```
kdmccormick: maybe also include a uuid:external_key pairing in the example, to showcase the full syntax?

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,153 @@
...
+                query=','.join(programs)
+            )
+        )
+        response = client.request('GET', url)
+        return response.text
+
+    def save_fixture_object(self, obj):
+        try:
+            obj.save()
+            logger.info('Saved {app_label}.{object_name}(pk={pk})'.format(
```
kdmccormick: the docs recommend `_meta.label` (or `.label_lower`) as a shortcut to `{app_label}.{object_name}`

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,153 @@
...
+                stored_seat_type, created = SeatType.objects.get_or_create(name=item.object.name)  # pylint: disable=unused-variable
+                seat_type_map[item.object.id] = (stored_seat_type, item)
+            elif isinstance(item.object, ProgramType):
+                stored_program_type, created = ProgramType.objects.get_or_create(name=item.object.name)
+                program_type_map[item.object.id] = (stored_program_type, item)
+            else:
+                # partner models are not included in the fixture
+                # replace partner with valid reference in this environment
+                try:
+                    item.object._meta.get_field('partner')
```
kdmccormick, line 123: What does this line do?

course_discovery/apps/course_metadata/management/commands/tests/test_load_program_fixture.py
```diff
@@ -0,0 +1,327 @@
...
+        call_command(
+            'load_program_fixture',
+            ','.join(program_uuids),
+            '--catalog-host', self.catalog_host,
+            '--oauth-host', self.oauth_host,
+            '--client-id', 'foo',
+            '--client-secret', 'bar',
+        )
+
+    def reset_db_state(self):
```
kdmccormick, line 96: I believe the DB will be reset after every test run automatically, I think this is superfluous.

kdmccormick (commented):
Looking good. Just a couple comments.

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+        url = urljoin(
+            catalog_host,
+            'extensions/api/v1/program-fixture/?programs={query}'.format(
+                query=','.join(programs)
+            )
+        )
+        response = client.request('GET', url)
+        return response.text
```
kdmccormick, line 62: If any of the UUIDs are invalid and prod gives back a 404, we fail ungracefully when we try to deserialize the "Not found" or whatever comes back

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
kdmccormick: I recommend instead/also printing the UUID (making it easier to connect input data with log output)

kdmccormick (commented):
A couple additional comments

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
zacharis278: this checks if the object we are going to create has a partner field and if so sets the partner value to edX

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
zacharis278: ah cool

course_discovery/apps/course_metadata/management/commands/tests/test_load_program_fixture.py
zacharis278: I'm running this in the middle of tests to clear the DB before calling the management command.  It's a bit awkward since generating the test data populates the DB which we would like to be empty when the command is called.

kdmccormick: Ahhhhh right, that is a strange situation. But I don't know any other way to solve it, so this looks good.

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
zacharis278: on second thought I left this one out.  Not everything we're saving has a uuid and it just gets messy checking/adding in a different print statement based on the field.

kdmccormick: Right, makes sense.

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,159 @@
+"""
+Populate catalog programs for masters sandbox environment
+"""
+import logging
+from posixpath import join as urljoin
+
+from django.core import serializers
+from django.core.exceptions import FieldDoesNotExist
+from django.core.management import BaseCommand, CommandError
+from django.db import DEFAULT_DB_ALIAS, DatabaseError, IntegrityError, connections, transaction
```
schenedx: #NIT: Since you need to use 5 objects from the `django.db` library, might as well just do `from django import db` and then use `db.<object>` to make it easier to read

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,159 @@
...
+                query=','.join(programs)
+            )
+        )
+
+        response = client.request('GET', url)
+
+        if response.status_code != 200:
+            raise CommandError('Unexpected response loading programs from discovery service: {code} {msg}'.format(
+                code=response.status_code,
+                msg=response.text,
```
schenedx, line 59: Are you sure the `response.text` property will do the right thing in case of 400 or 500?

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -0,0 +1,159 @@
...
+        partner = Partner.objects.get(short_code=options['partner_code'])
+
+        connection = connections[DEFAULT_DB_ALIAS]
+
+        with connection.constraint_checks_disabled():
+            self.load_fixture(fixture, partner)
+        try:
+            connection.check_constraints()
+        except Exception as e:
+            e.args = ("Problem checking constraints. Unable to save program(s): %s" % e,)
```
schenedx: Can you please make this error message clearer? Like `Checking database constraint failed trying to load fixtures. Unable to save program...`?

course_discovery/apps/course_metadata/management/commands/tests/test_load_program_fixture.py
```diff
@@ -0,0 +1,327 @@
...
+        Partner.objects.all().exclude(short_code='edx').delete()
```
schenedx, line 97: #NIT Aren't there other good ways to reset db? See https://github.com/django/django/blob/eac9ab7ebb1ce0cbbc79c4cf65e8f70b0635a240/django/core/management/commands/flush.py#L10? Maybe just invoke this management command?
Just thinking this list might change in the future.

schenedx (approved):
All my comments are NIT
I tested this command on the utaustinx sandbox and it works!

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
zacharis278: yeah that would be cleaner

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
zacharis278: yeah it just gives the raw body text.  E.g. the wrong uuid shows "Unexpected response loading programs from discovery service: 404 Could not load programs from UUIDs: [707acbed-0dae-4e69-a629-1fa20b87ccf2]"

PR #2016, opened 2019-07-11
Publish data from Publisher to Course Metadata

Dillon-Dumesnil:
In order to ensure we have data consistency, we are changing the migrate
course editors management command to include pushing course and course
run data from Publisher to Course Metadata. We will write all publisher
information into draft models as applicable.

DISCO-1144


course_discovery/apps/course_metadata/management/commands/migrate_publisher_to_course_metadata.py
```diff
@@ -41,7 +45,13 @@ def handle(self, *args, **options):
                 )
                 exception_org_keys.append(org_key)
                 continue
-            for discovery_course in Course.objects.filter(authoring_organizations__in=[org]):
+
+            for publisher_course in PublisherCourse.objects.filter(organizations__in=[org]):
+                for course_run in publisher_course.course_runs.all():
+                    partner = org.partner
```
Dillon-Dumesnil, line 51: Is this fine to do? I wasn't sure what the best way to get the partner would be since it's not coming from a request and that's how we usually do it.

Dillon-Dumesnil:
One thing to note is that any Courses in Publisher without any Course Runs won't be copied over because the publish to course metadata function requires a course run.

course_discovery/apps/course_metadata/management/commands/migrate_publisher_to_course_metadata.py
mikix: That's fine. Course objects also have them.

mikix (approved):
Seems fine? I'm very curious to see if we hit any issues with Curtin and migrating more than they expect - what fallout happens from that. But we'll see!

I did mention in person the issue about if a draft version already exists. But assuming that's fixed, let's try it.

course_discovery/apps/course_metadata/management/commands/migrate_publisher_to_course_metadata.py
Dillon-Dumesnil: Yeah, we just don't have access to the Course Metadata Course object at this point so figured I'd use the org

course_discovery/apps/course_metadata/management/commands/migrate_publisher_to_course_metadata.py
```diff
@@ -41,7 +45,13 @@ def handle(self, *args, **options):
...
+                    publish_to_course_metadata(partner, course_run, draft=True)
+
+            for discovery_course in Course.everything.filter(authoring_organizations__in=[org], draft=True):
```
jlajoie, line 54: Almost got very confused that a discovery_course is a course_metadata_course

course_discovery/apps/course_metadata/migrations/0185_migrate_publisher_to_course_metadata_config.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.21 on 2019-07-10 16:11
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
jlajoie, line 8: Not something I want you to fix, but I'm starting to feel like if this is the patter we're doing for migration configs to move them outside of the course_metadata tables, just feels a little bloated/unrelated but I understand why we've done it this way

course_discovery/apps/course_metadata/utils.py
```diff
@@ -462,6 +461,117 @@ def push_to_ecommerce_for_course_run(course_run):
     return True
 
 
+def publish_to_course_metadata(partner, course_run, draft=False):
```
jlajoie, line 464: nit: might be worth specifying this as publisher_course_run?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -462,6 +461,117 @@ def push_to_ecommerce_for_course_run(course_run):
...
+        if discovery_course:
+            ensure_draft_world(discovery_course)
+
+    discovery_course, created = Course.everything.update_or_create(
+        partner=partner, key=course_key, draft=draft, defaults=defaults
+    )
+    discovery_course.image.save(publisher_course.image.name, publisher_course.image.file)
+    discovery_course.authoring_organizations.add(*publisher_course.organizations.all())
+
+    subjects = [subject for subject in [
```
jlajoie, line 503: Wow I don't like that we have it set up this way but I know this isn't on you

course_discovery/apps/course_metadata/utils.py
```diff
@@ -462,6 +461,117 @@ def push_to_ecommerce_for_course_run(course_run):
...
+        'min_effort': course_run.min_effort,
+        'max_effort': course_run.max_effort,
+        'language': course_run.language,
+        'weeks_to_complete': course_run.length,
+        'has_ofac_restrictions': course_run.has_ofac_restrictions,
+        'external_key': course_run.external_key,
+        'expected_program_name': program_name or '',
+        'expected_program_type': expected_program_type,
+    }
+    discovery_course_run, __ = CourseRun.everything.update_or_create(
```
jlajoie, line 528: This works here because we've verified that the discovery_course is a draft, so it'll only ever be pointing to draft runs correct?

jlajoie (approved):
Logic seems sound and encompasses everything we need. Looks like it'll only ever hit drafts too. Just minor nits/overall complaints about how we've structured things in the past but I don't think those should be blockers. 👍

course_discovery/apps/course_metadata/utils.py
Dillon-Dumesnil: The discovery course might be a draft, it might not be. This function is called by the old left-publish code still and then by the management command. In the management command case, the discovery_course will be a draft. For the left-publish case, it won't be.

jlajoie: Makes sense 👍

PR #2020, opened 2019-07-12
Use new-style django MIDDLEWARE

mikix:
Update middleware dependencies to versions that can handle new-style MIDDLEWARE setting. And then switch to that style.


Dillon-Dumesnil (approved):
LGTM

PR #2021, opened 2019-07-12
Drop CourseMarketingSiteDataLoader

mikix:
We no longer import any interesting courses from the marketing site because they all have UUIDs listed on the marketing side. So lets drop this dead code and simplify the loader logic.

Also drops the load_drupal_data command.

https://openedx.atlassian.net/browse/DISCO-716

Follow-on work for edx.org:
- Drop the no-longer-used `migrate_drupal_tags` waffle switch from stage and prod.
- Delete the load_drupal_data jenkins job.


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -248,344 +237,3 @@ def process_node(self, data):
...
-        else:
-            logger.info(
-                'Course_run [%s] has uuid [%s] already on course about page. No need to ingest',
-                data['field_course_id'],
-                data['field_course_uuid']
-            )
-
-        # Once we've ingested the course and course run, we update the topic tags with any marketing site tags found.
-        if waffle.switch_is_active('migrate_drupal_tags') and course_run:
-            self.set_topic_tags(course_run.course, data)
```
mikix, line 367: This flag is no longer needed - confirmed with Diana.

pytest-no-xdist.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
 testpaths = course_discovery/apps
-addopts = --cov=course_discovery --cov-report term --cov-config=.coveragerc --ds=course_discovery.settings.test
+addopts = --cov=course_discovery --cov-report term --cov-config=.coveragerc --no-cov-on-fail --ds=course_discovery.settings.test
```
mikix, line 3: Unrelated, but a good change, I think.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -248,344 +237,3 @@ def process_node(self, data):
...
-        # NOTE (CCB): We need to include the nested taxonomy_term data since that is where the
-        # language information is stored.
-        kwargs['load-entity-refs'] = 'file,taxonomy_term'
-        return kwargs
-
-    def process_node(self, data):
-
-        course_run = self.get_course_run(data)
-
-        if not data.get('field_course_uuid'):
```
mikix, line 322: This is the big line - theoretically, every course run we would want to import already has a UUID. So we are never getting past this `if` now for active courses.

Jeff - we are still importing courses without UUIDs right now - but we are confident that we no longer care about those, right? (They are all archived?)

jlajoie: :+1

jlajoie (approved):
Confident in this :D thanks for putting it down! I suppose if we ever need to bring it back for a one-off we do a revert and can re-run, which isn't a small ask but not impossible.

PR #2031, opened 2019-07-16
DISCO-1144: Hopefully final fixes to migrate command

Dillon-Dumesnil:


mikix (approved):
I don't think you need any of the po/mo file changes?

PR #2032, opened 2019-07-16
Upgrade DRF

mikix:


course_discovery/settings/production.py
```diff
@@ -30,7 +30,7 @@
     CONFIG_FILE = environ['COURSE_DISCOVERY_CFG']
 
 with open(CONFIG_FILE, encoding='utf-8') as f:
-    config_from_yaml = yaml.load(f)
+    config_from_yaml = yaml.safe_load(f)
```
mikix: Deprecated in newer PyYAMLs (even though this PR actually goes backward on pyyaml version).

course_discovery/apps/api/mixins.py
```diff
@@ -3,7 +3,7 @@
 """
 # pylint: disable=not-callable
 
-from rest_framework.decorators import list_route
+from rest_framework.decorators import action
```
mikix, line 6: DRF drops list_route and detail_route in 3.10 (even though we aren't upgrading to it yet, just getting ready).

course_discovery/apps/api/renderers.py
```diff
@@ -62,4 +62,5 @@ class CourseRunCSVRenderer(CSVStreamingRenderer):
         'seats.credit.credit_provider',
         'seats.credit.credit_hours',
         'modified',
+        'course_key',
```
mikix: Something about drf-csv changed where this had to be specified (as best I can tell, it was in old CSV and this makes it still be there...) I don't fully understand this, and it needs testing in stage to be sure.

course_discovery/apps/api/v1/urls.py
```diff
@@ -19,7 +19,7 @@
 from course_discovery.apps.api.v1.views.user_management import UsernameReplacementView
 
 partners_router = routers.SimpleRouter()
-partners_router.register(r'affiliate_window/catalogs', AffiliateWindowViewSet, base_name='affiliate_window')
+partners_router.register(r'affiliate_window/catalogs', AffiliateWindowViewSet, basename='affiliate_window')
```
mikix, line 22: Old name is being dropped in DRF 3.11, this is just getting ready.

course_discovery/apps/api/renderers.py
```diff
@@ -10,56 +9,3 @@ class AffiliateWindowXMLRenderer(XMLRenderer):
...
-        'seats.verified.currency',
-        'seats.verified.upgrade_deadline',
-        'seats.credit.type',
-        'seats.credit.price',
-        'seats.credit.currency',
-        'seats.credit.upgrade_deadline',
-        'seats.credit.credit_provider',
-        'seats.credit.credit_hours',
-        'modified',
-    ]
```
mikix: OK, here's the deal. In 2016, we added the CSV endpoint with this nice field ordering above. In 2018, we broke that ordering. (Apparently the combo of the streaming CSV class and the version we were using didn't pay attention to our requested ordering - but newer versions fixed that bug).

I don't know how many or if anyone complained at the time. But I'm loathe to break the API a second time now that the upstream bug is fixed, just to have a more aesthetically pleasing ordering. So, I'm dropping this old nice ordering and just freezing our busted order in place.

course_discovery/apps/api/renderers.py
```diff
@@ -14,52 +14,69 @@ class AffiliateWindowXMLRenderer(XMLRenderer):
 
 class CourseRunCSVRenderer(CSVStreamingRenderer):
     """ CSV renderer for course runs. """
+    # This ordering is mostly alphabetical, for historical reasons. In 2016, we added this CSV endpoint with a nice
```
jlajoie, line 17: I _love_ this comment for context, thanks for adding this!

jlajoie (approved):
Overall looks good, great job on catching those tests and doing this upgrade! Just for sanity, we've verified the responses are the same between version upgrades? (I assume nothing would change, just figured might as well call it out)

mikix:
Hmmm... Looks like the default CSV code was also pulling in staff info into the CSV. And that makes the CSV in stage/prod look real messy (lists of serialized staff members does not flatten nicely).

I'm going to keep leaving that out of my new changes. It's too gross and non-relevant to keep in the CSV. So I guess I am breaking format after all a little bit... But I'll leave the alphabetical in place, to try to minimize the disruption I guess.

PR #2034, opened 2019-07-17
Adding course into create CourseRun params

Dillon-Dumesnil:


mikix (approved):
QQ

You also changed a second update_or_create, yah?

Dillon-Dumesnil:
Yah, on the entitlement with mode. I don't really have the same concerns there as I do with this change.

PR #2035, opened 2019-07-17
Ndalfonso/disco 645 course editor endpoint

xnick421:


course_discovery/apps/api/serializers.py
```diff
@@ -958,6 +960,30 @@ def create(self, validated_data):
...
+        model = CourseEditor
+        fields = (
+            'id',
+            'user',
+            'user_id',
+            'course',
+        )
+
+    def create(self, validated_data):
+        course = Course.objects.get(uuid=validated_data['course'])
```
mikix: Ah, always be very careful with `.get` that you are passing in enough flags to uniquely identify a row.

`Course` has a `unique_together` like:
```
        unique_together = (
            ('partner', 'uuid', 'draft'),
            ('partner', 'key', 'draft'),
        )
```
(We allow different partners - really different sites - to have the same courses in their catalog.)

Now, `Course.objects` only returns `draft=False` rows. So you're safe there. But you also should be passing in a partner. Which I believe you can get from `self.context['request'].site.partner`.

Except you actually do want draft versions. CourseEditors are always associated with a draft row, not an official row. So I'd probably also do `Course.everything.get(draft=True, ...`

And then consider - what if the course doesn't have a draft version yet? We could try to solve that case, but in practice I don't think that will happen. Easier to explain why in person, maybe poke Dillon about it or we can chat when I'm in.

course_discovery/apps/api/serializers.py
```diff
@@ -958,6 +960,30 @@ def create(self, validated_data):
...
+        course_editor = CourseEditor(
+            user=validated_data['user_id'],
+            course=course
+        )
+        course_editor.save()
```
mikix: You could drop this line by above making the object with `CourseEditor.objects.create` which automatically calls `save()`.

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+from course_discovery.apps.course_metadata.models import Course
+from course_discovery.apps.course_metadata.models import CourseEditor
+from course_discovery.apps.course_metadata.models import Organization
+
+
+class CourseEditorViewSet(mixins.CreateModelMixin,
+                          mixins.RetrieveModelMixin,
+                          mixins.ListModelMixin,
+                          mixins.DestroyModelMixin,
+                          viewsets.GenericViewSet):
```
mikix, line 18: You're listing these out rather than using `ModelViewSet` to avoid the `UpdateModelMixin` and having an `update()` method?

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_queryset(self):
+        uuid = self.request.query_params.get('uuid')
+        user = self.request.user
+
+        if user.is_staff:
+            query = CourseEditor.objects.all()
+        else:
+            user_orgs = Organization.objects.filter(organization_extension__group__in=user.groups.all())
+            query = CourseEditor.objects.filter(user__groups__organization_extension__organization__in=user_orgs)
```
mikix: I've so far been trying to keep all the gross logic around groups and org extensions and course editors inside class methods on the `CourseEditor` class. Simply so that it's all in one place if we ever change anything, because the queries are so weird.

So I'd vote for making most of this method a utility class method on `CourseEditor`. Something like `editors_for_user` or something.

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+
+        if uuid:
+            query = query.filter(course__uuid=uuid)
+
+        return query
+
+    def create(self, request):
+        course = get_object_or_404(Course, uuid=request.data['course'])
+        if not CourseEditor.is_course_editable(request.user, course):
+            raise PermissionDenied
```
mikix: There is view support for permission classes. See `IsCourseEditorOrReadOnly` for example - though you'd want different / simpler logic maybe. But I dunno, maybe this class is simple enough you don't need to be so fancy. Just mentioning it as an option if this class grows.

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+        ).distinct()
+
+        if editor not in users_in_authoring_orgs:
+            raise PermissionDenied('Editor does not belong to an authoring organization of this course.')
+
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
```
mikix: You could replace this whole stanza with `return super().create(request, *args, **kwargs)`

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def perform_create(self, serializer):
+        serializer.save()
```
mikix: You don't need this - that's just the same as the one provided by `CreateModelMixin`

mikix (commented):
I have not sat down and looked at what the serialization in and out would look like -  will do that next. But some comments below.

course_discovery/apps/api/v1/views/course_editors.py
mikix: I dunno, maybe it's fine here too. Since you also do a org check below in `create`, which probably should live in this class... Eh, your choice. I don't feel strongly about this.

course_discovery/apps/api/serializers.py
```diff
@@ -958,6 +960,30 @@ def create(self, validated_data):
         return Course.objects.create(**validated_data)
 
 
+class CourseEditorSerializer(serializers.ModelSerializer):
+    user = GroupUserSerializer(read_only=True)
+    user_id = serializers.PrimaryKeyRelatedField(queryset=User.objects.all(), write_only=True)
+    course = serializers.UUIDField(format='hex_verbose')
```
mikix: This bit isn't working - I get `str(course)` in my serialization. You might want `serializers.SlugRelatedField(slug_field='uuid')`?

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+from course_discovery.apps.course_metadata.models import CourseEditor
+from course_discovery.apps.course_metadata.models import Organization
+
+
+class CourseEditorViewSet(mixins.CreateModelMixin,
+                          mixins.RetrieveModelMixin,
+                          mixins.ListModelMixin,
+                          mixins.DestroyModelMixin,
+                          viewsets.GenericViewSet):
+    serializer_class = CourseEditorSerializer
```
mikix, line 21: You might also want to specify a `pagination_class` - not likely to have large sets of editors, but just for completeness. EdX's current preferred paginator is `CursorPagination` (even though it's not used in disco yet).

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+        if user.is_staff:
+            query = CourseEditor.objects.all()
+        else:
+            user_orgs = Organization.objects.filter(organization_extension__group__in=user.groups.all())
+            query = CourseEditor.objects.filter(user__groups__organization_extension__organization__in=user_orgs)
+
+        if uuid:
+            query = query.filter(course__uuid=uuid)
+
+        return query
```
mikix: In your tests, if you can have a test that checks query counts (ideally we're at 1, maybe 2?). If using `assertNumQueries`, pass in a threshold of 0, since there shouldn't be wiggle room on this - if we're doing more queries, we need to add some `select_related` or something.

course_discovery/apps/api/v1/views/course_editors.py
mikix: (It's the preferred paginator for performance reasons - the other ones have stuff like page_size and page_num - but to find where page_num starts in the database, is expensive. The CursorPagination class uses a start_at_index=xxx param (or something like that) and is much faster to find in the database.

Again, not concerns here. I'm more just curious if we can start using it and getting it in the code for future copying and pasting. :)

course_discovery/apps/api/v1/views/course_editors.py
```diff
@@ -0,0 +1,59 @@
...
+
+class CourseEditorViewSet(mixins.CreateModelMixin,
+                          mixins.RetrieveModelMixin,
+                          mixins.ListModelMixin,
+                          mixins.DestroyModelMixin,
+                          viewsets.GenericViewSet):
+    serializer_class = CourseEditorSerializer
+
+    def get_queryset(self):
+        uuid = self.request.query_params.get('uuid')
```
mikix: Oh also - you can add a Filter class that offers various ways of slicing down the queryset - look at like, ProgramFilter for example.

And I might name this `course` instead of `uuid` - otherwise it's not clear what the uuid is even referring to.

course_discovery/apps/api/filters.py
```diff
@@ -204,3 +204,14 @@ def _set_language(self, queryset, _, language_code):
     class Meta:
         model = Topic
         fields = ('slug', 'language_code')
+
+
+class CourseEditorFilter(filters.FilterSet):
+    course = filters.CharFilter(method='set_course_id')
```
mikix: You should be able to just do `filters.CharFilter(name='course__uuid')`

course_discovery/apps/api/permissions.py
```diff
@@ -9,11 +9,11 @@
 
 class ReadOnlyByPublisherUser(BasePermission):
     """
-    Custom Permission class to check user is a publisher user.
+    Custom Permission class to check user is a publisher user or a staff user.
     """
     def has_permission(self, request, view):
         if request.method == 'GET':
-            return request.user.groups.exists()
+            return request.user.is_staff or request.user.groups.exists()
```
mikix: This looks like a weird merge thing? You didn't make this change, right?

course_discovery/apps/api/v1/urls.py
```diff
@@ -32,19 +33,20 @@
...
+router.register(r'subjects', SubjectViewSet, base_name='subject')
+router.register(r'topics', TopicViewSet, base_name='topic')
+router.register(r'pathways', PathwayViewSet, base_name='pathway')
+router.register(r'programs', ProgramViewSet, base_name='program')
+router.register(r'program_types', ProgramTypeViewSet, base_name='program_type')
+router.register(r'search/all', search_views.AggregateSearchViewSet, base_name='search-all')
+router.register(r'search/courses', search_views.CourseSearchViewSet, base_name='search-courses')
+router.register(r'search/course_runs', search_views.CourseRunSearchViewSet, base_name='search-course_runs')
+router.register(r'search/programs', search_views.ProgramSearchViewSet, base_name='search-programs')
+router.register(r'search/people', search_views.PersonSearchViewSet, base_name='search-people')
```
mikix: Looks like a bad merge? We want `basename` not `base_name`

course_discovery/apps/publisher/api/serializers.py
```diff
@@ -77,6 +77,13 @@ def get_user_full_name(self, obj):
         return obj.get_full_name() or obj.username
 
 
+class ExtendedGroupUserSerializer(GroupUserSerializer):
+    email = serializers.CharField()
```
mikix: Honestly, I'd say just add email to the base GroupUserSerializer. I don't think email is any more priviledged than the full name.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -73,8 +73,8 @@ def __getattr__(self, name):
     def enabled(self):
         return self.salesforce_is_enabled()
 
-    def _query(self, soql):
-        return self.client.query(self.soql_escape(soql))
+    def _query(self, soql, *soql_args):
+        return self.client.query(soql.format(*[self.soql_escape(arg) for arg in soql_args]))
```
mikix: Again, bad merge?

course_discovery/apps/api/permissions.py
```diff
@@ -75,3 +75,18 @@ class CanReplaceUsername(BasePermission):
...
+        if request.method in SAFE_METHODS:
+            return True
+        else:
+            course = request.data.get('course')
+            if not course:
+                # Fail happily because OPTIONS goes down this path too with a fake POST.
+                # If this is a real POST, we'll complain about the missing course in the view.
+                return True
+
+            return CourseEditor.is_course_editable(request.user, view.course)
```
mikix, line 94: Feels odd to ignore the course value you just pulled from the request data and use view.course instead. Especially since that value is no longer cached. Maybe it would make more sense - and keep the view and permission classes more independent - if you did a lookup here.

(If you don't like that idea, maybe just add a comment about why the two courses.)

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+        self.org_ext = OrganizationExtensionFactory()
+
+    def test_list(self):
+        """Verify GET endpoint returns list of editors"""
+        CourseEditorFactory()
+        response = self.client.get(self.list_path)
+
+        assert len(response.data['results']) == 1
+
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
```
mikix, line 34: Maybe just a quick comment above like "now test with an unprivileged user" - or use ddt to make this dual test more obvious.

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+        CourseEditorFactory(course=self.course)
+        CourseEditorFactory()
+
+        response = self.client.get(self.list_path)
+
+        assert len(response.data['results']) == 2
+
+        response = self.client.get(self.list_path, {'course': self.course.uuid})
+
+        assert len(response.data['results']) == 1
```
mikix, line 50: Might want to check that the returned single editor is indeed the one you made above for self.course. Things would be very weird if not, but good to cover bases.

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+        self.staff_user.groups.add(self.org_ext.group)
+        course.authoring_organizations.add(self.org_ext.organization)
+
+        self.client.post(self.list_path, {'course': course.uuid}, format='json')
+        course_editor = CourseEditor.objects.first()
+
+        assert course_editor.course == course
+        assert course_editor.user == self.staff_user
+
+    def test_create_for_self_as_non_staff(self):
```
mikix: This and the test above it are good candidates for using ddt instead of repeating test code.

I'm guessing you haven't used ddt before? I might do something like:
```
@ddt.data(True, False)
def test_create_for_self(self, is_staff):
    self.user.staff = is_staff
    self.user.save()

   ... rest of test ...
```

Look around in code for more examples.

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.client.post(self.list_path, {'course': course.uuid}, format='json')
+        course_editor = CourseEditor.objects.first()
+
+        assert course_editor.course == course
+        assert course_editor.user == self.user
+
+    def test_create_for_self_as_non_staff_with_invalid_course(self):
+        """Verify non staff user cannot make them self an editor of a course they dont belong to"""
+        self.assertFalse(CourseEditor.objects.all().exists())
```
mikix: This is not a crazy sanity check, but the django test runner makes sure that the DB is wiped between tests - you probably don't need this at the beginning of your tests.

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+
+        response = self.client.post(self.list_path, {'course': self.course.uuid}, format='json')\
```
mikix: Weird little trailing backslash here.

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+
+        assert response.status_code == 403
+
+    def test_create_for_other_user_as_staff(self):
+        """Verify staff user can make another user an editor"""
+        self.assertFalse(CourseEditor.objects.all().exists())
+
+        self.user.groups.add(self.org_ext.group)
```
mikix, line 83: Does staff need to be in the group?

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -0,0 +1,128 @@
...
+        self.course.authoring_organizations.add(self.org_ext.organization)  # pylint: disable=no-member
```
mikix: Having the course be authored correctly feels like something that could be done in the setUp()

mikix (approved):
Got some nit comments about tests, but this looks good! Merge when you feel happy with it.

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
xnick421: thats for non staff :)

mikix: But this is a staff test.

xnick421: ok ya, self.user is the user they are making an editor.  the person signed in, in this test is self.staff_user.  we need to do this line because a non-staff user needs to be in the org in order to be an editor

PR #2037, opened 2019-07-18
Make it so staff users can read pathways

kdmccormick:
@edx/masters-neem 
https://openedx.atlassian.net/browse/EDUCATOR-4473

Fixes issue where `./manage.py lms cache_programs` was failing on sandboxes with a 403 from Discovery.

For context, the `pathways` endpoint (which was causing the 403) requires the `ReadOnlyByPublisherUser` permission, which very stringently requires that you _are part of a group_.

No other view in Discovery uses `ReadOnlyByPublisherUser`, and being "staff" is kinda like being in a group, so I figured that this fix would be reasonable.


iloveagent57 (approved):
Seems fine to me.

PR #2039, opened 2019-07-19
remove marketing_course_runs from courses

rgraber:
Note: this change is technically not backwards compatible, but it shouldn't matter because this field was only ever used by prospectus, which has been updated to no longer use it


bbaker6225 (approved):

bbaker6225:
Might be useful to add a note in the description that this isn't technically backwards compatible, but should be fine since it's only used by prospectus

PR #2040, opened 2019-07-19
Exempt certain fields from changing review status

jmyatt:
DISCO-1185

### DIFF
--- a/course_discovery/apps/api/utils.py
+++ b/course_discovery/apps/api/utils.py
@@ -51,19 +51,21 @@ def get_query_param(request, name):
     return cast2int(request.query_params.get(name), name)
 
 
-def data_has_changed(obj, new_key_vals):
+def reviewable_data_has_changed(obj, new_key_vals, exempt_fields=None):
     """
     Check whether serialized data for the object has changed.
 
     Args:
         obj (Object): Object representing the persisted state
         new_key_vals (dict_items): List of (key,value) tuples representing the new state
+        exempt_fields (list): List of field names where a change does not affect review status
 
     Returns:
-        bool for whether data for any fields has changed
+        bool for whether data for any reviewable fields has changed
     """
     changed = False
-    for key, new_value in new_key_vals:
+    exempt_fields = exempt_fields or []
+    for key, new_value in [x for x in new_key_vals if x[0] not in exempt_fields]:
         original_value = getattr(obj, key, None)
         if isinstance(new_value, list):
             field_class = obj.__class__._meta.get_field(key).__class__
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -568,26 +568,31 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         assert draft_course_run.official_version.status == CourseRunStatus.Unpublished
 
     @responses.activate
-    def test_patch_put_reset_status_transcript_languages(self):
+    def test_patch_put_non_review_fields_does_not_reset_status(self):
         self.mock_patch_to_studio(self.draft_course_run.key)
         self.mock_ecommerce_publication()
         self.draft_course_run.status = CourseRunStatus.Reviewed
+        self.draft_course_run.go_live_date = datetime.datetime(2031, 1, 1, tzinfo=pytz.UTC)
         self.draft_course_run.save()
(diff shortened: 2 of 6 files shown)


course_discovery/apps/api/utils.py
```diff
@@ -51,19 +51,21 @@ def get_query_param(request, name):
...
         new_key_vals (dict_items): List of (key,value) tuples representing the new state
+        exempt_fields (list): List of field names where a change does not affect review status
 
     Returns:
-        bool for whether data for any fields has changed
+        bool for whether data for any reviewable fields has changed
     """
     changed = False
-    for key, new_value in new_key_vals:
+    exempt_fields = exempt_fields or []
```
jmyatt, line 67: placing the default here instead of in the function signature to avoid pylint `dangerous-default-value`

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -568,26 +568,31 @@ def test_patch_put_reset_status(self):
         assert draft_course_run.official_version.status == CourseRunStatus.Unpublished
 
     @responses.activate
-    def test_patch_put_reset_status_transcript_languages(self):
```
jmyatt, line 571: confirmed with Product that this field should be exempt from needing changes to be reviewed, so I'm removing this test

course_discovery/apps/api/utils.py
Dillon-Dumesnil: I believe this is the proper way of using a default value of an empty list 👍

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
Dillon-Dumesnil: Do we still have a test that does reset status? Just want to make sure we didn't get rid of our one test that was testing status change post review.

course_discovery/apps/course_metadata/models.py
```diff
@@ -814,6 +814,18 @@ def editable_course_runs(cls, user, queryset):
 
 class CourseRun(DraftModelMixin, TimeStampedModel):
     """ CourseRun model. """
+    STATUS_CHANGE_EXEMPT_FIELDS = [
```
Dillon-Dumesnil: Nit: move these down to the bottom of the defined fields. I'm not sure where they are _supposed_ to be defined, but I feel like it is nicer to have all of the actual fields up top and this random definition lower down.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -300,7 +300,7 @@ def update_course(self, data, partial=False):
             image_data = ContentFile(base64.b64decode(imgstr), name='tmp.{extension}'.format(extension=ext))
             course.image.save(image_data.name, image_data)
 
-        changed = data_has_changed(course, serializer.validated_data.items())
+        changed = reviewable_data_has_changed(course, serializer.validated_data.items())
```
Dillon-Dumesnil, line 303: You have a test in test_courses.py about how image and video don't change the status, but those aren't listed here as exempted fields. How does that work?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
Dillon-Dumesnil: Same question on the test_courses.py file.

Dillon-Dumesnil (approved):
Couple of comments to respond to, but as long as the responses are all addressed already and I just missed them, Good to go.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
jmyatt: yes we still have `test_patch_put_reset_status` (course_runs) and `test_patch_resets_run_status` (courses) for this purpose

course_discovery/apps/api/v1/views/courses.py
jmyatt: those are fields that we handle [directly in the API view as opposed to via the serializer](https://github.com/edx/course-discovery/blob/4ea39e21f86707e6585c491cbe26359ecd9a0ee9/course_discovery/apps/api/v1/views/courses.py#L266-L269).  All of the fields in the 'requires review' list are handled by the serializer, so this code serves the purpose and we simply don't consider whether the image or video has changed.  Arguably the unit test should not include those in the edits made to prove non-reviewable edits don't trigger a re-review, but it made sense to me to leave them in, even though they are not part of the logic in this function

PR #2047, opened 2019-07-24
Reset our cache when memcache versions change

mikix:
Start using the currently installed python-memcached version in our
cache key prefix, so that the cache will automatically clear when
we update that package, which sometimes has format changes.


bbaker6225 (approved):
Seems reasonable

PR #2048, opened 2019-07-24
Try harder to match Pub runs

mikix:
1) If it already has a match in CM, use that
2) Else find the newest match in sibling runs
3) Else use the Pub course key

https://openedx.atlassian.net/browse/DISCO-1197

### DIFF
--- a/course_discovery/apps/publisher/utils.py
+++ b/course_discovery/apps/publisher/utils.py
@@ -136,9 +136,19 @@ def parse_datetime_field(date):
         return
 
 
-def get_course_key(publisher_course):
-    """ Construct the course key to look up a course_metadata course from a publisher course """
-    return '{org}+{number}'.format(
-        org=publisher_course.organizations.first().key,
-        number=publisher_course.number
-    )
+def find_discovery_course(publisher_course_run):
+    """ Returns the discovery course where this publisher run already lives or should live. """
+    # Ideal situation is that it already exists
+    if publisher_course_run.discovery_counterpart:
+        return publisher_course_run.discovery_counterpart.course
+
+    # OK, it hasn't been pushed to the course metadata tables yet. Let's find where it will live.
+    # We are intentionally not calling course.discovery_counterpart, because that simply looks at the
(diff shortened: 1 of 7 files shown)


course_discovery/apps/publisher/utils.py
```diff
@@ -136,9 +136,19 @@ def parse_datetime_field(date):
...
-def get_course_key(publisher_course):
-    """ Construct the course key to look up a course_metadata course from a publisher course """
-    return '{org}+{number}'.format(
-        org=publisher_course.organizations.first().key,
-        number=publisher_course.number
-    )
+def find_discovery_course(publisher_course_run):
+    """ Returns the discovery course where this publisher run already lives or should live. """
+    # Ideal situation is that it already exists
+    if publisher_course_run.discovery_counterpart:
```
Dillon-Dumesnil, line 142: This should probably be wrapped in a try/except since if it doesn't exist, it will return a DoesNotExist

course_discovery/apps/publisher/utils.py
```diff
@@ -136,9 +136,19 @@ def parse_datetime_field(date):
...
+        return publisher_course_run.discovery_counterpart.course
```
Dillon-Dumesnil, line 143: Since this is a cached property, calling it twice like this doesn't perform two database hits, right?

Dillon-Dumesnil (approved):
Curious if PubFE tests or these tests will be done first. LGTM to me once those are added

course_discovery/apps/publisher/utils.py
mikix: Right. Though this method in general is not shy about making queries.

course_discovery/apps/publisher/utils.py
mikix: GP

mikix: Ah, nope. The `@property` catches it for us.

mikix: Oddly enough, the CourseRun.discovery_counterpart catches it. The Course one does not. :man_shrugging:

PR #2049, opened 2019-07-24
Use base manager when building fixture.

kdmccormick:
Also, 500 when we fail to load objects instead of infinite-looping.

It'd be good to write a test before merging, but we can do that in another PR if you need this ASAP @schenedx .

@edx/masters-neem 
https://openedx.atlassian.net/browse/EDUCATOR-4531


iloveagent57 (approved):
Looks legit.

schenedx (approved):
👍 
I am fine merging this PR as it is for now, and tackle the unit tests later. 
We certainly want this fix to be in ASAP to unblock the Sandbox creation for Curtin.

PR #2052, opened 2019-07-25
Allow looking up org users by UUID

mikix:
Allow both UUID and the old database ID too.

The course serialization gives UUIDs for orgs, not IDs. Which seems appropriate. So when getting a list of org users, I figured it made sense to open up the API to UUIDs too.


Dillon-Dumesnil (approved):
LGTM. Write Tests

PR #2053, opened 2019-07-26
Add presave hooks to enforce external_key uniqueness

jansenk:
Add pre_save hooks to CourseRun, CurriculumCourseMembership, and Curriculum
to make sure that no two CourseRuns that have been 'linked' together within
a Course, Curriculum, or Program have the same external course key

### DIFF
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -162,12 +162,12 @@ def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # py
     If the course is not associated with a program, we will still verify that the external_key
     is unique within course runs in the course
     """
-    if instance.id:
-        old_course_run = CourseRun.everything.get(pk=instance.id)
-        if instance.external_key == old_course_run.external_key and instance.course == old_course_run.course:
-            return
     if not instance.external_key:
         return
+    if instance.id:
+        old_course_run = CourseRun.everything.get(pk=instance.pk)
+        if instance.external_key == old_course_run.external_key and instance.course == old_course_run.course:
+            return
 
     course = instance.course
     curricula = course.degree_course_curricula.select_related('program').all()
@@ -196,7 +196,7 @@ def ensure_external_key_uniquness__curriculum(sender, instance, **kwargs):  # py
     if not instance.id:
         return  # If not instance.id, we can't access course_curriculum, so we can't do anything
     if instance.program:
-        old_curriculum = Curriculum.objects.get(pk=instance.id)
+        old_curriculum = Curriculum.objects.get(pk=instance.pk)
         if old_curriculum.program and instance.program.id == old_curriculum.program.id:
             return
 
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -163,7 +163,7 @@ def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # py
     is unique within course runs in the course
     """
     if instance.id:
-        old_course_run = CourseRun.objects.get(pk=instance.id)
+        old_course_run = CourseRun.everything.get(pk=instance.id)
         if instance.external_key == old_course_run.external_key and instance.course == old_course_run.course:
             return
     if not instance.external_key:
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -166,6 +166,8 @@ def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # py
         old_course_run = CourseRun.objects.get(pk=instance.id)
         if instance.external_key == old_course_run.external_key and instance.course == old_course_run.course:
             return
+    if not instance.external_key:
+        return
 
     course = instance.course
     curricula = course.degree_course_curricula.select_related('program').all()
@@ -225,17 +227,14 @@ def _build_external_key_sets(course_runs):
     return external_key_set, course_run_ids
 
 
-def _duplicate_external_key_message(course_run, curriculum=None):
-    message = "Duplicate external_key found: external_key={} course_run={} course={}".format(
-        course_run.external_key,
-        course_run,
-        course_run.course
-    )
-    # Currently curriculum and program are never being provided
-    if curriculum:
-        message = message + " curriculum={}".format(curriculum)
-        if curriculum.program:
-            message = message + " program={}".format(curriculum.program)
+def _duplicate_external_key_message(course_runs):
+    message = 'Duplicate external_key{} found: '.format('s' if len(course_runs) > 1 else '')
+    for course_run in course_runs:
+        message += ' [ external_key={} course_run={} course={} ]'.format(
+            course_run.external_key,
+            course_run,
+            course_run.course
+        )
     return message
 
 
@@ -263,16 +262,16 @@ def check_curricula_and_related_programs_for_duplicate_external_key(curricula, c
 
     # Get the first course run in the curricula or programs that have a duplicate external key
     # but aren't the course runs we're given
-    course_run = CourseRun.objects.filter(
+    course_runs = CourseRun.objects.filter(
         ~Q(id__in=course_run_ids),
         Q(external_key__in=external_key_set),
         (
             Q(course__degree_course_curricula__program__in=programs) |
             Q(course__degree_course_curricula__in=programless_curricula)
         ),
-    ).select_related('course').distinct().first()
-    if course_run:
-        message = _duplicate_external_key_message(course_run)
+    ).select_related('course').distinct().all()
+    if course_runs:
+        message = _duplicate_external_key_message(course_runs)
         raise ValidationError(message)
 
 
@@ -292,5 +291,5 @@ def check_course_runs_within_course_for_duplicate_external_key(course, specific_
     for course_run in course.course_runs.all():
         external_key = course_run.external_key
         if external_key == specific_course_run.external_key and course_run != specific_course_run:
-            message = _duplicate_external_key_message(course_run)
+            message = _duplicate_external_key_message([course_run])
             raise ValidationError(message)
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -3,6 +3,7 @@ import logging
 import waffle
 from django.apps import apps
 from django.core.exceptions import ValidationError
+from django.db.models import Q
 from django.db.models.signals import post_delete, post_save, pre_delete, pre_save
 from django.dispatch import receiver
 
@@ -150,7 +151,7 @@ for model in apps.get_app_config('course_metadata').get_models():
 
 
 @receiver(pre_save, sender=CourseRun)
-def save_course_run(sender, instance, **kwargs):  # pylint: disable=unused-argument
+def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # pylint: disable=unused-argument
     """
     Pre-save hook to validate that when a course run is saved, that its external_key is
     unique.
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+        return  # If not instance.id, we can't access course_curriculum, so we can't do anything
+
+    course_run_map = {}
+    for course in instance.course_curriculum.all():
+        for course_run in course.course_runs.filter(external_key__isnull=False):
+            course_run_map[course_run.external_key] = course_run
+    check_curriculum(instance, course_run_map, set())
+
+
+def check_curriculum(curriculum, course_run_map, checked_courses):
```
jansenk: probably rename this

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+        course key as a course run in `external_course_map` (but isn't the course run in `external_course_map`),
+        we will return a tuple of:  (course_run, curriculum) of the 'offending' duplicate course key that we found
+    """
+    for curriculum in program.curricula.all():
+        offending_course_run = check_curriculum_for_duplicate_external_key(curriculum, course_run_map, checked_courses)
+        if offending_course_run:
+            return (offending_course_run, curriculum)
+
+
+def check_curriculum_for_duplicate_external_key(curriculum, course_run_map, checked_courses):
```
jansenk: add comment

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+    return message
+
+
+def check_program_for_duplicate_external_key(program, course_run_map, checked_courses):
+    """
+    Helper function for verifying the uniqueness of external course keys within a program
+
+    Parameters:
+        - program: program in which we are searching for potential duplicate course keys
+        - external_key_map: dict mapping external course keys to their course runs.
```
kdmccormick: Update docstring params (same for a couple of the other functions)

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+        checked_courses.add(course)
+
+
+def check_course_for_duplicate_external_key(course, course_run_map):
+    """
+    Helper function for verifying the uniqueness of external course keys within a course
+
+    Parameters:
+        - course: course in which we are searching for potential duplicate course keys
+        - external_key_map: dict mapping external course keys to their course runs.
```
kdmccormick: Docstring params

kdmccormick (commented):
Looks pretty comprehensive -- good stuff.

Still figuring out exactly how, but I think there are some places that you could remove code and save database time, while keeping the output of your algorithm the same.

jansenk:
> Looks pretty comprehensive -- good stuff.
> 
> Still figuring out exactly how, but I think there are some places that you could remove code and save database time, while keeping the output of your algorithm the same.

I agree. I'm still poking at it but i'm sure there's a way to optimize further

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+    course = instance.course
+    curricula = course.degree_course_curricula
+    if not curricula.all().exists():
+        duplicate_course_run = check_course_for_duplicate_external_key(course, course_run_map)
+        if duplicate_course_run:
+            message = _duplicate_external_key_message(duplicate_course_run)
+            raise ValidationError(message)
+    else:
+        seen_courses = set()
+        for curriculum in curricula.all():
```
kdmccormick: Imagine the course run is in both curricula X and Y, which are both part of program P.
This will:
* check_curricula(X):
    * check_program_for_dupes(P)
        * check_curriculum_for_dupes(X)
        * check_curriculum_for_dupes(Y)
* check_curricula(Y):
    * check_program_for_dupes(P)
        * check_curriculum_for_dupes(X)
        * check_curriculum_for_dupes(Y)

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -271,3 +274,254 @@ def test_course_curriculum_membership_side_effect_not_masters(self):
...
+    def setUpTestData(cls):
+        """
+        Sets up the tree for testing external course keys
+
+        program          1         2
+                        / \\       |
+        curriculum    1    2      3
+                      |     \\   / \\
+        course        1        2     3
+        """
```
kdmccormick, line 335: Love the ascii visuals

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -271,3 +274,254 @@ def test_course_curriculum_membership_side_effect_not_masters(self):
...
+            curriculum_4.program = self.program_1
+            curriculum_4.save()
+        curriculum_4.refresh_from_db()
+        self.assertEqual(curriculum_4.program, new_program)
+
+    """
+    ---------------------------------------------------------------------------------------
+    The tests below are to test that the validation still works in an incomplete hierarchy
+    ---------------------------------------------------------------------------------------
+    """
```
kdmccormick, line 497: If these all test a specific thing and don't share data with the other tests, I recommend splitting them out in to their own test class.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -271,3 +274,254 @@ def test_course_curriculum_membership_side_effect_not_masters(self):
...
+        cls.curriculum_3 = factories.CurriculumFactory(
+            name='curriculum_3',
+            program=cls.program_2,
+        )
+        cls.curriculums = [None, cls.curriculum_1, cls.curriculum_2, cls.curriculum_3]
+        cls._add_courses_to_curriculum(cls.curriculum_1, cls.course_1)
+        cls._add_courses_to_curriculum(cls.curriculum_2, cls.course_2)
+        cls._add_courses_to_curriculum(cls.curriculum_3, cls.course_2, cls.course_3)
+
+    @classmethod
```
kdmccormick: The helpers that don't use `cls`/`self` could be `@staticmethod`s

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -271,3 +274,254 @@ def test_course_curriculum_membership_side_effect_not_masters(self):
         for course_run in self.course_runs:
             for seat in course_run.seats.all():
                 self.assertNotEqual(seat.type, Seat.MASTERS)
+
+
+@ddt.ddt
+class ExternalCourseKeyTests(TestCase):
```
kdmccormick: Great exhaustive testing :100:

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+        - course: course in which we are searching for potential duplicate course keys
+        - external_key_map: dict mapping external course keys to their course runs.
+                            These are the external course keys that we looking for potential duplicates of
+
+    Returns:
+        If a course run is found under `course` that has the same external
+        course key as a course run in `external_course_map` (but isn't the course run in `external_course_map`),
+        this function will return the 'offending' course_run
+    """
+    for course_run in course.course_runs.all():
```
zacharis278, line 291: if you're looking for optimizations might I suggest querying course runs filtered by external keys in the map to start instead of calling .all() and looping in python.

I think you could actually expand this approach all the way up the stack via a giant join so you're only ever checking curricula, courses, course_runs that are a potential conflict.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+
+
+@receiver(pre_save, sender=CurriculumCourseMembership)
+def save_curriculum_course_membership(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Pre-save hook to validate that if a curriculum_course_membership is created or modified, the
+    external_keys for the course
+    """
+    course_runs = instance.course.course_runs.filter(external_key__isnull=False)
+    course_run_map = {
```
kdmccormick: The "make a map from ext keys to course runs" logic is duplicated across the three signal handlers; could you just pass in a list of course runs and have `check_curriculum` build the map?

kdmccormick (changes requested):

zacharis278:
What does the querycount look like on this as you add more curricula/courses.  I'm thinking this is going to run into n+1 problems at first glance

kdmccormick:
@Zacharis278 I agree with the n+1 concern, but I think starting with the course run with matching external keys and then "looking up" the tree would also be difficult, specifically when you have many course runs in question.

I think you can leverage the Django ORM like Dave was showing in all-hands this morning to do it efficiently top-down, something like:

```python

def check_program(program):

    # Grab all run/key pairs at once with two big queries
    course_ids = CurriculumCourseMembership.objects.filter(
        curriculum_id__in=program.curricula.all()
    ).values_list('course_id')
    run_pairs = CourseRun.objects.filter(course_id__in=course_ids).values_list('external_key', 'id')

    # Check for conflicts
    run_ids_by_ext_key = {}
    for ext_key, run_id in run_pairs:
        if not ext_key:
            continue
        existing_run_id = run_ids_by_ext_key.get(ext_key)
        if not existing_run_id:
            print('{} refers to {}'.format(ext_key, run_id))
            run_ids_by_ext_key[ext_key] = run_id
        elif existing_run_id != run_id:
            print('error: {} refers to both {} and {}'.format(ext_key, existing_run_id, run_id))
```

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
 for model in apps.get_app_config('course_metadata').get_models():
     for signal in (post_save, post_delete):
         signal.connect(api_change_receiver, sender=model)
+
+
+@receiver(pre_save, sender=CourseRun)
+def save_course_run(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
schenedx: Rather than name the function this way, I prefer name the function as summary of what it does. Maybe `ensure_external_key_uniqueness`?

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +147,147 @@ def _seat_type_exists(course_modes, seat_type):
...
+    """
+    course_runs = instance.course.course_runs.filter(external_key__isnull=False)
+    course_run_map = {
+        course_run.external_key: course_run for course_run in course_runs
+    }
+    check_curriculum(instance.curriculum, course_run_map, set())
+
+
+@receiver(pre_save, sender=Curriculum)
+def save_curriculum(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
schenedx: Again, it's best to rename it to tell next developer what this function does. I suggest `ensure_external_key_uniqueness_across_curriculum`

schenedx (commented):
Besides the two naming suggestion, I don't have any more feedback in addition to what you already got. Don't be afraid to name functions with a lot of words.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +148,151 @@ def _seat_type_exists(course_modes, seat_type):
...
+    programless_curricula = set()
+    for curriculum in curricula:
+        if curriculum.program:
+            programs.add(curriculum.program)
+        else:
+            programless_curricula.add(curriculum)
+
+    # Get the first course run in the curricula or programs that have a duplicate external key
+    # but aren't the course runs we're given
+    course_run = CourseRun.objects.filter(
```
kdmccormick: Nice :ok_hand:

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +148,151 @@ def _seat_type_exists(course_modes, seat_type):
...
+        ~Q(id__in=course_run_ids),
+        Q(external_key__in=external_key_set),
+        (
+            Q(course__degree_course_curricula__program__in=programs) |
+            Q(course__degree_course_curricula__in=programless_curricula)
+        ),
+    ).select_related('course').distinct().first()
+    if course_run:
+        message = _duplicate_external_key_message(course_run)
+        raise ValidationError(message)
```
kdmccormick, line 275: Consider putting *all* the offending course runs in the message. Might make it easier to fix an issue like someone accidentally linking the wrong curriculum and creating dozens of conflicts.

kdmccormick (approved):
Looks great! One small nit.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -144,3 +148,148 @@ def _seat_type_exists(course_modes, seat_type):
...
+    unique.
+
+    If the course is associated with a program through a Curriculum, we will verify that
+    the external course key is unique across all programs it is assocaited with.
+
+    If the course is not associated with a program, we will still verify that the external_key
+    is unique within course runs in the course
+    """
+    if instance.id:
+        old_course_run = CourseRun.everything.get(pk=instance.id)
```
mikix: Minor nit, but these lines should be consistent - either check for and pass in `instance.pk` or assume `id` is the primary key and get `id=instance.id`. Otherwise, you're using the `pk` abstraction to no purpose. (using `pk` in both places is my vote)

The pk/id mixing is done elsewhere in this file.

mikix (approved):
I don't have enough context on curricula and external_key to be super useful here. One small code comment below and one approach comment here:

This is adding a lot of complexity and at minimum, one extra lookup on every CourseRun save, which is a frequent thing. Could we not just ignore uniqueness on `save` but check it on access? Like, if a partner is hitting our API with their own external key, do a `get(external_key=incoming_key)`, catch the MoreThanOne exception, and present a very clear error like "You set the same external key for these runs, go fix it"?

If that was discussed and discarded, I trust ya. But just sanity checking.

PR #2054, opened 2019-07-26
DISCO-1205 sort transcript and content languages on a course run by name

xnick421:


mikix (approved):
I wrote all the below before realizing that all our language names are in English. I was thinking of the languages in a user profile, where they are displayed in native characters. Not the case here, no worriies. I left the stuff below in case it was new to you.

-----

So this is an interesting problem.

So I believe order_by is going to do the `dumb` thing and sort by code point (meaning the numerical value associated with the character) - because that's happening at the database level and it wouldn't know what locale to use.

Which means that z (0x7a) will sort before ŝ (0x15d) where we really want ŝ to sort in with the other esses.

This doesn't always matter. But those language tag names are going to be goofy, with all sorts of unicode characters. So we might want to try harder here, else finding a language is going to be a hunt.

To do this right, we'd have to sort in Python, using the locale of the request.

course_discovery/apps/api/serializers.py
```diff
@@ -716,11 +716,12 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
...
     course = serializers.SlugRelatedField(required=True, slug_field='key', queryset=Course.objects.filter_drafts())
     content_language = serializers.SlugRelatedField(
-        required=False, allow_null=True, slug_field='code', source='language', queryset=LanguageTag.objects.all(),
+        required=False, allow_null=True, slug_field='code', source='language',
+        queryset=LanguageTag.objects.all().order_by('name'),
         help_text=_('Language in which the course is administered')
     )
     transcript_languages = serializers.SlugRelatedField(
-        required=False, many=True, slug_field='code', queryset=LanguageTag.objects.all()
+        required=False, many=True, slug_field='code', queryset=LanguageTag.objects.all().order_by('name')
```
jlajoie, line 724: @xnick421 you're good here, any reference to this serializer will prefetch (which occurs down at line 750) using this query so it'll always be in the right order 👍

PR #2055, opened 2019-07-26
Retrieve a new access token at each stage of the refresh pipeline.

doctoryes:
https://openedx.atlassian.net/browse/DE-1606

The course download stage via the LMS course API is now taking 1 hour 45 minutes and retrieving 10,000+ course runs. This interval makes the access token expire before the next stage can connect to ECommerce. This PR makes the management command retrieve a new access token at each stage of the refresh pipeline.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
@@ -63,6 +63,21 @@ class Command(BaseCommand):
             help='The short code for a specific partner to refresh.'
         )
 
+    def get_access_token(self, url, client_id, client_secret, token_type):
+        """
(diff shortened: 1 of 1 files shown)


jlajoie (approved):
Seems super reasonable, thank you for making this change and diagnosing the problem! Do you know what caused this to be a problem now, if our token expiry periods changed or anythign? I wasn't under the impression that these jobs were taking longer than usual to complete, so it seems weird that our tokens are expiring.

pwnage101 (approved):
LGTM!

doctoryes:
The token started expiring when we shifted course-discovery from using OIDC DOP-based auth tokens to JWT DOT-based auth tokens. The expiry time is only 30 seconds for our JWT tokens issued in the edx-rest-api-client module.

PR #2058, opened 2019-07-29
Expose is_enrollable,is_marketable on runs in API

bbaker6225:
[DISCO-1242](https://openedx.atlassian.net/browse/DISCO-1242)


rgraber (approved):
LGTM

inventhouse (approved):
👍

course_discovery/apps/course_metadata/models.py
```diff
@@ -1430,7 +1430,7 @@ def is_marketable(self):
         """
         is_published = self.status == CourseRunStatus.Published
 
-        return is_published and self.seats.exists() and bool(self.marketing_url)
+        return is_published and bool(self.seats) and bool(self.marketing_url)
```
bbaker6225: `Additionally, if a some_queryset has not yet been evaluated, but you know that it will be at some point, then using some_queryset.exists() will do more overall work (one query for the existence check plus an extra one to later retrieve the results) than simply using bool(some_queryset), which retrieves the results and then checks if any were returned.` <-- django queryset docs

PR #2059, opened 2019-07-29
Allow user to select run to re-run

ciduarte:
DISCO-1219

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -172,13 +172,16 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
             self.mock_post_to_studio(new_key, rerun_key=self.draft_course_run.key)
             course.canonical_course_run = self.draft_course_run
             course.save()
+            rerun = self.draft_course_run.key
         else:
             self.mock_post_to_studio(new_key)
+            rerun = None
 
         response = self.client.post(url, {
             'course': course.key,
             'start': '2000-01-01T00:00:00Z',
             'end': '2001-01-01T00:00:00Z',
+            'rerun': rerun,
         }, format='json')
 
         self.assertEqual(response.status_code, 201)
@@ -315,6 +318,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
                 'course': course.key,
                 'start': '2000-01-01T00:00:00Z',
                 'end': '2001-01-01T00:00:00Z',
+                'rerun': self.draft_course_run.key,
             }, format='json')
         self.assertEqual(response.status_code, 201)
 
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -169,10 +169,10 @@ class CourseRunViewSet(viewsets.ModelViewSet):
         return super(CourseRunViewSet, self).list(request, *args, **kwargs)
 
     @classmethod
-    def push_to_studio(cls, request, course_run, create=False, old_course_run=None):
+    def push_to_studio(cls, request, course_run, create=False, old_course_run_key=None):
         if course_run.course.partner.studio_url:
             api = StudioAPI(course_run.course.partner.studio_api_client)
-            api.push_to_studio(course_run, create, old_course_run, user=request.user)
+            api.push_to_studio(course_run, create, old_course_run_key, user=request.user)
         else:
             log.info('Not pushing course run info for %s to Studio as partner %s has no studio_url set.',
                      course_run.key, course_run.course.partner.short_code)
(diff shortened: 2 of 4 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -224,12 +224,12 @@ def create_run_helper(self, run_data, request=None):
 
         # Set canonical course run if needed (done this way to match historical behavior - but shouldn't this be
         # updated *each* time we make a new run?)
-        if not old_course_run:
+        if not old_course_run_key:
             course.canonical_course_run = course_run
             course.save()
```
mikix, line 229: This is a behavior change - you're updating canonical_course_run whenever the course team wants to base off a different run. You might want to change this to `if not course.canonical_course_run:`

(Now, the idea of constantly updating the canonical course run is something we've toyed with, but haven't pulled the trigger on yet. But if we did, it would probably be to just remove the `if` condition altogether and make the new rerun the canonical run, rather than the base-run.)

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -212,7 +212,7 @@ def create_run_helper(self, run_data, request=None):
         # The serializer will attempt to retrieve the draft version of the Course
         course = Course.objects.filter_drafts().get(key=course_key)
         course = ensure_draft_world(course)
-        old_course_run = course.canonical_course_run
+        old_course_run_key = run_data.pop('rerun', '')
```
mikix: You already updated publisher-frontend? Because with this change, you're breaking this API a bit (current users that don't pass in `rerun` are no longer making reruns). But since we're the only consumers, that's fine, as long as you've updated PFE.

Although, do we ever not want to do a rerun? I would argue you should use the canonical course key run as a backup, if one exists. That would avoid breaking the API too (not that it matters).

Also as a very tiny nit, you use the empty string as signifier that there is no provided rerun -- I believe `None` is more the norm for that kind of status (not-provided / does-not-exist / etc), which means you can just drop the second arg to `pop`.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -172,13 +172,16 @@ def test_create_sets_canonical_course_run(self, has_canonical_run):
             self.mock_post_to_studio(new_key, rerun_key=self.draft_course_run.key)
             course.canonical_course_run = self.draft_course_run
             course.save()
+            rerun = self.draft_course_run.key
         else:
             self.mock_post_to_studio(new_key)
+            rerun = ''
```
mikix: I understand you're passing empty so that it winds its way down to the `if create and old_course_run_key` line, but it's not a super realistic test case - users are more likely to just not pass in the field altogether (which to be fair, in your current code is then upgraded to an empty string), but seems unlikely that they'd pass in the empty string from the start. 

Either way, this is fine. Just a comment about tests that hit endpoints ideally being implementation-dumb about the details of endpoint internals / having realistic code paths in tests.

mikix (approved):
How graceful is the error handling/message if I give a busted key like `mikesrealgoodcoursetrustmethisexists`?

Besides that, seems fine, modulo the behavior change around canonical test runs.

course_discovery/apps/api/v1/views/course_runs.py
ciduarte: I've updated PFE to pass in a rerun value on course run creation, it defaults to the course run with the latest start date if nothing is selected (per Carol). I believe the only time we don't want to do a rerun is when we're creating the first run of a course. In this case `old_course_run_key` is `None` in order to trigger `create_course_run_in_studio` rather than `create_course_rerun_in_studio` when it reaches `push_to_studio` in utils.py. Because there is a default value on PFE, I don't believe we need to use the canonical as backup. Thoughts?

RE: busted key, this change on PFE is just a drop down menu that's populated by `courseInfo` data. The user can't manually input a course key that doesn't exist. Do you think it's necessary to verify that the course run exists in studio prior to populating that drop down list?

course_discovery/apps/api/v1/views/course_runs.py
ciduarte: Good catch!

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
ciduarte: You're right, it's probably more correct to set `rerun = None`, since that's what the `pop('rerun', None)` will default as now rather than an empty string.

mikix: Well, even if you do that, you're still passing in json that's doing `rerun: None` - but it's more realistic to just not pass `rerun` in as part of the json. But that is an extremely minor nit. I'm just chatting, not reviewing at this point.

course_discovery/apps/api/v1/views/course_runs.py
mikix: `Because there is a default value on PFE, I don't believe we need to use the canonical as backup. Thoughts?`

That's assuming that PFE is your only consumer ever. Ideally we're a little more defensive in our endpoint design and don't tie ourselves so closely to the current code of the frontend at this moment in time.

It also means that an endpoint consumer always needs a rerun key ready to hand over (which PFE today has, but again, I'm talking about endpoint design in general now).

Or PFE might gain a bug where it doesn't send a rerun key.

All around, I would argue it's better to be defensive, since we know we do in fact want to base the rerun off the canonical run if we don't have a better idea (as we've been doing so far).

PR #2060, opened 2019-07-30
Fix is_enrollable logic to handle archived runs

bbaker6225:
We only consider enrollment_end and enrollment_start for the
enrollability check on course runs, since archived runs may have ended
but they can always be enrolled in (at least at the time of writing
this). This logic should also hold for all other types of runs,
assuming they are correctly setting enrollment end/start.


mikix (approved):
:+1: 

I don't know if it's worth it, but the docstring could mention that we specifically don't want to consider end date because of archived courses, to avoid fools like me coming by later and mucking it up with refactors.

mattdrayer:
Re: docstring -- it was worth it 😄

PR #2062, opened 2019-07-31
DISCO-1222: Allow changing course number after creation

Dillon-Dumesnil:


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -762,6 +763,8 @@ def test_update_success(self, method):
 
         course = Course.everything.get(uuid=self.course.uuid, draft=True)
         self.assertEqual(course.title, 'Course title')
+        expected_key = self.course.authoring_organizations.first().key + '+' + number  # pylint: disable=no-member
```
jlajoie, line 766: Isn't there some weird thing where we shouldn't need to have the Organization be part of the Course key? Or is this literally that situation

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -271,6 +271,13 @@ def update_course(self, data, partial=False):
         # Get and validate object serializer
         course = self.get_object()
         course = ensure_draft_world(course)  # always work on drafts
+
+        number = data.pop('number', None)
```
jlajoie, line 275: Is there a reason we're using this as `number` rather than `key`? Typically more comfortable keeping REST fields not needing to be something guessable

jlajoie (changes requested):
Minor request about naming of fields, if you have a good reason for it not being called `key` since that's what's representing in our database I'm all ears.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
Dillon-Dumesnil: This is just that situation. Keys always have the org that authored it so this is the way we construct keys.

jlajoie (approved):
👍

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: We made the decision to do the updated key creation in the backend so we could maintain all of the validation logic here. Unfortunately this does mean sending back a field that is not represented in the model or serializer because we compute it before saving.

PR #2063, opened 2019-08-02
Updates Salesforce Utils for Finalized Record Fields

jlajoie:


course_discovery/apps/course_metadata/models.py
```diff
@@ -337,6 +337,7 @@ class Organization(TimeStampedModel):
         null=True, blank=True, help_text=_('Logo to be displayed on certificates. If this logo is the same as '
                                            'logo_image_url, copy and paste the same value to both fields.')
     )
+    salesforce_id = models.CharField(max_length=255, null=True, blank=True)
```
Dillon-Dumesnil: Maybe throw in a comment about what this corresponds to in Salesforce. i.e. Org = Account, Course = Course, Case = Comments, Course Run = Course Run

course_discovery/apps/core/models.py
```diff
@@ -196,5 +196,11 @@ class SalesforceConfiguration(models.Model):
         verbose_name=_('Is a Salesforce Sandbox?'),
         default=True
     )
+    case_record_type_id = models.CharField(
+        max_length=255,
+        blank=True,
+        verbose_name=_('Case Record Type Id'),
+        default=''
```
Dillon-Dumesnil: Nit: Just use `null=True` so it has parity with all of the other id fields.

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: What is the format for the ids? Can we not use an int field or anything?

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
-        account_records = accounts.get('records', [])
-        if len(account_records) == 1:
-            return account_records[0]
-        raise Exception('{} records found for Name={}. Expected 1.'.format(len(account_records), name))
+    def create_course(self, course):
+        if not course.salesforce_id:
+            # If we don't have the parent Salesforce Organization ID for this course, create one
+            organization = course.authoring_organizations.first()
+            if not organization.salesforce_id:
+                self.create_publisher_organization(organization)
```
Dillon-Dumesnil, line 109: I think this line should be `self.create_account(organization)`

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
-                'Name': course.title,
-                'Course_Number__c': course.key,
-                'Account__c': account.get('Id'),
+    def create_course_run(self, course_run):
+        if not course_run.salesforce_id:
+            # If we don't have the parent Salesforce Course ID for this course run, create one
+            if not course_run.course.salesforce_id:
+                self.create_course(course_run.course)
+            sf_course_run = self.client.Course_Run__c.create({
+                'Course__c': course_run.course.salesforce_id,
```
Dillon-Dumesnil, line 134: I'm sure you'll catch it during testing, but I think you might need to refresh the course run from the db for this line to work right in the case that you had to create the course in salesforce. I would imagine the local course_run object you have would not get the updated course from the save inside of `self.create_course` so `course_run.course.salesforce_id` will still be `''`.

course_discovery/apps/course_metadata/salesforce.py
Dillon-Dumesnil: I also don't see why we need to do this here because it doesn't look like you end up doing anything with the organization or account that would have been created.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
+                'Link_to_Admin_Portal__c': '{url}/admin/course_metadata/courserun/{id}/change/'.format(
+                    url=self.partner.site, id=course_run.id
+                ),
+                'Course_Start_Date__c': course_run.end.isoformat(),
+                'Course_Run_Date__c': course_run.end.isoformat(),
+                'Publisher_Status__c': course_run.status,
+                'Course_Run_Name__c': course_run.title,
+                'Expected_Go_Live_Date__c': course_run.go_live_date,
```
Dillon-Dumesnil, line 142: Not sure if it matters, but if the idea is to expose this to PCs and such, I think they prefer `Publish Date` as opposed to `go_live_date`. We did `go_live_date` to try and help with overloading the term publish in the backend. If it's just a field though and the display name can be Publish Date, all good to go.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
-            account_key = course.authoring_organizations.first().key
-            # TODO: Check to see if first() is okay
-            account = self.get_or_create_account(account_key=account_key)
-            return self.client.Course__c.create({
-                'Name': course.title,
-                'Course_Number__c': course.key,
-                'Account__c': account.get('Id'),
+    def create_course_run(self, course_run):
+        if not course_run.salesforce_id:
+            # If we don't have the parent Salesforce Course ID for this course run, create one
```
Dillon-Dumesnil: Don't really care, but this comment doesn't seem to add much value. The code is rather self-documenting in this case.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
-            'Name',
-        ]
-        accounts = self._query("SELECT {} FROM Account WHERE Name='{}'", ','.join(fields), name)
-        account_records = accounts.get('records', [])
-        if len(account_records) == 1:
-            return account_records[0]
-        raise Exception('{} records found for Name={}. Expected 1.'.format(len(account_records), name))
+    def create_course(self, course):
+        if not course.salesforce_id:
+            # If we don't have the parent Salesforce Organization ID for this course, create one
```
Dillon-Dumesnil: Don't really care, but this comment doesn't seem to add much value. The code is rather self-documenting in this case.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
-            'Account__c',
-        ]
-        courses = self._query("SELECT {} FROM Course__c WHERE Course_Number__c='{}'", ','.join(fields), course_key)
-        course_records = courses.get('records', [])
-        if len(course_records) == 1:
-            return course_records[0]
-        raise Exception('{} records found for Name={}. Expected 1.'.format(len(course_records), course_key))
+    def create_case_for_course(self, course):
+        if not course.salesforce_case_id:
+            # If we don't have the parent Salesforce Course ID for this course, create one
```
Dillon-Dumesnil: Don't really care, but this comment doesn't seem to add much value. The code is rather self-documenting in this case.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
+                'Course__c': course.salesforce_id,
+                'Status': 'Open',
+                'Origin': 'Publisher',
+                'Subject': '{} Comments'.format(course.title),
+                'Description': 'This case is required to be Open for the Publisher comment service.'
+            }
+            case_record_type_id = self.partner.salesforce.case_record_type_id
+            # Only add the record type ID if it's configured, this is not a required field
+            if case_record_type_id:
+                case['RecordTypeId'] = self.partner.salesforce.case_record_type_id
```
Dillon-Dumesnil: Either use the defined `case_record_type_id` or update the if statement to `if self.partner.salesforce.case_record_type_id:` and delete the defining the variable.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
-        salesforce_course_run = self.get_course_run_by_name(course_run.key)
-        # Only create the Course Run if it doesn't already exist
-        if not salesforce_course_run:
-            course = self.get_or_create_course(course_run.course)
-            return self.client.Course_Runs__c.create({
-                'Course_Run_Name__c': course_run.title,
-                'Course_Run_Number__c': course_run.key,
-                'Parent_course_name__c': course.get('Id'),
+    def create_contact(self, user):
+        if not user.salesforce_id:
```
Dillon-Dumesnil: Was `user.salesforce_id` defined in a previous PR?

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
+            organization = Organization.user_organizations(user).first()
+            if not organization.salesforce_id:
+                self.create_account(organization)
+            # Last name cannot be None for the SF Contact Record, so we default to just the person's username
+            sf_contact = self.client.Contact.create({
+                'LastName': user.last_name or user.username,
+                'FirstName': user.first_name,
+                'Email': user.email,
+                'AccountId': Organization.user_organizations(user).first().salesforce_id
```
Dillon-Dumesnil: Can you just use `organization.salesforce_id`? Might need to refresh from db, but prevents having to run this same line twice.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
+    def _format_user_comment_body(user, body, course_run_key=None):
+        if user.first_name and user.last_name:
+            user_message = '[User]\n{first_name} {last_name} ({email})'.format(
+                first_name=user.first_name,
+                last_name=user.last_name,
+                email=user.email,
+            )
+        else:
+            user_message = '[User]\n{email}'.format(email=user.email)
+        course_run_message = '[Course Run]\n{course_run_key}\n\n'.format(course_run_key=course_run_key)
```
Dillon-Dumesnil: Nit: Use the check about `if course_run_key` up here to the definition of the variable instead as part of the return value formatting.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -86,105 +91,157 @@ def soql_escape(self, soql):
...
+                )
+            )
+            parsed_comments = [self._parse_user_comment_body(comment) for comment in comments.get('records')]
+            return parsed_comments
+        return []
+
+    @staticmethod
+    def _parse_user_comment_body(comment):
+        match = re.match(
+            r"\[User\]\n?(?:^(?:.*\s\(?)?(.+@[\w\.]+)\)?$\n\n)(?:\[Course Run\]\n^(.+)$\n\n)?\[Body\]\n^(.+)",
```
Dillon-Dumesnil: Assuming this is correct because Nope

Dillon-Dumesnil (commented):
Looks like a good start. I think a lot of my comments will probably get caught during test writing, but wanted to help just in case. Let me know if you want any clarification on any of my points.

course_discovery/apps/course_metadata/models.py
jlajoie: They're Strings, ```a0n63000003lqMfAAI```

course_discovery/apps/course_metadata/salesforce.py
jlajoie: We need the salesforce_id set on the organization, which the above method will do. And good catch here (came up in tests anyway) but this was the old schema we were working with before defaulting to the base Account type in SF.

course_discovery/apps/course_metadata/salesforce.py
jlajoie: Yeah just the key, the name doesn't have to be bound to anything specifically, I think right now it's "Expected Go Live Date" but that's configurable. Seemed preferable to have parity for field names between systems.

course_discovery/apps/course_metadata/salesforce.py
jlajoie: Whole method goes away actually 👍

course_discovery/apps/course_metadata/salesforce.py
jlajoie: As long as we control the write structure (which we do) this will be correct. Then intent is we try to parse out the content as if we wrote it, if we don't get a match we'll return the default author (service account) and the raw body.

jlajoie:
Moving to include API

PR #2065, opened 2019-08-05
Ndalfonso/disco 1188 internal workflow

xnick421:


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -61,6 +62,12 @@ class CourseRunViewSet(viewsets.ModelViewSet):
     # versions of this API should only support the system default, PageNumberPagination.
     pagination_class = ProxiedPagination
 
+    INTERNAL_REVIEW_FIELDS = (
```
jlajoie: nit: Might be worth defining these fields as a constant on the model? Just worth thinking about, doesn't really matter

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -303,6 +310,21 @@ def retrieve(self, request, *args, **kwargs):
         """ Retrieve details for a course run. """
         return super(CourseRunViewSet, self).retrieve(request, *args, **kwargs)
 
+    @action(detail=True, methods=['patch'], permission_classes=[InternalUserPermission])
+    def internal_review(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        for key in request.data.keys():
+            if key not in self.INTERNAL_REVIEW_FIELDS:
+                return Response(
+                    _('Can only update status, ofac restrictions, and ofac comment'),
+                    status=status.HTTP_403_FORBIDDEN
```
jlajoie: nit: I'd argue this is more of a 400 than a 403. It's just a bad request trying to do things the endpoint doesn't allow, not that your authorized user specifically is disallowed.

course_discovery/apps/course_metadata/models.py
```diff
@@ -814,6 +814,12 @@ def editable_course_runs(cls, user, queryset):
 
 class CourseRun(DraftModelMixin, TimeStampedModel):
     """ CourseRun model. """
+    OFAC_RESTRICTION_CHOICES = (
+        ('', '--'),
```
jlajoie, line 818: This is for "not reviewed yet" right?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -303,6 +310,21 @@ def retrieve(self, request, *args, **kwargs):
         """ Retrieve details for a course run. """
         return super(CourseRunViewSet, self).retrieve(request, *args, **kwargs)
 
+    @action(detail=True, methods=['patch'], permission_classes=[InternalUserPermission])
```
jlajoie: This feels pretty hard amounts of "not REST" to me, is there a reason we're not just using the default PATCH endpoint and overriding parts of it for these users?

jlajoie (changes requested):

course_discovery/apps/api/v1/views/course_runs.py
xnick421: could even be a 422, ill change to 400

course_discovery/apps/course_metadata/models.py
xnick421: yep

course_discovery/apps/api/v1/views/course_runs.py
xnick421: ill move 'em over

course_discovery/apps/api/v1/views/course_runs.py
xnick421: i suggested it and agreed on it with dillon and jason.  just seemed like a well defined way of separating this type of update since its a little different than other updates.

course_discovery/apps/course_metadata/models.py
```diff
@@ -888,10 +894,13 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
...
-    has_ofac_restrictions = models.BooleanField(
-        default=False,
-        verbose_name=_('Add OFAC restriction text to the FAQ section of the Marketing site')
+    has_ofac_restrictions = models.NullBooleanField(
+        blank=True,
+        choices=OFAC_RESTRICTION_CHOICES,
+        default=None,
+        verbose_name=_('Add OFAC restriction text to the FAQ section of the Marketing site'),
     )
+    ofac_comment = models.TextField(blank=True, help_text='Comment related to OFAC restriction')
```
mikix, line 903: This feels like such a process-oriented field on what is otherwise a definition-oriented model. The other fields are evergreen, while this one's content is tied to a moment in time. Plus, this comment field might be "temporary" until we move the country restriction list into Discovery and can just make that change directly from frontend.

Another option might be a separate model that is OneToOne with a CourseRun, to keep things cleaner? Lets you encode more metadata along with it too, and have room for expansion. Or just delete it wholesale once we no longer need `ofac_comment`.

Something like:
```
class LegalReview(TimestampedModel):
    course_run = OneToOneField...
    changed_by = Field...
    ofac_comment = Field...
```

But that is a big and subjective requested change. And is more about the design than just getting things done. If neither @jlajoie or @Dillon-Dumesnil agree, then forget it.

course_discovery/apps/course_metadata/models.py
```diff
@@ -918,6 +927,17 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
...
 
+    INTERNAL_REVIEW_FIELDS = (
+        'status',
+        'has_ofac_restrictions',
+        'ofac_comment',
+    )
+
+    INTERNAL_STATUS_TRANISTIONS = (
+        'review_by_internal',
+        'reviewed',
```
mikix: These should use CourseRunStatus.Reviewed etc naming, rather than hardcoded strings.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -275,29 +275,50 @@ def update(self, request, **kwargs):
...
+            return Response(serializer.data)
 
-        changed = reviewable_data_has_changed(
-            course_run, serializer.validated_data.items(), CourseRun.STATUS_CHANGE_EXEMPT_FIELDS)
-        save_kwargs = {}
-        response = self._update_course_run(course_run, draft, changed, serializer, request, save_kwargs)
+        else:
+            # Sending draft=False triggers the review process for unpublished courses
+            draft = request.data.pop('draft', True)  # Don't let draft parameter trickle down
+            request.data.pop('status', None)  # Status management is handled in the model
```
mikix: We popped `draft` above the serializater creation precisely because we didn't want it to be encoded in the serializer. Can we move it back up cleanly?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -275,29 +275,50 @@ def update(self, request, **kwargs):
...
+                    )
+
+            serializer.save()
+            return Response(serializer.data)
 
-        changed = reviewable_data_has_changed(
-            course_run, serializer.validated_data.items(), CourseRun.STATUS_CHANGE_EXEMPT_FIELDS)
-        save_kwargs = {}
-        response = self._update_course_run(course_run, draft, changed, serializer, request, save_kwargs)
+        else:
```
mikix: Eh, why bother having an `else` here? You unconditionally returned above.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -275,29 +275,50 @@ def update(self, request, **kwargs):
...
+            # Disallow updates on non internal review fields while course is in review
+            for key in request.data.keys():
+                if key not in CourseRun.INTERNAL_REVIEW_FIELDS:
+                    return Response(
+                        _('Can only update status, ofac restrictions, and ofac comment'),
+                        status=status.HTTP_400_BAD_REQUEST
+                    )
+
+            serializer.save()
+            return Response(serializer.data)
```
mikix: You sounded like you felt this method was getting too complicated. Maybe instead of a big `if` in the middle of this method, you could do something like:

```
def handle_internal_review_change(...):
         if (request.user.is_staff and course_run.in_review and 'status' in request.data and
                request.data['status'] in CourseRun.INTERNAL_STATUS_TRANISTIONS):
            # Disallow updates on non internal review fields while course is in review
            for key in request.data.keys():
                if key not in CourseRun.INTERNAL_REVIEW_FIELDS:
                    return Response(
                        _('Can only update status, ofac restrictions, and ofac comment'),
                        status=status.HTTP_400_BAD_REQUEST
                    )

            serializer.save()
            return Response(serializer.data)

        # Disallow internal review fields when course run is not in review
        for key in request.data.keys():
            if key in CourseRun.INTERNAL_REVIEW_FIELDS:
                return Response(
                    _('Invalid param'),
                    status=status.HTTP_400_BAD_REQUEST
                )

     return None
```

Then in the main method:
```
    response = self.handle_internal_review_change(...)
    if response:
        return response
```

Or something? Or whatever you like, just an option to keep this tighter. I agree its branching logic is getting a bit much.

course_discovery/apps/course_metadata/models.py
```diff
@@ -918,6 +927,17 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
         'pacing_type',
     ]
 
+    INTERNAL_REVIEW_FIELDS = (
+        'status',
+        'has_ofac_restrictions',
+        'ofac_comment',
+    )
+
+    INTERNAL_STATUS_TRANISTIONS = (
```
mikix: Typo in transitions.

course_discovery/apps/api/v1/views/course_runs.py
xnick421: ill break it up

Dillon-Dumesnil: Also could have 
```
if (condition):
    return handle_internal_status(params)
```

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: Also might make sense to move these status transitions into the CourseRunStatus model since that is what keeps track of all the statuses.

course_discovery/apps/api/v1/views/course_runs.py
mikix: But ideally he could also stuff the check for `!condition -> don't allow certain fields too` in there to keep all the logic in one place.

Dillon-Dumesnil: Yeah, I don't care either way. Just giving another potential instead of asking for a response that may or may not be there.

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: From my understanding, I don't have strong feelings about this.

PR #2066, opened 2019-08-06
Upgrade Django to latest micro release

mikix:
And some other upgrades alongside.


Dillon-Dumesnil (approved):
Are the .in files updated separately?

mikix:
@Dillon-Dumesnil the `.in` files don't need to be updated - they don't pin hard versions, they just say "give me Django<1.12" and the `.txt` files get output. So since there's a newer released version, the `.txt` files change after running `make upgrade`, but no `.in` changes.

PR #2070, opened 2019-08-08
Expand the SeatType table to include various meta fields

pwnage101:
DE-1410


course_discovery/apps/course_metadata/models.py
```diff
@@ -157,10 +157,50 @@ class Meta:
 class SeatType(TimeStampedModel):
     name = models.CharField(max_length=64, unique=True)
     slug = AutoSlugField(populate_from='name', slugify_function=uslugify)
+    is_paid_to_edx = models.BooleanField(
```
mikix: Whoa, don't encode one specific user of Open edX into its database tables. How about `is_paid_to_platform`?

course_discovery/apps/course_metadata/models.py
```diff
@@ -157,10 +157,50 @@ class Meta:
...
+    )
+    is_paid_to_partner = models.BooleanField(
+        default=False,
+        help_text=_('This mode is one that is paid to a partner, e.g. a masters course paid to the school.'),
+    )
+    is_id_verified = models.BooleanField(
+        default=False,
+        help_text=_('This mode requires ID verification.'),
+    )
+    is_certificate_eligible = models.BooleanField(
```
mikix: This feels redundant with a blank `certificate_type`? You could replace the column with an `@property` and avoid a risk of being out of sync.

pwnage101: Indeed, it does seem redundant, I will add the property.

course_discovery/apps/course_metadata/models.py
pwnage101: good catch.

pwnage101:
Note to self: this still needs at least a way of marking a seat type record as deprecated or inactive.  Either that or implement the version, version_start_date, and version_end_date fields described in the doc. https://openedx.atlassian.net/wiki/spaces/DE/pages/1031143609/Course+Modes+Table+for+Warehouse

Movement on this PR is also currently blocked on an upcoming meeting with masters and discovery to gain a better idea of how to architect this data correctly.

mikix:
This is being superceded by https://github.com/edx/course-discovery/pull/2185 I believe (a PR put together by the Publisher team, after talking about various needs). I'll close this out.

PR #2071, opened 2019-08-08
Add course editor to course with no draft rows

ciduarte:
DISCO-1295

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
@@ -50,14 +50,20 @@ class CourseEditorsViewSetTests(SerializationMixin, APITestCase):
         assert len(response.data['results']) == 1
         assert response.data['results'][0]['course'] == self.course.uuid
 
-    @ddt.data(True, False)
-    def test_create_for_self(self, is_staff):
-        """Verify can make self an editor. Test cases: as staff user, as non staff user"""
+    @ddt.data(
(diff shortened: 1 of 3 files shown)


Dillon-Dumesnil (approved):
Test for this? Otherwise LGTM

course_discovery/apps/api/v1/tests/test_views/test_course_editors.py
```diff
@@ -68,6 +68,23 @@ def test_create_for_self(self, is_staff):
         assert course_editor.course == course
         assert course_editor.user == self.user
 
+    @ddt.data(True, False)
+    def test_create_for_self_on_official_course(self, is_staff):
```
ciduarte: @Dillon-Dumesnil can you take a look at this test and tell me if it's valid? I basically copied the one above it, but changed line 77 to say `draft=False`

Dillon-Dumesnil: So good start for sure! What I would say to make it even better and less repeating code is update the above test to also take in a draft parameter. This can be done like so:
```
@ddt.data(
    (True, True), # Staff user on Draft Course
    (True, False), # Staff user on Official Course
    (False, True), # Non staff user on Draft Course
    (False, False), # Non staff user on Official Course
)
@ddt.unpack
def test_create_for_self(self, is_staff, draft):
```
And then maybe change the test name to better reflect this. Maybe just `test_create` or `test_create_user_and_course`. I don't know.

PR #2072, opened 2019-08-08
DISCO-1293 restrict access for new pub fe orgs

xnick421:
if a user has all of their orgs listed in the new ORGS_ON_NEW_PUB_FE setting variable, then don't allow them to access old publisher

### DIFF
--- a/course_discovery/apps/publisher/utils.py
+++ b/course_discovery/apps/publisher/utils.py
@@ -2,6 +2,8 @@
 import re
 
 from dateutil import parser
+from django.apps import apps
+from django.conf import settings
 
 from course_discovery.apps.core.models import User
 from course_discovery.apps.publisher.constants import (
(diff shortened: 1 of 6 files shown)


course_discovery/apps/publisher/utils.py
```diff
@@ -152,3 +154,29 @@ def find_discovery_course(publisher_course_run):
...
+    # We load the organization model like this due to a circular import with the course
+    # metadata models file
+    Organization = apps.get_model('course_metadata', 'Organization')
+    return Organization.user_organizations(user)
+
+
+def is_on_new_pub_fe(user):
+    """Returns if all the user's organizations have been moved to new publisher frontend"""
+    try:
+        orgs_on_new_pub_fe = base.ORGS_ON_NEW_PUB_FE.split(',')
```
jlajoie: We do this split because we can't set arrays in edx-internal I'm guessing?

course_discovery/static/sass/publisher/lib.scss
```diff
@@ -36,4 +36,3 @@
 .actions {
   margin-top: 30px;
 }
-
```
jlajoie: nit: might as well just drop this change since it's effectively nothing and we can track our git history a little better

course_discovery/apps/publisher/utils.py
xnick421: yes we just set a string in the following format:  "key1,key2,key3"

just like we do for the pub frontend whitelist

course_discovery/apps/publisher/utils.py
```diff
@@ -152,3 +154,29 @@ def find_discovery_course(publisher_course_run):
...
+
+def user_orgs(user):
+    # We load the organization model like this due to a circular import with the course
+    # metadata models file
+    Organization = apps.get_model('course_metadata', 'Organization')
+    return Organization.user_organizations(user)
+
+
+def is_on_new_pub_fe(user):
+    """Returns if all the user's organizations have been moved to new publisher frontend"""
```
jlajoie, line 167: Should we be doing this at a user level or at an org level? I'd argue that the orgs themselves should be the things determining whether or not this content is shown, not what the user belongs to? I'm guessing it's just for ease of blanket applying a big filter of "Has Access to old pub" vs not and we're cool if a user belongs to two orgs and still goes in and messes with a "new pub" course/course run?

course_discovery/static/sass/publisher/lib.scss
xnick421: ok no prob, i just saw we had an extra line in that file, ill revert

jlajoie (approved):
Just a couple of questions but otherwise looks good. Might as well put a link to the edx-internal PR in here as well just for traceability

course_discovery/apps/publisher/utils.py
xnick421: yea so if a user belong to any orgs that are not on this list, then they can still use the old site.

PR #2073, opened 2019-08-08
Report RCM pipeline failures

mikix:
Before, if a single RCM pipeline element failed, it would merely
be logged. Now, we report a command failure so it's easier to notice.

https://openedx.atlassian.net/browse/DISCO-1229

### DIFF
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -49,7 +49,9 @@ def timestamped_object_key_constructor(*args, **kwargs):  # pylint: disable=unus
     return TimestampedObjectKeyConstructor()(**kwargs)
 
 
-def set_api_timestamp(timestamp):
+def set_api_timestamp(msg):
+    timestamp = time.time()
+    logger.debug('{msg} Updating API timestamp to {timestamp}'.format(msg=msg, timestamp=timestamp))
     cache.set(API_TIMESTAMP_KEY, timestamp, None)
 
 
@@ -58,16 +60,7 @@ def api_change_receiver(sender, **kwargs):  # pylint: disable=unused-argument
     Receiver function for handling post_save and post_delete signals emitted by
     course_metadata models.
     """
-    timestamp = time.time()
-
-    logger.debug(
-        '{model_name} model changed. Updating API timestamp to {timestamp}.'.format(
-            model_name=sender.__name__,
-            timestamp=timestamp
-        )
-    )
-
-    set_api_timestamp(timestamp)
+    set_api_timestamp('{model_name} model changed.'.format(model_name=sender.__name__))
 
 
 class CompressedCacheResponse(CacheResponse):
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -218,10 +218,8 @@ class CoursesApiDataLoader(AbstractDataLoader):
             'enrollment_start': self.parse_date(body['enrollment_start']),
             'enrollment_end': self.parse_date(body['enrollment_end']),
             'hidden': body.get('hidden', False),
+            'license': body.get('license') or '',  # license cannot be None
         }
-
-        # NOTE: The license field is non-nullable.
-        defaults['license'] = body.get('license') or ''
 
         start = self.parse_date(body['start'])
         pacing_type = self.get_pacing_type(body)
--- a/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
@@ -1,7 +1,6 @@
 import concurrent.futures
 import itertools
 import logging
-import time
 
 import jwt
 import waffle
@@ -28,8 +27,10 @@ logger = logging.getLogger(__name__)
 def execute_loader(loader_class, *loader_args, **loader_kwargs):
     try:
         loader_class(*loader_args, **loader_kwargs).ingest()
+        return True
     except Exception:  # pylint: disable=broad-except
         logger.exception('%s failed!', loader_class.__name__)
+        return False
 
 
 def execute_parallel_loader(loader_class, *loader_args, **loader_kwargs):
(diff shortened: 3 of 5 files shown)


course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -48,7 +38,12 @@ def execute_parallel_loader(loader_class, *loader_args, **loader_kwargs):
     """
     connection.close()
 
-    execute_loader(loader_class, *loader_args, **loader_kwargs)
+    try:
+        loader_class(*loader_args, **loader_kwargs).ingest()
+        return True
+    except Exception:  # pylint: disable=broad-except
+        logger.exception('%s failed!', loader_class.__name__)
+        return False
```
mikix: These `return True/False` bits are the real crux of this PR's fix - to report back whether an individual pipeline element succeeded. Rest is mostly refactoring.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -57,174 +52,141 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument(
             '--partner_code',
-            action='store',
-            dest='partner_code',
-            default=None,
```
mikix, line 62: These are all defaults anyway.

course_discovery/settings/production.py
```diff
@@ -99,3 +99,6 @@
 # Enable offline compression of CSS/JS
 COMPRESS_ENABLED = True
 COMPRESS_OFFLINE = True
+
+# Use bucket-default ACLs
+AWS_DEFAULT_ACL = None
```
mikix, line 104: Unrelated change - this fixes our django-storages code to use an AWS's bucket's default ACLs rather than django-storages's default ACL of 'public-read' - not setting this causes a warning to be printed out about that unsafe default (which they are planning to change in some major release down the road).

`None` seemed safer.  Though I don't think discovery uses S3 storage, unless sending emails uses it? Doubt it though.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -218,11 +218,9 @@ def format_course_run_data(self, body, course=None):
             'enrollment_start': self.parse_date(body['enrollment_start']),
             'enrollment_end': self.parse_date(body['enrollment_end']),
             'hidden': body.get('hidden', False),
+            'license': body.get('license', ''),
         }
 
-        # NOTE: The license field is non-nullable.
-        defaults['license'] = body.get('license') or ''
```
mikix, line 224: Minor unrelated cleanup.

course_discovery/apps/api/cache.py
```diff
@@ -49,7 +49,9 @@ def timestamped_object_key_constructor(*args, **kwargs):  # pylint: disable=unus
     return TimestampedObjectKeyConstructor()(**kwargs)
 
 
-def set_api_timestamp(timestamp):
+def set_api_timestamp(msg):
+    timestamp = time.time()
+    logger.debug('{msg} Updating API timestamp to {timestamp}'.format(msg=msg, timestamp=timestamp))
```
mikix, line 54: Minor unrelated refactor to simplify the two calls to this function.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -57,174 +52,145 @@ class Command(BaseCommand):
...
-        timestamp = time.time()
-        logger.info(
-            'Data loading complete. Updating API timestamp to {timestamp}.'.format(timestamp=timestamp)
-        )
+        set_api_timestamp('Data loading complete.')
 
-        set_api_timestamp(timestamp)
+        if not all(f[2].result() for f in futures):
+            loader_names = ['%s:%s' % (f[0].short_code, f[1].__name__) for f in futures]
+            raise CommandError('The following data loaders failed: %s' % ', '.join(loader_names))
```
mikix: And boom, here's the other half of the actual fix - reporting back when a pipeline fails.

course_discovery/apps/api/cache.py
Dillon-Dumesnil: Should this maybe be a `logger.info`?

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -189,27 +183,20 @@ def handle(self, *args, **options):
                                     max_workers,
                                     is_threadsafe,
                                     **kwargs,
-                                )
+                                ))
+
+                success = success and all(f.result() for f in futures)
```
Dillon-Dumesnil, line 187: Fancy Mike Terry

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -160,26 +162,18 @@ def handle(self, *args, **options):
             )
 
             if waffle.switch_is_active('parallel_refresh_pipeline'):
-                for stage_num, stage in enumerate(pipeline):
+                futures = []
+                for stage in pipeline:
```
Dillon-Dumesnil: Nit: Remove this empty line.

Dillon-Dumesnil (approved):
Nice job!

course_discovery/apps/api/cache.py
mikix: Of the two calls to it, one used to be `debug` (the post-save call) and one used to be `info` (RCM). I think both are extremely low value messages. I'm inclined to keep it as `debug` just in case upgrading to `info` makes our logs very noisy now that every post-save would print one.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
mikix: You got it.

course_discovery/apps/api/cache.py
Dillon-Dumesnil: Ah, didn't realize info was actually higher than debug. All good. It mostly just bugs me in tests because it is extremely noisy and I'm not sure how much use it serves.

PR #2075, opened 2019-08-12
Seed new course runs with data

ciduarte:
Instructors, weeks to complete, min & max effort,
content, & transcript languages will be populated
from the user selected re-run

DISCO-1301

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -232,6 +232,16 @@ class CourseRunViewSet(viewsets.ModelViewSet):
             # This will prevent a breaking change if users of this endpoint don't choose to provide a key on rerun
             old_course_run_key = course.canonical_course_run.key
 
+        if old_course_run_key:
+            old_course_run = CourseRun.objects.filter_drafts().get(key=old_course_run_key)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -232,6 +232,16 @@ def create_run_helper(self, run_data, request=None):
             # This will prevent a breaking change if users of this endpoint don't choose to provide a key on rerun
             old_course_run_key = course.canonical_course_run.key
 
+        if old_course_run_key:
+            old_course_run = CourseRun.objects.filter_drafts().get(key=old_course_run_key)
+            course_run.staff = old_course_run.staff.all()
+            course_run.transcript_languages = old_course_run.transcript_languages.all()
```
mikix: Both `staff` and `transcript_languages` are ManyToMany fields. So these lines will trigger the following Django warning:

```
RemovedInDjango20Warning: Direct assignment to the forward side of a many-to-many set is deprecated due to the implicit save() that happens. Use staff.set() instead.
```

So I'd suggest moving those after the save and use `.set()` to make what's happening clearer and to be ready for Django 2.0.

mikix (approved):

PR #2076, opened 2019-08-12
Stop loading replica db from env

jdmulloy:


mikix (approved):
Looks reasonable - no more env settings, just from the config file?

coryleeio:
@mikix Thats right, basically what is happening is the ENVs are getting out of sync when we update the YML and we noticed it because we changed your DB host in stage.  If you guys just stop reading from the environment vars it'll unblock you.

I've filed this for Lahore to remove the ENVs from all the apps next week
https://openedx.atlassian.net/browse/OPS-4272

PR #2077, opened 2019-08-13
Add new publisher roles API

mikix:
Specifically, an endpoint that lists OrganizationUserRole entries.

This lives in the old publisher djangoapp and under the old
publisher URL scheme - just for parity with similar APIs that we
already use.

https://openedx.atlassian.net/browse/DISCO-1269

### DIFF
--- a/course_discovery/apps/publisher/api/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/tests/test_views.py
@@ -21,7 +21,9 @@ from course_discovery.apps.course_metadata.tests.factories import CourseRunFacto
 from course_discovery.apps.course_metadata.tests.factories import OrganizationFactory, PersonFactory
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
 from course_discovery.apps.publisher.api import views
-from course_discovery.apps.publisher.choices import CourseRunStateChoices, CourseStateChoices, PublisherUserRole
+from course_discovery.apps.publisher.choices import (
+    CourseRunStateChoices, CourseStateChoices, InternalUserRole, PublisherUserRole
+)
 from course_discovery.apps.publisher.constants import ADMIN_GROUP_NAME, INTERNAL_USER_GROUP_NAME
 from course_discovery.apps.publisher.models import (
     Course, CourseRun, CourseRunState, CourseState, OrganizationExtension, Seat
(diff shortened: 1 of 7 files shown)


course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -214,6 +216,85 @@ def test_get_organization_by_uuid(self):
         self.assertExpectedUsers(response)
 
 
+class OrganizationUserRoleViewTests(APITestCase):
+
```
Dillon-Dumesnil: Why you do this?

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -214,6 +216,85 @@ def test_get_organization_by_uuid(self):
...
+    def query(self, org_id=None, param=None):
+        if org_id is None:
+            org_id = self.organization.id
+        url = reverse('publisher:api:organization_user_roles', kwargs={'pk': org_id})
+        if param:
+            url += '?' + param
+        return self.client.get(path=url, content_type=JSON_CONTENT_TYPE)
+
+    def assertExpectedRoles(self, response, roles=None):
```
Dillon-Dumesnil, line 241: Nice

course_discovery/apps/publisher/api/tests/test_views.py
```diff
@@ -214,6 +216,85 @@ def test_get_organization_by_uuid(self):
...
+    def test_happy_path(self):
+        """
+        Verify that view returns list of roles associated with the given organization id.
+        """
+        response = self.query()
+        self.assertExpectedRoles(response)
+
+    def test_num_queries(self):
+        view = views.OrganizationUserRoleView(kwargs={'pk': str(self.organization.id)})
+        with self.assertNumQueries(1, threshold=0):  # one for Org Ext, one for users
```
Dillon-Dumesnil: Your comment leads me to believe the number of queries would be 2. Can we update it?

Dillon-Dumesnil (approved):
Couple of small nits. Are we filtering to get only the PCs on the frontend then since this view returns all the org user roles?

Looks good

course_discovery/apps/publisher/api/tests/test_views.py
mikix: Whoops, will do.

mikix:
> Are we filtering to get only the PCs on the frontend then since this view returns all the org user roles?

Yeah exactly.

PR #2079, opened 2019-08-13
Add new detail_url_slug field to course model

rgraber:


course_discovery/apps/api/serializers.py
```diff
@@ -876,7 +876,8 @@ def prefetch_queryset(cls, queryset=None, course_runs=None):
 
     class Meta:
         model = Course
-        fields = ('key', 'uuid', 'title', 'course_runs', 'entitlements', 'owners', 'image', 'short_description',)
+        fields = ('key', 'uuid', 'title', 'course_runs', 'entitlements', 'owners', 'image',
+                  'short_description', 'detail_url_slug', )
```
dianakhuang: Do we need to update any tests with this new field?

course_discovery/apps/course_metadata/models.py
```diff
@@ -544,7 +548,7 @@ class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
 
     # Do not record the slug field in the history table because AutoSlugField is not compatible with
     # django-simple-history.  Background: https://github.com/edx/course-discovery/pull/332
-    history = HistoricalRecords(excluded_fields=['slug'])
+    history = HistoricalRecords(excluded_fields=['slug', 'detail_url_slug'])
```
dianakhuang: Good to know.

course_discovery/apps/course_metadata/models.py
```diff
@@ -503,6 +503,10 @@ class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
     )
     key = models.CharField(max_length=255, db_index=True)
     title = models.CharField(max_length=255, default=None, null=True, blank=True)
+    detail_url_slug = AutoSlugField(populate_from='title', editable=True, slugify_function=uslugify,
```
mikix: Minor nit: `detail_url_slug` is very targetted for one of the possible purposes. Like, Prospectus is presumably using it for its course page URL? But feels odd for Discovery to be so specific about that. Do you think `title_slug` is too generic? Or `url_slug` at least is a little less specific... I dunno.

course_discovery/apps/course_metadata/models.py
```diff
@@ -503,6 +503,10 @@ class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
...
+                                    overwrite_on_add=False,
+                                    help_text=_('Leave this field blank to have the value generated automatically. '
+                                                + 'This field is currently unused.'))
```
mikix: I'd vote to leave the "Currently unused" bit out of a string where if we change it, translators need to retranslate and we need to generate a new migration for.

mikix (approved):

course_discovery/apps/course_metadata/models.py
rgraber: Does anyone outside of edX see admin? If only edX people see it it's fine but we were worried about confusing partners

course_discovery/apps/course_metadata/models.py
rgraber: point. I think url_slug makes sense. I don't love title_slug since it might be overridden by the user

PR #2082, opened 2019-08-15
Adds in SF Utils and Comments API

jlajoie:


course_discovery/apps/api/serializers.py
```diff
@@ -187,6 +187,10 @@ class TimestampModelSerializer(serializers.ModelSerializer):
     """Serializer for timestamped models."""
     modified = serializers.DateTimeField(required=False)
 
+class CommentSerializer(serializers.Serializer):
+    """Serializer for retrieving comments from Salesforce"""
+    pass
```
mikix, line 196: Add a comment about this?

course_discovery/apps/core/models.py
```diff
@@ -196,5 +196,11 @@ class SalesforceConfiguration(models.Model):
         verbose_name=_('Is a Salesforce Sandbox?'),
         default=True
     )
+    case_record_type_id = models.CharField(
```
mikix, line 199: It's a string, not a number? I guess even if it were a number today, we might not want to commit to that on our side...

course_discovery/apps/core/models.py
```diff
@@ -196,5 +196,11 @@ class SalesforceConfiguration(models.Model):
...
+        max_length=255,
+        blank=True,
+        verbose_name=_('Case Record Type Id'),
+        null=True,
```
mikix, line 203: Usually, Django recommends blank=True, null=False for CharFields. Allowing null is fine too, but just checking. Same comment on the other new fields too.

course_discovery/apps/course_metadata/models.py
```diff
@@ -337,6 +337,7 @@ class Organization(TimeStampedModel):
         null=True, blank=True, help_text=_('Logo to be displayed on certificates. If this logo is the same as '
                                            'logo_image_url, copy and paste the same value to both fields.')
     )
+    salesforce_id = models.CharField(max_length=255, null=True, blank=True)  # Publisher_Organization__c in Salesforce
```
mikix, line 340: Do we know if these 255 max_lengths are accurate to what Salesforce's ID max length is?

course_discovery/apps/api/v1/views/comments.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+    def list(self, request, *args, **kwargs):
+        partner = request.site.partner
+        course_uuid = request.query_params.get('course_uuid')
+        if not course_uuid:
+            return Response(
+                _('You must include a course_uuid in your query parameters.'),
+                status=status.HTTP_400_BAD_REQUEST
+            )
```
mikix, line 36: Maybe have the UUID as part of the URL then? Like `/comments/<uuid>/`

course_discovery/apps/api/v1/views/comments.py
```diff
@@ -0,0 +1,74 @@
...
+        course = get_object_or_404(Course, uuid=course_uuid)
```
mikix: Add a `partner=partner` too - that's in the unique_together with `uuid`.

And... `get_object_or_404` probably only calls `Course.objects.filter()` - which won't see draft rows. Might want to find  a way to use `Course.objects.filter_drafts()` or something.

course_discovery/apps/api/v1/views/comments.py
```diff
@@ -0,0 +1,74 @@
...
+        error_message = ''
+        if missing_values:
+            error_message += ''.join([_('Missing value for: [{name}]. ').format(name=name) for name in missing_values])
+        if error_message:
+            return Response((_('Incorrect data sent. ') + error_message).strip(), status=status.HTTP_400_BAD_REQUEST)
+
+        partner = self.request.site.partner
+        util = SalesforceUtil(partner)
+
+        course = get_object_or_404(Course, uuid=comment_creation_fields.get('course_uuid'))
```
mikix: Same `partner=partner` thing, and draft considerations here.

mikix (commented):
I didn't look at `salesforce.py` or `test_salesforce.py` very closely. Too much refactoring and salesforce-specific query logic for my tired brain.

course_discovery/apps/core/models.py
jlajoie: Yeah it's a string from SF, the first 3 characters resolve to the object type under the hood, the rest is the unique identifier, example is ```50063000006bLHrAAM```

course_discovery/apps/course_metadata/models.py
jlajoie: SF's is lower but I'm not sure on specifics of max length, most I've seen are 18.

course_discovery/apps/api/v1/views/comments.py
jlajoie: I thought about this, but it felt _weird_ to me since the opposite to me would be a GET /comments/<uuid> which seems like it would then be referencing a comment's UUID itself, which we don't have? Strong preferences here?

course_discovery/apps/api/v1/views/comments.py
jlajoie: 👍

course_discovery/apps/api/v1/views/comments.py
jlajoie: 👍

course_discovery/apps/api/v1/views/comments.py
mikix: `/comments-for-course/`? No real preference myself. Leaving as is, is good.

mikix (approved):
Once the partner=partner / draft stuff is fixed, looks good!

PR #2084, opened 2019-08-15
Add default to api request

rgraber:


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -171,6 +171,7 @@ def create(self, request, *args, **kwargs):
             'number': request.data.get('number'),
             'org': request.data.get('org'),
             'mode': request.data.get('mode'),
+            'url_slug': request.data.get('url_slug') or '',
```
dianakhuang: You should be able to write this as `request.data.get('url_slug', '')`. Also, are we okay with this returning a blank string instead of `None`?

rgraber: For reasons I don't quite understand, AutoSlugField will correctly overwrite '' with the slugified title but balks when it's None

dianakhuang (approved):

PR #2086, opened 2019-08-15
Clean up SiteMixin, other test setup

iloveagent57:
* Modify SiteMixin (in tests) to create test site in setUpClass(), rename the domain to just "testserver" (there was a failing test somewhere due to the `.fake` getting trimmed in some URL serializer).
* trim more test setup from test_models.py
* Change `make static` target to run `./manage.py compress` with verbosity 0.

### DIFF
--- a/course_discovery/apps/api/tests/mixins.py
+++ b/course_discovery/apps/api/tests/mixins.py
@@ -2,17 +2,27 @@ from django.conf import settings
 from django.contrib.sites.models import Site
 from django.test import RequestFactory
 
+from conftest import TEST_DOMAIN
 from course_discovery.apps.core.tests.factories import PartnerFactory, SiteFactory
 
 
 class SiteMixin(object):
-    def setUp(self):
-        super(SiteMixin, self).setUp()
-        domain = 'testserver.fake'
-        self.client = self.client_class(SERVER_NAME=domain)
-        Site.objects.all().delete()
-        self.site = SiteFactory(id=settings.SITE_ID, domain=domain)
-        self.partner = PartnerFactory(site=self.site)
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
 
-        self.request = RequestFactory(SERVER_NAME=self.site.domain).get('')
-        self.request.site = self.site
+        Site.objects.all().delete()
+        cls.site = SiteFactory(id=settings.SITE_ID, domain=TEST_DOMAIN)
+        cls.partner = PartnerFactory(site=cls.site)
+
+        cls.request = RequestFactory(SERVER_NAME=cls.site.domain).get('')
+        cls.request.site = cls.site
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        cls.partner.delete()
+
+    def setUp(self):
+        super().setUp()
+        self.client = self.client_class(SERVER_NAME=TEST_DOMAIN)
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -41,6 +41,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         self.refresh_index()
         self.request = APIRequestFactory().get('/')
         self.request.user = self.user
+        self.partner.save()
 
     def mock_patch_to_studio(self, key, access_token=True, status=200):
         if access_token:
@@ -404,6 +405,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
 
     def test_partial_update_no_studio_url(self):
         """ Verify we skip pushing when no studio url is set. """
+        orignal_partner_studio_url = self.partner.studio_url
         self.partner.studio_url = None
         self.partner.save()
 
@@ -425,6 +427,10 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
             self.draft_course_run.key,
             self.partner.short_code,
         ))
+
+        # reset the shared self.partner attribute
+        self.partner.studio_url = orignal_partner_studio_url
+        self.partner.save()
 
     def test_partial_update_bad_permission(self):
         """ Verify partially updating will fail if user doesn't have permission. """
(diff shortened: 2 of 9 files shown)


iloveagent57:
FYI @edx/masters-neem

course_discovery/apps/api/tests/mixins.py
```diff
@@ -6,13 +6,20 @@
 
 
 class SiteMixin(object):
-    def setUp(self):
-        super(SiteMixin, self).setUp()
-        domain = 'testserver.fake'
```
mikix, line 11: The failing test with a serializer trimming this `.fake` - that's not a bug that we're glad we caught? But rather, just some test weirdness? A domain name of just `testserver` feels less realistic than `testserver.fake`.

course_discovery/apps/api/tests/mixins.py
```diff
@@ -6,13 +6,20 @@
...
+
+        cls.request = RequestFactory(SERVER_NAME=cls.site.domain).get('')
+        cls.request.site = cls.site
 
-        self.request = RequestFactory(SERVER_NAME=self.site.domain).get('')
-        self.request.site = self.site
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        cls.partner.delete()
```
mikix, line 24: Do we need that? Doesn't Django reset the db between tests?

course_discovery/apps/api/v1/tests/test_views/mixins.py
```diff
@@ -217,4 +217,4 @@ def assertNumQueries(self, num, func=None, *args, **kwargs):
         may slightly modify the query count without having to update expected counts in tests,
         while still ensuring that we don't inflate the number of queries by an order of magnitude.
         """
-        return super().assertNumQueries(FuzzyInt(num, kwargs.pop('threshold', 2)), func=func, *args, **kwargs)
+        return super().assertNumQueries(FuzzyInt(num, kwargs.pop('threshold', 5)), func=func, *args, **kwargs)
```
mikix: Whoa that's a big change - lots of tests call this without an argument. Was this necessary to make a test less flaky or you just aren't a big fan of a tighter threshold?

course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
```diff
@@ -24,7 +24,7 @@ class AffiliateWindowViewSetTests(ElasticsearchTestMixin, SerializationMixin, AP
     def setUp(self):
         super(AffiliateWindowViewSetTests, self).setUp()
         self.user = UserFactory()
-        self.client.force_authenticate(self.user)
+        self.client.force_authenticate(self.user)  # pylint: disable=no-member
```
mikix: Is this stuff because of the self.client being at the class level?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -28,6 +28,11 @@
 
 @ddt.ddt
 class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin, APITestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls._orignal_partner_studio_url = cls.partner.studio_url
```
mikix: What's this trick for?

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -35,7 +35,7 @@ def get_response(self, query=None, path=None):
         return self.client.get(url)
 
     def build_facet_url(self, params):
-        return 'http://testserver.fake{path}?{query}'.format(
+        return 'http://testserver{path}?{query}'.format(
```
mikix: Could maybe just replace that with `self.domain` or something.

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -14,7 +14,7 @@
 
 
 @pytest.mark.django_db
-class TestAutocomplete:
+class TestAutocomplete(object):
```
mikix: You don't need to specify that in Python3 - all classes are new-style classes, whether you specify `object` or not.

course_discovery/apps/api/tests/mixins.py
iloveagent57: LOL, that's a good point, I'll double check that it behaves correctly in normal circumstances.

course_discovery/apps/api/tests/mixins.py
iloveagent57: My understanding is that the DB is destroyed after all the tests are run, but not between runs of individual test functions or classes. 
 https://docs.djangoproject.com/en/1.11/topics/testing/overview/#the-test-database

course_discovery/apps/api/v1/tests/test_views/mixins.py
iloveagent57: This was necessary to make a handful of tests less flaky - changing the `SiteMixin` class to create the mock site in `setUpClass` changed something about how the site and partner were read, somehow.  I can diff the set of queries to get a more precise explanation.

course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
iloveagent57: yeah

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
iloveagent57: `APITestCase` inherites from `SiteMixin`, which creates a `partner` test fixture in `setUpClass`.  There are a couple of tests below that change an attribute of `self.partner` - this keeps track of the partner's studio_url attribute as it was during initial creation, and then `setUp` below makes sure to reset it to that original value in between test method runs.

course_discovery/apps/api/v1/tests/test_views/test_search.py
iloveagent57: 👍

course_discovery/apps/course_metadata/tests/test_lookups.py
iloveagent57: I will never internalize this :)

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
iloveagent57: ...but now that I've described all of that, maybe I'll just create a different partner fixture within the tests that need to change a partner's attribute.

course_discovery/apps/api/tests/mixins.py
iloveagent57: I tried changing this to have the client created in `setUp`, so that it's re-created before each test method, and that doesn't show any issues when setting the domain to `testserver.fake`, so I think a test somewhere was over-writing the request's servername attribute before this test.  So I'll just create the client in `setUp`.

zacharis278 (approved):
this lgtm  🏃

course_discovery/apps/api/v1/tests/test_views/mixins.py
iloveagent57: somehow moving the client initialization back into `setUp` fixed this for all but one test, so I changed the threshold here back to 2.

PR #2087, opened 2019-08-16
DISCO-1307 status change exempt

xnick421:
If course run is reviewed, only revert back to legal review if fields that are changed are not listed in STATUS_CHANGE_EXEMPT_FIELDS

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -262,13 +262,16 @@ class CourseRunViewSet(viewsets.ModelViewSet):
     def _update_course_run(self, course_run, draft, changed, serializer, request, save_kwargs):
         # If changes are made after review and before publish, revert status to unpublished.
         # Unless we're just switching the status
-        if changed and course_run.status == CourseRunStatus.Reviewed:
+        non_exempt_update = changed and course_run.status == CourseRunStatus.Reviewed
+        if non_exempt_update:
             save_kwargs['status'] = CourseRunStatus.Unpublished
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -268,7 +268,8 @@ def _update_course_run(self, course_run, draft, changed, serializer, request, sa
             official_run.status = CourseRunStatus.Unpublished
             official_run.save()
 
-        if not draft and course_run.status != CourseRunStatus.Published:
+        if not draft and (course_run.status == CourseRunStatus.Unpublished or
+                          (changed and course_run.status == CourseRunStatus.Reviewed)):
```
mikix: A) This is a weird line / block. Could you maybe add a comment about what's actually happening while you're here? Like "user submitted for review, so we set status to the first review step" or something
B) Shouldn't this logic just be `if not draft and changed and course_run.status != CourseRunStatus.Published`? Why this extra complication?
C) But if we do need this complication, this has some duplicate `if` logic from the `if` block above. That could be consolidated into a temporary variable or something to avoid future mistakes.

mikix: Oh I see about the extra complication. OK, scratch B.

mikix: Maybe something for the comment in A though - easy to not follow why the `or` exists here.

xnick421: sounds good

PR #2090, opened 2019-08-16
Record slug history

rgraber:


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -231,7 +231,7 @@ def test_list_query(self):
         query = 'title:' + title
         url = '{root}?q={query}'.format(root=reverse('api:v1:course-list'), query=query)
 
-        with self.assertNumQueries(55, threshold=3):
+        with self.assertNumQueries(60, threshold=3):
```
mikix, line 234: The bumps in this file are 4, 5, and 6 big. I would have naively expected them to be the same.

Also, prefetching doesn't reduce that more? I thought prefetching got you some related data "for free" because it would grab it in the same SQL query. But maybe it just does the same amount of queries but up front and cached?

course_discovery/apps/course_metadata/models.py
```diff
@@ -721,6 +723,19 @@ def _update_or_create_official_version(self, course_run):
         return official_version
 
 
+@receiver(pre_save, sender=Course)
```
mikix: I'd say put signals in `signals.py`. While it does mean that some logic is in `models.py` and some in `signals.py`, it's the Django way, and we already have some logic there.

course_discovery/apps/course_metadata/models.py
```diff
@@ -721,6 +723,19 @@ def _update_or_create_official_version(self, course_run):
...
+def add_to_slug_history(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    if not instance.draft:
```
mikix: Extremely minor nit, but when doing a conditional that is either `do the thing or not`, I prefer an early-exit style of:
```
if instance.draft:
    return
```

Just to keep the method body focused on the thing you are actually trying to do, to keep indentation low, and so you can more naturally comment each early exit condition. But super doesn't matter in this case, just saying.

course_discovery/apps/course_metadata/models.py
```diff
@@ -721,6 +723,19 @@ def _update_or_create_official_version(self, course_run):
...
+        try:
+            obj = sender.objects.get(pk=instance.pk)
+        except sender.DoesNotExist:
+            pass
+        else:
+            if not obj.url_slug == instance.url_slug:
```
mikix: This feels like it would more naturally be `!=`

course_discovery/apps/course_metadata/models.py
```diff
@@ -721,6 +723,19 @@ def _update_or_create_official_version(self, course_run):
         return official_version
 
 
+@receiver(pre_save, sender=Course)
+def add_to_slug_history(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    if not instance.draft:
+        try:
+            obj = sender.objects.get(pk=instance.pk)
```
mikix: I really wish Django provided a "change set" feature for this kind of thing.

course_discovery/apps/course_metadata/models.py
```diff
@@ -721,6 +723,19 @@ def _update_or_create_official_version(self, course_run):
...
+        except sender.DoesNotExist:
+            pass
+        else:
+            if not obj.url_slug == instance.url_slug:
+                defaults = {'value': obj.url_slug}
+                CourseHistoricalSlug.objects.create(course=instance, **defaults)
```
mikix: `**defaults` isn't buying you anything here, because it's not an update_or_create call. I'd say just use kwarg params directly.

Also, this method will create double entries if the user changes back to a previous URL slug. Is that desired?

mikix (commented):
Out of curiousity, are we going to backpopulate this model from drupal data?

I'm assuming this is all to fuel Prospectus redirects. Do we have a plan for multiple URLs being claimed by different courses over time? Like course A used to have "cs50" slug, but doens't anymore. And now course B has it. It'll use timestamps for that, so that the newest slug wins?

(Do we also have any plan for course teams fighting over slugs? LIke if one partner wants `intro-to-computer-science`, is conflict resolution just done through PCs? -- this comment is more appropriate for the frontend PR, but here we are.)

course_discovery/apps/course_metadata/models.py
mikix: Also... pre-save has some problems. What if an error occurs during save? You'd have already recorded the new slug. Or what if a user clears out the slug? And it's going to be assigned during the `save` call? You'd save an empty string and not save the computed slug, right? Would post-save work?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -178,6 +178,7 @@ def get_expected_data(cls, course, request):
             'enrollment_count': 0,
             'recent_enrollment_count': 0,
             'topics': list(course.topics.names()),
+            'url_slug_history': [],
```
mikix, line 181: Should probably add a test that actually does have the `get_url_slug_history` method do something interesting.

course_discovery/apps/api/serializers.py
```diff
@@ -904,6 +904,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     original_image = ImageField(read_only=True, source='original_image_url')
     extra_description = AdditionalPromoAreaSerializer(required=False)
     topics = TagListSerializerField(required=False)
+    url_slug_history = serializers.SerializerMethodField()
```
mikix, line 911: Can you not do something with SlugRelatedField here?

PR #2093, opened 2019-08-19
DISCO-1317: Always write to draft when publishing

Dillon-Dumesnil:
We want to always write to the draft row and if the publish is coming from
old publisher, we want to also write to the official row. This will help
with data issues we were running into where the draft and official rows
were not in sync.

### DIFF
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -471,7 +471,7 @@ def push_to_ecommerce_for_course_run(course_run):
 
 
 @transaction.atomic
-def publish_to_course_metadata(partner, course_run, draft=False):
+def publish_to_course_metadata(partner, course_run, create_official=True):
     from course_discovery.apps.course_metadata.models import (
         Course, CourseEntitlement, CourseRun, ProgramType, Seat, SeatType, Video
     )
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/utils.py
```diff
@@ -471,7 +471,7 @@ def push_to_ecommerce_for_course_run(course_run):
 
 
 @transaction.atomic
-def publish_to_course_metadata(partner, course_run, draft=False):
+def publish_to_course_metadata(partner, course_run, from_publisher=True):
```
mikix: Nit: I would argue that `create_official` or something would be nicer. `from_publisher` means that CM code has to know what publisher needs. Leaky boundaries and all.

Although I guess below, you already have leaky boundaries with handling the image code. That's not our fault, but I think it's badly written. If old publisher wants to require that field, it should do it explictly, instead of waiting for an `image.name` to raise an AttributeError...

course_discovery/apps/course_metadata/utils.py
```diff
@@ -586,6 +583,11 @@ def publish_to_course_metadata(partner, course_run, draft=False):
         discovery_course.canonical_course_run = discovery_course_run
         discovery_course.save()
 
+    if from_publisher:
+        # This will update or create the official version for the Course,
+        # Course Run, Seats, and Entitlements
+        discovery_course_run.update_or_create_official_version()
```
mikix: This does a push to ecom. So either we can remove that from old-publisher or add a flag to this method to not do that.

mikix (commented):
I think this makes sense?

mikix (approved):
Once the ecom thing is fixed, approved!  Thanks for this work / investigation!

PR #2094, opened 2019-08-19
Refactor how we call delete_orphans

mikix:
Call it just once per RCM run, rather than once per loader.

https://openedx.atlassian.net/browse/DISCO-1223

### DIFF
--- a/course_discovery/apps/core/tests/test_utils.py
+++ b/course_discovery/apps/core/tests/test_utils.py
@@ -2,9 +2,9 @@ from django.db import models
 from django.test import TestCase
 from haystack.query import SearchQuerySet
 
-from course_discovery.apps.core.utils import SearchQuerySetWrapper, get_all_related_field_names
-from course_discovery.apps.course_metadata.models import CourseRun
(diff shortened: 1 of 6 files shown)


course_discovery/apps/core/tests/test_utils.py
```diff
@@ -41,6 +41,16 @@ def test_get_all_related_field_names(self):
         self.assertEqual(get_all_related_field_names(UnrelatedModel), [])
         self.assertEqual(set(get_all_related_field_names(RelatedModel)), {'foreignrelatedmodel', 'm2mrelatedmodel'})
 
+    def test_delete_orphans(self):
```
jlajoie, line 44: We happy we're not testing deletion of orphaned Images? 🙂

jlajoie (approved):
👍 lgtm

course_discovery/apps/core/tests/test_utils.py
mikix: Eh, I figured one model is the same as another, from core's point of view. Thanks for the review!

PR #2096, opened 2019-08-20
Don't assume nodes exist in Drupal

mikix:
In a couple places (publishing programs, deleting nodes), we are
now less confident that Drupal always has the node like we expect.
This will avoid a few 500 errors when asking Drupal to delete or
edit a node that doesn't actually exist.

https://openedx.atlassian.net/browse/PROD-560


Dillon-Dumesnil (approved):
LGTM

jlajoie (approved):
👍

PR #2098, opened 2019-08-20
Avoid assuming runs have end dates

mikix:
When migrating from old publisher to course metadata, we were
trying to migrate some pre-publisher runs that didn't have end
dates, like all modern runs do. Handle those gracefully.


Dillon-Dumesnil (approved):
Looks good.

PR #2100, opened 2019-08-21
DISCO-1039: Adds in bulk load salesforce comments command

jlajoie:


course_discovery/apps/api/v1/views/comments.py
```diff
@@ -8,6 +8,7 @@
 from rest_framework.response import Response
 
 from course_discovery.apps.api.serializers import CommentSerializer
+from course_discovery.apps.core.models import SalesforceConfiguration  # Comments
```
jlajoie: Remove the `# Comments`... dunno what I was doing there

course_discovery/apps/api/v1/views/comments.py
```diff
@@ -28,6 +29,11 @@ def get_queryset(self):
 
     def list(self, request):
         partner = request.site.partner
+        try:
```
jlajoie, line 32: Unrelated but by adding this in reference a partner that does not have a SF config does not return a 500, and instead shows that this endpoint isn't implemented via 404

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -167,15 +167,15 @@ def create_case_for_course(self, course):
...
+        user_comment_body = self.format_user_comment_body(user, body, course_run_key=course_run_key)
         self.client.FeedItem.create({
             'ParentId': course.salesforce_case_id,
             'Body': user_comment_body,
         })
         return self._create_comment_return_body(user, body, course_run_key)
 
     @staticmethod
-    def _format_user_comment_body(user, body, course_run_key=None):
+    def format_user_comment_body(user, body, course_run_key=None):
```
jlajoie, line 179: Moving to public so it's accessible in the management command

PR #2101, opened 2019-08-21
Prevent circular program structure

zacharis278:
@edx/masters-neem 

### [EDUCATOR-4203](https://openedx.atlassian.net/browse/EDUCATOR-4203)


course_discovery/apps/course_metadata/signals.py
```diff
@@ -96,6 +96,56 @@ def publish_masters_track(sender, instance, **kwargs):  # pylint: disable=unused
     _create_masters_track_on_lms_if_necessary(seat, partner)
 
 
+@receiver(pre_save, sender=Curriculum)
+def save_curriculum(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
jansenk: Since we have now multiple pre-save signal handlers for Curriculum, I think you should change this method name. Probably a good idea to do change the other one to something more specific as well

course_discovery/apps/course_metadata/signals.py
```diff
@@ -96,6 +96,56 @@ def publish_masters_track(sender, instance, **kwargs):  # pylint: disable=unused
...
+    program = instance.program
+    if _find_in_programs([program], curriculum=curriculum):
+        msg = 'Circular ref error. Program [{}] already contains Curriculum [{}]'.format(
+            program,
+            curriculum,
+        )
+        raise ValidationError(msg)
+
+
+def _find_in_programs(programs, curriculum=None, program=None):
```
jansenk: could you make the argument names a but more descriptive / add to the docstring?

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -278,6 +278,120 @@ def test_course_curriculum_membership_side_effect_not_masters(self):
...
 
+@ddt.ddt
+class ProgramStructureValidationTests(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        """
+        Set up program structure to test for cycles
+
+        program           p1
```
jansenk, line 289:
```
  _   _ _             _____ _                _   
 | \ | (_)           / ____| |              | |  
 |  \| |_  ___ ___  | |    | |__   __ _ _ __| |_ 
 | . ` | |/ __/ _ \ | |    | '_ \ / _` | '__| __|
 | |\  | | (_|  __/ | |____| | | | (_| | |  | |_ 
 |_| \_|_|\___\___|  \_____|_| |_|\__,_|_|   \__|
                                                 
```

course_discovery/apps/course_metadata/signals.py
```diff
@@ -96,6 +96,56 @@ def publish_masters_track(sender, instance, **kwargs):  # pylint: disable=unused
...
+        raise TypeError('_find_in_programs takes at least one of (curriculum, program)')
+
+    curricula = Curriculum.objects.filter(program__in=programs).prefetch_related('program_curriculum')
+
+    if curriculum in curricula or program in programs:
+        return True
+    if not programs:
+        return False
+
+    child_programs = [program for curriculum in curricula for program in curriculum.program_curriculum.all()]
```
jansenk, line 147: i don't ever think i've seen this syntax before and it's kinda melting my brain

jansenk (approved):
Just some small suggestions but overall looks great

course_discovery/apps/course_metadata/signals.py
```diff
@@ -96,6 +96,56 @@ def publish_masters_track(sender, instance, **kwargs):  # pylint: disable=unused
...
+
+def _find_in_programs(programs, curriculum=None, program=None):
+    """
+    Travese the stucture of a given list of programs for a curriculm or program node.
+    Returns True if an instance is found
+    """
+    if curriculum is None and program is None:
+        raise TypeError('_find_in_programs takes at least one of (curriculum, program)')
+
+    curricula = Curriculum.objects.filter(program__in=programs).prefetch_related('program_curriculum')
```
iloveagent57: I think you can get this to return sooner as follows:
```python
if not programs:
    return False
if program in programs:
    return True

curricula = Curriculum.objects.filter(program__in=programs).prefetch_related('program_curriculum')
if curriculum in curricula:
    return True
...
```

course_discovery/apps/course_metadata/signals.py
iloveagent57, line 147: You're overloading the `curriculum` variable here.

course_discovery/apps/course_metadata/signals.py
iloveagent57: I think this will read easier if you rename the parameters to `existing_programs`, `target_curriculum`, and `target_program`, respectively.

course_discovery/apps/course_metadata/signals.py
iloveagent57: ...which I think is actually a material change.  I'd try a different variable name and see if your tests break before proceeding.

iloveagent57 (dismissed):

course_discovery/apps/course_metadata/signals.py
kdmccormick: In Py2 yes, but in Py3 it won't leak out of the comprehension. But I agree that avoiding overloading is better anyway.

course_discovery/apps/course_metadata/signals.py
zacharis278: yeah I should probably just name this based on what it actually does instead

course_discovery/apps/course_metadata/signals.py
zacharis278: Thanks, TIL.  Changed the function parameters following Alex's suggestion below so that fixes the overloading issue.

kdmccormick:
Nice code and tests! They're easy to follow. 

`_find_in_program` will infinitely recurse if the program is already cyclic, in a way that `target_curriculum` and `target_program` are never reached. For example:

```
TgtCurr  
   |
  PgmX <- _find_in_program
   |
 CurrX
   |
  PgmX
   |
 CurrX
   |
  ...
```

I don't know whether that situation is worth worrying about, though, because I'm not sure if it'll ever come up. I just like thinking about graphs I guess.

jansenk:
> I don't know whether that situation is worth worrying about, though, because I'm not sure if it'll ever come up. I just like thinking about graphs I guess.

I don't believe we could ever get to that state, unless a cyclic structure already exists before this code is put in and then we add TgtCurr to the chain

zacharis278:
@kdmccormick yeah I think we'd have to patch that out separately if one exists.  However, the /programs endpoint would also be infinitely recursing if such a structure existed so I think someone would have noticed the mistake 🤞

zacharis278:
Instead of crossing fingers I just checked prod since we only have 4 programs with CurriculumCourseMembership.  None of them have cycles

PR #2102, opened 2019-08-21
Drop uses of CourseEntitlement.expires

mikix:
This is a never-used deprecated field. Step 1 of dropping the column
is to drop references. So that's what's happening here.

We keep serializing the field in our API as None though.

https://openedx.atlassian.net/browse/DISCO-820

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1529,6 +1529,8 @@ class CourseEntitlement(DraftModelMixin, TimeStampedModel):
     price = models.DecimalField(**PRICE_FIELD_CONFIG)
     currency = models.ForeignKey(Currency, models.CASCADE, default='USD')
     sku = models.CharField(max_length=128, null=True, blank=True)
+
+    # TODO: this field is deprecated and scheduled for removal
     expires = models.DateTimeField(null=True, blank=True)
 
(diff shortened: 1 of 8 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1529,6 +1529,8 @@ class CourseEntitlement(DraftModelMixin, TimeStampedModel):
     price = models.DecimalField(**PRICE_FIELD_CONFIG)
     currency = models.ForeignKey(Currency, models.CASCADE, default='USD')
     sku = models.CharField(max_length=128, null=True, blank=True)
+
+    # TODO: this field is deprecated and scheduled for removal
```
Dillon-Dumesnil, line 1533: Do we have a ticket we should link to here or is this going to be short lived enough that it doesn't matter? Not sure about timeline here.

Dillon-Dumesnil (approved):
:chef_kiss:

course_discovery/apps/course_metadata/models.py
mikix: Short lived. I'll propose that PR shortly.

mikix: Comment is just I dunno, in case I die before getting to it or whatever.

PR #2106, opened 2019-08-21
microb: Add MicroBachelors program_type

reillz10:


course_discovery/apps/course_metadata/utils.py
```diff
@@ -663,3 +663,28 @@ def headers(self):
             'Content-Type': 'application/json',
             'X-CSRF-Token': self.csrf_token,
         }
+
+
+def add_specific_program_types(seat_types, program_types):
```
bbaker6225: Nit: Might be nice to add a short comment to each of these utils just clarifying the context they are used in

bbaker6225 (approved):
Just a non-blocking comment about comments

course_discovery/apps/course_metadata/migrations/0187_add_microbachelors_program_type.py
```diff
@@ -0,0 +1,27 @@
...
+PROGRAM_TYPES = ('MicroBachelors',)
+
+add_program_types = add_specific_program_types(SEAT_TYPES, PROGRAM_TYPES)
+drop_program_types = drop_specific_program_types(PROGRAM_TYPES)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0186_update_migrate_publisher_command'),
```
mikix: This file's name and dependency are out of date - we're on 0190 now.

course_discovery/apps/course_metadata/utils.py
mikix: This does more of a `set` than an `add`, right? It clears out the current list first.

Oh, I guess it's adding the program type, but setting the seat types.

mikix (approved):
The way these utility methods are written is very migration specific (function generator to a migration runpython method prototype, etc). You didn't want to just put them in the migration file? Or maybe a migration_utils.py file or something? Feels weird to have them just hanging with rest of code.

bbaker6225:
@mikix Agree on the utils being very migration specific. The only thing I would add is that this pattern has been used a few times now, so I do think it makes sense to put it in some sort of utility file so that we don't keep rewriting it into every program type migration

mikix:
Well, copy and pasting every time.

I would argue migration files get more of a pass on DRY than normal code:

(A) it's not really code we have to maintain, so duplicated code doesn't hurt.

(B) It's not code we *want* to be in the business of maintaining, because changing the behavior of a function used in migration 0005 when you're writing migration 0030 means that you might be affecting how 0005 gets forward-applied, which is rarely a good thing.

I think that there's a benefit to keeping each migration file its own little trapped-in-amber code pile. But it's a matter of taste.

bbaker6225:
@mikix Yeah, good points. I agree with you now

reillz10:
@mikix & @bbaker6225 

Yes the thought was to move it to a util function so that when we have new program types in the future it will be less duplicated code. That said, moving it to the util function actually added a small amount of complexity because I had to use a closure to pass the program type to the resulting function that we want anyway. 

I don't have a strong preference. If the consensus is to move that code back into the migrations file, I'll do it!

reillz10:
@albemarle FYI -> I'm going to move the functions back to the migrations file as per the feedback above. Also going to correct the migration number.

iloveagent57:
This looks good as long as you correct the parent migration number.

PR #2110, opened 2019-08-23
Add Course Length to Affiliate API

zainab-amir:
Add course length to affiliate window api to target appropriate audience

[LEARNER-1144](https://openedx.atlassian.net/browse/LEARNER-1144)

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1551,6 +1551,7 @@ class AffiliateWindowSerializer(serializers.ModelSerializer):
     custom3 = serializers.SerializerMethodField()
     custom4 = serializers.SerializerMethodField()
     custom5 = serializers.CharField(source='course_run.short_description')
+    custom6 = serializers.SerializerMethodField()
 
     class Meta:
         model = Seat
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1538,6 +1538,7 @@ class AffiliateWindowSerializer(serializers.ModelSerializer):
     custom3 = serializers.SerializerMethodField()
     custom4 = serializers.SerializerMethodField()
     custom5 = serializers.CharField(source='course_run.short_description')
+    custom6 = serializers.CharField(source='course_run.weeks_to_complete')
```
waheedahmed: I believe we should append week/weeks like we do on about pages otherwise it's hard to understand what this is for by just looking at the numeric value, for example `1 week`, `4 weeks` etc.

waheedahmed (changes requested):

course_discovery/apps/api/serializers.py
```diff
@@ -1597,6 +1599,12 @@ def get_custom3(self, obj):
     def get_custom4(self, obj):
         return ','.join(org.name for org in obj.course_run.authoring_organizations.all())
 
+    def get_custom6(self, obj):
+        weeks = obj.course_run.weeks_to_complete
+        if weeks > 0:
+            return "{} {}".format(weeks, 'week' if weeks == 1 else 'weeks')
```
waheedahmed: Stick to single quotes.

waheedahmed (approved):
One minor comment otherwise 👍, also squash your commits before merging.

PR #2114, opened 2019-08-23
DISCO-1329 delete publisher_enable_course_import waffle

xnick421:


course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -4273,24 +4272,11 @@ def test_page_without_superuser(self):
...
         self.assertEqual(response.status_code, 200)
 
-    @override_switch('publisher_enable_course_import', False)
-    def test_page_with_superuser_without_waffle(self):
-        """ Verify that user can't access page when not logged in. """
-        response = self._make_users_valid()
-        self.assertRaises(Http404)
-        self.assertEqual(response.status_code, 404)
-
-    def test_page_with_waffle_without_superuser(self):
```
mikix, line 4288: Why delete this test?

course_discovery/apps/publisher/views.py
```diff
@@ -1465,17 +1465,14 @@ def get_context_data(self, **kwargs):
...
-        if self.request.user.is_superuser and waffle.switch_is_active('publisher_enable_course_import'):
+        if self.request.user.is_superuser:
             return super(AdminImportCourse, self).get(request, args, **kwargs)
         else:
             raise Http404
 
     def post(self, request, *args, **kwargs):
         """Post method for import page."""
 
-        if not (self.request.user.is_superuser and waffle.switch_is_active('publisher_enable_course_import')):
```
mikix, line 1476: You should keep this line. If we always take the `True` path for the waffle check, it becomes:
`if not (self.request.user.is_superuser and True):`
which becomes:
`if not self.request.user.is_superuser:`

mikix (changes requested):

course_discovery/apps/publisher/views.py
xnick421: i was switching back and forth between this one and 1331 where we want the false path.  prob forgot which branch i was on.  fixing

course_discovery/apps/publisher/tests/test_views.py
xnick421: this test is actually duplicated i believe, ill fix them up

mikix (approved):

PR #2115, opened 2019-08-23
DISCO-1330 delete publisher waffle flag

xnick421:
delete publisher_create_audit_seats_for_verified_course_runs waffle


mikix (approved):
Huh, I'm surprised that didn't break more tests (creating seats all of a sudden where before it didn't)

PR #2118, opened 2019-08-26
make url_slug unique

rgraber:
1. Make the url_slug unique (with draft status) in course metadata
2. Make url_slug unique in publisher
3. Add migrations to course metadata and publisher to backfill unique URL slugs for all existing courses, keeping course metadata and publisher in sync
4. Update error text for when slug is not unique


course_discovery/apps/publisher/migrations/0080_populate_url_slug.py
```diff
@@ -0,0 +1,31 @@
...
+def migrate_data_forward(apps, schema_editor):
+    for instance in Course.objects.all():
+        course_metadata_course = instance.discovery_counterpart
+        if course_metadata_course:
+            instance.url_slug = course_metadata_course.url_slug
+        instance.save()
+
+
+def migrate_data_backward(apps, schema_editor):
+    pass
```
mikix: I believe you can just neglect to pass a backward argument if you don't have anything to do.

course_discovery/apps/publisher/migrations/0080_populate_url_slug.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2019-08-27 21:10
+from __future__ import unicode_literals
+
+from django.db import migrations
+from course_discovery.apps.publisher.models import Course
```
mikix: This should be something like `Course = apps.get_model('publisher', 'course')` inside the forward method. See other migration files. This is done to avoid imports breaking as models go away over time but migration files stay the same.

course_discovery/apps/publisher/migrations/0080_populate_url_slug.py
```diff
@@ -0,0 +1,31 @@
...
+
+def migrate_data_forward(apps, schema_editor):
+    for instance in Course.objects.all():
+        course_metadata_course = instance.discovery_counterpart
```
mikix: This is an extra lookup per loop... Maybe that's fine, tables aren't huge.

I think it will also break once you do the suggestion above (`apps.get_model`) because the model class returned from that call is not the real model. It's some weird shim that Django provides. So it might not have `@property` methods like `discovery_counterpart`. I'm not sure if this whole paragraph is true or not.

course_discovery/apps/course_metadata/models.py
```diff
@@ -505,7 +505,7 @@ class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
     key = models.CharField(max_length=255, db_index=True)
     title = models.CharField(max_length=255, default=None, null=True, blank=True)
     url_slug = AutoSlugField(populate_from='title', editable=True, slugify_function=uslugify, overwrite_on_add=False,
-                             help_text=_('Leave this field blank to have the value generated automatically.'))
+                             help_text=_('Leave this field blank to have the value generated automatically.'),)
```
mikix: Seems like a leftover edit?

course_discovery/apps/course_metadata/models.py
```diff
@@ -573,6 +573,7 @@ class Meta:
         unique_together = (
             ('partner', 'uuid', 'draft'),
             ('partner', 'key', 'draft'),
+            ('url_slug', 'draft'),
```
mikix: Have you confirmed prod/stage don't have any dups?

course_discovery/apps/course_metadata/migrations/0195_save_all_url_slugs.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2019-08-27 21:10
+from __future__ import unicode_literals
+
+from django.db import migrations
+from course_discovery.apps.course_metadata.models import Course
```
mikix: Same import comment as other migration file.

course_discovery/apps/course_metadata/migrations/0195_save_all_url_slugs.py
```diff
@@ -0,0 +1,32 @@
...
+
+def migrate_data_forward(apps, schema_editor):
+    for instance in Course.objects.all():
+        instance.save()
```
mikix: Maybe a comment that you're doing this to set the slug.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -47,6 +48,23 @@ def inner(*args, **kwargs):
                             status=status.HTTP_400_BAD_REQUEST)
         except (PermissionDenied, Http404):
             raise  # just pass these along
+        except IntegrityError as ex:
+            if ex.args and ex.args[0] == 1062:
+                # special error message for slug uniqueness violation
+                error_regex = 'Duplicate entry \'(.*)-[01]\' for ' \
+                              'key \'course_metadata_course_url_slug_draft_4ddec6b0_uniq\''
```
mikix: That seems very brittle. Where is that hex coming from? What about Open edX users that might have different database backends? What about us in the future if we change backends?

You could maybe test if the slug already exists as a way of checking if that is why the IntegrityError happened? I dunno.

course_discovery/apps/course_metadata/models.py
rgraber: The full migration includes wiping out all existing url_slugs before doing the update since otherwise there's not really a great way to determine which ones get which version, and we weren't using them yet anyway.

rgraber: But there isn't a wipeout migration on publisher I just realized, so I'll add that.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -47,6 +48,29 @@ def inner(*args, **kwargs):
                             status=status.HTTP_400_BAD_REQUEST)
         except (PermissionDenied, Http404):
             raise  # just pass these along
+        except IntegrityError as ex:
+            # special handling for a url_slug uniqueness error when creating or editing a course
+            # need to handle this here because we can't make these queries inside an atomic block
```
Dillon-Dumesnil: Out of curiosity, why can't these checks be inside an atomic block?

Something like before saving, calling: 
```
if Course.everything.filter(url_slug=form_data['url_slug']).exists():
    raise ValidationError('The course URL slug [{slug}] is already in use. Please update this field and try again.'.format(slug=form_data['url_slug'])
```

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -47,6 +48,29 @@ def inner(*args, **kwargs):
...
+            # 1062 is the code for a uniqueness violation
```
Dillon-Dumesnil: I don't think this line is still needed after the changes you made from Mike Terry's comment earlier.

course_discovery/apps/course_metadata/migrations/0193_remove_course_url_slug.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2019-08-27 21:08
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
Dillon-Dumesnil: In general with all of these migrations in this PR, they can all be put into a single migration and you can dictate the order by the order they show up in for the operations. I.e.
```
    operations = [
        migrations.RemoveField(
            model_name='course',
            name='url_slug',
        ),
        migrations.AddField(
            model_name='course',
            name='url_slug',
            field=django_extensions.db.fields.AutoSlugField(blank=True, editable=False, help_text='Leave this field blank to have the value generated automatically.', populate_from='title'),
        ),
    ]

File: course_discovery/apps/api/v1/views/courses.py (hunk 9 above)
Comment (rgraber) [reply course_discovery/apps/api/v1/views/courses.py]: I kept getting "  "An error occurred in the current transaction. You can't "
TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block." Further research suggests there's a way around it but I don't know if this is enough of a potential problem to justify the time sink.

File: course_discovery/apps/course_metadata/migrations/0193_remove_course_url_slug.py [hunk 12]
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('course_metadata', '0193_migratecommentstosalesforce'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
Dillon-Dumesnil: Why are you removing this field just to re-add it again below? That adds a red flag in my mind.

course_discovery/apps/api/v1/views/courses.py
rgraber: Ah, I think the problem was I was originally trying to call it within an exception handler within the atomic transaction, which was not legal

course_discovery/apps/course_metadata/migrations/0196_save_all_url_slugs.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2019-08-27 21:10
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def migrate_data_forward(apps, schema_editor):
+    Course = apps.get_model('course_metadata', 'course')
+    for instance in Course.objects.all():
```
Dillon-Dumesnil: So I'm personally not a huge fan of needing to add in the `MigratableCourseDraftManager`. It has to do with `Course.objects` being a `DraftManager` and for whatever reason, that doesn't show up. I don't know, it's weird.

Anyway, I think there are two cleaner ways to handle this situation:
1) Just use the `everything` Manager since we know that works. So this line can just change to `for instance in Course.everything.filter(draft=False):`
2) Add `use_in_migrations` to the `DraftManager` class so that way `objects` works in Migrations for all of our models (it currently does for other models and is the reason `everything` already works)

course_discovery/apps/course_metadata/models.py
```diff
@@ -567,12 +567,13 @@ class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
...
 
     everything = CourseQuerySet.as_manager()
-    objects = DraftManager.from_queryset(CourseQuerySet)()
+    objects = MigratableCourseDraftManager()
 
     class Meta:
         unique_together = (
             ('partner', 'uuid', 'draft'),
             ('partner', 'key', 'draft'),
+            ('url_slug', 'draft'),
```
Dillon-Dumesnil: This should still have `'partner'`. The partner is like our different sites and we want to continue to enforce uniqueness between them

course_discovery/apps/course_metadata/models.py
```diff
@@ -567,12 +567,13 @@ class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
     additional_information = NullHtmlField(verbose_name=_('Additional Information'))
 
     everything = CourseQuerySet.as_manager()
-    objects = DraftManager.from_queryset(CourseQuerySet)()
+    objects = MigratableCourseDraftManager()
```
Dillon-Dumesnil: See above about MigratableCourseDraftManager

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -212,6 +214,46 @@ def test_publish(self):  # pylint: disable=too-many-statements
             **common_seat_kwargs
         )
 
+    @responses.activate
+    def test_publish_with_duplicate_url_slug(self):
+        publisher_course_run = self._create_course_run_for_publication()
+        publisher_course_run.course.url_slug = 'duplicate'
+        publisher_course_run.course.save()
+        publisher_course_run.save()
```
Dillon-Dumesnil: Is it necessary to save the course run here? Just curious if we can avoid another database write.

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -212,6 +214,46 @@ def test_publish(self):  # pylint: disable=too-many-statements
...
+        discovery_matching_course = CourseFactory()
```
Dillon-Dumesnil: You can just switch this to `CourseFactory(draft=False, url_slug='duplicate')`

If url_slug is not in the CourseFactory, I'd probably just recommend adding it as a FuzzyText

course_discovery/apps/publisher/migrations/0080_remove_course_url_slug.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2019-08-29 20:39
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
Dillon-Dumesnil: Same as above. You shouldn't need multiple migration files. Also want to know about the remove and add.

course_discovery/apps/publisher/migrations/0082_populate_url_slug.py
```diff
@@ -0,0 +1,35 @@
...
+    DiscoveryCourse = apps.get_model('course_metadata', 'course')
+    for instance in Course.objects.all():
+        # recreate organization() and partner() methods
+        organization = instance.organizations.all().first()
+        partner = organization.partner if organization else None
+        key = '{org}+{number}'.format(org=instance.organizations.first().key, number=instance.number)
+        try:
+            discovery_course = DiscoveryCourse.objects.get(partner=partner, key=key)
+            instance.url_slug = discovery_course.url_slug
+        except:
```
Dillon-Dumesnil: Blanket `except` here is a kind of scary thing. Do we not care what might go wrong and we should continue the migration? Are there expected issues that we can specifically catch? Are there _any_ scenarios that this migration should fail?

course_discovery/apps/course_metadata/managers.py
```diff
@@ -24,3 +26,8 @@ def get_draft(self, **kwargs):
         If a draft is not available, we give back the non-draft version.
         """
         return self.filter_drafts(**kwargs).get()
+
+
+class MigratableCourseDraftManager(DraftManager.from_queryset(CourseQuerySet)):
```
Dillon-Dumesnil: See comment below (in one of the migration files) regarding this class.

course_discovery/apps/publisher/migrations/0082_populate_url_slug.py
```diff
@@ -0,0 +1,35 @@
...
+def migrate_data_forward(apps, schema_editor):
+    Course = apps.get_model('publisher', 'course')
+    DiscoveryCourse = apps.get_model('course_metadata', 'course')
+    for instance in Course.objects.all():
+        # recreate organization() and partner() methods
+        organization = instance.organizations.all().first()
+        partner = organization.partner if organization else None
+        key = '{org}+{number}'.format(org=instance.organizations.first().key, number=instance.number)
+        try:
+            discovery_course = DiscoveryCourse.objects.get(partner=partner, key=key)
```
Dillon-Dumesnil: Did `discovery_counterpart` not work? I feel like that would be the cleaner way if possible.

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -212,6 +214,46 @@ def test_publish(self):  # pylint: disable=too-many-statements
...
+        discovery_matching_course.draft = True
+        discovery_matching_course.save()
+
+        with pytest.raises(IntegrityError):
+            url = reverse('publisher:api:v1:course_run-publish', kwargs={'pk': publisher_course_run.pk})
+            self.client.post(url, {})
+
+        publisher_course_run.course.url_slug = 'fixed'
+        publisher_course_run.course.save()
+        publisher_course_run.save()
```
Dillon-Dumesnil: Same here about is this necessary

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -212,6 +214,46 @@ def test_publish(self):  # pylint: disable=too-many-statements
...
+            self.client.post(url, {})
+
+        discovery_matching_course.draft = True
+        discovery_matching_course.save()
+
+        with pytest.raises(IntegrityError):
+            url = reverse('publisher:api:v1:course_run-publish', kwargs={'pk': publisher_course_run.pk})
+            self.client.post(url, {})
+
+        publisher_course_run.course.url_slug = 'fixed'
```
Dillon-Dumesnil: I think this should be broken off into a second test. With the test name `test_publish_with_duplicate_url_slug`, this part feels misleading. Then you have one test for the success case and one for the fail case with the uniqueness being broken.

Dillon-Dumesnil (changes requested):
It's a good start, just a couple things to fix up and I had a few questions throughout.

Also don't forget to add tests for uniqueness within the view and course metadata. There is only the test right now for the publisher app

course_discovery/apps/course_metadata/migrations/0193_remove_course_url_slug.py
rgraber: I needed to wipe all existing url_slugs because there was a chance they wouldn't be unique since it was originally added without that constraint

Dillon-Dumesnil: I would argue to wipe it as part of the function in the other migration then. Just null it to start and then do the save so it is automatically created. Removing and re-adding the same field feels hacky to get this result.

course_discovery/apps/publisher/migrations/0082_populate_url_slug.py
rgraber: No, discovery counterpart doesn't work because of the way models are created in migrations. It needs to be recreated by hand.

course_discovery/apps/course_metadata/migrations/0196_save_all_url_slugs.py
rgraber: I'm ok with using the .everything because the managers are themselves serialized and versioned in migrations, but FYI adding in use_in_migrations to the DraftManager class does not actually work (related to https://stackoverflow.com/questions/37764070/django-migrations-valueerror-could-not-find-manager-in-django-db-models-manager ).

Dillon-Dumesnil: Oh weird, I totally thought that would work. Weird. Oh well, I think the everything fix will be good for now.

course_discovery/apps/publisher/api/v1/tests/test_views.py
rgraber: url_slug should be auto-generated on save from the title, so I don't think they need to be in the CourseFactory

course_discovery/apps/publisher/api/v1/tests/test_views.py
rgraber: Since there already is a test for a successful publish I just added checking the url_slug into that

rgraber:
> It's a good start, just a couple things to fix up and I had a few questions throughout.
> 
> Also don't forget to add tests for uniqueness within the view and course metadata. There is only the test right now for the publisher app

I'm not sure it's necessary to test the course_metadata uniqueness since that's entirely enforced via a django constraint, and it doesn't seem right to unit test specific error messaging

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -197,6 +197,14 @@ def create(self, request, *args, **kwargs):
         if Course.objects.filter(partner=partner, key=course_creation_fields['key']).exists():
             raise Exception(_('A course with key {key} already exists.').format(key=course_creation_fields['key']))
 
+        # if a manually entered url_slug, ensure it's not already taken (auto-generated are guaranteed uniqueness)
+        if Course.everything.filter(url_slug=course_creation_fields['url_slug'],
```
Dillon-Dumesnil: I think I would update this to
```
if course_creation_fields['url_slug'] and Course.everything.filter(
    url_slug=course_creation_fields['url_slug'], 
    partner_id=course_creation_fields['partner']
).exists():
```
This way you can early exit if `url_slug` was never defined and don't have to hit the database unless necessary

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -197,6 +197,14 @@ def create(self, request, *args, **kwargs):
...
+                                    partner_id=course_creation_fields['partner']):
+            raise Exception(
+                _('Course edit was unsuccessful. '
```
Dillon-Dumesnil: Since this is its own message, I think we should be more explicit here that `Course creation was unsuccessful.` rather than just saying edit.

Nit: Write each line as far as it will go rather than break on sentences. Might be able to reduce overall line count.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -308,6 +316,14 @@ def update_course(self, data, partial=False):
         # (besides image and video, they are popped off above)
         changed = changed or reviewable_data_has_changed(course, serializer.validated_data.items())
 
+        if serializer.validated_data.get('url_slug'):
+            if Course.everything.filter(url_slug=serializer.validated_data['url_slug'],
+                                        partner=course.partner).exclude(key=course.key):
```
Dillon-Dumesnil: Nit: I would recommend switching this to exclude based on `UUID` rather than `key`

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -308,6 +316,14 @@ def update_course(self, data, partial=False):
         # (besides image and video, they are popped off above)
         changed = changed or reviewable_data_has_changed(course, serializer.validated_data.items())
 
+        if serializer.validated_data.get('url_slug'):
```
Dillon-Dumesnil: Combine these two ifs into an `if` with an `and`.
```
if serializer.validated_data.get('url_slug') and Course.everything.filter(
    url_slug=serializer.validated_data['url_slug'], partner=course.partner
).exclude(key=course.key):
```

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -197,6 +197,14 @@ def create(self, request, *args, **kwargs):
         if Course.objects.filter(partner=partner, key=course_creation_fields['key']).exists():
             raise Exception(_('A course with key {key} already exists.').format(key=course_creation_fields['key']))
 
+        # if a manually entered url_slug, ensure it's not already taken (auto-generated are guaranteed uniqueness)
+        if Course.everything.filter(url_slug=course_creation_fields['url_slug'],
+                                    partner_id=course_creation_fields['partner']):
+            raise Exception(
+                _('Course edit was unsuccessful. '
+                    'The course URL slug ‘{}’ is already in use. '
```
Dillon-Dumesnil: Please line up any successive lines with the line above it. Like, having the beginning quote at the same indentation level.
```
    _('Course edit was unsuccessful. '
      'The course URL...
```

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -308,6 +316,14 @@ def update_course(self, data, partial=False):
         # (besides image and video, they are popped off above)
         changed = changed or reviewable_data_has_changed(course, serializer.validated_data.items())
 
+        if serializer.validated_data.get('url_slug'):
+            if Course.everything.filter(url_slug=serializer.validated_data['url_slug'],
+                                        partner=course.partner).exclude(key=course.key):
+                raise Exception(
+                    _('Course edit was unsuccessful. '
+                        'The course URL slug ‘{}’ is already in use. '
```
Dillon-Dumesnil: Same comment about indentation level

Dillon-Dumesnil: Same comment about wrapping with brackets and using format variable.

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: I think it is our convention to wrap user input in brackets to help indicate to transifex users that it doesn't need to be translated. So `{}` -> [{}]. I would also recommend adding in a format variable just for additional clarity. So having The course URL slug [{slug}]...).format(slug=serializer.validated_data['url_slug']) instead of just [{}]

course_discovery/apps/course_metadata/migrations/0194_initialize_course_url_slug.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def migrate_data_forward(apps, schema_editor):
+        updated_pks = []
+        Course = apps.get_model('course_metadata', 'course')
+        for instance in Course.everything.all().order_by('draft'):
+            # will set the url_slug
+            if not instance.draft:
+                instance.url_slug = ''
+                instance.save()
+                # update corresponding draft instance
```
Dillon-Dumesnil, line 18: The remainder of this function needs to be wrapped in an `if instance.draft_version:` since not all official rows necessarily have draft versions.

course_discovery/apps/course_metadata/migrations/0194_initialize_course_url_slug.py
```diff
@@ -0,0 +1,41 @@
...
+# Generated by Django 1.11.23 on 2019-09-04 18:12
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    def migrate_data_forward(apps, schema_editor):
+        updated_pks = []
```
Dillon-Dumesnil: Nit: Change this variable name to `updated_draft_pks` for clarity since only those pks are added to this list and not all pks that are updated.

course_discovery/apps/course_metadata/models.py
```diff
@@ -573,6 +573,7 @@ class Meta:
         unique_together = (
             ('partner', 'uuid', 'draft'),
             ('partner', 'key', 'draft'),
+            ('partner', 'url_slug', 'draft'),
```
Dillon-Dumesnil, line 576: 👍

Dillon-Dumesnil (changes requested):
It's looking good. One key thing to fix in the course_metadata migration, but otherwise pretty much there. Nice job

course_discovery/apps/api/v1/views/courses.py
rgraber: I believe pycodestyle complains about that and calls it under-indented but I'll double check.

Dillon-Dumesnil: Oh weird. I could have sworn I've been yelled at in the past by pycode about that, but maybe things changed when we switched from pylint to pycode. I guess do whichever the linter is okay with and maybe it just accepts both. Sounds good. Thanks for checking!

rgraber: Turns out I was accidentally lining the text up with the underscore instead of the other text which was why it was complaining.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -197,6 +197,14 @@ def create(self, request, *args, **kwargs):
         if Course.objects.filter(partner=partner, key=course_creation_fields['key']).exists():
             raise Exception(_('A course with key {key} already exists.').format(key=course_creation_fields['key']))
 
+        # if a manually entered url_slug, ensure it's not already taken (auto-generated are guaranteed uniqueness)
+        if (course_creation_fields['url_slug'] and
+            Course.everything.filter(url_slug=course_creation_fields['url_slug'],
+                                     partner_id=course_creation_fields['partner']).exists()):
+            raise Exception(
+                _('Course creation was unsuccessful. The course URL slug ‘[{url_slug}]’ is already in use. Please '
```
Dillon-Dumesnil, line 205: Nit: Remove the ticks and just use the brackets.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -307,6 +315,14 @@ def update_course(self, data, partial=False):
         # If price didnt change, check the other fields on the course
         # (besides image and video, they are popped off above)
         changed = changed or reviewable_data_has_changed(course, serializer.validated_data.items())
+        validated_data = serializer.validated_data
+
+        if (validated_data.get('url_slug') and
+            Course.everything.filter(url_slug=validated_data['url_slug'],
+                                     partner=course.partner).exclude(uuid=course.uuid).exists()):
+            raise Exception(
+                _('Course edit was unsuccessful. The course URL slug ‘[{url_slug}]’ is already in use. '
```
Dillon-Dumesnil, line 324: Same nit here

Dillon-Dumesnil (approved):
Looks good! I had a small nit, but feel free to ignore

dianakhuang (approved):
:shipit:

PR #2119, opened 2019-08-27
Ndalfonso/disco 1331 delete publisher enable read only fields

xnick421:


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -220,13 +218,6 @@ def format_course_run_data(self, body, course=None):
         start = self.parse_date(body['start'])
         pacing_type = self.get_pacing_type(body)
```
mikix: Since you're removing the second place these two variables are used, you could probably move them inline to their single use below now.

course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -91,130 +80,6 @@ def _mock_ecommerce_api(self, publisher_course_run, status=200, body=None):
         url = '{root}publication/'.format(root=self.partner.ecommerce_api_url)
         responses.add(responses.POST, url, json=body, status=status)
 
-    @responses.activate
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
-    def test_publish(self):  # pylint: disable=too-many-statements
```
mikix, line 96: You still want this test. This tests publishing to ecom, discovery, and studio. It's not just about that waffle flag. I think you just need to remove the switch override and fix whatever behavior breaks in the test because it's off.

course_discovery/apps/publisher/models.py
```diff
@@ -511,36 +509,15 @@ def pacing_type_temporary(self):
             The progress of the above work will be tracked in the following ticket:
             https://openedx.atlassian.net/browse/EDUCATOR-3488.
```
mikix, line 510: Might want to fix up this docstring too - mention that this is old unused method. 

Or better yet, remove this method, now that we don't need it and fix up users of it. Same for other `_temporary` methods. But that all might be too much. Maybe better to just leave them as simple fall backs for now (with the comment on why) and we'll clean it up when decommissioning publisher.

course_discovery/apps/publisher/models.py
```diff
@@ -511,36 +509,15 @@ def pacing_type_temporary(self):
...
-        else:
-            return self.pacing_type
+        return self.pacing_type
 
     @pacing_type_temporary.setter
     def pacing_type_temporary(self, value):
-        if waffle.switch_is_active(PUBLISHER_ENABLE_READ_ONLY_FIELDS):
-            raise CourseRunEditException
-        else:
-            # Treat empty strings as NULL
```
mikix, line 529: You deleted this comment

course_discovery/apps/publisher/tests/test_admin.py
```diff
@@ -37,18 +35,6 @@ def setUp(self):
 
         self.assertFalse(CourseRun.objects.filter(lms_course_id__isnull=True).exists())
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
-    def test_update_course_form(self):
```
mikix, line 41: This test seems unrelated to the waffle feature? I think we still want it, just without the override.

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -244,75 +241,16 @@ def test_discovery_counterpart_failure_without_course(self):
 
         assert course_run.discovery_counterpart is None
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=False)
     def test_pacing_type_temporary_without_switch(self):
```
mikix: Might rename to just `test_pacing_type_temporary` - same comment throughout this file.

course_discovery/apps/publisher/tests/test_studio_api_utils.py
```diff
@@ -62,35 +59,6 @@ def assert_data_generated_correctly(course_run, expected_team_data):
     assert StudioAPI.generate_data_for_studio_api(course_run) == expected
 
 
-@pytest.mark.django_db
-@override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
-def test_generate_data_for_studio_api():
```
mikix, line 67: These two tests seem unrelated to the switch - I'd imagine we still want them, right?

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -936,21 +933,6 @@ def test_page_without_login(self):
             target_status_code=302
         )
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
-    def test_page_without_data(self):
```
mikix, line 940: I think we still want this.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -966,36 +948,6 @@ def _generate_credit_seat(self):
         """ Helper method to add credit seat for a course-run. """
         factories.SeatFactory(type='credit', course_run=self.course_run, credit_provider='ASU', credit_hours=9)
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
-    def test_course_run_detail_page_staff(self):
```
mikix, line 970: We want this.

course_discovery/apps/publisher/tests/test_views.py
```diff
@@ -1106,34 +1058,6 @@ def test_course_run_with_version(self):
         self.assertEqual(response.status_code, 200)
         self.assertNotContains(response, 'Enrollment Track')
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=True)
-    def test_detail_page_with_comments(self):
```
mikix, line 1110: We want this.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -109,27 +106,6 @@ def test_email_without_any_role(self):
         self.create_comment(content_object=self.course)
         self.assertEqual(len(mail.outbox), 0)
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=False)
-    def test_course_run_without_start_date(self):
```
mikix, line 113: We want this - this test had the flag disabled - nothing should change for this test. Though you might want to clean up the docstring.

mikix (changes requested):

course_discovery/apps/publisher_comments/tests/test_emails.py
xnick421: there is a note in this test should be deleted

mikix: Yeah, but remember that we are disabling this waffle flag in the `False` path. We no longer want the feature this waffle was designed to provide. And that note in the test was saying "once we remove this waffle flag as `True`, we don't need this test" - but we're removing it as `False` so test is still relevant.

course_discovery/apps/publisher_comments/tests/test_emails.py
```diff
@@ -109,12 +106,10 @@ def test_email_without_any_role(self):
...
         self.assertEqual(len(mail.outbox), 0)
 
-    @override_switch(PUBLISHER_ENABLE_READ_ONLY_FIELDS, active=False)
     def test_course_run_without_start_date(self):
         """ Verify that emails works properly even if course-run does not have a start date.
-
-            Note: Once PUBLISHER_ENABLE_READ_ONLY_FIELDS switch is ready to be removed, this test should
-            be deleted, as there will be no way to have a course run without a start date in Publisher.
+        Note: Once PUBLISHER_ENABLE_READ_ONLY_FIELDS switch is ready to be removed, this test should
+        be deleted, as there will be no way to have a course run without a start date in Publisher.
```
mikix: You don't need this comment anymore.

mikix (approved):

PR #2122, opened 2019-08-27
Handle mismatched draft/official rows

mikix:
When migrating, try to match up draft and official rows
a little harder.


Dillon-Dumesnil (approved):
Good catch

PR #2126, opened 2019-08-29
DISCO-1340 remove publisher waffles switches from DB

xnick421:


course_discovery/apps/publisher/migrations/0080_remove_publisher_waffle_switches.py
```diff
@@ -0,0 +1,35 @@
...
+WAFFLES = ('publisher_hide_features_for_pilot', 'publisher_add_instructor_feature',
+           'publisher_comment_widget_feature', 'publisher_approval_widget_feature',
+           'publisher_history_widget_feature')
+
+
+def delete_switches(apps, schema_editor):
+    """Delete the publisher switches."""
+    Switch = apps.get_model('waffle', 'Switch')
+    for waffle in WAFFLES:
+        Switch.objects.filter(name=waffle).delete()
```
mikix: You can just do:

```
Switch.objects.filter(name__in=WAFFLES).delete()
```

mikix (approved):
Looks good. Looks like we could have also done the `enable_publisher_email_notifications` flag during our removal. But no biggie.

PR #2127, opened 2019-08-29
Use proper serializer and limit result.

dianakhuang:
https://openedx.atlassian.net/browse/WEBSITE-64

Use the correct serializer in the correct places. Fixes page sizing and also `format=json`.


rgraber (approved):
LGTM

PR #2128, opened 2019-08-30
DISCO-1039: Adds in bulk load salesforce comments command

jlajoie:


course_discovery/apps/api/v1/views/comments.py
```diff
@@ -28,6 +29,11 @@ def get_queryset(self):
 
     def list(self, request):
         partner = request.site.partner
+        try:
+            util = SalesforceUtil(partner)
+        except SalesforceConfiguration.DoesNotExist:
+            raise Http404
```
mikix: Tiny nit: you use this twice. Maybe a utility method?

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    def handle(self, *args, **options):
+        config = MigrateCommentsToSalesforce.get_solo()
+        orgs = config.orgs.all()
+        if not orgs:
+            logger.error(
+                'No organizations were defined. Please add organizations to the MigrateCommentsToSalesforce model.'
+            )
+            raise CommandError('No organizations were defined.')
```
mikix: Strings aren't translated. Maybe we're the only ones that will ever run this though... 

You could probably save yourself a string by just having the long one be sent via CommandError. But doesn't matter. Maybe in some contexts the CommandError string isn't printed?

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+        partner = config.partner
+        if not partner:
+            logger.error('No partner was defined. Please add a partner to the MigrateCommentsToSalesforce model.')
```
mikix: These messages so verbose and helpful! :( Maybe just `Check yourself before you wreck yourself.`?

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+            raise CommandError('Salesforce configuration for {} does not exist'.format(partner.name))
+        for org in orgs:
+            logger.info('Executing for Organization: {}'.format(org.name))
+            courses = CourseMetadataCourse.objects.filter_drafts().filter(
+                authoring_organizations__in=[org]
+            )
+            logger.info('Found {} courses for {}'.format(len(courses), org.name))
+            for course in courses:
+                # Just need to create a case for the course here, as it will also create the Org and the Course itself
+                util.create_case_for_course(course)
```
mikix, line 50: I assume this is idempotent, too lazy to look (and the script in general?)

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+                        )
+                    except PublisherCourseRun.DoesNotExist:
+                        logger.warning('No PublisherCourseRun found for {}.'.format(course_run.key))
+                        continue
+                    publisher_course = publisher_course_run.course
+                    # Order by submit_date as that is when comments were added
+                    publisher_course_run_comments = Comments.objects.filter(
+                        content_type=ContentType.objects.get_for_model(PublisherCourseRun),
+                        object_pk=publisher_course_run.id,
+                    ).select_related('user')
```
mikix, line 73: I was like `Comments`?! What model is that?!

I've never touched that djangoapp.

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+                    for comment in publisher_course_run_comments:
+                        user_comment_body = util.format_user_comment_body(
+                            comment.user, comment.comment, course_run_key=course_run.key
+                        )
+                        comments.append({
+                            'ParentId': course.salesforce_case_id,
+                            'Body': user_comment_body,
+                            '_modified': comment.modified,  # To be removed later, just for ordering before bulk insert
```
mikix: Ah, I got confused by `later`. Made me think this was like, deprecated. Maybe `below`?

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+                    util.create_course_run(course_run)
+                    # Start at the course_run level because that's the only real way to get the Publisher Course
+                    try:
+                        publisher_course_run = PublisherCourseRun.objects.select_related('course').get(
+                            lms_course_id=course_run.key
+                        )
+                    except PublisherCourseRun.DoesNotExist:
+                        logger.warning('No PublisherCourseRun found for {}.'.format(course_run.key))
+                        continue
+                    publisher_course = publisher_course_run.course
```
mikix: These might be different publisher courses for the same runs inside one CM course...

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -0,0 +1,100 @@
...
+                        user_comment_body = util.format_user_comment_body(comment.user, comment.comment)
+                        comments.append({
+                            'ParentId': course.salesforce_case_id,
+                            'Body': user_comment_body,
+                            '_modified': comment.modified,  # To be removed later, just for ordering before bulk insert
+                        })
+                    if comments:
+                        for comment in comments:
+                            del comment['_modified']
+                        util.client.bulk.FeedItem.insert(comments)
```
mikix, line 99: This indentation level (under the `if publisher_course`) is probably fine, right? Every publisher run would have a link to a course.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -291,3 +292,15 @@ def _create_comment_return_body(user, body, course_run_key=None):
...
+    def _get_salesforce_equivalent(status):
+        # Note: these must match the equivalent 'picklistValues' for Salesforce's Course_Run__c.Publisher_Status__c
+        salesforce_statuses = {
+            CourseRunStatus.Unpublished: 'New/Unsubmitted Edits',
+            CourseRunStatus.LegalReview: 'In Legal Review',
+            CourseRunStatus.InternalReview: 'In PC Review',
+            CourseRunStatus.Reviewed: 'Scheduled',
+            CourseRunStatus.Published: 'Live',
+        }
+        return salesforce_statuses.get(status)
```
mikix, line 306: :+1:

mikix (approved):
I do think the one comment about potentially multiple Publisher courses for one CM course should be addressed/considered. But besides that, looks great!

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
mikix: That sounded like sarcasm, but they are helpful. And I was just making a joke suggestion. Stupid text with no tone.

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
jlajoie: I know! It directly inherits from what apparently is a default Django implementation of comments. Kinda neat

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
jlajoie: Yup! This method can be called any number of times.

One caveat I'm going to add in to our process about re-running for partially seeded Cases (for comments), that you'll need to manually drop those comments. The reason I don't want to just do it as part of the script is because of other people adding in comments and the harmful action against user submitted data that we'd jsut straight up lose.

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
jlajoie: Ahh damn is this true? I spoke w/ @Dillon-Dumesnil about this and he mentioned this is the most accurate way to associate a Publisher Course to a Course Metadata Course

jlajoie: ... thoughts on how to best approach that?

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
jlajoie: Yes was intentional, if there's no publisher course it'd feel a little weird to be inserting which is why I added that warning. Shouldn't fail it, but should be logged so we can at least trace back why comments weren't added.

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
jlajoie: I didn't think for internal commands we'd need to translate that's a super easy fix

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
mikix: He is right. You want to go run->run and then up. See find_discovery_course for example.

But the problem is that only works on a per-run basis. Like CM course A with runs B and C. B and C's Publisher counterpart runs might each point at different Publisher courses. So per-run, you can definitely find the right course. But your loop assumes that the Publisher course will be the same for all the runs. And that doesn't hold.

course_discovery/apps/api/v1/views/comments.py
jlajoie: I thought about it, I would have liked to put it in the SalesforceUtil itself but then it'd be one of those weird situations where "why is this raising a 404 when there's not web request" if you used it outside of the API. I can add it to the file though as a util 👍

PR #2129, opened 2019-08-30
Retry update index job upon sanity check failure

jmyatt:
We continue to see somewhat frequent failures of the job to update the ES search index due to the record count sanity check failing.  The common action taken when the alert is received is to manually re-run the jenkins job which usually results in job success.

After discussing with DevOps the complexities of adding a retry at the jenkins job config level, the approach in this PR of simply forcing a retry of the index build seems like the simplest approach to avoiding the noisy alerts.  This is an interim workaround until we put effort into upgrading our search backend, ES version, etc.

LEARNER-6364

### DIFF
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import time
 
@@ -53,9 +52,10 @@ class Command(HaystackCommand):
             for backend, index, alias, record_count in alias_mappings:
                 # Run a sanity check to ensure we aren't drastically changing the
                 # index, which could be indicative of a bug.
-                if not options.get('disable_change_limit', False):
-                    record_count_is_sane, index_info_string = \
-                        self.sanity_check_new_index(backend.conn, index, record_count)
+                if index in indexes_pending and not options.get('disable_change_limit', False):
+                    record_count_is_sane, index_info_string = self.sanity_check_new_index(
+                        backend.conn, index, record_count
+                    )
                     if record_count_is_sane:
                         self.set_alias(backend, alias, index)
                         indexes_pending.pop(index, None)
@@ -66,7 +66,7 @@ class Command(HaystackCommand):
                     indexes_pending.pop(index, None)
 
         if indexes_pending:
-            raise CommandError('Sanity check failed for new index(es). ' + json.dumps(indexes_pending))
+            raise CommandError('Sanity check failed for new index(es): {}'.format(indexes_pending))
 
     def percentage_change(self, current, previous):
         try:
--- a/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import time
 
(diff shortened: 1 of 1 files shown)


course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -1,3 +1,4 @@
+import json
```
jlajoie: Jason import json... had me a giggle at that.

But also can probably drop the json import if you're just using it for stringification purposes,

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -42,18 +43,30 @@ def handle(self, **options):
...
+                        self.set_alias(backend, alias, index)
+                        indexes_pending.pop(index, None)
+                    else:
+                        indexes_pending[index] = index_info_string
+                else:
+                    self.set_alias(backend, alias, index)
+                    indexes_pending.pop(index, None)
+
+        if indexes_pending:
+            raise CommandError('Sanity check failed for new index(es). ' + json.dumps(indexes_pending))
```
jlajoie: Mentioned above but you can probably drop the json.dumps for using things this way and just use the format interpolation.

e.g.

```
>>> a = [1,10,20,50,1000]
>>> a
[1, 10, 20, 50, 1000]
>>> 'This is a string with an array interpolated {} right before this point'.format(a)
'This is a string with an array interpolated [1, 10, 20, 50, 1000] right before this point'
```

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -42,18 +43,30 @@ def handle(self, **options):
...
+        indexes_pending = {key: '' for key in [x[1] for x in alias_mappings]}
+        while indexes_pending and run_attempts < 2:
+            run_attempts += 1
+            super(Command, self).handle(**options)
+
+            for backend, index, alias, record_count in alias_mappings:
+                # Run a sanity check to ensure we aren't drastically changing the
+                # index, which could be indicative of a bug.
+                if not options.get('disable_change_limit', False):
+                    record_count_is_sane, index_info_string = \
```
jlajoie: nit: I think we generally try to avoid line breaks in code, if you're trying to shorten the length then you might just be able to do the args on a new line? If just the 
```
record_count_is_sane, index_info_string = self.sanity_check_new_index(
``` 
is too long then whatever

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
```diff
@@ -42,18 +43,30 @@ def handle(self, **options):
...
-            if not options.get('disable_change_limit', False):
-                record_count_is_sane, index_info_string = self.sanity_check_new_index(backend.conn, index, record_count)
-                if not record_count_is_sane:
-                    raise CommandError('Sanity check failed for new index. ' + index_info_string)
-
-            self.set_alias(backend, alias, index)
+        run_attempts = 0
+        indexes_pending = {key: '' for key in [x[1] for x in alias_mappings]}
+        while indexes_pending and run_attempts < 2:
+            run_attempts += 1
```
jlajoie, line 49: Might be a little weird if we're getting inconsistencies?

Example:

3 indexes, A, B, C

First iteration:
Attempts set to 1
A and B succeed, popped form indexes_pending
C fails due to mismatch in count, stays in indexes_pending

Second iteration:
Attempts set to 2 (will terminate no matter what after this iteration)
Increments attempts to 2
A succeeds again, tries to pop from indexes_pending, defaults to None since not there
B fails this time, sets indexes_pending with the new index_info_string
C succeeds this time, pop from indexes_pending

We'll raise the CommandError now because B is inside of indexes_pending. I'd argue that it succeeded though?
 
I'd still say this is worth landing regardless to improve consistency but might be a case to consider.

jlajoie (approved):
Minor feedback, otherwise looks reasonable to me 👍

course_discovery/apps/edx_haystack_extensions/management/commands/update_index.py
jmyatt: 👍 made a change to fix that

PR #2135, opened 2019-09-04
Fixing course number doc to remove slack links since they expire

Dillon-Dumesnil:


jmyatt (approved):

jmyatt:
up to you, but you might want to prune/curate these a bit to include only the technically significant bits

Dillon-Dumesnil:
Deciding to leave them long so whoever ends up reading these in the future hopefully can gain a little more context through the conversations.

PR #2136, opened 2019-09-04
Fixes salesforce login for Prod environments

jlajoie:
With an empty string for domain, you receive a ```requests.exceptions.InvalidURL: URL has an invalid label.``` on login. However security token needs to be an empty string still. This update fixes that.


Dillon-Dumesnil (approved):
/shrug?

mikix (approved):
Sure man

PR #2137, opened 2019-09-04
Always pull start date from Studio

mikix:
We're switching to always editing the start date in Studio, not
Publisher. So always pull it in from Studio in RCM. And don't push
it to Studio from new Publisher unless we're creating the run.

Also use the new ENABLE_PUBLISHER feature flag - rather than whether
a marketing site is associated with a partner - to determine how
much data we pull from Studio.

https://openedx.atlassian.net/browse/DISCO-1313

### DIFF
--- a/course_discovery/apps/api/tests/test_utils.py
+++ b/course_discovery/apps/api/tests/test_utils.py
@@ -1,10 +1,13 @@
 import ddt
 import mock
 from django.test import TestCase
+from opaque_keys.edx.keys import CourseKey
 from rest_framework.request import Request
 from rest_framework.test import APIRequestFactory
 
-from course_discovery.apps.api.utils import cast2int, get_query_param
+from course_discovery.apps.api.utils import StudioAPI, cast2int, get_query_param
+from course_discovery.apps.core.utils import serialize_datetime
+from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory
 
 LOGGER_PATH = 'course_discovery.apps.api.utils.logger.exception'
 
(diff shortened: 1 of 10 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -208,17 +206,16 @@ def _update_instance(self, instance, validated_data, **kwargs):
...
             'key': body['id'],
+            'start': self.parse_date(body['start']),
             'end': self.parse_date(body['end']),
             'enrollment_start': self.parse_date(body['enrollment_start']),
             'enrollment_end': self.parse_date(body['enrollment_end']),
             'hidden': body.get('hidden', False),
             'license': body.get('license') or '',  # license cannot be None
         }
 
-        # When using a marketing site, only dates (excluding start) should come from the Course API.
```
mikix, line 218: I honestly don't know why we treated `start` differently. Does anyone else?

course_discovery/apps/api/tests/test_utils.py
```diff
@@ -40,3 +43,50 @@ def test_with_request(self):
...
+        return data
+
+    def test_create_rerun(self):
+        run1 = CourseRunFactory()
+        run2 = CourseRunFactory(course=run1.course)
+        self.api.create_course_rerun_in_studio(run2, run1.key)
+
+        expected_data = self.make_studio_data(run2)
+        self.assertEqual(self.client.course_runs.call_args_list, [mock.call(run1.key)])
+        self.assertEqual(self.client.course_runs.return_value.rerun.post.call_args_list[0][0][0], expected_data)
```
jlajoie, line 78: 👀 [0][0][0] haven't seen one of those in a while

course_discovery/apps/course_metadata/data_loaders/api.py
jlajoie: I honestly have no idea

course_discovery/apps/api/tests/test_utils.py
mikix: :) mock's call_args_list is overly complex.

mikix: Well, maybe appropriately so. But annoyingly complex for the common case.

course_discovery/apps/api/utils.py
```diff
@@ -150,10 +150,10 @@ def calculate_course_run_key_run_value(cls, course_num, start):
         return cls._get_next_run(run, '', related_course_runs)
 
     @classmethod
-    def generate_data_for_studio_api(cls, course_run, user=None):
+    def generate_data_for_studio_api(cls, course_run, creating, user=None):
```
jlajoie, line 153: Creating isn't a kwarg here but you use it as such down below with creating=True creating=False... is that fine?

course_discovery/apps/publisher/studio_api_utils.py
```diff
@@ -19,7 +19,7 @@ def _run_title(cls, course_run):
         return course_run.title_override or course_run.course.title
 
     @classmethod
-    def _run_times(cls, course_run):
+    def _run_times(cls, course_run, creating):
```
jlajoie, line 22: Doesn't look like you're using creating here?

course_discovery/apps/publisher/tests/test_studio_api_utils.py
```diff
@@ -84,7 +84,8 @@ def assert_data_generated_correctly(course_run, expected_team_data):
         'team': expected_team_data,
         'pacing_type': course_run.pacing_type_temporary,
     }
-    assert StudioAPI.generate_data_for_studio_api(course_run) == expected
+    # the publisher djangoapp doesn't care about the 'creating' flag passed below, so we just always set it False
+    assert StudioAPI.generate_data_for_studio_api(course_run, creating=False) == expected
```
jlajoie, line 88: nit: I think this is the first case I've seen of not using self.assertEquals but cool by me, know that's not a you change

jlajoie (approved):
Couple of nits but looks good!

course_discovery/apps/api/utils.py
mikix: Yeah, Python lets you do that. I did it because I didn't think there was a reasonable default? But passing just `True` or `False` to a method doesn't tell you what that boolean is for. So when calling this, I use the kwarg version for code-as-documentation purposes.

course_discovery/apps/publisher/studio_api_utils.py
mikix: No. Old Publisher will continue to always push start/end dates (we were leery of changing behavior there underneath folks). But maybe I should add a comment to that effect.

course_discovery/apps/publisher/tests/test_studio_api_utils.py
mikix: It mostly doesn't matter with pytest. And I think pytest upstream prefers the raw asserts? But I kinda don't. But yeah, didn't feel like changing it.

PR #2138, opened 2019-09-04
DISCO-1316 saleforce record updates

xnick421:
update course, course run, and organization in salesforce when they are update in the db, also push them to saleforce on creation

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_comments.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_comments.py
@@ -1,4 +1,6 @@
+import factory
 import mock
+from django.db.models.signals import post_save
 from rest_framework.reverse import reverse
 
 from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase, OAuth2Mixin
@@ -8,6 +10,7 @@ from course_discovery.apps.course_metadata.tests.factories import CourseFactory,
 
 
 class CommentViewSetTests(OAuth2Mixin, APITestCase):
+    @factory.django.mute_signals(post_save)
     def setUp(self):
         super(CommentViewSetTests, self).setUp()
         self.salesforce_config = SalesforceConfigurationFactory(partner=self.partner)
@@ -31,6 +34,7 @@ class CommentViewSetTests(OAuth2Mixin, APITestCase):
             self.assertEqual(response.status_code, 200)
             self.assertEqual(response.data, [])
 
+    @factory.django.mute_signals(post_save)
     def test_list_salesforce_case_id_set(self):
         self.course.salesforce_id = 'TestSalesforceId'
         self.course.save()
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -84,6 +84,25 @@ class DraftModelMixin(models.Model):
         abstract = True
 
 
+class CachedMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._cache = dict(self.__dict__)
+
+    def refresh_from_db(self, using=None, fields=None, **kwargs):
+        super().refresh_from_db(using, fields, **kwargs)
+        self.__dict__.pop('_cache', None)
+        self._cache = dict(self.__dict__)
+
+    def save(self, **kwargs):
+        super().save(**kwargs)
+        self.__dict__.pop('_cache', None)
+        self._cache = dict(self.__dict__)
+
+    def did_change(self, field):
+        return field in self.__dict__ and (field not in self._cache or getattr(self, field) != self._cache[field])
+
+
 class AbstractNamedModel(TimeStampedModel):
     """ Abstract base class for models with only a name field. """
     name = models.CharField(max_length=255, unique=True)
@@ -320,7 +339,7 @@ class AdditionalPromoArea(AbstractTitleDescriptionModel):
     pass
 
 
-class Organization(TimeStampedModel):
+class Organization(CachedMixin, TimeStampedModel):
     """ Organization model. """
     partner = models.ForeignKey(Partner, models.CASCADE, null=True, blank=False)
     uuid = models.UUIDField(blank=False, null=False, default=uuid4, editable=False, verbose_name=_('UUID'))
@@ -494,7 +513,7 @@ class PkSearchableMixin:
         return queryset.filter(pk__in=ids)
 
 
-class Course(DraftModelMixin, PkSearchableMixin, TimeStampedModel):
+class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
     """ Course model. """
     partner = models.ForeignKey(Partner, models.CASCADE)
     uuid = models.UUIDField(default=uuid4, editable=False, verbose_name=_('UUID'))
@@ -823,7 +842,7 @@ class CourseEditor(TimeStampedModel):
         return queryset.filter(user_can_edit, course__authoring_organizations__in=user_orgs).distinct()
 
 
-class CourseRun(DraftModelMixin, TimeStampedModel):
+class CourseRun(DraftModelMixin, CachedMixin, TimeStampedModel):
     """ CourseRun model. """
     OFAC_RESTRICTION_CHOICES = (
         ('', '--'),
@@ -1317,7 +1336,6 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
         # Push any product (seat, entitlement) changes to ecommerce as well
         if create_ecom_products:
             push_to_ecommerce_for_course_run(official_version)
-
         return official_version
 
     def handle_status_change(self, send_emails):
--- a/course_discovery/apps/course_metadata/salesforce.py
+++ b/course_discovery/apps/course_metadata/salesforce.py
@@ -6,6 +6,43 @@ from simple_salesforce import Salesforce, SalesforceExpiredSession
 from course_discovery.apps.core.models import User
 from course_discovery.apps.course_metadata.choices import CourseRunStatus
 
+ORGANIZATION_SALESFORCE_FIELDS = {
+    'organization': ('name', 'key')
+}
+COURSE_SALESFORCE_FIELDS = {
+    'course': ('title', 'has_ofac_restrictions', 'key'),
+}
+COURSE_RUN_SALESFORCE_FIELDS = {
+    'course_run': ('start', 'end', 'status', 'title', 'go_live_date', 'key'),
+}
+
+
+def requires_salesforce_update(source_of_edit, instance):
+    from course_discovery.apps.course_metadata.models import Course, CourseRun
+    relative_fields = ORGANIZATION_SALESFORCE_FIELDS
+    if isinstance(instance, Course):
+        relative_fields = COURSE_SALESFORCE_FIELDS
+    elif isinstance(instance, CourseRun):
+        relative_fields = COURSE_RUN_SALESFORCE_FIELDS
+    return any(attr in relative_fields[source_of_edit] and
+               instance.did_change(attr) for
+               attr in instance.__dict__)
+
+
+def populate_official_with_existing_draft(instance, util):
+    from course_discovery.apps.course_metadata.models import Course, CourseRun
+    created = False
+    if not instance.draft_version.salesforce_id:
+        if isinstance(instance, Course):
+            util.create_course(instance.draft_version)
+        elif isinstance(instance, CourseRun):
+            util.create_course_run(instance.draft_version)
+        created = True
+
+    instance.salesforce_id = instance.draft_version.salesforce_id
+    instance.save()
+    return created
+
 
 def salesforce_request_wrapper(method):
     """
@@ -96,10 +133,9 @@ class SalesforceUtil:
     @salesforce_request_wrapper
     def create_publisher_organization(self, organization):
         if not organization.salesforce_id:
-            sf_organization = self.client.Publisher_Organization__c.create({
-                'Organization_Name__c': organization.name,
-                'Organization_Key__c': organization.key,
-            })
+            sf_organization = self.client.Publisher_Organization__c.create(
+                self._build_publisher_organization_payload(organization)
+            )
             organization.salesforce_id = sf_organization.get('id')
             organization.save()
 
@@ -109,18 +145,9 @@ class SalesforceUtil:
             organization = course.authoring_organizations.first()
             if organization and not organization.salesforce_id:
                 self.create_publisher_organization(organization)
-            sf_course = self.client.Course__c.create({
-                'Course_Name__c': course.title,
-                'Link_to_Publisher__c': '{url}/courses/{uuid}'.format(
-                    url=self.partner.publisher_url.strip('/') if self.partner.publisher_url else '', uuid=course.uuid
-                ),
-                'Link_to_Admin_Portal__c': '{url}/admin/course_metadata/course/{id}/change/'.format(
-                    url=self.partner.site.domain.strip('/') if self.partner.site.domain else '', id=course.id
-                ),
-                'OFAC_Review_Decision__c': course.has_ofac_restrictions,
-                'Course_Key__c': course.key,
-                'Publisher_Organization__c': organization.salesforce_id if organization else None,
-            })
+            sf_course = self.client.Course__c.create(
+                self._build_course_payload(course, organization)
+            )
             course.salesforce_id = sf_course.get('id')
             course.save()
 
@@ -129,18 +156,9 @@ class SalesforceUtil:
         if not course_run.salesforce_id:
             if not course_run.course.salesforce_id:
                 self.create_course(course_run.course)
-            sf_course_run = self.client.Course_Run__c.create({
-                'Course__c': course_run.course.salesforce_id,
-                'Link_to_Admin_Portal__c': '{url}/admin/course_metadata/courserun/{id}/change/'.format(
-                    url=self.partner.site.domain.strip('/') if self.partner.site.domain else '', id=course_run.id
-                ),
-                'Course_Start_Date__c': course_run.start.isoformat() if course_run.start else None,
-                'Course_End_Date__c': course_run.end.isoformat() if course_run.end else None,
-                'Publisher_Status__c': self._get_salesforce_equivalent(course_run.status),
-                'Course_Run_Name__c': course_run.title,
-                'Expected_Go_Live_Date__c': course_run.go_live_date.isoformat() if course_run.go_live_date else None,
-                'Course_Number__c': course_run.key,
-            })
+            sf_course_run = self.client.Course_Run__c.create(
+                self._build_course_run_payload(course_run)
+            )
             course_run.salesforce_id = sf_course_run.get('id')
             course_run.save()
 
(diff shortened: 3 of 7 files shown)


course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -294,6 +300,41 @@ def _create_comment_return_body(user, body, course_run_key=None):
...
+                url=self.partner.site.domain.strip('/') if self.partner.site.domain else '', id=course_run.id
+            ),
+            'Course_Start_Date__c': course_run.start.isoformat() if course_run.start else None,
+            'Course_End_Date__c': course_run.end.isoformat() if course_run.end else None,
+            'Publisher_Status__c': self._get_salesforce_equivalent(course_run.status),
+            'Course_Run_Name__c': course_run.title,
+            'Expected_Go_Live_Date__c': course_run.go_live_date.isoformat() if course_run.go_live_date else None,
+            'Course_Number__c': course_run.key,
+        }
+
```
jlajoie, line 378: Just calling this out that once rebasing with prod some of this data may have changed, you'll just need to update to what's out there now

course_discovery/apps/course_metadata/signals.py
```diff
@@ -259,6 +260,48 @@ def ensure_external_key_uniquness__curriculum(sender, instance, **kwargs):  # py
...
+            util.update_course_run(instance)
+
+
+def _get_salesforce_util(partner):
+    """function to determine whether or not the post_save hook for Organizations, Courses,
+        and Course Runs should be triggered"""
+    try:
+        return SalesforceUtil(partner)
+    except SalesforceConfiguration.DoesNotExist:
+        return False
```
jlajoie: nit: Make this return None, you're evaluating for truthiness but in terms of type comparisons a null value makes more sense than either it's a boolean or it's a whole salesforce object

jlajoie (commented):
Looks like a great start, good stuff pulling the object payloads into external methods. Just the one comment about keeping track of some of those changes that may have occurred in master, and also don't forget about this AC ```Changes should be applied only for Official rows as that is what our Go Live date will be based on, not proposed changes```. You'll just want to check that on the updates if there exists a draft field (course/course run) and it isn't a draft, that's the time you'll want to write those changes out to salesforce since we'll only want real "live" data to exist in SF.

course_discovery/apps/api/v1/tests/test_views/test_comments.py
```diff
@@ -1,4 +1,6 @@
+import factory
 import mock
+from django.db.models.signals import post_save
```
Dillon-Dumesnil, line 3: Nit: Add an empty line here to keep the separation in imports.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -176,6 +161,31 @@ def create_comment_for_course_case(self, course, user, body, course_run_key=None
...
+        if organization.salesforce_id:
+            self.client.Publisher_Organization__c.update(
+                organization.salesforce_id, self._build_publisher_organization_payload(organization)
+            )
+
+    @salesforce_request_wrapper
+    def update_course(self, course):
+        """Triggered by the update_salesforce_course signal receiver"""
+        if course.salesforce_id:
+            organization = course.authoring_organizations.first()
```
Dillon-Dumesnil, line 213: I wonder if this should just be written into an `@property` on the Course model. Curious how many times it is used to determine the authoring org. Or maybe one day even turned into a many to one instead of m2m. Not something you need to change, I was just reflecting.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -259,6 +260,50 @@ def ensure_external_key_uniquness__curriculum(sender, instance, **kwargs):  # py
     check_curricula_and_related_programs_for_duplicate_external_key([instance], course_runs)
 
 
+@receiver(post_save, sender=Organization)
+def update_salesforce_organization(instance, created, **kwargs):  # pylint: disable=unused-argument
+    partner = instance.partner
+    util = _get_salesforce_util(partner)
+    if util:
+        if created:
+            util.create_publisher_organization(instance)
```
Dillon-Dumesnil, line 272: Probably not worth it, but I wonder if we could just change this all to an `update_or_create` and get rid of this whole if statement.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -259,6 +260,50 @@ def ensure_external_key_uniquness__curriculum(sender, instance, **kwargs):  # py
...
+    if util:
+        if created and instance.draft:
+            util.create_course(instance)
+        elif not created and not instance.draft:
+            util.update_course(instance)
+
+
+@receiver(post_save, sender=CourseRun)
+def update_salesforce_course_run(instance, created, **kwargs):  # pylint: disable=unused-argument
+    partner = instance.course.partner
```
Dillon-Dumesnil, line 295: Not sure it's really necessary, but a helper function to handle this stuff could be useful and potentially cleaner. It just raises a bit of a flag in my mind seeing how similar these three functions are and I'm curious how you could abstract it. Potentially only the first three lines, but might be an improvement.
```
    partner = instance.partner
    util = _get_salesforce_util(partner)
    if util:
```
I'm not diehard on it. Just might be nice

course_discovery/apps/course_metadata/signals.py
```diff
@@ -259,6 +260,50 @@ def ensure_external_key_uniquness__curriculum(sender, instance, **kwargs):  # py
     check_curricula_and_related_programs_for_duplicate_external_key([instance], course_runs)
 
 
+@receiver(post_save, sender=Organization)
+def update_salesforce_organization(instance, created, **kwargs):  # pylint: disable=unused-argument
```
Dillon-Dumesnil: Not sure if the naming of these functions has to be specific, but I would prefer all three of these to be `update_or_create_salesforce_xxxxx` since it is technically doing either.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -259,6 +260,50 @@ def ensure_external_key_uniquness__curriculum(sender, instance, **kwargs):  # py
...
+    if util:
+        if created and instance.draft:
+            util.create_course_run(instance)
+        elif not created and not instance.draft:
+            util.update_course_run(instance)
+
+
+def _get_salesforce_util(partner):
+    """
+    Function to determine whether or not the post_save hook for Organization, Course,
```
Dillon-Dumesnil: I think this docstring should be rewritten and potentially not even exist here. This function is trying to return to you a SalesforceUtil. You're writing the docstring about what the return value is going to be used for _by you_ and not what the function is actually doing. On that point, that's why I think this might not even best live here and potentially live as a helper function next to the Util itself and this file would just import the helper. Something to think about.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -124,6 +125,12 @@ def set_draft_state(obj, model, attrs=None):
 
     obj.save()
 
+    # If a course or course run is being saved, we refresh the instance from the db and set
```
Dillon-Dumesnil: I think this comment can be improved. Don't use the comment to explain what is happening in the code, that's obvious. Explain _why_ we need to do this. Mention the local object doesn't have salesforce_id so we have to refresh the db since it was saved there in a post_save hook. Give the why, not the what.

course_discovery/apps/course_metadata/tests/test_salesforce.py
```diff
@@ -7,10 +10,13 @@
 from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseRunFactory, OrganizationFactory
 
 
+@ddt.ddt
 class TestSalesforce(TestCase):
     def setUp(self):
         super(TestSalesforce, self).setUp()
         self.salesforce_config = SalesforceConfigurationFactory()
+        self.salesforce_path = 'course_discovery.apps.course_metadata.salesforce.Salesforce'
```
Dillon-Dumesnil, line 22: Nice job, I like these improvements.

course_discovery/apps/course_metadata/tests/test_salesforce.py
```diff
@@ -455,3 +450,45 @@ def test_format_and_parse(self):
 
             self.assertEqual(parsed_message.get('course_run_key'), course_run_key)
             self.assertEqual(parsed_message.get('comment'), body)
+
+    @ddt.data('test-id', None)
+    def test_update_publisher_organization(self, salesforce_id):
```
Dillon-Dumesnil, line 459: Is it possible to check any data changing with these three functions instead of just that the function is called? Probably not, just thought I'd ask

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -792,3 +792,54 @@ def test_collision_does_not_include_drafts(self):
...
+
+    def test_update_salesforce_course(self):
+        #  we mock the __init__ method in SalesforceUtil to ignore login functionality
+        with mock.patch(self.salesforce_util_path + '.__init__', return_value=None):
+            with mock.patch(self.salesforce_util_path + '.create_course') as mock_create_method:
+                with mock.patch(self.salesforce_util_path + '.update_course') as mock_update_method:
+                    course = factories.CourseFactory(draft=True)
+
+                    mock_create_method.assert_called()
+
```
Dillon-Dumesnil, line 831: Maybe just also do a save here while the Course is still draft to illustrate that drafts do not call the update method.

Dillon-Dumesnil (commented):
We're getting there! Couple comments. Let me know if you have any questions.

course_discovery/apps/course_metadata/tests/test_salesforce.py
xnick421: we would have to define the mock methods and have them return values

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -164,6 +145,10 @@ def create_case_for_course(self, course):
             sf_case = self.client.Case.create(case)
             course.salesforce_case_id = sf_case.get('id')
             course.save()
+            if course.official_version and not course.official_version.salesforce_case_id:
```
jlajoie, line 185: This only ever runs against draft versions, correct?

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -109,18 +108,9 @@ def create_course(self, course):
...
-                    url=self.partner.site.domain.strip('/') if self.partner.site.domain else '', id=course.id
-                ),
-                'OFAC_Review_Decision__c': course.has_ofac_restrictions,
-                'Course_Key__c': course.key,
-                'Publisher_Organization__c': organization.salesforce_id if organization else None,
-            })
+            sf_course = self.client.Course__c.create(
+                self._build_course_payload(course, organization)
+            )
             course.salesforce_id = sf_course.get('id')
```
jlajoie, line 151: Should we be trying to set the official version salesforce_id here as well?

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -129,18 +119,9 @@ def create_course_run(self, course_run):
...
-                'Course_End_Date__c': course_run.end.isoformat() if course_run.end else None,
-                'Publisher_Status__c': self._get_salesforce_equivalent(course_run.status),
-                'Course_Run_Name__c': course_run.title,
-                'Expected_Go_Live_Date__c': course_run.go_live_date.isoformat() if course_run.go_live_date else None,
-                'Course_Number__c': course_run.key,
-            })
+            sf_course_run = self.client.Course_Run__c.create(
+                self._build_course_run_payload(course_run)
+            )
             course_run.salesforce_id = sf_course_run.get('id')
```
jlajoie, line 162: Should we be trying to set the official version salesforce_id here as well? (for runs this time)

course_discovery/apps/course_metadata/signals.py
jlajoie: They're definitely all similar, but the way we grab the partner from each model varies and the subsequent method calls seemed distinct enough that it feels cleaner to me to just keep them separate. It's like, 3 lines of boiler plate and they all live together so it shouldn't be missed.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -124,6 +127,13 @@ def set_draft_state(obj, model, attrs=None):
 
     obj.save()
 
+    # We refresh the objects instance before we set it's salesforce_id because the instance in memory is
```
jlajoie: tiniest nit: objects -> object's

jlajoie: double tiny nit: its is possessive, it's is a contraction of it is, "we set it is salesforce_id"

course_discovery/apps/course_metadata/utils.py
```diff
@@ -683,3 +693,10 @@ def headers(self):
             'Content-Type': 'application/json',
             'X-CSRF-Token': self.csrf_token,
         }
+
+
+def get_salesforce_util(partner):
+    try:
+        return SalesforceUtil(partner)
+    except SalesforceConfiguration.DoesNotExist:
+        return None
```
jlajoie, line 702: 👍

jlajoie: nit: this method might be worth migrating to the get_util_or_404 method we have setup in the comments API? Just use this and instead of catching the exception in there just return the 404 if it's null? Don't care but seems like it could be worthwhile cleanup

jlajoie (changes requested):
Couple of nits, looks like a solid update and it looks like most of the headache was around what objects had the salesforce_id in memory that we operated on, vs where we had fetched and loaded course/course_runs into memory that were not operated upon in that salesforce object lifecycle. Makes sense!

course_discovery/apps/course_metadata/salesforce.py
xnick421: yea, but it may not have an official version yet, which is why we check.  so this logic is for the case where an official version is made without any previous comments, which means both draft and official have no case_id.  then the next time a comment comes in, it creates the case, the draft gets the id, and we copy over the id in this if block

course_discovery/apps/course_metadata/salesforce.py
xnick421: we do it in the set_draft_state util method.  we refresh the in memory instance there to reflect whats in the db.  we can talk more about this

course_discovery/apps/course_metadata/signals.py
```diff
@@ -281,6 +281,13 @@ def update_or_create_salesforce_course(instance, created, **kwargs):  # pylint:
         if created and instance.draft:
             util.create_course(instance)
         elif not created and not instance.draft:
+            if not instance.salesforce_id:
```
jlajoie: I think you'll need the else i there is a salesforce_id to just do the update like normal

course_discovery/apps/course_metadata/signals.py
```diff
@@ -293,6 +300,15 @@ def update_or_create_salesforce_course_run(instance, created, **kwargs):  # pyli
         if created and instance.draft:
             util.create_course_run(instance)
         elif not created and not instance.draft:
+            if not instance.salesforce_id:
```
jlajoie: Same here, if there is a salesforce_id just do the update

jlajoie (changes requested):
Just the one change, we'll need to handle the case of "hey things are okay this has the proper IDs set"

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -6,6 +6,52 @@
...
+    elif instance.__class__.__name__ == 'CourseRun':
+        relative_fields = COURSE_RUN_SALESFORCE_FIELDS
+    return any(attr in relative_fields[source_of_edit] and
+               getattr(instance, attr) != getattr(instance, '_old_' + attr) for
+               attr in instance.__dict__)
+
+
+def populate_official_with_existing_draft(instance, util):
+    if not instance.draft_version.salesforce_id:
+        if instance.__class__.__name__ == 'Course':
```
jlajoie: Might be best to change this to an isinstance check so in case we refactor class names (probably won't ever happen) this code won't break?

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -6,6 +6,52 @@
...
+            util.create_course(instance.draft_version)
+        elif instance.__class__.__name__ == 'CourseRun':
+            util.create_course_run(instance.draft_version)
+
+    instance.salesforce_id = instance.draft_version.salesforce_id
```
jlajoie, line 42: Might be worth even having an extra check that makes sure there is a draft_version.salesforce_id set? As it is now there are branches where you can start with no salesforce_id and can end up with no code being run other than ```instance.salesforce_id = instance.draft_version.salesforce_id```

course_discovery/apps/course_metadata/models.py
```diff
@@ -84,6 +84,25 @@ class Meta(object):
...
 
+class CachedMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._cache = dict(self.__dict__)
+
+    def refresh_from_db(self, using=None, fields=None, **kwargs):
+        super().refresh_from_db(using, fields, **kwargs)
+        if self._cache and self._cache['_cache']:
+            self._cache.pop('_cache')
```
mikix: This wouldn't stop it from getting in. You're popping it from the cache var before you overwrite it. Maybe move this block down after setting _cache?

You probably want this logic in `save` too?

course_discovery/apps/course_metadata/models.py
```diff
@@ -84,6 +84,25 @@ class Meta(object):
         abstract = True
 
 
+class CachedMixin:
```
jlajoie, line 87: Not even at the level of a nit, but I can't decide if Cached vs Cache is better,

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -6,6 +6,57 @@
...
+    #    for both course and site
+    # 3. wire up the update_or_create_salesforce_course_run to listen to the Course and Site models
+    'course': 'salesforce_id',
+    'site': 'domain',
+}
+
+
+def requires_salesforce_update(source_of_edit, instance):
+    relative_fields = ORGANIZATION_SALESFORCE_FIELDS
+    name = instance.__class__.__name__
```
jlajoie: I think we talked about maybe swapping this to an instanceof check with a localized import, did that not work out? If it didn't that's cool just want to double check

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -6,6 +6,57 @@
...
+    elif name == 'CourseRun':
+        relative_fields = COURSE_RUN_SALESFORCE_FIELDS
+    return any(attr in relative_fields[source_of_edit] and
+               instance.did_change(attr) for
+               attr in instance.__dict__)
+
+
+def populate_official_with_existing_draft(instance, util):
+    created = False
+    name = instance.__class__.__name__
```
jlajoie: Same as above with instanceof check

jlajoie (approved):
Just minor comments but looks good to go, tested locally as well

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -6,6 +6,57 @@
...
+    'organization': ('name', 'key')
+}
+COURSE_SALESFORCE_FIELDS = {
+    'course': ('title', 'has_ofac_restrictions', 'key'),
+    # TODO partner and organization values should be implemented
+    # 1. cache these fields in the constructor of Partner and Organization like we do for Course
+    # 2. add the below requires_salesforce_update method in the update_or_create_salesforce_course
+    #    for both partner and organization
+    # 3. wire up the update_or_create_salesforce_course to listen to the Partner and Organization models
+    'partner': ('publisher_url', 'domain'),
```
jlajoie: I think we can drop partner and organization here

PR #2145, opened 2019-09-06
Fix publisher migration to get discovery counterparts through course runs

rgraber:


dianakhuang (approved):

course_discovery/apps/publisher/migrations/0081_initialize_course_url_slug.py
```diff
@@ -10,22 +10,33 @@
 class Migration(migrations.Migration):
 
     def migrate_data_forward(apps, schema_editor):
+        DiscoveryCourseRun = apps.get_model('course_metadata', 'courseRun')
+
+        # need to recreate generated properties since the model managers aren't serialized
+        def get_discovery_course_run(course_run_instance):
+            if course_run_instance.lms_course_id:
+                return DiscoveryCourseRun.everything.filter(draft=False, key=course_run_instance.lms_course_id).first()
```
mikix, line 18: You _could_ just do `objects.filter` and skip the draft=False bit.

course_discovery/apps/publisher/migrations/0081_initialize_course_url_slug.py
```diff
@@ -10,22 +10,33 @@
...
+            else:
+                return None
+
+        # find the course metadata counterpart by finding the counterpart of the most recent course run and extracting
+        # the course object
+        def get_discovery_course(course_instance):
+            for course_run in course_instance.publisher_course_runs.all():
```
mikix, line 25: The code you're recreating also does `.order_by('-created')` which you might want - you probably want to prefer a link from the newest course rather than the oldest. Though... looking at the default ordering for `TimeStampedModel`, you'd get `('-modified', '-created',)` which is close enough...

mikix (approved):
Seems reasonable... Are you gonna conflict with any existing url_slugs as you go? Like, if the last course in the set has slug B right now for some reason, but the first course you iterate over wants to set B as the slug... you'll hit a problem.

Or did the previous migrations already wipe out the content?

course_discovery/apps/publisher/migrations/0081_initialize_course_url_slug.py
rgraber: objects doesn't work in migrations, annoyingly, because of the special manager

PR #2146, opened 2019-09-09
Revert "Fix publisher migration to get discovery counterparts through course runs"

rgraber:
Reverts edx/course-discovery#2145


dianakhuang (approved):
Looks reasonable to me.

PR #2147, opened 2019-09-09
Reconnect to Django in RCM

mikix:
Right at the end of the RCM main loop, when we are performing cleanup, try to reconnect to Django.

This will hopefully fix the recent consistent RCM failure.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
+++ b/course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
@@ -208,6 +208,8 @@ class Command(BaseCommand):
 
             # TODO Cleanup CourseRun overrides equivalent to the Course values.
 
+        connection.connect()  # reconnect to django outside of loop (see connect comment above)
+
         # Clean up any media orphans that we might have created
         delete_orphans(Image)
         delete_orphans(Video)


course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -208,6 +208,8 @@ def handle(self, *args, **options):
 
             # TODO Cleanup CourseRun overrides equivalent to the Course values.
 
+        connection.connect()  # reconnect to django outside of loop (see connect comment above)
```
mikix, line 211: Here is the referenced comment up above:

```
            # The Linux kernel implements copy-on-write when fork() is called to create a new
            # process. Pages that the parent and child processes share, such as the database
            # connection, are marked read-only. If a write is performed on a read-only page
            # (e.g., closing the connection), it is then copied, since the memory is no longer
            # identical between the two processes. This leads to the following behavior:
            #
            # 1) Newly forked process
            #       parent
            #              -> connection (Django open, MySQL open)
            #       child
            #
            # 2) Child process closes the connection
            #       parent -> connection (*Django open, MySQL closed*)
            #       child  -> connection (Django closed, MySQL closed)
            #
            # Calling connection.close() from a child process causes the MySQL server to
            # close a connection which the parent process thinks is still usable. Since
            # the parent process thinks the connection is still open, Django won't attempt
            # to open a new one, and the parent ends up running a query on a closed connection.
            # This results in a 'MySQL server has gone away' error.
            #
            # To resolve this, we force Django to reconnect to the database before running any queries.
```

Dillon-Dumesnil (approved):
Should be able to test on stage at least.

PR #2149, opened 2019-09-10
ensure slug_uniqueness with better migration

rgraber:


course_discovery/apps/course_metadata/migrations/0196_auto_20190910_1714.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
rgraber, line 1: This is necessary because while 0194 does the same thing, 0195 reverts 0194 and is already in master.

course_discovery/apps/publisher/migrations/0081_initialize_course_url_slug.py
```diff
@@ -2,30 +2,39 @@
...
-from django.core.exceptions import ObjectDoesNotExist
 from django.db import migrations
 import django_extensions.db.fields
 
 
 class Migration(migrations.Migration):
 
     def migrate_data_forward(apps, schema_editor):
-        Course = apps.get_model('publisher', 'course')
+        updated_pub_course_ids = []
```
rgraber, line 12: This is the only actual change from the previous PR. To get rid of the problem where multiple courses in publisher share the same counterpart in course_metadata, start from course_metadata and find a discovery counterpart going the opposite direction. There's a risk we will update the "wrong" publisher course, but it's unlikely because we are finding the publisher course based on the most recent course run in course_metadata.

course_discovery/apps/publisher/migrations/0083_publisher_course_unique_url_slug.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
rgraber, line 1: Also necessary because the reversion PR negated the uniqueness constraint

dianakhuang (approved):
🤞

course_discovery/apps/publisher/migrations/0081_initialize_course_url_slug.py
```diff
@@ -2,30 +2,39 @@
...
         DiscoveryCourse = apps.get_model('course_metadata', 'course')
-        for instance in Course.objects.all():
+        Course = apps.get_model('publisher', 'course')
+        CourseRun = apps.get_model('publisher', 'courseRun')
+
+        def get_publisher_course(discovery_course):
+            for discovery_course_run in discovery_course.course_runs.order_by('-created'):
+                if not discovery_course_run.draft:
```
mikix, line 19: This line *shouldn't* be needed (official course shouldn't have draft runs), but it's harmless, so no need to change.

mikix (approved):
Seems good? The double-loop approach of first syncing, then resetting the rest seems good.

PR #2151, opened 2019-09-10
Pull pacing_type from Studio

mikix:
When an org is on the new publisher, pull it in from Studio
and stop pushing it on updates. (i.e. move source-of-edit to
Studio for users of frontend-app-publisher)

https://openedx.atlassian.net/browse/DISCO-1341

This is related to https://github.com/edx/frontend-app-publisher/pull/258

### DIFF
--- a/course_discovery/apps/api/utils.py
+++ b/course_discovery/apps/api/utils.py
@@ -154,6 +154,7 @@ class StudioAPI:
         editors = cls._run_editors(course_run)
         org, number, run = cls._run_key_parts(course_run)
         start, end = cls._run_times(course_run, creating)
+        pacing = cls._run_pacing(course_run, creating)
 
         if user:
             editors.append(user)
@@ -177,8 +178,10 @@ class StudioAPI:
             'number': number,
             'run': run,
             'team': team,
-            'pacing_type': cls._run_pacing(course_run),
         }
+
+        if pacing:
+            data['pacing_type'] = pacing
 
         if start or end:
             data['schedule'] = {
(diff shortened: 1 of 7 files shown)


mikix:
Note: this shouldn't land until the frontend work is done as well.

course_discovery/apps/api/utils.py
```diff
@@ -177,9 +178,11 @@ def generate_data_for_studio_api(cls, course_run, creating, user=None):
             'number': number,
             'run': run,
             'team': team,
-            'pacing_type': cls._run_pacing(course_run),
         }
 
+        if pacing:
```
Dillon-Dumesnil, line 183: I don't love all of these ifs, but I don't have a better solution.

course_discovery/apps/publisher/utils.py
```diff
@@ -180,6 +180,26 @@ def is_on_new_pub_fe(user):
...
 
+def is_course_on_new_pub_fe(course):
+    """
+    Returns True if all the course's organizations have been moved to new publisher frontend
+
+    Args:
+        course: A course_metadata Course object (not a Publisher one)
+    """
+    try:
+        orgs_on_new_pub_fe = frozenset(settings.ORGS_ON_NEW_PUB_FE.split(','))
```
Dillon-Dumesnil: Is there not a cleaner way to do this without a try except? Like any sort of `getattr(settings, 'ORGS_ON_NEW_PUB_FE')` or something? Feels heavy handed to be waiting for an AttributeError when checking a setting.

Dillon-Dumesnil (approved):
Sure

course_discovery/apps/publisher/utils.py
mikix: Fair. I stole this format from `is_on_new_pub_fe`, because I was lazy. Getattr is better. I'll do that.

mikix: Done. And fixed the other function to follow suit.

PR #2152, opened 2019-09-11
Remove admin links in notification emails

jmyatt:


Dillon-Dumesnil (approved):
Looks fine. I might recommend having Mike Terry take a look just since he has a little more context on the emails and can better confirm if these are all the spots.

mikix (approved):
You might be able to also drop calculating/passing-in the `course_approval_url` template variable now, if it's no longer used.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -112,7 +112,6 @@ def send_email(course_run, template_name, subject, to_users, recipient_name, con
         'org_name': org.name,
         'contact_us_email': project_coordinator.email,
         'course_page_url': review_url,
-        'course_approval_url': approve_url,
```
mikix, line 115: You can also drop the calculation of approve_url above this.

jmyatt: d'oh!  thanks

PR #2154, opened 2019-09-12
Unpublish drafts along with old runs

mikix:
When the update_marketing_redirects command runs, it was only
unpublishing official rows of old inactive runs. This changes it
to also unpublish the draft rows, just to keep those in sync.


jlajoie (approved):
👍

PR #2155, opened 2019-09-12
Don't send double emails on publish

mikix:
We were accidentally sending two publish emails when publishing
from old publisher. This is because we now make drafts of rows
even when working in old publisher and the new email flow was
only checking if drafts existed.

https://openedx.atlassian.net/browse/DISCO-1357

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1320,7 +1320,7 @@ class CourseRun(DraftModelMixin, TimeStampedModel):
 
         return official_version
 
-    def handle_status_change(self):
+    def handle_status_change(self, send_emails):
         """
         If a row's status changed, take any cleanup actions necessary.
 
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1345,15 +1347,19 @@ def handle_status_change(self):
...
             else:  # The publish status check will send an email for go-live
-                emails.send_email_for_reviewed(self)
+                email_method = emails.send_email_for_reviewed
 
         elif self.status == CourseRunStatus.Published:
-            emails.send_email_for_go_live(self)
+            email_method = emails.send_email_for_go_live
+
+        if send_emails and email_method:
+            email_method(self)
```
jlajoie, line 1361: First class functions 🤗

course_discovery/apps/publisher/tests/test_models.py
```diff
@@ -896,17 +902,25 @@ def test_change_state(self, state):
         self.assertEqual(self.course_run_state.name, state)
 
     @responses.activate
+    @override_switch('enable_publisher_email_notifications', True)
     def test_published(self):
         person = PersonFactory()
+        org = OrganizationFactory()
```
jlajoie, line 908: Why the addition of Org info here? It doesn't look like anything above uses it except maybe as part of the e-mail template? Or I guess might be for just making sure the roles are set properly to have a more natural trigger of changing state?

jlajoie (approved):
One question about orgs but I think I figured it out, lgtm 👍

course_discovery/apps/publisher/tests/test_models.py
mikix: Well, that's tricky. It's because the new publisher email flow wants an org before it will send out an email. So to set up everything so that the new email flow **would** trigger if I didn't fix this bug, I made an org.

But it's brittle, because, like, I'm also fixing this bug. So if the new email flow also gained a requirement that, say, the course needs staff too to send a new email... this test wouldn't notice that and we wouldn't be fully testing what I want (that new emails don't get sent because of my fix, not because of things not being hooked up right). But that's the problem with testing a negative...

So. I dunno. I did what I could.

PR #2161, opened 2019-09-17
Add the necessary signal connect and disconnect to fix issues while saving the fixtures

schenedx:
This is a PR to fix a bug, where the program fixture loading management command was broken when new signals were introduced.
See the proof of fix at https://tools-edx-jenkins.edx.org/job/Sandboxes/job/UpdateMastersSandbox/

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
+++ b/course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
@@ -13,10 +13,12 @@ from edx_rest_api_client import client as rest_client
 
 from course_discovery.apps.core.models import Partner
 from course_discovery.apps.course_metadata.models import (
-    CourseRun, Curriculum, CurriculumCourseMembership, Program, ProgramType, Seat, SeatType
+    CourseRun, Curriculum, CurriculumCourseMembership, CurriculumProgramMembership, Program, ProgramType, Seat,
+    SeatType
 )
 from course_discovery.apps.course_metadata.signals import (
-    add_masters_track_on_course, ensure_external_key_uniquness__course_run, ensure_external_key_uniquness__curriculum,
+    add_masters_track_on_course, check_curriculum_for_cycles, check_curriculum_program_membership_for_cycles,
+    ensure_external_key_uniquness__course_run, ensure_external_key_uniquness__curriculum,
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -65,6 +75,12 @@ def disconnect_program_signals():
             publish_masters_track, sender=Seat,
         )
 
+        pre_save.connect(
+            check_curriculum_for_cycles, sender=Curriculum,
+        )
+        pre_save.connect(
+            check_curriculum_program_membership_for_cycles, sender=CurriculumProgramMembership,
+        )
         pre_save.connect(
```
kdmccormick: Why not have a list of tuples in the form `(signal_name, sender_model_class)` and then iterate through that to disconnect and re-connect? That way you know you that you re-connect everything you disconnected.

kdmccormick (commented):
Were you able to get that test working to ensure that new signals cause a test failure?

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
schenedx: Good suggestion for optimization. I will try that.

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -32,48 +34,61 @@ def disconnect_program_signals():
...
-    pre_save.disconnect(
-        ensure_external_key_uniquness__course_run, sender=CourseRun,
-    )
-    pre_save.disconnect(
-        ensure_external_key_uniquness__curriculum, sender=Curriculum,
-    )
-    pre_save.disconnect(
-        ensure_external_key_uniquness__curriculum_course_membership, sender=CurriculumCourseMembership,
-    )
+    signals_list = [
```
kdmccormick, line 46: Nice!

zacharis278 (approved):
this lgtm

schenedx:
@kdmccormick I got the test to fail before the change. The key is to have fixture file include `CurriculumProgramMembership` before `Curriculum`. This way, the signal would cause the failure during fixture file save. Once this build is green, I will merge.

PR #2163, opened 2019-09-17
update salesforce

xnick421:
handle cases where drafts already exist with no officials and officials already exist with no drafts before salesforce was implemented

### DIFF
--- a/course_discovery/apps/course_metadata/salesforce.py
+++ b/course_discovery/apps/course_metadata/salesforce.py
@@ -214,6 +214,8 @@ class SalesforceUtil:
             self.client.Course__c.update(
                 course.salesforce_id, self._build_course_payload(course, organization)
             )
+        else:
+            self.create_course(course)
 
     @salesforce_request_wrapper
     def update_course_run(self, course_run):
@@ -222,6 +224,8 @@ class SalesforceUtil:
             self.client.Course_Run__c.update(
                 course_run.salesforce_id, self._build_course_run_payload(course_run)
             )
+        else:
+            self.create_course_run(course_run)
 
     @staticmethod
     def format_user_comment_body(user, body, course_run_key=None):
--- a/course_discovery/apps/course_metadata/tests/test_signals.py
+++ b/course_discovery/apps/course_metadata/tests/test_signals.py
@@ -13,6 +13,7 @@ from waffle.testutils import override_switch
 
 from course_discovery.apps.api.v1.tests.test_views.mixins import FuzzyInt
 from course_discovery.apps.core.models import Currency
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
 from course_discovery.apps.course_metadata.models import (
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -848,13 +849,9 @@ def test_update_or_create_salesforce_course_run(self):
...
                             mock_update_method.assert_not_called()
 
-                            course_run.save()
-
-                            # This shows that an update to a draft does not hit the salesforce update method
-                            mock_update_method.assert_not_called()
-
                             course_run.draft = False
-                            course_run.name = 'changed'
+                            course_run.status = CourseRunStatus.Unpublished
```
Dillon-Dumesnil, line 853: Nit: If you're going to be setting status here, add a different status to the Factory call so it is clear this is actually being changed.

Factory does Published by default, but it's just an extra step for anyone looking at this.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -848,13 +849,9 @@ def test_update_or_create_salesforce_course_run(self):
...
                             course_run.save()
+                            print(course_run.__dict__)
```
Dillon-Dumesnil: Definitely remove

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -205,6 +205,8 @@ def update_publisher_organization(self, organization):
             self.client.Publisher_Organization__c.update(
                 organization.salesforce_id, self._build_publisher_organization_payload(organization)
             )
+        else:
```
Dillon-Dumesnil: Not needed because you always call this before calling this function. See `update_or_create_salesforce_organization` in course_metadata.signals.py

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -214,6 +216,8 @@ def update_course(self, course):
             self.client.Course__c.update(
                 course.salesforce_id, self._build_course_payload(course, organization)
             )
+        else:
+            self.create_course(course)
```
Dillon-Dumesnil, line 218: Rename to `update_or_create_course` for clarity on what it actually does.

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -222,6 +226,8 @@ def update_course_run(self, course_run):
             self.client.Course_Run__c.update(
                 course_run.salesforce_id, self._build_course_run_payload(course_run)
             )
+        else:
+            self.create_course_run(course_run)
```
Dillon-Dumesnil, line 228: Rename to `update_or_create_courserun` for clarity on what it actually does.

Dillon-Dumesnil (approved):
Couple of changes, but otherwise looks good

PR #2164, opened 2019-09-17
ENT-2312 | Adding prefetch_related to for loop to reduce DB queries for seat data

christopappas:
In the `CourseSearchSerializer`, there are two SerializerMethodFields. Within each is a for-loop that iterates over `result.object.course_runs.all()` (I actually didn't notice the 2nd loop until after committed a fix for the first loop!). When the serializer goes to create the representation for the `SearchResult` object (that's what the `instance` is in the `to_representation` call in this case), it calls both SerializerMethodField code separately. There is no query_set prefetch logic happening for `CourseSerializer` (we do a handful of others in search.py), so for each Course we put through the serializer, we were doing a lookup for all of its course_runs in order... and each of those course_runs' seats.

By running `prefetch_related_objects` in the `to_representation` method before those `SerializerMethodField` fields are called, we're able to avoid an abundance of DB queries. See below the difference in query count with my local setup that returns 534 results.

Locally, `search/all/` before:

```
138 queries in 88.16ms
```

Locally, `search/all/` after:

```
39 queries in 29.10ms
```

New relic corroborates the existence of many, many Seat queries:

Screen Shot 2019-09-17 at 3 21 47 PM

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -11,7 +11,7 @@ import waffle
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models.query import Prefetch
+from django.db.models.query import Prefetch, prefetch_related_objects
 from django.utils.text import slugify
 from django.utils.translation import ugettext_lazy as _
 from drf_dynamic_fields import DynamicFieldsMixin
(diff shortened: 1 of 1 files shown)


mikix (approved):
Sure. Nice!

dianakhuang (approved):

christopappas:
@mikix @dianakhuang Thanks for the quick review before! Re-requested review because I changed my approach. Dropped info in the description above, but, doing prefetching in the `to_representation` layer instead of in the SerailizerMethodFields themselves

mikix (approved):
That is not an approach I've seen elsewhere, but if it works, alright.

dianakhuang (approved):
Ditto to @mikix

PR #2166, opened 2019-09-18
DISCO-1366: Adds in max_retry logic for Salesforce login

jlajoie:
Enable retries for the 
```
requests.exceptions.ConnectionError: ('Connection aborted.', OSError("(104, 'ECONNRESET')",))
``` 
error.  (based on https://github.com/psf/requests/issues/4506)

For testing purposes I updated a Course Run on both my local and Stage which would send a request to Salesforce. After ~10 minutes of waiting, attempting to save and trigger a Salesforce write on Stage fails with the above error, while locally the request succeeds.

Alternatively, I could catch the OSError as part of the `salesforce_request_wrapper`, but that felt more harmful than enabling the max_retries.

### DIFF
--- a/course_discovery/apps/course_metadata/salesforce.py
+++ b/course_discovery/apps/course_metadata/salesforce.py
@@ -1,6 +1,8 @@
 import re
 from datetime import datetime, timezone
 
+import requests
+from requests.adapters import HTTPAdapter
 from simple_salesforce import Salesforce, SalesforceExpiredSession
 
 from course_discovery.apps.core.models import User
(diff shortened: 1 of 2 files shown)


jlajoie:
Also sneaking in the DOTALL flag here as part of https://openedx.atlassian.net/browse/DISCO-1367

course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -93,6 +95,11 @@ def salesforce_is_enabled(self):
             return bool(self.partner.salesforce)
 
         def login(self):
+            # Need to instantiate a session with multiple retries to avoid OSError
+            session = requests.Session()
+            adapter = HTTPAdapter(max_retries=2)
+            session.mount('https://', adapter)
```
mikix, line 101: I trust that this is the briefest way to write this, but feels weird that you have to be so lowlevel with it. (mucking with scheme mappings just for a retry)

mikix (approved):
Sure, love it.

course_discovery/apps/course_metadata/salesforce.py
jlajoie: Yeah, also felt weird that this was needed, dug through simple-salesforce to double check but didn't see anything. Only error I found was the one listed in the ticket, and this was the recommended approach to enable retries.

PR #2167, opened 2019-09-18
DISCO-1364: fix bug about always setting verified deadline

Dillon-Dumesnil:
Also found a different bug where on course creation, if there is
an error with the course run creation that is not caught and
raised and instead caught through our general exception catch,
it creates a response which the course endpoint then drops.

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -234,7 +234,9 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
                 'price': price,
                 'mode': course_creation_fields['mode'],
             })
-            CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
+            run_response = CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
+            if run_response.status_code != 201:
+                raise Exception(str(run_response.data))
 
         headers = self.get_success_headers(serializer.data)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1066,13 +1066,9 @@ class CourseRun(DraftModelMixin, CachedMixin, TimeStampedModel):
         # An unenrolled user may not enroll and purchase paid seats after the course or enrollment has ended.
         deadline = self.enrollment_deadline
 
-        # Note that even though we're sorting in descending order by upgrade_deadline, we will need to look at
-        # both the first and last record in the result set to determine which Seat has the latest upgrade_deadline.
-        # We consider Null values to be > than non-Null values, and Null values may sort to the top or bottom of
-        # the result set, depending on the DB backend.
-        latest_seat = seats[-1] if seats[-1].upgrade_deadline is None else seats[0]
-        if latest_seat.upgrade_deadline and (deadline is None or latest_seat.upgrade_deadline < deadline):
-            deadline = latest_seat.upgrade_deadline
+        seat = seats[0]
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1289,8 +1289,12 @@ def update_or_create_seats(self):
         deadline = None
         # only verified seats have a deadline specified
         if mode == Seat.VERIFIED:
-            deadline = self.end - datetime.timedelta(days=settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
-            deadline = deadline.replace(hour=23, minute=59, second=59, microsecond=99999)
+            verified_seat = Seat.everything.filter(course_run=self, type=Seat.VERIFIED).first()
```
mikix: `Everything`? We don't want `filter_drafts`? The unique_together for seats is madness.

Rest of code in file seems to use `_enrollable_paid_seats` to grab seats that might have upgrade deadline? Not just check verified.  Like, should this code be using `get_paid_seat_enrollment_end` or at least similar complicated logic?

Also, `type=mode` probably makes some more sense here.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -234,7 +234,9 @@ def create(self, request, *args, **kwargs):
                 'price': price,
                 'mode': course_creation_fields['mode'],
             })
-            CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
+            run_response = CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
+            if run_response.status_code != 201:
+                raise Exception(str(run_response.data))
```
mikix, line 239: I think you can just do `response.raise_for_status()` instead of checking status code?

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: I realized I forgot to add in `draft=True`. As of right now, it can only be called through the courses or course_runs endpoint which always deal with drafts. Although I guess I could make it agnostic by using `filter_drafts`. I'm not opposed.

We talked about this and did realize that verified seats are the only ones that make sense to have an upgrade deadline right now (since you can't upgrade to professional. You either are in it or you are not). I'll look deeper into that logic to see if it still makes sense to switch, but otherwise I might keep it reflecting what was in old pub

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: Unfortunately not :(
```
run_response.raise_for_status()
AttributeError: 'Response' object has no attribute 'raise_for_status'
```
`raise_for_status()` comes from using `requests` and the object you get back from doing `requests.get(...)`, but that response is a `Response` object we create from the rest_framework library

course_discovery/apps/course_metadata/models.py
```diff
@@ -1066,13 +1066,9 @@ def get_paid_seat_enrollment_end(self):
...
 
-        # Note that even though we're sorting in descending order by upgrade_deadline, we will need to look at
-        # both the first and last record in the result set to determine which Seat has the latest upgrade_deadline.
-        # We consider Null values to be > than non-Null values, and Null values may sort to the top or bottom of
-        # the result set, depending on the DB backend.
-        latest_seat = seats[-1] if seats[-1].upgrade_deadline is None else seats[0]
-        if latest_seat.upgrade_deadline and (deadline is None or latest_seat.upgrade_deadline < deadline):
-            deadline = latest_seat.upgrade_deadline
+        seat = seats[0]
+        if deadline is None or seat.upgrade_deadline < deadline:
```
mikix: I don't think you want to drop the `seat.upgrade_deadline and` bit here.

The first seat returned is very likely to have a None upgrade_deadline. (it gets sorted at front of list)

mikix (approved):
Nice! Left one comment.

PR #2169, opened 2019-09-19
DISCO-1368: Removes _modified to set CreatedDate in FeedItems

jlajoie:


course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -78,7 +78,8 @@ def handle(self, *args, **options):
                         comments.append({
                             'ParentId': course.salesforce_case_id,
                             'Body': user_comment_body,
-                            '_modified': comment.modified,  # To be removed below, just for ordering
+                            # Use modified for CreatedAt as that is the time this exact comment began to exist
```
Dillon-Dumesnil: Nit: Fix the comment to use `CreatedDate` instead of `CreatedAt`.

Weird there isn't a `comment.created` you could use.

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
```diff
@@ -91,11 +92,10 @@ def handle(self, *args, **options):
...
```
Dillon-Dumesnil: Nit: Fix the comment to use `CreatedDate` instead of `CreatedAt`.

Dillon-Dumesnil (approved):

course_discovery/apps/course_metadata/management/commands/migrate_comments_to_salesforce.py
jlajoie: There is a .created, but we order by in Old Pub by modified, so it kind of made sense that if we snapshot the state based on Modified, that's how it was "created" in Salesforce.

PR #2170, opened 2019-09-19
Track url_slug history

rgraber:
Create CourseUrlSlugHistory model to store previous urls of courses, use is_active to determine current URL


course_discovery/apps/course_metadata/migrations/0195_auto_20190909_1601.py
```diff
@@ -1,20 +0,0 @@
-# -*- coding: utf-8 -*-
```
rgraber: Would ordinarily leave this and migration 0196 in but doing so makes it impossible to roll back the new migration

course_discovery/apps/api/serializers.py
```diff
@@ -554,6 +556,21 @@ class Meta(object):
...
+    class Meta(object):
+        model = CourseUrlSlugHistory
+        fields = 'url_slug'
+
+
+class CourseUrlSlugHistorySerializer(MinimalCourseUrlSlugHistorySerializer):
+
+    class Meta(MinimalCourseUrlSlugHistorySerializer.Meta):
+        fields = (MinimalCourseUrlSlugHistorySerializer.Meta.fields,) + ('partner', 'is_active', 'previously_published',
+                                                                         'course',)
```
mikix: Does this work? Looks like you're making a tuple with a tuple inside? And the line is mad long. And the fields aren't sorted, which leads to Eldritch madness. Maybe do:
```
        fields = MinimalCourseUrlSlugHistorySerializer.Meta.fields + (
            'course',
            'is_active',
            'partner',
            'previously_published',
        )
```

course_discovery/apps/api/serializers.py
```diff
@@ -895,7 +912,7 @@ def prefetch_queryset(cls, queryset=None, course_runs=None):
     class Meta:
         model = Course
         fields = ('key', 'uuid', 'title', 'course_runs', 'entitlements', 'owners', 'image',
-                  'short_description', 'url_slug', )
+                  'short_description', )
```
mikix: This is an API break, yeah?

If we don't want to provide data in that field, we should probably at least still offer it with an always-blank (or null) value. See `MinimalPersonSerializer.get_email` for example.

course_discovery/apps/api/serializers.py
```diff
@@ -914,6 +931,8 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     original_image = ImageField(read_only=True, source='original_image_url')
     extra_description = AdditionalPromoAreaSerializer(required=False)
     topics = TagListSerializerField(required=False)
+    url_slug = serializers.SerializerMethodField()
```
mikix: I think you can just do a normal field with `source=active_url_slug` (like `original_image` above). (Or does it being an `@property` mean you can't do that?)

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -253,7 +253,7 @@ def test_list_uuid_filter(self):
         uuids = ','.join([str(course.uuid) for course in courses])
         url = '{root}?uuids={uuids}'.format(root=reverse('api:v1:course-list'), uuids=uuids)
 
-        with self.assertNumQueries(56):
+        with self.assertNumQueries(60):
```
mikix: I thought all the prefetching stuff was meant to prevent the query count increasing?

course_discovery/apps/course_metadata/migrations/0195_auto_20190909_1601.py
dianakhuang: Will this cause issues when we try to migrate prod? Have we checked this use case?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2404,6 +2447,26 @@ class Meta(object):
         verbose_name_plural = 'Person Areas of Expertise'
 
 
+class CourseUrlSlugHistory(DraftModelMixin, TimeStampedModel):
```
mikix: Why be draft? Partners aren't going to be editing draft versions of this, right? I assume not, in which case, you could just have it point to either a draft or official course.

course_discovery/apps/course_metadata/models.py
```diff
@@ -628,6 +627,13 @@ def marketing_url(self):
 
         return url
 
+    @property
+    def active_url_slug(self):
+        for entry in self.url_slug_history.all():
```
dianakhuang: Is there a reason why we're not using `filter` for this?

course_discovery/apps/course_metadata/models.py
```diff
@@ -745,6 +760,35 @@ def _update_or_create_official_version(self, course_run):
...
+                new_url_slug_kwargs = {
+                    'course': draft_version,
+                    'url_slug': slug,
+                    'is_active': True,
+                    'previously_published': False,
+                    'draft': True,
+                    'partner': self.partner
+                }
+                created = CourseUrlSlugHistory.everything.create(**new_url_slug_kwargs)
+            # clears the prefetch cache so the new slug registers
```
dianakhuang: Nit: indenting on the comment.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2404,6 +2447,26 @@ class Meta(object):
...
+    partner = models.ForeignKey(Partner, models.CASCADE)
+    url_slug = AutoSlugField(populate_from='course__title', editable=True, slugify_function=uslugify,
+                             overwrite_on_add=False)
+    is_active = models.BooleanField(default=False)
+    previously_published = models.BooleanField(default=False)
+
+    # ensure partner and draft matches course
+    def save(self, **kwargs):
+        self.partner = self.course.partner
+        super().save(**kwargs)
```
mikix, line 2605: This feels like a method of ensuring they match that leads to surprises. How about providing a `clean()` class method or a field level `validators` arg.
https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.clean
https://docs.djangoproject.com/en/1.11/ref/validators/

mikix (commented):
It seems like this PR is still being actively worked on? I'll come back later, but here are some comments.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2404,6 +2447,26 @@ class Meta(object):
         verbose_name_plural = 'Person Areas of Expertise'
 
 
+class CourseUrlSlugHistory(DraftModelMixin, TimeStampedModel):
+    course = models.ForeignKey(Course, models.CASCADE, related_name='url_slug_history')
+    # need to have these on the model separately for unique_together to work, but it should always match course.partner
+    partner = models.ForeignKey(Partner, models.CASCADE)
```
mikix, line 2588: You can also provide a custom `validate_unique` method to follow through the course object, so you don't need duplicate data here that can get out of sync.
https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.validate_unique

course_discovery/apps/api/v1/tests/test_views/test_courses.py
rgraber: I thought so too but was having trouble there. I'm not convinced I'm doing the prefetching correctly and may bother you about it later

course_discovery/apps/course_metadata/utils.py
```diff
@@ -583,6 +581,14 @@ def publish_to_course_metadata(partner, course_run, create_official=True):
             }
         )
 
+    try:
+        discovery_course.set_active_url_slug(publisher_course.url_slug)
+    except IntegrityError as ie:
+        if fail_on_url_slug:
```
dianakhuang, line 595: I'm curious about why we'd only want to fail on this sometimes. I would say leave a comment where we're calling `fail_on_url_slug=False` to explain the different code paths.

dianakhuang (commented):
Some comments.

course_discovery/apps/course_metadata/migrations/0195_auto_20190909_1601.py
rgraber: point. I also don't remember why it broke the rollback so I will investigate that again

course_discovery/apps/course_metadata/models.py
rgraber: My impression is that validate_unique and clean only work if you're going through a ModelForm, which isn't always the case

course_discovery/apps/course_metadata/models.py
rgraber: Having draft versions helps make the uniqueness constraint easier to enforce because we need to account for the draft status of the related course anyway (see other comment about why validate_unique isn't great for this). It also made transferring slugs from draft to published easier since it fits in the current publishing structure.

rgraber:
After sinking a day into trying to optimize the prefetching to reduce the number of queries, decided to punt that into a new story since it shouldn't be enough to have a noticeable performance impact. https://openedx.atlassian.net/browse/WEBSITE-161

dianakhuang (approved):
Works for me.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -583,6 +589,14 @@ def publish_to_course_metadata(partner, course_run, create_official=True):
...
+            raise ie
```
mikix: Nit: you can simpley say `raise` (and then you can drop `as ie` above).

course_discovery/apps/course_metadata/utils.py
```diff
@@ -583,6 +589,14 @@ def publish_to_course_metadata(partner, course_run, create_official=True):
...
+        url_slug = discovery_course.active_url_slug if discovery_course.active_url_slug else ''
+        discovery_course.set_active_url_slug(url_slug)
```
mikix: I don't understand this bit? You're just saving the same value back? Why the `x if x else ''` bit also? Isn't that (roughly) the same as just `x`?

course_discovery/apps/course_metadata/models.py
```diff
@@ -749,6 +765,35 @@ def _update_or_create_official_version(self, course_run):
 
         return official_version
 
+    def set_active_url_slug(self, slug):
+        with transaction.atomic():
```
mikix: Use the decorator instead. Saves you a whole indent and is more readable, I'd argue.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -855,7 +855,7 @@ def test_official_created(self):
 
         self.course_run.draft = True
         self.course_run.status = CourseRunStatus.Reviewed
-        self.course_run.course.draft = True
+        self.course_run.course = factories.CourseFactory(draft=True)
```
mikix: These are very different statements! One is preserving all the other fields and one is throwing them all away... Why does this test need this change?

course_discovery/apps/course_metadata/models.py
mikix: OK, so you didn't like the proposed validators because you wanted to cach saves that happen outside Forms and serializers (which also call the model validation).

Fair enough. But you did not address the root of my complaint -- simply overwriting the partner is likely surprising behavior for the caller of `save()`. Can you not throw an error instead if partners don't match?

course_discovery/apps/course_metadata/models.py
mikix: Fair. I guess you could do some validation in `save()` then, but your call on whether you think the extra partner column is better/worse.

course_discovery/apps/course_metadata/models.py
mikix: Why do you have to account for the draft status of the related course?

There are plenty of other "course-related data" that doesn't get its own draft status. Person. Video. Image. Because they are just "chunks of data" that don't need to be represented as an "in-progress" version. The draft version of a Course can simply link to a different Video object than the official Course. And that's a perfectly valid way to capture that the data is different between them. The Video object having a draft column would not help at all and simply end up confusing. I think the same is true here.

But maybe some aspect of your uniqueness problem is getting in the way here?  Can we maybe chat about that in person, for higher bandwidth?

course_discovery/apps/course_metadata/models.py
```diff
@@ -749,6 +765,35 @@ def _update_or_create_official_version(self, course_run):
...
+            draft_version = ensure_draft_world(self)
+            # nothing else is active
+            for url_slug_object in draft_version.url_slug_history.filter(course=draft_version,
+                                                                         is_active=True).exclude(url_slug=slug):
```
mikix: You shouldn't need the `course=draft_version` here, since you started with `draft_version.url_slug_history`.

course_discovery/apps/course_metadata/models.py
```diff
@@ -749,6 +765,35 @@ def _update_or_create_official_version(self, course_run):
...
+                    'course': draft_version,
+                    'url_slug': slug,
+                    'is_active': True,
+                    'previously_published': False,
+                    'draft': True,
+                    'partner': self.partner
+                }
+                created = CourseUrlSlugHistory.everything.create(**new_url_slug_kwargs)
+                # clears the prefetch cache so the new slug registers
+                self.url_slug_history.add(created)
```
mikix: Shouldn't need this line - it's automatically done by setting `'course': draft_version'` above. I'm not even sure this line works? Like, what does it do if you try to add an object with a different `course` property than `self`?

course_discovery/apps/course_metadata/models.py
```diff
@@ -749,6 +765,35 @@ def _update_or_create_official_version(self, course_run):
...
+                                                                         is_active=True).exclude(url_slug=slug):
+                url_slug_object.is_active = False
+                url_slug_object.save()
+            creating = not CourseUrlSlugHistory.everything.filter(course=draft_version, url_slug=slug).exists()  # pylint: disable=no-member
+            if creating:
+                new_url_slug_kwargs = {
+                    'course': draft_version,
+                    'url_slug': slug,
+                    'is_active': True,
+                    'previously_published': False,
```
mikix: Is defaulted to False, don't need to specify.

course_discovery/apps/course_metadata/models.py
```diff
@@ -749,6 +765,35 @@ def _update_or_create_official_version(self, course_run):
...
+    def set_active_url_slug(self, slug):
+        with transaction.atomic():
+            draft_version = ensure_draft_world(self)
+            # nothing else is active
+            for url_slug_object in draft_version.url_slug_history.filter(course=draft_version,
+                                                                         is_active=True).exclude(url_slug=slug):
+                url_slug_object.is_active = False
+                url_slug_object.save()
+            creating = not CourseUrlSlugHistory.everything.filter(course=draft_version, url_slug=slug).exists()  # pylint: disable=no-member
+            if creating:
```
mikix: This whole block could simply be an update_or_create call, yeah? Why do all that work manually?

course_discovery/apps/course_metadata/models.py
```diff
@@ -740,6 +747,15 @@ def _update_or_create_official_version(self, course_run):
...
                 set_official_state(entitlement, CourseEntitlement, {'course': official_version})
 
+        for url_slug_object in self.url_slug_history.all():
+            if url_slug_object.previously_published or url_slug_object.is_active:
+                set_official_state(url_slug_object, CourseUrlSlugHistory, {'course': official_version,
+                                                                           'previously_published': True})
+
+        active_url_slug_draft = CourseUrlSlugHistory.everything.get(course=self, is_active=True)  # pylint: disable=no-member
+        active_url_slug_draft.previously_published = True
+        active_url_slug_draft.save()
```
mikix: Hmm.... What does `previously_published` really mean? This is where we make an official version yes... But the `publish()` method of a CourseRun is where we actually make something published. Which is possibly months after we promote something to an official version.

course_discovery/apps/course_metadata/models.py
mikix: Seconded.

mikix (commented):
Was there an investigation into whether some django plugin existed out there that did a CurrentValue+History for you? This is a lot of custom code for what is surely a solved problem.

We use ConfigurationModel in LMS for a model-wide "active+history". I bet someone has also solved it at a field level. Then we don't have to maintain all this ourselves.

course_discovery/apps/course_metadata/models.py
rgraber: Good call, I can throw the error.

course_discovery/apps/course_metadata/models.py
```diff
@@ -749,6 +765,35 @@ def _update_or_create_official_version(self, course_run):
...
+                    'is_active': True,
+                    'previously_published': False,
+                    'draft': True,
+                    'partner': self.partner
+                }
+                created = CourseUrlSlugHistory.everything.create(**new_url_slug_kwargs)
+                # clears the prefetch cache so the new slug registers
+                self.url_slug_history.add(created)
+            else:
+                existing = CourseUrlSlugHistory.everything.get(course=draft_version, url_slug=slug)  # pylint: disable=no-member
```
mikix: You're also re-looking this up again, after throwing it away when setting `creating` above - don't query twice here, just save the value above. And this could more succinctly be written as `draft_version.url_slug_history.get(url_slug=slug)`

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -317,17 +323,23 @@ def update_course(self, data, partial=False):
...
-        if (validated_data.get('url_slug') and
-            Course.everything.filter(url_slug=validated_data['url_slug'],
-                                     partner=course.partner).exclude(uuid=course.uuid).exists()):
-            raise Exception(
-                _('Course edit was unsuccessful. The course URL slug ‘[{url_slug}]’ is already in use. '
-                  'Please update this field and try again.').format(url_slug=validated_data['url_slug']))
+        if url_slug:
+            validators.validate_slug(url_slug)
+            all_course_historical_slugs = CourseUrlSlugHistory.objects.filter(url_slug=url_slug,
+                                                                                 course__partner=course.partner)
```
mikix: You could just check `partner` directly? It's in the model. Same comment elsewhere.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2522,6 +2595,27 @@ class Meta(object):
...
+    # need to have these on the model separately for unique_together to work, but it should always match course.partner
+    partner = models.ForeignKey(Partner, models.CASCADE)
+    url_slug = AutoSlugField(populate_from='course__title', editable=True, slugify_function=uslugify,
+                             overwrite_on_add=False)
+    is_active = models.BooleanField(default=False)
+    is_active_on_draft = models.BooleanField(default=False)
+
+    # ensure partner matches course
+    def save(self, **kwargs):
+        if self.partner.pk != self.course.partner.pk:
```
mikix: Shouldn't need the `.pk` bit, but that's a super nit.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2522,6 +2595,27 @@ class Meta(object):
...
 
 
+class CourseUrlSlugHistory(TimeStampedModel):
+    course = models.ForeignKey(Course, models.CASCADE, related_name='url_slug_history')
+    # need to have these on the model separately for unique_together to work, but it should always match course.partner
+    partner = models.ForeignKey(Partner, models.CASCADE)
+    url_slug = AutoSlugField(populate_from='course__title', editable=True, slugify_function=uslugify,
+                             overwrite_on_add=False)
+    is_active = models.BooleanField(default=False)
+    is_active_on_draft = models.BooleanField(default=False)
```
mikix, line 2595: I don't fully grok this field yet, but I trust I will later -- maybe add a comment about what's going on here above the field definition?

mikix (commented):
This feels a little cleaner, I think.

Except get and set for `active_url` - those methods look like beasts. I kinda skimmed them.

course_discovery/apps/api/serializers.py
rgraber: ended up removing the serializer altogether, so marking this as resolved

course_discovery/apps/api/v1/tests/test_views/test_courses.py
rgraber: This has been logged as a separate optimization effort in order to speed up release. https://openedx.atlassian.net/browse/WEBSITE-161

course_discovery/apps/course_metadata/models.py
rgraber: this method has been overhauled, resolving

course_discovery/apps/course_metadata/models.py
rgraber: this bit was deleted, resolving

course_discovery/apps/course_metadata/utils.py
rgraber: comment added

course_discovery/apps/api/serializers.py
rgraber: this has been changed

course_discovery/apps/api/serializers.py
```diff
@@ -893,10 +895,13 @@ def prefetch_queryset(cls, queryset=None, course_runs=None):
             Prefetch('course_runs', queryset=MinimalCourseRunSerializer.prefetch_queryset(queryset=course_runs)),
         )
 
+    def get_url_slug(self, obj):  # pylint: disable=unused-argument
+        return None  # this has been removed from the MinimalCourseSerializer, set to None to not break APIs
```
mikix, line 899: Hey - when did we add the `url_slug` field to the serializer? If there hasn't been an open edx release since then (i.e. if this addition was all internal to git master), then I actually don't think we need to worry about API stability and could just delete.

mikix: Oh also - I just realized what was happening here - you're taking it off the minimal and adding it to the standard serializer. Why? Why "break API" instead of just leaving it? (You can keep the history in the standard, but might as well have the normal url_slug in minimal yeah?)

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -342,7 +350,10 @@ def update_course(self, data, partial=False):
                 course_run.official_version.status = CourseRunStatus.Unpublished
                 course_run.official_version.save()
 
-        return Response(serializer.data)
+        # hack to get the correctly-updated url slug into the response
```
mikix, line 352: Why would you need to do this? Wouldn't the serializer include it for you?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -893,8 +895,8 @@ def test_official_created(self):
     def test_official_canonical_updates_to_official(self):
         self.course_run.draft = True
         self.course_run.status = CourseRunStatus.Reviewed
-        self.course_run.course.canonical_course_run = self.course_run
         self.course_run.course.draft = True
+        self.course_run.course.canonical_course_run = self.course_run
```
mikix: This change is probably droppable yah?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -741,6 +741,8 @@ def test_get_video(self):
     @mock.patch('course_discovery.apps.course_metadata.emails.send_email_for_reviewed')
     def test_reviewed_with_go_live_date(self, when, published, mock_email):
         draft = factories.CourseRunFactory(draft=True, go_live_date=when, announcement=None)
+        draft.course.draft = True
+        draft.course.save()
```
mikix, line 745: Interesting. I wonder if it's possible to fix the factory to do that automatically. Not necessary for this PR, but I've not seen factories do that kind of thing before. I assume there is some plumbing there.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2522,6 +2585,30 @@ class Meta(object):
...
+    is_active = models.BooleanField(default=False)
+
+    # useful if a course editor decides to edit a draft and provide a url_slug that has already been associated
+    # with the course
+    is_active_on_draft = models.BooleanField(default=False)
+
+    # ensure partner matches course
+    def save(self, **kwargs):
+        if self.partner.pk != self.course.partner.pk:
+            raise Exception('Partner and course do not match')
```
mikix: Might be possible to find a more specific exception. Like ValidationError maybe? Not crucial, just boring Exceptions make life harder down the line when someone is trying to catch them.

Should also include course uuid, partner key, url slug and/or something else to help a future engineer seeing this in a log to figure out what happened.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1446,7 +1510,6 @@ def update_or_create_official_version(self, create_ecom_products=True):
             set_official_state(seat, Seat, {'course_run': official_version})
 
         official_course = self.course._update_or_create_official_version(official_version)  # pylint: disable=protected-access
-
```
mikix, line 1449: Leftover delta from some changes here?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2522,6 +2585,30 @@ class Meta(object):
         verbose_name_plural = 'Person Areas of Expertise'
 
 
+class CourseUrlSlugHistory(TimeStampedModel):
```
mikix: So what makes this model a history? It's just one slug. I guess the `active` field a little bit. But feels like the `History` word should just be used on the `url_slug_history` field and the model itself can just live its life in the open as a full on garbage person Slug.

I mean, I guess the table as a whole is a history. But like, so is any model table. And using the `History` word gets a little confusing with the `HistoricalRecords` tables used by data engineering.

This is more nit than anything, if you don't feel like regenerating your migrations. But the name does give me pause when I see it and then don't see any historical data in the defined columns.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2522,6 +2585,30 @@ class Meta(object):
...
+    # ensure partner matches course
+    def save(self, **kwargs):
+        if self.partner.pk != self.course.partner.pk:
+            raise Exception('Partner and course do not match')
+        super().save(**kwargs)
+
+    class Meta:
+        unique_together = (
+            ('partner', 'url_slug')
+        )
```
mikix, line 2610: Also, `('course', 'is_active')` would be true? Imma come back to this after looking at the rest of the PR.

course_discovery/apps/course_metadata/models.py
```diff
@@ -733,6 +732,19 @@ def marketing_url(self):
 
         return url
 
+    @property
+    def active_url_slug(self):
```
mikix, line 736: A docstring explaining what *should* happen in English might be nice - like "Official rows just return the current active slug, but drafts are more complicated..." - I dunno how deep you wanna get. While all code is technically self documenting, this has enough logic branches that English might help a newcomer.

course_discovery/apps/course_metadata/models.py
```diff
@@ -733,6 +732,19 @@ def marketing_url(self):
...
+        if not self.draft:
+            return self.url_slug_history.get(is_active=True).url_slug
```
mikix, line 741: You want to throw if there happens to not be one? I would normally expect a `None` or `''` value in that case.

course_discovery/apps/api/serializers.py
rgraber: Just to avoid having an extra query to CourseUrlSlugHistory in the minimal version.

course_discovery/apps/course_metadata/models.py
```diff
@@ -733,6 +732,19 @@ def marketing_url(self):
...
+        else:
+            try:
+                return self.url_slug_history.get(is_active=True).url_slug
+            except ObjectDoesNotExist:
+                # current draft url slug has already been published at least once, so get it from the official course
+                if self.official_version:
+                    return self.official_version.url_slug_history.get(is_active_on_draft=True).url_slug
+                return None
```
mikix, line 749: This could be something like... `filter(Q(course=self, is_active=True) | Q(course=self.official_version, is_active_on_draft=True))` to avoid two queries. Though I don't know what ordering is guaranteed if both exist - maybe an order_by could do that, or checking the value of first().draft or something.

Just a tiny performance nit.

course_discovery/apps/course_metadata/models.py
```diff
@@ -841,6 +853,8 @@ def _update_or_create_official_version(self, course_run):
             if entitlement.mode.slug in Seat.ENTITLEMENT_MODES:
                 set_official_state(entitlement, CourseEntitlement, {'course': official_version})
 
+        official_version.set_active_url_slug(official_version.draft_version.active_url_slug)
```
mikix: Nit: Odd construction. Is `draft_version` not reliable at this point because its object was taken over? Could use `self.active_url_slug` then.

course_discovery/apps/course_metadata/models.py
```diff
@@ -850,6 +864,56 @@ def _update_or_create_official_version(self, course_run):
...
         return official_version
 
+    @transaction.atomic
+    def set_active_url_slug(self, slug):
+        if self.draft:
+            active_draft_url_slug_object = None
+            try:
+                active_draft_url_slug_object = self.url_slug_history.get(is_active=True)
+            except ObjectDoesNotExist:
+                pass
```
mikix: More simply written as `active_draft_url_slug_object = self.url_slug_history.filter(is_active=True).first()` unless you are worried about the multiple-active case.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2522,6 +2585,30 @@ class Meta(object):
         verbose_name_plural = 'Person Areas of Expertise'
 
 
+class CourseUrlSlugHistory(TimeStampedModel):
+    course = models.ForeignKey(Course, models.CASCADE, related_name='url_slug_history')
```
mikix, line 2586: You don't want to `CASCADE`, do you? This URL is still "historically claimed" eh? Maybe SET_NULL?

course_discovery/apps/course_metadata/models.py
```diff
@@ -850,6 +864,56 @@ def _update_or_create_official_version(self, course_run):
...
+    @transaction.atomic
+    def set_active_url_slug(self, slug):
+        if self.draft:
+            active_draft_url_slug_object = None
+            try:
+                active_draft_url_slug_object = self.url_slug_history.get(is_active=True)
+            except ObjectDoesNotExist:
+                pass
+
+            # case 1: previously used on course
```
mikix: "previously used on course" isn't quite right to my ears. Maybe like "has an existing slug"

course_discovery/apps/course_metadata/models.py
```diff
@@ -850,6 +864,56 @@ def _update_or_create_official_version(self, course_run):
...
+                'url_slug': slug,
+            })[0]  # update_or_create returns an (obj, created?) tuple, so just get the object
+            # this line necessary to clear the prefetch cache
+            self.url_slug_history.add(obj)
+        else:
+            if self.draft_version:
+                try:
+                    self.draft_version.url_slug_history.get(is_active=True).delete()  # pylint: disable=no-member
+                except ObjectDoesNotExist:
+                    pass
```
mikix: You can just call `filter().delete()`

mikix (commented):
Random thought - does the frontend strip space on stuff for us? Or like, when we are accidentally sent a trailing space, does that get saved to db?

Here's another thought. Some of this logic is complicated by the whole draft/official thing and trying to keep a foreignkey up to date. What you really want is a foreign key to the **concept** of a course. What if you dropped the FK and used a course UUID instead? Then you wouldn't have to move anything around. Lookups and setting would be much simpler. It's not a common model design, since FKs have their benefits... But also, you may not want some of those benefits. Like, if the course goes away, you might not want your URL slug entry to go away.  But you'd still need to track `is_active` and `is_active_on_draft` separately.

Maybe the better way would be to go further into "This model is just a pile of URLs" and have `active_url_slug` be a FK. With either `unique=True` and you simply null out the draft value when migrating draft data to official. **Or** with a `save()` time check that any dups share the same UUID or something.

I'm just saying my brain hurts a bit with all this `active_on_draft` stuff.

course_discovery/apps/api/serializers.py
mikix: Could we prefetch the url history to avoid that?

rgraber: It would still add an extra DB call since prefetch would make a call to the CourseUrlSlugHistory table and then do the join in python

mikix: Oh weird. I'd think Django **could** grab the table via a subquery or something all in one go. But fair, if it doesn't do that, can't help it.

mikix: Although, it feels real weird to get a different answer from the same API...

course_discovery/apps/course_metadata/models.py
rgraber: I'll need to double check with product about this one to determine if a slug is still considered claimed if a course is deleted.

rgraber: Ok product says that a slug is no longer considered claimed if the course is deleted, so the CASCADE is correct here

course_discovery/apps/course_metadata/models.py
```diff
@@ -850,6 +864,56 @@ def _update_or_create_official_version(self, course_run):
 
         return official_version
 
+    @transaction.atomic
+    def set_active_url_slug(self, slug):
```
mikix, line 871: Does this never accept None? No way to unset a slug?

rgraber: There is no way to unset a slug right now, except to override it on the draft version. If it gets passed None it will just auto-generate the slug with uslugify

course_discovery/apps/api/v1/views/courses.py
rgraber: annoyingly no, because it's possible that what ends up in the course url_slug is not what came from the front end

mikix: But how? Didn't we just set the active slug? (I must be missing some context here)

rgraber:
re: whitespace
The SlugField validator catches them and will throw an error. We could conceivably strip them on the front end if we wanted to

course_discovery/apps/api/v1/views/courses.py
rgraber: The serializer doesn't update itself when the course updates, and we took the url_slug out of the incoming data before we call get_serializer so we could handle it separately

course_discovery/apps/course_metadata/models.py
rgraber: resolving this conversation since the approach was changed and it will make it easier to see which things have not actually been addressed yet

course_discovery/apps/course_metadata/models.py
rgraber: no, because there can be several inactive slugs

mikix: Great point :)

mikix (approved):
Sure? We've talked plenty about the model design. I still think the FK relationship should be flipped, but I get that there are time constraints.

I have a comment on active_url_slug about a better docstring and throwing when there isn't an existing slug that would be nice to fix. I didn't actually test any of this and didn't fully review the migration file.

But all in all, this is looking pretty good! Thanks Becca for sticking through all my nits.

PR #2171, opened 2019-09-20
Add key_for_reruns field to Courses

mikix:
This field will specify the course key used when creating new reruns under the course. This allows our course teams to "soft rename" course keys.

Their current approach for this is to create a run under a course with the desired key. Then move the run to the original course. Hopefully this will be easier for everyone.

https://openedx.atlassian.net/browse/DISCO-1358


docs/decisions/0006-disallow-course-number-changes-in-publisher.rst
```diff
@@ -65,10 +65,17 @@ create new Courses to have the desired result.
...
 --------
 
-At this time, we **disallow** modifying the Course number in both the Publisher
-app and the Publisher Microfrontend. I believe this could (and should) change
-in the future so I am documenting below the investigation I did including some
-of the changes that would need to happen before switching over.
+At this time, we **disallow** modifying the Course number in both the Publisher app and the Publisher Microfrontend.
+The Course number is used as a matching string between data sets in too many places.
+
+But we **will allow** setting a separate Course field that controls what key new reruns use as a course key.
```
jlajoie: tiny nit: make it just **allow** since this is something that's possible today, and not in the future?

course_discovery/apps/course_metadata/models.py
```diff
@@ -522,6 +522,10 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         default=None, null=True, blank=True,
     )
     key = models.CharField(max_length=255, db_index=True)
+    key_for_reruns = models.CharField(
```
jlajoie, line 525: Will we want an index on this? Like, will we ever be querying for a course based on the key_for_rerun? Doesn't need to be a now thing, and I'm _also_ assuming no but wanted to call it out

jlajoie (approved):
Seems simple enough! Minor questions but looks good to go 👍 

Also _60 files_ for this!

docs/decisions/0006-disallow-course-number-changes-in-publisher.rst
mikix: Ah. I was reading it like. "Your father and I have discussed it and we won't let you have a cookie, but we will allow you a carrot."

course_discovery/apps/course_metadata/models.py
mikix: Hmph. I don't think we'll query off this. I hope not anyway? Like, this field don't need to be unique or care about colliding with another key somewhere else. So there's not much use case for looking up based off of it.

So I'd say no index?

mikix: Good question though. I almost never think about indexes.

docs/decisions/0006-disallow-course-number-changes-in-publisher.rst
mikix: But I'll drop the will, since it's possibly confusing.

mikix: *will allow* -> *do allow*

PR #2172, opened 2019-09-20
DISCO-1355 disable views for pcs

xnick421:
Disable course detail and edit views for pcs in old publisher
if the bviewed courses organization has been moved over to new
publisher via the ORGS_ON_NEW_PUB_FE setting


course_discovery/apps/publisher/utils.py
```diff
@@ -163,37 +163,59 @@ def user_orgs(user):
     return Organization.user_organizations(user)
 
 
-def is_on_new_pub_fe(user):
-    """Returns if all the user's organizations have been moved to new publisher frontend"""
-    if user.is_staff:
-        return False
+def check_orgs_from_settings(method):
```
mikix, line 166: I don't love this approach. It's cumbersome for callers (you've had to add `user=` or `course=` to all callers. And it's not very pythonic (having to specify the arg name for a single-arg function is unusual).

You could get the same effect by using a helper function. Like:

```
def _check_orgs_from_settings(orgs):
        orgs_on_new_pub_fe = frozenset(filter(None, getattr(settings, 'ORGS_ON_NEW_PUB_FE', '').split(',')))
        if not orgs_on_new_pub_fe:
            return False

        for org in orgs:
            if org.key not in orgs_on_new_pub_fe:
                return False
        return True


def is_on_new_pub_fe(user):
    """Returns if all the user's organizations have been moved to new publisher frontend"""
    if user.is_staff:
        return False

    return _check_orgs_from_settings(user_orgs(user))
```

course_discovery/apps/publisher/utils.py
```diff
@@ -163,37 +163,59 @@ def user_orgs(user):
...
 
-    orgs = course.authoring_organizations.all()
+
+@check_orgs_from_settings
+def is_publisher_course_on_new_pub_fe(**kwargs):
+    """
+    Handles publisher courses, not course_metadata courses
+    Returns True if all the course's organizations have been moved to new publisher frontend
+    """
+    orgs = kwargs['course'].organizations.all()
```
mikix: This is probably good enough for our temporary purposes. But unfortunately enough, this isn't actually the logic for finding what orgs a publisher course has. Because our data is so mucked up.

See `find_discovery_course` for how we map from Publisher course run to Disco course (and its latter half maps from Pub course to Disco course). And then get orgs from disco stuff I guess.

That said, let's just leave it as it is for now. It's the 99% solution.

mikix (approved):
Looks good! Just got a couple comments about approach.

mikix (approved):
:man_cook: :kiss:

PR #2173, opened 2019-09-20
Increase default rest throttle rate

jmyatt:


mikix:
After talking about it, we're going to exempt publisher users instead:
https://github.com/edx/course-discovery/pull/2175

PR #2193, opened 2019-09-27
DISCO-1348 post-save-mute

xnick421:
create factories for Course, CourseRun, and Organization which have
post_save hooks muted

### DIFF
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -1,6 +1,8 @@
 from datetime import datetime
 
 import factory
+import requests
+from django.db.models.signals import post_save
 from factory.fuzzy import FuzzyChoice, FuzzyDateTime, FuzzyDecimal, FuzzyInteger, FuzzyText
 from pytz import UTC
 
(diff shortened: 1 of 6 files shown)


course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -81,7 +83,19 @@ class Meta:
...
+class SalesforceRecordFactory(factory.DjangoModelFactory):
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        from course_discovery.apps.course_metadata.tests.utils import build_salesforce_exception
+        try:
+            return super()._create(model_class, *args, **kwargs)
+        except requests.ConnectionError:
+            raise ConnectionError(build_salesforce_exception(model_class.__name__))
+        except Exception as e:
+            raise e
```
mikix: You can simply do:

```
        except Exception:
            raise
```

(A bare `raise` will re-raise the current exception.)

But actually, since you're not doing anything else with the exception, you can simply leave these two lines off entirely and non-`ConnectionError` exceptions would bubble up naturally.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -81,7 +83,19 @@ class Meta:
...
 
-class CourseFactory(factory.DjangoModelFactory):
+class SalesforceRecordFactory(factory.DjangoModelFactory):
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        from course_discovery.apps.course_metadata.tests.utils import build_salesforce_exception
+        try:
+            return super()._create(model_class, *args, **kwargs)
+        except requests.ConnectionError:
+            raise ConnectionError(build_salesforce_exception(model_class.__name__))
```
mikix, line 94: Maybe just a quick comment above this that you're doing all this to make a nicer log message for folks. I was initially very confused on why this code existed.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -81,7 +83,19 @@ class Meta:
...
+        except Exception as e:
+            raise e
+
+
+class CourseFactory(SalesforceRecordFactory, factory.DjangoModelFactory):
```
mikix: You should either make your salesforce class a Mixin (instead of a Factory) or drop the DjangoModelFactory inherit, since it's currently a base class of your salesforce class.

mikix (approved):
Nice

PR #2198, opened 2019-10-01
Updating Course endpoint to use new CourseType (and friends) models

Dillon-Dumesnil:
This also contains a fix to not pass 'mode' into the course
serializer since it is not necessary and making some error
messages more clear.

DISCO-1379


course_discovery/apps/course_metadata/utils.py
```diff
@@ -213,7 +213,7 @@ def create_missing_entitlement(course):
     return False
 
 
-def ensure_draft_world(obj):
+def ensure_draft_world(obj, course_type=None):
```
mikix, line 221: Would a bool be better here?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -265,7 +265,7 @@ def ensure_draft_world(obj):
         if original_course.entitlements.exists():
             for entitlement in original_course.entitlements.all():
                 set_draft_state(entitlement, CourseEntitlement, {'course': draft_course})
-        elif not create_missing_entitlement(draft_course):
+        elif not course_type and not create_missing_entitlement(draft_course):
```
mikix: We talked about this in person, but just so it doesn't get lost: we talked about moving the check after the `create_missing_entitlement`

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -94,6 +94,37 @@ def _create(cls, model_class, *args, **kwargs):
...
+
+    @factory.post_generation
+    def entitlement_types(self, create, extracted, **kwargs):
+        if create:  # pragma: no cover
+            add_m2m_data(self.entitlement_types, extracted)
+
+    @factory.post_generation
+    def course_run_types(self, create, extracted, **kwargs):
+        if create:  # pragma: no cover
+            add_m2m_data(self.course_run_types, extracted)
```
mikix, line 125: This little chunk of code we use a lot. Is it just like, the Right Way to do this? Why can't factory-boy just do it itself then? Or maybe we should make a mixin or something to make that easier.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1205,6 +1322,21 @@ def test_update_with_api_exception(self):
                     )
                 )
 
+    def test_update_fails_with_course_type(self):
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.uuid})
+
+        with LogCapture(course_logger.name) as log_capture:
+            # response = self.client.patch(url, {'type': '96bb4bbf-b65d-47c7-9581-90152e2f6872'}, format='json')
```
mikix: Left this comment in?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -170,23 +170,29 @@ def create(self, request, *args, **kwargs):
...
         error_message = ''
         if missing_values:
             error_message += ''.join([_('Missing value for: [{name}]. ').format(name=name) for name in missing_values])
         if not Organization.objects.filter(key=course_creation_fields['org']).exists():
-            error_message += _('Organization does not exist. ')
-        if not SeatType.objects.filter(slug=course_creation_fields['mode']).exists():
-            error_message += _('Entitlement Track does not exist. ')
+            error_message += _('Organization [{org}] does not exist. ').format(org=course_creation_fields['org'])
+        if mode and not SeatType.objects.filter(slug=mode).exists():
+            error_message += _('Entitlement Track [{mode}] does not exist. ').format(mode=mode)
```
mikix: Maybe add the same but for `type`?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -210,15 +216,21 @@ def create(self, request, *args, **kwargs):
         organization = Organization.objects.get(key=course_creation_fields['org'])
         course.authoring_organizations.add(organization)
 
+        # New flow for courses using CourseType model
+        if course.type:
+            entitlement_mode = course.type.entitlement_type
```
mikix: You know... We should just make all the entitlements listed, insead of just the first one. I realize it's the same now, but that would be the Correct thing.

Like just have `modes = xxx` and then later loop through them. Though that would require some careful handling of price too.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -229,19 +241,15 @@ def create(self, request, *args, **kwargs):
         # Note: We have to send the request object as well because it is used for its metadata
         # (like request.user and is set as part of the serializer context)
         if course_run_creation_fields:
-            course_run_creation_fields.update({
-                'course': course_creation_fields['key'],
-                'price': price,
-                'mode': course_creation_fields['mode'],
-            })
```
mikix, line 237: This is deeper in the weeds than I can handle - why don't we send these on anymore?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -282,18 +294,28 @@ def update_course(self, data, partial=False):
...
-        for entitlement_data in entitlements_data:
-            if entitlement_data == {}:
-                continue
-            entitlements.append(self.update_entitlement(course, entitlement_data, partial=partial))
-            # We set changed to True here if the price of the course is being updated
-            changed = format(float(entitlement_data['price']), '.2f') != str(course.entitlements.first().price)
+        # New flow for courses using CourseType model
+        if data.get('type'):
+            course_type = CourseType.objects.get(uuid=data.get('type'))
+            mode = course_type.entitlement_type
```
mikix: We should be careful with wording here. `mode` is a different thing now. Maybe the more verbose `entitlement_type`?

And the chunk below in the old path was multi-entitlement aware. No reason not to be here too? And just iterate over `entitlement_types`?

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: We don't use them anymore (or maybe ever?). Now it is pulled out of the Course's entitlement in update_or_create_seats in the CourseRun model

course_discovery/apps/api/serializers.py
```diff
@@ -653,14 +653,16 @@ class MinimalCourseRunSerializer(DynamicFieldsMixin, TimestampModelSerializer):
     start = serializers.DateTimeField(required=True)  # required so we can craft key number from it
     end = serializers.DateTimeField(required=True)  # required by studio
     type = serializers.CharField(read_only=True, source='type_legacy')
+    run_type = serializers.SlugRelatedField(required=False, allow_null=True, slug_field='uuid',
+                                            queryset=CourseRunType.objects.all())
```
mikix: You don't need a `source='type'` thing?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -211,15 +226,23 @@ def create(self, request, *args, **kwargs):
         organization = Organization.objects.get(key=course_creation_fields['org'])
         course.authoring_organizations.add(organization)
 
+        # DISCO-1399: No need to check the if else. It will always be course.type
```
mikix, line 229: This is true for us. For open edx, we should probably include a comment in the documentation for the next release about migrating?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -211,15 +226,23 @@ def create(self, request, *args, **kwargs):
...
-        mode = SeatType.objects.get(slug=course_creation_fields['mode'])
-        CourseEntitlement.objects.create(
-            course=course,
-            mode=mode,
-            partner=partner,
-            price=price,
-            draft=True,
-        )
+        # For example, the Audit Course Type has no entitlement_type
+        if entitlement_types:
```
mikix: You can drop the `if` statement yeah?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -277,25 +302,40 @@ def update_course(self, data, partial=False):
...
-                continue
-            entitlements.append(self.update_entitlement(course, entitlement_data, partial=partial))
-            # We set changed to True here if the price of the course is being updated
-            changed = format(float(entitlement_data['price']), '.2f') != str(course.entitlements.first().price)
+        # New flow for courses using CourseType model
+        # DISCO-1399: I think pretty much the whole 'else' can be removed.
+        if data.get('type'):
+            course_type = CourseType.objects.get(uuid=data.get('type'))
+            entitlement_types = course_type.entitlement_types
+            if entitlement_types:
```
mikix: Can drop this `if` too I think.

mikix (approved):

course_discovery/apps/api/v1/views/courses.py
mikix: To be clear, not a thing for your ticket. Just a thing for us to do in general.

Dillon-Dumesnil: We should make a ticket about how to document this change for openedx, yeah? Like it can probably be wrapped in Publisher documentation since it is through the endpoint and only Publisher uses that.

course_discovery/apps/api/serializers.py
Dillon-Dumesnil: I don't think so..? I didn't use one for course and SlugRelatedField doesn't even have a `source` option

mikix: But there is no backing field on the model called `run_type` right?

course_discovery/apps/api/v1/views/courses.py
mikix: Agreed. I have a note in my notebook about it, so consider it already done.

mikix (approved):
Yeah, seems good! All the miscellaneous test changes are fine I guess. Just annoying for you to have to have waded through.

PR #2200, opened 2019-10-02
DISCO-1319 remove all references to marketing_url_path

xnick421:


course_discovery/apps/course_metadata/models.py
```diff
@@ -483,8 +482,8 @@ def __str__(self):
 
     @property
     def marketing_url(self):
-        if self.marketing_url_path:
-            return urljoin(self.partner.marketing_site_url_root, self.marketing_url_path)
+        if self.key:
+            return urljoin(self.partner.marketing_site_url_root, self.key)
```
mikix: Key can include capital letters. You want a sluggified version of it. I had originally talked about adding a new field for the slug. But maybe we could get away with calling uslugify on it here? ... Not sure.

xnick421: this is only temporary state, we would merge the second pr right after. but i can do that for the 20 minutes this version will be deployed

course_discovery/apps/course_metadata/models.py
```diff
@@ -447,7 +447,6 @@ class Organization(CachedMixin, TimeStampedModel):
                                                        'than period, underscore or hyphen. This key will be used '
                                                        'in the course\'s course key.'))
     name = models.CharField(max_length=255)
-    marketing_url_path = models.CharField(max_length=255, null=True, blank=True)
```
mikix, line 450: I usually leave this in until the migration removal (that way other people don't accidentally get your change in the meantime). But it's fine this way too.

Unless, does this interfere with your later migration of data?

mikix (approved):
Looks good, but I do think it should be sluggified in the meantime.

PR #2201, opened 2019-10-02
DISCO-1319 org source of truth

xnick421:
- Remove SchoolMarketingSiteDataLoader and
SponsorMarketingSiteDataLoader.

- Always set name, description, and logo_image_url in lms loader
if these fields are not already written to in Course Discovery.

- Remove Organization from Course Discovery which do not exist
lms and are orphans.

- Add slug to organization, populate with marketing_url on
migration if it exists, otherwise use key.  Drop marketing_url.

- Add slug to serializer as read_only field.


mikix:
`Remove SchoolMarketingSiteDataLoader and SponsorMarketingSiteDataLoader.`

That commit comment belongs in the other PR.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -24,6 +25,7 @@
 
 class OrganizationsApiDataLoader(AbstractDataLoader):
     """ Loads organizations from the Organizations API. """
+    org_keys = []
```
mikix: Nit: probably a set makes more sense for just making a collection and testing existense. But since you're passing into django's ORM, not sure if you see the performance benefits of that. But still.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -51,22 +53,43 @@ def ingest(self):
...
+            org.save()
 
-        Organization.objects.update_or_create(key__iexact=key, partner=self.partner, defaults=defaults)
         logger.info('Processed organization "%s"', key)
+        self.remove_non_lms_orphans()
+
+    def remove_non_lms_orphans(self):
+        organizations = Organization.objects.exclude(key__in=self.org_keys)
+        for org in organizations:
+            delete_orphans(org)
```
mikix: Delete orphans works on like, a classname. Like `delete_orphans(Organization)`... You might have to augment it to take a new param like `exclude=self.org_keys` and make `org_keys` a set of pk values, so that `delete_orphans` doesn't need to know any field names by doing something like `.exclude(pk__in=exclude)`

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -51,22 +53,43 @@ def ingest(self):
...
+                setattr(org, attr, body[attr])
+            org.logo_image_url = logo
+        else:
+            for attr in ('name', 'description', 'logo_image_url'):
+                if not getattr(org, attr, None):
+                    requires_save = True
+                    if attr == 'logo_image_url':
+                        setattr(org, attr, logo)
+                    else:
+                        setattr(org, attr, body[attr])
```
mikix: Why not just have those fields in the `update_or_create`?  Looks like you're trying to handle the logo in a special way, but why?

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -176,83 +174,3 @@ def test_ingest_update(self):
 
         for datum in api_data:
             self.assert_subject_loaded(datum)
-
-
-class SchoolMarketingSiteDataLoaderTests(AbstractMarketingSiteDataLoaderTestMixin, TestCase):
```
mikix: You already did all this in the other PR yeah?

course_discovery/apps/course_metadata/migrations/0199_add_organization_slug_remove_marketing_url.py
```diff
@@ -0,0 +1,34 @@
...
+        migrations.AddField(
+            model_name='organization',
+            name='slug',
+            field=django_extensions.db.fields.AutoSlugField(blank=True, editable=False, populate_from='key'),
+        ),
+        migrations.RunPython(populate_slug, migrations.RunPython.noop),
+        migrations.RemoveField(
+            model_name='organization',
+            name='marketing_url_path',
+        ),
```
mikix: It's fine because of your stop-gap measure of looking at the key value in the other PR. But it might have been safer to (PR 1) add new field, populate it, use it everywhere, dropping usages of marketing_url_path and (PR 2) drop marketing_url_path.

course_discovery/apps/course_metadata/migrations/0199_add_organization_slug_remove_marketing_url.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.db import migrations
+import django_extensions.db.fields
+
+
+def populate_slug(apps, schema_editor):
+    Organization = apps.get_model('course_metadata', 'Organization')
+    for org in Organization.objects.all():
+        marketing_url = getattr(org, 'marketing_url_path', None)
+        org.slug = marketing_url if marketing_url else getattr(org, 'key', None)
```
mikix: None as default value?  Probably want `''`?  And if you do want `None`, you don't need to specify, that's already the default for `getattr`

Marketing path is gonna be like `school/mitx` - but you just want the `mitx` part, not the whole thing.

Maybe `marketing_url.rsplit('/')[-1] if marketing_url else getattr(org, 'key', '').lower()`?

course_discovery/apps/course_metadata/models.py
```diff
@@ -447,7 +447,7 @@ class Organization(CachedMixin, TimeStampedModel):
                                                        'than period, underscore or hyphen. This key will be used '
                                                        'in the course\'s course key.'))
     name = models.CharField(max_length=255)
-    marketing_url_path = models.CharField(max_length=255, null=True, blank=True)
+    slug = AutoSlugField(populate_from='key', editable=False, slugify_function=uslugify)
```
mikix, line 450: Why `editable=False`?

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -221,7 +221,7 @@ class OrganizationFactory(SalesforceRecordFactory):
     banner_image_url = FuzzyURL()
     certificate_logo_image_url = FuzzyURL()
     partner = factory.SubFactory(PartnerFactory)
-    marketing_url_path = FuzzyText()
+    slug = FuzzyText()
```
mikix: You may not need that? Without specifying, the AutoSlugField should provide a slugified version of the key, yeah?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1058,12 +1058,12 @@ def test_str(self):
     def test_marketing_url(self):
         """ Verify the property creates a complete marketing URL. """
         expected = '{root}/{slug}'.format(root=self.organization.partner.marketing_site_url_root.strip('/'),
-                                          slug=self.organization.marketing_url_path)
+                                          slug=self.organization.slug)
```
mikix, line 1063: You deleted these in the other PR. But you're right that they are valid tests even in the new system and should be added back, if you land the other PR first.

course_discovery/apps/course_metadata/data_loaders/api.py
xnick421: `name` and `description` are the exact properties an organization.  `logo` isn't.  it's `logo_image_url` on a org.  it needs to be mapped.  we cant put these additional fields in the original `update_or_create` on line 64 because we need to first check if they have values.  if they do we dont want to change them

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -48,6 +50,10 @@ def ingest(self):
 
         logger.info('Retrieved %d organizations from %s.', count, api_url)
 
+        delete_orphans(Organization, exclude=self.loaded_org_pks)
+
+        logger.info('Removed orphan Organizations excluding those which were loaded via OrganizationsApiDataLoader')
```
jlajoie, line 55: We don't do translations for RCM apparently? :( Not a thing to change, just noticed this now though

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -58,14 +64,18 @@ def update_organization(self, body):
...
         }
 
-        if not self.partner.has_marketing_site:
-            defaults.update({
-                'name': body['name'],
-                'description': body['description'],
-                'logo_image_url': logo,
-            })
+        for attr in ('name', 'description', 'logo'):
+            if body[attr]:
```
jlajoie: Might be worth using .get(attr) instead of [attr] so you avoid KeyErrors if you're checking that a value for a key exists

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -58,14 +64,18 @@ def update_organization(self, body):
...
+                if attr == 'logo':
+                    defaults['logo_image_url'] = logo
+                else:
+                    defaults[attr] = body[attr]
+
+        org, created = Organization.objects.update_or_create(key__iexact=key, partner=self.partner, defaults=defaults)  # pylint: disable=unused-variable
```
jlajoie: nit: if you aren't using created I think python standard is to do `org, _ = Organization...` and that way you can remove the linter comment

course_discovery/apps/course_metadata/migrations/0201_add_organization_slug_remove_marketing_url.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.db import migrations
+import django_extensions.db.fields
+
+
+def populate_slug(apps, schema_editor):
+    Organization = apps.get_model('course_metadata', 'Organization')
+    for org in Organization.objects.all():
+        marketing_url = getattr(org, 'marketing_url_path')
+        org.slug = marketing_url.rsplit('/')[-1] if marketing_url else getattr(org, 'key', '').lower()
```
jlajoie, line 13: Open question, we sure we want this to be empty string instead of None? Both are falsey values but DB queries against it would be a little different.

jlajoie (approved):
Couple of comments but looks good overall

PR #2205, opened 2019-10-04
fix migration to reset all slugs

rgraber:


course_discovery/apps/course_metadata/migrations/0199_url_slug_history.py
```diff
@@ -16,35 +16,22 @@ class Migration(migrations.Migration):
     ]
 
     def migrate_data_forward(apps, schema_editor):
-        updated_draft_course_pks = []
+        published_drafts = []
```
mikix, line 19: nit: a set would make more sense for this use case.

course_discovery/apps/course_metadata/migrations/0199_url_slug_history.py
```diff
@@ -16,35 +16,22 @@ class Migration(migrations.Migration):
...
         Course = apps.get_model('course_metadata', 'course')
         CourseUrlSlug = apps.get_model('course_metadata', 'courseUrlSlug')
         for instance in Course.everything.all().order_by('draft'):
```
mikix, line 22: Not part of this change I know, but order by draft? Because you are collecting the drafts up in this list?

course_discovery/apps/course_metadata/migrations/0199_url_slug_history.py
```diff
@@ -16,35 +16,22 @@ class Migration(migrations.Migration):
...
                 historical_slug = CourseUrlSlug.objects.create(
                     course=instance,
                     partner=instance.partner,
                     is_active=True,
-                    is_active_on_draft=current_slug_is_active_on_draft
+                    is_active_on_draft=True
                 )
-                # need to set slug after create so it's not overridden with the autoslugification
-                historical_slug.url_slug = instance.url_slug
                 historical_slug.save()
```
mikix, line 31: Can drop this line.

mikix (approved):
Sure? I'm less familiar with whether the logic changes make sense. But assume it's great.

course_discovery/apps/course_metadata/migrations/0199_url_slug_history.py
rgraber: yep. the goal is to only manually update drafts that don't have associated published courses

PR #2206, opened 2019-10-04
DISCO-1396: Sends e-mails when a comment is left via API

jlajoie:


course_discovery/apps/publisher/templates/publisher/email/comment.txt
```diff
@@ -1,11 +1,11 @@
 {% load i18n %}
 
 {% blocktrans with date=comment_date|date:'m/d/y' time=comment_date.time trimmed %}
-{{ user_name }} made the following comment on {{ course_name }} {{ date }} at {{ time }}.
+{{ user_name }} made the following comment on {{ course_name }} {{ date }} at {{ time }}
```
jlajoie: time gets rendered as a.m. or p.m. so... ```at 3:00 p.m..``` looked sort of dumb to me

course_discovery/apps/publisher/templates/publisher/email/comment.txt
```diff
@@ -1,11 +1,11 @@
...
 {% endblocktrans %}
 
 {{ comment_message }}
 
-{% trans "View comment in Publisher" %}{{ page_url }}
+{% trans "View comment in Publisher " %}{{ page_url }}
```
jlajoie: Came out as no space in text and looekd bad, added in the space, guessing we'll need to update trans for htis

mikix: I think you can add a space between the two blocks, like:
`{% trans "View comment in Publisher" %} {{ page_url }}`

course_discovery/apps/publisher/templates/publisher/email/comment.txt
mikix: Oh, I hate that rendering. But your change makes sense.

course_discovery/apps/publisher/templates/publisher/email/comment.html
```diff
@@ -14,7 +14,7 @@
         </p>
 
         <p>
-            <a href="{{ page_url }}">{% trans "View comment in Publisher" as tmsg %}{{ tmsg | force_escape }}</a>
+            <a href="{{ page_url }}">{% trans "View comment in Publisher " as tmsg %}{{ tmsg | force_escape }}</a>
```
mikix: Same thing elsewhere - in between blocks might be better.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -240,3 +241,60 @@ def send_email_for_go_live(course_run):
...
+            'user_name': comment.get('user', {}).get('username'),
+            'course_name': course.title,
+            'comment_date': dateutil.parser.parse(comment.get('created')),
+            'page_url': '{url}/courses/{path}'.format(
+                url=course.partner.publisher_url.strip('/'), path=course.uuid
+            )
+        }
+
+        txt_template = 'publisher/email/comment.txt'
+        html_template = 'publisher/email/comment.html'
```
mikix: I am a little grumbly at using a template from a different djangoapp. I might vote for copying them over (even if they end up being the same byte-for-byte). But that's a nit. I realize our djangoapps are all sorts of interdependent.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -240,3 +241,60 @@ def send_email_for_go_live(course_run):
...
+        org = course.authoring_organizations.first()
+        project_coordinator = get_project_coordinator(org)
+        recipients = list(CourseEditor.course_editors(course))
+        if project_coordinator:
+            recipients.append(project_coordinator)
+
+        # remove email of comment owner if exists
+        for user in recipients:
+            if user.email == comment.get('user', {}).get('email'):
+                recipients.remove(user)
```
mikix: Could be `recipients = filter(lambda x: x.email != user_email, recipients` or if you add back `author`, just `recipients.remove(author)`? But if you keep this, you could probably early exit too. Though I guess performance isn't really a concern.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -240,3 +241,60 @@ def send_email_for_go_live(course_run):
...
+        plain_content = template.render(context)
+        template = get_template(html_template)
+        html_content = template.render(context)
+        email_msg = EmailMultiAlternatives(
+            subject, plain_content, from_address, to_addresses
+        )
+        email_msg.attach_alternative(html_content, 'text/html')
+        email_msg.send()
+    except Exception:  # pylint: disable=broad-except
+        logger.exception('Failed to send email notifications for comment on course %s', course.uuid)
```
mikix, line 297: You can probably just try/except on the send, like the other function in this file. Saves you a whole-method indent.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -240,3 +241,60 @@ def send_email_for_go_live(course_run):
...
     send_email_to_project_coordinator(course_run, 'course_metadata/email/go_live', subject)
+
+
+def send_email_for_comment(comment, course):
+    """ Send the emails for a comment.
+
+        Arguments:
+            comment (Dict): Comment dict returned from salesforce.py
+            course (Course): Course object for the comment
+            author (User): User object who made the post request
```
mikix, line 265: You dropped `author` at some point.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -240,3 +241,60 @@ def send_email_for_go_live(course_run):
...
+        for user in recipients:
+            if user.email == comment.get('user', {}).get('email'):
+                recipients.remove(user)
+
+        to_addresses = [recipient.email for recipient in recipients if is_email_notification_enabled(recipient)]
+
+        from_address = settings.PUBLISHER_FROM_EMAIL
+
+        context = {
+            'comment_message': comment.get('comment'),
```
mikix: Can the comment contain any characters that might cause us problem in the html version? Might want to escape (separately for html / text I guess...)

mikix (approved):
I'm bummed at how much copied boilerplate is here. I bet the existing run-focused methods could be extended to be course-focused. But eh. It's not code that changes a lot.

Approach seems good though!

course_discovery/apps/course_metadata/emails.py
jlajoie: Yeah I'm cool with it, I had them copied over and I'm like "I shouldn't do this" then removed it. Will update

course_discovery/apps/course_metadata/emails.py
jlajoie: Early exit is probably a good call, I'll add that in 👍

course_discovery/apps/course_metadata/emails.py
jlajoie: Added a filter force_escape to comment.html

course_discovery/apps/course_metadata/emails.py
jlajoie: My intent here was to still return a 201 regardless of if the email succeeds or not, then just pipe out an exception. Think it's worth adding the try/except into the POST endpoint instead?

mikix: Ah no I like what you were doing. I just hate large indented blocks. Was suggesting moving the `try:` down to just above `email_msg.send()`

mikix (approved):

PR #2207, opened 2019-10-07
Change the UUID field of the django admin on course runs admin so we can update it

schenedx:
We do this because we want to update the duplicated UUID in course_run database to be unique again, so we can make credentials discovery data copy to be successful again.

@mikix please review.


mikix (approved):
If that is enough to show up in admin page, seems good.

mikix (approved):
Love it.

PR #2208, opened 2019-10-07
Add backpopulate_course_type command

mikix:
This command will go through an org and/or course and try to set the new 'type' fields on courses and course runs based on existing entitlements and seats.

This allows us to slowly backfill these new fields as we also add support for them.

https://openedx.atlassian.net/browse/DISCO-1383


course_discovery/apps/course_metadata/models.py
```diff
@@ -306,7 +306,7 @@ class CourseRunType(TimeStampedModel):
 
     See docs/decisions/0009-LMS-types-in-course-metadata.rst for more information.
     """
-    uuid = models.UUIDField(default=uuid4, editable=False, verbose_name=_('UUID'))
+    uuid = models.UUIDField(default=uuid4, editable=False, verbose_name=_('UUID'), unique=True)
```
mikix, line 309: Unrelated, just noticed that it made sense.  (And I used a separate migration for this, just to help indicate it's unrelated.)

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -94,6 +94,29 @@ def _create(cls, model_class, *args, **kwargs):
...
+    class Meta(object):
+        model = SeatType
+
+    name = FuzzyText()
+
+
+class ModeFactory(factory.DjangoModelFactory):
+    name = FuzzyText()
+    slug = FuzzyText()
+
```
Dillon-Dumesnil, line 107: Nit: I also added `certificate_type` in here on mine since it is the only other field that does not get a default value.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2609,6 +2609,23 @@ class Meta:
...
+class BackpopulateCourseTypeConfig(SingletonModel):
+    """
+    Configuration for the backpopulate_course_type management command.
+    """
+    class Meta:
+        verbose_name = 'backpopulate_course_type argument'
+
+    arguments = models.TextField(
+        blank=True,
+        help_text='Useful for manually running a Jenkins job. Specify like "--org=key1 --org=key2".',
```
Dillon-Dumesnil, line 2622: Nit: Not sure how descriptive you want to get with the help_text here. There are a lot of potential arguments that can go into this job so I could see this being a lot. No strong preference.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2609,6 +2609,23 @@ class Meta:
         )
 
 
+class BackpopulateCourseTypeConfig(SingletonModel):
```
Dillon-Dumesnil, line 2613: Nit: The name of this is a little misleading since we are doing more than just backpopulating course types. Maybe just `BackpopulateTypesConfig`? I don't know.

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -0,0 +1,187 @@
...
+# This command is designed to help fill out the new-style 'type' fields for Courses and CourseRuns.
+# These fields are a more explicit declaration for what sort of enrollment modes a course supports.
+# Whereas before, you'd have to examine the seats and entitlements for a course/run to see what sort of
+# course it was (i.e. is it credit? is it verified?).
+#
+# Which is what this command does - it tries to match the existing seat/entitlement profile for a course and
+# its runs. Then set a matching CourseType and CourseRunType for each.
+#
+# This is idempotent.
+# This does not change existing type fields.
```
Dillon-Dumesnil, line 22: Stronger Nit: Combine this line with the one beneath it. `... type fields, but it will validate... `...

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -0,0 +1,187 @@
...
+                return True
+
+        return False
+
+    def backpopulate(self, options):
+        # We look at both draft and official rows
+        courses = Course.everything.filter(partner__short_code=options['partner']).filter(
+            Q(uuid__in=options['course']) |
+            Q(authoring_organizations__key__in=options['org'])
+        ).distinct()
```
Dillon-Dumesnil, line 167: How does the `distinct` call work when it's possible to get duplicates? As in, will it favor draft or official? Does it matter which one it favors? Why not use `filter_drafts` here?

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -0,0 +1,187 @@
...
+        type_seat_types = set(run_type.tracks.values_list('seat_type__slug', flat=True))
+        return run_seat_types == type_seat_types
+
+    def match_course_type(self, course, course_type, commit=False):
+        matches = {}
+
+        # First, easy exit if entitlements don't match.
+        course_entitlement_modes = set(course.entitlements.values_list('mode', flat=True))
+        type_entitlement_modes = set(course_type.entitlement_types.values_list('id', flat=True))
+        # Allow old courses without entitlements by checking if it has any first (should we do this?)
```
Dillon-Dumesnil: Yeah, I think we should definitely try to match as many courses as possible

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -0,0 +1,187 @@
...
+        mismatched_entitlements = course_entitlement_modes and course_entitlement_modes != type_entitlement_modes
+        mismatched_existing_course_type = course.type and course.type != course_type
+        if mismatched_entitlements or mismatched_existing_course_type:
+            if mismatched_entitlements and course.type and course.type == course_type:
+                logger.info(
+                    _("Existing course type {type} for {key} ({id}) doesn't match its own entitlements.").format(
+                        key=course.key, id=course.id, type=course.type.name,
```
Dillon-Dumesnil: Nit: Put these in order of how they're called in the string

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -0,0 +1,187 @@
...
+                    )
+                )
+            return False
+        if not course.type:
+            matches[course] = course_type
+
+        # Now, let's look at seat types too. If any of our CourseRunType children match a run, we'll take it.
+        for run in course.course_runs.order_by('key'):  # ordered just for visible message reliability
+            # Catch existing type data that doesn't match this attempted type
+            if run.type and run.type not in course_type.course_run_types.all():
```
Dillon-Dumesnil: Should we save `course_type.course_run_types.all()` to a variable or is it a cached value?

Dillon-Dumesnil (approved):
Don't think I have any blocking comments. I still want to think about/maybe talk through a bit more how this command looks on multiple runs of a course and how the log statements will work out. Not sure how noisy this job can become and I hesitate to have it log out too much and then it's hard to find meaning.

course_discovery/apps/course_metadata/models.py
mikix: Yeah, but I don't want a huge line - especially if it were to be translated.

They can just do `--help` if they want the goods.

course_discovery/apps/course_metadata/models.py
mikix: Fair... I'm open to changing it.

In my head, the CourseType and all types below it was one unit. So setting the CourseType was the same as setting all the rest. I know that's not exactly what the script does, but that was the "vibe" I was going for.

I don't know if I like `BackpopulateTypes` because it's a little vague. Programs have types. Seats have types. But we're not touching those...

`CourseAndCourseRunTypes` is a bit much too. I dunno either.

Dillon-Dumesnil: I'm okay with it. Having read more of the actual command, you can't set any of the run types without the course type so it really is coupled together. It's fine

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
mikix: The `distinct` is to filter out course objects that would match both `--course` and `--org`. Shouldn't affect drafts/official at all, since it goes off primary keys.

I figured we should change both drafts and officials? But open to a different approach there.

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
mikix: Fair... I was going for the same order of those values in all the `format` calls in the file, rather than per-string. But I'll go your way.

Dillon-Dumesnil: I'm a per-string kinda guy, but am open to anything. Tis a nit.

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
Dillon-Dumesnil: Ah I see, okay. Yeah, I'm good with it. I was just thinking the distinct might affect the draft/official because we are grabbing them on uuids and those match, but totally makes sense that the distinct won't be based on that.

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
mikix: I don't think it's cached - I think the `all()` prevents that. I'll move to a variable.

Dillon-Dumesnil: And in case it helps, you also call this once more slightly lower in the function.

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
mikix: I think those are two separate promises. I think it's clearer separated, personally.

PR #2209, opened 2019-10-07
don't copy url history backwards

rgraber:


course_discovery/apps/course_metadata/utils.py
```diff
@@ -127,6 +127,11 @@ def set_draft_state(obj, model, attrs=None):
 
     obj.save()
 
+    # must be done after save so we have an id
+    if related_attrs:
+        for key, value in related_attrs.items():
+            getattr(obj, key).set(value)
```
Dillon-Dumesnil, line 133: Any reason for the `getattr().set` as opposed to the `setattr`? Just curious if it was required or what.

Dillon-Dumesnil (approved):
LGTM

course_discovery/apps/course_metadata/utils.py
rgraber: Django complains about using setattr on a related field

PR #2203, opened 2019-10-04
Exclude Ended course runs from endpoint api/v1/search/all/

zamanafzal:
This Pr code aimed to remove the Ended/Expired Course Runs from the Endpoint
/api/v1/search/all/
You can test this scenario by
1- Creating an Enterprise Customer
2- Creating an Enterprise Customer Catalog
3- Creating a course with Ended Course Runs and some Active course Runs.
4- Associate this Course with enterprise on Via E-COMMERCE COURSE ADMINISTRATION Panel.
5- Then run these two cammands on discover shell
python manage.py refresh_course_metadata
python manage.py update_index --disable-change-limit
6- Please check on the endpoint mentioned above.
PRE-PR Results:
All course runs will be visible in results.
POST-PR Results:
Course runs which are expired will not be visible in results.

PR #2211, opened 2019-10-08
Exclude ended/expired course runs from endpoint

zamanafzal:
This PR removes the Ended/Expired Course Runs from the Endpoint.
`/api/v1/search/all/`
**Important Note:**
Expired Courses will be removed only when `exclude_expire_course_run` QUERY PARAM will be added in the endpoint.

You can test this scenario by
1- Creating an Enterprise Customer
2- Creating an Enterprise Customer Catalog
3- Creating a course with Ended Course Runs and some Active course Runs.
4- Associate this Course with enterprise on Via E-COMMERCE COURSE ADMINISTRATION Panel.
5- Then run these two commands on discover shell
python manage.py refresh_course_metadata
python manage.py update_index --disable-change-limit
6- Please check on the endpoint mentioned above.
**PRE-PR Results:**
All course runs will be visible in results.
**POST-PR Results**:
Course runs which are expired will not be visible in results.

[ENT-1470](https://openedx.atlassian.net/browse/ENT-1470)

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1808,14 +1808,10 @@ class CourseSearchSerializer(HaystackSerializer):
 
     def get_course_runs(self, result):
         request = self.context['request']
-        query_params = request.GET
-        exclude_expire_course_run = query_params.get("exclude_expire_course_run", None)
-        # Check if exclude_expire_course_run is in queryparams then exclude the course runs whose end date is passed.
-        if exclude_expire_course_run:
-            course_runs = [course_run for course_run in result.object.course_runs.exclude(
-                end__lte=datetime.datetime.now(pytz.UTC))]
-        else:
-            course_runs = [course_run for course_run in result.object.course_runs.all()]
+        course_runs = result.object.course_runs.all()
+        # Check if exclude_expire_course_run is in query_params then exclude the course runs whose end date is passed.
+        if request.GET.get("exclude_expired_course_run"):
+            course_runs = course_runs.exclude(end__lte=datetime.datetime.now(pytz.UTC))
         return [
             {
                 'key': course_run.key,
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1825,7 +1825,7 @@ class CourseSearchSerializer(HaystackSerializer):
                 'estimated_hours': get_course_run_estimated_hours(course_run),
                 'first_enrollable_paid_seat_price': course_run.first_enrollable_paid_seat_price or 0.0
             }
-            for course_run in result.object.course_runs.all()
+            for course_run in result.object.course_runs.exclude(end__lte=datetime.datetime.now(pytz.UTC))
         ]
 
     def get_seat_types(self, result):
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1807,6 +1807,15 @@ class CourseSearchSerializer(HaystackSerializer):
         return super().to_representation(instance)
 
     def get_course_runs(self, result):
+        request = self.context['request']
+        query_params = request.GET
+        exclude_expire_course_run = query_params.get("exclude_expire_course_run", None)
+        # Check if exclude_expire_course_run is in queryparams then exclude the course runs whose end date is passed.
+        if exclude_expire_course_run:
+            course_runs = [course_run for course_run in result.object.course_runs.exclude(
+                end__lte=datetime.datetime.now(pytz.UTC))]
+        else:
+            course_runs = [course_run for course_run in result.object.course_runs.all()]
         return [
             {
                 'key': course_run.key,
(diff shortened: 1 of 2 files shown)


muhammad-ammar (changes requested):
Looks great. Can we add another test with multiple courseruns, make some the courseruns expiry and then verify that expired courseruns are exlcluded but non-expired are returned.

jmyatt:
is this or #2203 the correct PR for this change?

muhammad-ammar (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -1825,7 +1825,7 @@ def get_course_runs(self, result):
                 'estimated_hours': get_course_run_estimated_hours(course_run),
                 'first_enrollable_paid_seat_price': course_run.first_enrollable_paid_seat_price or 0.0
             }
-            for course_run in result.object.course_runs.all()
+            for course_run in result.object.course_runs.exclude(end__lte=datetime.datetime.now(pytz.UTC))
```
waheedahmed: I believe there are other consumers as well of this endpoint (e.g. marketing search) where we need archived runs. Maybe there is already a filter available which you can use to filter out the archived runs otherwise create one.

waheedahmed (changes requested):

course_discovery/apps/api/serializers.py
zamanafzal: @waheedahmed I have updated the PR and steps (in PR description) to verify the change. Now course runs will be excluded on the basis of query param. Please take a look.

course_discovery/apps/api/serializers.py
```diff
@@ -1807,6 +1807,15 @@ def to_representation(self, instance):
         return super().to_representation(instance)
 
     def get_course_runs(self, result):
+        request = self.context['request']
```
waheedahmed, line 1810: Can following query serve the purpose instead of implementing new one?
https://discovery.edx.org/api/v1/search/all/?content_type=course&availability=Current&availability=Starting+Soon&availability=Upcoming

course_discovery/apps/api/serializers.py
```diff
@@ -1807,6 +1807,15 @@ def to_representation(self, instance):
...
+        query_params = request.GET
+        exclude_expire_course_run = query_params.get("exclude_expire_course_run", None)
+        # Check if exclude_expire_course_run is in queryparams then exclude the course runs whose end date is passed.
+        if exclude_expire_course_run:
+            course_runs = [course_run for course_run in result.object.course_runs.exclude(
+                end__lte=datetime.datetime.now(pytz.UTC))]
+        else:
+            course_runs = [course_run for course_run in result.object.course_runs.all()]
```
muhammad-ammar: I think we don't need to iterate queryset here. Below code should serve the purpose.

```python
course_runs = result.object.course_runs.all() 
if request.GET.get("exclude_expired_course_run"):
    course_runs = course_runs.exclude(end__lte=datetime.datetime.now(pytz.UTC))
```

course_discovery/apps/api/serializers.py
```diff
@@ -1807,6 +1807,15 @@ def to_representation(self, instance):
         return super().to_representation(instance)
 
     def get_course_runs(self, result):
+        request = self.context['request']
+        query_params = request.GET
+        exclude_expire_course_run = query_params.get("exclude_expire_course_run", None)
```
muhammad-ammar: no need to create an extra variable,  we can do

```python
if request.GET.get("exclude_expired_course_run"):
```

also `dict.get` returns `None` as the default value so we don't need to.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -58,9 +58,9 @@ def json_date_format(datetime_obj):
     return datetime_obj and datetime.datetime.strftime(datetime_obj, "%Y-%m-%dT%H:%M:%S.%fZ")
 
 
-def make_request():
+def make_request(query_param=None):
     user = UserFactory()
-    request = APIRequestFactory().get('/')
+    request = APIRequestFactory().get('/', query_param)
```
muhammad-ammar: we should not add `query_param` in request if `query_param` is `None`.

course_discovery/apps/api/serializers.py
muhammad-ammar: `exclude_expire_course_run` -> `exclude_expired_course_run`

muhammad-ammar (changes requested):

course_discovery/apps/api/serializers.py
zamanafzal: Yeah, there is no need to iterate it. I will remove it. After removal this shouldn't be like
`if request.GET.get("exclude_expired_course_run"):
    course_runs = result.object.course_runs.exclude(end__lte=datetime.datetime.now(pytz.UTC))
else:
course_runs = result.object.course_runs.all() 
`
in this case we will get the desired data only and no need to filter it again if we need to Exclude Expired Course Runs

muhammad-ammar (approved):
LGTM

PR #2213, opened 2019-10-09
Adding path for using CourseRunType for Seat management

Dillon-Dumesnil:
This adds in functionality to the course and course run
end points to use the course run type model (and through it,
the Track and SeatType models) to inform how to manage Seats
for a course run.

DISCO-1379

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -251,7 +251,7 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         # Note: We have to send the request object as well because it is used for its metadata
         # (like request.user and is set as part of the serializer context)
         if course_run_creation_fields:
-            course_run_creation_fields.update({'course': course.key})
+            course_run_creation_fields.update({'course': course.key, 'price': price})
             run_response = CourseRunViewSet().create_run_helper(course_run_creation_fields, request)
             if run_response.status_code != 201:
                 raise Exception(str(run_response.data))
@@ -318,10 +318,10 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         entitlements = []
         # New flow for courses using CourseType model
         # DISCO-1399: I think pretty much the whole 'else' can be removed.
+        price = data.get('price', 0.00)
         if data.get('type'):
             course_type = CourseType.objects.get(uuid=data.get('type'))
             entitlement_types = course_type.entitlement_types.all()
-            price = data.get('price', 0.00)
             for entitlement_type in entitlement_types:
                 data = {'mode': entitlement_type.slug, 'price': price}
                 entitlements.append(self.update_entitlement(course, data, entitlement_type, partial=partial))
@@ -337,9 +337,12 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
                 )
         if entitlements:
             course.entitlements.set(entitlements)
-            # If entitlements were updated, we also want to update seats
-            for course_run in course.active_course_runs:
-                course_run.update_or_create_seats()
+            # DISCO-1399: This whole if can be removed. Updating or creating seats
+            # is being moved to the course run endpoint.
+            if not course.type:
+                # If entitlements were updated, we also want to update seats
+                for course_run in course.active_course_runs:
+                    course_run.update_or_create_seats(course_run.type, price)
 
         # Save video if a new video source is provided
         if (video_data and video_data.get('src') and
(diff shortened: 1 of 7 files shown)


Dillon-Dumesnil:
I think I covered the cases we need to support in my tests, but if there are any other scenarios you can think of that should be tested, I'm happy to update.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -339,7 +339,11 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
             course.entitlements = entitlements
             # If entitlements were updated, we also want to update seats
             for course_run in course.active_course_runs:
-                course_run.update_or_create_seats()
+                run_type_uuid = None
+                # DISCO-1399: Won't need the if anymore
```
mikix: But will need the stuff inside the `if`, yah? Maybe could be clearer, but super nit.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -339,7 +339,11 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
...
+                if data.get('type'):
```
mikix: Should this be `if course_run.type` instead? (To be more defensive around serializer being busted and to handle existing type values?)

course_discovery/apps/course_metadata/models.py
```diff
@@ -1434,68 +1434,106 @@ def search(cls, query):
...
+        # In the CourseType flow, we only send in a single price for all of the tracks (for now),
+        # so it would be possible to accidentally set an Audit seat to have a nonzero price because
+        # there is a Verified seat with a price.
+        if seat_type == Seat.AUDIT:
             price = 0.00
 
         seat, __ = Seat.everything.update_or_create(  # pylint: disable=no-member
             course_run=self,
-            type=mode,
+            type=seat_type,
```
mikix, line 1477: Nice.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1434,68 +1434,106 @@ def search(cls, query):
...
-                draft=True,
-                defaults={
-                    'price': 0.00,
-                    'upgrade_deadline': None,
-                },
-            )
-            seats.append(audit_seat)
+        return seat
+
+    def update_or_create_seats(self, run_type_uuid=None, price=0.00):
```
mikix: Why not just `run_type=None` and take an object? Would let you simplify the `if run.type` above and just pass in `run.type`. You'd have to do a `get()` in the create call, but that's not so bad...

course_discovery/apps/course_metadata/models.py
```diff
@@ -1434,68 +1434,106 @@ def search(cls, query):
...
+        course run exists. After an official version of the course run exists, it only supports
+        price changes or upgrades from Audit -> Verified.
+        """
+        seats = []
+        # DISCO-1399: Shouldn't need the whole if/else anymore. Can also look into if it's necessary
+        # to keep all of the helper functions (helper, validate, and upgrade deadline)
+        if run_type_uuid:
+            run_type = CourseRunType.objects.get(uuid=run_type_uuid)
+            seat_types = []
+            for track in run_type.tracks.filter(seat_type__isnull=False):
```
mikix: Could also `exclude(seat_type=None)` which *might* be clearer than filtering on a null being false. But it's a toss-up.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1434,68 +1434,106 @@ def search(cls, query):
...
+                    draft=True,
+                    defaults={
+                        'price': 0.00,
+                        'upgrade_deadline': None,
+                    },
+                )
+                seats.append(audit_seat)
+            else:
+                self.seats.exclude(type=seat_type).delete()
         self.seats.set(seats)
```
mikix: Does this line do anything? This is the wrong side of the relation to set yah? They are already in `self.seats` from pointing at self? Unless this is like a `refresh_from_db` but for one field?

mikix (approved):

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -337,9 +337,12 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
                 )
         if entitlements:
             course.entitlements.set(entitlements)
-            # If entitlements were updated, we also want to update seats
-            for course_run in course.active_course_runs:
-                course_run.update_or_create_seats()
+            # DISCO-1399: This whole if can be removed. Updating or creating seats
+            # is being moved to the course run endpoint.
+            if not data.get('type'):
```
mikix: This might be better as `if not course.type`?

PR #2215, opened 2019-10-09
DISCO-1389: Will not create children without parent IDs in Salesforce

jlajoie:


course_discovery/apps/course_metadata/salesforce.py
```diff
@@ -150,60 +159,69 @@ def create_publisher_organization(self, organization):
     def create_course(self, course):
         if not course.salesforce_id:
             organization = course.authoring_organizations.first()
-            if organization and not organization.salesforce_id:
-                self.create_publisher_organization(organization)
-            sf_course = self.client.Course__c.create(
-                self._build_course_payload(course, organization)
-            )
-            course.salesforce_id = sf_course.get('id')
-            course.save()
```
xnick421, line 159: this seems crazy that we had the logic like this in the first place, trying to think of an edge case which may have led us to having this?

xnick421 (approved):

course_discovery/apps/course_metadata/salesforce.py
jlajoie: I think it was just us being optimistic that it'd always work, or it'd error out above it.

PR #2217, opened 2019-10-10
DISCO-1388 return course_uuid in course run serializer

xnick421:


course_discovery/apps/api/serializers.py
```diff
@@ -739,6 +739,7 @@ def validate(self, attrs):
 class CourseRunSerializer(MinimalCourseRunSerializer):
     """Serializer for the ``CourseRun`` model."""
     course = serializers.SlugRelatedField(required=True, slug_field='key', queryset=Course.objects.filter_drafts())
+    course_uuid = serializers.ReadOnlyField(source='course.uuid')
```
Dillon-Dumesnil: Why do we need to add this into the serializer? Can't we just grab this info from off the course object we already serialize as part of the course run?

xnick421: the `course` field above this line you mean? thats a slug

xnick421: additionally, the front end is passing a `fields=course_uuid` param, so the payload is very small

Dillon-Dumesnil: Yeah, I saw that once I saw the frontend PR. Okay. Is this going to have any issues if course isn't set yet? Like is this going to error out or anything with a new course run when it is trying to source `course.uuid`, but `course` isn't set yet.

Dillon-Dumesnil (approved):

course_discovery/apps/api/serializers.py
xnick421: prob

mikix (commented):
Is there not a test that needs updating?

PR #2221, opened 2019-10-10
DISCO temp disable delete orphans in studio loader

xnick421:


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -71,10 +72,11 @@ def update_organization(self, body):
                 'logo_image_url': logo,
             })
 
-        org, _ = Organization.objects.update_or_create(key__iexact=key, partner=self.partner, defaults=defaults)
+        Organization.objects.update_or_create(key__iexact=key, partner=self.partner, defaults=defaults)
 
-        if org:
-            self.loaded_org_pks.add(org.pk)
+        # TODO add this back in when loading from drupal is completely removed
+        # if org:
```
mikix, line 78: When you add this back, I don't think you need this `if` -- `update_or_create` can't return None, I don't think. If it has a problem, it raises.

mikix (approved):

PR #2224, opened 2019-10-11
Initial rows for CourseType, CourseRunType, Track, and Mode

Dillon-Dumesnil:
DISCO-1382

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
@@ -6,7 +6,7 @@ from testfixtures import LogCapture, StringComparison
 
 from course_discovery.apps.course_metadata.management.commands.backpopulate_course_type import logger
 from course_discovery.apps.course_metadata.models import (
-    BackpopulateCourseTypeConfig, Course, CourseRunType, CourseType, Seat, SeatType
+    BackpopulateCourseTypeConfig, Course, CourseRunType, CourseType, Mode, Seat, SeatType, Track
 )
 from course_discovery.apps.course_metadata.tests import factories
 from course_discovery.apps.course_metadata.utils import ensure_draft_world
@@ -23,16 +23,14 @@ class BackpopulateCourseTypeCommandTests(TestCase):
         # Fill out a bunch of types and modes. Exact mode parameters don't matter, just the resulting seat types.
         self.audit_seat_type = SeatType.objects.get(slug=Seat.AUDIT)
         self.verified_seat_type = SeatType.objects.get(slug=Seat.VERIFIED)
-        self.audit_mode = factories.ModeFactory(name='Audit')
-        self.verified_mode = factories.ModeFactory(name='Verified')
-        self.audit_track = factories.TrackFactory(seat_type=self.audit_seat_type, mode=self.audit_mode)
-        self.verified_track = factories.TrackFactory(seat_type=self.verified_seat_type, mode=self.verified_mode)
-        self.audit_run_type = factories.CourseRunTypeFactory(name='Audit Only', tracks=[self.audit_track])
-        self.va_run_type = factories.CourseRunTypeFactory(name='Verified & Audit',
-                                                          tracks=[self.audit_track, self.verified_track])
-        self.va_course_type = factories.CourseTypeFactory(name='Verified & Audit',
-                                                          entitlement_types=[self.verified_seat_type],
-                                                          course_run_types=[self.audit_run_type, self.va_run_type])
+        self.audit_mode = Mode.objects.get(slug=Seat.AUDIT)
+        self.verified_mode = Mode.objects.get(slug=Seat.VERIFIED)
+        self.audit_track = Track.objects.get(seat_type=self.audit_seat_type, mode=self.audit_mode)
+        self.verified_track = Track.objects.get(seat_type=self.verified_seat_type, mode=self.verified_mode)
+        self.audit_run_type = CourseRunType.objects.get(slug=CourseRunType.AUDIT)
+        self.va_run_type = CourseRunType.objects.get(slug=CourseRunType.VERIFIED_AUDIT)
+        self.va_course_type = CourseType.objects.get(slug=CourseType.VERIFIED_AUDIT)
+        self.audit_course_type = CourseType.objects.get(slug=CourseType.AUDIT)
 
         # Now create some courses and orgs that will be found to match the above, in the simple happy path case.
         self.org = factories.OrganizationFactory(partner=self.partner, key='Org1')
@@ -140,7 +138,7 @@ class BackpopulateCourseTypeCommandTests(TestCase):
         self.assertIsNotNone(self.course.type)
 
     def test_normal_run(self):
-        self.run_command(log='Course .* matched type Verified & Audit')
+        self.run_command(log='Course .* matched type Verified and Audit')
         self.assertEqual(self.course.type, self.va_course_type)
         self.assertEqual(self.audit_run.type, self.audit_run_type)
         self.assertEqual(self.verified_run.type, self.va_run_type)
@@ -193,10 +191,13 @@ class BackpopulateCourseTypeCommandTests(TestCase):
                          {self.audit_run_type.id, self.va_run_type.id})
 
     def test_matches_earliest_course_type(self):
+        # This messes up this test due to Verified and Audit being a subset of Credit
+        # and Credit being created prior to 'Second'
+        CourseType.objects.get(slug=CourseType.CREDIT_VERIFIED_AUDIT).delete()
         second_type = factories.CourseTypeFactory(
             name='Second',
-            entitlement_types=self.va_course_type.entitlement_types.all(),  # pylint: disable=no-member
-            course_run_types=self.va_course_type.course_run_types.all(),  # pylint: disable=no-member
+            entitlement_types=self.va_course_type.entitlement_types.all(),
+            course_run_types=self.va_course_type.course_run_types.all(),
         )
 
         self.run_command()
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,134 @@
...
+    #     help_text=_('Completion can grant credit toward an organization’s degree.'),
+    # )
+    # certificate_type = models.CharField(
+    #     max_length=64, choices=CertificateType, blank=True,
+    #     help_text=_('Certificate type granted if this mode is eligible for a certificate, or blank if not.'),
+    # )
+    # payee = models.CharField(
+    #     max_length=64, choices=PayeeType, default=PayeeType.Platform,
+    #     help_text=_('Who gets paid for the course? Platform is the site owner, Organization is the school.'),
+    # )
```
Dillon-Dumesnil: I'll remove all the comments after review. Just put them in there for now so it's easier to see what all the possible fields are for these models. The fields are below the actual model variable declaration.

course_discovery/apps/course_metadata/models.py
```diff
@@ -259,7 +259,7 @@ class Mode(TimeStampedModel):
         help_text=_('Certificate type granted if this mode is eligible for a certificate, or blank if not.'),
     )
     payee = models.CharField(
-        max_length=64, choices=PayeeType, default=PayeeType.Platform,
+        max_length=64, choices=PayeeType, default=PayeeType.Platform, blank=True,
```
mikix: It's possible it might be nice to add an `@property` like `def is_paid():` or something to check a blank value here (as we have for certificate type). Not a blocker for this PR, just something to think about as we use these.

course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,134 @@
...
+        'name': 'Verified',
+        'is_id_verified': True,
+        'certificate_type': CertificateType.Verified,
+    }),
+)
+# The slugs are identical across Mode and SeatType
+TRACK_MODE_AND_SEAT_SLUGS = ('audit', 'credit', 'professional', 'verified',)
+COURSE_RUN_TYPE_NAMES_AND_TRACK_SLUGS = (
+    ('Audit Only', ['audit']),
+    ('Credit', ['audit', 'credit', 'verified']),
```
jmyatt: Julia confirmed that currently, all runs for Credit courses have all 3 of these track types

course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,134 @@
...
+MODE_SLUGS_AND_DEFAULTS = (
+    ('audit', {
+        'name': 'Audit',
+        'payee': '',
+    }),
+    ('credit', {
+        'name': 'Credit',
+        'is_id_verified': True,
+        'is_credit_eligible': 1,
+        'certificate_type': CertificateType.Credit,
```
mikix: Ideally we don't have to hard-import CetificateType. So maybe hardcode the string here, or move these definitions into the method where you can do like `apps.get_model`

course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,134 @@
...
+from django.db import migrations, models
+
+from course_discovery.apps.course_metadata.choices import CertificateType, PayeeType
+
+SEAT_TYPES = ('Audit', 'Credit', 'Professional', 'Verified',)
+# Contains the slug (unique identifier) and any fields that are not the default values
+MODE_SLUGS_AND_DEFAULTS = (
+    ('audit', {
+        'name': 'Audit',
+        'payee': '',
```
mikix: Is it weird that payee doesn't default to this?

course_discovery/apps/course_metadata/models.py
Dillon-Dumesnil: Yeah, I feel like we should maybe wait and see if that is needed before adding it. I'm not sure of the use case yet.

course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,142 @@
...
+        mode = Mode.objects.get(slug=slug)
+        seat_type = SeatType.objects.get(slug=slug)
+        Track.objects.update_or_create(mode=mode, seat_type=seat_type)
+    for name, track_slugs in COURSE_RUN_TYPE_NAMES_AND_TRACK_SLUGS:
+        tracks = []
+        for slug in track_slugs:
+            mode = Mode.objects.get(slug=slug)
+            seat_type = SeatType.objects.get(slug=slug)
+            tracks.append(Track.objects.get(mode=mode, seat_type=seat_type))
+        run_type, __ = CourseRunType.objects.update_or_create(name=name)
```
mikix: Why a dunder? One should be enough?

course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,142 @@
...
+        run_type.tracks.set(tracks)
+    for course_name, seat_type_slugs, run_type_names in COURSE_TYPE_NAMES_AND_SEAT_TYPE_SLUGS_AND_RUN_TYPE_NAMES:
+        entitlement_types = []
+        for slug in seat_type_slugs:
+            entitlement_types.append(SeatType.objects.get(slug=slug))
+        run_types = []
+        for run_name in run_type_names:
+            run_types.append(CourseRunType.objects.get(name=run_name))
```
mikix: Maybe:
```
entitlement_types = {SeatType.objects.get(slug=slug) for slug in seat_type_slugs}
run_types = {CourseRunType.objects.get(name=run_name) for run_name in run_names}
```

mikix: When grabbing CourseRunType by name, maybe do a `filter().first()`, since technically there could be multiple if someone made one after migration that made the model, but before this one...

I guess similar problem right below on the CourseType update_or_create... I dunno what to do about it.  I guess you could keep a live mapping of created-UUIDs to the names assigned. But meh.

Maybe just leave it all as-is.

mikix (approved):
Nice.

course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
```diff
@@ -23,16 +23,14 @@ def setUp(self):
...
-                                                          entitlement_types=[self.verified_seat_type],
-                                                          course_run_types=[self.audit_run_type, self.va_run_type])
+        self.audit_mode = Mode.objects.get(slug=Seat.AUDIT)
+        self.verified_mode = Mode.objects.get(slug=Seat.VERIFIED)
+        self.audit_track = Track.objects.get(seat_type=self.audit_seat_type, mode=self.audit_mode)
+        self.verified_track = Track.objects.get(seat_type=self.verified_seat_type, mode=self.verified_mode)
+        self.audit_run_type = CourseRunType.objects.get(name='Audit Only')
+        self.va_run_type = CourseRunType.objects.get(name='Verified and Audit')
+        self.va_course_type = CourseType.objects.get(name='Verified and Audit')
+        self.audit_course_type = CourseType.objects.get(name='Audit Only')
```
mikix: This kinda thing is annoying...  We should maybe add a slug field rather than/additionally to the model? Or make `name` unique, but that feels wrong to me.

course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
```diff
@@ -202,6 +200,9 @@ def test_matches_earliest_course_type(self):
         self.run_command()
         self.assertEqual(self.course.type, self.va_course_type)
 
+        # This messes up this test due to Verified and Audit being a subset of Credit
+        # and Credit being created prior to 'Second'
+        CourseType.objects.get(name='Credit').delete()
```
mikix: I would move this to the top of the test function, so it's true for all the mini-tests I do.

course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
```diff
@@ -223,39 +224,32 @@ def test_one_mismatched_run_ruins_whole_thing(self):
     def test_by_org(self):
         self.run_command(orgs=[self.org2])
         self.assertIsNone(self.course.type)
-        self.assertEqual(self.course2.type, self.va_course_type)
+        self.assertEqual(self.course2.type, self.audit_course_type)
```
mikix, line 227: Wait, why these changes? Course has runs with verified seats, yah?

course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
```diff
@@ -223,39 +224,32 @@ def test_one_mismatched_run_ruins_whole_thing(self):
...
         self.assertEqual(draft_course.entitlements.first().mode.slug, Seat.AUDIT)  # made a draft audit entitlement
         self.assertIsNone(self.course2.entitlements.first())  # but official is still empty
 
-        # Will fail without an audit-only CourseType
-        self.run_command(courses=[self.course2], fails=draft_course)  # self.course2 passes with V&A type
-
-        # Now make the missing audit-only CourseType and try again
-        audit_course_type = factories.CourseTypeFactory(name='Audit Only', entitlement_types=[],
-                                                        course_run_types=[self.audit_run_type])
-
```
mikix: I would prefer deleting the audit course type first before this stuff (and credit type?), so we can still test that we fail without audit-only.

course_discovery/apps/course_metadata/management/commands/tests/test_backpopulate_course_type.py
Dillon-Dumesnil: Nah, only audit

mikix (approved):
Nice! Slugs do look nice.

course_discovery/apps/course_metadata/migrations/0204_create_initial_coursetypes.py
```diff
@@ -0,0 +1,138 @@
...
+COURSE_RUN_TYPE_NAMES_SLUGS_AND_TRACK_SLUGS = (
+    ('Audit Only', 'audit', ['audit']),
+    ('Professional Only', 'professional', ['professional']),
+    ('Verified and Audit', 'verified-and-audit', ['audit', 'verified']),
+    ('Credit', 'credit', ['audit', 'credit', 'verified']),
+)
+COURSE_TYPE_NAMES_SLUGS_SEAT_TYPE_SLUGS_AND_RUN_TYPE_NAMES = (
+    ('Audit Only', 'audit', ['audit'], ['Audit Only']),
+    ('Professional Only', 'professional', ['professional'], ['Professional Only']),
+    ('Verified and Audit', 'verified-and-audit', ['verified'], ['Audit Only', 'Verified and Audit']),
```
mikix: Could go `verified` instead? Like, `credit` doesn't include other subtypes too. But super nit.

Dillon-Dumesnil: I wanted to keep this that way in case we introduce a verified only version one day. But don't disagree that including the subtypes could be problematic as we continue to increase.

`masters-verified-audit` here we come

PR #2227, opened 2019-10-16
DISCO-1389: Will not create children without parent IDs in Salesforce

jlajoie:


course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -822,21 +822,25 @@ def test_update_or_create_salesforce_course(self):
         with mock.patch(self.salesforce_util_path + '.__init__', return_value=None):
             with mock.patch(self.salesforce_util_path + '.create_course') as mock_create_method:
                 with mock.patch(self.salesforce_util_path + '.update_course') as mock_update_method:
-                    course = factories.CourseFactory(draft=True)
+                    with mock.patch(self.salesforce_util_path + '.create_publisher_organization'):
```
xnick421: what happens if we just mock the base salesforce_util_path instead of each individual method.   are the methods still available to call?

xnick421 (commented):
just quick questions about how we are mocking stuff

xnick421 (approved):
since we can get rid of all those mocks, i think its ready once they are all refactored 👍

PR #2228, opened 2019-10-16
DISCO-1397 Organization group users 404

xnick421:
Instead of returning a 404 error when an OrganizationExtension does
not exist for an organization, we return an empty queryset.

### DIFF
--- a/course_discovery/apps/publisher/api/tests/test_views.py
+++ b/course_discovery/apps/publisher/api/tests/test_views.py
@@ -196,10 +196,11 @@ class OrganizationGroupUserViewTests(APITestCase):
 
     def test_get_organization_not_found(self):
         """
-        Verify that view returns status=404 if organization is not found in OrganizationExtension.
+        Verify that view returns an empty list if organization is not found in OrganizationExtension.
         """
         response = self.query(org_id=0)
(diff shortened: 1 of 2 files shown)


jlajoie (commented):
Feels weird to me that we're swapping this from a 404 to a [] + 200, I think I'm missing some context? Typically I feel like if you're querying for something and don't get data, that's a not found type response?

xnick421:
You're right, the issue is though when the front end is on the course detail page, it hits this view in order to get a list if users that could potentially be assigned as a course editor.  In the case where the course has multiple organizations, and one organization does not have an OrganizationExtension set up, it raises a 404, which in turn causes the FE to break and show no users at all.

xnick421:
another approach would be to have the FE do all the logic, but i didn't think having the api return an empty list was necessarily a bad thing

jlajoie:
I feel like that's fine though, we just need to change our handling of it? Like in the promise on the frontend, instead of raising any failed request and failing the bunch, just use the ones that are available? 

This seems like a breaking enough change that it'd require a version bump which we don't have a plan for, as anyone who'd previously been expecting the 404s would now be operating off of empty lists. I'd say just handle it on the FE and treat 404s as empty arrays.

PR #2229, opened 2019-10-16
Backfill drupal redirects into new relationship on course

rgraber:


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -237,3 +243,83 @@ def process_node(self, data):
...
+        super().__init__(*args, **kwargs)
+        self.load_redirect_data()
+
+    def load_redirect_data(self):
+        with open(DRUPAL_REDIRECT_CSV_FILE) as redirect_csv:
+            reader = csv.DictReader(redirect_csv, fieldnames=('redirect_url', 'node', 'redirect_type'))
+            # Order data so that we can pull it out via node ids
+            for row in reader:
+                # Node id in CSV is stored as 'node/1234'.
+                # Ignore anything that doesn't match.
```
albemarle, line 266: Should we log if we find weird stuff?

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -237,3 +243,83 @@ def process_node(self, data):
...
+        try:
+            course = Course.objects.get(key__iexact=course_key)
+        except Course.DoesNotExist:
+            logger.info('No course found for %s', course_key)
+            return
+
+        for redirect_row in node_redirects:
+            redirect = redirect_row['redirect_url']
+            matched = self.standard_course_url_regex.match(redirect)
+            if matched:
```
albemarle, line 300: Same question, should we log if it doesn't parse?

albemarle (approved):

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
rgraber: No because that just means it's a different format of URL, so we'll save it in the url_redirects instead of url_slug_history

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
rgraber: I don't think so. Drupal has all kinds of weirdnesses in its data and it's doubtful that even if we find weird stuff we would want to do anything about it. This is a list of all the redirects (not just courses) so it's pretty likely there's junk in there that we don't care about.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
rgraber: No, but it's worth a code comment. If it doesn't parse it will just go into url_redirects instead of url_slug_history

PR #2230, opened 2019-10-16
Restriction on the version of MySql that we use for testing

staubina:
[PROD-830]

This is being added to deal with a failure in tests running with the
current version.

[PROD-830]: https://openedx.atlassian.net/browse/PROD-830


jlajoie (approved):
Thanks for getting this in 👍

PR #2232, opened 2019-10-16
WEBSITE-145: Add advertised_course_run_uuid field to api/v1/courses r…

reillz10:
…esponse

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1021,6 +1021,7 @@ class CourseEditorSerializer(serializers.ModelSerializer):
 
 class CourseWithProgramsSerializer(CourseSerializer):
     """A ``CourseSerializer`` which includes programs."""
+    advertised_course_run_uuid = serializers.SerializerMethodField()
     course_run_keys = serializers.SerializerMethodField()
     course_runs = serializers.SerializerMethodField()
     programs = NestedProgramSerializer(read_only=True, many=True)
@@ -1054,6 +1055,64 @@ class CourseWithProgramsSerializer(CourseSerializer):
             Prefetch('entitlements', queryset=CourseEntitlementSerializer.prefetch_queryset()),
         )
 
+    def get_advertised_course_run_uuid(self, course):
+        def get_verified_upgrade_deadline_for_course_run(course_run):
+            verified_seat_upgrade_deadline = None
+
+            for seat in course_run.seats.all():
+                if seat.type == Seat.VERIFIED:
+                    verified_seat_upgrade_deadline = seat.upgrade_deadline
+                    break
+
+            return verified_seat_upgrade_deadline
+
+        def is_upgradeable(course_run):
+            upgrade_deadline = get_verified_upgrade_deadline_for_course_run(course_run)
+            today_midnight = datetime.datetime.now(pytz.UTC).replace(hour=0, minute=0, second=0, microsecond=0)
+
+            if upgrade_deadline:
+                return upgrade_deadline >= today_midnight
+
+            return True
+
+        def user_has_time_to_finish_course(course_run):
+            # Users have time to finish course if purchasing verified cert
+            two_weeks_from_now = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=14)
+            two_weeks_from_now = two_weeks_from_now.replace(hour=0, minute=0, second=0, microsecond=0)
+
+            try:
+                return course_run.end >= two_weeks_from_now
+            except TypeError:
+                return True
+
+        def has_started(course_run):
+            return datetime.datetime.now(pytz.UTC) >= course_run.start
+
+        tier_one = []
+        tier_two = []
+        tier_three = []
+
+        marketable_course_runs = [course_run for course_run in course.course_runs.all() if course_run.is_marketable]
+
+        for course_run in marketable_course_runs:
+            if has_started(course_run) and is_upgradeable(course_run) and user_has_time_to_finish_course(course_run):
+                tier_one.append(course_run)
+            elif not has_started(course_run) and is_upgradeable(course_run):
+                tier_two.append(course_run)
+            else:
+                tier_three.append(course_run)
+
+        advertised_course_run_uuid = None
+
+        if tier_one:
+            advertised_course_run_uuid = sorted(tier_one, key=attrgetter('start'), reverse=True)[0].uuid
+        elif tier_two:
+            advertised_course_run_uuid = sorted(tier_two, key=attrgetter('start'))[0].uuid
+        elif tier_three:
+            advertised_course_run_uuid = sorted(tier_three, key=attrgetter('start'), reverse=True)[0].uuid
+
+        return advertised_course_run_uuid
+
     def get_course_run_keys(self, course):
         return [course_run.key for course_run in course.course_runs.all()]
 
@@ -1076,6 +1135,7 @@ class CourseWithProgramsSerializer(CourseSerializer):
             'programs',
             'course_run_keys',
             'editable',
+            'advertised_course_run_uuid'
         )
 
 
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -149,6 +149,7 @@ class CourseSerializerTests(MinimalCourseSerializerTests):
     @classmethod
     def get_expected_data(cls, course, request):
         expected = super().get_expected_data(course, request)
+
         expected.update({
             'short_description': course.short_description,
             'full_description': course.full_description,
@@ -258,6 +259,9 @@ class CourseEditorSerializerTests(TestCase):
 @ddt.ddt
 class CourseWithProgramsSerializerTests(CourseSerializerTests):
     serializer_class = CourseWithProgramsSerializer
+    YESTERDAY = datetime.datetime.now(UTC) - datetime.timedelta(days=1)
+    TOMORROW = datetime.datetime.now(UTC) + datetime.timedelta(days=1)
+    TWO_WEEKS_FROM_TODAY = datetime.datetime.now(UTC) + datetime.timedelta(days=14)
 
     @classmethod
     def get_expected_data(cls, course, request):
@@ -270,9 +274,24 @@ class CourseWithProgramsSerializerTests(CourseSerializerTests):
             ).data,
             'course_run_keys': [course_run.key for course_run in course.course_runs.all()],
             'editable': False,
+            'advertised_course_run_uuid': None,
         })
 
         return expected
+
+    def create_upgradeable_seat_for_course_run(self, course_run):
+        return SeatFactory(
+            course_run=course_run,
+            type='verified',
+            upgrade_deadline=self.TOMORROW
+        )
+
+    def create_not_upgradeable_seat_for_course_run(self, course_run):
+        return SeatFactory(
+            course_run=course_run,
+            type='verified',
+            upgrade_deadline=datetime.datetime(2014, 1, 1, tzinfo=UTC)
+        )
 
     def setUp(self):
         super().setUp()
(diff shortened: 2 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
+        def is_upgradeable(course_run):
+            upgrade_deadline = get_verified_upgrade_deadline_for_course_run(course_run)
+            today_midnight = datetime.datetime.now(pytz.UTC).replace(hour=0, minute=0, second=0, microsecond=0)
+
+            if (upgrade_deadline):
+                return upgrade_deadline >= today_midnight
+
+            return True
+
+        def has_time_to_finish(course_run):
```
albemarle: Is this not already a function somewhere else that we can reuse?

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
             Prefetch('entitlements', queryset=CourseEntitlementSerializer.prefetch_queryset()),
         )
 
+    def get_advertised_course_run_uuid(self, course):
+        def enrollment_start(course_run):
+            return course_run.start
+
+        def get_verified_upgrade_deadline_for_course_run(course_run):
```
rgraber, line 1059: All of these course run methods feel like they would make more sense on the course run model rather than in the serializer

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
+
+        for course_run in course.course_runs.all():
+            if has_started(course_run) and is_upgradeable(course_run) and has_time_to_finish(course_run):
+                tier_one.append(course_run)
+            elif not has_started(course_run) and is_upgradeable(course_run):
+                tier_two.append(course_run)
+            else:
+                tier_three.append(course_run)
+
+        active_course_run_uuid = None
```
rgraber: might as well change this to 'advertised_course_run_uuid' in the method as well

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
+
+        def has_started(course_run):
+            return datetime.datetime.now(pytz.UTC) >= course_run.start
+
+
+        tier_one = []
+        tier_two = []
+        tier_three = []
+
+        for course_run in course.course_runs.all():
```
rgraber: this should be limited to all marketable course runs. idk if you can do course.course_runs.filter(is_marketable=True) on a derived property so you may have to do it manualy

rgraber (changes requested):

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -285,6 +291,94 @@ def test_data(self):
...
+
+        expected_advertised_course_run = CourseRunFactory(course=self.course, start=self.YESTERDAY, end=self.TWO_WEEKS_FROM_TODAY)
+
+        not_upgradeable_course_run = CourseRunFactory(course=self.course, start=self.YESTERDAY, end=self.TWO_WEEKS_FROM_TODAY)
+        SeatFactory(course_run=not_upgradeable_course_run, type='verified', upgrade_deadline=datetime.datetime(2014, 1, 1, tzinfo=UTC))
+
+        for i in range(4):
+           CourseRunFactory(course=self.course, start=start_days[i], end=end_days[i])
+
+        expected = self.get_expected_data(self.course, self.request)
```
rgraber: do you really need the full expected object if you're just comparing the advertised course run uuid?

course_discovery/apps/api/serializers.py
reillz10: I don't think that it is, this business logic is brand new for course-discovery and used to only exist in the prospectus frontend. That said, I will look around and see if there is an existing function somewhere

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
+                tier_one.append(course_run)
+            elif not has_started(course_run) and is_upgradeable(course_run):
+                tier_two.append(course_run)
+            else:
+                tier_three.append(course_run)
+
+        active_course_run_uuid = None
+
+        if len(tier_one) > 0:
+            active_course_run_uuid = sorted(tier_one, key=enrollment_start, reverse=True)[0].uuid
```
mikix: `key=attrgetter('start')` is the more pythonic way to do this.

https://docs.python.org/3.5/howto/sorting.html#operator-module-functions

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
 
+    def get_advertised_course_run_uuid(self, course):
+        def enrollment_start(course_run):
+            return course_run.start
+
+        def get_verified_upgrade_deadline_for_course_run(course_run):
+            verified_seat_upgrade_deadline = None
+
+            for seat in course_run.seats.all():
+                if seat.type == 'verified':
```
mikix: Nano-nit: `Seat.VERIFIED`

But more broadly - why only verified? Professional and credit have upgrade deadlines too. Should this whole method just be replaced by a call to `CourseRun.get_paid_seat_enrollment_end`?

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
+                    break;
+
+            return verified_seat_upgrade_deadline
+
+        def is_upgradeable(course_run):
+            upgrade_deadline = get_verified_upgrade_deadline_for_course_run(course_run)
+            today_midnight = datetime.datetime.now(pytz.UTC).replace(hour=0, minute=0, second=0, microsecond=0)
+
+            if (upgrade_deadline):
+                return upgrade_deadline >= today_midnight
```
mikix, line 1074: Why midnight? Is that based off the expected jenkins job run times? If so, that's an odd thing to bake into the Open edX course-discovery repo.

course_discovery/apps/api/serializers.py
```diff
@@ -1052,6 +1053,66 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None, programs=No
...
+
+            return True
+
+        def has_time_to_finish(course_run):
+            two_weeks_from_now = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=14)
```
mikix, line 1080: Why two weeks?  Don't course runs have a `weeks_to_complete` field?

mikix (commented):
Why are you moving this from prospectus to course-discovery? Doesn't "deciding which runs you want to market" seem like the marketing site's domain?

This does feel like logic that we already have sitting around the place. Is "earliest active published course run" appropriate?  We have that sitting in `Course.update_marketing_redirects` and could be pulled out.

Or `CourseRun.is_current_and_still_upgradeable` which also does the two week check.

mikix:
(I realize it seems inconsistent to say "this should stay in marketing" and also "we already have logic for that here" -- but if possible, it would be nice to keep course-discovery from gaining even yet more business logic that maybe doesn't need to live in it. And this stuff seems like something that doesn't need to be shared between our frontends? And so doesn't need to live in a central repo? But maybe I'm wrong about your intended use cases.)

course_discovery/apps/api/serializers.py
mikix: Oh I guess I'm wrong about that. Verified might be only seats that have upgrade_deadlines.  See `CourseRun.get_seat_upgrade_deadline` though that hsa some of this logic.

course_discovery/apps/api/serializers.py
rgraber: That was a product decision

rgraber:
@mikix the reason we wanted to put this in discovery was so we were doing as little data transformation in prospectus as possible, and so we didn't have to recreate some of the logic that was already in disco and risk it getting out of sync. we made the same choice with the is_marketable and is_enrollable flags since they were originally coming from disco anyway but weren't defined the way we needed them to be.

course_discovery/apps/api/serializers.py
mikix: In talking with Becca about this, it sounds like the two weeks is more about "has time to see if this run is something they want to upgrade, which gives them past the end date to finish/get a cert" -- so maybe make that clearer here.

Probably a different function name than `has_time_to_finish` and maybe even a code comment?

mikix:
`risk it getting out of sync` -- but this PR didn't re-use any of the logic in Disco, so wouldn't have helped with the sync problem. It was just moving where logic lives. But hopefully this PR will start re-using existing logic and that justification for the move can start being true.

So is the argument that frontends shouldn't have any business logic at all? That the IDAs should own logic as a matter of course?

I mean, this business logic seems fairly light (just grabbing earliest marketable run, right?) - and not something that another IDA or frontend would need to reproduce? Disco already gives you start/end dates and is_marketable?

That said, the Publisher squad's current ownership policy around course-discovery does not require me to be on board.

mikix:
Sorry, `That said, the Publisher squad's current ownership policy around course-discovery does not require me to be on board.` sounded way more passive aggressive than I meant. I was just trying to say that my comments were nits rather than like, You Must Do This or I'll Hate It

course_discovery/apps/api/tests/test_serializers.py
reillz10: Good catch. I had done that for the following two tests and forgot to update the first one. Thank you!

reillz10: I suppose I don't! Thank you for catching my copy-pasta folly

cpennington:
Does this business logic need to be the same across many consumers? Is this data doing to be used/of interest to other clients?

course_discovery/apps/api/serializers.py
reillz10: 👍

course_discovery/apps/api/serializers.py
reillz10: Thanks at one point there was more logic in the enrollment_start function. Thank you for helping me become more pythonic.

jmyatt:
@cpennington yes that's a pivotal question here, and I *think* the answer is no, but open to other ideas on that

albemarle:
@cpennington @jmyatt I'm coming in a bit new here.  Which team actually owns this service these days?  Do they have architectural decision records or design docs on where business logic like this is meant to live, and if so could you point me at those?  (There looks to be a lot of good stuff in the /docs directory, but I don't see anything on this question.)

My understanding (and maybe @nasthagiri has more input here) has been that, edX-wide, front ends are expected to handle mostly display logic, and rely on the APIs to feed them data.  For instance, the ecommerce API does everything including calculating tax and calculating your discounted price, even though we could do the subtraction or multiplication in the front ends.

course_discovery/apps/api/serializers.py
reillz10: As we discuss irl: it feels weird to me to take these methods out of `get_advertised_course_run_uuid` since they are really just helper methods for that method. For now we can leave them here and move them out if they become more reusable.

course_discovery/apps/api/serializers.py
reillz10: This is more of a, we would rather a user see and be able to enroll into a future course_run than a current course_run that they won't have time to complete. We got 2 weeks as a minimum for that amount of time needed for a user to see a course_run (if there is a future run anyway) Given that, I'm not quite sure what to rename it to.

course_discovery/apps/api/serializers.py
reillz10: Fixed the NIT.

`CourseRun.get_seat_upgrade_deadline` feels like a weird function for this context. First, I would be passing a `seat_type` to it which would always be verified (something that function checks against because this function is only used currently for creating or updating a seat) From there, if there isn't any verified seats in the course_run it will return a new made up deadline, but because I'm not creating or updating a seat this deadline is bogus and isn't really tied to the seat at all. That function seems to be a helper function for `update_or_create_seat_helper`

course_discovery/apps/api/serializers.py
reillz10: went with `user_has_time_to_finish_course`

mikix: > This is more of a, we would rather a user see and be able to enroll into a future course_run than a current course_run that they won't have time to complete.

I know. I was suggesting that we *even more often* show the future run than the current run. We have a field called `weeks_to_complete` on a run. I'm guessing it's often more than 2. So hard coding 2 feels like a scammy bait and switch - get them to enroll for the current run, when we know they can't finish it. So I'd rather we use the real value and show the future run more often.

However, apparently the user can both see the course content *and* get a certificate after the course end date, if they upgrade? So the 2 weeks is more about sampling the course and deciding if you want to upgrade than anything tied to course completion.

So I still think `user_has_time_to_finish_course` is misleading about what we're trying to do -- otherwise some future engineer might reasonably think "it's a bug that they aren't using `weeks_to_complete`" and switch it.

mikix:
> Which team actually owns this service these days?

Publisher squad owns course-discovery. So, @jmyatt

> Do they have architectural decision records or design docs on where business logic like this is meant to live, and if so could you point me at those?

Not especially. I know our general architecture tries to follow the [Self-Contained System](http://scs-architecture.org/) pattern.

I don't live and breathe that philosophy, but I think it's less concerned with UI being stripped of logic than separating UI concerns into their own service (see their docs about how they are [different than microservices](http://scs-architecture.org/vs-ms.html)).

> My understanding (and maybe @nasthagiri has more input here) has been that, edX-wide, front ends are expected to handle mostly display logic, and rely on the APIs to feed them data. For instance, the ecommerce API does everything including calculating tax and calculating your discounted price, even though we could do the subtraction or multiplication in the front ends.

Sure... I agree with that ecommerce boundary decision. I hope I haven't come off as saying that all business logic should live in the UI?

mikix:
> My understanding ... has been that, edX-wide, front ends are expected to handle mostly display logic, and rely on the APIs to feed them data.

Maybe that's just my disconnect. This feels like "display logic" -- i.e. "what course run am I going to prominently display / promote to the user", and I would expect the UI to own that decision. It doesn't feel like "data" to me.

course_discovery/apps/api/serializers.py
rgraber: I can't think of a good function name, but if we can give it a decent comment, something like
"Users have time to finish course if purchasing verified cert"? Not great either, but something to be like 'yes, this is on purpose.'

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -289,6 +308,155 @@ def test_data(self):
         serializer = self.serializer_class(self.course, context={'request': self.request})
         self.assertDictEqual(serializer.data, expected)
 
+    def test_advertised_course_run_is_upgradeable_and_end_not_within_two_weeks(self):
```
rgraber, line 311: can this be done with ddt? Not a blocker but might be more in keeping with how other tests are done in disco

rgraber (approved):
Some nits, but nothing that should block merge

course_discovery/apps/api/tests/test_serializers.py
reillz10: @rgraber  I'm not familiar enough with ddt to know how to use it in this scenario. I've only used it in the past for unit test where different types of user are part of the test. Do you know how one would use ddt to clean up this test?

course_discovery/apps/api/serializers.py
mikix: >  "Users have time to finish course if purchasing verified cert"?

Again, the 2 weeks has *nothing* to do with finishing the course. Most courses have a value higher than 2 for `weeks_to_complete` -- looks like most have values in the 4-6 range.

It's more like a "long enough to sample the course to decide if you want to buy" -- a trial period maybe would be a good phrase for it?  `has_time_for_trial_period` or something? With a comment to that effect.

PR #2233, opened 2019-10-17
Change Seat.type from string to FK

mikix:
Solve long-standing TODO to make it point at a proper SeatType object.

Additionally, make SeatType.slug the unique field for SeatType, not name.

https://openedx.atlassian.net/browse/DISCO-1381


course_discovery/apps/course_metadata/models.py
```diff
@@ -1708,43 +1708,36 @@ def is_marketable(self):
...
         'decimal_places': 2,
         'max_digits': 10,
         'null': False,
         'default': 0.00,
     }
     course_run = models.ForeignKey(CourseRun, models.CASCADE, related_name='seats')
-    # TODO Replace with FK to SeatType model
-    type = models.CharField(max_length=63, choices=SEAT_TYPE_CHOICES)
+    # The 'type' field used to be a CharField but when we converted to a ForeignKey, we kept the db column the same,
+    # by specifying a bunch of these extra kwargs.
```
mikix, line 1745: The one thing that will change is that the column width will change from 63 to 50. 50 comes from the (default) width of the slug field we point to. I couldn't override it with `max_length=63`. But the width change does not destroy any contents (assuming we don't have data longer than 50). So it is a change... but I couldn't avoid it.

mikix:
Oh, another detail of this PR: I left the publisher tables alone. publisher.models.Seat still uses a string `type` field. But I figured it wasn't worth migrating that too.

course_discovery/apps/api/serializers.py
```diff
@@ -533,9 +534,7 @@ class Meta(object):
 
 class SeatSerializer(serializers.ModelSerializer):
     """Serializer for the ``Seat`` model."""
-    type = serializers.ChoiceField(
-        choices=[name for name, __ in Seat.SEAT_TYPE_CHOICES]
-    )
+    type = serializers.SlugRelatedField(slug_field='slug', queryset=SeatType.objects.all().order_by('name'))
```
Dillon-Dumesnil, line 539: Is the `order_by` necessary? Assuming the reason for this is so that it returns in the same order we used to?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -54,7 +54,7 @@ class CourseRunViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.CourseRunWithProgramsSerializer
     metadata_class = MetadataWithRelatedChoices
     metadata_related_choices_whitelist = (
-        'content_language', 'level_type', 'transcript_languages', 'expected_program_type'
+        'content_language', 'level_type', 'transcript_languages', 'expected_program_type', 'type'
```
Dillon-Dumesnil, line 57: Why this change?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -219,7 +219,7 @@
                 "attribute_values": [
                     {
                         "name": "certificate_type",
-                        "value": "honor"
+                        "value": "audit"
```
Dillon-Dumesnil, line 222: ok

course_discovery/apps/course_metadata/models.py
```diff
@@ -1720,43 +1720,36 @@ def is_marketable(self):
...
+    type = models.ForeignKey(SeatType, models.CASCADE,
+                             to_field='slug',  # this keeps the columns as a string, not an int
+                             db_column='type')  # this avoids renaming the column to type_id
```
Dillon-Dumesnil, line 1748: :chef_kiss:

course_discovery/apps/course_metadata/utils.py
```diff
@@ -383,7 +383,8 @@ def calculated_seat_upgrade_deadline(seat):
     Only verified seats have upgrade deadlines. If the instance does not have an upgrade deadline set, the value
     will be calculated based on the related course run's end date.
     """
-    if seat.type == seat.VERIFIED:
+    slug = seat.type if isinstance(seat.type, str) else seat.type.slug
```
Dillon-Dumesnil, line 386: Why is there a need for the if else? Not critiquing, just curious.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -97,9 +97,33 @@ def _create(cls, model_class, *args, **kwargs):
...
+        return SeatType.objects.get_or_create(name=Seat.MASTERS.capitalize())[0]  # name will create slug
+
+    @staticmethod
+    def professional():
+        return SeatType.objects.get(slug=Seat.PROFESSIONAL)
+
+    @staticmethod
+    def verified():
+        return SeatType.objects.get(slug=Seat.VERIFIED)
+
```
Dillon-Dumesnil, line 126: Clever

course_discovery/apps/course_metadata/models.py
```diff
@@ -1255,7 +1255,7 @@ def enrollable_seats(self, types=None):
         Returns seats, of the given type(s), that can be enrolled in/purchased.
 
         Arguments:
-            types (list of seat type names): Type of seats to limit the returned value to.
+            types (list of SeatTypes): Type of seats to limit the returned value to.
```
Dillon-Dumesnil, line 1254: Are there not concerns here about where/how this is called and changing the argument type? I didn't really see where it was called, but changing the argument type here stood out to me.

Dillon-Dumesnil (approved):
Couple questions, but everything looks good. I'm lovin it

course_discovery/apps/api/serializers.py
mikix: Yeah, I think it's making it easier for the frontend, so it doesn't have to order it. I copied it from the other place we use a slug related field for seat types.

course_discovery/apps/api/v1/views/course_runs.py
mikix: This was so that OPTIONS showed the possible values for seat.type

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
mikix: A test was failing because an honor SeatType didn't already exist. And I could have fixed that by pre-creating the type. But I figured, the test wasn't specifically caring about honor, it just old and pre-audit. So I updated the data.

course_discovery/apps/course_metadata/utils.py
mikix: This utility method is used by old publisher (which I left as a string type) and new publisher (now an object type)

course_discovery/apps/course_metadata/models.py
mikix: I believe I also changed all callers of it. As long as tests pass, I think we're OK here.

course_discovery/apps/api/v1/views/course_runs.py
Dillon-Dumesnil: Does this require writing the choices property on the SeatSerializer? I guess I would have thought this would show up more like `seat__type` or something

mikix: Naw, because of the same problem you hit - it will match the field anywhere in hierarchy (which like, is not a great feature - I should make that metadata class smarter).

mikix: Oh and you don't have to write the choices on the serializer because it's a related field and the choices are just the db from the queryset you give the slugrelatedfield.

PR #2235, opened 2019-10-18
Pull title from Studio in RCM

mikix:
We want to re-support Studio edits of course run titles. So lets actually pull that data in again.

https://openedx.atlassian.net/browse/DISCO-1390


jlajoie (approved):
👍

xnick421 (approved):

PR #2237, opened 2019-10-18
Avoid saving seats that don't match

mikix:
If a course has a CourseType or a course run has a CourseRunType, and the ecommerce RCM loader tries to save an incompatible seat or entitlement, refuse with a warning.

https://openedx.atlassian.net/browse/DISCO-1384

It does not fail the job... but other similar warnings didn't either. So I didn't change that behavior. But feels like maybe it should?

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -556,6 +556,15 @@ class EcommerceApiDataLoader(AbstractDataLoader):
                 seat_type=certificate_type, sku=sku
             )
             logger.warning(msg)
+            self.processing_failure_occurred = True
+            return
+        if course_run.type and not course_run.type.tracks.filter(seat_type=seat_type).exists():
+            logger.warning(
+                'Seat type {seat_type} is not compatible with course run type {run_type} for course run {key}'.format(
+                    seat_type=seat_type.slug, run_type=course_run.type.slug, key=course_run.key,
+                )
+            )
+            self.processing_failure_occurred = True
             return
 
         credit_provider = attributes.get('credit_provider')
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -557,6 +557,13 @@ def update_seat(self, course_run, product_body):
             )
             logger.warning(msg)
             return
+        if course_run.type and not course_run.type.tracks.filter(seat_type=seat_type).exists():
+            logger.warning(
+                'Seat type {seat_type} is not compatible with course run type {run_type} for course run {key}'.format(
+                    seat_type=seat_type.slug, run_type=course_run.type.slug, key=course_run.key,
+                )
+            )
+            return
```
jlajoie, line 568: (I know this isn't your doing and it's a tiny nit anyways) but I feel it is strange that we return None from the update_entitlement but  have no return from update_seat

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -689,6 +696,13 @@ def update_entitlement(self, body):
             )
             logger.warning(msg)
             return None
+        if course.type and mode not in course.type.entitlement_types.all():
+            logger.warning(
```
jlajoie, line 703: Question: should we be doing anything more than using the logger.warning? We'll surface it to splunk but personally I don't check that too often, will we revisit this?

jlajoie (approved):
Single question, looks good to merge regardless 👍

course_discovery/apps/course_metadata/data_loaders/api.py
mikix: Yeah agreed. Well, this code is for RCM, so you'd see it in the job. And we could reasonably fail for this case...

But the related code right above these blocks that checks if the type is even a real type don't raise an exception... So I figured I'd keep that going. Although maybe they should all raise?

PR #2240, opened 2019-10-22
Clean incoming HTML before validating

mikix:
Since users paste from word processors that inject all sorts of styling without their knowledge, we can't be strict about accepting the HTML.

So instead, let's also strip the incoming text for them.


course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -82,27 +79,6 @@ def clean_strings(cls, data):
...
-        html_converter = html2text.HTML2Text()
-        html_converter.wrap_links = False
-        html_converter.body_width = None
-        cleaned = html_converter.handle(cleaned).strip()
-        cleaned = markdown.markdown(cleaned)
-        cleaned = cls.MARKDOWN_CLEANUP_REGEX.sub(r'\1', cleaned)
-
-        # html2text does not handle ampersands properly.
-        # See https://github.com/Alir3z4/html2text/issues/109.
-        cleaned = cleaned.replace('&amp;', '&')
```
mikix, line 102: This got dropped - it got fixed upstream, and we want to keep entities as-is.

course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -82,27 +79,6 @@ def clean_strings(cls, data):
...
-        This method converts the HTML to a Markdown string (to remove styles, classes, and other unsupported
-        attributes), and converts the Markdown back to HTML.
-        """
-        cleaned = content.replace('&nbsp;', '')
-        html_converter = html2text.HTML2Text()
-        html_converter.wrap_links = False
-        html_converter.body_width = None
-        cleaned = html_converter.handle(cleaned).strip()
-        cleaned = markdown.markdown(cleaned)
-        cleaned = cls.MARKDOWN_CLEANUP_REGEX.sub(r'\1', cleaned)
```
mikix, line 98: This also got dropped - I don't think we care / want to strip surrounding `p` elements.

course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -145,8 +145,8 @@ def process_node(self, data):
         defaults = {
             'uuid': data['uuid'],
             'name': data['title'],
-            'description': self.clean_html(data['body']['value']),
-            'subtitle': self.clean_html(data['field_subject_subtitle']['value']),
+            'description': clean_html(data['body']['value']),
+            'subtitle': clean_html(data['field_subject_subtitle']['value']),
```
mikix, line 149: The fields in this field that still manually use `clean_html` would ideally be converted to be `HtmlField`s in their models. But I couldn't be bothered in this PR.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -635,3 +635,59 @@ def test_unlinked_draft_row(self):
...
...roke joinstyle="miter"/> <v:formulas>  <v:f eqn="if lineDrawn pixelLineWidth 0"/>  <v:f eqn="sum @0 1 0"/>  <v:f eqn="sum 0 0 @1"/>  <v:f eqn="prod @2 1 2"/>  <v:f eqn="prod @3 21600 pixelWidth"/>  <v:f eqn="prod @3 21600 pixelHeight"/>  <v:f eqn="sum @0 0 1"/>  <v:f eqn="prod @6 1 2"/>  <v:f eqn="prod @7 21600 pixelWidth"/>  <v:f eqn="sum @8 21600 0"/>  <v:f eqn="prod @7 21600 pixelHeight"/>  <v:f eqn="sum @10 21600 0"/> </v:formulas> <v:path o:extrusionok="f" gradientshapeok="t" o:connecttype="rect"/> <o:lock v:ext="edit" aspectratio="t"/></v:shapetype><v:shape id="Picture_x0020_1" o:spid="_x0000_i1025" type="#_x0000_t75" style=\'width:276pt;height:208pt;visibility:visible;mso-wrap-style:square\'> <v:imagedata src="file:////Users/myuser/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_image001.png"  o:title=""/></v:shape><![endif]--><!-- [if !vml]--><img src="file:////Users/myuser/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_image001.png" width="276" height="208" border="0" /><!--[endif]--></span></p><p class="MsoNormal">&nbsp;</p><p class="MsoNormal">&nbsp;</p><p><!--EndFragment--></p>',  # pylint: disable=line-too-long
```
jlajoie: 😐 _wow_ this is what we receive? Big yikes

jlajoie: So this is showing up as like, one MASSIVE line here, is that intended?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -719,3 +719,21 @@ def get_salesforce_util(partner):
         return SalesforceUtil(partner)
     except SalesforceConfiguration.DoesNotExist:
         return None
+
+
+def clean_html(content):
```
jlajoie, line 726: 👍 thanks for externalizing this, have definitely looked into using a util function before and we didn't have one so found an alternative approach.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -719,3 +719,21 @@ def get_salesforce_util(partner):
...
+    """Cleans HTML from a string.
+
+    This method converts the HTML to a Markdown string (to remove styles, classes, and other unsupported
+    attributes), and converts the Markdown back to HTML.
+    """
+    import html2text
+    import markdown
```
jlajoie: Out of curiosity why the import local to the method, just because if you import the full utils you won't import these libraries unless you use the clean_html method?

jlajoie (approved):
Looks good, just a few really minor questions but I trust that there's reasons for them 👍

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: Yeah... just because I felt like it would be more of an eyesore to have it multiline?

course_discovery/apps/course_metadata/utils.py
mikix: I was thinking that like, these are the only places we import them and was trying to localize them to where they are used. For no real reason.

Linter didn't like it anyway, so I moved them to toplevel.

course_discovery/apps/course_metadata/tests/test_utils.py
mikix: But per the big-yikes - I'm glad our validators stopped this from getting in at least  :)  But better to just strip for sure.

PR #2242, opened 2019-10-23
DISCO-1423: Catches OSError on Salesforce Requests

jlajoie:


Dillon-Dumesnil (approved):
Looks good. Don't really understand how the first OSError is caught, but the second one is raised, but I trust ya.

jlajoie:
> Looks good. Don't really understand how the first OSError is caught, but the second one is raised, but I trust ya.

Hah funnily enough that's been a problem with the code I copied this new addition from. Originally the intent was the try block happens around method() which will raise an OSError that is then caught. In the catch block we re-login, then call method() again, this time without any try/catch around it. The reason the second one raises an exception is because of the hardcoded side_effect in the mock.

Turns out the assert calls have _never_ run and codecov caught that this time due to nothing else being added.

PR #2244, opened 2019-10-24
Remove slug for non-staff users

rgraber:


Dillon-Dumesnil (approved):
I don't really know the publisher templates too well, but I'll assume you manually tested this and everything looks correct. Just looks like everything was moved under the `is_staff_user` so looks good 👍

PR #2245, opened 2019-10-24
Add in course type and run types to course options call

Dillon-Dumesnil:
DISCO-1386

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -186,6 +186,32 @@ class MetadataWithRelatedChoices(SimpleMetadata):
         return info
 
 
+class MetadataWithType(MetadataWithRelatedChoices):
+    """ A version of the MetadataWithRelatedChoices class that also includes logic for Course Types """
+
+    def create_type_options(self, info):
+        info['type_options'] = [{
+            'uuid': course_type.uuid,
+            'name': course_type.name,
+            'entitlement_types': [entitlement_type.slug for entitlement_type in course_type.entitlement_types.all()],
+            'course_run_types': [{
+                'uuid': course_run_type.uuid,
+                'name': course_run_type.name,
+            } for course_run_type in course_type.course_run_types.all()],
+        } for course_type in CourseType.objects.all()]
+        return info
+
+    def get_field_info(self, field):
+        info = super().get_field_info(field)
+
+        # This line is because a child serializer of Course (the ProgramSerializer) also has
+        # the type field, but we only want it to match with the CourseSerializer
+        if field.field_name == 'type' and isinstance(field.parent, self.view.get_serializer_class()):
+            return self.create_type_options(info)
+
+        return info
+
+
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -165,9 +165,28 @@ def determine_metadata(self, request, view):
         self.view = view  # pylint: disable=attribute-defined-outside-init
         return super().determine_metadata(request, view)
 
+    def create_type_options(self, info):
+        info['type_options'] = [{
+            'uuid': course_type.uuid,
+            'name': course_type.name,
+            'entitlement_types': [entitlement_type.slug for entitlement_type in course_type.entitlement_types.all()],
```
mikix: Just the slug, not a name for them either?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -68,6 +68,7 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
     serializer_class = serializers.CourseWithProgramsSerializer
     metadata_class = MetadataWithRelatedChoices
     metadata_related_choices_whitelist = ('mode', 'level_type', 'subjects',)
+    metadata_nested_info_whitelist = ('type',)
```
mikix: This code is so custom, I might say don't even attempt to look like you're a generic thing. Like, maybe `metadata_course_type_field = 'type'` or `metadata_show_course_type_info = True` or something dumb.

mikix (approved):

course_discovery/apps/api/v1/views/courses.py
Dillon-Dumesnil: I was debating it as well. I don't disagree. I'll update.

course_discovery/apps/api/serializers.py
Dillon-Dumesnil: Yeah, we don't actually surface this information. It is used in Frontend Applciation Publisher for
```
if (type.entitlement_types.length && type.entitlement_types[0] !== AUDIT_TRACK.key) {
```

course_discovery/apps/api/v1/views/courses.py
mikix: Or you know, use a subclass of MetadataWithRelatedChoices that added the logic. That's a reasonably clean way to isolate this code

Dillon-Dumesnil: I like that more. One sec

course_discovery/apps/api/serializers.py
```diff
@@ -165,6 +165,7 @@ def determine_metadata(self, request, view):
         self.view = view  # pylint: disable=attribute-defined-outside-init
         return super().determine_metadata(request, view)
 
+
```
mikix: Might be accidental, but who cares.

mikix (approved):

PR #2246, opened 2019-10-24
DISCO-1391 new dashboard

xnick421:
-add course run statuses and course editors to course endpoint
-allow this endpoint to be filterable by these new fields

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -12,7 +12,7 @@ from guardian.shortcuts import get_objects_for_user
 from rest_framework.exceptions import NotFound, PermissionDenied
 
 from course_discovery.apps.api.utils import cast2int
-from course_discovery.apps.course_metadata.choices import ProgramStatus
+from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
     Course, CourseEditor, CourseRun, Organization, Person, Program, Subject, Topic
 )
@@ -131,10 +131,20 @@ class FilterSetMixin:
 class CourseFilter(filters.FilterSet):
     keys = CharListFilter(field_name='key', lookup_expr='in')
     uuids = UUIDListFilter()
+    course_run_statuses = CharListFilter(method='filter_by_course_run_statuses')
+    editors = CharListFilter(field_name='editors__user__pk', lookup_expr='in', distinct=True)
 
     class Meta:
         model = Course
         fields = ('keys', 'uuids',)
+
+    def filter_by_course_run_statuses(self, queryset, _, value):
+        statuses = value
+        if 'in_review' in statuses:
+            statuses = statuses.replace(
+                'in_review', '{},{}'.format(CourseRunStatus.LegalReview, CourseRunStatus.InternalReview)
+            )
+        return queryset.filter(course_runs__status__in=statuses.split(',')).distinct()
 
 
 class CourseRunFilter(FilterSetMixin, filters.FilterSet):
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -942,6 +942,29 @@ class MinimalCourseSerializer(DynamicFieldsMixin, TimestampModelSerializer):
                   'short_description', 'type', 'url_slug',)
 
 
+class CourseEditorSerializer(serializers.ModelSerializer):
+    """Serializer for the ``CourseEditor`` model."""
+    user = GroupUserSerializer(read_only=True)
+    user_id = serializers.PrimaryKeyRelatedField(queryset=User.objects.all(), write_only=True)
+    course = serializers.SlugRelatedField(queryset=Course.objects.filter_drafts(), slug_field='uuid')
+
+    class Meta:
+        model = CourseEditor
+        fields = (
+            'id',
+            'user',
+            'user_id',
+            'course',
+        )
+
+    def create(self, validated_data):
+        course_editor = CourseEditor.objects.create(
+            user=validated_data['user_id'],
+            course=validated_data['course']
+        )
+        return course_editor
+
+
 class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     """Serializer for the ``Course`` model."""
     level_type = serializers.SlugRelatedField(required=False, allow_null=True, slug_field='name',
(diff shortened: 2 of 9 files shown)


course_discovery/apps/api/filters.py
```diff
@@ -131,11 +131,19 @@ def filter_marketable(self, queryset, name, value):
...
     keys = CharListFilter(field_name='key', lookup_expr='in')
     uuids = UUIDListFilter()
+    course_run_statuses = CharListFilter(method='filter_by_course_run_statuses')
+    editors = CharListFilter(field_name='editors__user__pk', lookup_expr='in', distinct=True)
 
     class Meta:
         model = Course
         fields = ('keys', 'uuids',)
 
+    def filter_by_course_run_statuses(self, queryset, _, value):
```
jlajoie, line 141: nit: It might be worth pylint disabling this one actually just so we know what field we aren't using, rather than the underscore notation, personally if I had to make a modification to this method I wouldn't know what param I'm missing out on.

course_discovery/apps/api/filters.py
```diff
@@ -131,11 +131,19 @@ def filter_marketable(self, queryset, name, value):
...
+        statuses = value
+        if 'in_review' in statuses:
```
jlajoie, line 143: tiny nit, maybe personal preference: could probably drop this if condition, I'd imagine the replace already does the check for existence and will just do the replace if it exists, else nothing happens anyways, though the condition DOES help increase readability, up to you

course_discovery/apps/api/serializers.py
```diff
@@ -916,6 +916,29 @@ class Meta:
                   'short_description', 'type', 'url_slug',)
 
 
+class CourseEditorSerializer(serializers.ModelSerializer):
```
jlajoie, line 945: Sanity check, this just got moved because you needed it in a lower serializer right?

course_discovery/apps/course_metadata/models.py
```diff
@@ -803,6 +803,14 @@ def first_enrollable_paid_seat_price(self):
 
         return None
 
+    @property
+    def course_run_statuses(self):
+        course_run_statuses = set()
+        course_runs = self.course_runs.all()
```
jlajoie: nit: Don't know how to do this in Django's ORM but I think I'd prefer that over this. If you care and it takes less than 3 minutes I'd say swap it, always a big fan of less data over the wire
```
SELECT DISTINCT stats FROM course_runs WHERE course_id=self
```

jlajoie (approved):
Know you haven't added tests yet but just make sure we aren't adding any n+1 queries, so be aware of the query count tests that break (if any do)

mikix:
Some comments from our slack/in-person convos, just for tracking purposes:
- Getting duplicated editors (we talked about a new API to solve this?)
- Should reset the page number if the filtering changes (like how we reset page number if search terms change)

course_discovery/apps/api/v1/urls.py
```diff
@@ -11,7 +11,7 @@
 from course_discovery.apps.api.v1.views.course_runs import CourseRunViewSet
 from course_discovery.apps.api.v1.views.courses import CourseViewSet
 from course_discovery.apps.api.v1.views.currency import CurrencyView
-from course_discovery.apps.api.v1.views.organizations import OrganizationViewSet
+from course_discovery.apps.api.v1.views import organizations as organization_views
```
mikix: We usually just import both classes we need rather than doing this.

course_discovery/apps/course_metadata/models.py
xnick421: oh you mean like have the orm do the distinct and not the set()....gotchya

course_discovery/apps/api/v1/views/organizations.py
```diff
@@ -43,10 +31,42 @@ def get_queryset(self):
...
+
+class OrganizationUsersViewSet(
+    CompressedCacheResponseMixin,
+    OrganizationQuerysetMixin,
+    mixins.ListModelMixin,
+    viewsets.GenericViewSet
+):
+    """ Users in requester's organization resource. """
+    serializer_class = GroupUserSerializer
+    pagination_class = ProxiedPagination
```
mikix: Since this is a new endpoint, you don't need to use the deprecated ProxiedPagination. Current edX recommendation is CursorPagination.

course_discovery/apps/api/v1/views/organizations.py
```diff
@@ -43,10 +31,42 @@ def get_queryset(self):
...
+
+    def get_queryset(self):
+        User = get_user_model()
+        organizations = super().get_queryset()
```
mikix: This feels weird to me - to use an Organization mixin but then completely swap out what the object for the queryset is?  Might feel more natural to just make a utility method like `get_orgs_for_user` or something in this file and use it in both places.

But it's additionally weird, because you aren't using any of the prefetch logic from the mixin.

Your filter is probably just `User.objects.filter(groups__organization_extension__group__in=user.groups).distinct()`? With a special case for staff.

(Which is really just `User.objects.filter(groups__in=user.groups).distinct()`, but making sure that we're talking about organization groups, not like "Internal Users" and "Legal Team" groups.)

course_discovery/apps/api/filters.py
xnick421: ok ill replace `_` with  `name`

course_discovery/apps/api/serializers.py
xnick421: yep

mikix (commented):
From our conversation yesterday, I would have expected this endpoint to be inside the publisher API in `course_discovery/apps/publisher/api/views.py`.

This endpoint is being surfaced at `/api/v1/organizations/users` - but we had talked about `/publisher/api/admins/organizations/users/` (our existing "get me all users in an org" query is at `/publisher/api/admins/organizations/<id>/users/` and we had talked about mirroring that).

I think there are reasonable arguments for both. Since this is kind of a weird publisher-oriented API, I'm more comfortable stuffing it into the garbage API structure of the publisher djangoapp? Rather than corrupting the more RESTful course-metadata API... But I'm persuadable. (@jlajoie - you often have opinions on APIs and paths and such)

xnick421:
@mikix does the `admins` part of that url that we discussed denote that this endpoint is only visible to admins?  i made that assumption which is why i chose this path, but i could be wrong in my assumption

course_discovery/apps/api/v1/views/organizations.py
xnick421: @mikix maybe no pagination since we want everybody?

mikix: Well, there's always a pagination, I think? (drf has some default pagination, I believe)

Else if we had a million users in an org, we'd DoS ourselves.

It's true that in practice, we hope to be lazy and not bother iterating over pages in FAP by just requesting a big enough page size to get them all in one go, though.

mikix:
> does the admins part of that url that we discussed denote that this endpoint is only visible to admins?

No, I'm not sure what "admins" is supposed to mean, except to maybe denote that this "isn't for normies" (rather for the privileged publisher users). It's just a bunch of the publisher API is back there. The permission model per-view is what controls who it's visible to. Mostly just restricted to publisher users.

mikix (approved):
Nice!

PR #2248, opened 2019-10-24
Updating prefetch related call to reduce DB queries

christopappas:
Was in new relic and noticed there was a resurgence of seat type DB calls. Updated a prefetch to bring that number back down (locally went from 200 queries to 56 on my search/all/ endpoint).

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1900,7 +1900,7 @@ class CourseSearchSerializer(HaystackSerializer):
         There are 2 course_run.all() calls in serializerMethodFields, so let's
         prefetch their objects to reduce the number of queries we have.
         """
-        prefetch_related_objects([instance.object], 'course_runs__seats')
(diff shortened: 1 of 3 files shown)


brittneyexline (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -1874,7 +1874,7 @@ def to_representation(self, instance):
         There are 2 course_run.all() calls in serializerMethodFields, so let's
         prefetch their objects to reduce the number of queries we have.
         """
-        prefetch_related_objects([instance.object], 'course_runs__seats')
+        prefetch_related_objects([instance.object], 'course_runs__seats__type')
```
mikix, line 1903: Do we want seats *and* type?

christopappas: That will actually prefetch both. The `course_runs__seats__type` entails the `course_runs__seats`

mikix (approved):
Thank you for fixing my mistake!

PR #2249, opened 2019-10-25
Allow credit/honor seats in Publisher

mikix:
For new Publisher-oriented endpoints that update seats, we want to allow but ignore existing credit or honor seats.

This is transitionary. We will eventually support them using the new CourseType support.

https://openedx.atlassian.net/browse/DISCO-1413

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -1269,6 +1269,59 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         official_audit_seat = official_course_run.seats.get(type__slug='audit')
         self.assertEqual(official_audit_seat.price, 0.00)
 
+    @ddt.data(
+        # Honor happy path (it's ignored)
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1269,6 +1269,59 @@ def test_patch_published_switch_audit_to_verified(self):
         official_audit_seat = official_course_run.seats.get(type__slug='audit')
         self.assertEqual(official_audit_seat.price, 0.00)
 
+    @ddt.data(
+        # Honor happy path (it's ignored)
+        (False, ['honor', 'verified'], ['audit', 'honor', 'verified'], 'verified', None),
```
mikix, line 1274: Uh, is this right? If we see an honor/verified situation, should we add a new audit seat?

Dillon-Dumesnil: I'd say yes. I believe our system expects verified seats to have honor with them

Dillon-Dumesnil (approved):
Looks good to me

PR #2254, opened 2019-10-29
DISCO-1376 remove has_ofac_restrictions from course

xnick421:


jlajoie (changes requested):
Won't be able to do the removal from the model and the migration in the same PR from what I remember, the whole two PR thing is what you'll need to do I think.
 https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations

jlajoie (approved):
👍

PR #2255, opened 2019-10-29
Ndalfonso/disco 1376 remove course has OFAC restrictions migrations

xnick421:


jlajoie (approved):
You'll need to rename 0210 to 0211, I got in a migration before you (I'm sorry!). Otherwise looks good

PR #2258, opened 2019-10-29
DISCO-1400: Creates unique index in Course Discovery Course Run to be unique for UUID+draft

jlajoie:
Will also migrate forward any duplicate UUIDs in case the current database is has these data integrity problems for Official rows. We don't think that many (if any) users of course-discovery will be impacted, but would rather be safe with applying this unique constraint.

Before migration
Screen Shot 2019-10-29 at 3 54 33 PM

After migration
Screen Shot 2019-10-29 at 3 55 21 PM


course_discovery/apps/course_metadata/migrations/0208_add_unique_together_to_courserun_uuid_and_draft.py
```diff
@@ -0,0 +1,53 @@
...
+                        duplicate_course_run.uuid = new_uuid
+                        duplicate_course_run.save()
+                        if duplicate_course_run.draft_version:
+                            duplicate_course_run.draft_version.uuid = new_uuid
+                            duplicate_course_run.draft_version.save()
+
+    operations = [
+        migrations.RunPython(
+            migrate_data_forward
+        ),
```
mikix: Minor nit: you might consider also adding `reverse_code=migrations.RunPython.noop` to make it possible to go backwards past this migration. Can't undo the data changes, but :shrug:

course_discovery/apps/course_metadata/migrations/0208_add_unique_together_to_courserun_uuid_and_draft.py
```diff
@@ -0,0 +1,53 @@
...
+        Data needs to be migrated forward for this migration to occur, in that we previously did not  require
+        any uniqueness on our Course Run UUIDs due to microsite support (though because of key uniqueness, if you had
+        an identical course run between two, you still were not able to save that course run). This created a new
+        UUID for a run where a key was modified to successfully save, but the UUID was not, and updates relevant drafts.
+        """
+        CourseRun = apps.get_model('course_metadata', 'CourseRun')
+
+        duplicate_uuid_values = CourseRun.everything.values('uuid').annotate(Count('uuid')).filter(
+            uuid__count__gt=1,
+            draft=False,  # Only official rows should ever need this migration
```
mikix: Why? Because we believe most of the mistakes are historical? I mean, I could go into our db right now and make duplicate uuids on drafts. But you are right I guess as long as edx.org doesn't have that state, because open edX folks are going to do this migration before running any draft-creating code. So alright, I buy it.  :)

You could just do CourseRun.objects instead? Which only looks at official rows. But then, this is a nice separate line comment.

course_discovery/apps/course_metadata/migrations/0208_add_unique_together_to_courserun_uuid_and_draft.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+        for duplicate_uuid_value in duplicate_uuid_values:
+            duplicate_uuid = duplicate_uuid_value.get('uuid')
```
mikix: Nano-nit: Might be able to avoid this line by using `values_list('uuid', flat=True)` instead of `values('uuid')` above?

mikix (approved):
Looks good to me.

PR #2259, opened 2019-10-30
Switch publisher org blacklist to a whitelist

jmyatt:


course_discovery/apps/publisher/tests/test_utils.py
```diff
@@ -252,82 +252,82 @@ def staff_user_always_false(user):
             self.assertFalse(is_on_new_pub_fe(user))
             user.is_staff = False
 
-        # When no ORGS_ON_NEW_PUB_FE list present
+        # When no ORGS_ON_OLD_PUBLISHER list present
         self.assertFalse(is_on_new_pub_fe(self.user))
```
jmyatt, line 256: an empty `ORGS_ON_OLD_PUBLISHER` will mean that users can access old publisher

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -236,7 +236,8 @@ def test_ingest(self, partner_uses_publisher, on_new_publisher):
         self.assertEqual(Course.objects.count(), 0)
         self.assertEqual(CourseRun.objects.count(), 0)
 
-        with self.settings(ORGS_ON_NEW_PUB_FE='MITx' if on_new_publisher else ''):
+        # Assume that while we are relying on ORGS_ON_OLD_PUBLISHER it will never be empty
+        with self.settings(ORGS_ON_OLD_PUBLISHER='MITx' if not on_new_publisher else 'OTHER'):
```
jmyatt, line 240: `ORGS_ON_OLD_PUBLISHER` will not be in the codebase for long - this assumes while it exists it will never be empty

course_discovery/apps/publisher/utils.py
```diff
@@ -164,54 +164,54 @@ def user_orgs(user):
...
 
 def check_orgs_from_settings(method):
-    """Decorator for retrieving orgs on new pub fe from settings"""
+    """Decorator for retrieving orgs on old publisher from settings"""
     def wrapper(obj):
-        orgs_on_new_pub_fe = frozenset(filter(None, getattr(settings, 'ORGS_ON_NEW_PUB_FE', '').split(',')))
-        if not orgs_on_new_pub_fe:
+        orgs_on_old_publisher = frozenset(filter(None, getattr(settings, 'ORGS_ON_OLD_PUBLISHER', '').split(',')))
+        if not orgs_on_old_publisher:
             return False
```
jmyatt, line 171: an empty ORGS_ON_OLD_PUBLISHER will mean that users can access old publisher

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
mikix: That seems fine - we can control that.

course_discovery/apps/publisher/utils.py
mikix: Might be worth a code comment here to that effect.

mikix (approved):
I didn't pour over your test changes, but approach seems fine to me. We'll just have to be careful when de-commissioning, but no biggie.

xnick421 (approved):
manually tested, behaves as expected! :)

PR #2263, opened 2019-11-01
Remove validation from additional info

mikix:
The additional_information Course field is filled in by administrators, not course teams. We need to allow special HTML in these fields for unusual use cases.


jlajoie (approved):
👍

PR #2268, opened 2019-11-04
DISCO-1359 manual course run keys

xnick421:
- make org extension auto_create_in_studio nullable column in db.
- add auto_generate_course_run_keys column to organization in db.
- populate new organization field with values from old org
  extension field.


course_discovery/apps/publisher/models.py
```diff
@@ -739,7 +739,7 @@ class OrganizationExtension(TimeStampedModel):
     organization = models.OneToOneField(Organization, models.CASCADE, related_name='organization_extension')
     group = models.OneToOneField(Group, models.CASCADE, related_name='organization_extension')
 
-    auto_create_in_studio = models.BooleanField(
+    auto_create_in_studio = models.NullBooleanField(
```
mikix, line 742: I forget why we need this to be null to remove it? But fine.

course_discovery/apps/course_metadata/models.py
```diff
@@ -470,6 +470,14 @@ class Organization(CachedMixin, TimeStampedModel):
         blank=True,
         help_text=_('Pick a tag from the suggestions. To make a new tag, add a comma after the tag name.'),
     )
+    auto_generate_course_run_keys = models.BooleanField(
```
mikix, line 475: I feel a _little_ squicky about encoding such edx-specific business logic in the CM models (like... why not have this flag be set per-course instead of per org? Or per-site instead of per-org?).

The publisher models were a little bit less of a commitment in my mind, but I suppose that's a fake feeling. I'm fine with them here. We can always adjust logic if open edx people need it. And we've been happy with the per-org flow for a while, so that's not likely to change.

mikix (approved):

course_discovery/apps/course_metadata/models.py
xnick421: ok cool

PR #2270, opened 2019-11-05
Disco 1359 manual course run keys 2

xnick421:


course_discovery/apps/publisher/tests/test_signals.py
```diff
@@ -120,9 +95,6 @@ def test_create_course_run_in_studio_as_rerun(self, mock_access_token):  # pylin
             course__number=number
         )
 
-        # We refresh because the signal should update the instance with the course run key from Studio
-        publisher_course_run.refresh_from_db()
```
mikix, line 124: That'd still be true, right?

course_discovery/apps/api/serializers.py
```diff
@@ -706,29 +706,21 @@ def get_marketing_url(self, obj):
...
-                # really wants.
-                if not organization.organization_extension.auto_create_in_studio:
-                    allow_key_override = True
-                    break
-            except ObjectDoesNotExist:
-                pass  # no organization extension
-
+            if not organization.auto_generate_course_run_keys:
+                allow_key_override = True
+                break
```
mikix, line 719: Nice cleanup. You _could_ go even further if you want, though maybe it's too long:
```
allow_key_override = course.authoring_organizations.filter(auto_generate_course_run_keys=False).exists()
```

course_discovery/apps/api/serializers.py
```diff
@@ -706,29 +706,21 @@ def get_marketing_url(self, obj):
...
         if allow_key_override and 'key' in data:
-            return
-
-        # Now, override whatever value for key was provided by looking at the start date
-        start = data['start']  # required
-        org, number = parse_course_key_fragment(course.key_for_reruns or course.key)
-        run = StudioAPI.calculate_course_run_key_run_value(number, start)
+            run = data['key']
```
mikix: Ah, when I wrote this initial (speculative) support, I envisioned the frontend passing the full key in. I realize we've now gone a different way and only will be passing in the term part. Might be less confusing to use `term` or `run` or something for the value from the frontend? `key` _usually_ (but not always) means a full course run key.

mikix (approved):
Nice.

You've got some noise in this PR that should shake out once you merge PR 1 - like the migration to Null. But that will work itself out.

course_discovery/apps/api/serializers.py
xnick421: aright that's fair enough!

PR #2276, opened 2019-11-07
Adds in the ability to order for course run types

Dillon-Dumesnil:
DISCO-1444


course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -111,7 +111,7 @@ def match_course_type(self, course, course_type, commit=False):
         if not course.type:
             matches[course] = course_type
 
-        course_run_types = course_type.course_run_types.all()
+        course_run_types = course_type.course_run_types.order_by('created')
```
Dillon-Dumesnil, line 114: Can you speak to if this is necessary or if it is only the course type that needs to be ordered by creation date?

mikix: Both. An Audit Only course run type vs a Verified+Audit course run type would need to be ordered correctly.

I still don't love doing this ordering by created stamp...  But maybe it will serve us enough to get everyone on CourseTypes.

course_discovery/apps/course_metadata/models.py
```diff
@@ -340,7 +340,7 @@ class CourseType(TimeStampedModel):
     name = models.CharField(max_length=64)
     slug = models.CharField(max_length=64, unique=True)
     entitlement_types = models.ManyToManyField(SeatType, blank=True)
-    course_run_types = models.ManyToManyField(CourseRunType)
+    course_run_types = SortedManyToManyField(CourseRunType)
```
mikix: nit: You *could* add a help text about the order affecting the display order? But maybe that's already obvious.

mikix (approved):

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
Dillon-Dumesnil: Yeah, it's temporary. I just wanted to make sure this change was needed. I'll leave it in. Thanks!

PR #2280, opened 2019-11-07
export new redirects and include final node redirect

rgraber:


course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
```diff
@@ -309,13 +309,19 @@ def test_ingest(self):
...
+        self.assertEqual(test_course_1.url_redirects.count(), 3)
 
         self.assertTrue(test_course_2.url_slug_history.filter(url_slug='health-numbers',
                                                               is_active=False,
                                                               is_active_on_draft=False).exists())
-        self.assertEqual(test_course_2.url_redirects.count(), 0)
-        self.assertEqual(test_course_3.url_redirects.count(), 0)
+        self.assertIn('node/354', test_course_2_paths)
+        self.assertEqual(test_course_2.url_redirects.count(), 1)
+        self.assertIn('node/563', test_course_3_paths)
```
xnick421: with three different cases in the same method, that may make this method a good candidate for ddt

albemarle (approved):

course_discovery/apps/course_metadata/data_loaders/tests/test_marketing_site.py
xnick421: also if we do use ddt here, it may be good to include a docstring at the top of the method explaining what those three test cases are

rgraber: i'm not sure it's worth a docstring since it's not actually different code paths (each node will actually be processed each time because we're testing full ingest), we're just separating which results we test

albemarle (approved):

jlajoie (approved):
👍

PR #2282, opened 2019-11-11
management command for removing redirects

rgraber:
Management command to remove redirects from courses. This is part of single course URL work.
Also a fix to standardize RCM to store all redirects without the preceding '/' so we don't end up with duplicates.
This PR also includes an update to the static redirects file.


course_discovery/apps/course_metadata/data_loaders/marketing_site.py
```diff
@@ -327,6 +327,9 @@ def process_node(self, data):
                 continue
             # redirect path is not /course/something, so add the full path to url_redirects
             try:
+                # remove preceding '/' if present for standardization
+                if redirect[0] == '/':
+                    redirect = redirect[1:]
```
mikix: A slightly more pythonic way of doing this would be `redirect = redirect.lstrip('/')` without the `if`

mikix (approved):
(I did not look closely at management code, but seems fine)

course_discovery/apps/course_metadata/management/commands/tests/test_remove_redirects_from_courses.py
```diff
@@ -0,0 +1,148 @@
...
+        # make sure active url_slugs remain
+        self.assertEqual(self.course1.active_url_slug, course1_active_url_slug)
+        self.assertEqual(self.course1.url_slug_history.count(), 1)
+        self.assertEqual(self.course1.url_redirects.count(), 0)
+
+        self.assertEqual(self.course2.active_url_slug, course2_active_url_slug)
+        self.assertEqual(self.course2.url_redirects.count(), 0)
+
+    @ddt.data(
+        # test both argument sources, with and without backslash (should be backslash-insensitive)
```
albemarle, line 58: 👍

PR #2283, opened 2019-11-12
Support Masters-style CourseTypes

mikix:
Specifically:
- Pay attention to CourseRunType.is_marketable wherever we calculate marketability
- Create CourseModes on the LMS side if a track without a seat is approved during internal review using the new Publisher flow

I've left the old Masters support in place for now. It just early exits if the run is using CourseRunTypes.

https://openedx.atlassian.net/browse/DISCO-1459

### DIFF
--- a/course_discovery/apps/course_metadata/query.py
+++ b/course_discovery/apps/course_metadata/query.py
@@ -39,6 +39,8 @@ class CourseQuerySet(models.QuerySet):
         marketable = (
             ~Q(course_runs__slug='') &
             Q(course_runs__seats__isnull=False) &
+            Q(course_runs__draft=False) &
+            ~Q(course_runs__type__is_marketable=False) &
             Q(course_runs__status=CourseRunStatus.Published)
         )
 
@@ -114,6 +116,10 @@ class CourseRunQuerySet(models.QuerySet):
         ).exclude(
             # This will exclude any course run without seats (e.g., CCX runs).
             seats__isnull=True
+        ).filter(
+            draft=False
+        ).exclude(
+            type__is_marketable=False
         ).filter(
             status=CourseRunStatus.Published
         )
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import random
 import string
 import uuid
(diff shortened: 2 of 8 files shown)


course_discovery/apps/course_metadata/utils.py
```diff
@@ -500,6 +504,56 @@ def push_to_ecommerce_for_course_run(course_run):
     return True
 
 
+def push_tracks_to_lms_for_course_run(course_run):
```
mikix, line 506: This method is mostly a copy-paste job from `signals.py`. Only difference is that it looks at tracks without a seat instead of a specific masters seat.

It also unconditionally sends a 0 USD price. This is what all the current master CourseModes in LMS have. So I think that's fine.

course_discovery/apps/course_metadata/query.py
```diff
@@ -39,6 +39,8 @@ def available(self):
         marketable = (
             ~Q(course_runs__slug='') &
             Q(course_runs__seats__isnull=False) &
+            Q(course_runs__draft=False) &
+            ~Q(course_runs__type__is_marketable=False) &
```
Dillon-Dumesnil, line 43: Why this instead of `Q(course_runs__type__is_marketable=True)`?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -500,6 +504,55 @@ def push_to_ecommerce_for_course_run(course_run):
...
+            'mode_display_name': track.mode.name,
+            'currency': 'usd',
+            'min_price': 0,
+        }
+        response = partner.lms_api_client.post(url, json=data)
+
+        if response.ok:
+            logger.info('Successfully published [%s] LMS mode for [%s].', track.mode.slug, course_run.key)
+        else:
+            logger.warning('Failed publishing [%s] LMS mode for [%s].', track.mode.slug, course_run.key)
```
Dillon-Dumesnil: Should this also log out the `response.error`?

Dillon-Dumesnil (approved):
Looks pretty good to me. Couple small questions, but loving it overall.

course_discovery/apps/course_metadata/utils.py
mikix: Yes. Will do.

course_discovery/apps/course_metadata/query.py
mikix: Because I wanted null types to still work.

PR #2285, opened 2019-11-12
Adds in org white list for showing course types

Dillon-Dumesnil:
DISCO-1461


course_discovery/apps/course_metadata/models.py
```diff
@@ -153,6 +153,68 @@ class Meta:
         abstract = True
 
 
+class Organization(CachedMixin, TimeStampedModel):
```
Dillon-Dumesnil, line 156: Moved this class higher because I was getting an Organization is not defined error when trying to set the M2M on CourseType

course_discovery/apps/api/serializers.py
```diff
@@ -2266,7 +2268,8 @@ def create_type_options(self, info):
             ],
         } for course_type in CourseType.objects.prefetch_related(
             'course_run_types__tracks__mode', 'entitlement_types'
-        ).all()]
+        ).all() if course_type.white_listed_orgs.count() == 0 or user.is_staff or
+            user.groups.filter(organization_extension__organization__in=course_type.white_listed_orgs.all())]
```
mikix: Might be able to do `.empty()` rather than `.count() == 0`

Looks like you're using a tertiary here, but no `else`?

mikix (approved):

PR #2287, opened 2019-11-15
Adding in ability to handle Honor and course runs with no seats

Dillon-Dumesnil:
This PR also adds in the Empty CourseType and Empty CourseRunType


course_discovery/apps/course_metadata/migrations/0215_add_empty_course_type.py
```diff
@@ -0,0 +1,33 @@
...
+
+EMPTY_NAME = 'Empty'
+EMPTY_SLUG = 'empty'
+
+def add_empty_course_type(apps, schema_editor):
+    CourseType = apps.get_model('course_metadata', 'CourseType')
+    CourseRunType = apps.get_model('course_metadata', 'CourseRunType')
+
+    CourseType.objects.update_or_create(name=EMPTY_NAME, slug=EMPTY_SLUG)
+    CourseRunType.objects.update_or_create(name=EMPTY_NAME, slug=EMPTY_SLUG, is_marketable=False)
```
mikix: These should probably only have `slug` spelled out in the kwargs and put the rest behind `defaults=`

course_discovery/apps/course_metadata/management/commands/backpopulate_course_type.py
```diff
@@ -112,6 +112,15 @@ def match_course_type(self, course, course_type, commit=False):
...
 
         course_run_types = course_type.course_run_types.order_by('created')
+        # We allow a subset of Course Run Types to be valid under the VA CourseType despite not
+        # being a part of the course_type.course_run_types. This is used to support legacy course runs
+        # that have Honor seats or potentially even no seats.
+        if course_type == self.va_course_type:
+            unmatched_course_run_types = CourseRunType.objects.filter(slug__in=[
+                CourseRunType.HONOR, CourseRunType.VERIFIED_HONOR, CourseRunType.VERIFIED_AUDIT_HONOR,
+                CourseRunType.EMPTY
+            ]).order_by()
```
mikix: Ah, this is all much more hard-coded than I had assumed we'd do. This is very edx.org-oriented.

In my head, we were adding new commandline arguments (like --allow-for=verified-audit:honor) so that we could be more flexible as we come up with more examples of this (or be permissive only for one course, but not others, etc).

That way we also don't need to have constants for HONOR in our code or force Open edX folks to use this same logic.

That said, this works for now and is already done.

mikix (approved):
Looks fine, but a couple comments

PR #2288, opened 2019-11-15
Send credit seats to ecom just like verified seats

mikix:
Uh, this might be all we need?

`NON_PRODUCT_MODES` is only used in one place: `push_to_ecommerce_for_course_run`, which is used by both new and old publisher. 

As far as I can tell, there should be no harm in sending it as well for old publisher.


Dillon-Dumesnil (approved):
/shrug. I guess?

PR #2290, opened 2019-11-18
Updating the Type ADR to talk about mismatches

Dillon-Dumesnil:


mikix (approved):
You might want to mention that this is usually for the case of historical weirdness, not something we encourage generally or as a best practice. But looks good!

PR #2294, opened 2019-11-19
Rename order field in LevelType (1/3)

pwnage101:
Stage 1: Add the new field and column but don't use it yet.

DE-1829

---

Subsequent stages:
2/3: https://github.com/edx/course-discovery/pull/2295
3/3: https://github.com/edx/course-discovery/pull/2296


jmyatt:
just curious, is the purpose of this change to use a more explicit field name and avoid the ambiguity (ecom implications) of the term `order`?

mikix:
I think `order` is a reserved word in a tool the Data Squad is using. So they want all models to avoid those words.

https://openedx.atlassian.net/wiki/spaces/DE/pages/1061913840/Reserved+Keywords+In+Various+DB+systems

Dillon-Dumesnil:
Does this mean we need to make these changes on all SortedM2M's we have in Disco? The newly SortedM2M of `CourseRunType` on the `CourseType` model comes to mind.

mikix:
> Does this mean we need to make these changes on all SortedM2M's we have in Disco? The newly SortedM2M of CourseRunType on the CourseType model comes to mind.

Only if the field name is reserved in Stitch. The change isn't to avoid any fields that have an order, but to avoid fields *named* `order`.

Dillon-Dumesnil:
> > Does this mean we need to make these changes on all SortedM2M's we have in Disco? The newly SortedM2M of CourseRunType on the CourseType model comes to mind.
> 
> Only if the field name is reserved in Stitch. The change isn't to avoid any fields that have an order, but to avoid fields _named_ `order`.

Ah my bad. It does a `sort_value`, not an order. Please disregard

mikix (approved):

hunytalk:
@pwnage101  Just curious why haven't we followed same practice you used for credentials, adding new field and adding references to this field in same release!

pwnage101:
@hunytalk which release are you referring to specifically?  This release (1/3) does not add any references besides the model field itself.

hunytalk:
I meant to say that why haven't you followed same release pattern here that you used for renaming field in credentials! (combining new field addition and writes to new field in first release)
Anyhow looks Ok to me.

hunytalk (approved):
image

pwnage101:
@hunytalk I see what you mean, in this case there was no need because the table is practically never written (it is essentially static).

PR #2297, opened 2019-11-20
Update our valid characters for course numbers

Dillon-Dumesnil:
The difference is the regexes are switching from using \w
to [a-zA-Z0-9].

DISCO-1479


jlajoie (approved):
Might be worth making a check that the opaque keys regex isn't used elsewhere in Disco as well, also maybe adding a test for this case? Regardless looks good by me

Dillon-Dumesnil:
> Might be worth making a check that the opaque keys regex isn't used elsewhere in Disco as well, also maybe adding a test for this case? Regardless looks good by me

Verified there are no other uses of the opaque keys regex in use

PR #2299, opened 2019-11-21
DISCO-1414 org source of truth

xnick421:
- add image fields to organization model for logo, certificate_logo,
  and banner_image
- push organization to studio on save
- remove OrganizationsApiDataLoader, SchoolMarketingSiteDataLoader,
  SponsorMarketingSiteDataLoader


course_discovery/apps/api/serializers.py
```diff
@@ -323,6 +326,21 @@ class Meta(MinimalOrganizationSerializer.Meta):
         )
         read_only_fields = ('slug',)
 
+    def get_certificate_logo_image_url(self, obj):
+        if obj.certificate_logo_image:
+            return obj.certificate_logo_image.url
+        return None
```
mikix: Feels like you might want to add a new serializer field into `api/fields.py` -- something like ImageUrlField?

course_discovery/apps/course_metadata/management/commands/download_organization_images.py
```diff
@@ -0,0 +1,42 @@
...
+class Command(BaseCommand):
+    help = 'Download organizations images from the url properties.'
+
+    def download_and_set_image(self, org, base_url, value):
+        try:
+            logo = requests.get('{}{}'.format(base_url, value))
+            try:
+                org.logo.save(org.uuid, ContentFile(logo.content))
+            except Exception as e:  # pylint: disable=broad-except
+                logger.error('{}: Could not set {} for organization {}'.format(e, value[:-4], org.uuid))
```
mikix: Why this inner except? Looks the same as outer one.

course_discovery/apps/course_metadata/management/commands/refresh_course_metadata.py
```diff
@@ -145,13 +144,8 @@ def handle(self, *args, **options):
             pipeline = (
                 (
                     (SubjectMarketingSiteDataLoader, partner.marketing_site_url_root, max_workers),
-                    (SchoolMarketingSiteDataLoader, partner.marketing_site_url_root, max_workers),
-                    (SponsorMarketingSiteDataLoader, partner.marketing_site_url_root, max_workers),
                     (CourseMarketingSiteDataLoader, partner.marketing_site_url_root, max_workers),
                 ),
-                (
-                    (OrganizationsApiDataLoader, partner.organizations_api_url, max_workers),
-                ),
```
mikix, line 132: :clap: :clap: :clap:

course_discovery/apps/course_metadata/models.py
```diff
@@ -170,6 +171,23 @@ class Organization(CachedMixin, TimeStampedModel):
         null=True, blank=True, help_text=_('Logo to be displayed on certificates. If this logo is the same as '
                                            'logo_image_url, copy and paste the same value to both fields.')
     )
+    logo_image = models.ImageField(
```
mikix, line 176: ImageField vs StdImageField? In my head, we liked StdImageField, but I don't know why or anything about the tradeoffs.

course_discovery/apps/course_metadata/management/commands/download_organization_images.py
```diff
@@ -0,0 +1,42 @@
...
+        except Exception as e:  # pylint: disable=broad-except
+            logger.error('{}: Could not get {} for organization {}'.format(e, value[:-4], org.uuid))
+
+    def handle(self, *args, **options):
+        base_url = getattr(settings, 'ORG_BASE_LOGO_URL', None)
```
mikix: Why are we using ORG_BASE_LOGO_URL here? That points at disco, yeah? And we're downloading from external services here, like Drupal?

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -336,9 +336,9 @@ class OrganizationFactory(SalesforceRecordFactory):
     name = FuzzyText()
     description = FuzzyText()
     homepage_url = FuzzyURL()
-    logo_image_url = FuzzyURL()
-    banner_image_url = FuzzyURL()
-    certificate_logo_image_url = FuzzyURL()
+    logo_image = FuzzyText()
```
mikix, line 339: Text?

course_discovery/apps/course_metadata/models.py
```diff
@@ -214,6 +232,25 @@ def marketing_url(self):
...
         return cls.objects.filter(organization_extension__group__in=user.groups.all())
 
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        key = self._cache['key']
+        super(Organization, self).save(*args, **kwargs)
+        key = key or self.key
+        partner = self.partner
+        data = {
+            'name': self.name,
+            'short_name': self.key,
```
mikix, line 251: Did you mean just `key` here?

course_discovery/apps/course_metadata/models.py
```diff
@@ -214,6 +232,27 @@ def marketing_url(self):
...
+            'description': self.description,
+        }
+        logo = self.certificate_logo_image
+        if logo:
+            base_url = getattr(settings, 'ORG_BASE_LOGO_URL', None)
+            logo_url = '{}{}'.format(base_url, logo) if base_url else logo.url
```
mikix, line 257: Does ImageField give back a relative path if it's local? When does `.url` not have the right value?

course_discovery/apps/course_metadata/management/commands/download_organization_images.py
```diff
@@ -0,0 +1,38 @@
...
+    def download_and_set_image(self, org, value, field):
+        try:
+            logo = requests.get('{}'.format(value))
+            try:
+                image_field = getattr(org, field[:-4])
+                image_field.save(value.split('/')[-1], ContentFile(logo.content))
+            except Exception as e:  # pylint: disable=broad-except
+                logger.error('{}: Could not set {} for organization {}'.format(e, field[:-4], org.uuid))
+        except Exception as e:  # pylint: disable=broad-except
+            logger.error('{}: Could not get {} for organization {}'.format(e, field[:-4], org.uuid))
```
nedbat: This would be clearer as:
```
try:
    ... get() ...
except Exception as e:
    logger.error("Couldn't get")
else:
    try:
        .... save() ...
    except Exception as e:
        logger.error("Couldn't set")
```
(I hope you don't mind a drive-by review...)

course_discovery/apps/course_metadata/models.py
xnick421: logo url will be populated correctly in stage and prod because we wont be using container names there.  However in dev, we use this because we refer to a container name.

course_discovery/apps/course_metadata/models.py
xnick421: i dont think so...key would be the cached value from before the organization record was saved

course_discovery/apps/course_metadata/models.py
xnick421: was running into a large amount of recursion errors which weren't straight forward, and changing it to the regular image field avoided them.  I can spend time looking into it if it's worth it.

course_discovery/apps/course_metadata/models.py
mikix: Ah, OK. I was missing why you were saving the cached key. Got it. Maybe a comment about the key shuffling being for renaming the key on LMS side?

PR #2301, opened 2019-11-22
DISCO-1463: Adds in override for verified_deadline, and ties deadline…

jlajoie:
… to end in loader


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -234,6 +243,15 @@ def update_course(self, official_course, draft_course, body):
         course = official_course or draft_course
         logger.info('Processed course with key [%s].', course.key)
 
+    def _update_verified_deadline_for_course_run(self, course_run):
+        if course_run:
+            seats = course_run.seats.exclude(type=Seat.VERIFIED, end=None)
```
mikix: You want to filter on verified, not exclude, right?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -234,6 +243,15 @@ def update_course(self, official_course, draft_course, body):
...
+            for seat in seats:
+                deadline = seat.course_run.end - datetime.timedelta(days=settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
+                deadline = deadline.replace(hour=23, minute=59, second=59, microsecond=99999)
```
mikix: This is kind of fussy logic. Is there room for consolidation with the other two places this logic is used?

Both `CourseRun.get_seat_upgrade_deadline` *and* `utils.calculated_seat_upgrade_deadline` use like exactly the same code (`get_seat...` can probably just call `calculated_seat...` actually). And that code is similar here too.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1780,7 +1780,8 @@ class Seat(DraftModelMixin, TimeStampedModel):
                              db_column='type')  # this avoids renaming the column to type_id
     price = models.DecimalField(**PRICE_FIELD_CONFIG)
     currency = models.ForeignKey(Currency, models.CASCADE, default='USD')
-    upgrade_deadline = models.DateTimeField(null=True, blank=True)
+    _upgrade_deadline = models.DateTimeField(null=True, blank=True, db_column='upgrade_deadline')
+    upgrade_deadline_override = models.DateTimeField(null=True, blank=True)
```
mikix, line 1789: Per Cale's efforts at making dates more relative, might this work better as an *offset* override? Like, a value to use instead of the default 10 day cutoff, so that if `end` changes, we don't have to re-calclulate the deadline like you do above.

But maybe that's not how this field gets used in practice. I dunno.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -192,8 +195,14 @@ def update_course_run(self, official_run, draft_run, body):
         new_pub_fe = is_course_on_new_pub_fe(run.course)
 
         validated_data = self.format_course_run_data(body, new_pub_fe=new_pub_fe)
+        end_has_updated = validated_data.get('end') != run.end
         self._update_instance(official_run, validated_data, suppress_publication=True)
         self._update_instance(draft_run, validated_data, suppress_publication=True)
+        if end_has_updated:
+            self._update_verified_deadline_for_course_run(official_run)
+            self._update_verified_deadline_for_course_run(draft_run)
```
mikix, line 202: So any custom deadline override gets overridden here when `end` changes? If that's true, why do we store the calculated value in the db at all?

course_discovery/apps/course_metadata/data_loaders/api.py
jlajoie: Yeah I can consolidate this, the reason I didn't want to use `calculated_seat_upgrade_deadline` is because of 
```
if seat.upgrade_deadline:
            return seat.upgrade_deadline
```

where we'd short circuit if the upgrade_deadline is already set, I suppose I can null it out though prior to the calculate call like we do elsewhere.

course_discovery/apps/course_metadata/models.py
jlajoie: So we'll need to recalculate and send the data along regardless when end changes as it needs to end up in ecom, it seemed more intuitive to me to have the same type as the field it's overriding. Also the way I viewed the use case wasn't "We want this course run deadline to occur 30 days before the end instead of 20" but rather "We'd like this to be two fridays before the course ends", though I suppose I could accept the date and calculate the offset and store that?

course_discovery/apps/course_metadata/data_loaders/api.py
mikix: Or another utility method for just the two lines that calculate the deadline? I dunno how best to refactor. Just felt like weird code to duplicate.

course_discovery/apps/course_metadata/data_loaders/api.py
jlajoie: Ahh sorry, little unclear here and maybe I can leave a comment alluding to this in the code. This ticket tackles two things, the first case being we want the default logic for verified_deadlines (the ones we compute with an offset) to be updated when some one changes the date, though agreed, we don't need to store it at all, we just need to send it along to ecom (though I'm not sure what other services might depend on having this data here in discovery).

The override never gets overridden by code, two separate fields, override only gets changed in django admin, and gets surfaced if it's set, otherwise we use the calculated value.

mikix (approved):

PR #2302, opened 2019-11-22
Support unsubmitted composite status

mikix:
For ease of frontend presentation, support filtering courses by a fake "unsubmitted" status that skips archived unpublished runs.
    
Also skip archived runs when reporting back course_run_statuses in course serializations.

https://openedx.atlassian.net/browse/DISCO-1455

Linked to frontend PR https://github.com/edx/frontend-app-publisher/pull/354

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -1,7 +1,9 @@
+import datetime
 import logging
 
+import pytz
 from django.contrib.auth import get_user_model
-from django.db.models import QuerySet
+from django.db.models import Q, QuerySet
 from django.utils.translation import ugettext as _
 from django_filters import rest_framework as filters
 from drf_haystack.filters import HaystackFacetFilter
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/filters.py
```diff
@@ -139,13 +141,24 @@ class Meta:
...
+            statuses.add(CourseRunStatus.InternalReview)
+        if 'unsubmitted' in statuses:  # unpublished and unarchived
+            statuses.remove('unsubmitted')
+            # "is not archived" logic stolen from CourseRun.has_ended
+            now = datetime.datetime.now(pytz.UTC)
+            or_queries.append(Q(course_runs__status=CourseRunStatus.Unpublished) & ~Q(course_runs__end__lt=now))
+
+        status_check = Q(course_runs__status__in=statuses)
+        for query in or_queries:
+            status_check |= query
```
jlajoie, line 159: Haven't seen that syntax before, neat!

course_discovery/apps/course_metadata/models.py
```diff
@@ -830,7 +830,14 @@ def first_enrollable_paid_seat_price(self):
...
     @property
     def course_run_statuses(self):
-        statuses = self.course_runs.filter(hidden=False).values_list('status', flat=True).distinct().order_by('status')
+        """
+        Returns all unique course run status values inside this course.
+
+        Note that it skips hidden and archived courses - this list is typically used for presentational purposes.
+        """
+        now = datetime.datetime.now(pytz.UTC)
+        runs = self.course_runs.exclude(hidden=True).exclude(status=CourseRunStatus.Unpublished, end__lt=now)
```
jlajoie, line 839: nit: Might be worth just having this all as a single exclude rather than chaining two?

jlajoie (approved):
👍 one nit that doesn't matter really, looks good

course_discovery/apps/course_metadata/models.py
mikix: I *think* if we did one exclude, the kwargs would have to all match. Like, we'd only exclude runs that were hidden, unpublished, **and** ended. But chained excludes are like `ors` in this case.

PR #2305, opened 2019-11-25
Drop flaky part of currency test

mikix:
TestCurrencyCurrencyView.test_get tests that the endpoint is cached.
Except sometimes that fails in Travis. Couldn't fail it locally.
For now, just disable that part of the test, with comment.


jlajoie (approved):
Reasonable enough for me, thanks 😸

PR #2319, opened 2019-12-06
Add partner into GET since it is part of uniqueness constraint

Dillon-Dumesnil:
DISCO-1483


jlajoie (approved):
👍 looks good, know probably not a warranted concern, but us using iexact... that can _also_ cause issues if we have COURSEKEY vs coursekey which would be permitted by our DB constraints right? But uhh, that'd never happen?

Dillon-Dumesnil:
I'm just going to leave it. I hope to deprecate all of this soon enough and it's been running with iexact for who knows how long now.

PR #2321, opened 2019-12-09
add program hooks to admin

rgraber:


course_discovery/apps/course_metadata/models.py
```diff
@@ -1870,6 +1870,8 @@ class Program(PkSearchableMixin, TimeStampedModel):
         help_text=_('The user-facing display title for this Program.'), max_length=255, unique=True)
     subtitle = models.CharField(
         help_text=_('A brief, descriptive subtitle for the Program.'), max_length=255, blank=True)
+    marketing_hook = models.CharField(
+        help_text=_('A brief hook for the marketing website'), max_length=255, blank=True)
```
mikix, line 1866: This is a temporary string, only meant for edx.org concerns, yah? I'd say it doesn't need translation then. Not a blocker, just might waste translators' time a bit.

mikix (approved):

albemarle (approved):

PR #2322, opened 2019-12-09
Avoid null Course and CourseRun types

mikix:
- Add a migration to fill in all null types to the special empty type and mark the types as blank=False.
- Require a type when serializing courses and runs.
- Fill in an empty type when making courses and runs via RCM.
- Delete old draft audit entitlements and stop making them.

https://openedx.atlassian.net/browse/DISCO-1427


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -302,52 +283,32 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
...
 
-        # First, update nested entitlements
-        entitlements = []
-        # New flow for courses using CourseType model
-        # DISCO-1399: I think pretty much the whole 'else' can be removed.
-        prices = data.get('prices', {})
-        if data.get('type'):
-            course_type = CourseType.objects.get(uuid=data.get('type'))
+        # First, update course entitlements
+        if data.get('type') or data.get('prices'):
```
Dillon-Dumesnil, line 299: Do we not need to pop `'prices'` off before? How does this play with the serializer above?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -210,6 +211,14 @@ def create_course_run(self, course, body):
         new_pub_fe = is_course_on_new_pub_fe(course)
         defaults = self.format_course_run_data(body, course=course, new_pub_fe=new_pub_fe)
 
+        # Set type to be the same as the most recent run as a best guess.
+        # Else mark the run type as empty and RCM will upgrade if it can.
+        latest_run = course.course_runs.order_by('-created').first()
```
Dillon-Dumesnil, line 216: Should this be a property on course runs? Do we use this logic a couple of times?

course_discovery/apps/course_metadata/migrations/0223_fill_in_course_types.py
```diff
@@ -0,0 +1,73 @@
...
+import django.db.models.deletion
+
+
+def null_to_empty(apps, _schema_editor):
+    """
+    Replaces any None CourseType or CourseRunType values with the special empty type.
+    """
+    for model_name in ('Course', 'CourseRun'):
+        type_model = apps.get_model('course_metadata', model_name + 'Type')
+        empty = type_model.objects.get(slug='empty')
```
Dillon-Dumesnil, line 15: Nit: `'empty'` -> `type_model.EMPTY`

Dillon-Dumesnil (approved):
Small comments. Looks great.

course_discovery/apps/course_metadata/data_loaders/api.py
mikix: I don't think we do?

course_discovery/apps/course_metadata/migrations/0223_fill_in_course_types.py
mikix: I dunno, same nit can go for the hardcoded 'audit', but part of me was like, "it's a migration file, don't reference model classes if possible"

course_discovery/apps/api/v1/views/courses.py
mikix: I dunno, but we didn't pop it before?

PR #2325, opened 2019-12-10
Updating RCM OAuth Client.

Dillon-Dumesnil:
This is being undertaken as part of a fix for receiving a
connection reset error in the ecommerce data loader.
The old client would not reauthenticate a request if the
old token expired. The new client will which should fix this
issue.


jlajoie (approved):
👍

PR #2326, opened 2019-12-12
Use Mode to determine ecom attributes

mikix:
For seats, we used to set is_id_verified and certificate_type ecom attributes by string comparision on seat type. But now we have Mode, which holds those flags.

So now we send those along now based on Mode. The only behavior difference is that we won't send any information for seats that aren't supposed to be in the that run's CourseRunType.


course_discovery/apps/publisher/api/v1/tests/test_views.py
```diff
@@ -93,341 +75,8 @@ def _mock_ecommerce_api(self, publisher_course_run, status=200, body=None):
         url = '{root}publication/'.format(root=self.partner.ecommerce_api_url)
         responses.add(responses.POST, url, json=body, status=status)
 
-    @responses.activate
-    def test_publish(self):  # pylint: disable=too-many-statements
```
mikix, line 97: This PR doesn't bother maintaining compatibility with old publisher publish flow. So I just dropped any tests broken by this change.

Dillon-Dumesnil (approved):
LGTM

PR #2327, opened 2019-12-12
Wait for the official course to be set on the official course run

Dillon-Dumesnil:
When setting official state for a course run, there is an
intermediate time after creating the official run, but before the
official course is set that would cause a duplicate external key
to be flagged. This now waits until the official course is set before
checking.

### DIFF
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -231,6 +231,13 @@ def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # py
     """
     if not instance.external_key:
         return
+    # This is for the intermediate time between the official course run being created through
+    # utils.py set_official_state and before the Course reference is updated to the official course.
+    # See course_metadata/models.py under the CourseRun model inside of the update_or_create_official_version
+    # function for when the official run is created and when several lines later, the official course
+    # is added to it.
+    if not instance.draft and instance.course.draft:
+        return
     if instance.id:
         old_course_run = CourseRun.everything.get(pk=instance.pk)
         if instance.external_key == old_course_run.external_key and instance.course == old_course_run.course:
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1565,6 +1565,9 @@ class CourseRun(DraftModelMixin, CachedMixin, TimeStampedModel):
         official_version.slug = self.slug
         official_version.course = official_course
 
+        # During this save, the pre_save hook `ensure_external_key_uniquness__course_run` in signals.py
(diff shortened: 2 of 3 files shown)


course_discovery/apps/course_metadata/signals.py
```diff
@@ -229,6 +229,13 @@ def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # py
     If the course is not associated with a program, we will still verify that the external_key
     is unique within course runs in the course
     """
+    # This is for the intermediate time between the official course run being created through
+    # utils.py set_official_state and before the Course reference is updated to the official course.
+    # See course_metadata/models.py under the CourseRun model inside of the update_or_create_official_version
+    # function for when the official run is created and when several lines later, the official course
+    # is added to it.
+    if not instance.draft and instance.course.draft:
```
schenedx, line 239: #NIT shouldn't these conditions go after the check on line 239?

schenedx (approved):
👍 Thank you!

course_discovery/apps/course_metadata/signals.py
Dillon-Dumesnil: Yeah, that's fine. It was more just wanting to make sure we treated this as an early exit. Running into some flakiness in the test so I'm going to see what I can do to improve it.

mikix (approved):
Sure.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -231,6 +231,13 @@ def ensure_external_key_uniquness__course_run(sender, instance, **kwargs):  # py
     """
     if not instance.external_key:
         return
+    # This is for the intermediate time between the official course run being created through
+    # utils.py set_official_state and before the Course reference is updated to the official course.
+    # See course_metadata/models.py under the CourseRun model inside of the update_or_create_official_version
+    # function for when the official run is created and when several lines later, the official course
+    # is added to it.
+    if not instance.draft and instance.course.draft:
```
matthugs, line 239: could we generalize
```suggestion
    if not instance.draft != instance.course.draft:
```

matthugs (approved):

course_discovery/apps/course_metadata/signals.py
Dillon-Dumesnil: Sure, one sec.

Dillon-Dumesnil: Reverting this change and going back to the old logic of `if not instance.draft and instance.course.draft:`. 

It was running into some issues (see https://travis-ci.org/edx/course-discovery/jobs/624283790?utm_medium=notification&utm_source=github_status) that I didn't feel equipped to investigate and resolve. FYI @matthugs

PR #2328, opened 2019-12-13
WEBSITE-361 set up mgmt command for bulk adding program hooks

rgraber:
Screen Shot 2019-12-13 at 4 46 19 PM


course_discovery/apps/course_metadata/models.py
```diff
@@ -2794,3 +2794,13 @@ class RemoveRedirectsConfig(SingletonModel):
     """
     remove_all = models.BooleanField(default=False, verbose_name=_('Remove All Redirects'))
     url_paths = models.TextField(default='', null=False, blank=True, verbose_name=_('Url Paths'))
+
+
+class BulkModifyProgramHookConfig(SingletonModel):
+    pass
```
mikix: Why have this at all?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2794,3 +2794,13 @@ class RemoveRedirectsConfig(SingletonModel):
...
+
+
+class ProgramHook(TimeStampedModel):
+    config = models.ForeignKey(BulkModifyProgramHookConfig, models.CASCADE, related_name='program_hooks')
```
mikix: This feels odd. A foreign key on a singleton model? What's the point?

Is this the reason for the singleton model? So you can use `config.program_hooks.all()`? Why not just directly do `ProgramHook.objects.all()`?

course_discovery/apps/course_metadata/management/commands/modify_program_hooks.py
```diff
@@ -0,0 +1,18 @@
...
+    """ Management command to bulk modify program hooks. Uses config BulkModifyProgramHooksConfig
+    to be filled with the correct mapping of uuid to program text. No command line interface for arguments
+    ./manage.py modify_program_hooks """
+
+    help = 'Modify program hooks in bulk with arguments from database'
+
+    def handle(self, *args, **options):
+        config = BulkModifyProgramHookConfig.get_solo()
+        for hook in config.program_hooks.all():
+            hook.program.marketing_hook = hook.marketing_hook
```
mikix: Where is `hook.marketing_hook` set? I feel like I don't grok the data flow here. You are trying to fill in `Program.marketing_hook` - but I miss how the data gets in the db in the first place - in the form of ProgramHooks - before running this cmd?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2794,3 +2794,13 @@ class RemoveRedirectsConfig(SingletonModel):
...
+    program = models.ForeignKey(Program, models.CASCADE)
+    marketing_hook = models.TextField(default=None, null=True, blank=False, verbose_name=_('Program hook'))
```
mikix: You said on slack that this whole thing was temporary? Don't need to bother with a translation then. That's just going to be annoying / waste time for our poor translators. And dropping it saves you from updating all the mo/po files.

course_discovery/apps/course_metadata/models.py
mikix: OK, I see from your screenshot now - this weird model setup is for convenience in django admin?

rgraber: Yes. We wanted to be able to add everything on one page. I agree it's kind of strange but I couldn't think of a way around it

course_discovery/apps/course_metadata/management/commands/modify_program_hooks.py
```diff
@@ -0,0 +1,38 @@
...
+    ./manage.py modify_program_hooks """
+
+    help = 'Modify program hooks in bulk with arguments from database'
+
+    def handle(self, *args, **options):
+        config = BulkModifyProgramHookConfig.get_solo()
+        lines = config.program_hooks.split('\n')
+        for line in lines:
+            tokenized = line.split(':', 1)
+            print(len(tokenized))
```
albemarle: leftover debugging?

albemarle (approved):

course_discovery/apps/course_metadata/management/commands/modify_program_hooks.py
rgraber: yup

course_discovery/apps/course_metadata/models.py
```diff
@@ -2810,3 +2810,7 @@ class RemoveRedirectsConfig(SingletonModel):
     """
     remove_all = models.BooleanField(default=False, verbose_name=_('Remove All Redirects'))
     url_paths = models.TextField(default='', null=False, blank=True, verbose_name=_('Url Paths'))
+
+
+class BulkModifyProgramHookConfig(SingletonModel):
+    program_hooks = models.TextField(blank=True, null=True)
```
mikix, line 2816: This is absolutely not worth changing - but Django docs recommend not using `null=True` for text fields, because then they have two empty states, and that can be confusing (like, is there a difference between the empty string and null?)

mikix (approved):

course_discovery/apps/course_metadata/models.py
rgraber: Good to know, thanks

PR #2329, opened 2019-12-13
Upgrade empty course & run types in RCM

mikix:
After loading seats from ecommerce, try filling in any empty course or course run types. This should only happen if we import runs from Studio, in practice.

If a run has seats but the course or course run does not have a type and we fail to find a matching type to use, we will fail the ecom loader (but will let it finish first).

https://openedx.atlassian.net/browse/DISCO-1427


course_discovery/apps/course_metadata/data_loaders/course_type.py
```diff
@@ -0,0 +1,138 @@
+import logging
```
mikix, line 1: Trust me that this whole file has no new logic - just moved code from the backpopulate job and then refactored to not talk about `self.`

(Except for moving from `if type:` to `if type.valid:`)

Dillon-Dumesnil (approved):
Small changes discussed in person over the property name. Otherwise, looks great. Wonderful use of iterator

PR #2330, opened 2019-12-13
DISCO-1414 org source of truth 1

xnick421:
-add certificate_name to org
-populate certificate_name in org data loader with name from
 lms


course_discovery/apps/course_metadata/migrations/0227_add_certificate_name_to_org.py
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies = [
+        ('course_metadata', '0226_historicalorganization'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='historicalorganization',
+            name='certificate_name',
+            field=models.CharField(help_text='If populated, this field will over-write name in platform.', max_length=255, null=True),
```
Dillon-Dumesnil: You're going to need to update or remove and rerun makemigrations so the help_text is updated here.

Dillon-Dumesnil (approved):
Update the migration and then looks good.

PR #2333, opened 2019-12-16
Stop unpublishing all runs in a course

mikix:
We accidentally were unpublishing all inactive runs in a course,
even if that left us with no published runs. But we are supposed
to always leave at least one published run so that a course is
still marketable.

This fixes that logic and adds new code to the
unpublish_inactive_runs management command to re-publish old runs
that we accidentally unpublished.


jlajoie (approved):
👍 looks good, thanks for fixing this, solid comment explaining what marketable_runs means and that we still need it. 

Tiniest of minor nits that's (of course) unrelated to your PR. I missed some string changes in the unpublish command around "redirect" renaming, could fix them in here or skip it, doesn't really matter just something I noticed.

mikix:
I see what you're saying about the redirect wording. If I have to re-push this PR for something, I'll update it.

PR #2343, opened 2019-12-19
Ndalfonso/disco 1414 org source of truth 2

xnick421:


course_discovery/apps/api/serializers.py
```diff
@@ -305,6 +305,27 @@ class Meta:
...
     tags = TagListSerializerField()
+    certificate_logo_image_url = serializers.SerializerMethodField()
+    logo_image_url = serializers.SerializerMethodField()
+    banner_image_url = serializers.SerializerMethodField()
+
+    def get_certificate_logo_image_url(self, obj):
+        image = getattr(obj, 'certificate_logo_image', None)
+        if image:
+            return image.url
+        return None
```
mikix, line 316: Why the `getattr`? We know the object is an `Organization`, right?

Would the following also work?
`return obj.certificate_logo_image and obj.certificate_logo_image.url`

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -28,68 +28,6 @@
 logger = logging.getLogger(__name__)
 
 
-class OrganizationsApiDataLoader(AbstractDataLoader):
```
mikix, line 31: :confetti_ball:

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -117,7 +117,7 @@ class BaseCourseIndex(OrganizationsMixin, BaseIndex):
 
     def prepare_logo_image_urls(self, obj):
         orgs = obj.authoring_organizations.all()
-        return [org.logo_image_url for org in orgs]
+        return [org.logo_image.url if org.logo_image else None for org in orgs]
```
mikix: Do you want a list that has `Nones` in it? I guess the previous version of this code had that.

Slightly tighter version of this would be `[org.logo_image and org.logo_image.url for org in orgs]`.

Or skipping None altogether: `[org.logo_image.url for org in orgs if org.logo_image]`?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -569,6 +569,7 @@ def test_get_lms_course_url(self):
         lms_course_url = get_lms_course_url_for_archived(partner, '')
         self.assertIsNone(lms_course_url)
 
+        partner.lms_url = 'http://127.0.0.1:8000'
```
mikix, line 572: Do you need all these test file `lms_url` changes, if we now catch the exception in `save`?

course_discovery/apps/course_metadata/models.py
```diff
@@ -218,6 +237,32 @@ def marketing_url(self):
...
+        """
+        We cache the key here before saving the record so that we can hit the correct
+        endpoint in lms.
+        """
+        key = self._cache['key']
+        super(Organization, self).save(*args, **kwargs)
+        key = key or self.key
+        partner = self.partner
+        data = {
+            'name': self.certificate_name if self.certificate_name else self.name,
```
mikix: `self.certificate_name or self.name`

course_discovery/apps/course_metadata/models.py
```diff
@@ -218,6 +237,32 @@ def marketing_url(self):
...
+        logo = self.certificate_logo_image
+        if logo:
+            base_url = getattr(settings, 'ORG_BASE_LOGO_URL', None)
+            logo_url = '{}{}'.format(base_url, logo) if base_url else logo.url
+            data['logo_url'] = logo_url
+        organizations_url = '{}organizations/{}/'.format(partner.organizations_api_url, key)
+        if partner.lms_api_client:
+            try:
+                partner.lms_api_client.put(organizations_url, json=data)
+            except Exception as e:  # pylint: disable=broad-except
```
mikix: I know I suggested this pattern, and I think it's still appropriate to catch the exception for several cases (at minimum, endpoint does not exist plus whatever stops the tests from failing). But there might be exceptions that are fatal?

Like, if Studio comes back and says "hey bud, uh just give me a moment" like a 503 and Studio doesn't rename its `short_name` like you asked - now you and Studio are out of sync. In that case, you might want to be fatal about it. So maybe only log and ignore for certain codes?

```
except HttpException as e:  # or whatever the exception class should be
   if e.code == 404:
       log()
   else:
       raise
```

mikix (approved):

course_discovery/apps/api/serializers.py
xnick421: i did this because we get an error saying something like `no file associated with 'certificate_logo_image'` if the org doesnt have one and we try to reference it.  so in that case, your suggestion doesnt work

PR #2345, opened 2019-12-19
Pop entitlements again, now that Publisher isn't sending it

mikix:
DISCO-1399


Dillon-Dumesnil (approved):
Best PR I've ever seen

PR #2347, opened 2019-12-23
Drop most of publisher

mikix:
This is part of removing the old publisher djangoapp in favor of the new frontend-app-publisher frontend.

I've left:
* Some of the permisison stuff around organization extensions, as we still use them
* The organization APIs that new publisher uses
* All models (minus attached logic) - `makemigrations` should be a no-op with this PR


Makefile
```diff
@@ -15,7 +15,6 @@ help: ## Display this help message
 static: ## Gather all static assets for production
 	$(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress --optimize-minimize
 	python manage.py collectstatic -v 0 --noinput
-	python manage.py compress -v0 --force
```
mikix, line 18: This would error out without templates to compress.  :man_shrugging:

jlajoie (approved):
Wasn't following the changes for the utils, but testing locally everything thing seems to work as expected with new pub still. Will merge this on Monday

PR #2349, opened 2019-12-27
Fix a few Django admin bugs

mikix:
- Fixed the canonical course run of a course from showing a value
  even if it didn't have one. Just made it a raw id, since this
  field isn't often changed and doesn't need to be pretty.

- Fixed the 'courses' field of a Program from including drafts and
  from not showing the correct ordering.

https://openedx.atlassian.net/browse/DISCO-1535

### DIFF
--- a/course_discovery/apps/course_metadata/forms.py
+++ b/course_discovery/apps/course_metadata/forms.py
@@ -8,22 +8,7 @@ from course_discovery.apps.course_metadata.choices import ProgramStatus
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, CurriculumCourseMembership, CurriculumCourseRunExclusion, Pathway, Program
 )
-
-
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/forms.py
```diff
@@ -8,55 +8,37 @@
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, CurriculumCourseMembership, CurriculumCourseRunExclusion, Pathway, Program
 )
-
-
-def filter_choices_to_render_with_order_preserved(self, selected_choices):
```
mikix, line 13: This just wasn't being called? I think maybe Django or sortedm2m or dal have evolved in a way that avoided this. Instead, the easiest way I found to modify the order was via `optgroups`, not `filter_choices_to_render`.

xnick421 (approved):

PR #2350, opened 2019-12-27
Return 403 on non-publisher user writes

mikix:
If a non-publisher user (a user without a group associated) tries to use publisher API, we now return a more forceful status code (rather than simply returning an empty set of courses or runs).

https://openedx.atlassian.net/browse/DISCO-1532


course_discovery/apps/publisher/utils.py
```diff
@@ -74,15 +74,15 @@ def is_project_coordinator_user(user):
...
+    """ Returns True if the user is allowed to use publisher functionality (member of a group or staff).
 
     Arguments:
         user (:obj:`User`): User whose permissions should be checked.
 
     Returns:
         bool: True, if user is an publisher user; otherwise, False.
     """
-    return user.groups.exists()
+    return user.is_superuser or user.is_staff or user.groups.exists()
```
mikix: Not necessary, but I figured while I was here. Hopefully this doesn't create too much conflict with my publisher-removal PR?

jlajoie (approved):
👍

PR #2351, opened 2020-01-02
Updating logs in RCM

Dillon-Dumesnil:
In an effort to reduce the total amount of logging
in RCM, we are updating the logs to remove any logs
in the analytics api and only log out information
when an error happens in the Ecommerce data loader.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -564,7 +564,6 @@ class EcommerceApiDataLoader(AbstractDataLoader):
 
     def update_seats(self, body):
         course_run_key = body['id']
-        logger.info('Processing seats for course with key [%s].', course_run_key)
         try:
             course_run = CourseRun.objects.get(key__iexact=course_run_key)
         except CourseRun.DoesNotExist:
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -633,14 +637,15 @@ def update_seat(self, course_run, product_body):
...
-        course_run.seats.update_or_create(
+        _, created = course_run.seats.update_or_create(
             type=seat_type,
             credit_provider=credit_provider,
             currency=currency,
             defaults=defaults
         )
 
-        logger.info('Processed seat for course with key [%s] and sku [%s].', course_run.key, sku)
+        if created:
```
Dillon-Dumesnil, line 644: I made the assumption that we only care about processing a seat if we have just created a new one as a way of reducing this log. This seems to be the only relevant case given we don't log out any other info about what we processed.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -605,13 +607,15 @@ def update_seat(self, course_run, product_body):
         try:
             seat_type = SeatType.objects.get(slug=certificate_type)
         except SeatType.DoesNotExist:
+            logger.info('Processing seats for course run with key [%s].', course_run.key)
```
jlajoie: Might make more sense to move this to the msg, seems a little misleading to say "processing" only for some and they're always expected to fail

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -605,13 +607,15 @@ def update_seat(self, course_run, product_body):
...
             msg = 'Could not find seat type {seat_type} while loading seat with sku {sku}'.format(
                 seat_type=certificate_type, sku=sku
             )
             logger.warning(msg)
             self.processing_failure_occurred = True
             return
         if not course_run.type.empty and not course_run.type.tracks.filter(seat_type=seat_type).exists():
+            logger.info('Processing seats for course run with key [%s].', course_run.key)
```
jlajoie: Can probably drop this one

jlajoie (approved):

PR #2355, opened 2020-01-08
DISCO-1414 org source of truth 3

xnick421:
- remove download org image command
- remove usages in org model to image url fields


jmyatt (changes requested):
~~it looks like there might be some usages of at least one of the deleted model fields that need to also be deleted in this PR (not sure this is the only one):~~

~~https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/admin.py#L173~~

nope my bad - different model

jmyatt (approved):

PR #2356, opened 2020-01-08
BLOCKED: DISCO-1540: Adds in id_verification_required to ecom push for entitlements

jlajoie:
see associated ecommerce PR https://github.com/edx/ecommerce/pull/2743

### DIFF
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -420,7 +420,7 @@ def serialize_seat_for_ecommerce_api(seat, mode):
     }
 
 
-def serialize_entitlement_for_ecommerce_api(entitlement):
+def serialize_entitlement_for_ecommerce_api(entitlement, mode):
     return {
         'price': str(entitlement.price),
         'product_class': 'Course Entitlement',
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/utils.py
```diff
@@ -443,31 +447,38 @@ def push_to_ecommerce_for_course_run(course_run):
...
     tracks = course_run.type.tracks.all()
     seats_with_modes = []
+    entitlements_with_modes = []
     for seat in course_run.seats.all():
         for track in tracks:
             if track.seat_type and seat.type == track.seat_type:
                 seats_with_modes.append((seat, track.mode))
                 break
+    for entitlement in course.entitlements.all():
+        for track in tracks:
```
Dillon-Dumesnil, line 462: Is there a chance that we lose any behavior here since we will now only send an entitlement if the course run has a track that matches it?

I'm thinking of the case where there might be a Verified Entitlement, but maybe the specific course run only has a Masters track or Audit track. Just something to think about. It might be fine.

Dillon-Dumesnil (approved):
Small question. Otherwise, looks good

PR #2357, opened 2020-01-08
remove check for up-to-date translations

karengu:
Translations will be automatically pushed to Transifex with a Jenkins job (https://github.com/edx/edx-internal/pull/1389, https://github.com/edx/edx-internal/pull/1390).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,6 @@ matrix:
 
       script:
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make docs'
-          - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make check_translations_up_to_date'
-          - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make validate_translations'
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make clean_static'
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make static'
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make quality'
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 .DEFAULT_GOAL := test
 NODE_BIN=$(CURDIR)/node_modules/.bin
 
-.PHONY: accept clean clean_static compile_translations detect_changed_source_translations dummy_translations extract_translations \
-	fake_translations help html_coverage migrate open-devstack production-requirements pull_translations quality requirements.js \
-	requirements start-devstack static stop-devstack test validate check_translations_up_to_date docs  static.dev static.watch
+.PHONY: accept clean clean_static detect_changed_source_translations extract_translations \
+	help html_coverage migrate open-devstack production-requirements pull_translations quality requirements.js \
+	requirements start-devstack static stop-devstack test docs static.dev static.watch
 
 include .travis/docker.mk
 
@@ -69,6 +69,7 @@ migrate: ## Apply database migrations
 html_coverage: ## Generate and view HTML coverage report
 	coverage html && open htmlcov/index.html
 
+# This Make target should not be removed since it is relied on by a Jenkins job (`edx-internal/tools-edx-jenkins/translation-jobs.yml`), using `ecommerce-scripts/transifex`.
 extract_translations: ## Extract strings to be translated, outputting .po and .mo files
 	# NOTE: We need PYTHONPATH defined to avoid ImportError(s) on Travis CI.
 	cd course_discovery && PYTHONPATH="..:${PYTHONPATH}" django-admin.py makemessages -l en -v1 --ignore="assets/*" --ignore="static/bower_components/*" --ignore="static/build/*" -d django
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -30,7 +30,6 @@ matrix:
 
       script:
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make docs'
-          - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make check_translations_up_to_date'
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make validate_translations'
```
Dillon-Dumesnil: I don't think we need this line either. No need to validate, we trust the push script is doing the correct thing.

Makefile
```diff
@@ -82,12 +82,10 @@ dummy_translations: ## Generate dummy translation (.po) files
 compile_translations: ## Compile translation files, outputting .mo files for each supported language
 	python manage.py compilemessages
 
-fake_translations: extract_translations dummy_translations compile_translations ## Generate and compile dummy translation files
-
 pull_translations: ## Pull translations from Transifex
 	tx pull -af --mode reviewed --minimum-perc=1
 
-push_translations: ## push source translation files (.po) from Transifex
+push_translations: ## Push source translation files (.po) to Transifex
```
Dillon-Dumesnil, line 85: 👏

Makefile
```diff
@@ -82,12 +82,10 @@ dummy_translations: ## Generate dummy translation (.po) files
 compile_translations: ## Compile translation files, outputting .mo files for each supported language
```
jlajoie: Looks like maybe we can drop this command as well since fake_translations was all that called it (unless it's relied on by the jenkins jobs)

jlajoie (approved):
Looks good, just the one comment on possibly being able to drop that command. Also don't forget to remove `fake_translations` from `.PHONY` as well since that command is gone

docs/advanced.rst
```diff
@@ -78,11 +78,7 @@ All user-facing strings should be marked for translation. edX runs this applicat
 Updating Translated Strings
 +++++++++++++++++++++++++++
 
-Like most edX projects, Discovery uses Transifex to translate content. Most of the translation process is automated. Transifex pulls new strings from this repo on a daily basis. Translated strings are merged back into the repo every week.
-
-If you change or introduce new strings, run ``make fake_translations`` to extract all strings marked for translation, generate fake translations in the Esperanto (eo) language directory, and compile the translations. You can trigger the display of the translations by setting your browser's language to Esperanto (eo), and navigating to a page on the site. Instead of plain English strings, you should see specially-accented English strings that look like this:
```
Dillon-Dumesnil, line 83: With the above comment about trying to keep things open source friendly, I think we should maybe just call out this section for the Open Source community and more clearly specify that for edX, the translations are pushed and pulled automatically.

Makefile
```diff
@@ -82,12 +82,10 @@ dummy_translations: ## Generate dummy translation (.po) files
...
 	python manage.py compilemessages
 
-fake_translations: extract_translations dummy_translations compile_translations ## Generate and compile dummy translation files
```
Dillon-Dumesnil, line 85: So I think you can basically remove any make commands that have to do with translations, except for `extract_translations` because it is used by the push job. I would add a comment to that specifying that that make command cannot be deleted because of the push job as well.

Makefile
```diff
@@ -82,12 +82,10 @@ dummy_translations: ## Generate dummy translation (.po) files
 compile_translations: ## Compile translation files, outputting .mo files for each supported language
 	python manage.py compilemessages
```
Dillon-Dumesnil: For all the commands you remove, please also remove them from the `.PHONY:` above

Makefile
Dillon-Dumesnil: But also I think we can just get rid of this now

docs/advanced.rst
```diff
@@ -78,11 +78,7 @@ All user-facing strings should be marked for translation. edX runs this applicat
...
-
-    Thé Fütüré øf Ønlïné Édüçätïøn Ⱡσяєм ι# Før änýøné, änýwhéré, änýtïmé Ⱡσяєм #
+Like most edX projects, Discovery uses Transifex to translate content. Most of the translation process is automated. Every week, changes to source code strings are extracted as translations, which are merged back to the repo and pushed to Transifex. Translated strings are also merged back into the repo every week.
```
Dillon-Dumesnil: Can change the part `Most of the translation process is automated.` to `The translation process is automated.`

Dillon-Dumesnil (commented):
Great start. Couple more things to remove (Yay!) and this should be looking good.

Also, please search the repo for anywhere we talk about translations and see if the language needs to be cleaned up. I know there is at least one in the README.rst and probably some other spots.

Makefile
karengu: I think `push.py` assumes that the repo defines `push_translations` (`ecommerce-scripts/transifex/utils.py`).

Dillon-Dumesnil: Ah totally correct. Cross reference all the `make` commands called in that file with the Makefile here to make sure we leave them all intact. Good catch!

docs/advanced.rst
```diff
@@ -73,16 +73,14 @@ Internationalization
 
 All user-facing strings should be marked for translation. edX runs this application in English, but our open source users may choose to use another language. Marking strings for translation ensures our users have this choice. Refer to edX's `i18n guidelines`_ for more details.
 
-.. _i18n guidelines: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/index.html
+.. _i18n guidelines: https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/internationalization/i18n.html
```
Dillon-Dumesnil, line 76: Great catch there

Makefile
```diff
@@ -71,23 +71,18 @@ html_coverage: ## Generate and view HTML coverage report
 
 extract_translations: ## Extract strings to be translated, outputting .po and .mo files
 	# NOTE: We need PYTHONPATH defined to avoid ImportError(s) on Travis CI.
+	# This Make target should not be removed since it is relied on by a Jenkins job (`edx-internal/tools-edx-jenkins/translation-jobs.yml`), using `ecommerce-scripts/transifex`.
```
Dillon-Dumesnil: Nit: Put the comments about the Make targets above the command name.

FYI nit just means it's a small personal preference of the reviewer, but you're welcome to ignore it

Dillon-Dumesnil (approved):
Nearly there! There is a use of `make validate_translations` in `course-discovery/.travis/run_tests.sh` that should be removed. Once that's out, you're good to merge!

And great job on the documentation!

PR #2358, opened 2020-01-08
DISCO-1414 org source of truth 4

xnick421:
-add migration for removal of url fields on organization
-also add migration from missed migration of previous change which
 makes blank=True for an org's certificate_name


Dillon-Dumesnil (approved):

jlajoie (approved):
Just double checking that you have the model changes in a previous PR

PR #2359, opened 2020-01-08
add brackets

karengu:
Add dummy change.


Dillon-Dumesnil (approved):

Dillon-Dumesnil:
Quality test will fail on purpose for translations not being pushed up. This is intentional so we are able to test the new push_translations job.

PR #2361, opened 2020-01-08
add fake translation step on extract translations

karengu:


Makefile
```diff
@@ -69,15 +69,17 @@ migrate: ## Apply database migrations
 html_coverage: ## Generate and view HTML coverage report
 	coverage html && open htmlcov/index.html
 
-extract_translations: ## Extract strings to be translated, outputting .mo files
+extract_translations: ## Extract strings to be translated, outputting .po and .mo files
```
Dillon-Dumesnil, line 72: Just commenting that we are setting `extract_translations` to mirror what `fake_translations` does so it generates the same .mo and .po files. We are unclear if all of the /locale/eo/ files are necessary, but are leaving that for a future time.

Dillon-Dumesnil (approved):

PR #2369, opened 2020-01-13
WS-543 add redirects from published course run slugs to the appropriate course

rgraber:


course_discovery/apps/course_metadata/management/commands/backfill_course_run_slugs_to_courses.py
```diff
@@ -0,0 +1,62 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('--all', action='store_true', help='Add redirects from all course run slugs')
+        parser.add_argument('-uuids', nargs="*", help='Add redirects from all course run slugs for specific courses')
+        parser.add_argument('--args-from-database', action='store_true',
+                            help=('Use arguments from the BackfillCourseRunSlugsConfig model instead of the '
+                                  'command line.')
+                            )
+
+    def handle(self, *args, **options):
+        # using mutually exclusive argument groups in management commands is only supported in Django 2.2
```
jlajoie, line 26: nit: I think I've seen this logic elsewhere in disco, would rather see this as a util somewhere instead of copy + pasting https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/remove_redirects_from_courses.py#L28-L40

course_discovery/apps/course_metadata/management/commands/backfill_course_run_slugs_to_courses.py
```diff
@@ -0,0 +1,62 @@
...
+        options_dict = options
+        if options_dict['args_from_database']:
+            options_dict = self.get_args_from_database()
+        self.add_redirects_from_course_runs(**options_dict)
+
+    def add_course_run_redirect_to_parent_course(self, course_run):
+        if course_run.status != CourseRunStatus.Published or course_run.draft:
+            return
+        try:
+            course_run.course.url_slug_history.update_or_create(
```
jlajoie: We shouldn't be using update_or_create if we're hitting uniqueness constraints, same way that you won't use .get() on a non-unique field. Why not update_or_create on what actually uniquely identifies this?

course_discovery/apps/course_metadata/management/commands/backfill_course_run_slugs_to_courses.py
```diff
@@ -0,0 +1,62 @@
...
+                url_slug=course_run.slug, course=course_run.course, partner=course_run.course.partner
+            )
+        except IntegrityError:
+            existing_slug = CourseUrlSlug.objects.get(url_slug=course_run.slug, partner=course_run.course.partner)
```
jlajoie: Alternatively it might just be better to do a pre-check. At the very least with this approach, I'd want to know what the IntegrityError is (uniqueness/fk constraint) failure so we aren't just swallowing errors that exist and just assume it's from a slug belonging to a course already. But again, if this is our normal and expected path for some of these courses to take, I'd also rather not see stacktraces for non actionable errors.

course_discovery/apps/course_metadata/management/commands/backfill_course_run_slugs_to_courses.py
```diff
@@ -0,0 +1,62 @@
...
+            logger.warning('Cannot add slug {slug} to course {uuid0}. Slug already belongs to course {uuid1}'.format(
```
jlajoie: Should this be info, warning, or an error? Is this actually problem that should fail the build? What's the goal by logging this? If I ran this command and it passed, I probably wouldn't hop in and look at the log messaging.

course_discovery/apps/course_metadata/management/commands/backfill_course_run_slugs_to_courses.py
```diff
@@ -0,0 +1,62 @@
...
+                slug=course_run.slug,
+                uuid0=course_run.course.uuid,
+                uuid1=existing_slug.course.uuid
+            ))
+
+    def add_redirects_from_course_runs(self, **kwargs):
+        if kwargs['uuids']:
```
jlajoie, line 55: nit: might want to swap this to a .get('uuids') since you'll get a KeyError if uuids is undefined (might never be but seemed better to be safe)

course_discovery/apps/course_metadata/models.py
```diff
@@ -1712,6 +1712,9 @@ def publish(self, send_emails=True):
         # It is likely that we are sunsetting an old run in favor of this new run, so unpublish old runs just in case
         self.course.unpublish_inactive_runs()
 
+        # Add a redirect from the course run URL to the canonical course URL if one doesn't already exist
+        self.course.url_slug_history.update_or_create(url_slug=self.slug,  # pylint: disable=no-member
```
jlajoie: You had the IntegrityError check up above, should that be here as well?

jlajoie (approved):
Looks fine, couple of nits/questions. Please fix your commit message before merging this in.

Also I've seen a lot of management commands around slugs/redirects being merged in, nothing wrong with that though it does come off as a bit of a code smell. I'm missing context/history but would love to know the reasons this data needs so much massaging.

course_discovery/apps/course_metadata/management/commands/backfill_course_run_slugs_to_courses.py
rgraber: The goal here is for us to know how many collisions we have. It's something we need to figure out before we make all redirects permanent.

PR #2372, opened 2020-01-16
EDUCATOR-4890 Empty strings should not be duplicated values

schenedx:
See https://openedx.atlassian.net/browse/EDUCATOR-4890
We should not have empty strings as values within the duplicated external_key check.

@edx/masters-devs Please help review.

### DIFF
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -218,6 +218,8 @@ def update_or_create_salesforce_course_run(instance, created, **kwargs):  # pyli
 def _build_external_key_sets(course_runs):
     """
     Helper function to extract two sets of ids from a list of course runs for use in filtering
+    However, the external_keys with null or empty string values are not included in the
+    returned external_key_set.
 
     Parameters:
         - course runs: a collection of course runs
@@ -228,7 +230,8 @@ def _build_external_key_sets(course_runs):
     external_key_set = set()
     course_run_ids = set()
     for course_run in course_runs:
-        external_key_set.add(course_run.external_key)
+        if course_run.external_key:
+            external_key_set.add(course_run.external_key)
         if course_run.id:
             course_run_ids.add(course_run.id)
 
--- a/course_discovery/apps/course_metadata/tests/test_signals.py
+++ b/course_discovery/apps/course_metadata/tests/test_signals.py
@@ -358,6 +358,39 @@ class ExternalCourseKeySingleCollisionTests(ExternalCourseKeyTestDataMixin, Test
         curriculum_4.refresh_from_db()
         self.assertEqual(curriculum_4.program, new_program)
 
+    @ddt.data(
+        None,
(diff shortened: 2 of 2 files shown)


course_discovery/apps/course_metadata/signals.py
```diff
@@ -228,9 +228,10 @@ def _build_external_key_sets(course_runs):
     external_key_set = set()
     course_run_ids = set()
     for course_run in course_runs:
-        external_key_set.add(course_run.external_key)
-        if course_run.id:
-            course_run_ids.add(course_run.id)
+        if course_run.external_key:
+            external_key_set.add(course_run.external_key)
+            if course_run.id:
+                course_run_ids.add(course_run.id)
```
iloveagent57: Should these two lines be de-dented?

mikix (commented):
I have the same question as Alex about de-denting.

This is a different code path than `Nones` were going through before, yah? Like, before this change, `None` was getting added to the set, right? Where did that eventually get ignored? And why not just ignore the empty string in the same place?

(In general, Django recommends leaving `null=False` (default value) for string-based model fields, since it's confusing to have two blank states - null and empty.)

course_discovery/apps/course_metadata/signals.py
```diff
@@ -228,9 +228,10 @@ def _build_external_key_sets(course_runs):
     external_key_set = set()
     course_run_ids = set()
     for course_run in course_runs:
-        external_key_set.add(course_run.external_key)
-        if course_run.id:
-            course_run_ids.add(course_run.id)
+        if course_run.external_key:
```
kdmccormick, line 233: Please note in the docstring that all falsy values are left out of the returned sets.

kdmccormick (commented):
@mikix The reason this wasn't an issue with `None` is because in SQL, `NULL` doesn't equal `NULL`, so [`Q(external_key__in=external_key_set)`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/signals.py#L275) left out all course runs with `external_key=None`

That being said, I do think changing the empty strings in the DB to `None` and then adding the constraint `null=False` is the way to go. If you are trying to get the bug fixed merged ASAP, I'm cool with that migration being a separate PR.

course_discovery/apps/course_metadata/signals.py
schenedx: Done.

schenedx:
@kdmccormick Thank you for helping me answer @mikix's question.
I think we should use this change to quickly unblock enrollments for Masters today.
We can talk about the database migration and fixes in the coming days.

course_discovery/apps/course_metadata/signals.py
schenedx: Updated the docstring for this point.

kdmccormick (approved):
:+1:

mikix (approved):

PR #2375, opened 2020-01-17
DISCO-1551 allow spans with lang attribute to be saved to backend 

karengu:
`html2text` automatically strips `spans`. Subclass the `HTML2Text` class to allow only `spans` with the `lang` attribute to be saved.


course_discovery/apps/course_metadata/utils.py
```diff
@@ -634,6 +634,23 @@ def get_salesforce_util(partner):
         return None
 
 
+class HTML2TextWithLangSpans(html2text.HTML2Text):
+    def __init__(self, out=None, baseurl="", bodywidth=html2text.config.BODY_WIDTH):
+        super().__init__(out, baseurl, bodywidth)
```
mikix: Might be easier / more future-proof to do:
```
    super().__init__(*args, **kwargs)
```

course_discovery/apps/course_metadata/utils.py
```diff
@@ -634,6 +634,23 @@ def get_salesforce_util(partner):
...
+    def handle_tag(self, tag, attrs, start):
+        super().handle_tag(tag, attrs, start)
+        if tag == 'span' and attrs:
+            attr_dict = dict(attrs)
+            if start and 'lang' in attr_dict:
+                self.outtextf(u'<span lang="{}">'.format(attr_dict['lang']))
+                self.in_lang_span = True
+            elif self.in_lang_span:
+                self.outtextf('</span>')
+                self.in_lang_span = False
```
mikix, line 655: Is there any risk of missing this line? Like should this be under a more general `if` or like should the outer `if` have an `else` that resets this var? Or would that imply misformed html that would error out in some way?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -634,6 +634,23 @@ def get_salesforce_util(partner):
...
+    def __init__(self, out=None, baseurl="", bodywidth=html2text.config.BODY_WIDTH):
+        super().__init__(out, baseurl, bodywidth)
+        self.in_lang_span = False
+
+    def handle_tag(self, tag, attrs, start):
+        super().handle_tag(tag, attrs, start)
+        if tag == 'span' and attrs:
+            attr_dict = dict(attrs)
+            if start and 'lang' in attr_dict:
+                self.outtextf(u'<span lang="{}">'.format(attr_dict['lang']))
```
mikix: Any non-lang attributes are already addressed by the call to super(), yah?

So this results in inline `<span>` tags in the markdown? Which markdown allows because it's kinda crazy. Neat. And if there's block-level content inside the span tags, that would have been the rich text editor's fault I guess...

mikix (approved):
In terms of approach... I don't see any option toggles on html2text that lets you do similar. So either we replace html2text with something else or something like this has to happen. :man_shrugging: This seems like the lesser evil right now.

jlajoie (approved):
👍Mostly left the comments in slack, but overall this implementation seems sane until we rethink our HTML stripping approach.

PR #2381, opened 2020-01-21
Drop some unused publisher tables

mikix:
https://openedx.atlassian.net/browse/DISCO-1472

There are a few changes in actual code (which you might think would affect blue/green deploys), but as far as I can tell, it's all in either unused or test code. So should be fine to all do in one PR.


course_discovery/apps/course_metadata/waffle.py
```diff
@@ -1,16 +0,0 @@
...
-This module contains configuration settings via
-waffle switches for the course_metadata app.
-"""
-import waffle
-
-# Switches
-MASTERS_COURSE_MODE_ENABLED = u'masters_course_mode_enabled'
-
-
-def masters_course_mode_enabled():
```
mikix, line 11: This was unrelated, just noticed these were left here from previous Publisher cleanup.

course_discovery/apps/publisher/models.py
```diff
@@ -1,283 +1,15 @@
-import logging
```
mikix, line 1: We're leaving just three models: UserAttributes (used to turn off emails for specific publisher users), OrganizationExtension, and OrganizationUserRole.

Dillon-Dumesnil (approved):
Just a reminder that the responsibility of the code touched is on the author and not the approver. So sure.

PR #2384, opened 2020-01-23
DISCO-1572 course type org whitelist

xnick421:
- make org whitelist for course types in django admin panel use a
  checklist instead of traditional many-to-many list

- this change also changes entitlement types for course types to
  use a checklist as well

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -1,5 +1,6 @@
 from adminsortable2.admin import SortableAdminMixin
 from django.contrib import admin, messages
+from django.forms import CheckboxSelectMultiple, ModelForm
 from django.http import HttpResponseRedirect
 from django.urls import reverse
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/admin.py
```diff
@@ -120,6 +122,9 @@ class CourseRunTypeAdmin(admin.ModelAdmin):
 class CourseTypeAdmin(admin.ModelAdmin):
     list_display = ['uuid', 'name']
     search_fields = ['uuid', 'name']
+    formfield_overrides = {
+        models.ManyToManyField: {'widget': CheckboxSelectMultiple},
```
mikix: Interesting. I would have expected this to be on a field-by-field basis (like, 'I want the "org_whitelist" field to use CheckboxSelectMultiple, but leave other many-to-many fields alone'). This is fine, just odd that django does it at this level.

mikix (approved):
I assume this is beautiful in the UI. Where's my @karengu style gif though?

course_discovery/apps/course_metadata/admin.py
xnick421: did some digging around, changed it so it only effects that one field

PR #2387, opened 2020-01-24
Make draft versions of Studio-made runs

mikix:
If a course already exists in Publisher as a draft world, when we notice new runs made in Studio, let's set up a draft version of the run as well as the official.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -826,16 +826,11 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         """ Official rows just return whatever slug is active, draft rows will first look for an associated active
          slug and, if they fail to find one, take the slug associated with the official course that has
          is_active_on_draft: True."""
-        if not self.draft:
-            return self.url_slug_history.get(is_active=True).url_slug
-        else:
-            try:
-                return self.url_slug_history.get(is_active=True).url_slug
-            except ObjectDoesNotExist:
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -826,16 +826,11 @@ def active_url_slug(self):
...
-            except ObjectDoesNotExist:
-                # current draft url slug has already been published at least once, so get it from the official course
-                if self.official_version:
-                    return self.official_version.url_slug_history.get(is_active_on_draft=True).url_slug
-                return None
+        active_url = self.url_slug_history.filter(is_active=True).first()
+        if not active_url and self.draft and self.official_version:
+            # current draft url slug has already been published at least once, so get it from the official course
+            active_url = self.official_version.url_slug_history.filter(is_active_on_draft=True).first()
+        return getattr(active_url, 'url_slug', None)
```
mikix, line 833: @rgraber Hey - during unit testing, I noticed that this method is fairly strict (prone to raise if the world doesn't look like it expects). We may validly have official versions that didn't go through normal Publisher flow for runs that directly come from Studio (which is how all open edx installations work).

So I reworked this method to be a little more forgiving about missing url slugs. Do my changes make sense?

Dillon-Dumesnil (approved):
LGTM

course_discovery/apps/course_metadata/models.py
rgraber: Makes sense to me, as long as we have tests that ensure that going through publisher will create url slugs correctly.

rgraber (approved):

PR #2389, opened 2020-01-27
BOM-1170 : Django 2.2 codemods applied

mzulqarnain1:
JIRA Link : [BOM-1170](https://openedx.atlassian.net/browse/BOM-1170)


mzulqarnain1:
Only instance of change codemods caught already had on_delete value as a positional argument.

PR #2395, opened 2020-01-30
Upgrade django-debug-toolbar

iamsobanjaved:
Related JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1229)

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -2,7 +2,6 @@
 django<1.12
 
 # FIXME: these need Django 2
-django-debug-toolbar<2
 django-guardian<2
 
 # FIXME: 4.x dropped UploadToAutoSlug, we need to port to django-dynamic-filenames
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -31,7 +31,7 @@ django-choices==1.7.1
 django-compressor==2.4
 django-contrib-comments==1.9.2
 django-cors-headers==2.5.3
-django-debug-toolbar==1.11
+django-debug-toolbar==2.2
 django-elasticsearch-debug-toolbar==2.0.0
 django-extensions==2.2.6
 django-filter==2.2.0


requirements/constraints.txt
```diff
@@ -2,7 +2,6 @@
 django<1.12
 
 # FIXME: these need Django 2
-django-debug-toolbar<2
```
mikix, line 5: Oh cool, I thought DDT had dropped support for django 1.x altogether, but I see it merely dropped some deprecated stuff.

mikix (approved):
Your tests might fail - our requirements bot upgrader has recently had tests fail on make upgrade and we haven't sorted it out yet - https://github.com/edx/course-discovery/pull/2393

requirements/constraints.txt
```diff
@@ -24,3 +23,5 @@ django-cors-headers<3
 
 # FIXME: 3.2 changed the 'id' field in responses from a number to a string - need to validate if that's a safe change
 django-autocomplete-light<3.2
+
+edx-rest-api-client<2.0
```
mikix: Can you add a comment to this line? Ideally later folks can always see why a constraint exists, so they know when they can remove it. In this case, I'd say something about it breaking tests because of the new caching (causing less requests than tests expect to be made).

mikix: (But awesome job tracking down which thing broke tests. Thanks!)

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -21,6 +21,9 @@ def api_url(self):  # pragma: no cover
 
     def assert_api_called(self, expected_num_calls, check_auth=True):
         """ Asserts the API was called with the correct number of calls, and the appropriate Authorization header. """
+        if responses.calls[0].request.path_url != '/oauth2/access_token':
+            expected_num_calls -= 1
```
mikix: That seems like a brittle fix.

1) What if the login request is not the first call made in the test?
2) If I write a new test in 2021, do I have to falsely inflate the number of calls expected, because this assert function tries to adjust for an off-by-one error in early 2020?

Would it be hard to instead decrement the number passed to `assert_api_called` in just the tests that failed?

requirements/constraints.txt
```diff
@@ -23,4 +22,4 @@ transifex-client<0.13
 django-cors-headers<3
 
 # FIXME: 3.2 changed the 'id' field in responses from a number to a string - need to validate if that's a safe change
-django-autocomplete-light<3.2
+django-autocomplete-light<3.2
```
mikix: Is this just git cruft?

iamsobanjaved: Yes

iamsobanjaved:
Will rebase with #2398 after it gets merge

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -21,6 +21,7 @@ def api_url(self):  # pragma: no cover
 
     def assert_api_called(self, expected_num_calls, check_auth=True):
         """ Asserts the API was called with the correct number of calls, and the appropriate Authorization header. """
+
```
mikix: We don't usually leave a gap after a docstring, do we?

mikix (approved):

jmbowman (approved):

PR #2398, opened 2020-01-31
Switch jsonfield to jsonfield2 BOM-1230

mraarif:
Relevant JIRA issue can be found [here](https://openedx.atlassian.net/browse/BOM-1230).

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -24,3 +24,6 @@ django-cors-headers<3
 
 # FIXME: 3.2 changed the 'id' field in responses from a number to a string - need to validate if that's a safe change
 django-autocomplete-light<3.2
+
+# jsonfield2 3.1.0 drops support for python 3.5
+jsonfield2==3.0.3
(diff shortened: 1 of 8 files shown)


awais786 (approved):

jlajoie (approved):

requirements/constraints.txt
```diff
@@ -24,3 +24,9 @@ django-cors-headers<3
 
 # FIXME: 3.2 changed the 'id' field in responses from a number to a string - need to validate if that's a safe change
 django-autocomplete-light<3.2
+
+# jsonfield2 3.1.0 drops support for python 3.5
+jsonfield2==3.0.3
+
+# edx-rest-api-client version >2 caches requests that fails the check for number of API calls
+edx-rest-api-client<2.0.0
```
jmbowman: @robrap Do you think this is a problem in edx-rest-api-client, or are some code changes needed here to use the new version?

jmbowman (approved):

requirements/constraints.txt
mikix: I'm guessing it's just test code that's like "expect 3 network requests" but the new caching client only does 2. So if true, we'd need to make some test code changes in Disco.

robrap: Hmm...I'm not looking at the code, but it might be more accurate to search the calls that the client is making for the call you expect, and ignore the auth calls that may or may not be made.  Or, you can clear the cache at the beginning of tests and update the count check to match today's truth.

jmbowman (approved):

PR #2399, opened 2020-01-31
Upgrade django-autocomplete-light BOM-1233

mraarif:
Relevant JIRA issue can be found [here](https://openedx.atlassian.net/browse/BOM-1233).

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -22,8 +22,5 @@ transifex-client<0.13
 # FIXME: 3+ require schemes in CORS_ORIGIN_WHITELIST URLs - remember to update configuration when you remove this
 django-cors-headers<3
 
-# FIXME: 3.2 changed the 'id' field in responses from a number to a string - need to validate if that's a safe change
-django-autocomplete-light<3.2
-
 # jsonfield2 3.1.0 drops support for python 3.5
 jsonfield2==3.0.3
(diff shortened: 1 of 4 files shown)


requirements/constraints.txt
```diff
@@ -22,5 +22,5 @@ transifex-client<0.13
 # FIXME: 3+ require schemes in CORS_ORIGIN_WHITELIST URLs - remember to update configuration when you remove this
 django-cors-headers<3
 
-# FIXME: 3.2 changed the 'id' field in responses from a number to a string - need to validate if that's a safe change
```
mikix, line 25: What about the number to string thing? Isn't that technically a breaking API change for any autocomplete consumers? Which are... django admin and Publisher I guess? If we assume nobody is doing their own scripting.

mraarif:
@jmbowman aside from `id` field is returned as a string, newer [version](https://github.com/yourlabs/django-autocomplete-light/blob/bc83aecdde2328e683b278149ff7cdce3ebf4065/src/dal_select2/views.py#L21) of `django-autocomplete-light` is returing an additional item in response `selected_text`.

What other checks can be performed (apart from tests) to make sure this change isn't going to break?

jmbowman:
You should be able to switch to your branch in devstack and open the relevant pages there.  The port is listed in the devstack README.  If you have trouble, the #always-available-squad will probably be better able to help than I am.

requirements/constraints.txt
mraarif: image

I've tested this change in admin while creating a Program and selecting the instructor ordering on that page, so far it seems to have worked fine.

mikix (approved):
Does adding an instructor work in Publisher too? (Go to http://localhost:18400/ , pick a course, and use the instructor search field on one of the course runs). If so, looks great.

mikix:
I tested Publisher locally - still works. :+1:

jmbowman (approved):

PR #2402, opened 2020-02-03
BOM-1234 --- Upgrade django-stdimage version

mzulqarnain1:
-Added django-dynamic-filenames dependency to replace UploadToAutoSlug
 with FilePattern
-Modified code to use FilePattern instead of deprecated method
 UploadToAutoSlug of stdimage
-Removed django-stdimage from pins
-Updated Requirements
-Updated Migrations
-Created new Migrations of ProgramType Model


mikix (commented):
How'd you get `make upgrade` to bump stdimage? It's pinned in constraints.txt as `django-stdimage<4`

Also, most of this PR (the make upgrade bits) are already done in like, 3 other PRs right now upgrading bits of Discovery. If just one of those lands, this PR will be tiny.

mzulqarnain1:
> How'd you get `make upgrade` to bump stdimage? It's pinned in constraints.txt as `django-stdimage<4`
> 
> Also, most of this PR (the make upgrade bits) are already done in like, 3 other PRs right now upgrading bits of Discovery. If just one of those lands, this PR will be tiny.

@mikix my bad, I forgot to add changes in constraint.txt to git. I have pushed that change too.

course_discovery/apps/course_metadata/models.py
```diff
@@ -308,8 +308,9 @@ class ProgramType(TimeStampedModel):
                               'associated courses, but enrolled in other seat types, will NOT have their completion '
                               'of the course counted toward the completion of the program.'),
     )
+    upload_to_pattern = FilePattern(filename_pattern='media/program_types/logo_images')
```
mikix: Usually model-level fields are model fields. But this isn't? Why not just have it inline like UploadToAutoSlug was?

course_discovery/apps/course_metadata/models.py
```diff
@@ -308,8 +308,9 @@ class ProgramType(TimeStampedModel):
...
     logo_image = StdImageField(
-        upload_to=UploadToAutoSlug(populate_from='name', path='media/program_types/logo_images'),
+        upload_to=upload_to_pattern,
```
mikix: You're dropping the `populate_from='name'` bit. Looks like you can use {instance.name} or something, like the examples in https://github.com/codingjoe/django-dynamic-filenames.

course_discovery/apps/course_metadata/models.py
mzulqarnain1: moved inline

course_discovery/apps/course_metadata/models.py
mzulqarnain1: Figuring this out.

course_discovery/apps/course_metadata/models.py
```diff
@@ -309,7 +309,7 @@ class ProgramType(TimeStampedModel):
                               'of the course counted toward the completion of the program.'),
     )
     logo_image = StdImageField(
-        upload_to=UploadToAutoSlug(populate_from='name', path='media/program_types/logo_images'),
+        upload_to=FilePattern(filename_pattern='media/program_types/logo_images/{instance.name:.40slug}{ext}'),
```
Ayub-Khan: @mikix I think this should resolve the name issue.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -286,13 +286,14 @@ def ensure_draft_world(obj):
...
-class UploadToFieldNamePath(UploadTo):
+class UploadToFieldNamePath(FilePattern):
     """
     This is a utility to create file path for uploads based on instance field value
     """
     def __init__(self, populate_from, **kwargs):
         self.populate_from = populate_from
-        super(UploadToFieldNamePath, self).__init__(populate_from, **kwargs)
+        kwargs['populate_from'] = populate_from
+        super(UploadToFieldNamePath, self).__init__(**kwargs)
```
mikix: Nit: might look more natural to do:
`super(UploadToFieldNamePath, self).__init__(populate_from=populate_from, **kwargs)`

course_discovery/apps/course_metadata/models.py
mikix: :+1:

mikix (approved):

course_discovery/apps/course_metadata/utils.py
Ayub-Khan: @mikix i think we need to revisit our custom class. Im looking into it.

Ayub-Khan:
@mikix review again.

course_discovery/apps/course_metadata/models.py
mikix: Could this be replaced with `UploadToFieldNamePath`?

requirements/local.txt
```diff
@@ -44,8 +45,8 @@ django-rest-swagger==2.2.0
 django-simple-history==2.8.0
 django-solo==1.1.3
 django-sortedm2m==3.0.0
-django-stdimage==3.2.0
-django-storages==1.8
+django-stdimage==5.0.2
+django-storages==1.9
```
mikix: Wait, this upgrade is bad -- it drops a boto backend we need. Disco now has a constraint on this, so re-run `make upgrade` please.

awais786 (approved):

jmbowman (approved):

mikix (approved):

PR #2404, opened 2020-02-03
Fix runs resetting status accidentally

mikix:
When submitting a run via Publisher, we would mark any reviewed courses as unreviewed, if the course had changed in a meaningful way.

But we had a bug in that 'meaningful' check that meant we'd always reset the runs if the course had an entitlement.

https://openedx.atlassian.net/browse/DISCO-1575


jmyatt (approved):

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -269,7 +269,7 @@ def update_entitlement(self, course, entitlement_type, price, partial=False):
             data = {'mode': entitlement_type.slug, 'price': price}
             serializer = CourseEntitlementSerializer(entitlement, data=data, partial=partial)
             serializer.is_valid(raise_exception=True)
-            return serializer.save(), entitlement.price != price
+            return serializer.save(), entitlement.price != float(price)
```
Dillon-Dumesnil, line 272: Nice catch

Dillon-Dumesnil (approved):

course_discovery/apps/api/v1/views/courses.py
mikix: Yes. No need to git blame the original line...

PR #2426, opened 2020-02-12
Adding dockerfile

coryleeio:


jlajoie (approved):
Looks reasonable, typically we access discovery in devstack on 18381, not sure if this is meant to have parity with that or if we'll need to change devstack and what ports get exposed or this is good as is.

coryleeio:
@jlajoie YThanks!

You don't actually have to change anything - the expose directive doesn't do much
https://docs.docker.com/engine/reference/builder/#expose

Since this is a production centric file I used the production port whichis 8381, for devstack you could simply set the command to use the python run server on 18381 in compose and proceed as normal

fredsmith (approved):

PR #2429, opened 2020-02-14
Filter unmarketable runs from searches

mikix:
Don't have course runs that are listed as non-marketable (like Masters or Empty runs) show up in endpoint searches that fuel the marketing site.

https://openedx.atlassian.net/browse/DISCO-1589

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes.py
+++ b/course_discovery/apps/course_metadata/search_indexes.py
@@ -40,6 +40,10 @@ TITLE_FIELD_BOOST = 25.0
 ORG_FIELD_BOOST = TITLE_FIELD_BOOST
 
 
+def filter_visible_runs(course_runs):
+    return course_runs.exclude(type__is_marketable=False)
+
+
 class OrganizationsMixin:
     def format_organization(self, organization):
         return '{key}: {name}'.format(key=organization.key, name=organization.name)
@@ -172,7 +176,7 @@ class CourseIndex(BaseCourseIndex, indexes.Indexable):
         return 'course:{}'.format(obj.key)
 
     def prepare_course_runs(self, obj):
-        return [course_run.key for course_run in obj.course_runs.all()]
+        return [course_run.key for course_run in filter_visible_runs(obj.course_runs)]
 
     def prepare_expected_learning_items(self, obj):
         return [item.value for item in obj.expected_learning_items.all()]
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/v1/views/search.py
```diff
@@ -22,6 +22,16 @@
...
 
 
+def filter_visible(queryset):
+    """
+    Filters out runs that shouldn't show up in a search.
+
+    Arguments:
+        queryset: a SearchQuerySet object over course runs
+    """
+    return queryset.filter(published=True).exclude(hidden=True).filter(is_marketable=True)
```
mikix: The addition here of `filter(is_marketable=True)` is the real change in this PR. Rest is refactoring or tests.

Dillon-Dumesnil (approved):
LGTM

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -222,7 +227,6 @@ class CourseRunIndex(BaseCourseIndex, indexes.Indexable):
     language = indexes.CharField(null=True, faceted=True)
     transcript_languages = indexes.MultiValueField(faceted=True)
     pacing_type = indexes.CharField(model_attr='pacing_type', null=True, faceted=True)
-    is_marketable = indexes.BooleanField(model_attr='type__is_marketable', faceted=True)
```
Dillon-Dumesnil, line 225: Does this require a migration or are indexes special?

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -254,6 +258,12 @@ def read_queryset(self, using=None):
             'seats__type',
         )
 
+    def index_queryset(self, using=None):
+        return filter_visible_runs(super().index_queryset(using=using))
+
+    def load_all_queryset(self):
+        return self.index_queryset()
```
Dillon-Dumesnil: Just to be on the same page, here you are just overwriting these to use the filter rather than the super version?

Dillon-Dumesnil (approved):
Very nice

course_discovery/apps/course_metadata/search_indexes.py
mikix: Apparently indexes don't need migrations.

mikix: I guess the migration is just "rebuild the index"

course_discovery/apps/course_metadata/search_indexes.py
mikix: Yes. `index_queryset` is the right place to do that. And I'm overloading `load_all_queryset` too because the base model is a redefinition of `index_queryset` rather than calling it. Which seems like a bug, but maybe there's a purpose there.

PR #2431, opened 2020-02-18
BOM-1235 Added tox for multiple Django versions

UsamaSadiq:
**Ticket:** [BOM-1235](https://openedx.atlassian.net/browse/BOM-1235)


tox.ini
```diff
@@ -0,0 +1,27 @@
+[tox]
+envlist = {py35}-django{111,20,21,22},quality
```
mikix: Do we actually care about supporting 2.0 and 2.1? They are EOL, right?

https://www.djangoproject.com/download/#supported-versions

Makefile
```diff
@@ -46,6 +46,10 @@ upgrade:
 	pip-compile --upgrade -o requirements/docs.txt requirements/docs.in
 	pip-compile --upgrade -o requirements/local.txt requirements/local.in
 	pip-compile --upgrade -o requirements/production.txt requirements/production.in
+	# Let tox control the Django version for tests
+	grep -e "^django==" requirements/local.txt > requirements/django.txt
+	sed '/^[dD]jango==/d' requirements/local.txt > requirements/local.tmp
+	mv requirements/local.tmp requirements/local.txt
```
mikix: Why not use `-i` instead of a move?

tox.ini
```diff
@@ -0,0 +1,27 @@
...
+skipsdist=true
+
+[pycodestyle]
+exclude = .git,.tox,migrations
+max-line-length = 120
```
mikix: A more advanced version of this already exists in `.pycodestyle` -- I'd say either move that config into this file directly or reference that config here.

tox.ini
```diff
@@ -0,0 +1,27 @@
...
+	npm install --unsafe-perm
+	$(NODE_BIN)/bower install --allow-root
+deps =
+    django111: -r requirements/django.txt
+    django20: Django>=2.0,<2.1
+    django21: Django>=2.1,<2.2
+    django22: Django>=2.2,<2.3
+	-r{toxinidir}/requirements/local.txt
+commands =
+    pytest {posargs}
```
mikix: This is going to chang how we run tests locally, yeah? We have some documentation about that in the README that will need updating too.

mikix:
Oh sorry, didn't see this was labelled as a WIP. (You can create a PR as a "draft PR" which won't ping the code owners when you first file it in github -- it's a little drop down arrow on the create button in github)

pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
+DJANGO_SETTINGS_MODULE = course_discovery.settings.test
 testpaths = course_discovery/apps
-addopts = -n2 --dist=loadscope --cov=course_discovery --cov-report term --cov-config=.coveragerc --no-cov-on-fail --ds=course_discovery.settings.test
```
jmbowman, line 3: Why were these options removed?  My guess is that the failing test usually gets called before the result of a particular query is cached, but since all the tests are now running on a single worker, there was an earlier test in the run that cached the results of the query.  The test should probably be fixed anyway to pass regardless of whether that query was cached beforehand or not.

Makefile
```diff
@@ -46,11 +46,15 @@ upgrade:
...
 	chmod a+rw requirements/*.txt
 
 test: clean ## Run tests and generate coverage report
 	## The node_modules .bin directory is added to ensure we have access to Geckodriver.
-	PATH="$(NODE_BIN):$(PATH)" pytest --ds=course_discovery.settings.test --durations=25
+	PATH="$(NODE_BIN):$(PATH)" pytest --ds=course_discovery.settings.test
```
jmbowman: Was this removed just to simplify the output?  It's kind of useful to have the slowest test durations, but I can understand removing it if we're not acting on this data yet.

requirements/local.txt
```diff
@@ -52,7 +52,6 @@ django-taggit-serializer==0.1.7
 django-taggit==1.2.0
 django-waffle==0.20.0
 django-webpack-loader==0.6.0
-django==1.11.28
```
jmbowman: `make requirements` should probably be updated to install both `local.txt` and `django.txt` in the same pip command, so dev environments get the correct version of Django.  Most of our other repos have separate test and dev requirements files, but this one doesn't.

jmbowman (approved):

jmbowman:
@mikix I think this is ready for final review now.

UsamaSadiq:
jenkins run all

tox.ini
```diff
@@ -0,0 +1,21 @@
+[tox]
+envlist = {py35}-django{111,20,21,22}
```
mikix, line 2: Again, let's not test 2.0 and 2.1 -- they are EOL and edX doesn't care about them (to my knowledge).

tox.ini
```diff
@@ -0,0 +1,21 @@
...
+skipsdist=true
+
+[pytest]
+DJANGO_SETTINGS_MODULE = course_discovery.settings.test
+testpaths = course_discovery/apps
+
+[testenv]
+deps =
+    django111: -r requirements/django.txt
```
mikix, line 11: Is this the only reason we separate out django.txt? Feels out of line with the rest of the django test envs below.

Maybe it would be better to align with our other repos and have a test.txt requirements file?

mikix (commented):
Looks mostly fine, I have a couple nit comments (`sed -i` and using test.txt) but would approve once you drop django20 and django21.

tox.ini
jmbowman: We are actually testing 2.0 and 2.1 because some deprecation warnings were added in those versions for code changes that take effect in 2.2.  We'll drop testing for them shortly after deploying to prod with 2.2.

tox.ini
jmbowman: We use the file rather than a range so that we're testing with the exact Django version that we currently deploy with.  And I'd be in favor of refactoring the requirements files to match what's done elsewhere, just not sure if this PR is the right place to do that (vs. a separate PR dedicated to that change).

tox.ini
mikix: OK

tox.ini
mikix: Sure. My beef was more with "django111: fromfile" -- which if you're doing that so that the file can move on ya, feels like you should do either "django: fromfile" or "django111: django==1.11" but not mix the two, you know? But this is fine for now.

mikix (approved):

Makefile
UsamaSadiq: updated with the use of `-i`.

PR #2434, opened 2020-02-18
Removing unused auth classes

jinder1s:
BearerAuthentication and BasicAuthentication classes are not used by our system.
Additionally, BearerAuthentication is deprecated and will soon be removed from edx-drf-extensions.


Dillon-Dumesnil (approved):
Sure

PR #2437, opened 2020-02-19
WS-681 use course run key in course run url slug

rgraber:


docs/decisions/0010-update-course-run-slugs.rst
```diff
@@ -0,0 +1,23 @@
+Course Run Keys in Course Run URL slugs
```
srwang, line 1: how do rst files work? how did you decide that this is 0010?  these are docs that are in the code base and for future devs to reference?

rgraber: 1. rst is just the rich text file extension, it means we can edit and view documents with fancy formatting in plain text editors
2. the docs are sequential, so the previous one was 0009
3. yes

srwang (approved):
🐌

mikix (approved):

docs/decisions/0010-update-course-run-slugs.rst
albemarle: @srwang RST was originally a Python documentation format.  It's actually not very common outside of that context, but it's similar to Markdown.

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -571,7 +571,7 @@ def setUpTestData(cls):
         )
 
     def test_draft_does_not_collide_with_draft(self):
-        with self.assertNumQueries(80, threshold=0):
+        with self.assertNumQueries(77, threshold=0):
```
albemarle, line 574: Did you ever discover the reason for this change?

docs/decisions/0010-update-course-run-slugs.rst
```diff
@@ -0,0 +1,23 @@
...
+
+Context
+-------
+
+The marketing site used to use the most recent (modulo some business logic) course run URL as the main URL for the whole course, redirecting all other course run URLs to the currently active course run url. At this time, course run URLs were generated from the course title, while course URLs (mostly unused) were generated from the course key. Now, the marketing site uses course URLs, and all course run URLs redirect to the single course URL. When the switch to single course URLs was made, these url slugs were updated to use the course title instead of the course key, since the course title was more human-readable and SEO-friendly. This has led to collisions between course URLs and course run URLs of courses with similar titles. 
+
+Decision
+--------
+
+In order to decrease the likelihood of collisions, going forward, all course run URLs will be generated from a combination of the course title and the course run key, eg http://edx.org/course/my-course-course-v1edxcmct12020 instead of http://edx.org/course/my-course
```
albemarle, line 20: +1 for example

course_discovery/apps/course_metadata/tests/test_signals.py
rgraber: yes, it was because django didn't have to work as hard to make a unique slug. when it is trying to generate a unique slug for something it will just make queries for Thing.filter(slug='my-slug-{n++}') until it gets an empty set

docs/decisions/0010-update-course-run-slugs.rst
```diff
@@ -0,0 +1,23 @@
...
+Status
+------
+
+Accepted
+
+
+Context
+-------
+
+The marketing site used to use the most recent (modulo some business logic) course run URL as the main URL for the whole course, redirecting all other course run URLs to the currently active course run url. At this time, course run URLs were generated from the course title, while course URLs (mostly unused) were generated from the course key. Now, the marketing site uses course URLs, and all course run URLs redirect to the single course URL. When the switch to single course URLs was made, these url slugs were updated to use the course title instead of the course key, since the course title was more human-readable and SEO-friendly. This has led to collisions between course URLs and course run URLs of courses with similar titles.
```
albemarle: This paragraph is a little hard for me to follow.

Suggestions:
* s/At this time/At that time/
* not sure that (mostly unused) helps here-- it made me wonder more about it
* maybe new paragraph starting at Now ?
* possible s/Now/Currently/ or "Since DATE" or...?
* sometimes "the marketing site" means Drupal and sometimes it means Prospectus-- is it worth differentiating?
* "these url slugs" - change to URL and also "these" is confusing
* I like the explanation of the change to course title
* replace "This" in last sentence with something like "However, because some courses have similar titles, we began seeing..."

(That's a lot, take what seems useful.  :-) )

albemarle:
Suggestion: do a search and replace sweep for lowercase url.

PR #2438, opened 2020-02-20
small PR to add the type slug to the Program response

staubina:
Small PR to add the slug for Program Type.


kdmccormick (approved):
Looks good :+1: 

You're obviously more in-tune with the architecture of this API than I am, but there may benefit to deprecating the `type` field or otherwise making it clear that `type_slug` is preferred. Take that as you will -- not a blocking issue.

staubina:
Deprecating fields is more difficult than adding fields to an API because of the potential for breaking the contract with users.

I think we need to form a V2 of this API eventually and slowly add to it what is used and needed rather than deprecating fields.

@kdmccormick thanks for the review 👍

jlajoie (approved):
👍

mikix:
In the future, we've been trying to move towards using uuids more often for this sort of thing. Especially across services, where renaming things is nigh-impossible.

CourseTypes and CourseRunTypes are referenced by UUID in our serializers. ProgramType ought to as well (though that would involve adding a uuid field to the model).

PR #2443, opened 2020-02-25
Missing migration for PR#2437

staubina:


jlajoie (changes requested):
Going to hold off on +1'ing this until we know what the intent is for this migration (and what caused the model change) and if it is needed or not.

rgraber:
This was a missing piece of https://github.com/edx/course-discovery/pull/2437 . we decided to add the course run key to the course run url slug to avoid the conflicts we'd been running into with single course URLs. There's an ADR in the PR if you want to know more

jlajoie (approved):
👍

PR #2445, opened 2020-02-25
Added the ProgramTypeAttrs to the serializer and model

staubina:
[MICROBA-150]

Updated the ProgramType to have a UUID and a Coaching support boolean.
Then serialized this data as a type_attrs for type attributes in the
API reponse for Programs.


course_discovery/apps/api/serializers.py
```diff
@@ -648,14 +648,36 @@ class Meta:
         fields = ('id', 'name', 'query', 'courses_count', 'viewers')
 
 
+class ProgramTypeSerializer(BaseModelSerializer):
+    """ Serializer for the Program Types. """
```
staubina, line 652: I just moved this serializer up

course_discovery/apps/course_metadata/migrations/0236_add_program_type_uuid_and_coaching.py
```diff
@@ -0,0 +1,50 @@
...
+from django.db import migrations, models
+import uuid
+
+
+def add_uuid_to_program_types(apps, schema_editor):
+    program_type = apps.get_model('course_metadata', 'ProgramType')
+
+    for obj in program_type.objects.all():
+        obj.uuid = uuid.uuid4()
+        obj.save()
```
mikix, line 14: In the rare case that you are reversing and forwarding migrations, this would overwrite any existing uuids. Might consider an `if` guard?

mikix (approved):
Looks great, thanks so much for doing this extra work.

course_discovery/apps/course_metadata/migrations/0236_add_program_type_uuid_and_coaching.py
staubina: So this migration will currently drop the column in a reverse migration.  Are you proposing we preserve it somehow?

mikix: Uh, nope. I forgot that it would be dropped. Carry on.

PR #2446, opened 2020-02-25
Add management command to dedupe course metadata

pwnage101:
Course and CourseRun metadata history tables largely contain duplicate
entries, where the only difference between subsequent records is the
`modified` column.  This breaks the clean_duplicate_history management
command provided by django-simple-history, so this PR provides a
slightly tweaked version of that management command to fix that issue
for our purposes.

DENG-20

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/deduplicate_course_metadata_history.py
+++ b/course_discovery/apps/course_metadata/management/commands/deduplicate_course_metadata_history.py
@@ -1,0 +1,36 @@
+"""
+Deduplicate course metadata history rows that were unnecessarily created while running
+refresh_course_metadata.
+
+This largely inherits the internals from the clean_duplicate_history management command
+provided by django-simple-history, with only one minor tweak (to ignore the `modified`
+field while comparing potentially duplicate records).
+
+Usage: identical to clean_duplicate_history:
+
+  python manage.py deduplicate_course_metadata_history --dry course_metadata.Course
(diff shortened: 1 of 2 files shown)


brianhw (approved):
LGTM, based on our earlier discussions.   👍

Dillon-Dumesnil (approved):
Looks great. Do you know if we only have this problem with our Course and CourseRun models? I know we also have the history enabled on Seat and CourseEntitlement models which are also affected in Refresh Course Metadata.

course_discovery/apps/course_metadata/management/commands/deduplicate_course_metadata_history.py
```diff
@@ -0,0 +1,40 @@
...
+  python manage.py deduplicate_course_metadata_history course_metadata.Course
+  python manage.py deduplicate_course_metadata_history course_metadata.CourseRun
+
+https://django-simple-history.readthedocs.io/en/latest/utils.html#clean-duplicate-history
+"""
+import logging
+
+from simple_history.management.commands import clean_duplicate_history
+
+logger = logging.getLogger(__name__)
```
doctoryes: Nit: Are the `import logging` and this logger get actually needed here?

doctoryes (approved):

course_discovery/apps/course_metadata/management/commands/deduplicate_course_metadata_history.py
pwnage101: Nope.

PR #2450, opened 2020-02-26
WS-655 index content for search

rgraber:


rgraber:
This is ready for review but cannot be committed until the Algolia setup is in place in configuration/edx-internal. This should not affect anything else in the code.

rgraber:
Algolia automatically provides the management command algolia_reindex as part of the library, which rebuilds the index based on whatever is in MyModel.objects.all()

course_discovery/apps/course_metadata/models.py
```diff
@@ -1003,6 +1004,64 @@ def set_active_url_slug(self, slug):
                 other_slug.is_active_on_draft = False
                 other_slug.save()
 
+    @cached_property
```
srwang, line 1007: is this a non-algolia related refactor or is it also being used somewhere

course_discovery/apps/course_metadata/migrations/0237_algoliaproxycourse_algoliaproxyprogram.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.28 on 2020-02-27 16:38
+from __future__ import unicode_literals
+
+from django.db import migrations
+import django.db.models.manager
+
+
+class Migration(migrations.Migration):
+
```
srwang, line 10: i am dumb and is this just fake data to feed algolia or algolia spits something back and this gets called on the front end instead of our api or neither

course_discovery/apps/course_metadata/index.py
```diff
@@ -0,0 +1,34 @@
...
+
+@register(AlgoliaProxyCourse)
+class CourseIndex(AlgoliaIndex):
+    search_fields = ('full_description', 'outcome', ('partner_name','partner'), 'short_description', 'title',)
+    # partner is also a facet field, but we already declared it above
+    facet_fields = ('availability', ('subject_names', 'subjects'), ('level_type_name', 'level'),
+                    ('active_run_language','language'),)
+    result_fields = ('active_run_key', 'active_run_start', 'active_run_type', 'active_url_slug', 'image_src', 'owners',
+                     'program_types', ('uuid', 'objectID'))
+    fields = search_fields + facet_fields + result_fields
```
srwang, line 16: search fields = stuff i can search thru to match text?
facet fields = attributes that can be calculated on courses?
result_fields = attributes that exist on courses?

course_discovery/apps/course_metadata/models.py
rgraber: this was originally a method on the courseWithProgram serializer since prospectus was the only thing that needed it. Now that we need it in Algolia as well I figured I'd move it back on the model and cache the property so we don't have to recalculate it for every field that uses it

course_discovery/apps/course_metadata/migrations/0237_algoliaproxycourse_algoliaproxyprogram.py
rgraber: I'm not sure I understand. The migration is just something Django does whenever a model is changed/added/removed in case the database needs to know about it

course_discovery/apps/course_metadata/index.py
rgraber: Facets are basically filters we're going to put in the sidebar (eg Subject, Level, Language, etc.). Result fields contain information we don't need to search, but will need to create the discovery cards out of the results

course_discovery/apps/course_metadata/models.py
reillz10: Do we have a hook to recalculate it? This field changes when there is a better course run available. I may just be missing where we have hooks to update this value.

rgraber: A cached_property in Django persists as long as the instance does. Since every time we hit the /courses endpoint we load a new instance we should be ok, especially because nobody else uses this value.

rgraber: We might have an inconsistent value if something changes *while* we're indexing, but it's probably better that we treat indexing as an atomic operation so we don't have details from different course runs on the card.

mikix (approved):
I don't understand the fine details of this, but we had talked about how to set up the indexes with proxy classes and that stuff looks good here. I think the overall approach is fine.

(How does this get exposed for search?)

rgraber:
@mikix  we query the hosted Algolia app directly from Prospectus, so there should be no new auth concerns in discovery.

PR #2451, opened 2020-02-27
WS-680 ensure that authoring orgs persist a configurable order. Since order determines order of logos on about page.

srwang:


rgraber (approved):
LGTM, sorry it was such a pain to get that second test

jlajoie (approved):
I'm all about more test coverage, though it looks like this is specific to determining the order of auth orgs from the sortedm2m field. I personally don't see a ton of value in testing a third party library's code, when that library already has its own tests, but sure 👍

albemarle (approved):

PR #2457, opened 2020-03-03
WS-655 wrap Algolia reindex management command

rgraber:


course_discovery/apps/course_metadata/management/commands/reindex_all_algolia_indices.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Command(BaseCommand):
+
+    help = 'Wrapper command for Algolia\'s algolia_reindex'
+
+    def handle(self, *args, **options):
+        for model in get_registered_model():
+            counts = reindex_all(model, batch_size=1000)
+            logger.info('\t* {} --> {}'.format(model.__name__, counts))
```
mikix: Might be interesting to log the old count?

Also, I know the other regular index command we have (update_index.py in Disco) does a check for like, is the new count more than 10 or 20 percent or whatever different, to catch dumb errors like accidentally filtering out most courses from the index.

And then it adds a flag `--disable-change-limit` to avoid that. So if you did mean it, you just ask devops to run the command with that flag.

A possible future option, anyway.

course_discovery/apps/course_metadata/management/commands/reindex_all_algolia_indices.py
```diff
@@ -0,0 +1,16 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+
+    help = 'Wrapper command for Algolia\'s algolia_reindex'
+
+    def handle(self, *args, **options):
+        for model in get_registered_model():
+            counts = reindex_all(model, batch_size=1000)
```
mikix: Can this raise an exception? Currently, you'd abort the reindex for future models if so. Is that desired?

If so, you may want to explicitly add a django db transaction around the loop, so you'd also roll back past model indexes in that case, to avoid any index updates.

If not, I'd say catch and log em.

mikix (approved):

course_discovery/apps/course_metadata/management/commands/reindex_all_algolia_indices.py
rgraber: I think I will save that for the future. That would require more opening up of the Algolia black box than I think is necessary at this moment.

course_discovery/apps/course_metadata/management/commands/reindex_all_algolia_indices.py
rgraber: The inner code of reindex_all takes care of trying/catching AlgoliaExceptions for us but you have a point that it might be worth wrapping it in a try in case something else happens.

rgraber:
Clarification on this PR: We're hoping to be able to go back to using the provided reindex_algolia management command. However, for some reason it doesn't seem to work when run on Jenkins so this is mostly a test to see if there's something inherently wrong with the command or if we just have trouble calling commands from external libraries for some reason

rgraber:
> Clarification on this PR: We're hoping to be able to go back to using the provided reindex_algolia management command. However, for some reason it doesn't seem to work when run on Jenkins so this is mostly a test to see if there's something inherently wrong with the command or if we just have trouble calling commands from external libraries for some reason

this is also why I'm excluding it from test coverage. just trying to diagnose the plumbing issues

PR #2459, opened 2020-03-05
WS-655 don't index data with missing required fields

rgraber:


course_discovery/apps/course_metadata/algolia_proxy_models.py
```diff
@@ -79,17 +88,16 @@ def expected_learning_items_values(self):
...
+        return get_owners(self)
 
     def course_titles(self):
         return [course.title for course in self.courses.all()]
 
     def program_type(self):
-        return self.type.slug
+        if self.type:
+            return self.type.slug
+        return None
```
mikix, line 99: Small nit, but for such a small check, I think the tertiary check looks more readable:
`return self.type.slug if self.type else None`

But that's definitely just a matter of opinion.

mikix (approved):
Sure!

course_discovery/apps/course_metadata/algolia_proxy_models.py
```diff
@@ -47,14 +60,10 @@ def image_src(self):
...
-            image = getattr(owner, 'logo_image', None)
-            if image:
-                return image.url
-            return None
+        return get_owners(self)
 
-        return [{'key': o.key, 'logoImageUrl': getattr(logo_image(o), 'url', None)} for o in
-                self.authoring_organizations.all()]
+    def should_index(self):
+        return len(self.owners()) > 0 and self.active_url_slug and self.partner.name == 'edX'
```
srwang, line 66: I don't understand why things should only be indexed if the partner is edx

rgraber: The term 'partner' is annoyingly overloaded. In this case, the owner of a course may be Harvard, MIT, or whoever but the partner being edX just means it's part of the greater edX catalog.

mikix: Is there a way to make that less weirdly hardcoded to edx.org usage?

mikix: Like, if the goal is to only index the "main" site, you could check `self.partner.site.id == settings.SITE_ID` or something. But even that feels a little wonky

rgraber: what feels wonky about that?

rgraber: Actually how bad would it be to fix this in a follow-up PR? it would help WS a lot to get this out today.

srwang (approved):

rgraber:
I will do a follow-up PR to fix the hardcoded site thing

course_discovery/apps/course_metadata/algolia_proxy_models.py
mikix: > what feels wonky about that?

About checking the site id? It feels needlessly hardcoded, to me. Like, why does only that one site get indexing? Maybe this should be a Partner model checkbox? Or why not just index all sites?

PR #2463, opened 2020-03-09
WS-711 search ranking

rgraber:


course_discovery/apps/course_metadata/models.py
```diff
@@ -1045,9 +1015,10 @@ def has_started(course_run):
         marketable_course_runs = [course_run for course_run in self.course_runs.all() if course_run.is_marketable]
 
         for course_run in marketable_course_runs:
-            if has_started(course_run) and is_upgradeable(course_run) and user_has_time_to_finish_course(course_run):
+            course_run_started = (not course_run.start) or (course_run.start and course_run.start < now)
```
srwang, line 1018: why is a course run that doesn't have a start date an already started course run?

rgraber: Those are usually self-paced courses that someone can start whenever. In most places when we're dealing with course start logic we count them as started.

srwang (approved):

course_discovery/apps/course_metadata/index.py
```diff
@@ -12,12 +12,20 @@ class CourseIndex(AlgoliaIndex):
...
     result_fields = ('active_run_key', 'active_run_start', 'active_run_type', 'active_url_slug', 'image_src', 'owners',
-                     'program_types', ('uuid', 'objectID'))
+                     'program_types', ('uuid', 'objectID'), 'availability_rank')
     fields = search_fields + facet_fields + result_fields
     settings = {
         # searchableAttributes ordered by highest value
-        'searchableAttributes': ['title', 'short_description', 'full_description', 'outcome', 'partner'],
-        'attributesForFaceting': ['partner', 'subject', 'level', 'language', 'availability']}
+        'searchableAttributes': [
+            'unordered(title)',  # AG best practice: position of the search term within the title doesn't matter
```
mikix, line 20: Would that logic not also hold for description fields?

mikix (approved):

course_discovery/apps/course_metadata/index.py
rgraber: Difficult to say, actually. Usually the beginning of the description is some kind of summary of the course content, so if the keyword is in the summary, it's probably more relevant than if the keyword is later on. We might end up changing this later if we feel our results are incorrectly skewed.

course_discovery/apps/course_metadata/index.py
```diff
@@ -30,7 +38,14 @@ class ProgramIndex(AlgoliaIndex):
                     ('active_languages', 'language'), 'card_image_url')
     result_fields = ('course_titles', 'marketing_url', 'program_type', 'owners', ('uuid', 'objectID'))
     fields = search_fields + facet_fields + result_fields
-    settings = {'searchableAttributes': ['title', 'subtitle', 'overview', 'expected_learning_items_values', 'partner'],
-                'attributesForFaceting': ['partner', 'subject', 'level', 'language', 'availability']}
+    settings = {
+        'searchableAttributes': [
+            'unordered(title)',
```
Dillon-Dumesnil, line 43: Nit: Have the same comment from above on this field too (it helped me when reading it). Or if you end up doing what MT says, perhaps as an overarching comment on top of the two `settings = {` lines.

Dillon-Dumesnil (approved):

course_discovery/apps/course_metadata/index.py
rgraber: We have a bunch of upcoming PRs that will affect this file, I'll make a note to add that comment in the next one.

PR #2477, opened 2020-03-17
WS-760 part II, add missing field and fix coverage

rgraber:


albemarle (approved):

course_discovery/apps/course_metadata/index.py
```diff
@@ -37,7 +37,7 @@ class ProductIndex(AlgoliaIndex):
     # Bit of a hack: Override get_queryset to return all wrapped versions of all courses and programs rather than an
     # actual queryset to get around the fact that courses and programs have different fields and therefore cannot be
     # combined in a union of querysets.
-    def get_queryset(self):
+    def get_queryset(self):  # pragma: no cover
```
srwang, line 40: mrao?

srwang (approved):

course_discovery/apps/course_metadata/index.py
rgraber: It means to exclude this from test coverage so we don't get dinged in coverage.py. As this is very much an algolia-specific hack there's no real reason to have tests around it.

jmyatt (approved):
LGTM

PR #2479, opened 2020-03-17
Add reserved keyword linter to CI

estute:
This adds the reserved keyword linter into the CI flow for this app. Once https://github.com/edx/edx-django-release-util/pull/35 lands, I will cut a proper release of edx-django-release-util and rerun `make upgrade` using the pypi hosted version of the code.

There are currently violations for both MYSQL and SNOWFLAKE keywords in this IDA, so I am only running the linter for STITCH keywords. In the future, if we change the field names in question, we can run the linter for those other database systems. I could also add the current violations to the whitelist/override file. I'm open to suggestions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ matrix:
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make clean_static'
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make static'
           - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make quality'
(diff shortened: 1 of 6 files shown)


Dillon-Dumesnil (approved):
Hey Stu, this looks great. Only question is about why we would merge this prior to the release of edx-django-release-util if the plan is to have to update this PR again once that is released. I trust your judgment, just curious.

estute:
@Dillon-Dumesnil this isn't quite done yet. I created it as pull request while still working as it is directly related the the CI flow of the repo. Once edx-django-release-utils has been updated and released, I will make the necessary changes to this PR.

Dillon-Dumesnil:
Ah okay! Sounds good. Sorry for the early review then. We are automatically tagged anytime a PR is opened in course discovery. If you make a draft PR though, it won't alert us so that's an option for WIP. In any case, you'll be free to merge this whenever you're ready

macdiesel (approved):

PR #2481, opened 2020-03-18
Remove non-Active programs and use ProgramType name instead of slug

rgraber:


course_discovery/apps/course_metadata/algolia_proxy_models.py
```diff
@@ -257,7 +257,7 @@ def course_titles(self):
     @property
     def program_types(self):
         if self.type:
-            return [self.type.slug]
+            return [self.type.name]
```
jlajoie, line 260: Didn't look at where all of the algolia stuff is used, but is this surfaced in the API anywhere? I'd consider this a breaking change if you're changing the expected return type from slug to name and it should either be built in a backwards compatible way (basically just add it as a new property), if not the PR is +1'd and is good to go

jlajoie (approved):
Just one comment, you're +1'd on this though depending on the answer to it since this PR looks good

course_discovery/apps/course_metadata/algolia_proxy_models.py
rgraber: It is indeed surfaced in the Algolia response API, but right now website are the only ones who are using this API and we're very much still in the process of finalizing it for our own use. Once more people are using it we'll be more careful about changes but this is effectively still in development.

PR #2483, opened 2020-03-19
Adding default permissions except view to run tests on django2.

awais786:
Testing permissions.
On Django2 tests are not running and giving errors.

Django2 is adding view permission by default and it conflicts with out custom permission.

https://docs.djangoproject.com/en/2.2/ref/models/options/#django.db.models.Options.default_permissions

**javascript_catalog** is removed.


course_discovery/urls.py
```diff
@@ -21,7 +21,7 @@
 from django.conf.urls.static import static
 from django.contrib import admin
 from django.utils.translation import ugettext_lazy as _
-from django.views.i18n import javascript_catalog
+from django.views.i18n import JavaScriptCatalog
```
mikix: This change is already in https://github.com/edx/course-discovery/pull/2476. It's fine to have it here too, but just heads up on possible conflicts.

course_discovery/apps/catalogs/models.py
```diff
@@ -125,3 +125,4 @@ class Meta(TimeStampedModel.Meta):
         permissions = (
             ('view_catalog', 'Can view catalog'),
         )
+        default_permissions = ('add', 'change', 'delete',)
```
mikix, line 128: I'm not super familiar with this aspect of a model. Does this not affect other models in Disco? Or just Catalog, because of how we partition out its permissions to partners?

mikix (approved):

course_discovery/apps/catalogs/models.py
awais786:
```
As per django.
Defaults to ('add', 'change', 'delete', 'view'). You may customize this list, for example, by setting this to an empty list if your app doesn’t require any of the default permissions. It must be specified on the model before the model is created by migrate in order to prevent any omitted permissions from being created.
```
My understanding is `'add', 'change', 'delete'` are already there in django1.11, they included the `view` in django2 which conflicts with our custom view migration. Two view migrations created and gives integrity error. 

To avoid this issue I just removed the view permission from default list other permissions stays there.

I don't think there is any permission effect on discovery. All tests are green. 
@jmbowman your thoughts on it.

awais786: We can hold this PR untill we have travis ready for django2.2 build.

Dillon-Dumesnil:
I just want to call out the similarities between this PR and #2488 so you two (Awais and Ayub) can avoid merge conflicts. Once it is resolved over who will be merging in the changes, I can review either or both PRs.

awais786:
> I just want to call out the similarities between this PR and #2488 so you two (Awais and Ayub) can avoid merge conflicts. Once it is resolved over who will be merging in the changes, I can review either or both PRs.

That PR is not merging it is just for testing purppose.

course_discovery/apps/catalogs/models.py
```diff
@@ -125,3 +125,4 @@ class Meta(TimeStampedModel.Meta):
         permissions = (
             ('view_catalog', 'Can view catalog'),
         )
+        default_permissions = ()
```
awais786: now3 tests will fail due to missing permissions.

PR #2486, opened 2020-03-19
DISCO-1335: Upgrades JS packages

jlajoie:
Summary:

* We don't use gulp anywhere, so that and it's associated gulpfile are now gone
* Webpack needed a major version upgrade
  * This cause all of the plugins to be updated to their latest as well
  * The beta tag is because of this PR, https://github.com/webpack/webpack/issues/6568 where the latest released version of extract-text-webpack-plugin 3.11.0 continues to fail for webpack4

We could get rid of most of the CSS plugins if we do away with templates/demo/query_preview.html 

We could also probably get rid of webpack entirely if we decide to drop support for the swagger pages


package.json
```diff
@@ -8,23 +8,22 @@
   "dependencies": {
     "bootstrap-sass": "^3.3.7",
     "bower": "^1.8.0",
-    "css-loader": "^0.28.4",
+    "css-loader": "^3.4.2",
     "datatables.net": "^1.10.15",
     "datatables.net-bs": "^1.10.15",
-    "extract-text-webpack-plugin": "^3.0.0",
-    "file-loader": "^0.11.2",
+    "extract-text-webpack-plugin": "^4.0.0-beta.0",
```
mikix, line 14: Yeah, beta is good?

mikix (approved):
Those are some big major version bumps.

We don't need gulp? (I don't know much about what it does)

package.json
jlajoie: Get on a call let me tell you the mess this was then I'll summarize here

jlajoie: Summarized in the summary 👍

PR #2489, opened 2020-03-20
Adding tests to make sure permissions exists in db for catalogs.

awais786:
Adding test to check permissions exists in db for catalogs.

This test will be useful in this PR https://github.com/edx/course-discovery/pull/2483

### DIFF
--- a/course_discovery/apps/catalogs/tests/test_models.py
+++ b/course_discovery/apps/catalogs/tests/test_models.py
@@ -1,4 +1,5 @@
 import ddt
+from django.contrib.auth.models import ContentType, Permission
 from django.test import TestCase
 
 from course_discovery.apps.catalogs.models import Catalog
@@ -97,3 +98,9 @@ class CatalogTests(ElasticsearchTestMixin, TestCase):
         with self.assertRaises(TypeError) as context:
             self.catalog.viewers = viewers
         self.assertEqual(context.exception.args[0], 'Viewers must be a non-string iterable containing User objects.')
+
+    @ddt.data('add_catalog', 'change_catalog', 'view_catalog', 'delete_catalog')
+    def test_catalogs_permissions(self, perm):
+        """ Validate that model has the all four permissions. """
+        cont_type = ContentType.objects.get(app_label='catalogs', model='catalog')
+        self.assertTrue(Permission.objects.get(content_type=cont_type, codename=perm))


UsamaSadiq (approved):
It seems fine for the required change.

Ayub-Khan (approved):

course_discovery/apps/catalogs/tests/test_models.py
```diff
@@ -97,3 +98,9 @@ def test_set_viewers_with_invalid_argument(self, viewers):
         with self.assertRaises(TypeError) as context:
             self.catalog.viewers = viewers
         self.assertEqual(context.exception.args[0], 'Viewers must be a non-string iterable containing User objects.')
+
+    @ddt.data('add_catalog', 'change_catalog', 'view_catalog', 'delete_catalog')
```
Dillon-Dumesnil, line 102: In #2483, you explicitly did not include the `view` permission. Why is it included as part of this test and set to return True?

awais786: @Dillon-Dumesnil its already in master branch https://github.com/edx/course-discovery/blob/master/course_discovery/apps/catalogs/models.py#L125

Dillon-Dumesnil: @awais786 So is the goal then that this test will fail for #2483 and when we change the permission and test to not include view, everything will pass again? Or is this set of permissions unrelated to the other set of permissions? Sorry, I'm a bit unclear about all of the catalog permissions that exist.

awais786: @Dillon-Dumesnil You can check this [link](https://docs.djangoproject.com/en/2.2/ref/models/options/#default-permissions).
```
Defaults to ('add', 'change', 'delete', 'view'). You may customize this list, for example, 
by setting this to an empty list if your app doesn’t require any of the default permissions. 
It must be specified on the model before the model is created by migrate in order to 
prevent any omitted permissions from being created.
```
So right now in django1.11 it has 4 permissions. 3 default and 1 custom `view` but in django2 `view` will come also and trigger errors. So I just maintained the `view` as it is and add 3 in default permission set. Thats why i added a test also in separate PR to make sure all pemissions remains there.

Please let me know if there is any confusion and feel free to suggest any improvements.

Dillon-Dumesnil: Ah okay, I see what you're saying now. Thanks for the explanation Awais.

Dillon-Dumesnil (approved):

PR #2496, opened 2020-03-23
DISCO-1607: Add stockrecord with sku into ecom request

Dillon-Dumesnil:
We use the sku in order to get the seat we want to update

### DIFF
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -415,6 +415,7 @@ def serialize_seat_for_ecommerce_api(seat, mode):
         'expires': serialize_datetime(calculated_seat_upgrade_deadline(seat)),
         'price': str(seat.price),
         'product_class': 'Seat',
+        'stockrecords': [{'partner_sku': getattr(seat, 'sku', None)}],
         'attribute_values': [
             {
                 'name': 'certificate_type',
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/utils.py
```diff
@@ -415,6 +415,7 @@ def serialize_seat_for_ecommerce_api(seat, mode):
         'expires': serialize_datetime(calculated_seat_upgrade_deadline(seat)),
         'price': str(seat.price),
         'product_class': 'Seat',
+        'stockrecords': [{'partner_sku': getattr(seat, 'sku', None)}],
```
mikix, line 418: We read the stockrecord back from the response, right? Might be a small race condition if there's an error in parsing the response or saving the stockrecord in disco. Then there's a gap before the next RCM run where we might send None for sku again, right?

mikix (approved):

course_discovery/apps/course_metadata/utils.py
Dillon-Dumesnil: Discussed offline and we think this is a low enough risk to allow this race condition. We believe even if there is a failure, a subsequent run would simply create a new seat for the course in ecommerce and we can remove or retire the seat that failed.

Dillon-Dumesnil:
Related to https://github.com/edx/ecommerce/pull/2849

This PR must land before the ecommerce PR.

PR #2508, opened 2020-03-27
Trying to get tox working

jinder1s:


.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: python
 
-python:
-    - "3.5"
```
jmbowman, line 4: Any reason this needed to be moved into each matrix entry definition?

.travis.yml
```diff
@@ -14,41 +11,50 @@ sudo: required
...
+    - python: 3.5
+      env: TOXENV=django20
+    - python: 3.5
+      env: TOXENV=django21
+    - python: 3.5
+      env: TOXENV=django22
+
+install:
+  - docker exec -t discovery bash -c 'sed -i "s|http://archive|http://us.archive|g" /etc/apt/sources.list' # US mirrors for speed
+  - docker exec -t discovery bash -c 'apt update && apt install -y --no-install-recommends firefox gettext xvfb'
```
jmbowman: Probably a change for a future PR, but there's an easier way to install Firefox on Travis workers.  Example: https://github.com/edx/bok-choy/blob/master/.travis.yml#L20-L21 .  And you shouldn't need xvfb if all the tests have been switched to use headless Firefox.

.travis.yml
```diff
@@ -14,41 +11,50 @@ sudo: required
...
+  - docker exec -t discovery bash -c 'sed -i "s/course_discovery.settings.devstack/course_discovery.settings.test/" /edx/app/discovery/discovery_env'
+  - docker exec -e TOXENV=$TOXENV -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make requirements'
+
+script:
+  - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make clean_static'
+  - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make static'
+  - docker exec -e TOXENV=$TOXENV -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && xvfb-run make test'
```
jmbowman: Again, shouldn't need xvfb-run if the tests are headless.

requirements/docs.txt
```diff
@@ -30,7 +30,7 @@ requests==2.23.0          # via requests-oauthlib, social-auth-core, sphinx
 six==1.14.0               # via cryptography, edx-sphinx-theme, packaging, social-auth-app-django, social-auth-core
 snowballstemmer==2.0.0    # via sphinx
 git+https://github.com/python-social-auth/social-app-django.git@c00d23c2b45c3317bd35b15ad1b959338689cef8#egg=social-auth-app-django  # via -r requirements/github.in
-social-auth-core==3.3.0   # via social-auth-app-django
+social-auth-core==3.3.2   # via social-auth-app-django
```
jmbowman: This should be pinned to 3.2.0 in the constraints file, the upgrade to 3.3.0 is what triggered users to lose superuser status; see [ARCHBOM-1078](https://openedx.atlassian.net/browse/ARCHBOM-1078) for details.

.travis.yml
jinder1s: I put it back, probably was added while debugging travis file earlier.

jlajoie (approved):

PR #2515, opened 2020-03-31
Upgrade Django version to 2.2

mraarif:
relevant JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1435).


requirements/constraints.txt
```diff
@@ -1,15 +1,12 @@
 # Stay on an LTS release
-django<1.12
+django<2.3
 
 # FIXME: these need Django 2
 django-guardian<2
```
mikix: Could maybe drop this constraint too then.

requirements/constraints.txt
```diff
@@ -1,15 +1,12 @@
...
 
 # FIXME: 1.x is an ancient version that we should upgrade off of
 elasticsearch<2
 
-# FIXME: 1.8.5 breaks the way we use serialization methods
-# FIXME: 1.8.3 breaks something about OneToOne field validators
-drf-haystack<1.8.3
 djangorestframework<3.10  # newer versions drop features that old drf-haystack above still uses
```
mikix: If you're dropping drf-haystack constraint, you can drop the djangorestframework one too.

mikix: Though from test failures, seems like the drf-haystack issues are still present -- are you removing the constraint for django 2.2 reasons?

mraarif: yes the `drf-haystack` version 1.8.2 was pinned at Django 2.0

requirements/constraints.txt
mraarif: we might need this constraint because django-gaurdian >=2 drops support for django 1.11, since we still want to be able to test against django 1.11

jmbowman:
@mikix Could you or someone else on your team help Arif with the last failing test here?  We think this is the last blocker to getting discovery ready for deployment on Django 2.2.  There's a little more context in https://openedx.atlassian.net/browse/BOM-1435 .

mikix:
Looking...

mikix:
I don't have a final answer yet and have to sign off for now, but look at CourseRunSerializer.to_internal_value, I bet it's doing something to mess this up.

(And note that Publisher sends `staff` as a list of uuids, so you may not want to edit that test to send dictionaries, but instead keep the ability to send a list of uuid slugs, and then fix whatever is complaining about that.)

mikix:
Ah, hm. Looks like the `fields` property of the serializer, which our hacky little `to_internal_value` relies on, gained the `cached_property` decorator. So our hack is no longer working... Hm.

mikix:
OK, just merged https://github.com/edx/course-discovery/pull/2525 -- try rebasing and dropping your changes to test_partial_update_staff.

jmbowman (approved):
I don't have the background context on the serializer changes, but otherwise looks fine.

jmbowman:
@mraarif Looks like this needs a rebase now. That should also get the correct Travis build result, we just switched the repo from travis-ci.org to travis-ci.com for better build reporting reliability.

@mikix This should be otherwise ready for final review.

course_discovery/apps/api/serializers.py
```diff
@@ -87,6 +88,24 @@
 }
 
 
+def get_default_field_kwargs(model, field):
```
mikix, line 96: Is this to keep existing behavior in the face of drf-haystack or drf changing some defaults? Might be worth a comment to that effect here. Otherwise, it's not clear what this is trying to achieve (to me anyway).

course_discovery/settings/base.py
```diff
@@ -400,8 +400,10 @@
     'DEFAULT_THROTTLE_RATES': {
         'user': '100/hour',
     },
+    'DEFAULT_SCHEMA_CLASS': 'rest_framework.schemas.coreapi.AutoSchema'
```
mikix, line 403: Did this use to be the default and they changed it on us?

mikix (approved):
I also don't really understand the serializer changes. But I trust that they are sensible.

Excited for django 2.2!

course_discovery/settings/base.py
jmbowman: Yes: https://www.django-rest-framework.org/coreapi/#legacy-coreapi-schemas-docs

mraarif: it was required after DRF 3.10 [release](https://www.django-rest-framework.org/community/3.10-announcement/#continuing-to-use-coreapi)

course_discovery/apps/api/serializers.py
mraarif: comment added.

mikix (approved):

mikix:
Looks like the get_default_field_kwargs changes aren't getting covered in tests.

mikix (approved):

PR #2517, opened 2020-04-01
Exclude to-be decommisioned partners from RCM

Dillon-Dumesnil:
We are excluding Harvard Medical School Global Academy and MITProfessionalX as part of this work.


jmyatt (approved):

feanil:
@georgebabey does it make sense to also just start excluding xpro right now?

jmyatt (approved):

georgebabey (approved):

PR #2519, opened 2020-04-03
Fixing RCM

Dillon-Dumesnil:
RCM has been failing because it has not had access to all of the
courses because the service user was being treated as an
AnonymousUser instead of a staff user. By passing in the username,
it will be treated as staff and get all courses.


mikix (approved):
Beautiful

PR #2520, opened 2020-04-06
Upgrade the django-app-conf to use latest version

mraarif:
Upgrade `django-app-conf` to `1.0.4` that has added. support for Django 2.2.
Relevant JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1232).

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -6,7 +6,7 @@
 #
 alabaster==0.7.12         # via sphinx
 babel==2.8.0              # via sphinx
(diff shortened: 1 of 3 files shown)


awais786 (approved):

jmbowman (approved):
Looks like it was actually upgraded a few days ago in https://github.com/edx/course-discovery/pull/2519, but these upgrades are useful also.

mraarif:
ah!, I didn't notice after rebase.

PR #2523, opened 2020-04-06
Catalog courses query improvement

cpennington:


course_discovery/apps/course_metadata/query.py
```diff
@@ -50,14 +50,14 @@ def available(self):
...
-        query = self.filter(enrollable & not_ended & marketable)
+        ids = self.filter(enrollable & not_ended & marketable).values('id')
 
         # By itself, the query above performs a join across several tables and would return
         # a copy of the same course multiple times (a separate copy for each available
         # seat in each available run).
         # We use distinct() to make sure it only returns a single copy of each available
         # course.
-        return query.distinct()
+        return self.filter(id__in=ids)
```
mikix, line 60: Should fix the comment above too.

Is distinct so bad, that a subquery is more efficient? Hmm, I see that passing in fields to distinct only works on postgres?

mikix (approved):
Seems good!

course_discovery/apps/course_metadata/query.py
cpennington: I think it just has to do a bunch more work to check the distincts over all the fields vs doing the subquery (which often isn't much different than a join).

PR #2527, opened 2020-04-08
Default to ordering Person by id, because that has an index, unlike c…

cpennington:
…reated


xnick421 (commented):
Is the migration supposed to be part of this pr?

cpennington:
There's a few changes from before whose migrations got missed. This change does require a migration as well. I could split them up, though

Dillon-Dumesnil (approved):
Looks good. Jeff and Becca talked about the other parts of the migration already, so those are fine to keep as well. I know Stuart Young has a PR that was also picking these other migrations up so maybe just talk to him to avoid any conflicts (I'm not sure if anything breaks if you have the same operation in two separate migrations).

Dillon-Dumesnil:
Doesn't look like Stu has an open PR so it might be good to just shoot him a message that he doesn't need to worry about those after a rebase and you're good to merge.

PR #2530, opened 2020-04-10
drop foreign keys on course_metadata.historicalcourserun

estute:
This pull request represents the first step in the plan to rename the start/end fields in course_metadata_courserun (runbook: https://openedx.atlassian.net/wiki/spaces/DE/pages/1334706563/DENG-20+rename+start+column+in+Discovery+course+metadata+courserun+table+-+alternate+version)\

In order to move forward with a gh-ost migration on the course_metadata_historicalcourserun table, we need to (temporarily) remove any foreign keys. Luckily, this table only has 1:

```
mysql> SELECT TABLE_NAME, COLUMN_NAME, CONSTRAINT_NAME, REFERENCED_TABLE_NAME, REFERENCED_COLUMN_NAME FROM INFORMATION_SCHEMA. KEY_COLUMN_USAGE WHERE REFERENCED_TABLE_SCHEMA = 'discovery' AND TABLE_NAME = 'course_metadata_historicalcourserun';
+-------------------------------------+-----------------+------------------------------------------------------------+-----------------------+------------------------+
| TABLE_NAME                          | COLUMN_NAME     | CONSTRAINT_NAME                                            | REFERENCED_TABLE_NAME | REFERENCED_COLUMN_NAME |
+-------------------------------------+-----------------+------------------------------------------------------------+-----------------------+------------------------+
| course_metadata_historicalcourserun | history_user_id | course_metadata_hist_history_user_id_03c38d19_fk_core_user | core_user             | id                     |
+-------------------------------------+-----------------+------------------------------------------------------------+-----------------------+------------------------+
1 row in set (0.00 sec)

mysql> SELECT TABLE_NAME, COLUMN_NAME, CONSTRAINT_NAME, REFERENCED_TABLE_NAME, REFERENCED_COLUMN_NAME FROM INFORMATION_SCHEMA. KEY_COLUMN_USAGE WHERE REFERENCED_TABLE_SCHEMA = 'wwdiscovery' AND TABLE_NAME = 'course_metadata_historicalcourserun';
Empty set (0.00 sec)
```

Generated SQL:
```
BEGIN;
--
-- Alter field history_user on historicalcourserun
--
ALTER TABLE `course_metadata_historicalcourserun` DROP FOREIGN KEY `course_metadata_hist_history_user_id_03c38d19_fk_core_user`;
COMMIT;
```


mikix:
I mean, I guess in theory, do whatever you guys need.

But... like, isnt foreign key relationships useful parts of a history? In this case, looks like the user that caused a change to the table. Seems bad to wipe that?

pwnage101:
> I mean, I guess in theory, do whatever you guys need.
> 
> But... like, isnt foreign key relationships useful parts of a history? In this case, looks like the user that caused a change to the table. Seems bad to wipe that?

This field is typically filled in by an handler provided by django-simple-history that we never really implement (I've never seen us implement it).  If this was a table that we edit via django-admin, sure it would contain useful data about who clicked the save button in the interface (IIRC this is the default behavior when the handler is unimplemented).  However my guess is that every row of this table just has this field set to the user ID of the automation user that runs in jenkins.

@estute might be worth confirming this.

mikix:
`select count(distinct u.email) from course_metadata_historicalcourserun cr join core_user u on cr.history_user_id = u.id;`

That gives 295 different users, and if you drop the count() bit, they seem like real people?

Dillon-Dumesnil:
Not to be intrusive, but I have at least used the user id on the history table in the past to look who made a particular change. I think it ended up being the PC for that org, so we could have gotten there anyway, but it was still helpful having it recorded.

bmedx:
FWIW this isn't removing the field or data in it, only the MySQL-side constraint that the user id exist in the users table. If we aren't seeing errors about FK violation in Splunk this should have no effect on what's stored in the database. If we are, we should probably take a look at that. :)

Dillon-Dumesnil:
> FWIW this isn't removing the field or data in it, only the MySQL-side constraint that the user id exist in the users table. If we aren't seeing errors about FK violation in Splunk this should have no effect on what's stored in the database. If we are, we should probably take a look at that. :)

Ah my bad! I'm just following along with the comments and haven't done a thorough look at the PR. 👍

mikix:
Oh, OK. I also missed that from the description `we need to (temporarily) remove any foreign keys` -- missing a word constraint I guess.

But temporarily dropping the constraint seems fine.

mikix (approved):

estute:
Sorry about any confusion with the wording on that message. We will put the db constraints back in place after this work has completed.

PR #2533, opened 2020-04-10
WIP - DO NOT MERGE - adding start_date and end_date fields to CourseRun

estute:
This pull request represents the second step in the plan to rename the start/end fields in course_metadata_courserun (runbook: https://openedx.atlassian.net/wiki/spaces/DE/pages/1334706563/DENG-20+rename+start+column+in+Discovery+course+metadata+courserun+table+-+alternate+version)\

I am adding the two fields to the model, and have split the resulting migrations into two separate migration files, for the regular and historical tables, as they should be applied in different steps. There were some other uncaught migrations that I picked up prior to this work, so I added them as well.

As stated in the doc, this should not be merged and deployed to production until the first ghost migrations have completed (to add the `x_date` columns to the history table).


estute:
We are no longer moving forward with this work.

PR #2537, opened 2020-04-13
Only update courseruns when they change

cpennington:


Dillon-Dumesnil (approved):
LGTM. Thanks for writing this Cale!
I think us having this `_update_instance` is making this a lot easier (thanks @mikix )

mikix (approved):
Nice.

PR #2539, opened 2020-04-13
WS-875 don't index internal/hidden courses to algolia

rgraber:


albemarle (approved):

Dillon-Dumesnil (approved):

jlajoie (approved):
Looks good, two things that might be worth considering but not blockers.

* Could be useful as the `should_index` condition continues to grow that there's a docstring or comment explaining the logic as well.
* Might be worth testing that some course_runs being hidden as long as they aren't the advertised_course_run doesn't impact the indexing

PR #2543, opened 2020-04-14
ENT-2724 | removed get_cache_key utility and using it from edx-django-utils

HammadAhmadWaqas:


course_discovery/apps/core/api_client/lms.py
```diff
@@ -4,12 +4,11 @@
 import logging
 
 from django.core.cache import cache
+from edx_django_utils.cache.utils import get_cache_key
 from edx_rest_api_client.client import EdxRestApiClient
 from edx_rest_api_client.exceptions import SlumberBaseException
 from requests.exceptions import ConnectionError, Timeout  # pylint: disable=redefined-builtin
 
-from course_discovery.apps.api.utils import get_cache_key
```
mikix, line 11: So this swap will have the effect of wiping out our cache. (The external version uses `sorted()` on its kwargs and we did not. So it will generate different keys for the same args.)

Are we worried about that? I guess it will generate heavier load on the LMS. I don't have my IDE open right now to see where all this class is used.

mikix: (To be clear, this is a good change and we should make it, just want to make sure we understand impact first.)

HammadAhmadWaqas: @mikix Thank you for raising concern.
we have only one use in this repo(course-discovery) [here](https://github.com/edx/course-discovery/blob/5d2cbed3a96a72185f9ed86c9aaf17f595fa4ff6/course_discovery/apps/core/api_client/lms.py#L59)

Yes, you are right. This change may wipe out our cache for once. but this change is made to resolve a bug reported in [ENT-2724](https://openedx.atlassian.net/browse/ENT-2724) reported by @georgebabey

mikix (approved):
I see. The `sorted` _is_ the fix. :) Alright, cache wipe can't be helped.

ziafazal (approved):
👍

PR #2548, opened 2020-04-16
BOM-1519

awais786:
https://openedx.atlassian.net/browse/BOM-1519
Removing django<2.2 versions from tox/travis.
Adding BOM-1533 python3.8 requirement also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
 
 matrix:
   include:
-    - env: TOXENV=django111
+    - env: TOXENV=django22
     - env: COMMAND=test:quality
       install:
         - docker exec -t discovery bash -c 'sed -i "s/course_discovery.settings.devstack/course_discovery.settings.test/" /edx/app/discovery/discovery_env'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -33,9 +35,11 @@ matrix:
         - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make static'
         - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make quality'
         - docker exec -t discovery bash -c 'source /edx/app/discovery/discovery_env && cd /edx/app/discovery/discovery/ && make check_keywords'
-    - env: TOXENV=django20
-    - env: TOXENV=django21
-    - env: TOXENV=django22
+    - python: 3.5
+      env: TOXENV=django22
```
mikix: This is a duplicate, right? You have same config above.

.travis.yml
```diff
@@ -21,7 +21,9 @@ before_install:
 
 matrix:
   include:
-    - env: TOXENV=django111
+    - python: 3.5
```
mikix: You don't need to specify python here, there's a global python version defined, right?

.travis.yml
awais786: Yes, thanks for catching this.

.travis.yml
awais786: I added just for readability. I have removed it.

mikix (approved):

PR #2549, opened 2020-04-16
BOM-1535

awais786:
https://openedx.atlassian.net/browse/BOM-1535

Upgrading django-guardian after removing this pin.
```
django-guardian>=2 drops support for Django 1.11	
```

**NOTE:  It brings new migration also from guardian package.**
Applying [guardian.0002_generic_permissions_index](https://github.com/django-guardian/django-guardian/blob/devel/guardian/migrations/0002_generic_permissions_index.py)


mikix (approved):

requirements/pip_tools.txt
```diff
@@ -5,5 +5,8 @@
 #    pip-compile --output-file=requirements/pip_tools.txt requirements/pip_tools.in
 #
 click==7.1.1              # via pip-tools
-pip-tools==4.5.1          # via -r requirements/pip_tools.in
+pip-tools==5.0.0          # via -r requirements/pip_tools.in
```
jmbowman: Note that this upgrade requires pip 20+, and we're still installing 19.3.1 in all our containers and VMs.  Probably should constraint this until we update the configuration repo to start installing a newer pip version by default.

jmbowman (commented):
Looks good other than the one noted issue.

requirements/pip_tools.txt
awais786: constraint added.

PR #2552, opened 2020-04-21
Update devstack settings JWT_AUTH to enable inter-service communicati…

iloveagent57:
…on during local development.  The problem is that the `devstack` setting inherits the useless JWT_ISSUER defaults from `base`, and edx-drf-extensions (or something) fails hard on a mismatched JWT_ISSUER now (even if there is a good entry in JWT_ISSUERS (plural)).


mikix (approved):
Sure

Dillon-Dumesnil (approved):

PR #2559, opened 2020-04-23
DISCO-1617 admin autocomplete

xnick421:
- replace autocomplete.ModelSelect2 with new autocomplete_fields
  property newly available in django2.


Dillon-Dumesnil (approved):
Assuming you tested the UI and functionality between the old and new admin pages, this looks great. It was a little weird seeing `list_display` and `search_fields` removed so just want to make sure everything is the same.

mikix (approved):

PR #2563, opened 2020-04-27
WS-925 make program type translatable: add new model/field

rgraber:


course_discovery/apps/course_metadata/migrations/0242_auto_20200427_1343.py
```diff
@@ -0,0 +1,37 @@
...
+            name='ProgramTypeTranslation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('language_code', models.CharField(db_index=True, max_length=15, verbose_name='Language')),
+                ('name_t', models.CharField(max_length=32, unique=True)),
+                ('master', parler.fields.TranslationsForeignKey(editable=False, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='translations', to='course_metadata.ProgramType')),
+            ],
+            options={
+                'unique_together': {('language_code', 'master')},
+                'verbose_name': 'program type Translation',
```
jlajoie: nit: might just want to fix the casing here to be consistent, Program Type Translation or program type translation

course_discovery/apps/course_metadata/migrations/0242_auto_20200427_1343.py
```diff
@@ -0,0 +1,37 @@
...
+                'default_permissions': (),
+                'db_table': 'course_metadata_programtype_translation',
+                'db_tablespace': '',
+                'managed': True,
```
jlajoie: Not sure about what these options are for. Were these manually input or did they come from the migration? I did a quick search and didn't find any references for `db_tablespace` elsewhere in course-discovery.

course_discovery/apps/course_metadata/models.py
```diff
@@ -295,14 +295,15 @@ def __str__(self):
...
-class ProgramType(TimeStampedModel):
+class ProgramType(TranslatableModel, TimeStampedModel):
     XSERIES = 'xseries'
     MICROMASTERS = 'micromasters'
     PROFESSIONAL_CERTIFICATE = 'professional-certificate'
     PROFESSIONAL_PROGRAM_WL = 'professional-program-wl'
     MASTERS = 'masters'
 
     name = models.CharField(max_length=32, unique=True, null=False, blank=False)
+    translations = TranslatedFields(name_t=models.CharField(max_length=32, unique=True, null=False, blank=False))
```
jlajoie: So I've never used TranslatedFields before, but is this what creates the above migration? Out of curiosity, if we wanted to translate a second field, would we just pass in a second kwarg to this field?

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -15,11 +15,12 @@
     AlgoliaProxyCourse, AlgoliaProxyProduct, AlgoliaProxyProgram
 )
 from course_discovery.apps.course_metadata.choices import CourseRunStatus
+# pylint: disable=no-name-in-module
```
jlajoie: Question about this, is this because we don't have an actual `ProgramTypeTranslation` class? Does this mean we'll need to add this entry everywhere we want to use and access translations for ProgramTypes?

jlajoie (changes requested):
Just a couple of questions since I'm not fully informed on translations, but depending on the answers this will probably be good to go.

course_discovery/apps/course_metadata/migrations/0242_auto_20200427_1343.py
rgraber: Came from running makemigrations.

course_discovery/apps/course_metadata/models.py
rgraber: Yes, this is what created the migration. I think we would indeed just pass kwargs for all fields we wanted translated

course_discovery/apps/course_metadata/tests/test_signals.py
rgraber: Eventually there will be a ProgramTypeTranslation class once everything is properly moved around. This is just a stopgap.

rgraber: On further reflection it will be better to create it here.

jlajoie (approved):

PR #2564, opened 2020-04-27
WS-926 copy programType.name to programTypeTranslatable.name_t

rgraber:


course_discovery/apps/course_metadata/migrations/0244_migrate_programtype_name_to_translatable_model.py
```diff
@@ -0,0 +1,50 @@
...
+from django.db import migrations
+
+logger = logging.getLogger(__name__)
+
+
+def forwards_func(apps, schema_editor):
+    ProgramType = apps.get_model('course_metadata', 'ProgramType')
+    ProgramTypeTranslation = apps.get_model('course_metadata', 'ProgramTypeTranslation')
+
+    for programType in ProgramType.objects.all():
```
jlajoie: nit: snake_case over camelCase for python

course_discovery/apps/course_metadata/migrations/0244_migrate_programtype_name_to_translatable_model.py
```diff
@@ -0,0 +1,50 @@
...
+            language_code=settings.PARLER_DEFAULT_LANGUAGE_CODE,
+            name_t=programType.name,
+        )
+
+
+def backwards_func(apps, schema_editor):
+    ProgramType = apps.get_model('course_metadata', 'ProgramType')
+    ProgramTypeTranslation = apps.get_model('course_metadata', 'ProgramTypeTranslation')
+
+    for programType in ProgramType.objects.all():
```
jlajoie: same as above nit

course_discovery/apps/course_metadata/migrations/0244_migrate_programtype_name_to_translatable_model.py
```diff
@@ -0,0 +1,50 @@
...
+        try:
+            translation = ProgramTypeTranslation.objects.get(master_id=programType.pk, language_code=settings.LANGUAGE_CODE)
+            programType.name = translation.name_t
+            programType.save()  # Note this only calls Model.save()
+        except ObjectDoesNotExist:
+            # nothing to migrate
```
jlajoie, line 38: This comment threw me off here, nothing to migrate sounds benign but is then followed by a log.exception, might be worth just removing or clarifying what the intended course of action is. Looking at this I'd be surprised if this is ever hit as the assumption would be you roll forwards, then roll back if things don't happen, but I guess some one in that period could delete an entry in django admin. Definitely agree it's worth logging, but maybe not at the exception level as what could you really do about this?

jlajoie (approved):
Will +1 it to not block you later, but I left a couple of comments that I think are worthwhile to address

PR #2565, opened 2020-04-27
WS-927 remove references to ProgramType name

rgraber:


course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -120,10 +120,10 @@ def test_retrieve_basic_curriculum(self, django_assert_num_queries):
         self.create_curriculum(program)
 
         # Notes on this query count:
-        # 36 queries to get program without a curriculum and no courses
+        # 37 queries to get program without a curriculum and no courses
         # +2 for curriculum details (related courses, related programs)
         # +8 for course details on 1 or more courses across all sibling curricula
-        with django_assert_num_queries(46):
+        with django_assert_num_queries(47):
```
jlajoie, line 126: Quick question about this since I didn't see any changes around JOINs, but if we had 2 programs, does this query count go up to 48? Just want to avoid n+1 queries and making sure we have the proper select_relateds set up

course_discovery/apps/course_metadata/models.py
```diff
@@ -365,7 +365,7 @@ def get_program_type_data(pub_course_run, program_model):
 class ProgramTypeTranslation(TranslatedFieldsModel):  # pylint: disable=model-no-explicit-unicode
     master = models.ForeignKey(ProgramType, models.CASCADE, related_name='translations', null=True)
 
-    name_t = models.CharField(max_length=32, blank=False, null=False)
+    name_t = models.CharField("name", max_length=32, blank=False, null=False)
```
jlajoie, line 368: Question on what "name" here is, is that just the db_column? Does this need to be a kwarg?

course_discovery/apps/course_metadata/models.py
```diff
@@ -330,7 +330,7 @@ class ProgramType(TranslatableModel, TimeStampedModel):
         },
         help_text=_('Please provide an image file with transparent background'),
     )
-    slug = AutoSlugField(populate_from='name', editable=True, unique=True, slugify_function=uslugify,
+    slug = AutoSlugField(populate_from='name_t', editable=True, unique=True, slugify_function=uslugify,
```
jlajoie, line 333: Do you know what the order of operations is, if this will only work for new entries, or if it will backpopulate (and in doing so, cause IntegrityErrors since this field is unique)

jlajoie (approved):
Looks good, couple of questions but depending on answers this should be good to go 👍

course_discovery/apps/course_metadata/models.py
rgraber: The only thing I think it determines is the label in Admin but I will double check on that.

course_discovery/apps/course_metadata/models.py
rgraber: This happens on a pre-save hook and only if the field is otherwise blank, so it should be fine.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
rgraber: This particular test just tests getting a single program. The number of queries didn't go up in the tests that get multiple programs so I think we're ok

course_discovery/apps/course_metadata/models.py
rgraber: Yes, the first argument to any field is the verbose name, so just for display and admin labeling.

PR #2571, opened 2020-04-29
ENT-2663 Enterprise API: Include additional detail-level fields in Enterprise Catalog Courses List

zamanafzal:
Enterprise API: Include additional detail-level fields in Enterprise Catalog "Courses" list endpoint
The fields are: 
course detail -> outcome
course detail -> modifed
course detail -> level_type
course run detail -> staff
course run detail -> content_language

**Testing Scenario Before PR**:
1-Create an Enterprise Catalog.
2- Copy the Catalog_UUID and hit the URL
`http://localhost:18000/enterprise/api/v1/enterprise_catalogs/Catalog_UUID/?detail_fields=True
`
You will not observe the above-mentioned detail-level fields in the response.

**After PR:**
Hit the URL again
`http://localhost:18000/enterprise/api/v1/enterprise_catalogs/Catalog_UUID/?detail_fields=True
`
You will observe the above-mentioned detail-level fields in the response.

**Then**
Remove the `detail_fields` `query_param` from the request and detail-level fields will not be in the response.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1931,34 +1931,58 @@ class CourseSearchSerializer(HaystackSerializer):
     course_runs = serializers.SerializerMethodField()
     seat_types = serializers.SerializerMethodField()
 
+    def __init__(self, *args, **kwargs):
+        super(CourseSearchSerializer, self).__init__(*args, **kwargs)
+        request = self.context['request']
+        detail_fields = request.GET.get("detail_fields")
+        # if detail_fields query_param not in request than do not add the following fields in serializer response.
+        if not detail_fields:
+            self.fields.pop('level_type')
+            self.fields.pop('modified')
+            self.fields.pop('outcome')
+
     @staticmethod
     def _get_default_field_kwargs(model, field):
         return get_default_field_kwargs(model, field)
+
+    @staticmethod
+    def course_run_detail(request, detail_fields, course_run):
+        course_run_detail = {
+            'key': course_run.key,
+            'enrollment_start': course_run.enrollment_start,
+            'enrollment_end': course_run.enrollment_end,
+            'go_live_date': course_run.go_live_date,
+            'start': course_run.start,
+            'end': course_run.end,
+            'modified': course_run.modified,
+            'availability': course_run.availability,
+            'pacing_type': course_run.pacing_type,
+            'enrollment_mode': course_run.type_legacy,
+            'min_effort': course_run.min_effort,
+            'max_effort': course_run.max_effort,
+            'weeks_to_complete': course_run.weeks_to_complete,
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1926,34 +1926,59 @@ class CourseSearchSerializer(HaystackSerializer):
...
+            'end': course_run.end,
+            'modified': course_run.modified,
+            'availability': course_run.availability,
+            'pacing_type': course_run.pacing_type,
+            'enrollment_mode': course_run.type_legacy,
+            'min_effort': course_run.min_effort,
+            'max_effort': course_run.max_effort,
+            'weeks_to_complete': course_run.weeks_to_complete,
+            'estimated_hours': get_course_run_estimated_hours(course_run),
+            'content_language': course_run.language.code if course_run.language else None,
```
jmyatt: I don't think you want to add this here, since it's one of the new fields you're adding conditionally based on the presence of `detail_fields`

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -138,6 +139,9 @@ def prepare_sponsoring_organizations(self, obj):
     def prepare_level_type(self, obj):
         return obj.level_type.name if obj.level_type else None
 
+    def prepare_outcome(self, obj):
+        return obj.outcome if obj.outcome else None
```
jmyatt: simply `return obj.outcome` would behave the same way, no?  But also, it looks like you can skip this function altogether and just add a `model_attr` to line 120 where you define the `outcome` field

course_discovery/apps/course_metadata/search_indexes.py
```diff
@@ -117,6 +117,7 @@ class BaseCourseIndex(OrganizationsMixin, BaseIndex):
     logo_image_urls = indexes.MultiValueField()
     sponsoring_organizations = indexes.MultiValueField(faceted=True)
     level_type = indexes.CharField(null=True, faceted=True)
+    outcome = indexes.CharField(null=True, faceted=True)
```
jmyatt: I don't think you want `faceted=True` here since this is a long form text field (used for search and display but not faceting, like short_description and long_description)

jmyatt (changes requested):

course_discovery/apps/course_metadata/search_indexes.py
zamanafzal: yes, you are right. let me add the `model_attr`.

zamanafzal:
@jmyatt Thank you for the feedback. I have incorporated it. Can you please spare some time and take a look again. Thanks

jmyatt (approved):
LGTM

PR #2575, opened 2020-04-30
WS-942 replace levelType.name with levelTypeTranslation.name_t

rgraber:
Replace all uses of LevelType.name with LevelType.name_t. Prefetching didn't seem to make a difference in query counts so I left it out. Also includes a new type of field to allow us to set the level on a course when all we get from the serializer is a translated name.


course_discovery/apps/api/fields.py
```diff
@@ -65,6 +67,19 @@ def to_internal_value(self, data):
...
+    """ Use in place of SlugRelatedField when the slug field is a TranslatedField """
+
+    def to_internal_value(self, data):
+        full_translated_field_name = 'translations__{slug_field}'.format(slug_field=self.slug_field)
+        try:
+            return self.get_queryset().get(**{full_translated_field_name: data})
+        except ObjectDoesNotExist:
+            self.fail('does_not_exist', slug_name=self.slug_field, value=smart_text(data))
+        except (TypeError, ValueError):
+            self.fail('invalid')
```
jlajoie: Haven't seen self.fail before, but if this swallows the exception it might be worth clarifying what the error actually is instead of just outputting invalid.

course_discovery/apps/api/fields.py
```diff
@@ -65,6 +67,19 @@ def to_internal_value(self, data):
...
 
+class SlugRelatedTranslatableField(serializers.SlugRelatedField):
+    """ Use in place of SlugRelatedField when the slug field is a TranslatedField """
+
+    def to_internal_value(self, data):
+        full_translated_field_name = 'translations__{slug_field}'.format(slug_field=self.slug_field)
+        try:
+            return self.get_queryset().get(**{full_translated_field_name: data})
+        except ObjectDoesNotExist:
+            self.fail('does_not_exist', slug_name=self.slug_field, value=smart_text(data))
```
jlajoie: Might be worth just surfacing the exception here if we're going to fail anyways? Same as below

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -359,6 +359,7 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
         if url_slug:
             course.set_active_url_slug(url_slug)
 
+
```
jlajoie: Might just want to get rid of this newline here since nothing in this file is changed otherwise

jlajoie (approved):
Just had some questions around self.fail() and why we chose to go that route. If there's a good reason I'm fine with it. Other than that it looks good to go, so I'll +1 you.

course_discovery/apps/api/fields.py
rgraber: Full disclosure this was copy-pasted from the SlugRelatedField code

PR #2577, opened 2020-05-01
Encode bytestring to concat with cached object in log output

jmyatt:


course_discovery/apps/api/cache.py
```diff
@@ -83,7 +83,7 @@ def process_cache_response(self, view_instance, view_method, request, args, kwar
                 truncated_response_triple = response_triple
             else:
                 truncated_response_triple = (
-                    response_triple[0][:100] + "...",
+                    response_triple[0][:100] + "...".encode('ascii'),
```
mikix, line 86: Probably safer to do `utf8`, but this might be sufficient.

mikix (approved):

jlajoie (approved):
I was under the impression that python's default encoding was ascii, so I'm surprised this is necessary, but 👍

mikix:
> I was under the impression that python's default encoding was ascii, so I'm surprised this is necessary

I believe it's s a matter of not auto-converting bytes to strings in the first place. So we have to do it manually.

PR #2580, opened 2020-05-04
WS-873 create Spanish index

rgraber:


course_discovery/apps/course_metadata/algolia_proxy_models.py
```diff
@@ -10,10 +10,15 @@
 # Utility methods used by both courses and programs
 def get_active_language(course):
     if course.advertised_course_run and course.advertised_course_run.language:
-        return course.advertised_course_run.language.name
+        return course.advertised_course_run.language.macro_language
     return None
 
 
+def get_and_activate_index_language(product):
```
jlajoie: nit: might be worth renaming this, `get_` to me implies it'll return a value

course_discovery/apps/course_metadata/index.py
```diff
@@ -43,3 +41,80 @@ def get_queryset(self):  # pragma: no cover
...
+                      'is_prof_cert_program')
+    result_fields = (('product_marketing_url', 'marketing_url'), ('product_card_image_url', 'card_image_url'),
+                     ('product_uuid', 'uuid'), 'active_run_key', 'active_run_start', 'active_run_type', 'owners',
+                     'course_titles')
+    # Algolia needs this
+    object_id_field = (('custom_object_id', 'objectID'), )
+    fields = search_fields + facet_fields + ranking_fields + result_fields + object_id_field
+    settings = {
+        'searchableAttributes': [
+            'unordered(title)',  # AG best practice: position of the search term in plain text fields doesn't matter
```
jlajoie: AG is Algolia?

course_discovery/apps/course_metadata/index.py
```diff
@@ -43,3 +41,80 @@ def get_queryset(self):  # pragma: no cover
...
+        'attributesForFaceting': ['partner', 'availability', 'subject', 'level', 'language', 'product', 'program_type'],
+        'customRanking': ['asc(availability_rank)', 'asc(is_prof_cert_program)', 'desc(recent_enrollment_count)']
+    }
+    index_name = 'spanish_product'
+    should_index = 'should_index'
+
+    # Bit of a hack: Override get_queryset to return all wrapped versions of all courses and programs rather than an
+    # actual queryset to get around the fact that courses and programs have different fields and therefore cannot be
+    # combined in a union of querysets.
+    def get_queryset(self):  # pragma: no cover
```
jlajoie, line 75: Like the comment, but confused as to the why. I'm not seeing this called anywhere, but if it's still being used as a queryset I'd expect to be able to chain .filter() calls onto it, is that still the case here?

course_discovery/apps/course_metadata/index.py
```diff
@@ -43,3 +41,80 @@ def get_queryset(self):  # pragma: no cover
...
+
+    def update_obj_index(self, instance):
+        for indexer in self.model_index:
+            indexer.update_obj_index(instance)
+
+    def delete_obj_index(self, instance):
+        for indexer in self.model_index:
+            indexer.delete_obj_index(instance)
+
+    def raw_search(self, query='', params={}):
```
jlajoie, line 101: Just double checking if there's any escaping that needs to be done here or if indexer.raw_search() covers that

course_discovery/apps/course_metadata/algolia_proxy_models.py
```diff
@@ -263,8 +272,9 @@ def course_titles(self):
 
     @property
     def program_types(self):
```
jlajoie, line 274: I'm surprised program_types returns an array of exactly length 1. Feels more like "type" but I don't expect this to be changed here

course_discovery/apps/course_metadata/index.py
```diff
@@ -43,3 +41,80 @@ def get_queryset(self):  # pragma: no cover
...
+                     'secondary_description', 'tertiary_description')
+    facet_fields = (('availability_level', 'availability'), ('subject_names', 'subject'), ('levels', 'level'),
+                    ('active_languages', 'language'), ('product_type', 'product'), ('program_types', 'program_type'))
+    ranking_fields = ('availability_rank', ('product_recent_enrollment_count', 'recent_enrollment_count'),
+                      # TEMPORARY: Demote Professional Certificate programs until promoted programs are in place
+                      'is_prof_cert_program')
+    result_fields = (('product_marketing_url', 'marketing_url'), ('product_card_image_url', 'card_image_url'),
+                     ('product_uuid', 'uuid'), 'active_run_key', 'active_run_start', 'active_run_type', 'owners',
+                     'course_titles')
+    # Algolia needs this
```
jlajoie: nit: Might be worth clarifying here what it needs and why? Something like "AlgoliaIndex requires setting object_id_field"

jlajoie (approved):
Couple of nits, only real question is around the get_queryset(). I appreciate the comment there but it's hard for me to tell if that's something we should keep doing since I'm missing context.

course_discovery/apps/course_metadata/index.py
rgraber: updated the comment. Algolia treats this purely as an iterable.

course_discovery/apps/course_metadata/index.py
rgraber: indexer should handle that

course_discovery/apps/course_metadata/algolia_proxy_models.py
rgraber: a course can have multiple program types, and we need the fields to have the same name across courses and programs to make the combined index

course_discovery/apps/course_metadata/index.py
rgraber: yes. updating to spell out

course_discovery/apps/course_metadata/index.py
rgraber: updated comment

PR #2581, opened 2020-05-04
Log data used to generate ProgramListView keys

cpennington:


course_discovery/apps/api/cache.py
```diff
@@ -78,23 +79,27 @@ def process_cache_response(self, view_instance, view_method, request, args, kwar
             kwargs=kwargs
         )
         response_triple = self.cache.get(key)
+        logger.info(view_instance.__class__.__name__)
```
mikix: Is this supposed to be here?

mikix (approved):
You're really digging into this section of code, eh? :)

course_discovery/apps/api/cache.py
cpennington: Nope.

cpennington:
I mean, something is definitely bizarre around this code, to be triggering as many cache misses as we're seeing.

mikix:
> I mean, something is definitely bizarre around this code, to be triggering as many cache misses as we're seeing.

Yeah for sure, wasn't suggesting focus is bad. Just this investigation has you very in the weeds. You'll be able to finish this ticket with an edX degree in `cache.py`

PR #2582, opened 2020-05-04
Revert "WS-942 replace levelType.name with levelTypeTranslation.name_t"

rgraber:
Reverts edx/course-discovery#2575


jlajoie (approved):
👍 for revert

PR #2590, opened 2020-05-05
add is_enrollable field to search/all endpoint for course and courseruns

adamstankiewicz:
The `is_enrollable` field is already included in the list/detail response for course runs. However, the `is_enrollable` field is not present on the "/search/all" endpoint.

For Enterprise, we'd like to avoid duplicating logic that already exists in course-discovery. That said, we'd like to expose the `is_enrollable` field in the response for the "/search/all" endpoint. This will effectively eliminate our need to duplicate the logic that already exists to compute `is_enrollable`.


bbaker6225 (approved):
Makes sense to me so we continue having discovery act as the source of truth for this

jlajoie (approved):
👍

PR #2603, opened 2020-05-11
Fixing deprecated allow-tags attribute.

awais786:
https://openedx.atlassian.net/browse/BOM-1625

Fixing deprecated allow-tags attribute from django-admin


mzulqarnain1 (approved):

jmbowman (approved):

UsamaSadiq (approved):

awais786:
@mikix Please review.

mikix (commented):
We don't need to replace the functionality with something like `mark_safe`?

awais786:
> We don't need to replace the functionality with something like `mark_safe`?


its using make_safe inside https://docs.djangoproject.com/en/2.2/_modules/django/utils/html/#format_html

mikix (approved):
Awesome, thanks!

PR #2608, opened 2020-05-13
WS-980 expand Chinese language options

rgraber:


course_discovery/apps/ietf_language_tags/models.py
```diff
@@ -21,6 +21,9 @@ def macrolanguage(self):
     def translated_macrolanguage(self):
         return self.name_t.split('-')[0].strip()
 
+    def is_chinese(self):
+        return self.code.startswith("zh")
```
mikix: Sorry, just to understand this PR in general -- this change is to start returning full language tag for Chinese, so like zh_TW and zh_CN etc, instead of just zh?

Is Chinese the only language that wants that treatment? I'm used to pt_BR being a different language tag from pt. There might be others. But maybe we don't use those same tags.

Even if the answer to the above is "no", this PR might feel less hardcoded if this method wasn't called `is_chinese` but rather something like `uses_full_language_tag` or something.

Or maybe even better there was just an `@property` like `translated_macrolanguage` above that hid all that for the consumer. Something like `translated_tag` or whatever. I don't know what a good name is. But it just feels weird that `get_active_language` has to know it should ask if something is Chinese and then it has to know what to do about that.

rgraber: At the moment, the idea of separating out the different versions of Chinese but not of other languages is search-specific, which is why I put the actual logic for returning the full name instead of the macrolanguage in the search model code. This doesn't apply to the language code, just to the display string. I'm fine with pushing it up, but at some point we do have to specify that this only applies to Chinese languages since that's the actual criteria. FWIW we do the same in Elasticsearch.

rgraber: Would it be better to rename get_active_language to get_active_language_display_string?

mikix: Got it, display string, not tag code.

If the logic is search-only, it makes sense to live in the proxy model code, sure. But at that point, I'd move the `is_chinese` call into the proxy model too.

Like, image a Person model with a `is_mike` method on it because some other module living elsewhere cared about Mikes. Just feels like the wrong boundary between modules.

Like, what if you had a `should_display_full_name` method in `algolia_proxy_models.py` that had the same contents of `is_chinese` but just with a different name and isolated to the same place that business logic is used.

Or maybe just put the `startswith('zh')` code directly into `get_active_language` if you don't envision needing such abstractions.

> FWIW we do the same in Elasticsearch

Does that mean we have Chinese-checking code in ES code too? Should we share that logic?

rgraber: unfortunately the logic is not easily shareable. there's just this in the prepare_language code
```
def _prepare_language(self, language):
        if language:
            # ECOM-5466: Render the macro language for all languages except Chinese
            if language.code.startswith('zh'):
                return language.name
            else:
                return language.macrolanguage
        return None
```
the only way i can think to share it would be to put it in the LanguageTag code, which as you said is a bit funny

mikix: Oh that seems fine then. Wouldn't like, a `get_display_name` in LanguageTag be fine then? That basically does that logic above and could be used by ES and Algolia?

My beef with `is_chinese` is how it forces consumers to know to ask about it and then act on it. Bad logic bleeding. (And what if we do later add pt_BR to that list? We add a `is_portugeuese_brazil`?)

But a method like `get_display_name` on the language tag seems like an appropriate encapsulation, yeah? Since this doesn't seem to be an Algolia-only-thing after all. And lets us share this code between modules.

rgraber: Sure, that makes sense

mikix (approved):

albemarle:
Good call on logical boundaries, @mikix!  👍

PR #2611, opened 2020-05-14
WS-1012 boost Spanish language results in spanish_product index

rgraber:


albemarle (approved):

course_discovery/apps/course_metadata/tests/test_algolia_models.py
```diff
@@ -173,9 +166,26 @@ def attach_published_course_run(self, course, run_type="archived"):
             course=course,
             start=course_start,
             end=course_end,
-            status=CourseRunStatus.Published
+            status=CourseRunStatus.Published,
+            **kwargs
         )
 
+
+@override_settings(SETTING_DICT=ChainMap({'AUTO_INDEXING': 'False'}, settings.ALGOLIA))
```
srwang, line 174: what does this line do?

rgraber: explicitly turns off a setting wherein Algolia tries to update the index every time a model is saved. the setting is off anyway, but for some reason it needs to be explicitly set in tests.

course_discovery/apps/course_metadata/algolia_proxy_models.py
```diff
@@ -190,6 +196,13 @@ def product_card_image_url(self):
     def owners(self):
         return get_owners(self)
 
+    @property
+    def promoted_in_spanish_index(self):
+        language_tag = get_active_language_tag(self)
+        if language_tag:
+            return language_tag.code.startswith('es')
```
mikix, line 203: Checking `zh` and `es` this way is fine. But we may want a more robust check than just `startswith` as we add more of this kind of logic, since some languages don't have a two-letter code (ISO 639-1) and instead use a three letter code (ISO 639-2). Take Estonian for example. It's two-letter is `et` but it's three-letter is `est`. If it *didn't* have a two-letter, I assume `language_tag.code` would be `est` and this check would be insufficient. But there are none that actually conflict today with these two prefixes.

Ideally we could ask for language_tag.only_language_family_part_of_code and then do a straight compare instead of startswith. Does that information exist? We could always simulate it with like, `.split('_')[0] == 'es'` I guess.

But anyway, this is fine. Just thinking out loud.

https://www.loc.gov/standards/iso639-2/php/English_list.php

mikix (approved):

course_discovery/apps/course_metadata/algolia_proxy_models.py
rgraber: Worth revisiting for sure. We'll be in this code a fair bit over the next few weeks, I suspect, so we may be able to update it sensibly.

PR #2621, opened 2020-05-19
WS-1025 make default search results configurable in admin

rgraber:
Note: I know the comment about debugging sortable-select is not ideal, but I have a request from product to get this out sooner rather than later with a follow up project to figure out why making the courses sortable causes unexpected UI behavior


reillz10 (approved):
👍

course_discovery/apps/course_metadata/index.py
```diff
@@ -1,80 +1,55 @@
...
+            return [course_rule, program_rule]
+        return []
+
+    # Rules aren't automatically set in regular reindex_all, so set them explicitly
+    def reindex_all(self, batch_size=1000):
+        super().reindex_all(batch_size)
+        self._AlgoliaIndex__index.replace_all_rules(self.get_rules())  # pylint: disable=no-member
+
+
+class ProductIndex(BaseProductIndex):
```
Dillon-Dumesnil: If we are going to have a new class per Language, maybe we should explicitly call out that this it the `EnglishProductIndex`? And require subclasses of `BaseProductIndex` to specify a language? Just a thought, but you know these classes and how they will be used significantly better so I might just be off base here.

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -327,3 +328,13 @@ def should_index(self):
                 self.status == ProgramStatus.Active and
                 self.availability_level and
                 self.partner.name == 'edX')
+
```
Dillon-Dumesnil, line 331: Curious: Is there a reason for the rename of this file?
Looks slightly odd to me now where we import Proxy models from a file named algolia_models instead of algolia_proxy_models.

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -327,3 +328,13 @@ def should_index(self):
...
+
+class SearchDefaultResultsConfiguration(models.Model):
+    index_name = models.CharField(max_length=32, unique=True)
+    programs = SortedManyToManyField(
+        Program, blank=True
+    )
+    courses = SortedManyToManyField(
+        Course, blank=True
+    )
```
Dillon-Dumesnil: Nit: one line these

Dillon-Dumesnil (approved):
Sure, couple small comments, no blockers.

course_discovery/apps/course_metadata/algolia_models.py
rgraber: I added a non-proxy model and figured it might be useful to keep all model-like things in one place

nedbat:
This has been cherry-picked onto Juniper.

PR #2624, opened 2020-05-28
DISCO-1640: sort search results

mikix:
This adds a new ordering param to some search endpoints like /api/v1/search/all. You can now specify ordering=aggregation_key just to get a sensible reliable ordering.

https://openedx.atlassian.net/browse/DISCO-1640

### DIFF
--- a/course_discovery/apps/api/v1/views/search.py
+++ b/course_discovery/apps/api/v1/views/search.py
@@ -2,7 +2,7 @@ import uuid
 
 from django.db.models import Q
 from django.http import QueryDict
-from drf_haystack.filters import HaystackFilter
+from drf_haystack.filters import HaystackFilter, HaystackOrderingFilter
 from drf_haystack.mixins import FacetMixin
 from drf_haystack.viewsets import HaystackViewSet
 from haystack.backends import SQ
@@ -11,7 +11,6 @@ from haystack.query import SearchQuerySet
 from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import ParseError, ValidationError
-from rest_framework.filters import OrderingFilter
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.renderers import BrowsableAPIRenderer, JSONRenderer
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/views/search.py
```diff
@@ -25,8 +24,9 @@
 # pylint: disable=useless-super-delegation
 class BaseHaystackViewSet(mixins.DetailMixin, FacetMixin, HaystackViewSet):
     document_uid_field = 'key'
-    facet_filter_backends = [filters.HaystackFacetFilterWithQueries, filters.HaystackFilter, OrderingFilter]
+    facet_filter_backends = [filters.HaystackFacetFilterWithQueries, filters.HaystackFilter, HaystackOrderingFilter]
```
mikix, line 27: This swap was not required, but just felt like a good idea. Haystack's is a subclass of DRF's with a few more checks. Since we're using haystack here, seemed wise.

Dillon-Dumesnil (approved):
Sure

course_discovery/apps/api/v1/views/search.py
```diff
@@ -25,8 +24,9 @@
...
     ordering_fields = ('start',)
+    ordering = ('aggregation_key',)  # default ordering (needed, otherwise results may change as you paginate)
```
adamstankiewicz: Does `aggregation_key` need to be specified in the `ordering_fields`?

adamstankiewicz (approved):

course_discovery/apps/api/v1/views/search.py
mikix: No. `ordering_fields` is the valid set of fields one can specify using an `?ordering=xxx` flag. The default ordering specified by `ordering` is not validated against that list.

adamstankiewicz: Cool, that's what I figured but just wanted to verify :) Thanks!

mikix:
Updated this to not provide a default ordering, but just let you specify an `ordering=aggregation_key` parameter to get stable sorting.

adamstankiewicz:
> Updated this to not provide a default ordering, but just let you specify an `ordering=aggregation_key` parameter to get stable sorting.

Makes sense. We can add this ordering query parameter on our call to /search/all.

iloveagent57 (approved):
Thank you!

PR #2631, opened 2020-06-11
DISCO-1643: Add in Honor Course Types to be created by default

Dillon-Dumesnil:
This will create the Course Types of Honor Only, Verified and Honor,
and Credit with Honor (Credit, Verified, Honor). Although Honor
is not used internally to edX, it is widely used in the Open edX
community and this work is to continue to support their use cases.

### DIFF
--- a/course_discovery/apps/course_metadata/migrations/0251_add_honor_course_type.py
+++ b/course_discovery/apps/course_metadata/migrations/0251_add_honor_course_type.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+# Contains the slug (unique identifier) and any fields that are not the default-model values
+MODE_SLUGS_AND_DEFAULTS = (
+    ('honor', {
+        'name': 'Honor',
+        'certificate_type': 'honor',
+    }),
+)
+# The slugs are identical across Mode and SeatType
+TRACK_MODE_AND_SEAT_SLUGS = ('honor',)
+COURSE_RUN_TYPE_NAMES_SLUGS_AND_TRACK_SLUGS = (
+    ('Honor Only', 'honor', ['honor']),
+    ('Verified and Honor', 'verified-honor', ['honor', 'verified']),
+    ('Credit with Honor', 'credit-verified-honor', ['credit', 'honor', 'verified']),
+)
+COURSE_TYPE_NAMES_SLUGS_SEAT_TYPE_SLUGS_AND_RUN_TYPE_SLUGS = (
+    ('Honor Only', 'honor', [], ['honor']),
+    ('Verified and Honor', 'verified-honor', ['verified'], ['honor', 'verified-honor']),
+    ('Credit with Honor', 'credit-verified-honor', ['verified'], ['honor', 'verified-honor', 'credit-verified-honor']),
+)
+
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/migrations/0251_add_honor_course_type.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+# Contains the slug (unique identifier) and any fields that are not the default values
```
mikix: You mean default-model values, but the variable is called `MODE_SLUGS_AND_DEFAULTS` and you're setting the `defaults` dict for a create. So it's a *little* confusing to say "not the default values". But I get you.

course_discovery/apps/course_metadata/migrations/0251_add_honor_course_type.py
```diff
@@ -0,0 +1,84 @@
...
+COURSE_RUN_TYPE_NAMES_SLUGS_AND_TRACK_SLUGS = (
+    ('Honor Only', 'honor', ['honor']),
+    ('Verified and Honor', 'verified-honor', ['honor', 'verified']),
+    ('Credit with Honor', 'credit-verified-honor', ['credit', 'honor', 'verified']),
+)
+COURSE_TYPE_NAMES_SLUGS_SEAT_TYPE_SLUGS_AND_RUN_TYPE_NAMES = (
+    ('Honor Only', 'honor', [], ['Honor Only']),
+    ('Verified and Honor', 'verified-honor', ['verified'], ['Honor Only', 'Verified and Honor']),
+    ('Credit with Honor', 'credit-verified-honor', ['verified'], ['Honor Only', 'Verified and Honor', 'Credit with Honor']),
+)
```
mikix: Why do we use the run type name as they key here instead of a slug?

mikix (approved):
Nice work!

course_discovery/apps/course_metadata/migrations/0251_add_honor_course_type.py
Dillon-Dumesnil: Yes i do. This was mostly copy-pasted from the original course type migration. Just a miss there as well. I'll update.

course_discovery/apps/course_metadata/migrations/0251_add_honor_course_type.py
Dillon-Dumesnil: Uhm... another legacy thing from the copy paste. Will update as well.

mikix (approved):
Beautiful

mikix (approved):
I did not pore over the migration code, but looks generally good. Like the new documentation stuff, very helpful!

jmyatt (approved):
👍  It's testament to the versatility of the CourseType/CourseRunType design that pulling in Honor is pretty straightforward

pomegranited:
Hi @Dillon-Dumesnil , thanks for this! Will it get cherry-picked into `juniper.master`?

Dillon-Dumesnil:
Hi @pomegranited, yes, it will. I'm working with Ned Batchelder at the moment to add this in.

nedbat:
This has been cherry-picked onto Juniper.

pomegranited:
Brilliant, thank you both!

PR #2639, opened 2020-06-22
WS-1115 hide hidden programs from search

rgraber:


course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -327,7 +327,8 @@ def should_index(self):
                 self.program_types and
                 self.status == ProgramStatus.Active and
                 self.availability_level and
-                self.partner.name == 'edX')
+                self.partner.name == 'edX' and not
+                self.hidden)
```
mikix: `not` on the previous line? Seems like it obcures what the condition is.

mikix (approved):

JackedX (approved):
Good to me

albemarle (approved):

course_discovery/apps/course_metadata/tests/test_algolia_models.py
```diff
@@ -458,3 +458,9 @@ def test_do_not_index_if_not_active(self):
         assert not unpublished_program.should_index
         assert not retired_program.should_index
         assert not deleted_program.should_index
+
+    def test_do_not_index_if_hidden(self):
+        program = AlgoliaProxyProgramFactory(partner=self.__class__.edxPartner, hidden=True)
+        program.authoring_organizations.add(OrganizationFactory())
+        self.attach_archived_course(program=program)
+        assert not program.should_index
```
mikix, line 466: I have a general comment here -- not something worth delaying this cat1 fix, but just saying.

For these kinds of tests, I prefer to see a pattern like "Do it one way (wrong or right). Change one thing. Do it again with different results."

I know that _you know_ that the `hidden=True` flag is the only difference with the "happy path" test in this class. But imagine someone comes by in a year and says that for a program to be indexed, its title needs to start with T. And then the happy path test fails. So they fix it. And they they don't know about all the other tests with implicit "everything but this one change is the happy path" logic.

So now you have a test here with two things preventing it from being indexed. And we are no longer truly testing the hidden condition.

The "do once, change, re-examine" pattern is slower (more work done in each test), but I think it makes for more robust tests. Besides the contrived example above, I've seen people write tests for a change like yours that accidentally had two gating conditions in the test when they wrote it (the one they added and just another one that they forgot about) and they miss a bug in their new code because of it.

And it's also more DRY because you can just move the happy path test into a method like `assert_happy_path` or `set_up_happy_path` or whatever. Rather than having the happy path setup duplicated in each of these tests. Then have a new test like yours do:

```
self.assert_should_index()
self.program.hidden = True
self.program.save()
assert not program.should_index
```

Or similar.

PR #2645, opened 2020-06-30
WS-1116 update person endpoint to accept slug as well as uuid

rgraber:


course_discovery/apps/api/v1/views/people.py
```diff
@@ -20,15 +24,44 @@ class PersonViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
...
+    # override get_object to allow multiple lookup fields
+    def get_object(self):
+
+        def is_valid_uuid(uuid_to_test, version=4):
+            try:
+                uuid_obj = UUID(uuid_to_test, version=version)
+            except ValueError:
+                return False
+
+            return str(uuid_obj) == uuid_to_test
```
Dillon-Dumesnil: Is it possible for this to not be True since the uuid_obj was created using the uuid_to_test? Might be more clear to just return True if not.

course_discovery/apps/api/v1/views/people.py
```diff
@@ -20,15 +24,44 @@ class PersonViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
...
     pagination_class = PageNumberPagination
     metadata_class = MetadataWithRelatedChoices
     metadata_related_choices_whitelist = ('organization',)
 
+    course_uuid_regex = re.compile(COURSE_UUID_REGEX)
+
+    # override get_object to allow multiple lookup fields
+    def get_object(self):
+
+        def is_valid_uuid(uuid_to_test, version=4):
```
Dillon-Dumesnil: Nit: This is a sub-function inside of get_object. You can just pass version=4 straight into UUID() and remove it as a kwarg. It's not going to be called by anyone else so you have complete control over the version either way.

Dillon-Dumesnil (approved):
Small clarifying question and then a nit, but looks good! Nice 👍

course_discovery/apps/api/v1/views/people.py
rgraber: oops, copypasta. good catch.

reillz10 (approved):
lgtm

PR #2650, opened 2020-07-01
WS-1143 adding degree.micromasters_path

JackedX:
Added micromasters_path to DegreeSerializer

some regex logic to remove the host from the url


Dillon-Dumesnil (approved):
Looks fine to me. Just check in on the tests and quality

PR #2651, opened 2020-07-02
WS-1165 add org logo to position serializer

rgraber:


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1770,6 +1771,7 @@ def test_position_with_no_org(self):
             'organization_id': None,
             'organization_override': None,
             'organization_marketing_url': None,
+            'organization_logo_image_url': None,
```
mikix, line 1774: You ought to have a test that actually tests the new code path too.

course_discovery/apps/api/serializers.py
```diff
@@ -309,7 +310,7 @@ class Meta:
         model = Position
         fields = (
             'title', 'organization_name', 'organization', 'organization_id', 'organization_override',
-            'organization_marketing_url', 'organization_uuid',
+            'organization_marketing_url', 'organization_uuid', 'organization_logo_image_url'
```
mikix, line 313: Feels like we should have started this serializer with a nested serializer. Ah well.

mikix (approved):
There's a missing test, but approving now to not be a blocker.

course_discovery/apps/api/tests/test_serializers.py
rgraber: Not sure I understand. Isn't that covered in the previous test?

mikix: Yeah, looks like it! Missed that line, when I was reviewing.

albemarle (approved):

PR #2654, opened 2020-07-07
WS-1143: adding degree.micromasters_path to discovery

JackedX:
checking degree micromasters_url is a string before calling regex for…… micromasters_path in serializer.js


Dillon-Dumesnil (approved):

srwang (approved):
what was the issue?

JackedX:
> what was the issue?

I needed to check that the degree.micromasters_url was a string before calling regex functions

rgraber:
Do we know under what circumstances it's not a string?

JackedX:
> Do we know under what circumstances it's not a string?

I coded and tested for the two following conditions
1. There are micromasters without micromasters_url
2. Programs without degrees

PR #2656, opened 2020-07-09
Swang/ws 1009

reillz10:


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -239,7 +239,7 @@ def test_list_query(self):
 
         # Known to be flaky prior to the addition of tearDown()
         # and logout() code which is the same number of additional queries
-        with self.assertNumQueries(39):
+        with self.assertNumQueries(42):
```
mikix, line 242: Could this be avoided by actually using the prefetch_queryset call that was added?

reillz10: Potentially, although it didn't avoid it locally when I updated the code to use it. That said, I might be doing something wrong.

course_discovery/apps/ietf_language_tags/migrations/0004_auto_20200618_1744.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.12 on 2020-06-18 17:44
```
rgraber: remove this file. this one is one me, i need to wrap up this migration, but that will be a separate ticket

rgraber (changes requested):
We need to test the collaborator serializer beyond the empty set

course_discovery/apps/ietf_language_tags/migrations/0004_auto_20200618_1744.py
reillz10: 👍

JackedX (approved):
⛵️LGTM

course_discovery/apps/ietf_language_tags/migrations/0004_auto_20200618_1744.py
```diff
@@ -0,0 +1,18 @@
...
+        ('ietf_language_tags', '0003_auto_20200506_1245'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='languagetagtranslation',
+            name='name_t',
+            field=models.CharField(max_length=255, verbose_name='Name for translation'),
+        ),
+    ]
```
mikix: This was a forgotten migration that you're throwing in?

course_discovery/apps/course_metadata/models.py
```diff
@@ -724,6 +724,26 @@ def search(cls, query, queryset=None):
         return queryset.filter(pk__in=ids)
 
 
+class Collaborator(TimeStampedModel):
```
mikix, line 727: This seems awfully similar to a Person model, just dropping some (blankable) fields you don't care about. Is there a reason not to use that? Person models are used for Endorsements and as staff for CourseRuns.

So we have CourseEditors for Publisher writers. Person for instructors. And Collaborators for Studio writers?

(I also have a gut WW2 reaction to the word collaborators, but maybe that's just me.)

course_discovery/apps/api/v1/tests/test_views/test_courses.py
mikix: I think that means it wasn't being used?

I mean, this test doesn't make any Collaborators so this is 3 extra queries to find nothing. I bet that should either be 0 or 1 as part of a big prefetch join?

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -261,7 +261,7 @@ def test_list_uuid_filter(self):
         uuids = ','.join([str(course.uuid) for course in courses])
         url = '{root}?uuids={uuids}'.format(root=reverse('api:v1:course-list'), uuids=uuids)
 
-        with self.assertNumQueries(39):
+        with self.assertNumQueries(42):
```
mikix, line 264: You don't have any tests that actually create and test the serialization of a Collaborator, with a factory and all that.

course_discovery/apps/course_metadata/models.py
reillz10: Its probably more similar to a Organization model (although, presumably the collaborator could be an organization or a person), with the blankable fields and the specific dimensions that we want for the image. I wasn't sure if that would have any negative effects such as new school pages being built for these organizations.

Screen Shot 2020-07-10 at 9 40 44 AM

course_discovery/apps/api/v1/tests/test_views/test_courses.py
reillz10: added one in `test_serializers.py`

course_discovery/apps/ietf_language_tags/migrations/0004_auto_20200618_1744.py
reillz10: removed

course_discovery/apps/api/v1/tests/test_views/test_courses.py
rgraber: I have a hunch about fixing this but would like to put that in the next PR (should be a pretty fast follow) to unblock some work in prospectus that will consume the new field.

rgraber (approved):

mikix (approved):
Sure!

PR #2657, opened 2020-07-09
Revert "WS-728: Make default search results Contentful-driven"

srwang:
Reverts edx/course-discovery#2647

Reverting until we can bring pre-search state back up with a better user experience.


course_discovery/apps/course_metadata/migrations/0252_auto_20200624_1533.py
```diff
@@ -1,16 +0,0 @@
-# Generated by Django 2.2.12 on 2020-06-24 15:33
```
mikix, line 1: You can't just delete migrations. You need to generate a new migration that undoes it (in this case, creates the model SearchDefaultResultsConfiguration again).

PR #2658, opened 2020-07-09
WS-728: Undo moving default products to contentful

srwang:
Putting default products back into Algolia config until we can get pre-search state back up with a better user experience.


mikix (commented):
This is mostly a straight revert, except for the migration?

Is planned to be re-reverted again? I ask because if this change is staying like this, you might want to also squash migrations so that the drop&add of the model doesn't hit open edx installations in the next open edx release, deleting all their data for that model.

srwang:
> This is mostly a straight revert, except for the migration?
> 
> Is planned to be re-reverted again? I ask because if this change is staying like this, you might want to also squash migrations so that the drop&add of the model doesn't hit open edx installations in the next open edx release, deleting all their data for that model.

yes, the idea is to re-revert later.

mikix (approved):

PR #2662, opened 2020-07-16
WS-1180: Collaborator endpoints

JackedX:
Adds a GET, POST,  LIST endpoint at /api/v1/collaborators

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -110,6 +110,7 @@ class CharListFilter(filters.CharFilter):
 
         return super(CharListFilter, self).filter(qs, value)
 
+
 class UUIDListFilter(CharListFilter):
     """ Filters a field via a comma-delimited list of UUIDs. """
 
@@ -259,8 +260,9 @@ class CourseEditorFilter(filters.FilterSet):
         model = CourseEditor
         fields = ('course',)
 
+
 class CollaboratorFilter(filters.FilterSet):
+
     class Meta:
         model = Collaborator
         fields = ('name', 'uuid',)
-
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -16,7 +16,7 @@ from rest_framework.exceptions import NotFound, PermissionDenied
 from course_discovery.apps.api.utils import cast2int
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
-    Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject, Topic
+    Collaborator, Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject, Topic
 )
 
 logger = logging.getLogger(__name__)
@@ -108,7 +108,6 @@ class CharListFilter(filters.CharFilter):
             value = value.split(',')
 
         return super(CharListFilter, self).filter(qs, value)
-
 
 class UUIDListFilter(CharListFilter):
     """ Filters a field via a comma-delimited list of UUIDs. """
@@ -258,3 +257,9 @@ class CourseEditorFilter(filters.FilterSet):
     class Meta:
         model = CourseEditor
         fields = ('course',)
+
+class CollaboratorFilter(filters.FilterSet):
+    class Meta:
+        model = Collaborator
+        fields = ('name', 'uuid',)
+
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -16,7 +16,8 @@ from rest_framework.exceptions import NotFound, PermissionDenied
 from course_discovery.apps.api.utils import cast2int
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
-    Collaborator, Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject, Topic
+    Collaborator, Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject,
+    Topic
 )
 
 logger = logging.getLogger(__name__)
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -259,10 +259,3 @@ class CourseEditorFilter(filters.FilterSet):
     class Meta:
         model = CourseEditor
         fields = ('course',)
-
-
-class CollaboratorFilter(filters.FilterSet):
-
-    class Meta:
-        model = Collaborator
-        fields = ('name', 'uuid',)
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -16,8 +16,7 @@ from rest_framework.exceptions import NotFound, PermissionDenied
 from course_discovery.apps.api.utils import cast2int
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
-    Collaborator, Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject,
-    Topic
+    Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject, Topic
 )
 
 logger = logging.getLogger(__name__)
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1079,7 +1079,8 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     course_run_statuses = serializers.ReadOnlyField()
     editors = CourseEditorSerializer(many=True, read_only=True)
     collaborators = SlugRelatedFieldWithReadSerializer(slug_field='uuid', required=False, many=True,
-                                                       queryset=Collaborator.objects.all(), read_serializer=CollaboratorSerializer())
+                                                       queryset=Collaborator.objects.all(),
+                                                       read_serializer=CollaboratorSerializer())
 
     @classmethod
     def prefetch_queryset(cls, partner, queryset=None, course_runs=None):  # pylint: disable=arguments-differ
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -278,9 +278,9 @@ class CollaboratorSerializer(BaseModelSerializer):
             return Collaborator.objects.all()
         return None
 
-    def get_image_url(self, obj):
+    def image_url(self, obj):
         if obj.image:
-            return obj.get_image_url
+            return obj.image_url
         return None
 
     def create(self, validated_data):
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -274,9 +274,7 @@ class CollaboratorSerializer(BaseModelSerializer):
 
     @classmethod
     def prefetch_queryset(cls):
-        if Collaborator.objects:
-            return Collaborator.objects.all()
-        return None
+        return Collaborator.objects.all()
 
     def image_url(self, obj):
         if obj.image:
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -268,15 +268,23 @@ class SubjectSerializer(DynamicFieldsMixin, BaseModelSerializer):
 
 class CollaboratorSerializer(BaseModelSerializer):
     """Serializer for the ``Collaborator`` model."""
-    image = StdImageSerializerField()
+    image = StdImageSerializerField(required=False)
+    image_url = serializers.SerializerMethodField()
 
     @classmethod
     def prefetch_queryset(cls):
-        return Collaborator.objects.all()
+        if Collaborator.objects:
+            return Collaborator.objects.all()
+        return None
+
+    def get_image_url(self, obj):
+        if obj.image:
+            return obj.get_image_url
+        return None
 
     class Meta:
         model = Collaborator
-        fields = ('name', 'image', 'uuid')
+        fields = ('name', 'image', 'image_url', 'uuid')
 
     @property
     def choices(self):
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -285,6 +285,7 @@ class CollaboratorSerializer(BaseModelSerializer):
 
     def create(self, validated_data):
         return Collaborator.objects.create(**validated_data)
+
     class Meta:
         model = Collaborator
         fields = ('name', 'image', 'image_url', 'uuid')
@@ -1083,8 +1084,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     course_run_statuses = serializers.ReadOnlyField()
     editors = CourseEditorSerializer(many=True, read_only=True)
     collaborators = SlugRelatedFieldWithReadSerializer(slug_field='uuid', required=False, many=True,
-                                               queryset=Collaborator.objects.all(),
-                                               read_serializer=CollaboratorSerializer())
+                                                       queryset=Collaborator.objects.all(), read_serializer=CollaboratorSerializer())
 
     @classmethod
     def prefetch_queryset(cls, partner, queryset=None, course_runs=None):  # pylint: disable=arguments-differ
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1105,13 +1105,13 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
             'expected_learning_items',
             'prerequisites',
             'subjects',
+            'collaborators',
             'topics',
             'url_slug_history',
             'url_redirects',
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
             Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
             Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
-            Prefetch('collaborators', queryset=CollaboratorSerializer.prefetch_queryset()),
         )
 
     class Meta(MinimalCourseSerializer.Meta):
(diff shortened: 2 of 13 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -270,6 +270,10 @@ class CollaboratorSerializer(BaseModelSerializer):
     """Serializer for the ``Collaborator`` model."""
     image = StdImageSerializerField()
 
+    @classmethod
```
rgraber, line 275: Why do we need this?

course_discovery/apps/api/filters.py
```diff
@@ -258,3 +259,10 @@ class CourseEditorFilter(filters.FilterSet):
     class Meta:
         model = CourseEditor
         fields = ('course',)
+
```
rgraber: we only need to get by uuid so i don't think we need this filter

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,81 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=useless-super-delegation
+class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
+    """ CollaboratorSerializer resource. """
+
+    filter_backends = (DjangoFilterBackend,)
+    filterset_class = filters.CollaboratorFilter
```
rgraber: I think we can get rid of the filter logic

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,81 @@
...
+    lookup_field = 'uuid'
+    lookup_value_regex = '[0-9a-f-]+'
+    permission_classes = (DjangoModelPermissionsOrAnonReadOnly,)
+    queryset = serializers.CollaboratorSerializer.prefetch_queryset()
```
rgraber, line 21: I don't think we need to specify this if the queryset is just Collaborators.objects.all()

course_discovery/apps/core/tests/factories.py
```diff
@@ -13,6 +14,12 @@ def add_m2m_data(m2m_relation, data):
         m2m_relation.add(*data)
 
 
+class CollaboratorFactory(factory.DjangoModelFactory):
+    class Meta:
+        model = Collaborator
+    name = (factory.Faker('first_name'), factory.Faker('last_name'))
```
rgraber: I think there are factories for images that would be useful to have

rgraber (changes requested):
Couple notes, also it would be good to include the update to the /courses endpoint (allowing it to update a course's collaborators based on a list of uuids) as part of this PR so we don't have to go through discovery release twice

course_discovery/apps/api/v1/views/collaborators.py
JackedX: if I take out this queryset I get this error message

'CollaboratorViewSet' should either include a `queryset` attribute, or override the `get_queryset()` method.

course_discovery/apps/core/tests/factories.py
JackedX: K I'll work on this.

course_discovery/apps/api/v1/views/collaborators.py
rgraber: Did not know that, thanks

course_discovery/apps/api/serializers.py
JackedX: Yeah we do actually :(

course_discovery/apps/api/v1/views/collaborators.py
JackedX: Done

course_discovery/apps/api/serializers.py
```diff
@@ -270,10 +270,19 @@ class CollaboratorSerializer(BaseModelSerializer):
...
+    def prefetch_queryset(cls):
+        return Collaborator.objects.all()
+
     class Meta:
         model = Collaborator
         fields = ('name', 'image', 'uuid')
 
+    @property
+    def choices(self):
+        # choices shows the possiblity values via HTTP's OPTIONS verb
```
albemarle: typo in "possible"

course_discovery/apps/api/filters.py
JackedX: removed

course_discovery/apps/core/tests/factories.py
JackedX: added FakeURL calls

course_discovery/apps/api/serializers.py
```diff
@@ -268,11 +269,31 @@ def choices(self):
...
+    def create(self, validated_data):
+        return Collaborator.objects.create(**validated_data)
 
     class Meta:
         model = Collaborator
-        fields = ('name', 'image', 'uuid')
+        fields = ('name', 'image', 'image_url', 'uuid')
+
+    @property
+    def choices(self):
```
rgraber: this is unused

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -142,7 +142,7 @@ def test_retrieve_curriculum_with_child_programs(self, django_assert_num_queries
             curriculum=curriculum
         )
 
-        with django_assert_num_queries(FuzzyInt(60, 2)):
+        with django_assert_num_queries(FuzzyInt(63, 2)):
```
rgraber, line 145: I'm surprised this went up since we already had collaborators in the CourseSerializer. Any idea what caused it?

course_discovery/apps/api/v1/tests/test_views/test_collaborators.py
```diff
@@ -0,0 +1,33 @@
+from rest_framework.reverse import reverse
+
+from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase, OAuth2Mixin, SerializationMixin
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
+from course_discovery.apps.course_metadata.tests.factories import CollaboratorFactory
+
+
+class CollaboratorViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
```
rgraber, line 8: Need tests for /update

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -33,7 +33,7 @@
 class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
     def setUp(self):
         super(CourseViewSetTests, self).setUp()
-        self.user = UserFactory(is_staff=True)
+        self.user = UserFactory(is_staff=True, is_superuser=True)
```
rgraber: This is a little concerning, do you know why it had to be updated to superuser?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -218,6 +219,12 @@ def create(self, request, *args, **kwargs):
         organization = Organization.objects.get(key=course_creation_fields['org'])
         course.authoring_organizations.add(organization)
 
+        collaborators = request.data.get('collaborators')
+        for uuid in collaborators:
+            collaborator = Collaborator.objects.get(uuid=uuid)
+            course.collaborators.add(collaborator)
+            course.save()
```
rgraber: A bulk add would save us a few trips to the database

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -341,6 +348,12 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
             image_data = ContentFile(base64.b64decode(imgstr), name='tmp.{extension}'.format(extension=ext))
             course.image.save(image_data.name, image_data)
 
+        if data.get('collaborators'):
```
rgraber, line 350: see previous comment about bulk adds

rgraber (changes requested):
Needs tests around update and I had a few questions, but is mostly there

course_discovery/apps/api/v1/tests/test_views/test_courses.py
JackedX: yeah we don't need that anymore

course_discovery/apps/api/serializers.py
JackedX: removed

course_discovery/apps/api/v1/tests/test_views/test_programs.py
JackedX: This happened on the build serve... I haven't been able to fail it locally

course_discovery/apps/api/v1/tests/test_views/test_collaborators.py
JackedX: will do

course_discovery/apps/api/v1/tests/test_views/test_programs.py
JackedX: reverted

rgraber (approved):
LGTM

reillz10 (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -268,11 +269,26 @@ def choices(self):
...
 class CollaboratorSerializer(BaseModelSerializer):
     """Serializer for the ``Collaborator`` model."""
-    image = StdImageSerializerField()
+    image = StdImageSerializerField(required=False)
+    image_url = serializers.SerializerMethodField()
+
+    @classmethod
+    def prefetch_queryset(cls):
+        if Collaborator.objects:
+            return Collaborator.objects.all()
```
mikix: Why would the collaborator class not have a model manager?

course_discovery/apps/course_metadata/models.py
```diff
@@ -740,6 +740,12 @@ class Collaborator(TimeStampedModel):
     name = models.CharField(max_length=255, default='')
     uuid = models.UUIDField(default=uuid4, editable=False, verbose_name=_('UUID'))
 
+    @property
+    def get_image_url(self):
```
mikix: It seems confusing to name a property with a verb like that. I know `Person` does that with `get_profile_image_url`, but that is more of a mistake to not repeat than example to follow. I'd say follow the footsteps of `Course` and `CourseRun` and just call this `image_url`.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -105,7 +105,7 @@ def test_retrieve(self, django_assert_num_queries):
         """ Verify the endpoint returns the details for a single program. """
         program = self.create_program()
 
-        with django_assert_num_queries(FuzzyInt(55, 2)):
+        with django_assert_num_queries(FuzzyInt(55, 3)):
```
mikix: Did this get actually get flakier/fuzzier or did the base value just increase?

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,84 @@
...
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=useless-super-delegation
+class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
+    """ CollaboratorSerializer resource. """
+
+    lookup_field = 'uuid'
+    lookup_value_regex = '[0-9a-f-]+'
+    permission_classes = (IsAuthenticated, IsCourseEditorOrReadOnly,)
```
mikix: `IsCourseEditorOrReadOnly` expects to be working on `Course` objects -- you can see that from its `has_object_permissions` code. You may want your own `IsCollaboratorEditorOrReadOnly` or something that is named better.

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,84 @@
...
+    queryset = serializers.CollaboratorSerializer.prefetch_queryset()
+    serializer_class = serializers.CollaboratorSerializer
+    pagination_class = PageNumberPagination
```
mikix: Please use `CursorPagination` for all new endpoints. It's got much better performance when iterating on larger data sets (not that this endpoint is likely to be in thousands, but better safe than sorry).

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,84 @@
...
+    metadata_class = MetadataWithRelatedChoices
```
mikix: Why this metadata class? That's useful if you want to include the full set of choices for a related field in your model in OPTIONS calls. Collaborator doesn't have related fields, right?

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def create(self, request, *args, **kwargs):
```
mikix, line 25: Is this method definition useful? It's exactly the same as `super().create` except that it has special handling for an exception from `perform_create`. Which, I'm not sure how useful that is.

Maybe we can just call the super instead of redefining it?

Same with update and partial_update below. I don't think we need any of this custom logic.

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,84 @@
...
+    def update(self, request, *args, **kwargs):
+        return self.update_collaborator(request.data)
+
+    def partial_update(self, request, *_args, **_kwargs):
+        """ Partially update details of collaborator. """
+        return self.update_collaborator(request.data, partial=True)
+
+    def list(self, request, *args, **kwargs):
+        """ Retrieve a list of all collaborators. """
+        return super(CollaboratorViewSet, self).list(request, *args, **kwargs)
```
mikix: Python3 style is to just call `super().list(...)` without the class name and self there.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -218,6 +219,14 @@ def create(self, request, *args, **kwargs):
         organization = Organization.objects.get(key=course_creation_fields['org'])
         course.authoring_organizations.add(organization)
 
+        collaborators_uuid = request.data.get('collaborators')
+        collaborators = []
+        if collaborators_uuid:
+            for uuid in collaborators_uuid:
+                collaborators.append(Collaborator.objects.get(uuid=uuid))
```
mikix: `Collaborator.objects.filter(uuid__in=collaborators_uuid)` is just one lookup, not N. Same below in the update call.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -218,6 +219,14 @@ def create(self, request, *args, **kwargs):
...
+            course.collaborators.add(*collaborators)
+            course.save()
```
mikix: You don't need `save()` here -- `add()` does the save already.

course_discovery/apps/core/tests/factories.py
```diff
@@ -13,6 +17,15 @@ def add_m2m_data(m2m_relation, data):
         m2m_relation.add(*data)
 
 
+class CollaboratorFactory(factory.DjangoModelFactory):
+    name = (factory.Faker('first_name'), factory.Faker('last_name'))
```
mikix: Is that a tuple or a string that ends up being set?

course_discovery/apps/api/serializers.py
JackedX: Well Django actually creates a default manager for all models... I don't need to do anything custom with this so I think a manager may just be extra overhead.

"By default, Django adds a Manager with the name objects to every Django model class. However, if you want to use objects as a field name, or if you want to use a name other than objects for the Manager, you can rename it on a per-model basis. "

https://docs.djangoproject.com/en/3.0/topics/db/managers/

course_discovery/apps/api/v1/views/collaborators.py
JackedX: removed

course_discovery/apps/core/tests/factories.py
JackedX: suppose to be a string

course_discovery/apps/api/serializers.py
mikix: Right. So what I'm saying is, why check `if Collaborator.objects:`? It should always be present.

course_discovery/apps/core/tests/factories.py
mikix: Right, but you are setting name as a tuple of strings. Like, in a console, play with `('a', 'b')` vs `'a' + 'b'` or whatever. Maybe the factory code does something smart with that though.

But it doesn't matter, you can just use `name = factory.Faker('name')` instead of breaking it up and stitching it back.

course_discovery/apps/api/serializers.py
JackedX: ah that's what you meant... Python Phun

course_discovery/apps/core/tests/factories.py
JackedX: yeah I eventually did that ;)

course_discovery/apps/api/serializers.py
JackedX: removed the if

course_discovery/apps/api/v1/tests/test_views/test_programs.py
JackedX: Not sure how to tell the difference

course_discovery/apps/api/v1/views/collaborators.py
JackedX: Thanks for the feedback updated...

I spoke with Rebecca and she's ok with replicating the Course permission for collaborators

course_discovery/apps/api/v1/tests/test_views/test_programs.py
JackedX: thanks for explaining the subtleties of this queries

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -0,0 +1,22 @@
...
+from rest_framework.permissions import IsAuthenticated
+
+from course_discovery.apps.api import serializers
+from course_discovery.apps.api.cache import CompressedCacheResponseMixin
+from course_discovery.apps.api.permissions import IsInOrgOrReadOnly
+
+logger = logging.getLogger(__name__)
+
+
+class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
```
mikix, line 15: Removing the stub methods that just called super() is fine. But just FYI for why you'll see that pattern in Discovery is for the autogenerated swagger docs (http://localhost:18381/api-docs/)

The stubs were included just for their docstrings being a little specific ("list all Collaborators") vs whatever the default text is in swagger. I'm ambivalent on the value of that. But that's what they are doing when you see the pattern.

mikix (approved):
Looks good!  Thanks for dealing with me.  :)

course_discovery/apps/api/v1/views/collaborators.py
JackedX: ahh that makes sense I'll put them back

JackedX:
@mikix It was a learning experience! Thank you so much!

PR #2702, opened 2020-08-13
WS-1258 update collaborator image size reqs

rgraber:


albemarle (approved):

srwang (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -733,9 +733,9 @@ class Collaborator(TimeStampedModel):
         blank=True,
         null=True,
         variations={
-            'original': (80, 80),
+            'original': (200, 100),
```
mikix, line 736: Does this have any implications for existing data that might want to be converted or anything like that? (Or do you not have existing collaborators yet?)

mikix (approved):

course_discovery/apps/course_metadata/models.py
rgraber: we only have 2 collaborators right now, and we are planning to just manually remove/replace them

PR #2703, opened 2020-08-13
If we have a UserThrottle row for staff, still throttle them

mikix:
Before this change, even if we added a custom throttle for a staff user, it wouldn't take effect.


Dillon-Dumesnil (approved):

course_discovery/apps/core/tests/test_throttles.py
```diff
@@ -16,52 +19,52 @@ class RateLimitingExceededTest(SiteMixin, APITestCase):
...
+        super().setUp()
 
         self.url = reverse('api_docs')
         self.user = UserFactory()
         self.client.login(username=self.user.username, password=USER_PASSWORD)
-        self.default_rate_patcher = mock.patch.dict(
-            settings.REST_FRAMEWORK['DEFAULT_THROTTLE_RATES'],
-            {'user': '10/hour'}
-        )
-        self.default_rate_patcher.start()
```
mikix, line 28: This wasn't actually working, as the throttling class didn't notice the change. It was sticking with its default of 100/hour (which meant that these tests were doing 100 endpoint requests each).

course_discovery/apps/core/tests/test_throttles.py
```diff
@@ -16,52 +19,52 @@ class RateLimitingExceededTest(SiteMixin, APITestCase):
...
             Response: Response of the last request.
         """
-        default_num_requests = OverridableUserRateThrottle().num_requests
-        for __ in range(default_num_requests + 1):
-            response = self.client.get(self.url)
+        count = count or 6
+        user_rates = {'user': '5/hour'}
+        new_drf = deepcopy(settings.REST_FRAMEWORK)
+        new_drf['DEFAULT_THROTTLE_CLASSES'] = ('course_discovery.apps.core.throttles.OverridableUserRateThrottle',)
+        with override_settings(REST_FRAMEWORK=new_drf):
```
mikix: Need to use `override_settings`, because that emits a special django "settings changed" signal that is required to get DRF to notice we swapped the classes on them.

course_discovery/apps/core/tests/test_throttles.py
```diff
@@ -16,52 +19,52 @@ class RateLimitingExceededTest(SiteMixin, APITestCase):
...
+            with patch('course_discovery.apps.core.throttles.OverridableUserRateThrottle.THROTTLE_RATES', user_rates):
```
mikix: But that doesn't work for this setting, because the class doesn't pay attention to the signal and just encodes the global value at import time. :-/

PR #2704, opened 2020-08-14
WS-1249: logging to debug why course slugs sporadically update

srwang:


mikix (approved):
This feels like the act of a desperate programmer. :)  Hope you're keeping your sanity over there.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1013,6 +1013,11 @@ def _update_or_create_official_version(self, course_run):
 
     @transaction.atomic
     def set_active_url_slug(self, slug):
+        # logging to help debug error around course url slugs incrementing
+        logger.info('The current slug is {}'.format(self.url_slug))
```
albemarle: For logging like this, it's easier to pull out of Splunk if it's all in one line, because you can just do a grep and then quickly scan the results.  That said, I don't think that's a blocker.

albemarle (approved):

PR #2708, opened 2020-08-18
WS-1292 Explicitly allow prospectus worker to get all orgs

rgraber:


mikix (approved):
Presumably we'll drop this once we fix the user up again.

rgraber:
> Presumably we'll drop this once we fix the user up again.

that's the plan. SRE is aware of the issue.

albemarle (approved):
Can you add the Jira ticket to the commit message?

rgraber:
> Can you add the Jira ticket to the commit message?

Will do

PR #2711, opened 2020-08-18
Upgrade factory-boy to 3.0.1 (and upgrade other packages)

cpennington:


course_discovery/apps/catalogs/tests/factories.py
```diff
@@ -1,10 +1,11 @@
 import factory
```
Dillon-Dumesnil, line 1: Can we remove this import then?

Also nit: `from factory.django import DjangoModelFactory`

mikix (approved):

Dillon-Dumesnil (approved):
Just one comment that applies to most of the files here about removing the old factory import. Otherwise, LGTM

course_discovery/apps/catalogs/tests/factories.py
cpennington: No, we can't remove the other import, because it's used by `factory.post_generation`. Stylistically, we could switch imports around, but I didn't feel like doing that as part of this commit.

Dillon-Dumesnil: Whoop, missed the post_generation call. Darn purple text on black background. My bad. All approvals all around.

waheedahmed:
cleaning the old stuff.

PR #2718, opened 2020-08-28
Serve static files in decentralized devstack ARCHBOM-1473

jmbowman:
Explicitly add static file serving to the URL configuration in decentralized devstack, since we aren't using runserver (which does this automatically).  Also:

* Move all devstack-specific steps out of the base image creation
* Build the static assets just for the devstack image
* Turn on `DEBUG` for the devstack image (since otherwise no static or media files will be served)
* Build, tag, and upload the new devstack image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial as app
+FROM ubuntu:xenial as discovery
 
 # System requirements.
 RUN apt-get update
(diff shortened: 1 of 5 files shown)


devstack.yml
```diff
@@ -38,6 +38,7 @@ DATABASES:
         PASSWORD: password
         PORT: 3306
         USER: discov001
+DEBUG: true
```
robrap, line 42: > Turn on DEBUG for the devstack image (since otherwise no static or media files will be served)

This would be very non-obvious (at least to me), so could this be added as a comment?

robrap (approved):
See single comment.

devstack.yml
jmbowman: Added.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial as app
```
jinder1s: app seems to be the name given to base image(non devstack and new relic) in other services: https://github.com/edx/enterprise-catalog/blob/kdmccormick/spike/Dockerfile#L1

jinder1s (approved):

PR #2723, opened 2020-09-01
ARCHBOM-1496: Changing container names

jinder1s:
Container names are now postfixed with .local.edx.org instead of prefixed with edx.devstack.

This takes advantage of the fact *.local.edx.org resolves to 127.0.0.1.


jinder1s:
Closigng because we are no longer changing container name for enterprise-catalog decentralized devstack prototype

PR #2724, opened 2020-09-02
[BD-20][RACCOON GANG][NBorovenskiy] Haystack replacement to es-dsl. Upgrade to ElasticSearch 7

NikolayBorovenskiy:
**Description:** 
This PR replaces the django-haystack and drf-haystack libraries with elasticsearch-dsl, django-elasticsearch-dsl, django-elasticsearch-dsl-drf. Also migrates Elasticserch to version 7.

**Reviewers:**
- @dianakhuang 
- @mikix 
- @robrap 

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] Documentation in source code updated
- [ ] All related Confluence documentation is updated (including deployment documentation)
- [ ] Commits are squashed

**Post merge:**
- [x] Delete working branch


openedx-webhooks:
Thanks for the pull request, @NikolayBorovenskiy! I've created [BLENDED-574](https://openedx.atlassian.net/browse/BLENDED-574) to keep track of it in Jira.  (The original issue OSPR-4938 has been deleted.)   More details are on the [BD-20](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1596162490/BD-19+BD-20+Elasticsearch+Haystack+Upgrades) project page.

When this pull request is ready, tag your edX technical lead.

natabene:
@NikolayBorovenskiy Hello, is this an intentional PR to the Open edX platform, or has it been created by mistake?

NikolayBorovenskiy:
> @NikolayBorovenskiy Hello, is this an intentional PR to the Open edX platform, or has it been created by mistake?

@natabene hello,
 It's not a mistake. We are collaborating with EDX under ES1 to ES7 migration. I think that the guys on the `Reviewers:` list understand what is happening and what to do next.

natabene:
@NikolayBorovenskiy Got it. Should you open more PRs for this line of work in the future, please preface your PR title with [BD-20]. Then it will be processed automatically.

NikolayBorovenskiy:
> @NikolayBorovenskiy Got it. Should you open more PRs for this line of work in the future, please preface your PR title with [BD-20]. Then it will be processed automatically.

Got you. My mistake. I'll do that now. Thanks.

mikix (approved):
Carrying over my approval from the WIP version of this PR.

mikix:
Obviously, actually merging this will have to wait until we have an upgrade plan in place. But this is great!

feanil:
@NikolayBorovenskiy can you update the branch to fix the test failures? It looks like mostly we need to update query counts because there is one more new query?  And it looks like there are also some upstream changes that conflict as well, can you update for that?

course_discovery/settings/base.py
```diff
@@ -440,102 +443,35 @@
...
+ELASTICSEARCH_DSL = {
+    'default': {'hosts': '127.0.0.1:9200'}
 }
+ELASTICSEARCH_INDEX_NAMES = {
+    'course_discovery.apps.course_metadata.search_indexes.documents.course': 'course',
+    'course_discovery.apps.course_metadata.search_indexes.documents.course_run': 'course_run',
+    'course_discovery.apps.course_metadata.search_indexes.documents.person': 'person',
+    'course_discovery.apps.course_metadata.search_indexes.documents.program': 'program',
+}
+ELASTICSEARCH_DSL_INDEX_SETTINGS = {'number_of_shards': 1, 'number_of_replicas': 1}
```
feanil, line 463: Elasticsearch's default number of shards is `5`, why set it to 1 here?

course_discovery/settings/production.py
```diff
@@ -62,13 +62,10 @@
...
-    'URL': ELASTICSEARCH_URL,
-    'INDEX_NAME': ELASTICSEARCH_INDEX_NAME,
-    'KWARGS': {
-        'verify_certs': True,
-        'ca_certs': certifi.where(),
-    },
+# To attach certifi elasticsearch host you should  do pip install certifi
+# should do the trick. elasticsearch-py will automatically look it up.
+ELASTICSEARCH_DSL['default'].update({
+    'hosts': ELASTICSEARCH_URL,
```
feanil: To make cutover easier, can you pull this data from `ELASTICSEARCH_CLUSTER_URL` instead?  This will be a new variable so that if(like for edx.org) the config is auto updating, we don't break the old machines when updating the new machines.

Please update any relevant docs as well.

README.rst
```diff
@@ -38,6 +38,46 @@ run:
...
+
+.. code-block:: shell
+
+    $ ./manage.py search_index --populate [--models [app[.model] app[.model] ...]] [--parallel]
+
+Recreate and repopulate the indices:
+
+.. code-block:: shell
+
+    $ ./manage.py search_index --rebuild [-f] [--models [app[.model] app[.model] ...]] [--parallel]
```
feanil, line 74: @mikix are there management commands that re-build the ES index regularly that need to be updated as a part of the upgrade process?

mikix: I don't believe so. The `search_index` command is new, as a low-level toolbox. But there's still the same `update_index` command.

course_discovery/settings/base.py
NikolayBorovenskiy: Hi @feanil you are right that it was 5, but as far as I know, it is not so.
In es7, this approach was abandoned when the default is 5 shards per index. In short, because of the overhead. Here you can read it
https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#:~:text=Index%20creation%20no%20longer%20defaults,now%20one%20shard%20per%20index.
https://stackoverflow.com/questions/60929071/elasticsearch-7-5-2-cluster-is-creating-only-one-shard-for-an-index-by-default#:~:text=2%20Answers&text=Earlier%20Elasticsearch%20had%20default%20number,seeing%20just%201%20primary%20shard.

In any case, there is no hard binding here and we can use `ELASTICSEARCH_DSL_INDEX_SETTINGS` to set the amount that we need. It's just a setting, and the setting can and should be changed. We can also set the number of shards and replicas for each index separately. This PR also has such a mechanism.

Why did I even set this setting, because the default is the same? The answer is simple, to quote the `Zen of Python` - *explicit is better than implicit*.

README.rst
NikolayBorovenskiy: Yes, Michael is right here. The fact is that `update_index` is the key and main command for `django-haystack`, but we switched to `elasticsearch-dsl`, where the main command is `search_index`, now it plays an important role. But at the same time, `update_index` with the **sanity check** feature remained. We have implemented it from zero in order to keep the old interface as much as possible and not cause surprise and annoyance.

course_discovery/settings/production.py
NikolayBorovenskiy: If I got you right here it is the commit about it https://github.com/edx/course-discovery/pull/2724/commits/ef714b1e7e0d9c7165c7e6ab1fc433d9c69fb03f

feanil:
@NikolayBorovenskiy looks like there is one more test failure to fix and then we should be good to merge this and try to ship it to stage.

NikolayBorovenskiy:
@feanil looks like we are pretty good and it's time to move on another level to begin test it out.

feanil:
Sounds good, we'll merge to deploy to stage today.

PR #2728, opened 2020-09-09
WS-1293: Update collaborator endpoint logging to include the POST/PAT…

reillz10:
…CH body.

### DIFF
--- a/course_discovery/apps/api/v1/views/collaborators.py
+++ b/course_discovery/apps/api/v1/views/collaborators.py
@@ -23,15 +23,13 @@ class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
     pagination_class = CursorPagination
 
     def create(self, request, *args, **kwargs):
-        # logging to help debug missing collaborator images
-        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
+        logger.info('The raw collaborator data coming from the publisher POST is {}.'.format(request.data))
 
         return super().create(request, *args, **kwargs)
 
     def update(self, request, *args, **kwargs):
-        # logging to help debug missing collaborator images
-        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
-        
+        logger.info('The raw collaborator data coming from the publisher PATCH is {}.'.format(request.data))
+
         return super().update(request, *args, **kwargs)  # pylint: disable=no-member
 
     def partial_update(self, request, *args, **kwargs):
--- a/course_discovery/apps/api/v1/views/collaborators.py
+++ b/course_discovery/apps/api/v1/views/collaborators.py
@@ -23,9 +23,15 @@ class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
     pagination_class = CursorPagination
 
     def create(self, request, *args, **kwargs):
+        # logging to help debug missing collaborator images
+        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
+
         return super().create(request, *args, **kwargs)
 
     def update(self, request, *args, **kwargs):
+        # logging to help debug missing collaborator images
+        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -23,9 +23,15 @@ class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
...
 
     def create(self, request, *args, **kwargs):
+        # logging to help debug missing collaborator images
+        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
+
         return super().create(request, *args, **kwargs)
 
     def update(self, request, *args, **kwargs):
+        # logging to help debug missing collaborator images
+        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
```
jlajoie: nit: Might be worth differentiating the log message here between update/create for easier splunk querying

albemarle (approved):

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -23,9 +23,15 @@ class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
     pagination_class = CursorPagination
 
     def create(self, request, *args, **kwargs):
+        # logging to help debug missing collaborator images
+        logger.info('The raw collaborator data coming from publisher is {}.'.format(request.data))
+
         return super().create(request, *args, **kwargs)
 
     def update(self, request, *args, **kwargs):
+        # logging to help debug missing collaborator images
```
jlajoie: nit: Not sure if this comment really provides any extra value, could probably just drop it since it just says what the next line does which is already pretty clear.

Dillon-Dumesnil (approved):

course_discovery/apps/api/v1/views/collaborators.py
```diff
@@ -23,9 +23,15 @@ class CollaboratorViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
     pagination_class = CursorPagination
 
     def create(self, request, *args, **kwargs):
```
jlajoie, line 25: nit: Know you didn't make this in this PR, but if you're just overriding then calling super you don't need to override the method at all, just something to keep in mind when removing the logging.

jlajoie (approved):
This is fine to merge in, but I'm not sure if this will get you closer to the problem you're trying to diagnose. I'm missing some context here, but from what it looks like, you're expecting collaborators to have images and currently they sometimes don't. If it's required, should that not just be a DB/Model/Serializer constraint so it's never possible for it to be null on an update/create?

Especially since we aren't doing any custom logic here, we just sort of put into the DB the request body we receive. I'd say it's pretty safe to assume (if there aren't external deletions/modifications of collaborators outside of this endpoint) that if images are being dropped, it's just because we are performing a create/update without an image.

If there aren't historical tables, adding them to the collaborators could be helpful since that'll track more than just the API usage which can modify the DB, and looking through the admin logs for this model could be beneficial as well if you haven't checked that yet.

reillz10:
> This is fine to merge in, but I'm not sure if this will get you closer to the problem you're trying to diagnose. I'm missing some context here, but from what it looks like, you're expecting collaborators to have images and currently they sometimes don't. If it's required, should that not just be a DB/Model/Serializer constraint so it's never possible for it to be null on an update/create?
> 
> Especially since we aren't doing any custom logic here, we just sort of put into the DB the request body we receive. I'd say it's pretty safe to assume (if there aren't external deletions/modifications of collaborators outside of this endpoint) that if images are being dropped, it's just because we are performing a create/update without an image.
> 
> If there aren't historical tables, adding them to the collaborators could be helpful since that'll track more than just the API usage which can modify the DB, and looking through the admin logs for this model could be beneficial as well if you haven't checked that yet.

That is a great point @jlajoie. I assumed (incorrectly) the the image was a required field, but it turns out that it wasn't required in the serializer. I have updated this PR to include that logic as well. Thanks for the tip!

PR #2729, opened 2020-09-09
Make sure fixture-created course runs get program type info

matthugs:


course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -214,10 +214,21 @@ def load_fixture(self, fixture_text, partner):
...
+                           ' No matching ProgramType in fixture')
                     logger.warning(msg.format(
                         program_type=obj.object.type_id,
-                        program=obj.object.id
+                        program=obj.object.id,
+                    ))
+                    raise
+            else if isinstance(obj.object, CourseRun) and obj.object.expected_program_type:
+                try:
+                    obj.object.expected_program_type = program_type_map[obj.object.expected_program_type][0]
```
Dillon-Dumesnil: I'm not going to claim great understanding of this code, but these lines feel a little odd to me. We check that the object already has an expected_program_type defined and then using that expected_program_type, we assign a new one? Would it not return the same program type? Or is there perhaps a remapping that is happening here?

No actual issues, just wanted to call this out and make sure this is all the desired functionality.

Dillon-Dumesnil (approved):
Looks good to me. Are there tests we can add to check for this new case for the management command?

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
matthugs: This is the desired functionality.

We play the same game elsewhere in this file with program_types attached directly to programs. (We also do this for seat types.) This is remapping, yes.

course_discovery/apps/course_metadata/management/commands/tests/test_load_program_fixture.py
```diff
@@ -221,6 +227,36 @@ def test_update_existing_program_type(self):
         self.assertEqual(len(stored_seat_types), 1)
         self.assertEqual(stored_seat_types[0].name, self.seat_type_verified.name)
 
+    @responses.activate
+    def test_remapping_courserun_programtype(self):
```
schenedx, line 231: #NIT, do we also need a test case to test the code block of the exception case `Failed to assign program type to course run`?

PR #2733, opened 2020-09-16
Unpin django-storages and upgrade packages

cpennington:


Dillon-Dumesnil (commented):
Would you mind updating this PR to be more directed to specifically the upgrade you want to make (django-storages)? We are seeing some weirdness with local `make upgrade` right now vs the jenkins job for python requirements and want to be careful until we are able to investigate. A ticket has already been created to look into this.

Example of weirdness would be ton look at `cffi` in local.txt that is actually being downgraded.

cpennington:
@Dillon-Dumesnil: Done.

Dillon-Dumesnil (approved):
Thanks for updating! Looks great

PR #2745, opened 2020-09-22
Retry any data loader that fails up to 3 times

cpennington:


course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -15,6 +15,7 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
         PAGE_SIZE (int): Number of items to load per API call
     """
 
+    LOADER_MAX_RETRY = 3
```
Dillon-Dumesnil, line 18: So this adds a default of 3 times to all of our data loaders, but the Ecommerce data loader then overloads it to be only 2. If we are seeing the majority of our issues in the ecom data loader, wouldn't we want that one to have the higher number?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -316,27 +315,9 @@ def __init__(self, partner, api_url, max_workers=None, is_threadsafe=False, **kw
...
-                    attempt_count=attempt_count,
-                    max=EcommerceApiDataLoader.LOADER_MAX_RETRY
-                ))
-
-            logger.info('Refreshing ecommerce data from %s...', self.partner.ecommerce_api_url)
-            self._load_ecommerce_data()
-
-            if self.processing_failure_occurred:  # pragma: no cover
-                logger.warning('Processing failure occurred caused by an exception on at least on of the threads, '
-                               'blocking deletes.')
```
Dillon-Dumesnil, line 335: So this used to try and avoid deleting entitlements if there was an exception. The changes are a little hard to follow, but is that behavior still maintained? I think so because where we would set a processing failure, we will now raise an exception after the retries, but I just want to confirm.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -510,19 +491,8 @@ def _delete_entitlements(self):
         entitlements_to_delete.delete()
 
     def _check_future_and_process(self, future, process_fn, request_fn, page):
```
Dillon-Dumesnil: Might be able to just get rid of this function if it is now reduced down to 2 lines. I think I only wrote this because of all the OSError catching.

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -714,16 +682,13 @@ def update_entitlement(self, body):
...
-                'Seat type {seat_type} is not compatible with course type {course_type} for course {uuid}'.format(
-                    seat_type=mode.slug, course_type=course.type.slug, uuid=course_uuid,
-                )
+            msg = 'Seat type {seat_type} is not compatible with course type {course_type} for course {uuid}'.format(
+                seat_type=mode.slug, course_type=course.type.slug, uuid=course_uuid,
             )
-            self.processing_failure_occurred = True
-            return None
+            logger.warning(msg)
+            raise Exception(msg)
```
Dillon-Dumesnil: Will these raising Exceptions now cause early exits to happen? In the past, setting the processing_failure allowed the remaining content to still be refreshed.

Or is the idea just to early exit here on an Exception and then let the general retry code catch the Exception and redo the whole data loader?

I wonder how that will play for errors that need manual intervention to fix. It will just fail the number of times of Max retry I suppose. Just thinking out loud.

Dillon-Dumesnil (approved):
Nothing blocking, but I do have some questions about the functionality. Mostly sanity checking stuff.

course_discovery/apps/course_metadata/data_loaders/__init__.py
cpennington: I wanted to respect the existing ecommerce data loader. I don't actually know where we're seeing the most issues, just that we saw an issue w/ ecommerce most recently.

course_discovery/apps/course_metadata/data_loaders/api.py
cpennington: Yes, this is maintained. Everywhere that `self.processing_failure_occurred` used to be set now raises an exception. Those `Exception`s are caught by the code outside `ingest`, so the `delete_entitlements` call will never happen if there's an exception.

course_discovery/apps/course_metadata/data_loaders/api.py
cpennington: Yeah, fair.

course_discovery/apps/course_metadata/data_loaders/api.py
cpennington: Oh, I see. I misunderstood the design intent.

Just to clarify: in the existing code, if things get updated, and then the whole process retries, it re-fetches the things that it just recently updated, right?

Dillon-Dumesnil: Yes, I believe that is true. I remember seeing logs where the course run type failure would show up in the middle, but then the remaining seats/entitlements/enrollments would still be processed.

cpennington: Ok, so the only advantage to the current code vs the new setup is in the case where the whole thing retries and doesn't succeed, in which case we'll at least have a partial update.

Let me see if there's a clean way to still use exceptions on any given task being submitted to the ThreadPool without making the whole ThreadPool shut down on the first exception.

cpennington: Findings (in the new code):
1. if `is_threadsafe`, then an exception inside the `self._load_*_data` functions each execute individually, and an exception in one won't break others.
2. if not `is_threadsafe`, then an exception inside `self._request_*` will be raised when `.results()` is called inside `_check_future_and_process`, which will halt processing of the remaining futures.
3. Because the futures are processed in order, we won't process them as fast as the results come back from ecommerce. This could be improved by using https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed to receive the results as they are ready.
4. The new errors being added inside `update_seat` stop processing on that seat, exactly like the old code used to. Likewise inside `update_entitlement`. Those methods are called by the `_process_course_runs` and `_process_entitlements` methods, which would be interrupted by the raised exceptions. This is a change from the prior code.

Given that, I think reverting back to the previous logic that sets `self.processing_failure_occurred` is going to be the most reliable way forward.

Dillon-Dumesnil (approved):

PR #2746, opened 2020-09-23
ws-1216: Only include instructors for a program's courses' advertised…

reillz10:
… course runs

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -38,7 +38,9 @@ from course_discovery.apps.course_metadata.publishers import (
     CourseRunMarketingSitePublisher, ProgramMarketingSitePublisher
 )
 from course_discovery.apps.course_metadata.tests import factories
-from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, ImageFactory, SeatFactory, SeatTypeFactory
+from course_discovery.apps.course_metadata.tests.factories import (
+    CourseRunFactory, ImageFactory, SeatFactory, SeatTypeFactory
+)
 from course_discovery.apps.course_metadata.tests.mixins import MarketingSitePublisherTestMixin
 from course_discovery.apps.course_metadata.utils import ensure_draft_world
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1897,9 +1897,23 @@ def test_price_ranges_with_multiple_course_runs_and_none_dates(self):
         self.assertEqual(program.price_ranges, expected_price_ranges)
 
     def test_staff(self):
+        TWO_WEEKS_FROM_TODAY = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=14)
```
rgraber, line 1902: can we test that we *don't* include staff from non-active course runs? since that was the problem we were originally trying to fix i think it'd be good to be explicit
also probably want to test the empty case if there's not a test for it already (no active course runs to be found)

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1897,11 +1897,39 @@ def test_price_ranges_with_multiple_course_runs_and_none_dates(self):
...
+            end=TWO_WEEKS_FROM_TODAY,
+            status=CourseRunStatus.Published,
+            enrollment_end=TWO_WEEKS_FROM_TODAY,
+            staff=set(expected_staff)
+        )
+        SeatFactory(
+            course_run=advertised_course_run,
+            type=SeatTypeFactory.verified(),
+            upgrade_deadline=TOMORROW
+        )
```
Dillon-Dumesnil, line 1918: Out of curiosity, why was this needed? It doesn't seem to be used in the old version of the test.

Dillon-Dumesnil (approved):

course_discovery/apps/course_metadata/tests/test_models.py
reillz10: This makes the course_run qualify as an `advertised_course_run`. If it doesn't have any seats then it can't be advertised.

rgraber (approved):
LGTM

PR #2755, opened 2020-09-30
Remove journals app after Juniper deprecation

tuchfarber:


Dillon-Dumesnil (approved):
LGTM. Thanks for doing this! A purely red PR is always a beautiful thing to see.

PR #2756, opened 2020-10-01
Updates to improve performance of the Program and affiliate apis

staubina:
This PR removes the usage of a filter from the `entitlements` method. When you use filter is breaks the usage of prefetch because another queryset is generated.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1020,7 +1020,7 @@ class MinimalCourseSerializer(DynamicFieldsMixin, TimestampModelSerializer):
         # queryset passed in happens to be empty.
         queryset = queryset if queryset is not None else Course.objects.all()
 
-        return queryset.select_related('partner', 'type').prefetch_related(
+        return queryset.select_related('partner', 'type', 'canonical_course_run').prefetch_related(
             'authoring_organizations',
             Prefetch('entitlements', queryset=CourseEntitlementSerializer.prefetch_queryset()),
             Prefetch('course_runs', queryset=MinimalCourseRunSerializer.prefetch_queryset(queryset=course_runs)),
@@ -1111,6 +1111,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
             Prefetch('course_runs', queryset=CourseRunSerializer.prefetch_queryset(queryset=course_runs)),
             Prefetch('authoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
             Prefetch('sponsoring_organizations', queryset=OrganizationSerializer.prefetch_queryset(partner)),
+            Prefetch('entitlements', queryset=CourseEntitlementSerializer.prefetch_queryset()),
         )
 
     class Meta(MinimalCourseSerializer.Meta):
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -105,7 +105,7 @@ class TestProgramViewSet(SerializationMixin):
         """ Verify the endpoint returns the details for a single program. """
         program = self.create_program()
 
-        with django_assert_num_queries(FuzzyInt(56, 2)):
+        with django_assert_num_queries(FuzzyInt(57, 2)):
             response = self.assert_retrieve_success(program)
         # property does not have the right values while being indexed
         del program._course_run_weeks_to_complete
@@ -158,7 +158,7 @@ class TestProgramViewSet(SerializationMixin):
             partner=self.partner)
         # property does not have the right values while being indexed
         del program._course_run_weeks_to_complete
-        with django_assert_num_queries(FuzzyInt(44, 1)):  # travis is often 43
+        with django_assert_num_queries(FuzzyInt(40, 1)):  # travis is often 43
             response = self.assert_retrieve_success(program)
         assert response.data == self.serialize_program(program)
         assert course_list == list(program.courses.all())
@@ -167,7 +167,7 @@ class TestProgramViewSet(SerializationMixin):
         """ Verify the endpoint returns data for a program even if the program's courses have no course runs. """
         course = CourseFactory(partner=self.partner)
         program = ProgramFactory(courses=[course], partner=self.partner)
-        with django_assert_num_queries(FuzzyInt(31, 2)):
+        with django_assert_num_queries(FuzzyInt(32, 2)):
             response = self.assert_retrieve_success(program)
         assert response.data == self.serialize_program(program)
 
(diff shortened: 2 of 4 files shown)


mikix (approved):
Sure

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -105,7 +105,7 @@ def test_retrieve(self, django_assert_num_queries):
         """ Verify the endpoint returns the details for a single program. """
         program = self.create_program()
 
-        with django_assert_num_queries(FuzzyInt(56, 2)):
+        with django_assert_num_queries(FuzzyInt(60, 2)):
```
staubina: The change in the entitlements method increased queries, but more than made up for it in reduced query duplication. The filter in the previous version caused every call to ignore the prefetch done at higher levels in the code.

course_discovery/apps/api/serializers.py
```diff
@@ -680,7 +680,7 @@ class CourseEntitlementSerializer(BaseModelSerializer):
 
     @classmethod
     def prefetch_queryset(cls):
-        return CourseEntitlement.everything.all().select_related('currency', 'mode')
+        return CourseEntitlement.everything.all().select_related('course', 'currency', 'mode')
```
cpennington: I think the `all` here is irrelevant?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2249,8 +2249,15 @@ def canonical_seats(self):
...
     def entitlements(self):
-        applicable_seat_types = set(self.type.applicable_seat_types.all())
-        return CourseEntitlement.objects.filter(mode__in=applicable_seat_types, course__in=self.courses.all())
+        """
+        Property to retrieve all of the entitlements in a Program.
+        """
+        return [
+            entitlement
+            for course in self.courses.all()
+            for entitlement in course.entitlements.all()
```
cpennington, line 2261: Might be worth making a comment that this is doing list comprehension in order to not break prefetching. If you were feeling ambitious, you could also check to see if the courses and entitlements have been prefetched (but that requires using an undocumented api).

course_discovery/apps/api/v1/tests/test_views/test_programs.py
cpennington: I'm confused. If it reduced query duplication, what queries did it add? Is it indicating that there's code that isn't prefetching `course` and `course.entitlement`?

cpennington (commented):
More broadly, I wonder if there's an avenue towards better results by starting the query from the `Entitlements` class, since that's what you're trying to return. Not sure if that would end up better or not, and it may be counter to the current DRF layout.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2249,8 +2249,15 @@ def canonical_seats(self):
 
     @property
     def entitlements(self):
-        applicable_seat_types = set(self.type.applicable_seat_types.all())
-        return CourseEntitlement.objects.filter(mode__in=applicable_seat_types, course__in=self.courses.all())
+        """
+        Property to retrieve all of the entitlements in a Program.
+        """
+        return [
```
tuchfarber, line 2258: Since we're changing the return type of this function from a queryset to a list, are there any consuming functions that assume a queryset that we need to update?

tuchfarber (approved):
Assuming the answer to my question is "no", :+1:

course_discovery/apps/api/serializers.py
staubina: I was adding to the prefetch. This was preexisting so I would rather not change it. I will remove the added prefetch because I do not think it helped.

course_discovery/apps/course_metadata/models.py
staubina: Good idea on the comment.

staubina:
@cpennington I was attempting to alter the DRF structure as little as possible so I would like to avoid making the changed to the way the entitlements are retrieved for now.  I do agree that we should look at approaching these endpoints in a cleaner way. Maybe even consider a V2 sometime soon.

I think in general we need to take a deeper look at this affiliate endpoint problem. I do not like that we are using APIs to provide a data dump to an affiliate marketing service in general. 

My goal is to improve performance with as little structural alteration as possible.

course_discovery/apps/course_metadata/models.py
staubina: I will double-check. All tests seem to be functioning fine and manual tests all seem to be working also.

staubina: This change is fine because the data is just iterated over and non of the special functions of a queryset are used after being retrieved. I also only see this method being used here and by the API endpoint.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
cpennington: It looks like `ProgramSerializer` (used by the program-detail view) prefetches `course` using the `CourseSerializer` (https://github.com/edx/course-discovery/blob/dc48f8a50541c608be363abfeef0845d78fdc628/course_discovery/apps/api/serializers.py#L1629), but that `CourseSerializer` doesn't prefetch `entitlements` (and neither does `ProgramSerializer`). Perhaps that's why we're not actually seeing query reductions here?

staubina: This is odd to me because I see a substantial drop is local duplicated queries on the affiliates endpoint and a small drop on the detail endpoint for Programs. Once I clear up another test issue I will look at this.

staubina: My theory was that because we were using filter before we were performing the work we are doing in fewer querries, but because it was ignoring the prefetch cache the querries were being duplicated. 

Now the queries for a single call are higher, and this is caught in the test, but the duplicated calls drop from 130 to 90 in my local manual testing with 3 programs.

mikix (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -2249,8 +2249,19 @@ def canonical_seats(self):
 
     @property
     def entitlements(self):
-        applicable_seat_types = set(self.type.applicable_seat_types.all())
-        return CourseEntitlement.objects.filter(mode__in=applicable_seat_types, course__in=self.courses.all())
+        """
+        Property to retrieve all of the entitlements in a Program.
+
+        Warning: The choice to not use a filter method on the queryset here was deliberate. The filter
```
cpennington: Nit: This should be a comment (because it's directed at code authors), rather than a docstring (which is directed at users of the function). You could add to the docstring the information about it returning a list, rather than a querystring.

cpennington (approved):

course_discovery/apps/api/v1/tests/test_views/test_programs.py
cpennington: That makes sense. It might be worth adding a test that call the endpoint w/ 1/10/100 programs (or entitelments, or whatever things need to scale), and making sure that the number of queries is constant.

PR #2758, opened 2020-10-02
ENT-3334 | SAP SF customer is seeing courses as not open for enrollment when they should be

HammadAhmadWaqas:
added status in CourseSearchSerializer to get status on search/all endpoint.
this status will be used to check the availability of the course.


hasnain-naveed (approved):
LGTM

Dillon-Dumesnil (dismissed):
Sure

Dillon-Dumesnil:
So I just want to get some clarity on how this status can help/how it will be used to understand if this is the correct fix. Thank you!

muhammad-ammar (approved):

HammadAhmadWaqas:
@Dillon-Dumesnil actually we make decisions about the `course` using its course runs information, that information includes its `enrollment_start` `enrollment_end` `availability` as well as its `status`. if all that information is valid then we say that course is active.
Here is [PR](https://github.com/edx/edx-enterprise/pull/795) that first time tried to use `status` for deciding `active` value.

Dillon-Dumesnil (approved):
Ah right, I forgot we show stuff at the Course level and not Course Run level so you're able to aggregate information. Sounds good.

PR #2760, opened 2020-10-02
DISCO-1675: Use mysql 5.7 in docker/travis

mikix:


Makefile
```diff
@@ -1,17 +1,13 @@
 .DEFAULT_GOAL := help
 NODE_BIN=$(CURDIR)/node_modules/.bin
-TOX = ''
+TOX := tox
```
mikix, line 3: This and the change below is to make tests work even without TOXENV being defined (the empty string for TOX means that running tests fails). It made more sense to me that we'd run all environments (no TOXENV) than none (no TOX).

I ran into this during testing, when I went to run the travis container locally.

.travis/docker-compose-travis.yml
```diff
@@ -2,14 +2,11 @@ version: "2"
...
-    image: mysql:5.6.45
+    image: mysql:5.7
     container_name: db
     command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
     environment:
-      MYSQL_ROOT_PASSWORD: ""
-      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-      MYSQL_USER: "discov001"
-      MYSQL_PASSWORD: "password"
+      MYSQL_ROOT_PASSWORD: "password"
```
mikix, line 9: This is because according to mysql docs, the USER/PASSWORD/DATABASE combo grants that user full access to the named database. But does not say anything about other databases.

The tests try to make additional test databases (test_discovery_gw0 and _gw1) during test setup. For some reason, mysql 5.6 was allowing this (doesn't seem like it should have according to docs?) and 5.7 didn't. So I just switched to using the root user for the tests, since they can make new databases fine.

https://dev.mysql.com/doc/refman/5.6/en/docker-mysql-more-topics.html
https://dev.mysql.com/doc/refman/5.7/en/docker-mysql-more-topics.html

mikix (commented):
Notes below

course_discovery/settings/production.py
```diff
@@ -53,14 +47,21 @@
...
+    DB_MIGRATION_ENGINE='ENGINE',
+    DB_MIGRATION_USER='USER',
+    DB_MIGRATION_NAME='NAME',
+    DB_MIGRATION_HOST='HOST',
+    DB_MIGRATION_PORT='PORT',
 )
+for override, db_key in DB_OVERRIDES.items():
+    if override in environ:
+        DATABASES['default'][db_key] = environ.get(override)
+        DATABASES['read_replica'][db_key] = environ.get(override)
```
mikix, line 64: This and the changes above were made so that when an override variable is specified, we override both the replica and the default db.

adzuci (approved):
LGTM

.travis/docker-compose-travis.yml
```diff
@@ -44,7 +41,7 @@ services:
       DB_NAME: "discovery"
       DB_PASSWORD: "password"
       DB_PORT: "3306"
-      DB_USER: "discov001"
+      DB_USER: "root"
```
adzuci, line 44: How was the discov001 user getting created before and is there a reason to switch to root?

mikix: See the comment for the code above this

ciduarte (approved):

course_discovery/settings/base.py
```diff
@@ -115,7 +115,7 @@
 WSGI_APPLICATION = 'course_discovery.wsgi.application'
 
 # Database
-# https://docs.djangoproject.com/en/1.8/ref/settings/#databases
+# https://docs.djangoproject.com/en/2.2/ref/settings/#databases
```
Dillon-Dumesnil, line 118: :chefkiss:

course_discovery/settings/production.py
```diff
@@ -23,13 +23,7 @@
 # This may be overridden by the YAML in DISCOVERY_CFG, but it should be here as a default.
 MEDIA_STORAGE_BACKEND = {}
 
-# TODO Drop the try-except block once https://github.com/edx/configuration/pull/3549 is merged and we are using the
```
Dillon-Dumesnil, line 26: Do I even want to know how old this comment is?

Dillon-Dumesnil (approved):
Sure

PR #2761, opened 2020-10-02
Updates to improve performance of the Program and affiliate apis

staubina:


course_discovery/apps/api/serializers.py
```diff
@@ -680,7 +680,7 @@ class CourseEntitlementSerializer(BaseModelSerializer):
 
     @classmethod
     def prefetch_queryset(cls):
-        return CourseEntitlement.everything.all().select_related('currency', 'mode')
+        return CourseEntitlement.everything.all().select_related('course', 'currency', 'mode')
```
cpennington, line 683: So, this serializer doesn't look like it uses the course at all, based on the set of fields above. Where were you seeing the course being fetched?

course_discovery/apps/api/serializers.py
```diff
@@ -799,7 +799,7 @@ def prefetch_queryset(cls, queryset=None):
         # queryset passed in happens to be empty.
         queryset = queryset if queryset is not None else CourseRun.objects.all()
 
-        return queryset.select_related('course', 'type').prefetch_related(
+        return queryset.select_related('course', 'type', 'language').prefetch_related(
```
cpennington, line 802: Is language used in this serializer?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2178,20 +2178,22 @@ def course_runs(self):
         Otherwise, this method will incur many, many queries when fetching related courses and course runs.
         """
         excluded_course_run_ids = [course_run.id for course_run in self.excluded_course_runs.all()]
-
+        course_run_list = list()
```
cpennington, line 2181: Why force this to be a list here, rather than at the caller?

cpennington (commented):
A bunch of these changes are very non-obvious (as to why they would improve things), so it might be worth some comments about why you're making the choices you're making.

PR #2773, opened 2020-10-08
Switch to Ubuntu 20.04 focal

mikix:


.travis.yml
```diff
@@ -1,3 +1,4 @@
+dist: focal
```
mikix, line 1: Not technically necessary -- this travis script just installs the discovery docker images and runs everything inside them.

But nice to have all our environments be as the same if possible.

Dockerfile
```diff
@@ -1,21 +1,19 @@
-FROM ubuntu:xenial as app
+FROM ubuntu:focal as app
 
 # System requirements.
 RUN apt-get update && \
-	apt-get install -y software-properties-common && \
-	apt-add-repository -y ppa:deadsnakes/ppa && \
-	apt-get update && \
```
mikix, line 7: Now that python3.8 is built into focal, we don't need the deadsnakes repo. I could have left it in there for the next time we bump python... but I felt it was cleaner to just remove the unused code.

Dockerfile
```diff
@@ -59,7 +57,7 @@ RUN pip install -r requirements/production.txt
 # every time any bit of code is changed.
 COPY . .
 
-# Expose canoncal Discovery port
+# Expose canonical Discovery port
```
Dillon-Dumesnil, line 60: If ever there was a time for the :miketerry: emoji

Dillon-Dumesnil (approved):
Sure

PR #2791, opened 2020-10-20
WS-1414 upload program card images to discovery

rgraber:


course_discovery/apps/api/v1/views/programs.py
```diff
@@ -107,3 +112,21 @@ def list(self, request, *args, **kwargs):
...
+            # base64 encoded image - decode
+            file_format, imgstr = image_data.split(';base64,')  # format ~= data:image/X;base64,/xxxyyyzzz/
+            ext = file_format.split('/')[-1]  # guess file extension
+            image_data = ContentFile(base64.b64decode(imgstr), name=f'tmp.{ext}')
+            program.card_image.save(image_data.name, image_data)
+            msg = 'Successfully updated program card image for program {uuid}: {title}'.format(uuid=program.uuid,
+                                                                                               title=program.title)
+            return Response(msg)
+        else:
+            return Response('Bad image data in request', status=status.HTTP_400_BAD_REQUEST)
```
ciduarte, line 132: Could we write a test for this response?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2058,6 +2058,14 @@ class Program(PkSearchableMixin, TimeStampedModel):
...
     banner_image_url = models.URLField(null=True, blank=True, help_text='DEPRECATED: Use the banner image field.')
+    card_image = StdImageField(
+        upload_to=UploadToFieldNamePath(populate_from='uuid', path='media/programs/card_images'),
+        blank=True,
+        null=True,
+        variations={
+            'card': (378, 225),
+        }
+    )
     card_image_url = models.URLField(null=True, blank=True, help_text=_('Image used for discovery cards'))
```
ciduarte, line 2069: I notice we have both `card_image` (which we use to grab `card_image.url`) and `card_image_url`. Is there a plan to phase out `card_image_url` in the future?

ciduarte (approved):
Two questions, but looks good to me! 👍

course_discovery/apps/course_metadata/models.py
rgraber: Yes, we're hoping to deprecate it in the same way we've deprecated banner_image_url

rgraber: We already took it out of admin so it shouldn't be possible to change card_image_url independently of card_image except through the code, which is basically the same as banner_image_url

srwang (approved):

PR #2792, opened 2020-10-22
WS-1414 followup: update algolia index with new card image

rgraber:


AlasdairSwan (approved):
:shipit:

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -270,6 +270,8 @@ def custom_object_id(self):
 
     @property
     def product_card_image_url(self):
+        if self.card_image:
+            return getattr(self.card_image, 'url', None)
```
Dillon-Dumesnil: In your last PR, you didn't use a `getattr` to make sure the card_image had the url attribute (in the serializer). Should that code be updated to use the getattr as well or could this last line be changed to `return self.card_image.url`?

Dillon-Dumesnil (approved):
One small question, but otherwise LGTM

course_discovery/apps/course_metadata/algolia_models.py
rgraber: Hm, good question. Both were copy-paste, tbh. self.card_image.url should work here

albemarle (approved):

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -270,6 +270,8 @@ def custom_object_id(self):
 
     @property
     def product_card_image_url(self):
+        if self.card_image:
```
albemarle, line 273: Is it documented anywhere what the difference between these fields is, and why we had to do this?  If not this might be a good place.

PR #2794, opened 2020-10-22
Add exec ed type

tuchfarber:
We are creating a new course type in LMS called "Executive Education". While most of the hardcode bits were happily fixed via the Modes/Seats/Types system, these two spots still require a hard code in order for things to set up correctly. 

Without `ENTITLEMENT_MODES`, the call to ecommerce won't create an entitlement for us.

Without the `CertificateType` choice, we can't make a seat of `executive-education` type. 

Let me know if you need more context :)


Dillon-Dumesnil (commented):
Hey Matt, would you mind adding in some details about what you want this PR to accomplish or any related tickets/PRs so we can have a full picture on the change you want to make here?

tuchfarber:
Aw shucks I forgot this auto pinged you guys. Sorry about that, gimme a sec.

Dillon-Dumesnil:
No rush if you're not ready, feel free to just move this PR into a draft PR and ping us in slack when you're ready :)

tuchfarber:
Updated! Also yes, I hate putting edx.org product name stuff in here. And in ecommerce. And in LMS. Each and every line pains me.

Dillon-Dumesnil:
Great, thank you! I'll take a look at this in the morning.

Dillon-Dumesnil:
I thought when we last talked about this, executive education was going to be implemented as no-id-professional. Are you saying now that you decided to make a new LMS mode, Ecommerce Product Type (seat and entitlement) for executive education and this is the work to support that?

tuchfarber:
Yeah, that plan died for a couple of reasons (one of which was that course certs are differentiable by mode and we wanted a new course cert type). So the plan is now to make a new mode, seat, and all that. 

Related PRs are:
 - https://github.com/edx/ecommerce/pull/3221
 - https://github.com/edx/edx-platform/pull/25381

Dillon-Dumesnil:
Looking at the edx-platform PR, are you not just using the verified cert? Like lines like [this](https://github.com/edx/edx-platform/pull/25381/files#diff-364f09d5284c2c3b0f680e4992b6dfae29f34c8a3e0d74015c3a1ce425153839R302) have me a bit confused. Curious where the new mode is actually coming into play.

For context, I'm just trying to figure out what all is actually needed. Because I don't _think_ I see anything to actually support an Executive Education Product in Ecommerce, but it looks like it is simply verified under the hood. Which would lead me to think the changes here might not be necessary, but I think I might be missing some parts.

tuchfarber:
Good catch on the platform PR. I think I got a bit overzealous in my "follow MASTERS" plan. I updated it. 

My understanding, based on my conversation with @staubina, was that Ecommerce doesn't need to know much about new seat types because they are stored as strings, so they mostly accept whatever they are told from discovery/publisher. Looking through the code and testing I saw that:
1. The certificate type edited in this PR gets sent to ecommerce when you publish the course in publisher
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/utils.py#L419-L426

2. Ecommerce reads that cert type and that's what's used when it created the seat
image

3. Ecommerce sends that attribute over to LMS during fulfillment so they get enrolled in the exec-ed mode of the course. 

This is mostly new code to me though, so I'm going to pull AJ in to make sure those words are right.

Dillon-Dumesnil:
Okay, I'm happy to hear from AJ as well, but I definitely trust that you/your team has done your due diligence.

To round out the conversation, can you update me on what changes you'll be making outside of code that are also related to this? I assume the creation of the SeatType, Seat, Track, Mode, CourseRunType, CourseType. I think that might be it, but some sanity checking from your end would be very helpful too!

tuchfarber:
AJ sent me "that all looks correct to me" on Slack, so he probably won't post here. 

(ExecEd = "executive-education" in the DB)

As far as DB additions, we'll be creating
- An ExecEd `Mode`
- An ExecEd `SeatType`
- An ExecEd `Track` linking the `Seat` and `Mode`
- An ExecEd `CourseRunType` that only uses the ExecEd track 
- An ExecEd `CourseType` that only uses the ExecEd `CourseRunType` and "Entitlement Type". Possible org whitelisting. Still waiting to hear back from product.

Dillon-Dumesnil (approved):
Okay, that all sounds good to me then. Luckily, this is all internal so if we need to make changes after the fact, that can be supported.

One thing I will say, when you are making the tracks, set the org-whitelist to only include edX to start. If you don't set it at creation time, it will show up for all of our partners and that will likely lead to a lot of questions for our PCs from our partners. Easier to initially put edX and reverse that to the desired state later.

PR #2813, opened 2020-10-29
EDUCATOR-5376: Remove individual endorsements from program form in django admin

jansenk:
[EDUCATOR-5376](https://openedx.atlassian.net/browse/EDUCATOR-5376)

The individual endorsements have been removed from the Django Admin.

Screen Shot 2020-10-29 at 11 44 31 AM

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -198,14 +198,15 @@ class ProgramAdmin(admin.ModelAdmin):
     exclude = ('card_image_url',)
 
     # ordering the field display on admin page.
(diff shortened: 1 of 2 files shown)


Dillon-Dumesnil:
Hey @jansenk , thanks for the Pull Request. Before approving the code, I'm hoping to get some more insight on this change. Looking at the ticket, it looks like this change is designed to help out edX employees, but I want to be sure that we aren't making any decisions that will adversely affect the Open edX community. My thinking here is if the Open edX community uses Individual Endorsements, we are now removing their ability to easily add and manage them.

PR #2823, opened 2020-11-03
Avoid bad string decode call

jmyatt:
https://openedx.atlassian.net/browse/DISCO-1701

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -948,6 +948,11 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
                 log_capture.check(
                     (
                         course_logger.name,
+                        'INFO',
+                        'DISCO-1701 Exception class: IntegrityError',
+                    ),
+                    (
+                        course_logger.name,
                         'ERROR',
                         'Failed to set data: Error',
                     )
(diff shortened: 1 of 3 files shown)


Dillon-Dumesnil (commented):
Going to wait to approve until we figure out what's happening with the tests.
And looking at the ticket, perhaps the issue is that this one course is somehow returning a list instead of a string?

jmyatt:
the test failures are related to the code change (bytestring / unicode string conversion) so before I change the test I need to make sure that's the right thing to do.  The list being returned is in the Exception's `content` attr, but a similar error occurs with that attr contains a unicode string instead of a list (in that case the error is `'str' object has no attribute 'decode'`)

mikix:
Is it possible that python3 guarantees `e.content` is a unicode string if it exists? If so, we can simply drop the `decode` call.

jmyatt:
that's what I tried first, but some tests failed due to the fact that they explicitly set bytestrings in that `content` attr.  So I took this approach to cover the case where we might actually get a bytestring in runtime.  But maybe the test should just be changed?  

the test lines I'm referring to:
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_courses.py#L932
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py#L618

mikix:
What exception class is hitting this code path? (i.e. what library is the exception type that provides `content`? I looked real quick at django and drf, but didn't see exceptions like that)

jmyatt:
yes good point - it's unclear which Exception class this is since the AttributeError on the decode call is happening in the `except` block for the original Exception.  I'll change this PR to temporarily log that information, then will be able to follow up with another change to clean it up.

mikix (approved):

PR #2833, opened 2020-11-12
Revert "Switch to es7"

mikix:
Reverts edx/course-discovery#2830

Found a few more possible issues. Might need another round of fixes.


mavidser:
Hi @mikix - checking if there's an updated timeline to upgrade discovery to ES7 yet? It seems to perhaps be the last service which is dependent on ES1.5, [according to the devstack configuration](https://github.com/edx/devstack/blob/33c792693d14188ba43673ffcdf2877e1259e725/docker-compose.yml#L201).

---

Context: Setting up openedx from `configuration` [is broken on master right now](https://groups.google.com/g/open-edx-btr-notifications/c/Qy68eTORzto), because forum expects ES7 now, [while configuration only installs 1.5.2](https://github.com/edx/configuration/blob/08b7b10ce38d4ac668059a109d748c90a3f78e54/playbooks/roles/elasticsearch/defaults/main.yml#L16). One way to fix that would be to pin forum to a pre-ES7 version (edx/configuration#6093), but I also thought to check here to see if we'd just be able to directly upgrade configuration to use ES7 anytime soon.

mikix:
Hello @mavidser - next week we might try again after having fixed some issues found in the latest run through.

PR #2837, opened 2020-11-19
DISCO-1695: Order serialized runs inside a course

mikix:
When we serialize a course, we used to just use the default ordering (run creation order) for its runs. But this commit changes that to an explicit start date ordering.

This matches up with what a human would expect the ordering to be.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -9,6 +9,7 @@ import responses
 from django.conf import settings
 from django.db import IntegrityError
 from django.db.models.functions import Lower
+from django.db.models.query import Prefetch
 from rest_framework.reverse import reverse
 from testfixtures import LogCapture
 
@@ -49,6 +50,17 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
     def mock_ecommerce_publication(self):
         url = f'{self.course.partner.ecommerce_api_url}publication/'
         responses.add(responses.POST, url, json={}, status=200)
+
+    def serialize_course(self, course, many=False, format=None, extra_context=None):  # pylint: disable=redefined-builtin
+        # The course view sorts courses by key and their course runs by start date. Let's emulate that here.
+        queryset = Course.everything.prefetch_related(
+            Prefetch('course_runs', queryset=CourseRun.everything.order_by('start', 'id')),
+        )
+        if many:
+            obj = queryset.filter(id__in=[c.id for c in course]).order_by(Lower('key'))
+        else:
+            obj = queryset.get(id=course.id)
+        return super().serialize_course(obj, many=many, format=format, extra_context=extra_context)
 
     def test_get(self):
         """ Verify the endpoint returns the details for a single course. """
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1250,7 +1253,7 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None):
 
     def get_course_runs(self, course):
         return CourseRunSerializer(
-            course.course_runs,
+            course.course_runs.order_by('start'),
```
mikix: CatalogCourseSerializer

course_discovery/apps/api/serializers.py
```diff
@@ -2054,7 +2057,7 @@ def course_run_detail(request, detail_fields, course_run):
 
     def get_course_runs(self, result):
         request = self.context['request']
-        course_runs = result.object.course_runs.all()
+        course_runs = result.object.course_runs.order_by('start')
```
mikix: CourseSearchSerializer

course_discovery/apps/api/serializers.py
```diff
@@ -1193,12 +1193,15 @@ def get_advertised_course_run_uuid(self, course):
             return course.advertised_course_run.uuid
         return None
 
+    def ordered_course_runs(self, course):
+        return course.course_runs.order_by('start')
+
```
mikix: CourseWithProgramsSerializer (what Publisher sees)

Dillon-Dumesnil (approved):

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -38,10 +38,10 @@ def get_queryset(self):
         # which happens when the queryset is stored in a class property.
         partner = self.request.site.partner
         q = self.request.query_params.get('q')
-        queryset = None
+        queryset = Program.objects.filter(partner=partner).order_by('id')
 
         if q:
-            queryset = Program.search(q)
+            queryset = Program.search(q, queryset=queryset)
```
mikix, line 44: This was unrelated to my main fix. While running tests, I noticed that the program test `test_filter_by_types` (at least) was flaky. Afaict, it's because the view never applied any ordering by default, and Programs don't have model ordering. So the returned order was technically undefined.

Which seems bad for testing and for pagination. So I set up a default ordering by id.

(This also fixes a bug where a `q=` parameter would have skipped partner-partitioning and searched over all partners.)

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -50,6 +51,17 @@ def mock_ecommerce_publication(self):
         url = f'{self.course.partner.ecommerce_api_url}publication/'
         responses.add(responses.POST, url, json={}, status=200)
 
+    def serialize_course(self, course, many=False, **kwargs):
+        # The course view sorts courses by key and their course runs by start date. Let's emulate that here.
```
mikix: So a quick note on my overall approach here. I wanted to sort the serialized course runs inside a course. 

There are four natural places to do that:
- On the django model by default
  - I didn't want to do it on the model level, since (A) that felt heavy-weight and (B) I really only care about sorting a *course's* runs, not all runs relative to each other.
- In the serializer get_course_runs
  - If the previous solution was sorting at the earliest possible moment, this is the last. And it's too late, it's already after the queryset has been resolved and the prefetched results are sitting there. So we don't want to have to requery the db.
- In the serializer prefetch_queryset
  - This would be fine, except all serializers that have course runs lists would end up getting this ordering (even if the runs are across courses). But the greater sin in my mind was that all the prefetch_queryset code currently is strictly performance optimization. This would be sullying that code with actual business logic.
- In the view
  - This felt appropriate, given the above and given that the view does all the other queryset tranformations.

OK. So now I'm sorting in the view. But none of the test code expects that. They just look up a course, serialize it directly, and compare with what the view returns. So I needed to kind of fake what the view does a bit, so the tests still work.

As you can see, several of the tests were already doing course-sorting themselves by key. I condensed all that into this one view-specific-override of the serialize_course utility method.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -226,14 +256,13 @@ def test_list(self):
         self.assertEqual(response.status_code, 200)
         self.assertListEqual(
             response.data['results'],
-            self.serialize_course(Course.objects.all().order_by(Lower('key')), many=True)
+            self.serialize_course(Course.objects.all(), many=True),
```
Dillon-Dumesnil: Shouldn't need the comma at the end here. I think the function is only expecting two arguments.

course_discovery/apps/api/v1/views/programs.py
Dillon-Dumesnil: Should this order by id be set on the model level rather than the view? Or am I misunderstanding this?

Dillon-Dumesnil (approved):
Couple small questions, but yeah, I mean, it all looks fine to me.

Were you able to do any load testing? Just curious how much of an effect this sorting could have on the endpoint.

course_discovery/apps/api/v1/views/programs.py
mikix: I think that's maybe a philosophical question. Should every model have a `ordering = ['id']` by default? Maybe, but there's a theoretical performance penalty.

In this case, the view is the one that cares about repeatable order. I'm not sure if other code does.

PR #2843, opened 2020-11-24
DISCO-1695: Sort course runs in courses in programs too

mikix:
Prevoiusly, we only sorted them inside of the course endpoint.


Dillon-Dumesnil (approved):
Works for me

PR #2845, opened 2020-11-30
Add org logo to minimal program serializer

tuchfarber:
Similar to https://github.com/edx/course-discovery/pull/2844 we need
to store this value when we scan so we can use a cached version in
services like credentials.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -361,24 +361,25 @@ class PositionSerializer(BaseModelSerializer):
 
 
 class MinimalOrganizationSerializer(BaseModelSerializer):
-    class Meta:
-        model = Organization
-        fields = ('uuid', 'key', 'name', 'auto_generate_course_run_keys',)
-        read_only_fields = ('auto_generate_course_run_keys',)
-
-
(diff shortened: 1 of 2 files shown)


staubina (approved):
Overall this PR makes sense to me. We should just make sure it does not bloat the number of SQL queries.

tuchfarber:
Verified that due to existing prefetching we stay at 16 queries (checked via DDT).

course_discovery/apps/api/serializers.py
```diff
@@ -361,9 +361,17 @@ def get_organization_logo_image_url(self, obj):
 
 
 class MinimalOrganizationSerializer(BaseModelSerializer):
+    certificate_logo_image_url = serializers.SerializerMethodField()
+
+    def get_certificate_logo_image_url(self, obj):
+        image = getattr(obj, 'certificate_logo_image', None)
+        if image:
+            return image.url
+        return None
```
Dillon-Dumesnil, line 370: Nit: `return image and image.url`

Dillon-Dumesnil (approved):

course_discovery/apps/api/serializers.py
tuchfarber: Oh yeah, I just copied and pasted this from it's old location. I'll clean it up.

PR #2864, opened 2021-01-04
Rsgraber/ws 1468 fix drag and drop

rgraber:
This makes the following changes Search Default Configuration admin form:
1. Courses are now sortable
2. Adding a course object should no longer add duplicates
3. Adding a program should no longer add duplicates UNLESS we have two active programs with the same name, which is a lot harder to solve

This does not address the issue of courses/programs sometimes being added to the middle of the list instead of the end. Still hunting that one.


srwang (approved):
will there be tests?

rgraber:
> will there be tests?

The issue is the only thing that makes sense to test is the javascript (otherwise we're just testing that Django fields work), and course-discovery is not set up for js testing

Dillon-Dumesnil (approved):

PR #2867, opened 2021-01-04
update course skills on course metadata field update

muhammad-ammar:
**Description:** If a skills relavant course field is updated than fire signal so that a background task in taxonomy update the course skills. Currently we are only interested in `full_description` course field.

**JIRA:** [ENT-3758](https://openedx.atlassian.net/browse/ENT-3758)

**Blocked By:** https://github.com/edx/taxonomy-connector/pull/34

### DIFF
--- a/course_discovery/apps/api/utils.py
+++ b/course_discovery/apps/api/utils.py
@@ -94,9 +94,10 @@ def reviewable_data_has_changed(obj, new_key_vals, exempt_fields=None):
         exempt_fields (list): List of field names where a change does not affect review status
 
     Returns:
-        bool for whether data for any reviewable fields has changed
+        list of changed field names
     """
     changed = False
+    changed_fields = []
     exempt_fields = exempt_fields or []
     for key, new_value in [x for x in new_key_vals if x[0] not in exempt_fields]:
         original_value = getattr(obj, key, None)
@@ -114,7 +115,13 @@ def reviewable_data_has_changed(obj, new_key_vals, exempt_fields=None):
                         changed = True
         elif new_value != original_value:
             changed = True
-    return changed
+        else:
+            changed = False
+
+        if changed:
+            changed_fields.append(key)
+
+    return changed_fields
 
 
 def conditional_decorator(condition, decorator):
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -586,3 +586,47 @@ EMSI_API_ACCESS_TOKEN_URL = 'https://auth.emsicloud.com/connect/token'
 EMSI_API_BASE_URL = 'https://emsiservices.com'
 EMSI_CLIENT_ID = ''
 EMSI_CLIENT_SECRET = ''
+
+
+################################### BEGIN CELERY ###################################
+
+# Message configuration
+CELERY_SERIALIZER = 'json'
+CELERY_RESULT_SERIALIZER = 'json'
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_COMPRESSION = 'gzip'
+CELERY_MESSAGE_COMPRESSION = 'gzip'
+
+# Events configuration
+CELERY_TRACK_STARTED = True
+CELERY_SEND_EVENTS = True
+CELERY_SEND_SENT_EVENT = True
+
+# Prevent Celery from removing handlers on the root logger. Allows setting custom logging handlers.
+# See http://celery.readthedocs.io/en/latest/userguide/configuration.html#worker-hijack-root-logger.
+CELERYD_HIJACK_ROOT_LOGGER = False
+
+CELERY_DEFAULT_EXCHANGE = 'discovery'
+CELERY_DEFAULT_ROUTING_KEY = 'discovery'
+CELERY_DEFAULT_QUEUE = 'discovery.default'
+
+# Celery Broker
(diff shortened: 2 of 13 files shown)


course_discovery/settings/base.py
```diff
@@ -651,3 +651,64 @@
...
+# Celery Broker
+# These settings need not be set if CELERY_ALWAYS_EAGER == True, like in Standalone.
+# Devstack overrides these in its docker-compose.yml.
+# Production environments can override these to be whatever they want.
+CELERY_BROKER_TRANSPORT = os.environ.get("CELERY_BROKER_TRANSPORT", "redis")
+CELERY_BROKER_HOSTNAME = os.environ.get("CELERY_BROKER_HOSTNAME", "edx.devstack.redis:6379")
+CELERY_BROKER_VHOST = os.environ.get("CELERY_BROKER_VHOST", "10")
+CELERY_BROKER_USER = os.environ.get("CELERY_BROKER_USER", "")
+CELERY_BROKER_PASSWORD = os.environ.get("CELERY_BROKER_PASSWORD", "password")
+CELERY_BROKER_URL = "{0}://{1}:{2}@{3}/{4}".format(
```
iamsobanjaved: You should not add `CELERY_` prefix here until you have added `namespace='CELERY'` in `config_from_object` call. You should remove `CELERY_` prefix from here. This setting name should be `BROKER_URL`

https://docs.celeryproject.org/en/stable/userguide/configuration.html#broker-url

course_discovery/settings/production.py
```diff
@@ -87,3 +87,11 @@
 
 # Have images and such that we upload be publicly readable
 AWS_DEFAULT_ACL = 'public-read'
+
+CELERY_BROKER_URL = "{0}://{1}:{2}@{3}/{4}".format(
```
iamsobanjaved: Same here, removal of prefix here too.

course_discovery/settings/test.py
```diff
@@ -64,3 +66,15 @@
...
+
+################################### BEGIN CELERY ###################################
+
+CELERY_ALWAYS_EAGER = True
+
+CELERY_IGNORE_RESULT = True
+
+CELERY_RESULT_BACKEND = 'file://{}'.format(tempfile.TemporaryDirectory())
+
+CELERY_BROKER_URL = 'memory://localhost/'
```
iamsobanjaved, line 78: Removal of prefix here too

muhammad-ammar:
@iamsobanjaved I would prefer to use `CELERY_` prefix in `config_from_object`. Any concerns?

iamsobanjaved:
> @iamsobanjaved I would prefer to use `CELERY_` prefix in `config_from_object`. Any concerns?

Nope, then you should mention it there and make sure every celery setting is prefixed. 

> \# - namespace='CELERY' means all celery-related configuration keys
> \#   should have a CELERY_ prefix.
> app.config_from_object('django.conf:settings', namespace='CELERY')

https://docs.celeryproject.org/en/stable/django/first-steps-with-django.html#django-first-steps

course_discovery/apps/api/utils.py
```diff
@@ -114,7 +116,11 @@ def reviewable_data_has_changed(obj, new_key_vals, exempt_fields=None):
                         changed = True
         elif new_value != original_value:
             changed = True
-    return changed
+
+        if changed:
```
moconnell1453, line 121: I don't think this will work properly because **changed** never gets set back to False on each iteration. Once you encounter a changed field, all subsequent fields will be added to the **changed_field_names** list, regardless of whether their values have changed or not.

course_discovery/apps/api/utils.py
```diff
@@ -94,9 +94,11 @@ def reviewable_data_has_changed(obj, new_key_vals, exempt_fields=None):
         exempt_fields (list): List of field names where a change does not affect review status
 
     Returns:
-        bool for whether data for any reviewable fields has changed
+        tuple of two elements where first item is bool to tell whether data for any reviewable fields has changed and
+        second item tells which fields have changed
```
jmyatt: what purpose does the boolean serve?  Wouldn't an empty `changed_field_names` achieve the same thing?

moconnell1453: Looking at this logic more closely, I agree. We can treat an empty list as `changed == False`

moconnell1453 (approved):

jmyatt:
@muhammad-ammar will you be coordinating this with the associated terraform PR to setup the worker infrastructure?  Should this PR be marked blocked on that step?

muhammad-ammar:
> @muhammad-ammar will you be coordinating this with the associated terraform PR to setup the worker infrastructure? Should this PR be marked blocked on that step?

@jmyatt Currently signal is disable under a settings which means celery task is disabled too. I was thinking to merge this PR. Do you think there will be any issues if we merge the changes incrementally? I am working on terraform and other changes for worker.

jmyatt:
> > @muhammad-ammar will you be coordinating this with the associated terraform PR to setup the worker infrastructure? Should this PR be marked blocked on that step?
> 
> @jmyatt Currently signal is disable under a settings which means celery task is disabled too. I was thinking to merge this PR. Do you think there will be any issues if we merge the changes incrementally? I am working on terraform and other changes for worker.

ok yes that sounds like a good approach - thanks for the clarification

jmyatt (approved):

PR #2896, opened 2021-01-19
Added the workflow to use Github Actions to upgrade requirements

mraarif:
this PR intends to replace its Jenkins job for periodically checking for dependency updates and creating a PR for it.
this job/Github workflow can be invoked manually (once it's merged on master) and we can also specify the `target_branch` against which we want to run this job.

Relevant JIRA issue: https://openedx.atlassian.net/browse/BOM-2137


.github/workflows/requirements-upgrade.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Clone testeng-ci
+      run: git clone https://github.com/edx/testeng-ci.git
+
+    - name: setup default branch
+      if: ${{ github.event.inputs.branch }}==''
+      run: echo "target_branch=$(if ['${{ github.event.inputs.branch }}' = '']; then echo 'master'; else echo '${{ github.event.inputs.branch }}'; fi)" >> $GITHUB_ENV
+
+    - name: setup testeng-ci
+      run: |
+        echo "target branch: ${{ github.event.inputs.branch }}"
```
feanil, line 46: Is this echo necessary still or was it just for debugging during development?

requirements/constraints.txt
```diff
@@ -38,3 +38,5 @@ edx-lint<=1.6
 
 # latest version causing issues in IDAs.
 edx-auth-backends==3.3.0
+
+pip-tools==5.3.1
```
feanil: Why are we rolling back the pip-tools upgrade? When will it be okay to un-pin it? Can you add a comment with the details?

waheedahmed (approved):

PR #2899, opened 2021-01-20
[WIP][BD-14] Ensure editing organizations reactivates them in LMS/Studio

kdmccormick:
```
Since LMS/Studio organizations have been completely backfilled,
there are now hundreds of Inactive organizations in the
edxapp database. Inactive organizations are not surfaced to
users, and hence are effectively non-existent.

If an organization is created or edited in Discovery, one
would expect that it is a "real" organization and should
therefore appear as Active in LMS/Studio. This commit
ensures that saving an organization in Discovery tells
LMS to mark the organization as Active if it currently
exists as Inactive.
```

https://openedx.atlassian.net/browse/TNL-7913


kdmccormick:
Closed in favor of https://github.com/edx/edx-organizations/pull/165

PR #2904, opened 2021-01-22
add log to debug course skills update issue

muhammad-ammar:


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -400,9 +400,14 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
                     if settings.FIRE_UPDATE_COURSE_SKILLS_SIGNAL:
                         # If a skills relavant course field is updated than fire signal
                         # so that a background task in taxonomy update the course skills
-                        if any(field in COURSE_FIELDS_FOR_SKILLS for field in changed_fields):
-                            UPDATE_COURSE_SKILLS.send(self.__class__, course_uuid=course.uuid)
+                        relevant_fields_changed = any(field in COURSE_FIELDS_FOR_SKILLS for field in changed_fields)
+                        logger.info(
+                            '[UPDATE_COURSE_SKILLS_DEBUG] RelevantFieldsChanged: [%s]',
+                            relevant_fields_changed
+                        )
```
Dillon-Dumesnil, line 407: I don't think this is going to log out what you want it to.
Since relevant_fields_changed is an `any` call, it will return True or False. If that is what you want, feel free to merge. If you want the actual relevant fields that have changed, this needs to be updated.

Dillon-Dumesnil (approved):

course_discovery/apps/api/v1/views/courses.py
muhammad-ammar: @Dillon-Dumesnil Thank You for the review. actual changed fields are already logged. it looks like `any` is not working.

PR #2930, opened 2021-02-10
ENT-3892: Course Skills view link should not appear for draft courses, Also added success message when course skills are updated successfully.

saleem-latif:
__Description:__
Course skills page was throwing 500 error for draft courses, this PR fixes that, as well as hides the course skill page link for draft courses.

### DIFF
--- a/course_discovery/apps/course_metadata/views.py
+++ b/course_discovery/apps/course_metadata/views.py
@@ -1,7 +1,7 @@
 from django.contrib import admin, messages
 from django.contrib.auth import get_permission_codename
 from django.http import Http404, HttpResponseRedirect
-from django.shortcuts import render
+from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/views.py
```diff
@@ -139,10 +139,9 @@ def post(self, request, course_pk):
             for skill_id in form.cleaned_data['include_skills']:
                 remove_course_skill_from_blacklist(course.key, skill_id)
 
-        context = self._build_context(request, course_pk)
```
sameenfatima78, line 142: Any reason to remove context?

sameenfatima78 (approved):
LGTM, just one question

course_discovery/apps/course_metadata/views.py
saleem-latif: Yes, I thought it would be better to direct the user to the same page and let `get` handle the page serve instead of duplicating the code in `post`.

waheedahmed (approved):

PR #2936, opened 2021-02-10
Update analytics_api.py

MatthewPiatetsky:
https://edx-internal.slack.com/archives/G01J7T0CLCS/p1612813879012800

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/analytics_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/analytics_api.py
@@ -12,7 +12,7 @@ logger = logging.getLogger(__name__)
 
 class AnalyticsAPIDataLoader(AbstractDataLoader):
 
-    API_TIMEOUT = 120  # time in seconds
+    API_TIMEOUT = 300  # time in seconds, was increased to address timeouts when connecting to the analytics service
 
     def __init__(self, partner, api_url, max_workers=None, is_threadsafe=False):
         super().__init__(partner, api_url, max_workers, is_threadsafe)


course_discovery/apps/course_metadata/data_loaders/analytics_api.py
```diff
@@ -12,7 +12,7 @@
 
 class AnalyticsAPIDataLoader(AbstractDataLoader):
 
-    API_TIMEOUT = 120  # time in seconds
+    API_TIMEOUT = 300  # time in seconds
```
Dillon-Dumesnil: Can you add in a comment about why we are doing this and when we might be able to reduce this back to its current state?

Dillon-Dumesnil (approved):
Would you mind also updating the PR message with more context. Our slack messages are erased every six months so if someone finds this PR in >6 months, they will be left without any context.

course_discovery/apps/course_metadata/data_loaders/analytics_api.py
mikix: I don't think we plan to reduce it. Brian Mesick looked at the service, didn't see anything particularly wrong with it. Just a lot of data to sift. Mentioned maybe pagination as a possible improvement.

So could add a comment like "Consider reducing this if we ever get pagination" maybe?

DawoudSheraz:
Closing this PR because this is an old PR and there have been no updates on it. It can be re-opened if needed.

openedx-webhooks:
Although this pull request is already closed,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #2938, opened 2021-02-12
Give analytics service a larger timeout, during RCM

mikix:
We were seeing some timeouts during normal RCM runs. We think it just needs a bit more time.


ciduarte (approved):

mikix:
Already another PR for this, whoops.

PR #2968, opened 2021-03-05
add staff in raw fields

uzairr:
### Background
Lately the PC team has seen Discovery taking a lot of time to load, specially when it comes to course run pages.
e.g: https://discovery.edx.org/admin/course_metadata/courserun/21473/change/

### Investigation
This was happening because of the fact that `django-admin` needs to fetch all of the related data to populate `admin` page(provided above) for each `course_run`.In order to optimize the page load time, there was a need to fetch only relevant data.During the discovery, following fields are identified which would be fetched only on-demand:
1. `draft_version` in Inline Seat Model.
2. `upgrade_deadline` in Inline Seat Model.
3. 'Staff' list.
4. 'Transcript Languages' list.
5. 'Video Translation Languages` list.
### Actions
1. `draft_version` and `currency` fields inside Inline Seat Model are added to `raw_id_fields`.Now, they would be loaded only on demand.
#### Before
Screen Shot 2021-03-08 at 6 47 21 PM

#### After
Screen Shot 2021-03-08 at 6 50 55 PM

2. A custom widget is added to the `CourseRunAdmin` which will give suggestions as per the input of three characters and load only on selection.
#### Before
Screen Shot 2021-03-08 at 6 52 48 PM

####  After
Screen Shot 2021-03-07 at 6 03 29 AM

[DISCO-1677](https://openedx.atlassian.net/browse/DISCO-1677)

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -18,7 +18,9 @@ from course_discovery.apps.course_metadata.constants import COURSE_SKILLS_URL_NA
 from course_discovery.apps.course_metadata.exceptions import (
     MarketingSiteAPIClientException, MarketingSitePublisherException
 )
-from course_discovery.apps.course_metadata.forms import CourseAdminForm, PathwayAdminForm, ProgramAdminForm
+from course_discovery.apps.course_metadata.forms import (
+    CourseAdminForm, CourseRunAdminForm, PathwayAdminForm, ProgramAdminForm
+)
 from course_discovery.apps.course_metadata.models import *  # pylint: disable=wildcard-import
 from course_discovery.apps.course_metadata.views import CourseSkillsView
 
@@ -72,8 +74,8 @@ class ProgramEligibilityFilter(admin.SimpleListFilter):
 class SeatInline(admin.TabularInline):
     model = Seat
     extra = 1
-    readonly_fields = ('_upgrade_deadline',)
-    raw_id_fields = ('draft_version', )
+    readonly_fields = ('_upgrade_deadline', )
+    raw_id_fields = ('draft_version', 'currency')
 
 
 class PositionInline(admin.TabularInline):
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/admin.py
```diff
@@ -72,7 +74,7 @@ def queryset(self, request, queryset):
 class SeatInline(admin.TabularInline):
     model = Seat
     extra = 1
-    readonly_fields = ('_upgrade_deadline',)
+    readonly_fields = ('_upgrade_deadline', 'currency')
```
waheedahmed: Can put it in the raw id fields list instead of read-only? It might break some PC workflow and not sure about it.

course_discovery/apps/course_metadata/forms.py
```diff
@@ -90,6 +90,21 @@ class Meta:
...
+        widgets = {
+          'staff': SortedModelSelect2Multiple(
+              url='admin_metadata:person-autocomplete',
+              attrs={
+                  'data-minimum-input-length': 3,
+                  'class': 'sortable-select',
+              },
+          ),
+        }
+        fields = '__all__'
```
waheedahmed: Move this above the widgets and after the model variable.

course_discovery/apps/course_metadata/forms.py
```diff
@@ -90,6 +90,21 @@ class Meta:
         exclude = ('slug', 'url_slug', )
 
 
+class CourseRunAdminForm(forms.ModelForm):
+    class Meta:
+        model = CourseRun
+        widgets = {
+          'staff': SortedModelSelect2Multiple(
```
waheedahmed: I guess, taking advantage of this work we can add `transcript_languages` and `video_translation_languages` into this widget using `language_tags:language-tag-autocomplete` which will improve the performance further of this page.

waheedahmed (changes requested):

waheedahmed (approved):

PR #2980, opened 2021-03-11
WS-1738: Course Recommendation endpoint QA

Vyraal1:
Refined to use querysets for de-duping and exclusion of calling course. Added an additional test to check for matching both subjects and organizations.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1134,28 +1134,16 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         Recommended set of courses for upsell after finishing a course. Returns de-duped list of Courses that:
         A) belong in the same program as given Course
         B) share the same subject AND same organization (or at least one)
-        in priority over of A then B
+        in priority of A over B
         """
-        recommended_courses = []
-        # flatten a list of programs that has a list of courses
-        recommended_courses.extend(list(itertools.chain.from_iterable(
-            [program.courses.all() for program in self.programs.all()])))
-        course_subjects = [Q(subjects=subject) for subject in self.subjects.all()]
-        organizations = [Q(authoring_organizations=org) for org in self.authoring_organizations.all()]
-        recommended_courses.extend(list(Course.objects
-                                        .filter(*course_subjects)
-                                        .filter(*organizations)
-                                        .all()))
-        # filters out calling course
-        recommended_courses = [course for course in recommended_courses if course.key != self.key]
-        deduped_courses = []
-        seen = set()
-        for course in recommended_courses:
-            if course not in seen:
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1134,28 +1134,16 @@ def recommendations(self):
...
-                seen.add(course)
-
-        return deduped_courses
+        in priority of A over B
+        """
+        recommended_courses = (
+            Course.objects.filter(
+                subjects__in=self.subjects.all(),
+                authoring_organizations__in=self.authoring_organizations.all()
+            ) | Course.objects.filter(
```
rgraber: does the '|' mean there's an OR in the query, or we'll only look for matching programs if we don't find any of the same subj/org?

rgraber: also were you able to test this with an empty subject list? that's an edge case so NBD if not, but I'm curious.

Vyraal1: It's a SQL OR, so it'll add a course as long as it matches either: 
`prog1 OR prog2....`
OR
`(subject1 OR subject2...) AND (org1 OR org2...)`

rgraber (commented):
We need to return the ones from the same program first in the list

Vyraal1:
Should be ready now, had to use python native list since `union` and using other methods of maintain order such as `annotate` with `order_by` did not preserve the order.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1134,28 +1134,23 @@ def recommendations(self):
...
+            .distinct()
+            .all())
+
+        subject_org_courses = list(
+            Course.objects.filter(
+                subjects__in=self.subjects.all(),
+                authoring_organizations__in=self.authoring_organizations.all()
+            ).exclude(key=self.key)
+            .distinct()
+            .all())
```
waheedahmed, line 1151: Can we combine these two queries using `Q`?

waheedahmed (approved):
Just one question otherwise looks good 👍

course_discovery/apps/course_metadata/models.py
Vyraal1: I tried originally using Q's, but I had trouble preserving the order for `program_courses` over `subject_org_courses`. 

Approaches I tried were using `annotate` to have a `rec_order` with a value of 1 for the program courses, and a value of 0 for the subject courses and ordering by that, but I couldn't get it and eventually decided the trade-off wasn't worth spending more time on it for a short-term experiment. 

If the experiment is a win, there will be more emphasis on further refining the query.

waheedahmed (approved):
👍 once build is green

albemarle (approved):

srwang (approved):

PR #2984, opened 2021-03-16
WS-1804: Fix Recommendations

Vyraal1:
Added deduping upon union of query results while maintaining order. Also added a check to see if a course is marketable/has a marketable course run as it would lead to courses missing on Prospectus (only marketable courses have pages).

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1129,6 +1129,7 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
 
         return advertised_course_run
 
+    # Experiment WS-1681: Course recommendations
     def has_marketable_run(self):
         for course_run in self.course_runs.all():
             if course_run.is_marketable:
@@ -1158,7 +1159,15 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
             .distinct()
             .all())
 
-        return [*program_courses, *subject_org_courses]
+        # coercing to set destroys order, looping to remove dupes on union
+        seen = set()
+        deduped = []
+        for course in [*program_courses, *subject_org_courses]:
+            if course not in seen and course.has_marketable_run():
+                deduped.append(course)
+                seen.add(course)
+        return deduped
+        # End Experiment WS-1681: Course recommendations
 
 
 class CourseEditor(TimeStampedModel):
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1129,6 +1129,12 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
 
         return advertised_course_run
 
+    def has_marketable_run(self):
+        for course_run in self.course_runs.all():
+            if course_run.is_marketable:
+                return True
+        return False
+
     def recommendations(self):
         """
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1129,15 +1129,24 @@ def advertised_course_run(self):
 
         return advertised_course_run
 
+    # Experiment WS-1681: Course recommendations
+    def has_marketable_run(self):
+        for course_run in self.course_runs.all():
+            if course_run.is_marketable:
+                return True
+        return False
```
waheedahmed: Since `is_marketable` is a boolean field, we can simplify it.
```suggestion
        return any(run.is_marketable for run in self.course_runs.all())
```

waheedahmed: And if you want to avoid any further iterations if we find a marketable run at the start then you could simply return the `return course_run.is_marketable`, no need to return `True` explicitly.

waheedahmed (approved):
One suggestion otherwise LGTM 👍

rgraber (approved):
LGTM

course_discovery/apps/course_metadata/models.py
```diff
@@ -1150,7 +1156,15 @@ def recommendations(self):
             .distinct()
             .all())
 
-        return [*program_courses, *subject_org_courses]
+        # coercing to set destroys order, looping to remove dupes on union
+        seen = set()
+        deduped = []
+        for course in [*program_courses, *subject_org_courses]:
```
albemarle, line 1162: program_courses is already deduped, right?  So you can start by making deduped a copy of that array, and only do the checking for subject_org_courses.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1150,7 +1156,15 @@ def recommendations(self):
             .distinct()
             .all())
 
-        return [*program_courses, *subject_org_courses]
+        # coercing to set destroys order, looping to remove dupes on union
```
albemarle, line 1159: Can you improve the clarity of this comment?

course_discovery/apps/course_metadata/models.py
albemarle: Oh nm, you have to check for has_marketable_run().  In that case I would delete the comment on line 1159 altogether.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -2422,30 +2422,57 @@ def setUp(self):
...
-        self.course2_with_subject = factories.CourseFactory(key='course2', subjects=[cooking])
-        self.course3_with_different_subject = factories.CourseFactory(key='course3', subjects=[brewing])
-        self.course4_with_2_subjects = factories.CourseFactory(key='course4', subjects=[brewing, cooking])
-        self.course5_with_subject = factories.CourseFactory(key='course5', subjects=[not_cooking])
-        self.course6_with_subject = factories.CourseFactory(key='course6', subjects=[not_brewing])
+        self.course1_with_subject = CourseFactory(key='course1', subjects=[cooking])
+        self.course2_with_subject = CourseFactory(key='course2', subjects=[cooking])
+        self.course3_with_different_subject = CourseFactory(key='course3', subjects=[brewing])
+        self.course4_with_2_subjects = CourseFactory(key='course4', subjects=[brewing, cooking])
+        self.course5_with_subject = CourseFactory(key='course5', subjects=[not_cooking])
```
albemarle, line 2429: :-)

PR #2985, opened 2021-03-19
Add duplicate of description field for spanish translations

zainab-amir:
Created new field for school pages in Disco to allow for alternate translated text to be manually inputted.

- Updated model to include `description_es` field
- Updated serializer to include `description_es`
-----

Locally `/api/v1/organizations/` now includes the newly added field in it's response:

Screenshot 2021-03-19 at 5 16 15 PM

[DISCO-1737](https://openedx.atlassian.net/browse/DISCO-1737)


zainab-amir:
Removed `null=True` from the `description_es` field, details:
```
     = Avoid using null on string-based fields such as CharField and TextField.
       If a string-based field has null=True, that means it has two possible
       values for "no data": NULL, and the empty string. In most cases, it's
       redundant to have two possible values for "no data;" the Django
       convention is to use the empty string, not NULL.
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -168,6 +168,7 @@ class Organization(CachedMixin, TimeStampedModel):
     )
     slug = AutoSlugField(populate_from='key', editable=False, slugify_function=uslugify)
     description = models.TextField(null=True, blank=True)
+    description_es = models.TextField(blank=True)
```
waheedahmed: You might need to set `null=True` as well.

zainab-amir: If I set `null=True` the CI / semgrep check fails with the above error

zainab-amir: I think we can keep it this way and inform website squad that the default value will be empty string

waheedahmed: Got it.

waheedahmed: Specify `verbose_name=_('Spanish Description')` and a help text specifying the purpose of the field.

waheedahmed (approved):
A couple of suggestions and don't forget to recreate the migration after these changes.

waheedahmed (approved):

PR #2988, opened 2021-03-25
ENT-4105: Filter out the blacklisted courses skills in API response.

saleem-latif:
__Description:__:
This PR filters out the blacklisted course's skills in API response.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -21,7 +21,7 @@ from rest_framework.fields import CreateOnlyDefault, UUIDField
 from rest_framework.metadata import SimpleMetadata
 from rest_framework.relations import ManyRelatedField
 from taggit_serializer.serializers import TaggitSerializer, TagListSerializerField
-from taxonomy.models import CourseSkills
+from taxonomy.utils import get_whitelisted_course_skills
 
 from course_discovery.apps.api.fields import (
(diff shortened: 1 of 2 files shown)


sameenfatima78 (approved):

waheedahmed (approved):
👍 once tests passes.

course_discovery/apps/api/serializers.py
```diff
@@ -1133,7 +1133,7 @@ def get_canonical_course_run_key(self, obj):
         return None
 
     def get_skill_names(self, obj):
-        course_skills = CourseSkills.objects.select_related('skill').filter(course_id=obj.key)
+        course_skills = get_whitelisted_course_skills(obj.key)
         return list(set(course_skill.skill.name for course_skill in course_skills))
```
HammadAhmadWaqas, line 1138: can we use query to get skill names instead of using the loop?
`course_skills.values_list('skill__name', flat=True)`

Also, can we update the relevant test to add some skills so that it can make sure we are getting the correct skill names?
here are few tests: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/tests/test_serializers.py

saleem-latif: @HammadAhmadWaqas I have updated the tests, please take a look. I can not use your other suggestion as `get_whitelisted_course_skills` is supposed to return a list of CourseSkill objects and not a queryset.

HammadAhmadWaqas (approved):
looks good to me once tests pass.

PR #2990, opened 2021-03-25
add content_language_search_facet_name to CourseRunSerializer

adamstankiewicz:
In order to add the "Language" search facet to the Enterprise Learner Portal and have an exact match of language names to correspond with edx.org/search, I added the content language search facet name to the `CourseRunSerializer` such that it appears in the necessary API response for enterprise-catalog, e.g.:

```
{
  "content_language": "zh-cmn",
  "content_language_search_facet_name": "Chinese - Mandarin",
  ...
}
```

See https://github.com/edx/enterprise-catalog/pull/278 for how this will be used by enterprise-catalog.

**Alternatives considered**
1. Using the `langcodes` package in enterprise-catalog to parse the existing `content_language` field that includes the language code, e.g. "en-us". However, this approach did not work due to how edx.org/search formats the Chinese languages (e.g., "Chinese - Mandarin" vs. "Mandarin Chinese", which throws off the A-Z sorting).
2. Data duplication of the `LanguageTag` model in enterprise-catalog.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
@@ -316,7 +316,7 @@ class CatalogViewSetTests(ElasticsearchTestMixin, SerializationMixin, OAuth2Mixi
             # to be included.
             filtered_course_run = CourseRunFactory(course=course)
 
-            with self.assertNumQueries(31, threshold=3):
+            with self.assertNumQueries(35, threshold=3):
                 response = self.client.get(url)
 
             assert response.status_code == 200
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/tests/test_views/test_catalogs.py
```diff
@@ -316,7 +316,7 @@ def test_courses(self, state):
             # to be included.
             filtered_course_run = CourseRunFactory(course=course)
 
-            with self.assertNumQueries(31, threshold=3):
+            with self.assertNumQueries(35, threshold=3):
```
iloveagent57, line 319: OMG, I forgot how crazy these assertions drive me.

iloveagent57 (approved):

waheedahmed (approved):

PR #2993, opened 2021-03-29
make Course/Run UUID field editable with a waffle flag.

waheedahmed:
Made Course and CourseRun draft version UUID field editable if waffle flag is enabled.

DISCO-1745


course_discovery/apps/course_metadata/admin.py
```diff
@@ -209,11 +222,23 @@ class CourseRunAdmin(admin.ModelAdmin):
     )
     ordering = ('key',)
     raw_id_fields = ('course', 'draft_version',)
-    readonly_fields = ('uuid', 'enrollment_count', 'recent_enrollment_count', 'hidden', 'key')
+    readonly_fields = ('enrollment_count', 'recent_enrollment_count', 'hidden', 'key')
```
mikix, line 225: It might be less confusing for someone scanning this code to still see `uuid` in this list (and just remove it in `get_readonly_fields`). But just a nit.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -209,11 +222,23 @@ class CourseRunAdmin(admin.ModelAdmin):
...
     search_fields = ('uuid', 'key', 'title_override', 'course__title', 'slug', 'external_key')
     save_error = False
     form = CourseRunAdminForm
 
+    def get_readonly_fields(self, request, obj=None):
+        """
+        Make UUID field editable for draft if flag is enabled.
+        """
+        if obj and obj.draft:
+            flag_name = f'{obj._meta.app_label}.{obj.__class__.__name__}.make_uuid_editable'
```
mikix, line 235: This ends up as `course_metadata.courserun.make_uuid_editable`?

mikix (approved):

course_discovery/apps/course_metadata/admin.py
waheedahmed: @mikix `course_metadata.CourseRun.make_uuid_editable` to be exact.

waheedahmed: And for Course `course_metadata.Course.make_uuid_editable`

course_discovery/apps/course_metadata/admin.py
waheedahmed: @mikix I tried that but since it is a tuple there is no straight way to remove the item from a tuple, that's why I choose concatenation instead of removing.

PR #2997, opened 2021-04-02
feat: Management command to add assets to edx Org

Tj-Tracy:
This command, when used by a provisioning script, will load logos for
the edX organization so that a program certificate can be created
locally.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/add_logos_to_organization.py
+++ b/course_discovery/apps/course_metadata/management/commands/add_logos_to_organization.py
@@ -24,7 +24,7 @@ class Command(BaseCommand):
         try:
             # We search by partner here because right now, the default provisioned org does not have a name, but has "edX" as the partner
             organization = Organization.objects.filter(
-                partner=self._get_partner_by_name(partner_name),
+                partner__short_code=partner_name
             )[:1].get()
             assets = self._open_assets(
                 logo=kwargs.get("logo"),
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/add_logos_to_organization.py
```diff
@@ -0,0 +1,46 @@
...
+        parser.add_argument('--certificate_logo', default='/edx/app/discovery/discovery/provision-temp/assets/demo-asset-certificate-logo.png')
+        parser.add_argument('--banner_image', default='/edx/app/discovery/discovery/provision-temp/assets/demo-asset-banner-image.png')
+
+    def handle(self, *args, **kwargs):
+        partner_name = kwargs.get('partner')
+        logo_path = kwargs.get("logo")
+        certificate_logo_path = kwargs.get('certificate_logo')
+        banner_image_path = kwargs.get('banner_image')
+        try:
+            organization = Organization.objects.get(partner=self._get_partner_by_name(partner_name))
```
waheedahmed: I guess no need to make an extra query for the partner and instead, you can directly do this:
```suggestion
            organization = Organization.objects.get(partner__name=partner_name)
```

waheedahmed (commented):
Also, instead of the partner name can we use `short_code` here?

course_discovery/apps/course_metadata/management/commands/add_logos_to_organization.py
Tj-Tracy: Thanks for the suggestion. 🔧

course_discovery/apps/course_metadata/management/commands/add_logos_to_organization.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class Command(BaseCommand):
+    """
+    Uploads given images to an organization. This command is meant to be run
+    from a provisioning script and not by hand.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument("--partner", default="edX")
```
crice100: I think this should be edx (all lower case). Also on line 25.

crice100 (approved):

waheedahmed (approved):
👍 once the build is green

PR #3001, opened 2021-04-06
ENT-4285 | Added `Refresh Course Skills` update review into CourseAdmin view.

HammadAhmadWaqas:
Screen Shot 2021-04-06 at 6 10 32 PM
Screen Shot 2021-04-06 at 6 10 59 PM

### DIFF
--- a/course_discovery/apps/course_metadata/views.py
+++ b/course_discovery/apps/course_metadata/views.py
@@ -5,11 +5,13 @@ from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import TemplateView, UpdateView, View
+from taxonomy.signals.signals import UPDATE_COURSE_SKILLS
 from taxonomy.utils import (
     blacklist_course_skill, get_blacklisted_course_skills, get_whitelisted_course_skills,
     remove_course_skill_from_blacklist
 )
 
+from course_discovery.apps.course_metadata.constants import REFRESH_COURSE_SKILLS_URL_NAME
 from course_discovery.apps.course_metadata.forms import CourseRunSelectionForm, ExcludeSkillsForm
 from course_discovery.apps.course_metadata.models import Course, Program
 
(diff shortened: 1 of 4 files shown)


moconnell1453 (approved):

zamanafzal (approved):

course_discovery/apps/course_metadata/views.py
```diff
@@ -45,6 +47,85 @@ def get_success_url(self):
...
+        return HttpResponseRedirect(
+            reverse(f'admin:{REFRESH_COURSE_SKILLS_URL_NAME}', args=(course_pk,))
+        )
+
+    def _update_course_skills(self, course_uuid):
+        """
+        Call the management command to refresh data.
+        """
+        try:
+            management.call_command('refresh_course_skills', '--course', course_uuid, '--commit')
```
waheedahmed: Is this a blocking call? Since this command `refresh_course_skills` is making calls to third-party services outside edX, I guess using a background task here would be a good idea.

waheedahmed (commented):
One question otherwise looks good.

course_discovery/apps/course_metadata/views.py
HammadAhmadWaqas: @waheedahmed this command is called from taxonomy-connector which is installed it course-discovery itself.

waheedahmed: Sorry, I didn’t understand.

waheedahmed: Actually, my question is more related to what that command `refresh_course_skills` does but not where it lives.

HammadAhmadWaqas: Sorry for not making it clear.

And Yes `refresh_course_skills` command is getting data from a Third-party service. but as it's only for a single course, this should not take time.

@waheedahmed Do you still think we should add this into the async task?

HammadAhmadWaqas:
@waheedahmed PR updated as per suggestion.

Thank you for the feedback.

waheedahmed (approved):

PR #3008, opened 2021-04-14
ENT-4444: Upgraded taxonomy-connector version to 1.9.0

saleem-latif:
**JIRA Ticket:** [ENT-4444](https://openedx.atlassian.net/browse/ENT-4444)

__Description:__
This PR upgrades taxonomy-connector version to 1.9.0, new version contains the following changes.

1. A data migration to copy the values from the course_id field into the course_key field. 
2. Removal of all references to the `course_id` field in the code, NOT removing the `course_id` field from the model in the code.
3. Changing `null` back to `False` and `editable` back to `True` on the `course_key` field.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -171,6 +171,7 @@ class SkillFactory(factory.django.DjangoModelFactory):
 
 class CourseSkillsFactory(factory.django.DjangoModelFactory):
     course_id = FuzzyText()
+    course_key = FuzzyText()
     skill = factory.SubFactory(SkillFactory)
     confidence = FuzzyDecimal(0.0, 1.0)
     is_blacklisted = False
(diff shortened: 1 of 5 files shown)


sameenfatima78 (approved):

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -171,6 +171,7 @@ class Meta:
 
 class CourseSkillsFactory(factory.django.DjangoModelFactory):
     course_id = FuzzyText()
```
waheedahmed, line 173: Do we still need this `course_id` or can it be removed now?

waheedahmed (approved):
One question otherwise looks good.

course_discovery/apps/course_metadata/tests/factories.py
saleem-latif: At the moment, we need this because removing it causes an Integrity Error on db insertions. However, it will be removed in the next release as part of the 3rd sub task of [ENT-4402](https://openedx.atlassian.net/browse/ENT-4402)

PR #3010, opened 2021-04-16
ENT-4436: Add feature flag in Organization to display course skills in Publisher

sameenfatima78:
**JIRA:** 

This PR adds a feature flag in the Organization model which will be False by default. If enabled, the skills will show up in the Publisher in course edit page. 

- https://openedx.atlassian.net/browse/ENT-4332
- https://openedx.atlassian.net/browse/ENT-4436

**Screenshots:** 

Screenshot 2021-04-16 at 2 46 24 PM


course_discovery/apps/course_metadata/models.py
```diff
@@ -205,6 +205,13 @@ class Organization(CachedMixin, TimeStampedModel):
             " generated.  When this flag is disabled, the key can be manually set."
         )
     )
+    enable_skills_in_publisher = models.BooleanField(
+        default=False,
+        verbose_name=_('Enable skills in Publisher'),
+        help_text=_(
+            "When this flag is enabled, the skill tags will show up in the course edit page in Publisher."
```
saleem-latif: nit: use single quotes for strings. i.e. `'When this flag is enabled, the skill tags will show up in the course edit page in Publisher.'`

saleem-latif (approved):
One small comment, otherwise Looks Good 👍

waheedahmed (approved):
👍 once double-quotes changed to single.

waheedahmed:
Actually, I have one question. Are you going to use this flag in some API endpoint or at the publisher frontend to determine whether you need to show the skills or not?

sameenfatima78:
> Actually, I have one question. Are you going to use this flag in some API endpoint or at the publisher frontend to determine whether you need to show the skills or not?

@waheedahmed The latter one; I will be using it in the Publisher
Screenshot 2021-04-16 at 4 40 31 PM

course_discovery/apps/course_metadata/models.py
```diff
@@ -205,6 +205,13 @@ class Organization(CachedMixin, TimeStampedModel):
             " generated.  When this flag is disabled, the key can be manually set."
         )
     )
+    enable_skills_in_publisher = models.BooleanField(
+        default=False,
+        verbose_name=_('Enable skills in Publisher'),
+        help_text=_(
+            'When this flag is enabled, the skill tags will show up in the course edit page in Publisher.'
+        )
+    )
```
muhammad-ammar, line 214: @sameenfatima78 I think backend should not send skills to frontend if `enable_skills_in_publisher` is set `False`. Validations on frontend can be bypassed easily plus one can see the skills in the API response.

muhammad-ammar (approved):

course_discovery/apps/course_metadata/models.py
sameenfatima78: As discussed on call, this feature flag will be removed in the future and is only in place right now for testing on 2-3 partners so no action required here.

course_discovery/apps/course_metadata/models.py
sameenfatima78: Done

PR #3039, opened 2021-05-21
fix: update slug for official version

waheedahmed:
If there is an official course version exists but no active or published course run, slug change does not gets updated for official version.


rgraber (approved):
LGTM

Vyraal1 (approved):

uzairr (approved):

PR #3042, opened 2021-05-26
ENT-4158: Add skill description to search all and course detail endpoint

sameenfatima78:
JIRA: https://openedx.atlassian.net/browse/ENT-4158

This PR adds a new field called "skills" in `search/all` and course detail endpoint which returns an array of objects, holding skills belonging to a course alongwith their descriptions. 

Example:
Screenshot 2021-05-26 at 1 51 27 PM

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -21,7 +21,7 @@ from rest_framework.fields import CreateOnlyDefault, UUIDField
 from rest_framework.metadata import SimpleMetadata
 from rest_framework.relations import ManyRelatedField
 from taggit_serializer.serializers import TaggitSerializer, TagListSerializerField
-from taxonomy.utils import get_whitelisted_course_skills
+from taxonomy.utils import get_whitelisted_course_skills, get_whitelisted_serialized_skills
 
 from course_discovery.apps.api.fields import (
     HtmlField, ImageField, SlugRelatedFieldWithReadSerializer, SlugRelatedTranslatableField, StdImageSerializerField
@@ -1077,6 +1077,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
                                                        queryset=Collaborator.objects.all(),
                                                        read_serializer=CollaboratorSerializer())
     skill_names = serializers.SerializerMethodField()
+    skills = serializers.SerializerMethodField()
 
     @classmethod
     def prefetch_queryset(cls, partner, queryset=None, course_runs=None):  # pylint: disable=arguments-differ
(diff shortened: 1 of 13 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1145,6 +1147,15 @@ def get_skill_names(self, obj):
         course_skills = get_whitelisted_course_skills(obj.key)
         return list(set(course_skill.skill.name for course_skill in course_skills))
 
+    def get_skills(self, obj):
+        course_skills = get_whitelisted_course_skills(obj.key)
+        if course_skills:
+            return [{
+                'name': course_skill.skill.name,
+                'description': course_skill.skill.description,
+            } for course_skill in course_skills]
```
waheedahmed: Since `get_whitelisted_course_skills()` returns a queryset, can we use `.values()` here and at all other places?

waheedahmed: And, can we make this whole logic a utility function here in `course-discovery` or in `taxonomy-connector` instead of duplicating at each place?

waheedahmed (commented):
Couple of questions otherwise looks good

course_discovery/apps/api/serializers.py
sameenfatima78: Good suggestion, I have created a utility function and made changes accordingly. 

Regarding using `.values()`, I think it won't be possible to fetch Skill data in that case. Queryset holds a reference to the Skill object inside CourseSkill... from which we are fetching the name..

course_discovery/apps/course_metadata/utils.py
```diff
@@ -414,6 +414,16 @@ def calculated_seat_upgrade_deadline(seat):
     return None
 
 
+def serialize_course_skills(course_skills):
```
saleem-latif: We should write a serializer in `taxonomy-connector` with implementation something like
```
class SkillSerializer(serializers.ModelSerializer):
        class Meta:
        model = Skill
        fields = ('name', 'description')
```
And then add a utility method in `taxonomy.utils`

```
def serialize_course_skills(course_skills):
    """
    Get a list of serialized course skills.
    
    Arguments:
        course_skills (list<CourseSkill>): A list of course skills that need to be serialised.
     
     Returns:
         (dict): A dictionary containing the following key-value pairs
             1.  name: 'Skill name'
             2. description: "Skill Description"
    """
    skills = [course_skill.skill for course_skill in course_skills]
    serializer = SkillSerializer(skills, many=True)
    return serializer.data
```
You can also do this part in a follow up PR if you need to merge this now, the above suggestion is only refactor work and should not affect the functionality.

saleem-latif (approved):
One comment, otherwise looks good 👍

waheedahmed (approved):

course_discovery/apps/course_metadata/utils.py
sameenfatima78: Done

PR #3045, opened 2021-06-01
fix: allow publisher to remove video from course

waheedahmed:
Currently, there is no way in the Publisher to remove video links from a course. Now PCs and course teams can remove course about video from the course using Publisher.

DISCO-1736


albemarle (approved):

rgraber (approved):
LGTM

PR #3080, opened 2021-08-05
feat: add taxonomy jobs data in /search/all and course detail apis

muhammad-ammar:
**JIRA:** https://openedx.atlassian.net/browse/ENT-4583

**Description:** This will add `jobs` data from taxonomy in `/search/all` and `course detail` endpoint.


sameenfatima78 (commented):
We also need to pass the 'jobs' field in the Meta class of `CourseSearchDocumentSerializer` and `CourseRunSearchDocumentSerializer` 

See this: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/search_indexes/serializers/course.py#L129

DawoudSheraz:
Closing this PR as it has not been updated since 2021. It can be re-opened if it is still needed.

PR #3098, opened 2021-08-18
fix: learner dashboard shows updated course image (AA-992)

sofiayoon:
## Description

Course image updated via Publisher was not updated in the learner dashboard. This PR fixes this bug by updating the course image both in Studio and the learner dashboard.  

## Supporting information

https://openedx.atlassian.net/browse/AA-992


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -389,6 +389,7 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
                 if course_run.status == CourseRunStatus.Published:
                     # This will also update the course
                     course_run.update_or_create_official_version()
+                    CourseRunViewSet.update_course_run_image_in_studio(course_run)
```
mikix, line 392: This seems good and workable - could land.

But I think there's also a version of this fix where this line gets pushed into `CourseRun.update_or_create_official_version`, in its `notify_services` stanza, along with where we update ecommerce and the LMS tracks.

It feels like a reasonable fit there. But... that change might be more complicated (means you could remove a custom call in `course_runs.py` too probably - maybe move some code around). And might result in us updating the image more often then needed. Which :shrug:

But ultimately, the image is derived from the course. So it's reasonable to add this call to `courses.py`. The above thought was just a possible larger-scope refactor.

mikix (approved):

waheedahmed (approved):

PR #3121, opened 2021-09-09
fix: bug with how translated fields are loaded from fixture

matthugs:
JIRA:[EDUCATOR-5112](https://openedx.atlassian.net/browse/EDUCATOR-5112)

It always confused me why we had so much asymmetry to how we loaded these fixtures. I tested it without and it seems to work just fine.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -271,7 +271,7 @@ class CourseViewSetTests(OAuth2Mixin, SerializationMixin, APITestCase):
         keys = ','.join([course.key for course in courses])
         url = '{root}?keys={keys}'.format(root=reverse('api:v1:course-list'), keys=keys)
 
-        with self.assertNumQueries(46):
+        with self.assertNumQueries(49):
             response = self.client.get(url)
         self.assertListEqual(response.data['results'], self.serialize_course(courses, many=True))
 
--- a/course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
+++ b/course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
@@ -13,7 +13,7 @@ from edx_rest_api_client import client as rest_client
 
 from course_discovery.apps.core.models import Partner
 from course_discovery.apps.course_metadata.models import (
-    CourseRun, Curriculum, CurriculumCourseMembership, CurriculumProgramMembership, Program, ProgramType, SeatType
+    CourseRun, Curriculum, CurriculumCourseMembership, CurriculumProgramMembership
 )
 from course_discovery.apps.course_metadata.signals import (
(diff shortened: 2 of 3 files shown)


zacharis278:
@matthugs getting back from vacation and just seeing this now. I'll have to spend a bit of time reasoning the 'why' on this and get back to you b/c off the top of my head I don't recall.

matthugs:
@Zacharis278 thanks for taking a look! Based on what I can see in the tests (and the reasons the CI isn't green on this PR yet) the desire was to leave the command still idempotent even when previously pulled-in models were deleted and recreated. In that use case the ids for the models would be different, but some of the data would be the same. So it seems to be preventing duplicate entries in that sort of case.

If my assessment is correct, I think ripping out the extra complexity is a good idea. Maybe at the time program types were new it was likely that they'd change over the lifetime of an integration sandbox, but I don't think that's a likely scenario any more.

zacharis278:
@matthugs ok yeah that makes sense. Originally we wanted this to be runnable on existing sandboxes to sync  updated/added/removed data. Since you're the only person really using this, if you don't see any need for that complexity let's just make things simpler.

zacharis278:
might be worth changing the docstring for this command to mention it's intended for a fresh sandbox and DB id's may conflict otherwise. Would hate for someone to unknowingly to go back down that rabbit hole.

course_discovery/apps/course_metadata/management/commands/load_program_fixture.py
```diff
@@ -79,6 +79,11 @@ class Command(BaseCommand):
     """
     Command to populate catalog database with programs from another environment
     using the /program-fixtures endpoint
+    
+    Idempotent up to changes in the ids of records in the source
+    environment; if a record previously copied by this command has
+    been deleted in the source environment and then recreated under a
+    new primary key then this command will likely fail.
```
matthugs, line 86: @Zacharis278 done

zacharis278 (approved):
👍

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -271,7 +271,7 @@ def test_list_key_filter(self):
         keys = ','.join([course.key for course in courses])
         url = '{root}?keys={keys}'.format(root=reverse('api:v1:course-list'), keys=keys)
 
-        with self.assertNumQueries(46):
+        with self.assertNumQueries(49):
```
matthugs, line 274: This test was consistently failing before I made this change, despite my not having made a change upstream of this test. I noticed, however, that there were exactly 3 queries in the test output ([e.g. here](https://github.com/edx/course-discovery/runs/3679046418?check_suite_focus=true)) which referenced translations. So my loose theory is that my updating some tests to more explicitly reference translations/translated fields caused django-parler and its associated modules to be more widely loaded into the testing runtime. But /shrug on the details. I'm gonna' move ahead with this because I am #tired of this PR

PR #3123, opened 2021-09-10
fix: update course_run api

mubbsharanwar:
- update course_run api to edit upgrade_deadline date using publisher app.

DISCO-1622

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -270,7 +270,7 @@ class CourseRunViewSet(ValidElasticSearchQueryRequiredMixin, viewsets.ModelViewS
         return response
 
     @writable_request_wrapper
-    def _update_course_run(self, course_run, draft, changed, serializer, request, prices):
+    def _update_course_run(self, course_run, draft, changed, serializer, request, prices, upgrade_deadline_override):
         save_kwargs = {}
         # If changes are made after review and before publish, revert status to unpublished.
         # Unless we're just switching the status
@@ -289,7 +289,7 @@ class CourseRunViewSet(ValidElasticSearchQueryRequiredMixin, viewsets.ModelViewS
         course_run = serializer.save(**save_kwargs)
 
         if course_run in course_run.course.active_course_runs:
-            course_run.update_or_create_seats(course_run.type, prices)
+            course_run.update_or_create_seats(course_run.type, prices, upgrade_deadline_override,)
 
         self.push_to_studio(request, course_run, create=False)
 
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -179,6 +179,7 @@ def list(self, request, *args, **kwargs):
 
     @classmethod
     def push_to_studio(cls, request, course_run, create=False, old_course_run_key=None):
+        print('studio api link ', course_run.course.partner.studio_url)
```
waheedahmed: nit: remove this print statement.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1739,12 +1739,14 @@ def get_seat_upgrade_deadline(self, seat_type):
...
 
-    def update_or_create_seat_helper(self, seat_type, prices):
+    def update_or_create_seat_helper(self, seat_type, prices, upgrade_deadline):
         defaults = {
             'upgrade_deadline': self.get_seat_upgrade_deadline(seat_type),
         }
         if seat_type.slug in prices:
             defaults['price'] = prices[seat_type.slug]
+        if upgrade_deadline:
+            defaults['upgrade_deadline'] = upgrade_deadline
```
waheedahmed: We were supposed to expose the `upgrade_deadline_override` in Publisher instead of `upgrade_deadline` field.

waheedahmed (changes requested):
Also, update the Publisher PR.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1751,10 +1751,11 @@ def update_or_create_seat_helper(self, seat_type, prices):
             type=seat_type,
             draft=True,
             defaults=defaults,
+            upgrade_deadline_override=upgrade_deadline_override,
```
waheedahmed: We need to add a check above this otherwise it will set null if the value is not sent in the PATCH request.

waheedahmed: Also, in my testing, it created duplicate seat objects instead of updating the existing ones and also added the upgrade deadline override value to the Audit seat type which shouldn't have happened.

waheedahmed (changes requested):

course_discovery/apps/course_metadata/models.py
waheedahmed: Screenshot 2021-09-17 at 7 39 06 PM

course_discovery/apps/course_metadata/models.py
```diff
@@ -1739,13 +1739,16 @@ def get_seat_upgrade_deadline(self, seat_type):
...
 
-    def update_or_create_seat_helper(self, seat_type, prices):
+    def update_or_create_seat_helper(self, seat_type, prices, upgrade_deadline_override):
         defaults = {
             'upgrade_deadline': self.get_seat_upgrade_deadline(seat_type),
         }
         if seat_type.slug in prices:
             defaults['price'] = prices[seat_type.slug]
 
+        if upgrade_deadline_override and seat_type.slug != Seat.AUDIT:
```
waheedahmed: I guess the upgrade deadline is only meant for Verified seat types and we can update this condition like this:
```suggestion
        if upgrade_deadline_override and seat_type.slug == Seat.VERIFIED:
```

waheedahmed (changes requested):
One change otherwise 👍

course_discovery/apps/course_metadata/models.py
mubbsharanwar: if upgrade deadline is only meant for Verified seat types then condition should be like 
`if upgrade_deadline_override and seat_type.slug == Seat.VERIFIED:`

waheedahmed (approved):

PR #3130, opened 2021-09-21
fix: fixing tests case due to django-stdimage and django32 changes wi…

awais786:
fixing tests case due to django-stdimage and django32 changes with setUpTestData().

Issue reported in stdimage
https://github.com/codingjoe/django-stdimage/issues/246

Release notes
https://docs.djangoproject.com/en/3.2/releases/3.2/#tests


awais786:
After this merge I will check django-autocomplete package with django32 compatibility.

waheedahmed (approved):

PR #3199, opened 2021-10-14
fix: add django version condition on access of response headers

iamsobanjaved:
Need to add django version condition to support both Django 3.2 and Django 2.2


awais786 (approved):

DawoudSheraz:
Closing this old PR. This can be re-opened if it is still required.

PR #3213, opened 2021-11-03
build: program fixtures now exclude language tags

matthugs:
(also hopefully exclude their translations)
rationale is that the languages are created primarily through
migrations (which is a weird practice for picklist tables, y'all!), so
it's not really necessary to recreate them. I'm a bit worried about
that approach, though, since not all languages defined in production
discovery are created on a fresh provision, but it seems very likely
that this will cover our use case so /shrug


waheedahmed (approved):
👍 once tests are passing.

PR #3220, opened 2021-11-23
feat: add recent_enrollment_count in program list endpoint

sameenfatima78:
We index `recent_enrollment_count` field for courses but this field is not available for programs in Algolia. We use programs list endpoint to fetch all content metadata in enterprise_catalogs that is then indexed. This PR is about making `recent_enrollment_count` field available for programs as well. 
JIRA: https://openedx.atlassian.net/browse/ENT-5142

Changes:
- Added `recent_enrollment_count` in `MinimalProgramSerializer`
- Removed `recent_enrollment_count` from `ProgramSerializer`

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1549,7 +1549,7 @@ class MinimalProgramSerializer(FlexFieldsSerializerMixin, BaseModelSerializer):
             'uuid', 'title', 'subtitle', 'type', 'type_attrs', 'status', 'marketing_slug', 'marketing_url',
             'banner_image', 'hidden', 'courses', 'authoring_organizations', 'card_image_url',
             'is_program_eligible_for_one_click_purchase', 'degree', 'curricula', 'marketing_hook',
-            'total_hours_of_effort',
+            'total_hours_of_effort', 'recent_enrollment_count',
         )
         read_only_fields = ('uuid', 'marketing_url', 'banner_image')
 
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1709,7 +1709,7 @@ class Meta(MinimalProgramSerializer.Meta):
             'faq', 'credit_backing_organizations', 'corporate_endorsements', 'job_outlook_items',
             'individual_endorsements', 'languages', 'transcript_languages', 'subjects', 'price_ranges',
             'staff', 'credit_redemption_overview', 'applicable_seat_types', 'instructor_ordering',
-            'enrollment_count', 'recent_enrollment_count', 'topics', 'credit_value',
```
saleem-latif, line 1712: why it is removed from `MinimalProgramSerializer` ?

course_discovery/apps/api/serializers.py
```diff
@@ -1549,7 +1549,7 @@ class Meta:
             'uuid', 'title', 'subtitle', 'type', 'type_attrs', 'status', 'marketing_slug', 'marketing_url',
             'banner_image', 'hidden', 'courses', 'authoring_organizations', 'card_image_url',
             'is_program_eligible_for_one_click_purchase', 'degree', 'curricula', 'marketing_hook',
-            'total_hours_of_effort',
+            'total_hours_of_effort', 'recent_enrollment_count',
```
saleem-latif, line 1552: Is this field `recent_enrollment_count` already present on `Program` model?

saleem-latif (approved):
A couple of questions, otherwise looks good 👍🏼

course_discovery/apps/api/serializers.py
sameenfatima78: Yes https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L2187

course_discovery/apps/api/serializers.py
sameenfatima78: It is actually not removed but added in the `MinimalProgramSerializer`, it's removed from `ProgramSerializer` because it uses `MinimalProgramSerializer` fields along with its own. The API endpoint from where we fetch content metdata in `enterprise-catalogs` uses `MinimalProgramSerializer`, that is why I moved the field there.

zainab-amir (approved):
LGTM :shipit:

PR #3223, opened 2021-11-25
feat: add LearnerPathwayCourse model

muhammad-ammar:


course_discovery/apps/learner_pathway/models.py
```diff
@@ -80,3 +84,25 @@ def get_skills(self):
                     skills_aggregated.append(skill)
                     already_added_skills.add(skill['name'])
         return skills_aggregated
+
+
+class LearnerPathwayCourse(LearnerPathwayNode):
+
+    course = models.OneToOneField(Course, on_delete=models.CASCADE)
```
saleem-latif: Shouldn't it be 1-M field?

One course can appear in different pathways. but if we make it `OneToOneField` field then that won't be possible.

saleem-latif (changes requested):

course_discovery/apps/learner_pathway/tests.py
```diff
@@ -1,3 +1,55 @@
...
+        """ Verify that `LearnerPathwayCourse.get_estimated_time_of_completion` method is working as expected """
+
+        learner_pathway_course = LearnerPathwayCourse(course=self.course)
+        estimated_time_of_completion = learner_pathway_course.get_estimated_time_of_completion()
+        assert estimated_time_of_completion == get_course_run_estimated_hours(self.advertised_course_run)
+
+    def test_get_skills(self):
+        """ Verify that `LearnerPathwayCourse.get_skills` method is working as expected """
+
+        expected_skills = [{'name': 'skill name 1', 'description': 'skill description 1'}]
```
irfanuddinahmad, line 48: shouldn't this be just skill names?

muhammad-ammar: discussed on ticket and for now we will go with skill name and description.

saleem-latif (approved):

PR #3241, opened 2021-12-29
feat: csv data loader

DawoudSheraz:
### Description

This PR adds a management command that uses CSVDataLoader, also added via this PR, to import course and course information from CSV into discovery. For a single row, the CSV loader works as follows:

- Check the existence of the course and course run in the database. They should be draft if already present
- If the course and course run are not present, create an initial entry
- Update course and course run through separate API calls with their respective fields. 
- Mark the run as reviewed. This essentially publishes the course if applicable.

The loader uses existing course and course run view sets to make API calls to create or update objects. The consumption of APIs as opposed to writing ORM queries is that the CSV loaded is mimicking the way courses and course runs are authored on the publisher. Eventually, all the course/course runs will be accessible and editable from the publisher, so the data representation of API will help achieve that. Also, the APIs already have certain data validation and status update procedures in place which will ensure no bad/missing information is being ingested into discovery.

### Assumptions

The csv loader works under some assumptions, at least at the time of creating this PR:
- Course and course run should be draft if already present before ingesting.
- LevelType and related subjects should be present in the database
- Collaborators(for the course) and Staff(for course runs) are created during ingestion if not located within DB. Staff's name is saved only
- The dates are in UTC as compared to EST. This might require a change in the near future if the provided dates in the sample are in EST
- If the content and/or transcript language are not in IETF language format, course run data is not updated
- Some CSV keys might need an update once there is a sample data set
- The default course editor of the imported course will be the user associated with OAuth application used for making API calls.
 

### Testing

You can use the management command to check the loader, replace csv path value with a valid csv path.
```
python manage.py import_course_metadata --csv_path='/edx/app/discovery/discovery/course_discovery/apps/course_metadata/data_loaders/test.csv'
```

Make sure you have the following LevelType entries in the database:
- Introductory
- Advanced
- Intermediate

Also, add the following subjects to the DB, if not present:

- Mathematics
- Physics
- Computer Science
- Social Sciences

### Test CSVs
- [test.csv](https://github.com/openedx/course-discovery/files/7904309/test.csv)
- [test.csv](https://github.com/openedx/course-discovery/files/7904316/test.csv)

### Testing on Sandbox

dawoudsheraz.sandbox.edx.org/

- Download the CSV generator from ticket https://openedx.atlassian.net/browse/PROD-2613?focusedCommentId=589805
- Generate a new CSV
- Connect to VPN
- Copy the csv to sandbox via command `scp test.csv <github_user>@dawoudsheraz.sandbox.edx.org:/tmp`(Assuming command is run from directory where csv is present)
- SSh into sandbox `ssh <github_user>@dawoudsheraz.sandbox.edx.org:`. 
- cd `/edx/app/discovery/discovery`
- run `source ../venvs/discovery/bin/activate`
- Run `python manage.py import_course_metadata --csv_path='/tmp/test.csv'`
- Verify the data on https://discovery-dawoudsheraz.sandbox.edx.org/admin (email: admin@example.com, edx)
- Verify the data on LMS/Studio (admin@example.com, Hamilton)
- Verify data on ecommerce

### Todos
to be catered in follow-up PRs:

- Change price to proper enrollment track-type instead of hard-coding to verified
- Update how the staff/person ids are passed
- Update date parsing to EST once the time values are confirmed
- Add verified deadline override verification checks

Also, in the ticket, there is a script to generate CSV with dummy data. You should generate some other test csv using that script and verify the behavior.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -1,0 +1,398 @@
+"""
+Data loader responsible for creating course and course runs entries in Database
+provided a csv containing the required information.
+"""
+import csv
+import logging
+
+from django.conf import settings
+from django.db.models import Q
+from django.urls import reverse
+
+from course_discovery.apps.core.utils import serialize_datetime
+from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoader
+from course_discovery.apps.course_metadata.models import (
+    Collaborator, Course, CourseRun, CourseRunPacing, CourseRunType, CourseType, Organization, Person, ProgramType
+)
+from course_discovery.apps.course_metadata.utils import download_and_save_course_image
+from course_discovery.apps.ietf_language_tags.models import LanguageTag
+
+logger = logging.getLogger(__name__)
+
+
+class CSVDataLoader(AbstractDataLoader):
+    PROGRAM_TYPES = [
+        ProgramType.XSERIES,
+        ProgramType.MASTERS,
+        ProgramType.MICROMASTERS,
+        ProgramType.MICROBACHELORS,
+        ProgramType.PROFESSIONAL_PROGRAM_WL,
+        ProgramType.PROFESSIONAL_CERTIFICATE
+    ]
+
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None):
+        super().__init__(partner, api_url, max_workers, is_threadsafe)
+
+        try:
+            self.reader = csv.DictReader(open(csv_path, 'r'))
+        except FileNotFoundError:
+            logger.exception("Error opening csv file at path {}".format(csv_path))
+            raise  # re-raising exception to avoid moving the code flow
+
+    def ingest(self):  # pylint: disable=too-many-statements
+        logger.info("Initiating CSV data loader flow.")
+        for row in self.reader:
+
+            row = self.transform_dict_keys(row)
+            course_title = row['title']
+            org_key = row['organization']
+
+            logger.info('Starting data import flow for {}'.format(course_title))
+            if not Organization.objects.filter(key=org_key).exists():
+                logger.error("Organization {} does not exist in database. Skipping CSV loader for course {}".format(
+                    org_key,
+                    course_title
+                ))
+                continue
+
+            try:
+                course_type = CourseType.objects.get(name=row['course_enrollment_track'])
+                course_run_type = CourseRunType.objects.get(name=row['course_run_enrollment_track'])
+            except CourseType.DoesNotExist:
+                logger.exception("CourseType {} does not exist in the database.".format(
+                    row['course_enrollment_track']
+                ))
+                continue
+            except CourseRunType.DoesNotExist:
+                logger.exception("CourseRunType {} does not exist in the database.".format(
+                    row['course_run_enrollment_track']
+                ))
+                continue
+
+            course_key = self.get_course_key(org_key, row['number'])
+
+            course = Course.objects.filter_drafts(key=course_key, partner=self.partner).first()
+            if course:
+                course_run = CourseRun.objects.filter_drafts(course=course).first()
+                logger.info("Course {} is located in the database.".format(course_key))
+            else:
+                logger.info("Course key {} could not be found in database, creating the course.".format(course_key))
+                try:
+                    _ = self._create_course(row, course_type.uuid, course_run_type.uuid)
+                except Exception:  # pylint: disable=broad-except
+                    logger.exception("Error occurred when attempting to create a new course against key {}".format(
+                        course_key
+                    ))
+                    continue
+                course = Course.everything.get(key=course_key, partner=self.partner)
+                course_run = CourseRun.everything.filter(course=course).first()
+
+            is_downloaded = download_and_save_course_image(course, row['image'])
+            if not is_downloaded:
+                logger.error("Unexpected error happened while downloading image for course {}".format(
+                    course_key
+                ))
+                continue
+
+            try:
+                self._update_course(row, course)
+            except Exception:  # pylint: disable=broad-except
+                logger.exception("An unknown error occurred while updating course information")
+                continue
+
+            # No need to update the course run if the run is already in the review
+            if not course_run.in_review:
+                try:
+                    self._update_course_run(row, course_run)
+                    course_run.refresh_from_db()
+                except Exception:  # pylint: disable=broad-except
+                    logger.exception("An unknown error occurred while updating course run information")
+                    continue
+
+            try:
+                self._complete_run_review(row, course_run)
+            except Exception:  # pylint: disable=broad-except
+                logger.exception("An unknown error occurred while completing course run review")
+
+            logger.info("Course and course run updated successfully for course key {}".format(course_key))
+        logger.info("CSV loader ingest pipeline has completed.")
+
+    def _create_course_api_request_data(self, data, course_type_uuid, course_run_type_uuid):
+        """
+        Given a data dictionary, return a reduced data representation in dict
+        which will be used as input for course creation via course api.
+        """
+        pricing = self.get_pricing_representation(data['verified_price'])
+
+        # TODO: make appropriate timezone adjustment when it is confirmed if time values are in EST or UTC
+        course_run_creation_fields = {
+            'pacing_type': self.get_pacing_type(data['course_pacing']),
+            'start': self.get_formatted_datetime_string(f"{data['start_date']} {data['start_time']}"),
+            'end': self.get_formatted_datetime_string(f"{data['end_date']} {data['end_time']}"),
+            'run_type': str(course_run_type_uuid),
+            'prices': pricing,
+        }
+        return {
+            'org': data['organization'],
+            'title': data['title'],
+            'number': data['number'],
+            'type': str(course_type_uuid),
+            'prices': pricing,
+            'course_run': course_run_creation_fields
+        }
+
+    def _update_course_api_request_data(self, data, course):
+        """
+        Create and return the request data for making a patch call to update the course.
+        """
+        collaborator_uuids = self.process_collaborators(data['collaborators'], course.key)
+        subjects = self.get_subject_slugs(
+            data.get('primary_subject'),
+            data.get('secondary_subject'),
+            data.get('tertiary_subject')
+        )
+
+        update_course_data = {
+            'draft': False,
+            'key': course.key,
+            'uuid': str(course.uuid),
+            'url_slug': course.url_slug,
+            'type': str(course.type.uuid),
+            'subjects': subjects,
+            'collaborators': collaborator_uuids,
+            'prices': self.get_pricing_representation(data['verified_price']),
+
+            'title': data['title'],
+            'syllabus_raw': data['syllabus'],
+            'level_type': data['course_level'],
+            'outcome': data['what_will_you_learn'],
+            'faq': data['frequently_asked_questions'],
+            'video': {'src': data['about_video_link']},
+            'prerequisites_raw': data['prerequisites'],
+            'full_description': data['long_description'],
+            'short_description': data['short_description'],
+            'learner_testimonials': data['learner_testimonials'],
+            'additional_information': data['additional_information'],
+        }
+        return update_course_data
+
+    def _update_course_run_request_data(self, data, course_run):
+        """
+        Create and return the request data for making a patch call to update the course run.
+        """
+        program_type = data['expected_program_type']
+        staff_uuids = self.process_staff_names(data['staff'], course_run.key)
+        content_language = self.verify_and_get_language_tags(data['content_language'])
+        transcript_language = self.verify_and_get_language_tags(data['transcript_language'])
+
+        update_course_run_data = {
+            'run_type': str(course_run.type.uuid),
+            'key': course_run.key,
+            'prices': self.get_pricing_representation(data['verified_price']),
+            'staff': staff_uuids,
+            'draft': False,
+
+            'weeks_to_complete': data['length'],
+            'min_effort': data['minimum_effort'],
+            'max_effort': data['maximum_effort'],
+            'content_language': content_language[0],
+            'expected_program_name': data['expected_program_name'],
+            'transcript_languages': transcript_language,
+            'go_live_date': self.get_formatted_datetime_string(data['publish_date']),
+            'expected_program_type': program_type if program_type in self.PROGRAM_TYPES else None,
+            'upgrade_deadline_override': self.get_formatted_datetime_string(
+                f"{data['upgrade_deadline_override_date']} {data['upgrade_deadline_override_time']}"
+            ),
+        }
+        return update_course_run_data
+
+    def get_formatted_datetime_string(self, date_string):
+        """
+        Return the datetime string into the desired format %Y-%m-%dT%H:%M:%SZ
+        """
+        return serialize_datetime(self.parse_date(date_string))
+
+    def get_pacing_type(self, pacing):
+        """
+        Return appropriate pacing selection against a provided pacing string.
+        """
+        if pacing:
+            pacing = pacing.lower()
+
+        if pacing == 'instructor-paced':
+            return CourseRunPacing.Instructor
+        elif pacing == 'self-paced':
+            return CourseRunPacing.Self
+        else:
+            return None
+
+    def verify_and_get_language_tags(self, language_str):
+        """
+        Given a string of language tags or names, verify their existence in the database
+        and return a list of language codes.
+        """
+        languages_codes_list = []
+        languages_list = language_str.split(',')
+        for language in languages_list:
+            language = language.strip()
+            language_obj = LanguageTag.objects.filter(
+                Q(name=language) | Q(code=language)
+            ).first()
+            if not language_obj:
+                raise Exception(
+                    'Language {} from provided string {} is either missing or an invalid ietf language'.format(
+                        language, language_str
+                    )
+                )
+            languages_codes_list.append(language_obj.code)
+        return languages_codes_list
+
+    def _call_course_api(self, method, url, data):
+        """
+        Helper method to make course and course run api calls.
+        """
+        response = self.api_client.request(
+            method,
+            url,
+            json=data,
+            headers={'content-type': 'application/json'}
+        )
+        response.raise_for_status()
+        return response
+
+    def _create_course(self, data, course_type_uuid, course_run_type_uuid):
+        """
+        Make a course entry through course api.
+        """
+        course_api_url = reverse('api:v1:course-list')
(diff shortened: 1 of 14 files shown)


course_discovery/apps/course_metadata/utils.py
```diff
@@ -683,3 +685,31 @@ def clean_html(content):
...
+
+def download_and_save_course_image(course, image_url):
+    """
+    Helper method to download an image from a provided image url and save it
+    as course card image.
+    """
+    try:
+        response = requests.get(image_url)
+
+        if response.status_code == requests.codes.ok:  # pylint: disable=no-member
```
Ali-D-Akbar, line 697: is this different from rest_framework.status? If not, we can use them instead.

DawoudSheraz:
cc: @ziafazal ready for review

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -1,9 +1,15 @@
 from unittest import mock
 
 import responses
+import six
```
ziafazal: Using `six` does not make sense in in python 3.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,394 @@
...
+    def ingest(self):  # pylint: disable=too-many-statements
+        logger.info("Initiating CSV data loader flow.")
+        for row in self.reader:
+
+            row = self.transform_dict_keys(row)
+            course_title = row['title']
+            org_key = row['organization']
+
+            logger.info('Starting data import flow for {}'.format(course_title))
+            if not Organization.objects.filter(key=org_key).exists():
```
ziafazal, line 51: Should not we create the organization if it does not exist?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,394 @@
...
+        Given a comma-separated string of collaborator names, return the list of collaborator
+        uuids after processing.
+
+        Processing involves the following
+            * Checking if the collaborator value is valid
+            * Checking for existence of collaborator in DB
+            * Create collaborator if not present
+        """
+        collaborators = collaborators.split(',')
+        collaborators = [collaborator for collaborator in collaborators if collaborator.strip()]
```
ziafazal: nit: strip the value being added in the list `[collaborator.stript() for collaborator in collaborators if collaborator.strip()]`

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,394 @@
...
+                logger.exception("CourseRunType {} does not exist in the database.".format(
+                    row['course_run_enrollment_track']
+                ))
+                continue
+
+            course_key = self.get_course_key(org_key, row['number'])
+
+            try:
+                # The imported courses and course runs will not be draft.
+                course = Course.everything.get(key=course_key, partner=self.partner)
```
ziafazal: I think we should not use `get` with `everything` here and in statements below it sometimes return more than one record.
```
(Pdb) Course.everything.get(key=course_key, partner=self.partner)
*** course_discovery.apps.course_metadata.models.Course.MultipleObjectsReturned: get() returned more than one Course -- it returned 2!
(Pdb) Course.objects.get(key=course_key, partner=self.partner)
<Course: edX+MDVLL: id porro>
```

ziafazal (changes requested):
@DawoudSheraz I have completed first round of review overall looks good. I have few suggestions. 
One thing I have noticed if we have invalid data in CSV we can get stuck. For example the csv I used had **Upgrade Deadline Override Date** set to 06/23/2022 which resulted in this error
```Failed to publish commerce data for course-v1:edX+1RN8K+1T2022 to LMS. Verification deadline must be after the course mode upgrade deadlines.```
I then updated the **Upgrade Deadline Override Date** to 02/22/2022 to make it prior to **End Date** but the error did not go away(possibly because we can't update it from publisher APIs).
I think we should add some data validation in our loader to avoid such errors.

File: course_discovery/apps/course_metadata/data_loaders/tests/mixins.py (hunk 2 above)
Comment (DawoudSheraz) [reply course_discovery/apps/course_metadata/data_loaders/tests/mixins.py]: Right, I will update this.

File: course_discovery/apps/course_metadata/data_loaders/csv_loader.py (hunk 3 above)
Comment (DawoudSheraz) [reply course_discovery/apps/course_metadata/data_loaders/csv_loader.py:51]: Since the organization has other fields too(logo, description), we should not create org if not found.

File: course_discovery/apps/course_metadata/data_loaders/csv_loader.py (hunk 5 above)
Comment (DawoudSheraz) [reply course_discovery/apps/course_metadata/data_loaders/csv_loader.py]: Yes, I have noted that down in https://openedx.atlassian.net/browse/PROD-2618. However, if the course is present in DB, what variant should we get ideally? Draft or live? The choice here matters. When filtering with `objects`, it filters draft out of queryset. That's why I was using `everything`

**Comment by DawoudSheraz** [comment]
> @DawoudSheraz I have completed first round of review overall looks good. I have few suggestions. One thing I have noticed if we have invalid data in CSV we can get stuck. For example the csv I used had **Upgrade Deadline Override Date** set to 06/23/2022 which resulted in this error `Failed to publish commerce data for course-v1:edX+1RN8K+1T2022 to LMS. Verification deadline must be after the course mode upgrade deadlines.` I then updated the **Upgrade Deadline Override Date** to 02/22/2022 to make it prior to **End Date** but the error did not go away(possibly because we can't update it from publisher APIs). I think we should add some data validation in our loader to avoid such errors.

Yes, already noted on https://openedx.atlassian.net/browse/PROD-2618. However, we are not sure if the ExEd courses will have this verification deadline override applicable. Once we are certain of that, we can add appropriate checks where needed.

File: course_discovery/apps/course_metadata/data_loaders/csv_loader.py [hunk 6]
```diff
@@ -0,0 +1,398 @@
+"""
+Data loader responsible for creating course and course runs entries in Database
```
azanbinzahid: We may add more along these lines: ...Database, push to studio and other services

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+        ProgramType.MICROBACHELORS,
+        ProgramType.PROFESSIONAL_PROGRAM_WL,
+        ProgramType.PROFESSIONAL_CERTIFICATE
+    ]
+
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None):
+        super().__init__(partner, api_url, max_workers, is_threadsafe)
+
+        try:
+            self.reader = csv.DictReader(open(csv_path, 'r'))
```
azanbinzahid, line 37: do we have a plan to sanitize csv input?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+                logger.exception("CourseRunType {} does not exist in the database.".format(
+                    row['course_run_enrollment_track']
+                ))
+                continue
+
+            course_key = self.get_course_key(org_key, row['number'])
+
+            course = Course.objects.filter_drafts(key=course_key, partner=self.partner).first()
+            if course:
+                course_run = CourseRun.objects.filter_drafts(course=course).first()
```
azanbinzahid, line 76: do we need to check if `course_run` is found or not?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+            'draft': False,
+
+            'weeks_to_complete': data['length'],
+            'min_effort': data['minimum_effort'],
+            'max_effort': data['maximum_effort'],
+            'content_language': content_language[0],
+            'expected_program_name': data['expected_program_name'],
+            'transcript_languages': transcript_language,
+            'go_live_date': self.get_formatted_datetime_string(data['publish_date']),
+            'expected_program_type': program_type if program_type in self.PROGRAM_TYPES else None,
```
azanbinzahid, line 202: What will `None` represent here?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+        """
+        languages_codes_list = []
+        languages_list = language_str.split(',')
+        for language in languages_list:
+            language = language.strip()
+            language_obj = LanguageTag.objects.filter(
+                Q(name=language) | Q(code=language)
+            ).first()
+            if not language_obj:
+                raise Exception(
```
azanbinzahid, line 242: where will this exception get caught?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+            'course_run': course_run_creation_fields
+        }
+
+    def _update_course_api_request_data(self, data, course):
+        """
+        Create and return the request data for making a patch call to update the course.
+        """
+        collaborator_uuids = self.process_collaborators(data['collaborators'], course.key)
+        subjects = self.get_subject_slugs(
+            data.get('primary_subject'),
```
azanbinzahid, line 150: do we need to add default return?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+    def get_pricing_representation(self, verified_price):
+        """
+        Return dict representation of prices.
+        """
+        return {
+            'verified': verified_price,  # TODO: temporary value to verified
+            # Actual value from course type -> entitlement_tracks -> slugs
+        }
+
+    def get_subject_slugs(self, *subjects):
```
azanbinzahid, line 342: do we need to verify value/entry from subject model?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+        Given a comma-separated string of collaborator names, return the list of collaborator
+        uuids after processing.
+
+        Processing involves the following
+            * Checking if the collaborator value is valid
+            * Checking for existence of collaborator in DB
+            * Create collaborator if not present
+        """
+        collaborators = collaborators.split(',')
+        collaborators = [collaborator.strip() for collaborator in collaborators if collaborator.strip()]
```
azanbinzahid, line 361: why we are striping twice?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -0,0 +1,398 @@
...
+            * Checking for existence of staff member in DB
+            * Create staff if not present
+        """
+
+        staff_names_list = staff_names.split(',')
+        staff_names_list = [staff_name for staff_name in staff_names_list if staff_name.strip()]
+        staff_uuids = []
+
+        # TODO: This is a fragile approach. It is possible for two people to have same name within a partner.
+        # TODO: CSV would need to provide more information to identify staff members from other than names
```
azanbinzahid, line 386: For this I guess we are planning to add staff in separate function, right.

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -35,3 +40,193 @@ def assert_api_called(self, expected_num_calls, check_auth=True):
...
+
+    def _write_csv(self, csv, lines_dict_list):
+        """
+        Helper method to write given list of data dictionaries to csv, including the csv header.
+        """
+        header = ''
+        lines = ''
+        for key in self.CSV_DATA_KEYS_ORDER:
+            title_case_key = key.replace('_', ' ').title()
+            header = '{}{},'.format(header, title_case_key)
```
azanbinzahid, line 115: we can append header names in list and then use `",".join(header)`

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -35,3 +40,193 @@ def assert_api_called(self, expected_num_calls, check_auth=True):
...
+        header = f"{header[:-1]}\n"
+
+        for line_dict in lines_dict_list:
+            for key in self.CSV_DATA_KEYS_ORDER:
+                lines = '{}"{}",'.format(lines, line_dict[key])
+            lines = f"{lines[:-1]}\n"
```
azanbinzahid, line 121: same impl as in above comment

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3070,3 +3070,65 @@
     'uuid': '6b8b779f-f567-4e98-aa41-a265d6fa073a',
     'vuuid': 'e0f8c80a-b377-4546-b247-1c94ab3a218a'
 }
+
+VALID_COURSE_AND_COURSE_RUN_CSV_DICT = {
```
azanbinzahid, line 3074: in dates please use date greater than 12 so that the format is understood easily. e.g. `12/25/2022`

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: Sure.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: https://openedx.atlassian.net/browse/PROD-2615 is responsible for this, once this PR merges.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: The course creation call creates one run for it too. Therefore, the course run will be present. But I guess having a check would not hurt.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: For serializer to ignore this value.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: I think the function will be returning the empty list if none of the subject entries are present.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: https://github.com/openedx/course-discovery/pull/3241/files#diff-c981dbbc87edba29cfa41634f17d194a25a3bd209560c130628f027714f2c7d9R109

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: Ah, this is a tricky situation here. The subject objects should be present beforehand in the DB. If that assumption breaks, then yes. We can cater it in input validation ticket(shared above)

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: For both if check and the assigned value.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: Yeah, this is here for an initial approach to adding staff.

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
DawoudSheraz: Yeah, I can check it.

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
DawoudSheraz: I will give it a try

course_discovery/apps/course_metadata/utils.py
DawoudSheraz: How?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -682,3 +684,33 @@ def clean_html(content):
...
+            content_type = response.headers['Content-Type'].lower()
+            extension = IMAGE_TYPES.get(content_type)
+
+            if extension:
+                filename = '{uuid}.{extension}'.format(uuid=str(course.uuid), extension=extension)
+                course.image.save(filename, ContentFile(response.content))
```
waheedahmed, line 703: Course image field has some specific criteria, for example:

1. Image must be 1134x675 pixels in size
2. It must be JPEG or PNG
3. etc.

Are we handling those?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -682,3 +684,33 @@ def clean_html(content):
     cleaned = markdown.markdown(cleaned)
 
     return cleaned
+
+
+def download_and_save_course_image(course, image_url):
```
waheedahmed, line 689: Looks like specific to the data loader only, can we make it a class function instead?

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
DawoudSheraz: This was causing an error due to some headers missing out in the CSV. Plus, we still need the for loop to convert the key from snake_case to Title case.

azanbinzahid: This should work but not so important at this point.

```
        header = []
        for key in self.CSV_DATA_KEYS_ORDER:
            title_case_key = key.replace('_', ' ').title()
            header.append(title_case_key)
        
        header = ",".join(header)
```

ziafazal (approved):
👍

waheedahmed (commented):
A couple of questions otherwise looks good!

course_discovery/apps/course_metadata/utils.py
DawoudSheraz: Actually, no, I made this util function by moving out download and save logic from the management command https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/download_course_images.py.

course_discovery/apps/course_metadata/utils.py
DawoudSheraz: Valid point, we can add those checks in a follow-up PR, covered under https://openedx.atlassian.net/browse/PROD-2615

waheedahmed (approved):

PR #3248, opened 2022-01-13
refactor: remove unused old slug field from course

waheedahmed:
Deleted disco’s old course URL field (slug) from the model and all its usages.

DISCO-1770


brianhw:
Note: it would probably be good to check in with #data-support (or file a BI ticket) to check on how this field is being used in Snowflake.

attiyaIshaque (approved):

attiyaIshaque (approved):
After the build is green.

PR #3251, opened 2022-01-15
feat: ENT-5274 Augmented the pathway models with new fields

irfanuddinahmad:
**Description**: This PR updated the learner pathway and pathway step models with the following new fields:
                            Status: active=published, inactive=not published
                            Scope: single enterprise, multiple enterprise, all enterprises
                            Enterprise Customer: UUID(s) of customers who should see this pathway. If blank, no one sees it. If blank AND type=all enterprises, all enterprises see it.
                            Banner image:  Banner image.
                            Overview: The text field to describe the pathway.
                            Minimum Requirement: Minimum number of nodes required for step completion.

**JIRA**: [ENT-5274](https://openedx.atlassian.net/browse/ENT-5274)


course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,47 @@ def get_node(cls, uuid):
...
+   Fields:
+       uuid (UUIDField, PRIMARY KEY): Enterprise Customer code
+
+   .. no_pii:
+   """
+    class Meta:
+        verbose_name = _("Enterprise Customer")
+        verbose_name_plural = _("Enterprise Customers")
+
+    uuid = models.UUIDField(primary_key=True, default=uuid4, editable=False)
```
saleem-latif: We should not have a default for this.

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,47 @@ def get_node(cls, uuid):
...
 class LearnerPathway(models.Model):
     """
     Top level model for learner pathway.
     """
     uuid = models.UUIDField(default=uuid4, editable=False, unique=True, verbose_name=_('UUID'))
     name = models.CharField(max_length=255, null=False, blank=False, help_text=_('Pathway name'))
-
+    status = models.CharField(
+        help_text=_('The active/inactive status of this Pathway.'),
+        max_length=16, null=False, blank=False, default=PathwayStatus.Inactive,
```
saleem-latif: I think the default `null=False` and `blank=False` are default values so do not need to pass them explicitly.

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,47 @@ def get_node(cls, uuid):
...
+        choices=PathwayStatus.choices, validators=[PathwayStatus.validator]
+    )
+    scope = models.CharField(
+        help_text=_('The enterprise scope(single, multiple, all) of this Pathway.'),
+        max_length=16, null=False, blank=False, default=PathwayScope.Single,
```
saleem-latif: I think the default `null=False` and `blank=False` are default values so do not need to pass them explicitly.

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,47 @@ def get_node(cls, uuid):
...
+        choices=PathwayScope.choices, validators=[PathwayScope.validator]
+    )
+    enterprise_customer_uuids = models.ManyToManyField(
+        EnterpriseCustomer,
+        related_name='enterprise_uuids',
```
saleem-latif: I think related name here should be `learner_pathways`

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,47 @@ def get_node(cls, uuid):
...
+        blank=True,
+        help_text=("List of enterprises that can access the pathway. If blank, no one sees it. "
```
saleem-latif: nit: change indentation to 
```
 help_text=_(
    "List of enterprises that can access the pathway. If blank, no one sees it. If blank AND scope=all, all enterprises get access"
)
 


```

saleem-latif: ref: https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#breaking-long-lines

course_discovery/apps/learner_pathway/models.py
```diff
@@ -116,6 +150,20 @@ def __repr__(self):
 class LearnerPathwayStep(models.Model):
     uuid = models.UUIDField(default=uuid4, editable=False, unique=True, verbose_name=_('UUID'))
     pathway = models.ForeignKey(LearnerPathway, related_name='steps', on_delete=models.CASCADE)
+    min_requirement = models.PositiveSmallIntegerField(
+        blank=True,
```
saleem-latif: Wouldn't it be better to skip `blank=True` since this column has a default value?

saleem-latif (changes requested):

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,49 @@ def get_node(cls, uuid):
...
+    enterprise_customer_uuids = models.ManyToManyField(
+        EnterpriseCustomer,
+        related_name='learner_pathways',
+        blank=True,
+        help_text=(
+            "List of enterprises that can access the pathway. If blank, no one sees it. "
+            "If blank AND scope=all, all enterprises get access"
+        )
+    )
+    banner_image_url = models.URLField(blank=True, null=True)
```
muhammad-ammar: How the image will be uploaded to S3 bucket?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -116,6 +152,19 @@ def __repr__(self):
 class LearnerPathwayStep(models.Model):
     uuid = models.UUIDField(default=uuid4, editable=False, unique=True, verbose_name=_('UUID'))
     pathway = models.ForeignKey(LearnerPathway, related_name='steps', on_delete=models.CASCADE)
+    min_requirement = models.PositiveSmallIntegerField(
+        default=1,
+        validators=[
+            MinValueValidator(1)
```
muhammad-ammar, line 144: Can you please explain how this validation works?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,49 @@ def get_node(cls, uuid):
...
+   Fields:
+       uuid (UUIDField, PRIMARY KEY): Enterprise Customer code
+
+   .. no_pii:
+   """
+    class Meta:
+        verbose_name = _("Enterprise Customer")
+        verbose_name_plural = _("Enterprise Customers")
+
+    uuid = models.UUIDField(primary_key=True, editable=False)
```
muhammad-ammar: How the data will be populated in this model?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,49 @@ def get_node(cls, uuid):
...
     name = models.CharField(max_length=255, null=False, blank=False, help_text=_('Pathway name'))
-
+    status = models.CharField(
+        help_text=_('The active/inactive status of this Pathway.'),
+        max_length=16, default=PathwayStatus.Inactive,
+        choices=PathwayStatus.choices, validators=[PathwayStatus.validator]
+    )
+    scope = models.CharField(
+        help_text=_('The enterprise scope(single, multiple, all) of this Pathway.'),
+        max_length=16, default=PathwayScope.Single,
```
muhammad-ammar: How we will ensure that in case of `single` scope, a pathway can be associated with multiple `enterprise_customer_uuids`?

course_discovery/apps/learner_pathway/choices.py
```diff
@@ -0,0 +1,12 @@
...
+from djchoices import ChoiceItem, DjangoChoices
+
+
+class PathwayStatus(DjangoChoices):
+    Active = ChoiceItem('active', _('Active'))
+    Inactive = ChoiceItem('inactive', _('Inactive'))
+
+class PathwayScope(DjangoChoices):
+    Single = ChoiceItem('single', _('Single'))
+    Multiple = ChoiceItem('multiple', _('Multiple'))
```
muhammad-ammar: Do we have a definition of `multiple`?  Do we have a `max`?

course_discovery/apps/learner_pathway/models.py
muhammad-ammar: Shouldn't this be a required field?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,49 @@ def get_node(cls, uuid):
...
+        EnterpriseCustomer,
+        related_name='learner_pathways',
+        blank=True,
+        help_text=(
+            "List of enterprises that can access the pathway. If blank, no one sees it. "
+            "If blank AND scope=all, all enterprises get access"
+        )
+    )
+    banner_image_url = models.URLField(blank=True, null=True)
+    overview = models.TextField(null=True, blank=True)
```
muhammad-ammar: Shouldn't this be a required field?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -116,6 +152,19 @@ def __repr__(self):
...
     pathway = models.ForeignKey(LearnerPathway, related_name='steps', on_delete=models.CASCADE)
+    min_requirement = models.PositiveSmallIntegerField(
+        default=1,
+        validators=[
+            MinValueValidator(1)
+        ],
+        help_text=_('Minimum number of nodes to complete this step of the pathway')
+    )
+
+    def clean(self):
```
muhammad-ammar, line 149: Please note the model clean method will only be called when saving a form. It will not be called when we save a model using model `save` method manually. Is this expected?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -116,6 +152,19 @@ def __repr__(self):
...
+        # Ensure that the minimum requirement is not more than the number of nodes
+        node_count = len(self.get_nodes())
+        if self.min_requirement > node_count:
+            self.min_requirement = node_count
```
muhammad-ammar: I think this will be confusing. As per my understanding, min requirement should be updated explicitly only through field.

course_discovery/apps/learner_pathway/models.py
muhammad-ammar: Also what about the backend call mentioned in the ticket?

muhammad-ammar (changes requested):
I am done with first pass of the review. Let me know once the PR is ready for second review.

course_discovery/apps/learner_pathway/models.py
```diff
@@ -69,14 +70,49 @@ def get_node(cls, uuid):
...
+    """
+   Enterprise Customer as defined and maintained in the edx-enterprise repository
+
+   Fields:
+       uuid (UUIDField, PRIMARY KEY): Enterprise Customer code
+
+   .. no_pii:
+   """
+    class Meta:
+        verbose_name = _("Enterprise Customer")
```
saleem-latif: nit: it would be better to be consistent and use single quoted string everywhere

saleem-latif (approved):
One more comment, other than that looks good 👍🏼

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: replaced with imagefield

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: MinValueValidator¶
class MinValueValidator(limit_value, message=None)¶
Raises a ValidationError with a code of 'min_value' if value is less than limit_value, which may be a callable.
https://docs.djangoproject.com/en/4.0/ref/validators/

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: Coordinating with ENT-5294 assignee. Separate PR required for the work to fetch the uuids from edx-enterprise

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: To be managed at the form level.

course_discovery/apps/learner_pathway/choices.py
irfanuddinahmad: No max. Multiple is any number greater than 1.

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: done

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: No

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: Can you suggest an alternative?

course_discovery/apps/learner_pathway/models.py
irfanuddinahmad: Raised a ValidationError

course_discovery/apps/learner_pathway/utils.py
```diff
@@ -13,3 +14,25 @@ def get_advertised_course_run_estimated_hours(course):
                 return get_course_run_estimated_hours(course_run)
 
     return None
+
+
+class UploadToFieldNamePath(FilePattern):
```
muhammad-ammar: This class is already present [here](https://github.com/openedx/course-discovery/blob/f51ac139fb68ad901878eee0c62cae3551a45b11/course_discovery/apps/course_metadata/utils.py#L296-L315) We can use the that in our app instead of duplicating.

course_discovery/apps/learner_pathway/models.py
```diff
@@ -116,6 +138,19 @@ def __repr__(self):
...
+            MinValueValidator(1)
+        ],
+        help_text=_('Minimum number of nodes to complete this step of the pathway')
+    )
+
+    def clean(self):
+        # Ensure that the minimum requirement is not more than the number of nodes
+        node_count = len(self.get_nodes())
+        if self.min_requirement > node_count:
+            raise ValidationError(_('Please use a minimum requirement less than the total number of requirements.'))
```
muhammad-ammar, line 153: How about `Minimum requirement can not be greater than the total number of steps in the pathway`?

irfanuddinahmad: then we have to say minimum steps instead of minimum requirement ...

muhammad-ammar (approved):

PR #3256, opened 2022-01-20
fix: textual improvements

muhammad-ammar:


course_discovery/apps/learner_pathway/models.py
```diff
@@ -99,6 +99,10 @@ class LearnerPathway(models.Model):
     )
     overview = models.TextField(blank=True)
 
+    class Meta:
```
saleem-latif, line 102: I think you might need to create migrations for this. Can you check if running `makemigrations` creates any migration files?

saleem-latif (approved):
Just one comment, other than looks good : 👍

course_discovery/apps/learner_pathway/models.py
muhammad-ammar: Great catch. Thank You!

PR #3257, opened 2022-01-21
feat: added course_run's availability property in program detailed serializers

HammadAhmadWaqas:
`availability`  property is available in course_run, course, and search/all endpoints but it is missing in the program detail endpoint. It is required in the programe endpoint as well to determine the program's availability.


saleem-latif (approved):
Looks Good 👍🏼

waheedahmed (approved):

PR #3266, opened 2022-01-26
feat: add additional metadata model in course metadata

azanbinzahid:
## Description:
This PR adds `AdditionalMetadata` model in course_metadata app to hold data related to 2U courses. The model is using `ForeignKey` relation instead of  `OneToOne` because of this [ADR](https://github.com/openedx/course-discovery/blob/master/docs/decisions/0004-publisher-draft.rst#consequences). 

## Linked Ticket:
[PROD-2647](https://openedx.atlassian.net/browse/PROD-2647)


course_discovery/apps/course_metadata/models.py
```diff
@@ -783,6 +792,11 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
     extra_description = models.ForeignKey(
         AdditionalPromoArea, models.CASCADE, default=None, null=True, blank=True, related_name='extra_description',
     )
+    additional_metadata = models.ForeignKey(
+        AdditionalMetadata, models.CASCADE,
+        related_name='additional_metadata_for_course',
```
DawoudSheraz: the related name will use by **AdditionalMetadata** model object to reference the courses it is linked to. A sample `courses` or `related_courses` will be semantically correct.

course_discovery/apps/course_metadata/models.py
```diff
@@ -588,7 +588,16 @@ class Meta:
...
+class AdditionalMetadata(TimeStampedModel):
+    """
+    This model (will) hold(s) 2U related additional fields
+    """
+
+    external_identifier = models.CharField(max_length=255, blank=True, null=True)
+    external_url = models.URLField(blank=True, null=True)
 
+    def __str__(self):
+        return self.external_identifier
```
DawoudSheraz: add external url alongside identifier too.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -404,6 +408,15 @@ def courses(self, obj):
...
         ])
 
+@admin.register(AdditionalMetadata)
+class AdditionalMetadataAdmin(admin.ModelAdmin):
+    list_display = ('external_identifier', 'external_url', 'courses')
+    search_fields = ('external_identifier',)
+
+    def courses(self, obj):
+        return ', '.join([
+            course.key for course in obj.additional_metadata_for_course.all()
```
DawoudSheraz: once the related name is changed, you will need to change the reference here too.

course_discovery/apps/course_metadata/models.py
```diff
@@ -783,6 +795,11 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
     extra_description = models.ForeignKey(
         AdditionalPromoArea, models.CASCADE, default=None, null=True, blank=True, related_name='extra_description',
     )
+    additional_metadata = models.ForeignKey(
```
ziafazal, line 798: @azanbinzahid I think one course should have one addional metadata.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -261,7 +261,7 @@ def test_list_query(self):
 
         # Known to be flaky prior to the addition of tearDown()
         # and logout() code which is the same number of additional queries
-        with self.assertNumQueries(48):
+        with self.assertNumQueries(51):
```
ziafazal, line 264: why these query count increased?

course_discovery/apps/api/serializers.py
```diff
@@ -1158,6 +1167,23 @@ def get_skills(self, obj):
...
+        external_identifier = additional_metadata.get('external_identifier')
+
+        if external_url:
+            additional_metadata, _ = AdditionalMetadata.objects.get_or_create(external_url=external_url)
+            if external_identifier:
+                additional_metadata.external_identifier = external_identifier
+            instance.additional_metadata = additional_metadata
+            instance.save()
+
+    def update(self, instance, validated_data):
```
ziafazal, line 1181: Could you please explain why we have to override `update` method here?

ziafazal (commented):
@azanbinzahid Thanks for making these changes I have posted few questions.

course_discovery/apps/course_metadata/models.py
azanbinzahid: Since each course has a (duplicate) draft version of it, therefore OneToOne relation was raising issues as specified in the consequences of this [ADR](https://github.com/openedx/course-discovery/blob/master/docs/decisions/0004-publisher-draft.rst#consequences).

course_discovery/apps/api/serializers.py
azanbinzahid: Update method was not supporting nested serialized objects update via `API` by default and therefore needs to be override. [Ref](https://www.django-rest-framework.org/api-guide/serializers/#writing-update-methods-for-nested-representations).

course_discovery/apps/api/v1/tests/test_views/test_courses.py
azanbinzahid: I believe this due to the addition of new model and its `prefetch_queryset`. The count increased in same pattern as for the addition of similar fields in model.

course_discovery/apps/course_metadata/models.py
```diff
@@ -589,6 +589,18 @@ class Meta:
         verbose_name = _('Topic model translations')
 
 
+class AdditionalMetadata(TimeStampedModel):
+    """
+    This model holds 2U related additional fields
+    """
+
+    external_url = models.URLField(blank=False, null=False)
+    external_identifier = models.CharField(max_length=255, blank=True, null=False)
```
ziafazal, line 598: I think if we allow blank then null should be `True` to store blank values as `NULL'

ziafazal (approved):
👍

course_discovery/apps/course_metadata/models.py
uzairr: by default value is `False` for both blank and null

azanbinzahid: It violates `semgrep` rule as follows:

image

uzairr (approved):
just one update, otherwise LGTM 👍

DawoudSheraz (approved):

PR #3272, opened 2022-02-02
Bpant/ent 5221 add fields to all programs

binodpant:
The api/v1/programs/ endpoint currently does not list these fields

`price_ranges`
`expected_learning_items`

We would like to add these fields to the endpoint serializer so that we can index those in algolia for use in enterprise public catalog MFE

### This PR contains

A change to 'move' the fields from the programSerializer to the MinimalProgramSerializer. The former already utilizes all the fields in Meta from the latter so both endpoint should now report these fields

JIRA: ENT-5221

### Alternative considered: 
We could, from enterprise catalog, call the `api/v1/programs/<uuid>` for each program item, but that would hammer discovery a lot more, than this trade off of increased payload size for the list endpoint

### Testing notes
Locally visited 
http://localhost:18381/api/v1/programs/2199b158-eff7-4c16-b5a6-0f48570f1d0e/

and 

http://localhost:18381/api/v1/programs

and verified these fields exist in both responses


binodpant:
it has been recommended to make the fields optionally returned only upon request, will work on that change

PR #3283, opened 2022-02-08
feat: new optional query param that extends fields 

alex-sheehan-edx:
when `extra_fields` is passed as a query param to the program list view, it will include two additional fields of data- `expected_learning_items` and `price_ranges`

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1668,6 +1668,52 @@ class MinimalProgramSerializer(FlexFieldsSerializerMixin, BaseModelSerializer):
         return obj.card_image_url
 
 
+class MinimalExtendedProgramSerializer(MinimalProgramSerializer):
+    """
+    Extended minimal program serializer.
+
+    Identical to the `MinimalProgramSerializer` except with two additional fields
+
+    When using the FlexFieldsSerializerMixin to get the courses field on a program,
+    you will also need to include the fields you want on the course object
+    since the course serializer also uses drf_dynamic_fields.
+    Eg: ?fields=courses,course_runs
+    """
+
+    authoring_organizations = MinimalOrganizationSerializer(many=True)
+    banner_image = StdImageSerializerField()
+    courses = serializers.SerializerMethodField()
+    type = serializers.SlugRelatedField(slug_field='name_t', queryset=ProgramType.objects.all())
+    type_attrs = ProgramTypeAttrsSerializer(source='type')
+    degree = DegreeSerializer()
+    curricula = CurriculumSerializer(many=True)
+    card_image_url = serializers.SerializerMethodField()
+    expected_learning_items = serializers.SlugRelatedField(many=True, read_only=True, slug_field='value')
+
+    @classmethod
+    def prefetch_queryset(cls, partner, queryset=None):
+        # Explicitly check if the queryset is None before selecting related
+        queryset = queryset if queryset is not None else Program.objects.filter(partner=partner)
+
+        return queryset.select_related('type', 'partner').prefetch_related(
+            'excluded_course_runs',
+            'expected_learning_items',
+            # `type` is serialized by a third-party serializer. Providing this field name allows us to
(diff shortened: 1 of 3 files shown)


waheedahmed (changes requested):
Instead of copying all the code from`MinimalProgramSerializer` can we simply extend this new class with it and just define the extra fields?

course_discovery/apps/api/v1/views/programs.py
```diff
@@ -30,6 +30,8 @@ class ProgramViewSet(CompressedCacheResponseMixin, viewsets.ReadOnlyModelViewSet
 
     def get_serializer_class(self):
         if self.action == 'list':
+            if hasattr(self, 'request') and self.request.query_params.get('extra_fields'):
```
waheedahmed: nit: can we rename the query param to `extended` instead? `extra_fields` seems vague to me.

waheedahmed: And, I don't think we need `hasattr(self, 'request')` check since it's a DRF view so it should always be available.

alex-sheehan-edx: there's a test that doesn't include the request object so I added it as a buffer

alex-sheehan-edx: https://github.com/openedx/course-discovery/pull/3283/files#diff-cd4cde02ab46f0b84879b2513d7e2fa7c18b12a7fefc4c27ee6d2f4880bbc4a5R351-R353 this is the problematic test

course_discovery/apps/api/serializers.py
```diff
@@ -1668,6 +1668,52 @@ def get_card_image_url(self, obj):
         return obj.card_image_url
 
 
+class MinimalExtendedProgramSerializer(MinimalProgramSerializer):
+    """
+    Extended minimal program serializer.
+
+    Identical to the `MinimalProgramSerializer` except with two additional fields
+
+    When using the FlexFieldsSerializerMixin to get the courses field on a program,
```
binodpant, line 1677: do we actually use this mixin? is this doc line up to date?

alex-sheehan-edx: I wanted to have all the same functionality as the minimal serializer which uses that mixin.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -210,6 +210,21 @@ def test_list(self):
 
         self.assert_list_results(self.list_path, expected, 19)
 
+    def test_extended_query_param_fields(self):
+        """ Verify that the `extra_fields` query param will result in an extended amount of fields returned. """
```
binodpant: change `extra_fields` to `extended`

course_discovery/apps/api/serializers.py
```diff
@@ -1668,6 +1668,52 @@ def get_card_image_url(self, obj):
...
+    you will also need to include the fields you want on the course object
+    since the course serializer also uses drf_dynamic_fields.
+    Eg: ?fields=courses,course_runs
```
binodpant, line 1680: discussed offline with Alex. Even though the filters options does work for some fields appraently it does not seem to work for courses. Hence we can't use it as is rightnow and this seems to be right approach for the time being. We can revisit if we can make the filters mixin work in lieu of this serializer in the future

binodpant (approved):
lgtm, tested locally by pulling branch

course_discovery/apps/api/v1/views/programs.py
waheedahmed: If there is only one test, could we just pass the request there? I believe we should not add checks to fix some tests and instead correct them.

waheedahmed (approved):
One last suggestion otherwise 👍

course_discovery/apps/api/v1/views/programs.py
alex-sheehan-edx: I totally agree! Let me see what I can do.

PR #3288, opened 2022-02-14
feat: add external identifier in csv loader

DawoudSheraz:
### Description
- Add external identifier in additional metadata payload in CSV Data loader
- Fix the additional metadata object not saving the updates to the external identifier by calling explicit save(though I am not sure if that is a best practice endorsed by DRF. Tried Google but could not find any helpful link)
- Updated the tests

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1175,7 +1175,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
             additional_metadata, _ = AdditionalMetadata.objects.get_or_create(external_url=external_url)
             if external_identifier:
                 additional_metadata.external_identifier = external_identifier
-            additional_metadata.save()
+                additional_metadata.save()
             instance.additional_metadata = additional_metadata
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1175,6 +1175,7 @@ def update_additional_metadata(self, instance, additional_metadata):
             additional_metadata, _ = AdditionalMetadata.objects.get_or_create(external_url=external_url)
             if external_identifier:
                 additional_metadata.external_identifier = external_identifier
+            additional_metadata.save()
```
azanbinzahid: shouldn't this be inside the `if` just above?

azanbinzahid (changes requested):

course_discovery/apps/api/serializers.py
DawoudSheraz: Yes, please suggest this change so I can commit it directly.

azanbinzahid (approved):

uzairr (approved):

PR #3290, opened 2022-02-15
fix: don't allow same URL slug with different case

waheedahmed:
Recently, changed the collation to utf8_bin which makes the field case sensitive and allows different entries of the same slug but in a different case. Adding a check to don't allow them for different courses.

DISCO-1790

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -12,7 +12,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import FileExtensionValidator
-from django.db import models, transaction
+from django.db import IntegrityError, models, transaction
 from django.db.models import F, Q
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1075,6 +1075,14 @@ def set_active_url_slug(self, slug):
         logger.info('The current slug is {}; The slug to be set is {}; Current course is a draft: {}'
                     .format(self.url_slug, slug, self.draft))
 
+        if slug:
+            # case 0: if slug is already in use with another, rasie an IntegrityError
+            excluded_course = self.official_version if self.draft else self.draft_version
+            slug_already_in_use = CourseUrlSlug.objects.filter(url_slug__iexact=slug.lower()).exclude(
+                course__in=[self, excluded_course]).exists()
```
zainab-amir, line 1082: just a minor confirmation: The check for course runs of the same course is already in place right? So it will make sure that a previous course run of the same course doesn't have the same slug?

zainab-amir (commented):
Should we add a test for this change?

course_discovery/apps/course_metadata/models.py
waheedahmed: @zainab-amir Course run slugs are separate from these and are generated from the course title and key. They cannot be edited manually, so we don't have to worry about them here.

waheedahmed: And the tests are already in place for this scenario with the default collation type.
https://github.com/openedx/course-discovery/blob/renovate/npm-ansi-regex-vulnerability/course_discovery/apps/course_metadata/tests/test_models.py#L807-L821

This check is specific to our case where we changed the collation for our DB instances only so not sure how to test that separately.

zainab-amir (approved):

PR #3291, opened 2022-02-15
feat: add utils to transform 2u data

uzairr:
## Description:
This PR will add utils method to transform 2u course fields.Following are the details:
Util methods to clean/update the following fields from WIP data CSV:

**Short** **Description**

> Remove Duration information from the first line, if present

**Minimum-maximum effort**

> Split Minimum-effort string from CSV based on the pattern

> If only a single number is present(minimum), the maximum will be calculated by +1 in min.

> Utility method to decode base64 urls

## Linked Ticket:
for more details please visit: [PROD-2675](https://openedx.atlassian.net/browse/PROD-2675)

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
@@ -3135,7 +3135,6 @@ INVALID_LANGUAGE = {
     'content_language': 'gibberish-language',
 }
 
-
 COURSE_TESTIMONIALS = [
     {
         "name": "Test User 1",
@@ -3185,3 +3184,9 @@ COURSE_CURRICULAM = {
         },
     ]
 }
+
+EFFORT_FORMATS = [
+    '7–10 hours per week', '7 hours per week'
+]
+
+BASE64_STRING = 'aHR0cHM6Ly93d3cuZ29vZ2xlLmNvbS8='
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_utils.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_utils.py
@@ -5,7 +5,9 @@ Unit tests for utils.
 from django.test import TestCase
 
 from course_discovery.apps.course_metadata.data_loaders.tests import mock_data
-from course_discovery.apps.course_metadata.data_loaders.utils import format_curriculum, format_faqs, format_testimonials
+from course_discovery.apps.course_metadata.data_loaders.utils import (
+    format_base64_strings, format_curriculum, format_effort_info, format_faqs, format_testimonials
+)
 
 
 class FormattingTests(TestCase):
@@ -51,3 +53,25 @@ class FormattingTests(TestCase):
             "</p><p><b>Module 2: </b>Spelling, consistency and style</p></div>"
         )
         assert format_faqs({}) == ""
+
+    def test_format_effort_info(self):
+        """
+        Test the utility function format_effort_info
+        """
+
+        output = format_effort_info(mock_data.EFFORT_FORMATS[0])
+        assert output == (7, 10)
+
+        output = format_effort_info(mock_data.EFFORT_FORMATS[1])
+        assert output == (7, 8)
+
+        # returns None if empty string is passed
+        self.assertIsNone(format_effort_info(""))
+
+    def test_format_base64_strings(self):
+        """
+        Test the utility function format_base64_strings
+        """
+
+        output = format_base64_strings(mock_data.BASE64_STRING)
+        assert output == "https://www.google.com/"
--- a/course_discovery/apps/course_metadata/data_loaders/utils.py
+++ b/course_discovery/apps/course_metadata/data_loaders/utils.py
@@ -1,6 +1,9 @@
 """
 Utils for loaders to format or transform field values.
 """
+import base64
+import re
+
 from lxml.html import clean
 
 from course_discovery.apps.course_metadata.validators import HtmlValidator
@@ -79,3 +82,26 @@ def format_testimonials(data):
         formatted_html += p_tag(f"<i>\"{text}\"</i>") + p_tag(f"-{name} ({title})")
 
     return cleaner.clean_html(formatted_html) if formatted_html else formatted_html
+
+
+def format_effort_info(data):
+    """
+    Format the effort info in the form of maximum and minimum effort
+    """
+    if len(data) > 0:
+        temp = re.findall(r'\d+', data)
+        effort_vals = tuple(map(int, temp))
+        if len(effort_vals) == 1:
+            effort_vals = (effort_vals[0], effort_vals[0] + 1)
+        return effort_vals
+    return None
+
+
+def format_base64_strings(data):
+    """
+    Decode and returns the encoded base64 strings if encoded
+    """
+    try:
+        return base64.b64decode(data).decode('utf-8')
+    except Exception:  # pylint: disable=broad-except
+        return data


course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -0,0 +1,38 @@
...
+import re
+import base64
+
+
+def format_short_description(data):
+    """
+    Format the short description field if it has duration info at its beginning
+    """
+    lowercase = data.lower()
+    index_if_exist = lowercase.find('duration')
```
DawoudSheraz: nit: we can do `>duration:` to be more inclusive there, we don't want to remove the actual word duration in the content if Duration info is not present on the first line. This might remove the need for `lowercase.find('>', 0, index_if_exist)`

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    if len(effort_vals) == 1:
+        effort_vals = (effort_vals[0], effort_vals[0] + 1)
+    return effort_vals
+
+
+def format_base64_strings(data):
+    """
+    Decode and returns the encoded base64 strings
+    """
+    return base64.b64decode(data).decode('utf-8')
```
DawoudSheraz: what happens if the content is not base64? To be on the safe side, we can add a check to decode only if the content is encoded in base64.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3134,3 +3134,11 @@
...
     'content_language': 'gibberish-language',
 }
+
+SHORT_DESCRIPTION = '<p><strong>Duration:</strong> 10 weeks (excluding orientation)<br>\r\n</p><p>An effective digital marketing strategy\r\n</p>'
+
+EFFORT_FORMATS = [
+    '7–10 hours per week', '7 hours per week'
+]
+
+BASE64STRING = 'aHR0cHM6Ly93d3cuZ29vZ2xlLmNvbS8='
```
DawoudSheraz: nit: BASE64_STRING

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3134,3 +3134,11 @@
     **VALID_COURSE_AND_COURSE_RUN_CSV_DICT,
     'content_language': 'gibberish-language',
 }
+
+SHORT_DESCRIPTION = '<p><strong>Duration:</strong> 10 weeks (excluding orientation)<br>\r\n</p><p>An effective digital marketing strategy\r\n</p>'
```
DawoudSheraz: We should add a 2nd variant where no removal happens if duration information is not present.

course_discovery/apps/course_metadata/data_loaders/tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class FormattingTests(TestCase):
+    """
+    Test Suite for formatting utils
+    """
+
+    def test_format_short_description(self):
+        """
+        Test the utility function format_testimonials
```
DawoudSheraz: docstring updates

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+def format_effort_info(data):
+    """
+    Format the effort info in the form of maximum and minimum effort
+    """
+    temp = re.findall(r'\d+', data)
+    effort_vals = tuple(map(int, temp))
+    if len(effort_vals) == 1:
+        effort_vals = (effort_vals[0], effort_vals[0] + 1)
+    return effort_vals
```
azanbinzahid: What will it return if an empty string is passed?

azanbinzahid (commented):
Few more test cases would be needed to verify its functionality.

course_discovery/apps/course_metadata/data_loaders/utils.py
uzairr: good catch 👍 
partially it is correct but in the later one there is a possibility of a nested tag so the mentioned codeline would catch that use case i.e.
For example: `<p><strong>Duration:</strong> 10 weeks (excluding orientation)<br>\r\n</p><p>An effective digital marketing strategy\r\n</p>`

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        if len(effort_vals) == 1:
+            effort_vals = (effort_vals[0], effort_vals[0] + 1)
+        return effort_vals
+
+
+def format_base64_strings(data):
+    """
+    Decode and returns the encoded base64 strings if encoded
+    """
+    if not is_base64(data):
```
DawoudSheraz: this code pattern here seems confusing, if data is not base64, why would the base64 decode be called against it. Renaming would do the job here.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3134,3 +3134,14 @@
     **VALID_COURSE_AND_COURSE_RUN_CSV_DICT,
     'content_language': 'gibberish-language',
 }
+
+SHORT_DESCRIPTION = [
+    '<p><strong>Duration:</strong> 10 weeks (excluding orientation)<br>\r\n</p><p>An effective digital marketing strategy\r\n</p>',
+    '<p>An effective digital marketing strategy\r\n</p>'
```
DawoudSheraz: Like Azan said, we can have more scenarios here for test, just to be safe.

DawoudSheraz (approved):
follow-up nits, otherwise 👍🏽

azanbinzahid:
Please make sure to test on local publisher on various inputs.

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    lowercase = data.lower()
+    index_if_exist = lowercase.find('>duration')
+    if index_if_exist > 0:
+        upper_index = lowercase.find('>', 0, index_if_exist)
+        start_tag = lowercase[0: upper_index + 1]
+        end_tag = '</{}'.format(start_tag[1:])
+        start_index = lowercase.find(end_tag, index_if_exist) + len(end_tag)
+        data = data[start_index:]
+    return data
```
DawoudSheraz: As discussed internally, the duration information will remain within the HTML. We can remove this method.

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -0,0 +1,52 @@
...
+            effort_vals = (effort_vals[0], effort_vals[0] + 1)
+        return effort_vals
+    return None
+
+
+def format_base64_strings(data):
+    """
+    Decode and returns the encoded base64 strings if encoded
+    """
+    if not is_base64(data):
```
DawoudSheraz: When testing on the local, this check caused issues when data was equal to empty string. Either we can hide the function behind if check or update this method.

DawoudSheraz (changes requested):
Added comments/small changes to be done based on discussions.

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -79,3 +82,34 @@ def format_testimonials(data):
...
     return cleaner.clean_html(formatted_html) if formatted_html else formatted_html
+
+
+def is_base64(data):
+    """
+    Verifies whether the given string is base64 or not
+    """
+    if len(data) > 0:
+        return base64.b64encode(base64.b64decode(data).decode('utf-8')) == data
+    return False
```
DawoudSheraz: Seems strange that we are decoding and encoding the value to check if the data is base64. If the data is not base64, the exception will be raised and that has not been handled in the caller function. This util method might not be needed if the decode method is wrapped in try/catch in the caller function and return original value if the decode fails.

azanbinzahid (approved):

DawoudSheraz (approved):

PR #3295, opened 2022-02-16
feat: offset based reading

ansabgillani:
This is a feature that allows offset-based and batch-size reading for csv.

Ticket LInk: [PROG-2636](https://openedx.atlassian.net/browse/PROD-2636?atlOrigin=eyJpIjoiYjQxYmZmMDBiM2QxNDdkNGI3MjZkMzEwYmVkNmQxZWIiLCJwIjoiaiJ9)

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -49,7 +49,7 @@ class CSVDataLoader(AbstractDataLoader):
             if index < self.start_index:
                 # Skip n number of rows while traversal
                 continue
-            if index >= self.batch_size + self.start_index or index == len(self.reader_len):
+            if index >= self.batch_size + self.start_index or index == self.reader_len:
                 # Break when traversal exceeds batch size
                 # There is a possibility that the index is starting from ith location.
                 # So the batch size + start index will be cut 
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -30,11 +30,13 @@ class CSVDataLoader(AbstractDataLoader):
         ProgramType.PROFESSIONAL_CERTIFICATE
     ]
 
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None, is_draft=False):
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None, is_draft=False, start_index=0, batch_size=-1):
         super().__init__(partner, api_url, max_workers, is_threadsafe)
 
         try:
             self.reader = csv.DictReader(open(csv_path, 'r'))
+            self.start_index = start_index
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -30,20 +30,31 @@ class CSVDataLoader(AbstractDataLoader):
...
             self.is_draft = is_draft
         except FileNotFoundError:
             logger.exception("Error opening csv file at path {}".format(csv_path))
             raise  # re-raising exception to avoid moving the code flow
 
     def ingest(self):  # pylint: disable=too-many-statements
         logger.info("Initiating CSV data loader flow.")
-        for row in self.reader:
-
+        for index, row in enumerate(self.reader):
```
DawoudSheraz, line 48: When `len(list(self.reader))` is called in __init__, iterating on the reader again might not be possible. The reader is iterator(https://docs.python.org/3/library/csv.html#csv.reader). You would need to store data in list first and use that list here.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -30,20 +30,31 @@ class CSVDataLoader(AbstractDataLoader):
...
+            if index < self.start_index:
+                # Skip n number of rows while traversal
+                continue
+            if index >= self.batch_size + self.start_index or index == self.reader_len:
```
DawoudSheraz, line 52: index == reader_len - 1. index is starting at 0

course_discovery/apps/course_metadata/management/commands/import_course_metadata.py
```diff
@@ -37,10 +37,22 @@ def add_arguments(self, parser):
...
+            help='The integer value, startiing from 0, which is used to give initial index of the batch. Default is 0.',
+            type=int,
+            required=False
+        )
+        parser.add_argument(
+            '--batch_size',
+            help='The integer value, startiing from 1, which is used to give size of the batch. Default is length of csv.',
+            type=int,
+            required=False
+        )
```
DawoudSheraz, line 51: give default values if they are not required.

course_discovery/apps/course_metadata/management/commands/tests/test_import_course_metadata.py
```diff
@@ -113,3 +113,42 @@ def test_success_flow(self, jwt_decode_patch):  # pylint: disable=unused-argumen
...
+        """
+        self._setup_prerequisites(self.partner)
+        self.mock_studio_calls(self.partner)
+        self.mock_ecommerce_publication(self.partner)
+        _, image_content = self.mock_image_response()
+        batch_size = 5  
+
+        with NamedTemporaryFile() as csv:
+            csv = self._write_csv(
+                csv, [mock_data.VALID_COURSE_AND_COURSE_RUN_CSV_DICT for i in range(10)])
```
DawoudSheraz, line 130: We are essentially writing the same course/row 10 times. You would need to change id/number to have unique entries. Otherwise the test will fail.

PR #3309, opened 2022-02-28
feat: stop marketing Old Mongo courses

mikix:
They are being sunset and we no longer want to link to them.

Because I started paying attention to key structure, I also updated the default course factories to use new-style keys and updated some tests to match. 

(My first swing at this was on the LMS side to mark them `hidden`, but prospectus still generates pages for those courses, just doesn't include them in search/browse results. So this is a step further.)

[DEPR-58](https://openedx.atlassian.net/browse/DEPR-58)

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
@@ -13,7 +13,7 @@ class CatalogQueryViewSetTests(APITestCase):
         self.user = UserFactory(is_staff=True, is_superuser=True)
         self.client.force_authenticate(self.user)
         self.course = CourseFactory(partner=self.partner, key='simple_key')
-        self.course_run = CourseRunFactory(course=self.course)
+        self.course_run = CourseRunFactory(course=self.course, key='simple/key/run')
         self.url_base = reverse('api:v1:catalog-query_contains')
         self.error_message = 'CatalogQueryContains endpoint requires query and identifiers list(s)'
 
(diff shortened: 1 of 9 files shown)


course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
```diff
@@ -13,7 +13,7 @@ def setUp(self):
         self.user = UserFactory(is_staff=True, is_superuser=True)
         self.client.force_authenticate(self.user)
         self.course = CourseFactory(partner=self.partner, key='simple_key')
-        self.course_run = CourseRunFactory(course=self.course)
+        self.course_run = CourseRunFactory(course=self.course, key='simple/key/run')
```
mikix, line 16: I tried but failed to figure out how to properly escape a modern-style key for elasticsearch (presumably have to encode the colon and plus somehow), so I just used an old-style key instead. The point of the test is whether we can pass the key around, not the encoding specifically, so I just called it a day. No worse than before anyway.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1982,6 +1983,8 @@ def could_be_marketable(self):
         """
         if not self.type.is_marketable:
             return False
+        if CourseKey.from_string(self.key).deprecated:  # Old Mongo courses are not marketed
+            return False
```
mikix, line 1988: This is the actual point of this PR - rest is test fixes to accomodate this.

waheedahmed (approved):

PR #3317, opened 2022-03-07
feat: ENT-5469 Added nested inlines for learner pathway admin

irfanuddinahmad:
***Description***: This PR updates the django model admin pages of learnerpathway to incorporate multi level inline steps, courses and programs

***JIRA***: [ENT-5469](https://openedx.atlassian.net/browse/ENT-5469)

### DIFF
--- a/course_discovery/apps/learner_pathway/admin.py
+++ b/course_discovery/apps/learner_pathway/admin.py
@@ -1,7 +1,8 @@
 """
 Admin definitions for learner_pathway app.
 """
-
+# pylint: disable=no-member
+import nested_admin
 from django.contrib import admin
 from django.urls import reverse
 from django.utils.html import format_html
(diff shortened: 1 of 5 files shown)


saleem-latif (approved):

HammadAhmadWaqas (changes requested):
I think we don't need to use a third-party package here. We can use Django's built-in class.

Please read Django docs here: https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#inlinemodeladmin-objects

we are already doing this here: https://github.com/openedx/edx-enterprise/blob/master/enterprise/admin/__init__.py#L170

HammadAhmadWaqas (approved):
As nested Inline admins are not fully supported, Using 3rd party library make sense.

PR #3320, opened 2022-03-09
feat: add certificate_info and facts fields to additional_metdata

azanbinzahid:
## Description:
This PR adds `certificate_info` and `facts` model/fields to `AdditionalMetadata` model in `course_metadata` app.

## Linked Ticket:
[PROD-2713](https://openedx.atlassian.net/browse/PROD-2713)

## Screenshots
| Empty | Not Empty |
| :---: | :---: |
| image | image |

## TODO:

- [x] Write facts serializer
- [x] Internal squad review
- [ ] Owner squad review
- [x] Cater for orphan objects
- [x] decide between text vs html field for blurb and add heading/blurb model if req

## Testing:
I am finding some inconsistencies while making patch requests from publisher browser console. It can be related to existing courses or draft vs non-draft version of the courses. This needs to be tested in detail before merging any change. 
- [x] Fixed


course_discovery/apps/course_metadata/models.py
```diff
@@ -598,6 +606,13 @@ class AdditionalMetadata(TimeStampedModel):
     external_url = models.URLField(blank=False, null=False)
     external_identifier = models.CharField(max_length=255, blank=True, null=False)
     lead_capture_form_url = models.URLField(blank=True, null=False)
+    facts = models.ManyToManyField(
+        Fact, default=None, blank=True, related_name='realted_course_additional_metadata',
```
DawoudSheraz: related_course_additional_metadata

course_discovery/apps/course_metadata/models.py
```diff
@@ -598,6 +606,13 @@ class AdditionalMetadata(TimeStampedModel):
     external_url = models.URLField(blank=False, null=False)
     external_identifier = models.CharField(max_length=255, blank=True, null=False)
     lead_capture_form_url = models.URLField(blank=True, null=False)
+    facts = models.ManyToManyField(
```
DawoudSheraz, line 623: I am still juggling b/w ForeignKey and M2M relation type here.

course_discovery/apps/api/serializers.py
```diff
@@ -1167,13 +1187,33 @@ def get_skills(self, obj):
     def create(self, validated_data):
         return Course.objects.create(**validated_data)
 
+    def update_facts(self, instance, facts):
+        # TODO: Need suggestions on how to update this ManyToManyField
+        pass
```
DawoudSheraz: what issue is being faced here? We can get or create facts objects and add the objects to instance facts field.

azanbinzahid: Like what logic we should use while updating or creating these? how to edit (existing) list of fact and/or associated course or instance objects.

azanbinzahid:
Looping in @ziafazal for final review and testing

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1345,6 +1349,23 @@ class TestAbstractTitleDescriptionModel(AbstractTitleDescriptionModel):
         assert str(instance) == title
 
 
+class AbstractHeadingBlurbModelTests(TestCase):
+    """ Tests for AbstractTitleDescriptionModel. """
+
+    def test_str(self):
+        class TestAbstractHeadingBlurbModel(AbstractHeadingBlurbModel):
+            pass
```
DawoudSheraz: nit: a docstring of why this model is required in the tests could be good.

DawoudSheraz (commented):
The changes look good, going to test on the local to identify any issues if present.

course_discovery/apps/api/serializers.py
```diff
@@ -1167,13 +1193,36 @@ def get_skills(self, obj):
...
+        for fact in facts_data:
+            fact_obj = Fact.objects.create(**fact)
+            instance.facts.add(fact_obj)
+
+    def update_certificate_info(self, instance, certificate_info_data):
+
+        if instance.certificate_info:
+            CertificateInfo.objects.filter(id=instance.certificate_info.id).update(**certificate_info_data)
+        else:
+            instance.certificate_info = CertificateInfo.objects.create(**certificate_info_data)
```
DawoudSheraz, line 1208: Since the instance here is additional_metadata and not a course object, a save call is needed here. On local testing, the cert object was created but not associated with additional_metadata object. Adding the save call resolved the issue(though not sure if the tests would be impacted by it).

DawoudSheraz (approved):
one comment to address before the merge

course_discovery/apps/api/serializers.py
```diff
@@ -1167,13 +1193,36 @@ def get_skills(self, obj):
     def create(self, validated_data):
         return Course.objects.create(**validated_data)
 
+    def update_facts(self, instance, facts_data):
+
+        instance.facts.all().delete()
+        for fact in facts_data:
+            fact_obj = Fact.objects.create(**fact)
+            instance.facts.add(fact_obj)
```
DawoudSheraz, line 1201: even if the facts_data is the same, the prev objects will be deleted and new ones with the same data created. Maybe not in this PR but if we can add a mechanism to only delete or update those that have been modified in a follow-up PR, it can save DB operations in an already expensive API call.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1021,31 +1021,62 @@ def test_update_with_level_type(self):
 
     @responses.activate
     def test_update_with_additional_metadata(self):
+        course = CourseFactory(additional_metadata=None)
```
DawoudSheraz, line 1024: just out of curiosity: why is a separate course needed here? Why didn't self.course work here?

DawoudSheraz (approved):

course_discovery/apps/api/v1/tests/test_views/test_courses.py
azanbinzahid: The self.course that is being setup in setUp() of test suite, already has an additional metadata attached to it via factories. It makes unable to test adding correct additional_metadata instance via API.

DawoudSheraz: hmmm, shouldn't patch requests update the already attached/factory created additional metadata object?

azanbinzahid: it does, but it doesn't test the case when object is created. Before, it can only test updating objects already created. That is why we missed `instance.save()` for cert info.

uzairr (approved):

ziafazal (approved):
👍

PR #3322, opened 2022-03-11
ENT-5431: Fixed a few bugs related to learner pathway found in testing. Also renamed name field to title in learner pathway model.

saleem-latif:
__Jira Ticket:__ [ENT-5431](https://openedx.atlassian.net/browse/ENT-5431)

__Description:__
This PR fixes the following 2 issues found during testing

1. Enterprise catalog expected the product (course, program, or learner pathway) to have a title attribute but in pathways, the attribute key is name instead of title. We should rename this field to title to conform to the other 2 product types (i.e. course and program). I think we can achieve that by simply renaming the column name to title.
2. Partner should default to edX, also update the [LearnerPathwayDocument](https://github.com/openedx/course-discovery/blob/master/course_discovery%2Fapps%2Fcourse_metadata%2Fsearch_indexes%2Fdocuments%2Flearner_pathway.py#L48) to make sure that objects having None as the partner do not cause an exception.


sameenfatima78 (approved):

course_discovery/apps/course_metadata/search_indexes/documents/learner_pathway.py
```diff
@@ -45,7 +45,7 @@ def prepare_aggregation_key(self, obj):
         return 'learnerpathway:{}'.format(obj.uuid)
 
     def prepare_partner(self, obj):
-        return obj.partner.short_code
+        return obj.partner.short_code if obj.partner is not None else ''
```
HammadAhmadWaqas: Can we just add `if obj.partner else ''`

HammadAhmadWaqas (approved):
1 nit otherwise looks good once tests pass.

waheedahmed (approved):

PR #3342, opened 2022-03-21
refactor: make handling of optional fields flexible in CSV loader

DawoudSheraz:
### Description

- Handle optional fields for CSV loader in a flexible way as to not break the flow if the data dict or column is not present in CSV
- If the data is present for an optional field, it would work as before. 
- Allow the override code field to be blank in the serializer.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -180,7 +180,7 @@ class CSVDataLoader(AbstractDataLoader):
         """
         Create and return the request data for making a patch call to update the course.
         """
-        collaborator_uuids = self.process_collaborators(data['collaborators'], course.key)
+        collaborator_uuids = self.process_collaborators(data.get('collaborators', ''), course.key)
         subjects = self.get_subject_slugs(
             data.get('primary_subject'),
             data.get('secondary_subject'),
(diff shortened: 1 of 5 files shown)


uzairr (approved):
ensure the green build please

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -186,37 +185,26 @@ def _update_course_api_request_data(self, data, course):
...
-                'certificate_info': self.process_heading_blurb(
-                    data['certificate_header'], data['certificate_text']
-                ),
-                'facts': self.process_stats(
-                    data['stat1'], data['stat1_text'],
-                    data['stat2'], data['stat2_text']
-                )
-            },
+            'learner_testimonials': self.get_dict_key_value(data, 'learner_testimonials'),
+            'additional_information': self.get_dict_key_value(data, 'additional_information'),
```
azanbinzahid: `data.get(key, '')` was also fine.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -468,3 +457,38 @@ def process_stats(self, stat1, stat1_text, stat2, stat2_text):
...
+        return data.get(key, value)
+
+    def get_additional_metadata_dict(self, data):
+        """
+        Return the appropriate additional metadata dict representation, skipping the keys that are not
+        present in the input data dict.
+        """
+        additional_metadata = {
+            'external_url': data['redirect_url'],
+            'external_identifier': data['external_identifier'],
```
azanbinzahid, line 482: `dict/key` function here?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -468,3 +457,38 @@ def process_stats(self, stat1, stat1_text, stat2, stat2_text):
         if stat2_dict:
             stats.append(stat2_dict)
         return stats
+
+    def get_dict_key_value(self, data, key, value=''):
```
azanbinzahid: change value to `default_value`

azanbinzahid: What if some fields accept `None` instead of `''`?

azanbinzahid (changes requested):
I don't like the overall idea of `get_dict_key_value` utility function given a builtin method already exist. Please comment if it has additional benefits.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: These are required fields for 2U Product lines.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: This is the default value. The caller can always provide a different value if '' is not the default value for a field.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: Well, the default value is set to '' for many string/html fields. Dictionary's get(key) returns None if the value is not found. That's why the util function was added to avoid repeating dict.get(key, '') pattern.

azanbinzahid (approved):

PR #3350, opened 2022-03-30
feat: add organic_url to additional metadata

azanbinzahid:
## Description:
This PR adds `organic_url` to `AdditionalMetadata` model in `course_metadata` app. It also increases the `max_len` of url fields to `511` and adds `help_text`. 

## Linked Ticket:
[PROD-2758](https://openedx.atlassian.net/browse/PROD-2758)


azanbinzahid:
@DawoudSheraz should we add loader changes along with it?

uzairr (approved):
ensure the green build otherwise LGTM 👍

course_discovery/apps/course_metadata/models.py
```diff
@@ -620,6 +620,7 @@ class AdditionalMetadata(TimeStampedModel):
     external_url = models.URLField(blank=False, null=False)
     external_identifier = models.CharField(max_length=255, blank=True, null=False)
     lead_capture_form_url = models.URLField(blank=True, null=False)
+    organic_url = models.URLField(blank=True, null=False)
```
DawoudSheraz: lets add max_length=511 here. The length issue faced in external_url(yet to be fixed) might pop-up here too. Also, adding a help_text for both organic and external URLs will help in creating a distinction among these URLs.

azanbinzahid: can we add `max_len` to `external_url` as well?

DawoudSheraz: Yes, that would be +1. Just make sure to test it out on local so that changing column length does not affect the existing data.

azanbinzahid: ..updated and tested locally

DawoudSheraz:
> @DawoudSheraz should we add loader changes along with it?

Sorry, I missed this question earlier. Let's target that in a follow-up PR.

DawoudSheraz (approved):
Please update the PR description for all the changes done. Otherwise, 👍🏽

PR #3351, opened 2022-03-31
feat: add product type field

uzairr:
Add course type field in serializer which represents type slug to distinguish among courses.

PROD-2749

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1032,6 +1032,7 @@ class MinimalCourseSerializer(FlexFieldsSerializerMixin, TimestampModelSerialize
     type = serializers.SlugRelatedField(required=True, slug_field='uuid', queryset=CourseType.objects.all())
     uuid = UUIDField(read_only=True, default=CreateOnlyDefault(uuid4))
     url_slug = serializers.SerializerMethodField()
+    course_type = serializers.SerializerMethodField()
 
     @classmethod
     def prefetch_queryset(cls, queryset=None, course_runs=None):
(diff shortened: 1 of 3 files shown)


DawoudSheraz (commented):
Looks good from my end, although a ddt test to verify the product types for different courses is missing.

course_discovery/apps/api/serializers.py
```diff
@@ -1045,6 +1046,12 @@ def prefetch_queryset(cls, queryset=None, course_runs=None):
             cls.prefetch_course_runs(MinimalCourseRunSerializer, course_runs),
         )
 
+    def get_course_type(self, obj):
+        course_type_obj = getattr(obj, 'type', None)
+        if course_type_obj:
+            return course_type_obj.slug
+        return None
```
DawoudSheraz: type is always present on the course. Did you run into a scenario where it was not the case?

uzairr: No i didn't, it is just a safer way to access a custom written attribute!

DawoudSheraz: Well, I would say there is no need for that, since the course will always have a course type.

course_discovery/apps/api/serializers.py
```diff
@@ -1032,6 +1032,7 @@ class MinimalCourseSerializer(FlexFieldsSerializerMixin, TimestampModelSerialize
     type = serializers.SlugRelatedField(required=True, slug_field='uuid', queryset=CourseType.objects.all())
     uuid = UUIDField(read_only=True, default=CreateOnlyDefault(uuid4))
     url_slug = serializers.SerializerMethodField()
+    course_type = serializers.SerializerMethodField()
```
DawoudSheraz, line 1035: Make this a read-only field.

azanbinzahid: How this addition will return type slug? What if we use `SlugRelated` field?

azanbinzahid (changes requested):
Please add field presence test in api test course views.

course_discovery/apps/api/serializers.py
uzairr: by default it is read-only: https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L1874

azanbinzahid (approved):

DawoudSheraz (approved):
- Once the build is green
- Update PR description to include the updated changes

PR #3352, opened 2022-03-31
feat: Added course recommendation model.

saleem-latif:
__JIRA Ticket:__ [ENT-5553](https://openedx.atlassian.net/browse/ENT-5553)

__Description:__
This PR adds a new model `CourseRecommendation` to store course recommendations. This is implementation is part of the [ENT-5552](https://openedx.atlassian.net/browse/ENT-5552)


course_discovery/apps/taxonomy_support/models.py
```diff
@@ -0,0 +1,46 @@
...
+        on_delete=models.CASCADE,
+        help_text=_('This is the course recommended after learner completes the original course.'),
+    )
+    skills_intersection_ratio = models.FloatField(
+        help_text=_('Ratio of number of common skills and total skill count.'),
+    )
+    skills_intersection_length = models.IntegerField(
+        help_text=_('Count of common skills between the original course and recommended course.'),
+    )
+    subjects_intersection_ratio = models.FloatField(
```
sameenfatima78, line 36: The total subject count here is the # of subjects that the original course has, right? We should still check with Nathan if we want to save subject ratio like this... now that it's been identified that a course can have multiple subjects.

course_discovery/apps/taxonomy_support/models.py
```diff
@@ -0,0 +1,46 @@
...
+because they have direct dependency with models from course discovery.
+"""
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from model_utils.models import TimeStampedModel
+from course_discovery.apps.course_metadata.models import Course
+
+
+class CourseRecommendation(TimeStampedModel):
```
sameenfatima78, line 12: Shouldn't this table reside in course metadata app? Since the models belonging to Taxonomy app are already in the taxonomy-connector repository, not sure if we should create a new models file inside this app here.

sameenfatima78 (approved):
Some comments but other than that looks good! Also, test cases are missing for new model.

course_discovery/apps/taxonomy_support/models.py
saleem-latif: I placed the model here as these are more related to the `taxonomy-connector` then course metadata app. Course recommendations are solely based on the skills data coming from taxonomy and will not work without it. That is why it seemed better to keep it here.

course_discovery/apps/taxonomy_support/models.py
saleem-latif: total subject count is (count of subject in original course + count if subjects in recommended course), i.e. the same logic as is in skills.

Regarding the other question, I have posted a comment in the ticket for Nathan.

waheedahmed (approved):

PR #3369, opened 2022-04-13
feat: add course type filtering in courses api

DawoudSheraz:
### Description

- Add course_type filtering in course API to quickly filter the courses based on the course type
- The exception in course type slug is `open-courses`, where all the course type courses except Executive Education and Bootcamp are returned. [Figma Design Reference](https://www.figma.com/file/M4jkaKcn1kwbN1OcEVMiPr/Project-Board%3A-Publisher?node-id=0%3A1)
- If the slug is incorrect/missing, no validation is done. By default, the filter will be attempted and empty results will be returned if the slug is incorrect.

### Testing

- Use different course types on `http://localhost:18381/api/v1/courses/?course_type=executive-education-2u&editable=1` and verify the results

### Reference

- https://django-filter.readthedocs.io/en/main/ref/filters.html#keyword-only-arguments

### DIFF
--- a/course_discovery/apps/api/filters.py
+++ b/course_discovery/apps/api/filters.py
@@ -13,7 +13,7 @@ from rest_framework.exceptions import NotFound, PermissionDenied
 from course_discovery.apps.api.utils import cast2int
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
-    Course, CourseEditor, CourseRun, LevelType, Organization, Person, Program, ProgramType, Subject, Topic
+    Course, CourseEditor, CourseRun, CourseType, LevelType, Organization, Person, Program, ProgramType, Subject, Topic
 )
 
 logger = logging.getLogger(__name__)
@@ -88,6 +88,7 @@ class CourseFilter(filters.FilterSet):
     uuids = UUIDListFilter()
     course_run_statuses = CharListFilter(method='filter_by_course_run_statuses')
     editors = CharListFilter(field_name='editors__user__pk', lookup_expr='in', distinct=True)
+    course_type = filters.CharFilter(method='filter_by_course_type')
 
     class Meta:
         model = Course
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/filters.py
```diff
@@ -113,6 +114,20 @@ def filter_by_course_run_statuses(self, queryset, _, value):
 
         return queryset.filter(status_check, course_runs__hidden=False).distinct()
 
+    def filter_by_course_type(self, queryset, _, value):
+        """
+        Filter the courses by the course type by validating if the provided type slug is a valid slug.
+
+        If the filter value is `open-courses`, all the courses except Bootcamp and Executive Education
```
azanbinzahid, line 121: Why it is named `open-courses`?

azanbinzahid (approved):
Looks good.

course_discovery/apps/api/filters.py
DawoudSheraz: To correspond with Figma navigation design https://www.figma.com/file/M4jkaKcn1kwbN1OcEVMiPr/Project-Board%3A-Publisher?node-id=0%3A1

course_discovery/apps/api/filters.py
```diff
@@ -113,6 +114,20 @@ def filter_by_course_run_statuses(self, queryset, _, value):
...
+        courses would be returned.
+        """
+        filter_value = None
+        if value == 'open-courses':
+            filter_value = ~Q(type__slug__in=[CourseType.EXECUTIVE_EDUCATION_2U, CourseType.BOOTCAMP_2U])
+        elif CourseType.objects.filter(slug=value).exists():
+            filter_value = Q(type__slug=value)
+        return queryset.filter(filter_value) if filter_value else queryset
```
uzairr: It might return `None` instead returning the queryset, semantically nothing would be returned in case of a mismatched filter value ... 💭 💭

uzairr (approved):
Just a little semantic thought, otherwise LGTM 👍

course_discovery/apps/api/filters.py
DawoudSheraz: Sure, I will update it.

course_discovery/apps/api/filters.py
```diff
@@ -113,6 +114,19 @@ def filter_by_course_run_statuses(self, queryset, _, value):
...
+        """
+        Filter the courses by the course type by validating if the provided type slug is a valid slug.
+
+        If the filter value is `open-courses`, all the courses except Bootcamp and Executive Education
+        courses would be returned.
+        """
+        if value == 'open-courses':
+            filter_value = ~Q(type__slug__in=[CourseType.EXECUTIVE_EDUCATION_2U, CourseType.BOOTCAMP_2U])
+        else:
+            filter_value = Q(type__slug=value)
```
zainab-amir, line 127: Do we need to show any error if the type of slug is not valid?

zainab-amir (approved):
One comment otherwise :shipit:

course_discovery/apps/api/filters.py
DawoudSheraz: I did a similar thing before. But like Uzair pointed out above, the missing/incorrect slug should return an empty result. Therefore, filtering on an incorrect slug automatically creates an empty queryset.

PR #3370, opened 2022-04-14
feat: add card_image in LearnerPathway model

muneebGH:
[ticket](https://openedx.atlassian.net/browse/ENT-5644)


sameenfatima78 (approved):

course_discovery/apps/learner_pathway/models.py
```diff
@@ -98,7 +98,20 @@ class LearnerPathway(models.Model):
...
+        upload_to=UploadToFieldNamePath(populate_from='uuid', path='media/learner_pathway/card_images'),
+        blank=True,
+        null=True,
+        variations={
+            'large': (1440, 480),
+            'medium': (726, 242),
+            'small': (435, 145),
+            'x-small': (348, 116),
+        },
+        help_text='image that will be displayed on learner pathway cards',
```
HammadAhmadWaqas, line 111: Should not card_image have different size than banner_image?
see this example: https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L2239

HammadAhmadWaqas (commented):
Do we need this data in algolia?

muneebGH:
@HammadAhmadWaqas yes we would need to change the algolia logic to take card_image instead of banner_image. banner_image would be fetched when modal is opened from discovery through api call.

course_discovery/apps/course_metadata/search_indexes/serializers/learner_pathway.py
```diff
@@ -18,6 +19,7 @@ class LearnerPathwaySearchDocumentSerializer(ModelObjectDocumentSerializerMixin,
 
     steps = LearnerPathwayStepSerializer(many=True, source='object.steps')
     banner_image = StdImageSerializerField(source='object.banner_image')
+    card_image_url = serializers.SerializerMethodField()
```
saleem-latif, line 22: Why do we need `SerializerMethodField` here? wouldn't it be better to use something like `banner_image`?

saleem-latif (approved):
One comment, other than that looks good 👍

course_discovery/apps/course_metadata/search_indexes/serializers/learner_pathway.py
muneebGH: in tried using that, but it didnt work somehow, so found this approach [here ](https://github.com/openedx/course-discovery/blob/6ef5139f2106617e388c92a839d47d337c4270e4/course_discovery/apps/course_metadata/search_indexes/serializers/person.py#L19)and implemented it, it worked.

PR #3384, opened 2022-04-18
fix: Upgrade Node version in Dockerfile

Jawayria:
As we are moving course-discovery to Node 16, we need to update Dockerfile accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,7 @@ WORKDIR /edx/app/discovery
 # Copy just JS requirements and install them.
 COPY package.json package.json
 COPY package-lock.json package-lock.json
-RUN nodeenv /edx/app/nodeenv --node=12.11.1 --prebuilt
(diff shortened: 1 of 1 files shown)


awais786 (approved):

Dockerfile
```diff
@@ -43,7 +43,7 @@ WORKDIR /edx/app/discovery
 # Copy just JS requirements and install them.
 COPY package.json package.json
 COPY package-lock.json package-lock.json
-RUN nodeenv /edx/app/nodeenv --node=12.11.1 --prebuilt
+RUN nodeenv /edx/app/nodeenv --node=16.14.2 --npm=8.5.x --prebuilt
```
iamsobanjaved, line 46: Have you tested creating `nodeenv` with `x` in the patch version locally? I have a feeling that this will fail, as it failed for me when I tried `16.x.x` in the node version.

iamsobanjaved: I have tested this locally and this command works fine.

iamsobanjaved (approved):

PR #3397, opened 2022-04-25
feat: adr for middleware api

uzairr:
Add ADR document for the middleware api approaches to sync
external data for course(s).

PROD-2754


davidjoy:
First, quick note: 

This ADR is at the path `docs/decisions/docs/decisions` - looks like it should be moved up two directories.

Diving into the content a bit:

I think we need to be more specific about what decision we're making in this ADR and why it needs to be made.  The title is "Sync external course(s) in discovery" and the PR is "adr for middleware api". That makes me think we're writing an ADR about our decision to create a "middleware API" so we can sync external courses in discovery.  

The decision we're document, then, is why this API is necessary.  The alternatives considered jump to implementation details about whether it's asynchronous or synchronous, but we haven't yet answered the question about whether we need a new API at all!  

Some questions we could ask ourselves and document in this ADR:

- What problem are we trying to solve?
  - _Potential_ problems, for instance, could be: volume of data is too high and might crash the service, we need better data validation, the existing APIs are too slow, the existing APIs do not give good feedback, resulting in a bad developer experience, etc.  
- Why does that problem mean we need a new API?
  - This might be answered by describing what's wrong with the existing APIs and how they don't address the problems above.  
- Why is this a "middleware"?  (I think that's a misleading term and would advocate for not using it...)
  - I feel like this is a term we said at some point and have started using, but I don't think what we're talking about is a middleware... 
- Why is it important for us to think about whether or not its asynchronous, assuming it needs to exist?
  - The answer to this is probably in response to one of the problems and the shortcomings in our existing APIs.

I don't know the answers to these questions, nor do I know whether we even need to do this project!  The idea of adding a new API to course-discovery was an idea by some folks who aren't in the code (me, for instance!) and aren't familiar with how it works.  It's the responsibility of the team to determine whether that idea is even worth doing, and if so, to document why here.  So I'd be very careful not to assume that we _need_ to do this.  If we've determined we _do need to do it_, that's how we made that decision is what should be in this ADR.

PR #3411, opened 2022-05-06
fix: Updated the logic to calculate estimated effort to complete a node in learner pathway.

saleem-latif:
__JIRA Ticket:__ [ENT-5646](https://openedx.atlassian.net/browse/ENT-5646)

__Description:__
Updated the model so that the step reflects the min/max range of values.

### DIFF
--- a/course_discovery/apps/learner_pathway/models.py
+++ b/course_discovery/apps/learner_pathway/models.py
@@ -17,7 +17,7 @@ from course_discovery.apps.course_metadata.models import Course, Program
 from course_discovery.apps.course_metadata.utils import UploadToFieldNamePath
 from course_discovery.apps.learner_pathway import constants
 from course_discovery.apps.learner_pathway.choices import PathwayStatus
-from course_discovery.apps.learner_pathway.utils import get_advertised_course_run_estimated_hours
+from course_discovery.apps.learner_pathway.utils import avg, get_advertised_course_run_estimated_hours
 
 
 class AbstractModelMeta(ABCMeta, type(models.Model)):
@@ -35,7 +35,7 @@ class LearnerPathwayNode(models.Model, metaclass=AbstractModelMeta):
         abstract = True
 
     @abstractmethod
-    def get_estimated_time_of_completion(self) -> str:
+    def get_estimated_time_of_completion(self) -> int:
         """
         Subclasses must implement this method to calculate and return the estimated time of completion of the node.
         """
@@ -123,7 +123,7 @@ class LearnerPathway(models.Model):
         """
         completion_time = 0
         for step in self.steps.all():
-            completion_time += step.get_estimated_time_of_completion()
+            completion_time += avg(step.get_estimated_time_of_completion())
         return completion_time
 
     @property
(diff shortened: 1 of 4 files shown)


sameenfatima78 (approved):

course_discovery/apps/learner_pathway/models.py
```diff
@@ -181,8 +181,21 @@ def remove_node(self, uuid):
...
+            (tuple<int, int>): A tuple containing the min and max of estimated time of completion.
+        """
+        estimated_completion_times_of_nodes = sorted(
+            [node.get_estimated_time_of_completion() for node in self.get_nodes()]
+        )
+        estimated_completion_times_of_nodes = sorted(estimated_completion_times_of_nodes)
+        return (
+            sum(estimated_completion_times_of_nodes[:self.min_requirement]),
+            sum(estimated_completion_times_of_nodes[-self.min_requirement:]),
+        )
```
muhammad-ammar: Can we please add some explanation regarding how the above code works?

muhammad-ammar (approved):
LGTM. Please see my comment regarding improving code documentation.

course_discovery/apps/learner_pathway/models.py
```diff
@@ -181,8 +181,21 @@ def remove_node(self, uuid):
...
+        """
+        Get a range of estimated time of completion.
+
+        Returns:
+            (tuple<int, int>): A tuple containing the min and max of estimated time of completion.
+        """
+        estimated_completion_times_of_nodes = sorted(
+            [node.get_estimated_time_of_completion() for node in self.get_nodes()]
+        )
+        estimated_completion_times_of_nodes = sorted(estimated_completion_times_of_nodes)
```
muneebGH: we are sorting the list two times(one in upper line) is this ok?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -123,7 +123,7 @@ def time_of_completion(self) -> float:
         """
         completion_time = 0
         for step in self.steps.all():
-            completion_time += step.get_estimated_time_of_completion()
+            completion_time += avg(step.get_estimated_time_of_completion())
```
muneebGH, line 126: if there is a block in the step and that returns 'Not known' through [this](https://github.com/openedx/course-discovery/blob/850b8b97a22ef94bec08121bfbd76c92620ead0a/course_discovery/apps/learner_pathway/models.py#L321)
that string will be a part of list and an exception might be thrown, what do you? plus do we need a separate function for avg? can we like paste the one liner here?

course_discovery/apps/learner_pathway/models.py
saleem-latif: Done

course_discovery/apps/learner_pathway/models.py
saleem-latif: I forgot to remove this line, good catch

course_discovery/apps/learner_pathway/models.py
saleem-latif: I have updated the return type of block to `int` all `get_estimated_time_of_completion` methods should have the same type as the base class `LearnerPathwayNode`. 
Regarding `avg` it seemed more readable to add a separate. function for this.

muneebGH (approved):
LGTM

course_discovery/apps/learner_pathway/models.py
muneebGH: oky cool

PR #3416, opened 2022-05-09
feat: ADR for syncing external course(s)

uzairr:
Add ADR document for the approaches to sync product lines from external source(s).

PROD-2754


docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+Decision
+========
+
+Pending
+
+Context
+=======
+
+In order to ingest course data from external source so that its relevant course
+instance would be created in discovery system, a data payload is needed.
```
DawoudSheraz: - Creation is not the sole reason. Course and course run updates are too.
- LoBs managing data their on end, discovery a central place to hold all data

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+
+The discovery must sync the courses without causing downtime or poor performance
```
DawoudSheraz: no need to make this a heading. This should be text under **Context** heading

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+================================================================================
+
+To avoid any unexpected event, one should ensure that the payload must be made
+compatible with the course creation api(s).While syncing course(s) data from
```
DawoudSheraz: Creation + update

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
+External course(s) data in discovery
```
DawoudSheraz: External product lines as opposed to courses.

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+external source, data transformers must be placed before the invocation of
+course creation api(s) so that the payload is reshaped and validated as per the
+api(s) need.One should also keep in account that multiple internal and external
+api(s) are needed to be invoked for a course to be created like instructor,
+ecommerce etc.
+
+Following are potential arguments that may cause downtime:
+
+- High volume of data may slows down the transformer causing downtime.
+- The invocation order of api(s) must be ensured to avoid any dirty data in the DB.
```
DawoudSheraz: This is more of a data integrity issue as compared to downtime.

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Possible Solutions
```
DawoudSheraz: This should be renamed to Alternative Considered. After giving the context, we should explain what our selected approach is (separate API). Once we are done with that, we can explain the alternate approaches and why we didn't go that way.

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+discovery. In terms of the async operations, a single api can serve the purpose
+better than multiple api(s).
+
+Suggested API implementation approaches ...
+===========================================
+
+As we already know, multiple including external and internal api(s) are invoked
+to create a course which may affect the response time due to different
+architectural contraints in systems.This led to the inquiry of a feasible
+mechanism in terms of execution.
```
DawoudSheraz: Here, we can go more in-detail of why a separate API is needed. Some itemss are touched, but we still can add details like:
- Separate course and course run APIs, the different payload format
- No API for degrees yet, its payload will be different
- Each LoB will interact with different API which can cause communication gaps
- Future goal of Event based design

docs/decisions/0012-external-course-data.rst
```diff
@@ -0,0 +1,87 @@
...
+
+To avoid downtime and poor-performance, celery workers should be consumed.
+This approach demands a separate api to track the status of an event/celery task.
+
+Synchronous execution
+=====================
+
+To avoid downtime and poor-performance, the incoming request to the api is
+served under controlled conditions like throttling. To have a better idea,
+NR(new-relic) statistics can be utilized to decide about the synchronous
```
DawoudSheraz: We should not mention newRelic here. We should focus more on Why as opposed to How

docs/decisions/0012-external-course-data.rst
uzairr: but it can lead to a downtime, thoughts ??

DawoudSheraz: In the case of data integrity issues, it would be a one-time server error, no?

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Decision
+========
+
+Pending
+
+Context
+=======
+
+In order to sync course data in discovery from external source without causing
```
DawoudSheraz: replace: from external source -> with data from other lines of businesses.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+downtime or poor performance so that its relevant course and course run instance(s)
```
DawoudSheraz: suggestion: After "poor performance", we should break the line. It is a long line and difficult to read.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+are created and updated later on down the line, an interface which is readily
+available and a data payload is needed.To avoid any unexpected event, one should
```
DawoudSheraz: update: interface accepting data payload in a pre-defined format.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+updation api(s).While syncing course(s) data from external source, data transformers
+must be placed before the invocation of creation/updation api(s) so that the
+payload is reshaped and validated as per the api(s) need.One should also keep
+in account that multiple internal and external api(s) are needed to be invoked
+for a course to be created or updated like instructor, ecommerce etc.
+
+Following are potential arguments that may cause downtime:
+
+- High volume of data may slows down the operation causing downtime.
+- The invocation order of api(s) must be ensured to attain a sequence.
```
DawoudSheraz: Instead of a list, we can turn it into a paragraph and add a bit more details.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Alternatives Considered
```
DawoudSheraz: This heading here does not make sense. It should be Possible  Approaches or something like that.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+=======================
+
+To better solve the above potential causes, there are two approaches either of
+them can be utilized.
+
+- a management command responsible to populate course(s) data which can ensure
+data management and invocation order of api(s).
+- a new api which can be written to serve the same purpose.
+
+Separate API, why ?
```
DawoudSheraz: Same, the heading should be Selected Approach

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+===================
+
+There is a need to track the operation in terms of event name which will track
```
DawoudSheraz: We can explain the selected approach in 1-2 lines and then explain why it has been selected.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+the scope of a requested change i.e. either the change has an impact
+on instructor, degree, course, course-run etc or a complete operation
+of a course creation.This flexibility in operation demands different payload(s)
+as per the nature of a needed change.The api(s) already part of the system
+would create communication gap with external bodies because each of them
+represents a line of business(LoB) with its own status.
```
DawoudSheraz: Along with the comm gap, they will have to deal with multiple APIs for various operations which can be overwhelming.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Having a separate api for this job is better suited based on the context,
+above requirements and already existed functional constraints compared to
+management command(s) which are needed to be triggered manually while an api is
+readily available via url.Currently, there is no api in discovery which can be
```
DawoudSheraz: this line is hard to understand. It can be re-worded and broken down.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+- invocation order of the api(s).
+- A better developer experience of feedback to locate the exact point of failure.
+- bridging different line of businesses(LoBs)
+- Payload management based on event name
+
+Note: This api act as a gateway by which external course(s) are created and
+updated in discovery. In terms of the async operations, a single api can serve
+the purpose better than multiple api(s).
+
+Suggested API implementation approaches ...
```
DawoudSheraz: We should explain the selected(sync approach) in this section only.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,91 @@
...
+===========================================
+
+As we already know, multiple including external and internal api(s) are invoked
+to create a course which may affect the response time due to different
+architectural contraints in systems.This led to the inquiry of a feasible
+mechanism in terms of execution.
+
+Asynchronous execution
```
DawoudSheraz: At the end, have a new section, Alternates Considered/Rejected Approaches and explain async along with other rejected approaches and why they were rejected.

DawoudSheraz (commented):
- Added a few comments, most related to data-position/small tweaks to the text

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+Pending
+
+Context
+=======
+
+In order to sync course data in discovery from other lines of businesses
+without causing downtime or poor performance, an interface accepting data payload
+in a pre-defined format is needed.
+To avoid any unexpected event, one should ensure that the payload must be made
+compatible with the course creation and updation api(s).While syncing course(s)
```
DawoudSheraz: We can remove the word `course` from here and instead add products' data API(cause we will be adding Degrees here too).

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+data from external source, data transformers must be placed before the invocation
+of creation/updation api(s) so that the payload is reshaped and validated as
+per the api(s) need.One should also keep in account that multiple internal and
+external api(s) are needed to be invoked for a course to be created or updated
```
DawoudSheraz: same point.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+like instructor, ecommerce etc.
+
+Following are potential situations that may lead a downtime:
```
DawoudSheraz: this line does not make sense here anymore, please remove it.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+
+At any given time, there is a probability that the discovery may encounter a
+high flux of data from different lines of businesses(LoBs).The current api(s)
+are not much well-written to validate all the edge-cases in data payload that
```
DawoudSheraz: Well, the wrapper API will also be calling these APIs. The well-written argument seems a bit out of place here.  Edge cases is a valid point here though.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+- invocation order of the api(s).
+- A better developer experience of feedback to locate the exact point of failure.
+- bridging different line of businesses(LoBs)
+- Payload management based on event name
+
+Note: This api act as a gateway by which external course(s) are created and
+updated in discovery. In terms of the async operations, a single api can serve
+the purpose better than multiple api(s).
+
+Suggested API implementation approache(s) ...
```
DawoudSheraz: We can remove this heading and bring the below paragraph under Feasible approach section

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Pending
+
+Context
+=======
+
+In order to sync course data in discovery from other lines of businesses
+without causing downtime or poor performance, an interface accepting data payload
+in a pre-defined format is needed.
+To avoid any unexpected event, one should ensure that the payload must be made
```
azanbinzahid: We could directly name respective stakeholders instead of `one`

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+
+At any given time, there is a probability that the discovery may encounter a
+high flux of data from different lines of businesses(LoBs).The current api(s)
+are not much well-written to validate all the edge-cases in data payload that
+may result in flooding validation errors.Other than that, database operations
+would be drastically affected in terms of availability.
+
+Moreover, in order to create/update a course, there are multiple apis that are
+needed to be invoked in a series pattern.So, if the invocation order is not
+maintained then data integrity will be compromised.This situation would create
```
azanbinzahid: We can be be more specific, like what APIs and in which order.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+an overhead in terms of communication because multiple teams need to be connected
+to resolve the data integrity conflict.
+
+Possible Approaches
+===================
+
+To better solve the above potential causes, there are two approaches either of
```
azanbinzahid: causes or issues?

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,99 @@
...
+the purpose better than multiple api(s).
+
+Suggested API implementation approache(s) ...
+===========================================
+
+To avoid downtime and poor-performance, the incoming request to the api is
+served under controlled conditions like throttling.There is an added benefit
+to this approach is, a separate status api is not needed which would lessen the
+overhead in terms of communication and implementation with concerned external bodies.
+
```
azanbinzahid, line 81: How we will get the status of bulk ingestion?

azanbinzahid (commented):
api -> API
could be more specific instead of general
grammar/punctuation errors.

docs/decisions/0012-external-product-lines.rst
uzairr: thanks for mentioning! here `well-written to validate all the edge-cases` is used in the same sense that you suggested!

docs/decisions/0012-external-product-lines.rst
uzairr: there are multiple ways to track the status like adding logs while implementing the new API ...

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+Decision
+========
+
+Pending
+
+Context
+=======
+
+To sync product(s) data in discovery from other lines of businesses
```
DawoudSheraz: nit: Product(s) -> Products; to be updated across all the doc. We know we will be dealing with multiple products, so let's keep the terms consistent.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+without causing downtime or poor performance, the interface accepts data payload
+in a pre-defined format is needed.
```
DawoudSheraz: the interface that will accept data payload in a pre-defined format is required.

docs/decisions/0012-external-product-lines.rst
DawoudSheraz: Additionally, we can be specific about unexpected event, such as successful course creation, update, instructors update, etc,

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+To avoid any unexpected event, one should ensure that the payload must be made
+compatible with the API, which would be used to add product(s).While syncing product(s)
```
DawoudSheraz: When reading the sentence, the part `, which would be used to add product(s)` does not make sense. It can be removed.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+data from an external source, data transformers must be placed before the invocation
```
DawoudSheraz: We can explain in one line that there will be some differences in data representation across discovery and external LoBs. The transformers would be more inclined toward fields mapping and now the actual data transformation as such. The actual data representation will be done by LoB as per data contract.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+of creation/update API(s) so that the payload is reshaped and validated as
+per the API(s) need. One should also keep in account that multiple internal and
+external API(s) are needed to be invoked for a product to be created or updated
+like the instructor, eCommerce etc.
```
DawoudSheraz: We can specify which are internal vs external in instructor and ecommerce example.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+
+At any given time, there is a probability that the discovery may encounter a
+high flux of data from different lines of businesses(LoBs).The current API(s)
+are not well-written to validate all the edge-cases in data payload that may
+result in flooding validation errors. Other than that, database operations would
+be drastically affected in terms of availability.
```
DawoudSheraz: nit: database operations would be drastically affected which can result in availability issues

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Moreover, to create/update a product, multiple APIs need to be invoked in a
+series pattern. So, if the invocation order is not maintained, data integrity
+will be compromised. This situation would create overhead in communication
+because multiple teams need to be connected to resolve the data integrity conflict.
```
DawoudSheraz: Please add an explicit example here.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Possible Approaches
+===================
+
+To better solve the above potential issue(s), there are two approaches either of
```
DawoudSheraz: nit: issue(s) -> issues

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+them can be utilized.
+
+- a management command responsible for populating product(s) data which can ensure
+data management and invocation order of API(s).
+- A new API that can be written to serve the same purpose.
+
+Feasible Approach
```
DawoudSheraz: Rename this to Selected Approach.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+=================
+
+A separate API for this job is better suited, which would take care of the management
+of data payload as well as invocation sequence of API(s).
+
+There is a need to track the operation in terms of the event name, which will
```
DawoudSheraz: Update: There is a need to track the operation in terms of the event name -> By modeling the changes in LoB products as events

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+follow the scope of a requested change, i.e., whether the change impacts
+instructor, degree, course, course-run etc. or a complete operation of product
+creation. This flexibility in the process demands different payload(s) as per
+the nature of a needed change. The API(s) are already part of the system would
+create a communication gap as they would have to deal with multiple APIs for
+various operations, which can be overwhelming.
+
+Based on the context and already existing functional constraints, a separate API
+is better suited because it would be readily available. In contrast to the API,
```
DawoudSheraz, line 65: The existing APIs will also be readily available. This does not come as a benefit of a new API.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+API(s) in async operations.
+
+Alternates considered/rejected approache(s)
+===========================================
+
+Async execution can also be considered via celery workers. If this approach is
+considered then it will demand another API which gives the feedback related to
+event/celery task.
+
+In addition to this, it may cost infrastructure in terms of creation and
```
DawoudSheraz: We had another rejected approach, please add it too, alongside the information on why it was not accepted.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,93 @@
...
+like the instructor, eCommerce etc.
+
+At any given time, there is a probability that the discovery may encounter a
+high flux of data from different lines of businesses(LoBs).The current API(s)
+are not well-written to validate all the edge-cases in data payload that may
+result in flooding validation errors. Other than that, database operations would
+be drastically affected in terms of availability.
+
+Moreover, to create/update a product, multiple APIs need to be invoked in a
+series pattern. So, if the invocation order is not maintained, data integrity
```
ansabgillani: Let's rename this to queue pattern (Recognized DS).

docs/decisions/0012-external-product-lines.rst
uzairr: benefit in terms of calling a single API rather than multiple

docs/decisions/0012-external-product-lines.rst
uzairr: can you remind me that one ?

docs/decisions/0012-external-product-lines.rst
DawoudSheraz: That is ease of use, not an availability benefit.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Decision
+========
+
+Pending
+
+Context
+=======
+
+To sync products data in discovery from other lines of businesses without causing
```
davidjoy, line 17: We should try to make sure the ADR isn't 2U-specific since this repository is part of Open edX.

I'd suggest we phrase this in more general terms.  i.e., "provide an API to sync courses from external data sources".

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+======
+
+In Review
+
+Decision
+========
+
+Pending
+
+Context
```
davidjoy, line 14: The "Context" section is about the situation which led us to need to make a decision.  So it's more about our current state and the problems we're facing, rather than how we're solving them.  So it'd be appropriate to say things like: 

1. External sources have their own product schemas and today the owners of course-discovery are responsible for transforming each of those into our own internal schema.  
2. We want the owners of the external data sources to be responsible for that translation and for course-discovery to present a unified, consistent API which external callers should conform to.
3. We don't think the current set of APIs provided by course-discovery is very approachable/usable for external consumers and requires a great deal of domain knowledge to use correctly.  This is a great place to talk about the ways it's difficult, and give some examples of the sequencing required to call them correctly.
4. We don't think the current set of APIs give very good error messages or feedback, contributing to that lack of approachability.
5. We have concerns about course-discovery's ability to handle an influx of data at scale. (Side note: I'm not sure if our current approach actually addresses this, but I see it called out in this ADR)

Those are the things that come to mind for me as problems with our current state - did I miss any?

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+internal and external API(s) must be invoked, like the instructor and eCommerce,
+respectively.
+
+At any given time, there is a probability that the discovery may encounter a
+high flux of data from different lines of businesses(LoBs).The current API(s)
+are not well-written to validate all the edge-cases in data payload that may
+result in flooding validation errors. Other than that, database operations would
+be drastically affected which can result in availability issues.
+
+Moreover, to create/update a product, multiple APIs need to be invoked in a
```
davidjoy, line 35: This paragraph is great context. 👍🏻

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+queue pattern. So, if the invocation order is not maintained, data integrity
+will be compromised. This situation would create overhead in communication
+because multiple teams need to be connected to resolve the data integrity conflict.
+
+Possible Approaches
```
davidjoy, line 40: I think we can remove this short section in favor of the next two; I don't think it's adding much.  The approach we chose will be explained in detail in the next section, and any rejected alternatives will be discussed below that.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+===================
+
+To better solve the above potential issues, there are two approaches either of
+them can be utilized.
+
+- a management command responsible for populating products data which can ensure
+data management and invocation order of API(s).
+- A new API that can be written to serve the same purpose.
+
+Selected Approach
```
davidjoy, line 50: This section is generally called "Decision" and talks about how we've chosen to address the issues described in Context.  For consistency with other ADRs, I'd suggest just using that title.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+=================
+
+A separate API for this job is better suited, which would take care of the management
+of data payload as well as invocation sequence of API(s).
+
+By modeling the changes in LoB products as events, which will
```
davidjoy, line 56: I'm not quite sure what we're saying in this paragraph.  

I think this is saying that there are multiple types of models which we want external data sources to be able to add to course-discovery, so our APIs will need to handle all of them.  To me that implies that maybe we have multiple APIs we're creating?  I think we should be wary of creating one big single API that accepts multiple types of data - in my experience usually that results in something that's very difficult to maintain and reason about.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+will be compromised. This situation would create overhead in communication
+because multiple teams need to be connected to resolve the data integrity conflict.
+
+Possible Approaches
+===================
+
+To better solve the above potential issues, there are two approaches either of
+them can be utilized.
+
+- a management command responsible for populating products data which can ensure
```
davidjoy, line 46: I think this "management command" idea, for instance, should be moved down to Rejected Alternatives and described in more detail, I'm not sure if I understand how it would be used.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+is better suited because it would be readily available. In contrast to the API,
+the management command must be triggered manually. Currently, there is no API
+in the discovery that can be used to handle the data from an external source and
+create/update products objects accordingly, so if a new API is designed, it is
+expected to be responsible for:
+
+- Management of external products data.
+- Invocation order of the API(s).
+- A better developer experience of feedback to locate the exact point of failure.
+- Bridging different lines of businesses(LoBs)
```
davidjoy, line 74: I don't quite understand what it means to bridge different lines of business... is this talking about how different external data sources have the same sort of data type (i.e., a course) but with differing schemas? 

Reminder, we should remove language around "line of business" in favor of just talking about external data sources.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+- Payload management based on an event name
+
+To avoid downtime and poor performance, the incoming request to the API is
```
davidjoy, line 77: Thinking about downtime/performance, some thoughts using some examples of scale from the edX/2U instances:

1. Today, course-discovery regularly gets course data dumps from edx-platform.  In terms of scale, edx.org and our instance of edx-platform has many more courses than any other known data source (including other 2U LoBs).
2. I can't imagine the volume or frequency of changes from other external data sources would put more strain on course-discovery than edx.org's data already does. 
3. If we want to control API pressure on course-discovery, in the long term we probably want to handle that with a queue or event bus and not rely solely on throttling logic, which puts the burden of retry logic, etc., on the API's consumers.

When I put all that together, I'd want to propose that we:

1. Sanity check these assumptions!
2. Assume that in the short term, course-discovery will be able to handle the volume coming from other external data sources, but add the throttling logic anyway for safety.  (i.e., we don't really expect anyone will actually get throttled)
3. Plan to work with Arch-BOM on inserting an event bus into the system to decouple course-discovery from external data sources trying to put data into the system, effectively using the event bus as a queue allowing course-discovery to pull events off of it at its own pace.  

In support of the 3rd bullet there - one small thing we could do is to not point our consumers directly at course-discovery, but at an API Gateway so that we can later introduce an event bus or queue without changing the contract.  I'm happy to help work with Arch-BOM to get this set up.

It's not clear to me that we'd need a "status API" in either case, so long as the external data source/consumer gets a return status code on their request that indicates success.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+served under controlled conditions like throttling. An added benefit
+to this approach is that a separate status API is not needed, which would lessen the
+overhead in terms of communication and implementation with concerned external bodies.
+
+Note: This API act as a gateway by which external products are created and
+updated in discovery. A single API can serve the purpose better than multiple
+API(s) in async operations.
+
+Rejected Alternatives
```
davidjoy, line 86: As mentioned above, I'd move the content about the management command idea down here since we're pretty sure it's not a thing we're going to do.

docs/decisions/0012-external-product-lines.rst
```diff
@@ -0,0 +1,94 @@
...
+=====================
+
+Async execution can also be considered via celery workers. If this approach is
```
davidjoy, line 89: @robrap, you and I had briefly discussed whether or not we thought it was a good idea to consider using a celery queue for this API.  Just tagging you in case you have any thoughts on this rejected alternative.

robrap: @uzairr: I'd recommend that you and your team get familiar with https://openedx.atlassian.net/wiki/spaces/AC/pages/1074397222/Architecture+Manifesto+WIP, especially the section on "Asynchronous beats Synchronous". Hopefully this will help you see better the benefits and reasons why we rely on asynchronous tasks. This can be discussed further once that is better understood.

robrap: @uzairr: I'd recommend that you and your team get familiar with https://openedx.atlassian.net/wiki/spaces/AC/pages/1074397222/Architecture+Manifesto+WIP, especially the section on "Asynchronous beats Synchronous". Hopefully this will help you see better the benefits and reasons why we rely on asynchronous tasks. This can be discussed further once that is better understood.

davidjoy: In thinking a little more about the nature of this new API and the type of requests we expect it to handle - and in line with Robert's suggestion above - I think we should strongly consider using celery to create asynchronous tasks for catalog processing.

We want to keep inter-service communication APIs performant; this seems like a prime candidate for asynchronously processing since we expect to be receiving data in bulk.

I'd also call out that I don't think we necessary need a "status" API if we go this route.  I don't expect there will be anything on the caller's side monitoring status for these use cases necessarily.  If anything, we need to understand the human workflows and our consumers' expectations before we could even say.

robrap: I'm wondering if "status" was the results of data validation errors, etc.? Also, note that if this is meant to be a bulk upload of changes, that could be made more clear in the ADR language.

PR #3422, opened 2022-05-16
feat: Add a learner pathway endpoint to return uuids of all pathways having course run key or program uuid

muhammad-ammar:


course_discovery/apps/learner_pathway/api/v1/views.py
```diff
@@ -33,6 +33,30 @@ def snapshot(self, request, uuid):
...
+        )
+        pathway_programs = models.LearnerPathwayProgram.objects.filter(
+            program__uuid__in=program_uuids
+        ).values_list(
+            'step__pathway__uuid',
+            flat=True
+        )
+
+        ids = list(pathway_courses) + list(pathway_programs)
+        return Response(set(ids))
```
saleem-latif, line 58: Would it be better to return a response like

```
{ "results": 
  [
    {
        "uuid": "<course-uuid>",
        "content_type": "course",
    },
    {
        "uuid": "<program-uuid>",
        "content_type": "program",
    }
  ]
}
```

saleem-latif (approved):
One comment, other than that looks good  👍

irfanuddinahmad (approved):
You could add the input and output format in the PR description.

PR #3427, opened 2022-05-20
feat: Add model to store addtional_metadata fields for degrees

AliAdnanSohail:
[add a new additional metadata model for degree/program](https://2u-internal.atlassian.net/browse/PROD-2805)

### Description: 
We need to store new fields for Degrees, so we created a new model named DegreeAdditionalMetadata which has one-to-one relation with Degree model. While creating Degree object we will send these new fields which will be added into DegreeAdditionalMetadata model of that Degree.

### Testing Instruction:
For now to test it we need to use Django Admin 
URL: <BASE_URL>/admin/course_metadata/degree/add/
Screen Shot 2022-05-26 at 11 57 47 PM

From drop down user can select existing or can create a new object by clicking on '+'

### Notes:
I've added this relation in DegreeAdditionalMetadata model to get input of these fields with Degree fields in django admin using inline input. But if we do this and delete DegreeAdditionalMetadata object then it will delete Degree object as well. And this way user can create independent DegreeAdditionalMetadata objects without relating to Degree

So now I'm adding this relation to Degree model and user will have to select/create DegreeAdditionalMetadata from dropdown, and if use delete DegreeAdditionalMetadata object then it will not effect to Degree object. Also user can not create DegreeAdditionalMetadata independently, user must have to create degree as well


course_discovery/apps/course_metadata/models.py
```diff
@@ -2750,6 +2750,33 @@ def __str__(self):
...
+        blank=True, null=False, max_length=511,
+        help_text=_('The URL of the paid landing page on external site')
+    )
+
+    degree = models.OneToOneField(
+        Degree,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+        default=None
```
DawoudSheraz: Please add related_name here as well.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2750,6 +2750,33 @@ def __str__(self):
         return str(f'Degree: {self.title}')
 
 
+class DegreeAdditionalMetadata(TimeStampedModel):
+    """
+    This model holds 2U degree related additional fields
+    """
+
+    external_url = models.URLField(
```
azanbinzahid: Why are we not allowing it to be null/blank?

course_discovery/apps/course_metadata/models.py
```diff
@@ -2750,6 +2750,33 @@ def __str__(self):
...
+        blank=False, null=False, max_length=511,
+        help_text=_('The redirect URL of the course on external site')
+    )
+    external_identifier = models.CharField(max_length=255, blank=True, null=False)
+    organic_url = models.URLField(
+        blank=True, null=False, max_length=511,
+        help_text=_('The URL of the paid landing page on external site')
+    )
+
+    degree = models.OneToOneField(
```
azanbinzahid: can we make this relation in degree model?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -653,6 +653,11 @@ class DegreeCostInlineAdmin(admin.StackedInline):
     extra = 1
 
 
+@admin.register(DegreeAdditionalMetadata)
+class DegreeAdditionalMetadataAdmin(admin.ModelAdmin):
+    list_display = ('id', 'external_url', 'external_identifier', 'organic_url')
```
DawoudSheraz: Instead of id, the degree UUID and degree title can be displayed here.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2743,6 +2762,15 @@ class Degree(Program):
         null=True,
     )
 
+    degree_additional_metadata = models.OneToOneField(
```
DawoudSheraz, line 2766: nit: additional_metadata
The model has Degree in the name, hence is no need to add it to the field.

course_discovery/apps/course_metadata/admin.py
AliAdnanSohail: Rather displaying degree UUID and title, I'm gonna display whole degree model like this

Screen Shot 2022-05-23 at 9 08 16 PM

course_discovery/apps/course_metadata/models.py
AliAdnanSohail: It is recommended to keep the column name same as the OneToOne related table name

course_discovery/apps/api/serializers.py
```diff
@@ -637,6 +637,14 @@ class Meta:
         )
 
 
+class DegreeAdditionalMetadataSerializer(BaseModelSerializer):
```
DawoudSheraz, line 640: Add it within DegreeSerializer to expose the data within API

course_discovery/apps/course_metadata/models.py
AliAdnanSohail: I've added this relation in DegreeAdditionalMetadata model to get input of these fields with Degree fields in django admin using inline input. But if we do this and delete DegreeAdditionalMetadata object then it will delete Degree object as well. And this way user can create independent DegreeAdditionalMetadata objects without relating to Degree

So now I'm adding this relation to Degree model and user will have to select/create DegreeAdditionalMetadata from dropdown, and if use delete DegreeAdditionalMetadata object then it will not effect to Degree object. Also user can not create DegreeAdditionalMetadata independently, user must have to create degree as well

azanbinzahid (commented):
Looks good other than missing test suits.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: ☝🏽 Add this to the PR description and the related internal ticket.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Can you reference this, please? Thank you

DawoudSheraz:
Apart from the field name, it is good overall. Add missing unit tests, fix the failing tests, and rebase the branch.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2619,6 +2619,25 @@ def __str__(self):
...
+    """
+
+    external_url = models.URLField(
+        blank=True, null=False, max_length=511,
+        help_text=_('The redirect URL of the degree on external site')
+    )
+    external_identifier = models.CharField(max_length=255, blank=True, null=False)
+    organic_url = models.URLField(
+        blank=True, null=False, max_length=511,
+        help_text=_('The URL of the paid landing page on external site')
```
DawoudSheraz: organic url is not paid landing page. **external_url** is.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: @AliAdnanSohail

DawoudSheraz (approved):
- Rebase and make sure the build is green
- Nit comments to address
- Update PR with screenshot and the field placement decision. Copy that to internal ticket too.

course_discovery/apps/course_metadata/models.py
AliAdnanSohail: Reference link: https://www.geeksforgeeks.org/python-relational-fields-in-django-models/


Screen Shot 2022-05-26 at 11 49 32 PM

azanbinzahid (approved):
I do think we should have `test_degrees` inside [test_views](https://github.com/openedx/course-discovery/tree/master/course_discovery/apps/api/v1/tests/test_views). However, if it requires more effort, we can make another PR for this. 

Also, add TODO may be in desc that more fields [needs to | will] be added in this model as we get more clarity. 

Once the build is green!

AliAdnanSohail:
@azanbinzahid yes we can add degree view test when we will work on its API

secondly I think we don't need to add TODO for future fields, when we need them to add we can simply add them. 
TODOs help us when we are missing something in current scope. For future we don't need to add TODO

PR #3430, opened 2022-05-24
feat: update refresh_course_metadata command

zainab-amir:
Added option to run the dependant stages separately. This can unblock other stages if one fails. Also will reduce time to run the command.

Currently we can run parallelised execution of the data loading pipeline. The pipeline is structured so that stages that can run independently of others are interleaved using separate processes. However the management command runs as one and takes a lot of time to finish.

This argument is not required and we can ignore it to run the command as it runs today.

[VAN-941](https://2u-internal.atlassian.net/browse/VAN-941)


waheedahmed (approved):
Can we add a test case for this, otherwise looks good!

PR #3434, opened 2022-05-25
fix: pin django-stdimage version

UsamaSadiq:


UsamaSadiq:
Seems like this PR is not needed since the test failures were only occurring due to flaky test failures in course-discovery.

PR #3436, opened 2022-05-25
fix: remove skip-review step from csv loader

DawoudSheraz:
### Description

- Remove the step of setting course run status as reviewed if draft=False. This means that a course create from scratch via loader will move in legal view and only have draft versions of the course
    - This was added to move a course directly into published status, skipping internal and legal reviews. Not needed anymore
- Update course and course run query method in test to `everything` instead of `objects`. objects filter out draft versions of the course
- Adding a temporary, default User Agent for the org logo overrides download requests call, to avoid 403 from 2U CDNs.

### Notes
- If the course and course run are already published, running the loader with no draft flag will push the changes to both draft and non-draft versions of the course and course run.
- If the CSV contains an entry for a new course and the draft flag is not provided, the course and course run will be created, and the course run will move into the LegalReview phase.
    - In a follow-up PR, the draft flag will be removed from CSV loader. Instead, using course run status, the draft flag value will be determined for each row. If a course has any published course run, the draft will be False. Otherwise true. Reference: https://github.com/openedx/frontend-app-publisher/blob/master/src/components/EditCoursePage/index.jsx#L204-L213

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -134,7 +134,12 @@ class CSVDataLoader(AbstractDataLoader):
                 is_logo_downloaded = download_and_save_course_image(
                     course,
                     row['organization_logo_override'],
-                    'organization_logo_override'
+                    'organization_logo_override',
+                    # TODO: Temporary addition of User agent to allow access to data CDNs
+                    headers={
+                        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 '
+                                      '(KHTML, like Gecko) Chrome/101.0.4951.64 Safari/537.36'
+                    }
                 )
                 if not is_logo_downloaded:
                     logger.error("Unexpected error happened while downloading override logo image for course {}".format(  # lint-amnesty, pylint: disable=logging-format-interpolation
@@ -146,17 +151,10 @@ class CSVDataLoader(AbstractDataLoader):
             if not course_run.in_review:
                 try:
                     self._update_course_run(row, course_run, course_type)
-                    course_run.refresh_from_db()
                 except Exception:  # pylint: disable=broad-except
                     logger.exception("An unknown error occurred while updating course run information")
                     self.messages_list.append('[COURSE RUN UPDATE ERROR] course {}'.format(course_title))
                     continue
-
-            if not self.is_draft:
-                try:
-                    self._complete_run_review(row, course_run)
-                except Exception:  # pylint: disable=broad-except
-                    logger.exception("An unknown error occurred while completing course run review")
 
             logger.info("Course and course run updated successfully for course key {}".format(course_key))  # lint-amnesty, pylint: disable=logging-format-interpolation
             self.course_uuids[str(course.uuid)] = course_title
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -134,7 +134,12 @@ def ingest(self):  # pylint: disable=too-many-statements
                 is_logo_downloaded = download_and_save_course_image(
                     course,
                     row['organization_logo_override'],
-                    'organization_logo_override'
+                    'organization_logo_override',
+                    # TODO: Temporary addition of User agent to allow access to data CDNs
+                    headers={
+                        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 '
+                                      '(KHTML, like Gecko) Chrome/101.0.4951.64 Safari/537.36'
+                    }
```
azanbinzahid, line 142: we can define it inside the function.

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -72,7 +72,7 @@ class CSVLoaderMixin:
         'external_identifier', 'syllabus', 'frequently_asked_questions'
     ]
     BASE_EXPECTED_COURSE_DATA = {
-        'draft': False,
+        'draft': True,
```
azanbinzahid, line 77: Can we put a comment here to document why we are doing it?

azanbinzahid (commented):
a few nits to address.

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
DawoudSheraz: Sure.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
DawoudSheraz: which function?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -146,18 +151,11 @@ def ingest(self):  # pylint: disable=too-many-statements
             if not course_run.in_review:
                 try:
                     self._update_course_run(row, course_run, course_type)
-                    course_run.refresh_from_db()
                 except Exception:  # pylint: disable=broad-except
                     logger.exception("An unknown error occurred while updating course run information")
                     self.messages_list.append('[COURSE RUN UPDATE ERROR] course {}'.format(course_title))
                     continue
 
-            if not self.is_draft:
```
mulby, line 155: Will this impact our first boot camps ingestion?

DawoudSheraz: If bootcamps follow the same ingestion path as Executive Education, then no.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
azanbinzahid: the function that is being called: `download_and_save_course_image`

DawoudSheraz: This temporary header is only needed for logo override images downloaded from 2U CDN, that's why I added it here as opposed to the util method. In the util method, the header would have applied to every request.

azanbinzahid (approved):

uzairr (approved):

PR #3440, opened 2022-05-27
feat: update csv loader draft behavior to match publisher

DawoudSheraz:
### Description

- Update CSV loader draft behavior to be course and course run specific. This makes the loader behavior consistent with publisher. The publisher's behavior for draft flag is:
    - Course: If any of the course run is in published state, the draft is false, else true ([Reference](https://github.com/openedx/frontend-app-publisher/blob/master/src/components/EditCoursePage/index.jsx#L204-L213)).
    - Course Run: If target run or status is published. CSV Loader expects only one run/course, so the status of the fetched course run is checked (https://github.com/openedx/frontend-app-publisher/blob/master/src/components/EditCoursePage/index.jsx#L117-L124)

### Testing
- Make sure you have Executive education course type setup https://2u-internal.atlassian.net/wiki/spaces/IM/pages/18974245/Executive+Education+Course+Type#Unpaid-and-Paid-Track%2FEnrollment-Type-Variants
- Create the subjects "Economics & Finance" and "Design" from admin
- Create "Introductory" level type from admin
- Use the file: [dt.csv](https://github.com/openedx/course-discovery/files/8787129/dt.csv)
- Run the loader inside discovery shell `./manage.py import_course_metadata --csv_path=dt.csv`
- Both courses will be in an Unpublished state.
- ~~Now, publish one from Published, make edits to draft version from admin, and re-run the loader~~
- ~~You would see the draft False for the published course in the logs and the admin changes would be present in non-draft version too.~~
- Now, publish one from Published, make edits to CSV against a few fields, and re-run the loader
- The CSV updates will be present for both draft and non-draft versions of the course (Check via admin)

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -224,7 +224,7 @@ class CSVDataLoader(AbstractDataLoader):
         No 1 is not applicable at the moment. For 2, CSV loader right now only expects
         one course run for each course, hence the status of the single fetched course run is checked.
         """
-        return not (course_run.status == CourseRunStatus.Published)
+        return not course_run.status == CourseRunStatus.Published
 
     def _update_course_api_request_data(self, data, course, is_draft):
         """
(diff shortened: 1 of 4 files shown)


azanbinzahid (approved):
Tested locally, working as expected.

uzairr (approved):

AliAdnanSohail (approved):

Ali-D-Akbar (approved):

PR #3441, opened 2022-05-27
fix: move one to one relation in DegreeAdditionalMetadata model

AliAdnanSohail:
[add a new additional metadata model for degree/program](https://2u-internal.atlassian.net/browse/PROD-2805)

## Description
In our previous [PR](https://github.com/openedx/course-discovery/pull/3427) we added the One-to-One relation in Degrees model, but that was causing issue and is not recommended because by deleting Degree, DegreeAdditionalMetadata was not deleting but by deleting DegreeAdditionalMetadata, Degree was deleting. Which was not required. So in this PR we have reverted previous changes and add relation in DegreeAdditionalMetadata model to get expected behavior


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1357,7 +1358,7 @@ def test_degree_marketing_data(self):
             'costs_fine_print': degree.costs_fine_print,
             'deadlines_fine_print': degree.deadlines_fine_print,
             'title_background_image': degree.title_background_image,
-            'degree_additional_metadata': expected_degree_additional_metadata,
+            'additional_metadata': expected_degree_additional_metadata,
```
azanbinzahid, line 1361: do we need to rename this as well? `expected_additional_metadata`

DawoudSheraz (commented):
The changes look good overall, I will give the final 👍🏽 once I test it out on the local. Please fix CI issues.

course_discovery/apps/api/tests/test_serializers.py
AliAdnanSohail: @azanbinzahid other attributes are also named with prefix `expected_degree` for example `expected_degree_costs` So to keep it consistent I did the same

course_discovery/apps/course_metadata/migrations/0279_auto_20220527_1244.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.13 on 2022-05-27 12:44
```
DawoudSheraz, line 1: Moving forward, make sure to give migrations a meaningful name rather than the default timestamp.

DawoudSheraz (approved):

azanbinzahid (approved):

PR #3444, opened 2022-05-30
feat: Add new model to save specialization of Degree and update serializer

AliAdnanSohail:
[add specialization model for Degrees](https://2u-internal.atlassian.net/browse/PROD-2810)

## Description
We need to store specialization for Degrees. So this model will contain specialization value and association with degree

## Testing
To test it you need to go Add/Edit Degree page from Django Admin and you can create new or select existing specialization for that degree
Screen Shot 2022-05-30 at 6 57 29 PM


DawoudSheraz (commented):
The changes look good. I will give a final review after the local testing. Any reason why this PR is in draft?

DawoudSheraz (approved):
If it can be done in this PR, good. Not a blocker. The specialization returned in the API is list of dict. 
```
"specializations": [
                    {
                        "value": "Networking"
                    },
                    {
                        "value": "IT"
                    },
                    {
                        "value": "Blockchain"
                    }
                ]
```
Since this is a single-valued model, the serializer output can be list of strings instead of list of dicts.

azanbinzahid (approved):
As @DawoudSheraz mentioned, what we are missing is dedicated test suit for degree (listing) view.

DawoudSheraz (approved):
Quality failure to fix before merge

PR #3423, opened 2022-05-17
fix: Revert "feat: add course_type to search/all response json"

johnnagro:
## Description

 - Reverts openedx/course-discovery#3420
 - probably missing fields in `BaseAggregateSearchViewSet` like https://github.com/openedx/course-discovery/pull/2880/files
 - need to investigate an error:
 ```
2022-05-16 19:55:57,697 ERROR 1271 [django.request] /edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/django/utils/log.py:224 - Internal Server Error: /api/v1/search/all/
Traceback (most recent call last):
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/elasticsearch_dsl/utils.py", line 159, in __getattr__
    return self.__getitem__(attr_name)
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/elasticsearch_dsl/utils.py", line 178, in __getitem__
    return _wrap(self._d_[key])
KeyError: 'course_type'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/rest_framework/fields.py", line 457, in get_attribute
    return get_attribute(instance, self.source_attrs)
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/rest_framework/fields.py", line 97, in get_attribute
    instance = getattr(instance, attr)
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/elasticsearch_dsl/utils.py", line 161, in __getattr__
    raise AttributeError(
AttributeError: 'Hit' object has no attribute 'course_type'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/newrelic/hooks/framework_django.py", line 552, in wrapper
    return wrapped(*args, **kwargs)
  File "/edx/app/discovery/venvs/discovery/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
[...]

PR #3420, opened 2022-05-13
feat: add course_type to search/all response json

johnnagro:
## Description

The `course_type` field has been added to `Course` objects distinguish new 2U additions to our catalog, like Executive Education and Bootcamp content. Downstream consumers of Discovery - like Enterprise Catalog Service - require `search/all` responses to contain the new field for indexing/filtering.
 
## References

- [ENT-5823](https://2u-internal.atlassian.net/browse/ENT-5823)
- https://github.com/openedx/course-discovery/pull/3351
- https://github.com/openedx/enterprise-catalog/pull/448

PR #3454, opened 2022-06-02
feat: add course_type to search/all response json

johnnagro:
## Description

The `course_type` field has been added to `Course` objects distinguish new 2U additions to our catalog, like Executive Education and Bootcamp content. Downstream consumers of Discovery - like Enterprise Catalog Service - require `search/all` responses to contain the new field for indexing/filtering.

## References

- Reverts openedx/course-discovery#3423
- Previous attempt https://github.com/openedx/course-discovery/pull/3420
- [ENT-5823](https://2u-internal.atlassian.net/browse/ENT-5823)
- https://github.com/openedx/course-discovery/pull/3351
- https://github.com/openedx/enterprise-catalog/pull/448


course_discovery/apps/course_metadata/search_indexes/documents/course.py
```diff
@@ -119,6 +120,9 @@ def prepare_prerequisites(self, obj):
     def get_queryset(self):
         return super().get_queryset().prefetch_related('course_runs__seats__type')
 
+    def prepare_course_type(self, obj):
```
alex-sheehan-edx, line 123: how is this used?

alex-sheehan-edx (approved):
context needed over the prepare class method + I'm a little confused how course type would be set but otherwise looks good

course_discovery/apps/course_metadata/search_indexes/documents/course.py
johnnagro: it is the way our elasticsearch library processes the fields https://django-elasticsearch-dsl.readthedocs.io/en/latest/fields.html#using-prepare-field since we need to grab a specific field from a sub-object this seems to be the way

PR #3461, opened 2022-06-07
feat: suppress 2u products in catalog

uzairr:
## Description:
Exclude 2u products in course data returned from catalogs. This way 2u products will not appear on edX pages.

## Linked Ticket:
for more details please visit: [PROD-2835](https://2u-internal.atlassian.net/browse/PROD-2835)

### DIFF
--- a/course_discovery/apps/api/utils.py
+++ b/course_discovery/apps/api/utils.py
@@ -11,6 +11,7 @@ from opaque_keys.edx.keys import CourseKey
 from requests.exceptions import HTTPError
 from sortedm2m.fields import SortedManyToManyField
 
+from course_discovery.apps.core.api_client.lms import LMSAPIClient
 from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.models import CourseRun
 
@@ -142,6 +143,37 @@ def decode_image_data(image_data):
     ext = file_format.split('/')[-1]  # guess file extension
     image_data = ContentFile(base64.b64decode(img_str), name=f'tmp.{ext}')
     return image_data.name, image_data
+
+
+def check_catalog_api_access(partner, user):
+    """
+    Uses LMSAPIClient to check the catalog api access for a
+    given user
+
+    Arguments:
+        user (User): Django User.
+
+    Returns:
+        (dict): ApiAccessRequests for the given user.
+
+    Example:
+        {
+            "id": 1,
+            "created": "2017-09-25T08:37:05.872566Z",
+            "modified": "2017-09-25T08:37:47.412496Z",
+            "user": 5,
+            "status": "approved",
+            "website": "https://example.com/",
+            "reason": "Example Reason",
+            "company_name": "Example Inc",
+            "company_address": "Example Address",
+            "site": 1,
+            "contacted": True
+        }
+    """
+    lms_client = LMSAPIClient(partner)
+    api_access_response = lms_client.get_api_access_request(user)
+    return api_access_response
 
 
 class StudioAPI:
--- a/course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
@@ -3,6 +3,7 @@ import xml.etree.ElementTree as ET
 from os.path import abspath, dirname, join
 
 import ddt
+import mock
 import pytz
 from lxml import etree
 from rest_framework import status
@@ -15,7 +16,7 @@ from course_discovery.apps.core.tests.factories import UserFactory
 from course_discovery.apps.core.tests.helpers import make_image_file
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
 from course_discovery.apps.course_metadata.choices import CourseRunStatus
-from course_discovery.apps.course_metadata.models import ProgramType, Seat, SeatType
+from course_discovery.apps.course_metadata.models import CourseType, ProgramType, Seat, SeatType
 from course_discovery.apps.course_metadata.tests.factories import (
     CourseRunFactory, ProgramFactory, ProgramTypeFactory, SeatFactory, SeatTypeFactory
 )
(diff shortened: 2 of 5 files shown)


ansabgillani (approved):

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -91,9 +92,11 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin, unuse
         ---
         serializer: serializers.CatalogCourseSerializer
         """
+        products_lookup_query_for_2u = \
+            Q(type__slug=CourseType.EXECUTIVE_EDUCATION_2U) | Q(type__slug=CourseType.BOOTCAMP_2U)
         catalog = self.get_object()
-
-        queryset = catalog.courses()
+        # exclude 2u products so that they cannot appear on edx pages
+        queryset = catalog.courses().exclude(products_lookup_query_for_2u)
```
azanbinzahid: Can be refactored to something like this:

```
course_types_2U = [CourseType.EXECUTIVE_EDUCATION_2U, CourseType.BOOTCAMP_2U]
queryset = catalog.courses().exclude(type__slug__in=course_types_2U)
```

azanbinzahid (commented):
- Since this is a temp solution, please use `temp` as prefix
- I think filtering this view won't filter affiliate view, just to be sure, add a unit test for awin as well
- please make sure that filtering here does not affect prospectus data

course_discovery/apps/api/v1/views/catalogs.py
DawoudSheraz: +1 to what Azan has already said on various items. Affiliate has a separate endpoint, so this filter needs to be moved there too. Not certain if this is temporary but we can use feat! to imply breaking changes.

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -92,8 +92,12 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin, unuse
         serializer: serializers.CatalogCourseSerializer
         """
         catalog = self.get_object()
-
         queryset = catalog.courses()
+        # exclude 2u products for partners other than edX
+        if request.site.partner.name != 'edX':
```
DawoudSheraz: We should avoid using hard-coded strings for partner/site names in the comparison here.

uzairr: @DawoudSheraz i was of the same view unless i found this:
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/algolia_models.py#L257

DawoudSheraz: But we can keep our code clean. I don't agree with hard-coded in algolia_models either. I am certain there will already be a matching setting variable/constant for this.

course_discovery/apps/api/v1/views/affiliates.py
```diff
@@ -35,6 +35,10 @@ def retrieve(self, request, pk=None):
             raise PermissionDenied
 
         courses = catalog.courses()
+        if request.site.partner.name != 'edX':
```
azanbinzahid: I don't think edX consume this awin endpoint, hence we can remove this check here and exclude courses directly. Albeit, this needs to be verified.

uzairr: @azanbinzahid but i think we should make this check part of the system for better validation!

azanbinzahid (approved):

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -92,8 +93,12 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin, unuse
         serializer: serializers.CatalogCourseSerializer
         """
         catalog = self.get_object()
-
         queryset = catalog.courses()
+        # exclude 2u products for partners other than edX
+        if request.site.partner.name not in settings.PARTNER_NAMES:
+            course_types_2U = [CourseType.EXECUTIVE_EDUCATION_2U, CourseType.BOOTCAMP_2U]
+            queryset = queryset.exclude(type__slug__in=course_types_2U)
+
```
DawoudSheraz: Are we certain the enterprise users are coming in via different site and not edX?

DawoudSheraz (commented):
- Taking the conversation on the approach internally before a final review.

course_discovery/apps/api/v1/tests/test_views/test_affiliate_window.py
```diff
@@ -213,6 +214,44 @@ def test_with_closed_enrollment(self):
...
+                'course_discovery.apps.api.v1.views.affiliates.check_catalog_api_access',
+                return_value={'company_name': company_name}
+        ):
+            response = self.client.get(self.affiliate_url)
+            assert response.status_code == 200
+            root = ET.fromstring(response.content)
+            if company_name:
+                assert 1 == len(root.findall('product'))
+            else:
+                assert 3 == len(root.findall('product'))
```
DawoudSheraz: nit: we can add 1,3 as expected product count as part of ddt data item.

DawoudSheraz (commented):
I have not tested it on the local yet. A few small question/nits before final 👍🏽

DawoudSheraz (approved):
Locally tested.

course_discovery/apps/api/utils.py
```diff
@@ -144,6 +145,16 @@ def decode_image_data(image_data):
...
 
 
+def check_catalog_api_access(partner, user):
+    """
+    Given a partner and user instance to validate whether it is granted
+    to access catalog or not
+    """
+    lms_client = LMSAPIClient(partner)
+    api_access_response = lms_client.get_api_access_request(user)
+    return api_access_response
```
azanbinzahid, line 176: what kind of response it will give?

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -94,6 +95,12 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin, unuse
         catalog = self.get_object()
 
         queryset = catalog.courses()
+        catalog_api_access_response = check_catalog_api_access(request.site.partner, request.user)
+        # exclude 2u products for partners other than edX
+        if catalog_api_access_response and catalog_api_access_response.get('company_name'):
```
azanbinzahid: can you explain in words how this condition becoming true will decide we should exclude 2U products?

course_discovery/apps/api/utils.py
uzairr: it will return the user data related to the catalog api access
`{
                "id": 1,
                "created": "2017-09-25T08:37:05.872566Z",
                "modified": "2017-09-25T08:37:47.412496Z",
                "user": 5,
                "status": "approved",
                "website": "https://example.com/",
                "reason": "Example Reason",
                "company_name": "Example Inc",
                "company_address": "Example Address",
                "site": 1,
                "contacted": True
            }`

course_discovery/apps/api/v1/views/catalogs.py
uzairr: if the requesting user has `catalog api access` then its `status` attribute can be utilized to filter the product lines.

Note: Just noticed that using `status` field is more appropriate than any other field(`company_name`). Updating this check

course_discovery/apps/api/utils.py
DawoudSheraz: We can add the sample response in `get_api_access_request` if not present already.

course_discovery/apps/api/v1/views/affiliates.py
```diff
@@ -35,6 +36,12 @@ def retrieve(self, request, pk=None):
             raise PermissionDenied
 
         courses = catalog.courses()
+        catalog_api_access_response = check_catalog_api_access(request.site.partner, request.user)
+        # exclude 2u products for affiliates
```
DawoudSheraz: Some explanation/comments here on the approach will be helpful in the long run

course_discovery/apps/api/v1/views/catalogs.py
```diff
@@ -94,6 +95,12 @@ def courses(self, request, id=None):  # pylint: disable=redefined-builtin, unuse
         catalog = self.get_object()
 
         queryset = catalog.courses()
+        catalog_api_access_response = check_catalog_api_access(request.site.partner, request.user)
+        # exclude 2u products for partners other than edX
```
DawoudSheraz: same comment here.

DawoudSheraz (approved):

azanbinzahid (approved):

Ali-D-Akbar (approved):

PR #3463, opened 2022-06-07
feat: update max_workers for course data loader

zainab-amir:
Ticket: https://2u-internal.atlassian.net/browse/VAN-941


cdeery (approved):
Excellent. This should damp down the Gateway timeouts.

PR #3465, opened 2022-06-08
feat!: remove uniqueness on program title field

DawoudSheraz:
### Description

This PR removes the uniqueness requirement on the Program model's title field. The presence of this constraint hinders the curation of Programs and Degrees with the same title from the same/different organizations but with different content. It is not clear why this check was introduced in the first place. One potential justification could be that the field "marketing_slug" is auto-generated from the Title field. But marketing_slug is manually entered and has its own uniqueness check in place. The slug needs to be unique as it appears in the marketing pages' url. But it is not generated from Title, so it is just a hindrance to creating different programs with the same title.

### Testing

- Create a program from Disco admin
- Create another program with the same title. It will be saved once the migration is applied.


azanbinzahid (approved):
We need to communicate this, that from now on programs/degrees' factories and tests should use slug as key. I will update Degree Loader as this get's merged. All logs will now use slugs as identifier.

azanbinzahid:
Do you think [this](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L2328) needs to be updated as well?

AliAdnanSohail (approved):

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -2070,6 +2070,70 @@ def test_credit_value(self):
...
+        program_2 = factories.ProgramFactory(
+            title='Test Program',
+            marketing_slug='test-program-2',
+            overview='This is Test Program with marketing slug test-program-2',
+            weeks_to_complete=15
+        )
+
+        assert Program.objects.filter(title='Test Program').count() == 2
+        assert program_1.title == program_2.title
+        assert program_1.overview != program_2.overview
```
Ali-D-Akbar, line 2092: Seems like unnecessary checks to me.

Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: Yes, sort of. Just for a little distinction two programs of the same title with different content are 2 separate entities.

DawoudSheraz:
> Do you think [this](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L2328) needs to be updated as well?

It can be updated, yes.

PR #3469, opened 2022-06-09
feat: use blurb from GEAG api response

Ali-D-Akbar:
[PROD-2815](https://2u-internal.atlassian.net/browse/PROD-2815)

The current EE data loader didn't have any short description field and used to add title as short description. 
This PR uses `blurb` field to be mapped onto short description which is now being exposed through getsmarter GEAG api. 

Note: Currently, there's no validation being performed on the response of GEAG api so if there isn't any `blurb` field, the title will be used as the fallback `short_description`

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
+++ b/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
@@ -297,7 +297,7 @@ class Command(BaseCommand):
             'title': partially_filled_csv_dict.get('title') or product_dict['altName'] or product_dict['name'],
             '2u_title': product_dict['name'],
             'edx_title': product_dict['altName'],
-            'short_description': partially_filled_csv_dict.get('title') or product_dict['name'],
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -297,7 +297,7 @@ def get_transformed_data(self, partially_filled_csv_dict, product_dict):
             'title': partially_filled_csv_dict.get('title') or product_dict['altName'] or product_dict['name'],
             '2u_title': product_dict['name'],
             'edx_title': product_dict['altName'],
-            'short_description': partially_filled_csv_dict.get('title') or product_dict['name'],
+            'short_description': product_dict.get('blurb') or partially_filled_csv_dict.get('title'),
```
DawoudSheraz: For the fallback, we should use product_dict['name'] because partially filled CSV is not used anymore.

DawoudSheraz (approved):

AliAdnanSohail (approved):

uzairr (approved):

PR #3470, opened 2022-06-09
feat: add location models and fields

bseverino:
https://2u-internal.atlassian.net/browse/WS-2986

- Country and state fields added to Course and Program
- Location restriction models: for course or program, add list of countries and/or states, with the ability to specify allowlist or blocklist


course_discovery/apps/course_metadata/models.py
```diff
@@ -922,6 +955,10 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         validators=[FileExtensionValidator(['png'])]
     )
 
+    location_restriction = models.ForeignKey(
+        CourseLocationRestriction, models.SET_NULL, related_name='courses', default=None, null=True, blank=True
+    )
```
bseverino, line 964: I'm not a huge fan of adding this as a foreign key to `Course`, but we can't add `course` as a one-to-one field on `CourseLocationRestriction` as there are technically two versions of each course (draft and non-draft). This makes it a little clunky to update in admin as there is a reverse relationship.

noisecapella (approved):

zzzarius (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -834,6 +839,32 @@ def __str__(self):
         return f'{self.name}'
 
 
+class LocationRestrictionModelMixin(TimeStampedModel):
```
DawoudSheraz: nit: Instead of Mixin, please use AbstractLocationRestrictionModel or BaseLocationRestrictionModel. The working of both is nearly the same but the semantics are different.

course_discovery/apps/course_metadata/models.py
```diff
@@ -834,6 +839,32 @@ def __str__(self):
...
+
+    class Meta:
+        abstract = True
+
+
+class CourseLocationRestriction(LocationRestrictionModelMixin):
+    """
+    Course location restriction model.
+
+    We have to set this as a foreign key field on course rather than a one to one relationship, due to draft courses.
```
DawoudSheraz, line 864: This comment would make more sense in Course model when defining the relationship

course_discovery/apps/course_metadata/models.py
```diff
@@ -863,6 +894,8 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
     authoring_organizations = SortedManyToManyField(Organization, blank=True, related_name='authored_courses')
     sponsoring_organizations = SortedManyToManyField(Organization, blank=True, related_name='sponsored_courses')
     collaborators = SortedManyToManyField(Collaborator, null=True, blank=True, related_name='courses_collaborated')
+    country = CountryField(null=True, blank=True)
+    state = USStateField(null=True, blank=True)
```
DawoudSheraz, line 898: How are these fields different from the linked model? Does this imply the origin of the product?

DawoudSheraz: Plus, why can't they be made part of CourseLocationRestriction model?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -744,3 +745,35 @@ class CollaboratorAdmin(admin.ModelAdmin):
...
+    inlines = (CourseInline,)
+
+
+@admin.register(ProgramLocationRestriction)
+class ProgramLocationRestrictionAdmin(admin.ModelAdmin):
+    list_display = ('program', 'restriction_type',)
+    fields = ('program', 'restriction_type', 'countries', 'states', 'created', 'modified')
+    readonly_fields = ('created', 'modified')
+    raw_id_fields = ('program',)
+    search_fields = ('program__name', 'program__marketing_slug')
```
DawoudSheraz: Out of curiosity, where isn't there any ProgramRestrictionInline for program admin?

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -125,6 +126,18 @@ def product_marketing_url(self):
...
+    def product_country(self):
+        return self.get_country_code()
+
+    @property
+    def product_state(self):
+        return self.state
+
+    @property
+    def product_location_restriction(self):
+        return self.location_restriction
```
DawoudSheraz: location_restriction is an object here, right? How will its representation be determined there?

DawoudSheraz (commented):
- Added some comments
- Will these fields become part of Publisher? Or are they expected to be utilized via admin only?
- How will these impact external products?

course_discovery/apps/course_metadata/models.py
bseverino: Yes, these fields represent the origin of the product and are unrelated to location restriction - there are plans to utilize these in search (i.e. rank products based on the user's home country/state)

course_discovery/apps/course_metadata/admin.py
bseverino: I didn't think to add it, but that would be helpful so I will!

course_discovery/apps/course_metadata/algolia_models.py
bseverino: IIRC it resolved correctly to a JSON object, but let me test this manually again to make sure

bseverino:
@DawoudSheraz 

> Will these fields become part of Publisher? Or are they expected to be utilized via admin only?

We have tickets open to populate this data for existing products, as well as to add the fields to Publisher.

> How will these impact external products?

By external products do you mean EE/Boot Camps/etc?

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Alright. This means there is an expectation that certain courses will have values for origin fields only and no associated model for restriction, right?

DawoudSheraz:
> @DawoudSheraz
> 
> > Will these fields become part of Publisher? Or are they expected to be utilized via admin only?
> 
> We have tickets open to populate this data for existing products, as well as to add the fields to Publisher.
> 
> > How will these impact external products?
> 
> By external products do you mean EE/Boot Camps/etc?

`By external products do you mean EE/Boot Camps/etc?` Yes.

bseverino:
@DawoudSheraz Those products should be able to use these fields as well.

course_discovery/apps/course_metadata/models.py
bseverino: @DawoudSheraz That's right - after some discussion though, we decided this was confusing and I've opted to leave these fields out of the PR and to implement them at a future date.

bseverino:
Because of the changes in permission, though, I no longer have access to push to this branch. I created a new PR here (basically identical, but with country/state fields removed): https://github.com/openedx/course-discovery/pull/3500

PR #3476, opened 2022-06-13
chore: use GEAG API as source of org short code overrides

DawoudSheraz:
### Description

- Map altUniversityAbbreviation field to "Organization Short Code Override" column in EE data transformation and population management command

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
+++ b/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
@@ -26,7 +26,7 @@ class Command(BaseCommand):
 
     # The list to define order of the header keys in csv.
     OUTPUT_CSV_HEADERS = [
-        'organization', '2u_organization_code', 'edx_organization_code', 'title', '2u_title', 'edx_title', 'number',
+        '2u_organization_code', 'title', '2u_title', 'edx_title', 'number',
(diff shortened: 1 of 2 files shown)


azanbinzahid (approved):

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -267,8 +268,7 @@ def get_transformed_data(self, partially_filled_csv_dict, product_dict):
 
         return {
             **default_values,
-            'organization': product_dict['altUniversityAbbreviation'] or product_dict['universityAbbreviation'],
-            'edx_organization_code': product_dict['altUniversityAbbreviation'],
+            'organization_short_code_override': product_dict['altUniversityAbbreviation'],
```
Ali-D-Akbar, line 271: Can we use `get()` instead in case the `altUniversityAbbreviation` is None? or are we just assuming the value will always be there?

Ali-D-Akbar (approved):
One question before merge.

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
DawoudSheraz: Yes, it will be there. If we get into a lot of cases where the data is not populated in GEAG, it can be changed to `get`. But this piece of information is somewhat required now.

PR #3479, opened 2022-06-14
feat: Add default value for status in Program Model and remove hardcoded value from degree loader

AliAdnanSohail:
[remove hard-coded status in degree loader](https://2u-internal.atlassian.net/browse/PROD-2853)

## Description
In degree CSV loader, each time when degree data was loaded by using CSV loader it was assigning `Unpublished` status to Degree, which in the time of creation is acceptable but in the time of update this value should not b hardcoded. So now we have added a default value in Database, due to which when a new degree will be created then status of that Degree will be by default Unpublished. And when Degree is updating it will not effect the status


course_discovery/apps/course_metadata/models.py
```diff
@@ -2246,7 +2246,7 @@ class Program(PkSearchableMixin, TimeStampedModel):
     type = models.ForeignKey(ProgramType, models.CASCADE, null=True, blank=True)
     status = models.CharField(
         help_text=_('The lifecycle status of this Program.'), max_length=24, null=False, blank=False, db_index=True,
-        choices=ProgramStatus.choices, validators=[ProgramStatus.validator]
+        choices=ProgramStatus.choices, validators=[ProgramStatus.validator], default=ProgramStatus.Unpublished
```
DawoudSheraz, line 2261: to verify this addition in an existing unit test or a new unit test in test_models

DawoudSheraz (approved):
- Rebase before the merge
- 1-2 more reviews internally

azanbinzahid (approved):

uzairr (approved):

PR #3489, opened 2022-06-20
fix: prevent making additional metadata object for non 2U courses

azanbinzahid:
### Description:
This PR adds a check to create additional metadata object only if the course type is of 2U. 

### Linked Ticket:
[PROD-2855](https://2u-internal.atlassian.net/browse/PROD-2855)

### Testing:
1. Run unit tests
2. Send additional metadata via api where courses type is not 2U, it shouldn't create additional metadata object in db. It will just ignore.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1260,7 +1260,11 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     def update(self, instance, validated_data):
         # Handle writing nested additional_metadata separately
         if 'additional_metadata' in validated_data:
-            self.update_additional_metadata(instance, validated_data.pop('additional_metadata'))
+            # Handle additional metadata only for 2U courses else just pop
+            additional_metadata_data = validated_data.pop('additional_metadata')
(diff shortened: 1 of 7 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1260,7 +1260,11 @@ def update_additional_metadata(self, instance, additional_metadata):
     def update(self, instance, validated_data):
         # Handle writing nested additional_metadata separately
         if 'additional_metadata' in validated_data:
-            self.update_additional_metadata(instance, validated_data.pop('additional_metadata'))
+            # Handle additional metadata only for 2U courses else just pop
+            additional_metadata_data = validated_data.pop('additional_metadata')
```
Ali-D-Akbar, line 1264: nit: unnecessary variable. Can be used directly.

DawoudSheraz (approved):

Ali-D-Akbar (approved):
nit but not a blocker. 👍

course_discovery/apps/api/serializers.py
azanbinzahid: using it directly doesn't pop it from validated data in case next condition is not met.

PR #3493, opened 2022-06-23
fix: course run key was overflowing after 'z'

AliAdnanSohail:
[Investigate and fix why the bootcamp ingestion for a set of courses is failing on production](https://2u-internal.atlassian.net/browse/PROD-2869)

## Issue:
We are increasing Course Run if course run already exist for any course_num.

For example if we have course run = `course-v1:TestX+Testing101x+1T2017`
the new course run will `course-v1:TestX+Testing101x+1T2017a` 
append `a` at the last. And if we have `course-v1:TestX+Testing101x+1T2017a` then it will replace `a` with `b` to make it unique like `course-v1:TestX+Testing101x+1T2017b`, So each time course run is increasing it will change alphabet. And we are not handling boundary case, like if we get `z` at the end then after increment it was replaced with `{` which is not acceptable.

## Fix:
We need to update logic to handle boundary case that if we get `z` at the last then new course run will `course-v1:TestX+Testing101x+1T2017aa`

### DIFF
--- a/course_discovery/apps/api/tests/test_utils.py
+++ b/course_discovery/apps/api/tests/test_utils.py
@@ -12,7 +12,9 @@ from rest_framework.request import Request
 from rest_framework.test import APIRequestFactory
 from rest_framework.views import APIView
 
-from course_discovery.apps.api.utils import StudioAPI, cast2int, decode_image_data, get_query_param
+from course_discovery.apps.api.utils import (
+    StudioAPI, cast2int, decode_image_data, get_query_param, increment_character, increment_str
+)
 from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase, OAuth2Mixin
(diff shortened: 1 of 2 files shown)


DawoudSheraz (approved):
This can be a potential breaking change. Please see https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html#type on how to show that in the commit. Other than that, good to go. Locally tested on 100 courses.

connorhaugh (approved):
Looks good and well tested to me. Should not effect existing courses, which is good. if you could:
- Make sure this gets into the Olive release notes.
- Notify TNL and Partner support when you do merge so we can detect any impacts to partners early. 
That would be excellent. Thanks so much!

azanbinzahid (approved):
Logic-wise, it works correct. Just make sure that our platform does not have limitation for one word suffix just in case.

PR #3494, opened 2022-06-24
fix: update existing facts instead of deleting and creating new

azanbinzahid:
### Description
This PR improves the logic for handling facts update.

### Linked Ticket
[PROD-2856](https://2u-internal.atlassian.net/browse/PROD-2856)

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1226,10 +1226,24 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
 
     def update_facts(self, instance, facts_data):
 
-        instance.facts.all().delete()
-        for fact in facts_data:
-            fact_obj = Fact.objects.create(**fact)
-            instance.facts.add(fact_obj)
+        existing_facts = instance.facts.all()
+        if len(existing_facts) == len(facts_data):
+            for i, fact in enumerate(existing_facts):
+                Fact.objects.filter(id=fact.id).update(**facts_data[i])
+        else:
+            instance.facts.clear()
+            for fact in facts_data:
+                # we can query all orphan facts and use one of them but that is more computation
+                # therefore, just checking if a fact with same data exists
+                facts_queryset = Fact.objects.filter(**fact)
+                fact_obj = facts_queryset.first() if facts_queryset.exists() else None
+
+                # create a new fact object if (a) it does not exist or (b) it exists but it is not orphan
+                # related object count > 0 means it is not orphan and updating it will cause an unwanted update
+                if not fact_obj or fact_obj.related_course_additional_metadata.count():
+                    fact_obj = Fact.objects.create(**fact)
+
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1158,6 +1158,95 @@ def test_no_update_with_additional_metadata_if_type_not_2U(self, type_slug):
...
+        url_2 = reverse('api:v1:course-detail', kwargs={'key': course_2.uuid})
+        url_3 = reverse('api:v1:course-detail', kwargs={'key': course_3.uuid})
+
+        response_1 = self.client.patch(url_1, {'additional_metadata': additional_metadata_1}, format='json')
+        assert response_1.status_code == 200
+        assert Fact.objects.count() == 2    # created two new objects
+
+        response_2 = self.client.patch(url_2, {'additional_metadata': additional_metadata_2}, format='json')
+        assert response_2.status_code == 200
+        assert Fact.objects.count() == 3    # created 1 new object even fact wih same data exisisted
```
DawoudSheraz: nit: existed

course_discovery/apps/api/serializers.py
```diff
@@ -1226,10 +1226,25 @@ def create(self, validated_data):
...
+            for fact in facts_data:
+                # we can query all orphan facts and use one of them but that is more computation
+                # therefore, just checking if a fact with same data exists
+                facts_queryset = Fact.objects.filter(**fact)
+                fact_obj = facts_queryset.first() if facts_queryset.exists() else None
+
+                # only make relation with the orphan fact object else create a new one
+                # updating a fact (as in above if) attached to more than one course may
+                # result in unwanted update
+                if not fact_obj or fact_obj.related_course_additional_metadata.count():
```
DawoudSheraz, line 1243: Hmmm, why is there no `not` before `fact_obj.related_course_additional_metadata.count()`? Because if additional metadata is present, we are opting for a new fact object creation. I am a bit confused on the behavior here.

azanbinzahid: Yes, if the count > 0 that means the `fact_obj` that we fetched from the db has some other association (read: not orphan). I am suggesting not to use it and create a new one instead to maintain 1-1 relation. Otherwise, we might update (read: overwrite) it through some other course update call (which we won't want).

azanbinzahid: Your confusion is right, this is a modification to the proposed approach in ticket.

DawoudSheraz: Ok, we can update the in-code comment to be more descriptive, like the above comment.

azanbinzahid: Updated

DawoudSheraz (approved):
Once the build is green.

Ali-D-Akbar (approved):

PR #3498, opened 2022-06-27
feat: use config model for csv file and update course import metadata

Ali-D-Akbar:
Using a config model instead of using the csv file path in `import_course_metadata` command. A file can be uploaded on the config model through django admin which will be used in the current version of the `import_course_metadata` command. We want to be able to run this command on Jenkins.

The reason behind changing this approach is to avoid uploading file on discovery server through ansible (for which, an additional amount of work was required to create a seedjob in Jenkins) so using this approach, the file is accessible from the server and the management command is run along with other commands using a single seed job. We will be adding this command in our stage/prod jenkins in the following PR: https://github.com/edx/edx-internal/pull/6671

Codewise, this PR does the following:
1. Adds `django-config-models` and run upgrade
2. Adds new config model which is also available in admin
3. Updates Unit Tests for our new approach
4. Fixes pylint changes due to the recent requirement upgrade


course_discovery/apps/course_metadata/management/commands/import_course_metadata.py
```diff
@@ -26,19 +28,13 @@ def add_arguments(self, parser):
             default='edx',
             type=str,
         )
-        parser.add_argument(
-            '--csv_path',
-            help='Path to the CSV file',
-            type=str,
-            required=True
-        )
```
DawoudSheraz, line 36: Why was this removed? Didn't we discuss internally to keep this argument but make it optional? Or did you get into any issues with that?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3137,9 +3137,9 @@
     'course_pacing': 'self-paced',
     'course_run_enrollment_track': 'Paid Executive Education',
     'staff': 'staff_1,staff_2',
-    'minimum_effort': 4,
-    'maximum_effort': 10,
-    'length': 10,
+    'minimum_effort': '4',
+    'maximum_effort': '10',
+    'length': '10',
```
DawoudSheraz, line 3142: why is the change from int -> str required?

course_discovery/apps/course_metadata/management/commands/tests/test_import_course_metadata.py
```diff
@@ -31,6 +32,14 @@ def setUp(self) -> None:
         self.mock_access_token()
         self.user = UserFactory.create(username="test_user", password=USER_PASSWORD, is_staff=True)
         self.client.login(username=self.user.username, password=USER_PASSWORD)
+        csv_file_content = ','.join(list(mock_data.VALID_COURSE_AND_COURSE_RUN_CSV_DICT)) + '\n'
+        csv_file_content += ','.join(f'"{key}"' for key in list(
```
DawoudSheraz, line 36: why can't the content be read from the temp file? Why re-write the logic?

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -353,13 +353,13 @@ def test_ensure_draft_world_not_draft_course_run_given(self):
...
-            assert draft_seats[i].course_run.uuid == not_draft_seats[i].course_run.uuid
-            assert draft_seats[i].official_version == not_draft_seats[i]
-            assert not_draft_seats[i].draft_version == draft_seats[i]
+        for i, draft_seat in enumerate(draft_seats):
+            assert draft_seat.price == not_draft_seats[i].price
+            assert draft_seat.sku == not_draft_seats[i].sku
+            assert draft_seat.course_run != not_draft_seats[i].course_run
+            assert draft_seat.course_run.uuid == not_draft_seats[i].course_run.uuid
+            assert draft_seat.official_version == not_draft_seats[i]
+            assert not_draft_seats[i].draft_version == draft_seat
```
DawoudSheraz, line 362: 😄

course_discovery/apps/course_metadata/models.py
```diff
@@ -3230,6 +3231,17 @@ class Meta:
...
+class CSVDataLoaderConfiguration(ConfigurationModel):
+    """
+    Configuration to store a csv file that will be used in import_course_metadata.
+    """
+    csv_file = models.FileField(
+        validators=[FileExtensionValidator(allowed_extensions=['csv'])],
+        help_text=_("It expects the data will be provided in a csv file format "
+                    "with first row containing all the headers.")
+    )
+
```
DawoudSheraz, line 3285: Are other fields auto-generating from ConfigurationModel?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
Ali-D-Akbar: In order to write this in csv file with comma separated using `','.join()`

course_discovery/apps/course_metadata/management/commands/tests/test_import_course_metadata.py
Ali-D-Akbar: I already discussed the issue it causes with the temp file. `SimpleUploadedFile` can be used to upload file in `FIleFIeld`.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: yes, `change_date`, `changed_by` are already in there.

course_discovery/apps/course_metadata/management/commands/import_course_metadata.py
Ali-D-Akbar: Yes we did. I think keeping both of these approaches will be redundant. The config model is a more clean way rather than copying the file and giving a path to the command.

Open to suggestions.

DawoudSheraz: @azanbinzahid What are your thoughts? On local, the ability to provide files via cmd line argument will make it easier to test and ingest data. Two options are better than 1.

course_discovery/apps/course_metadata/management/commands/import_course_metadata.py
```diff
@@ -30,15 +32,15 @@ def add_arguments(self, parser):
...
         )
 
     def handle(self, *args, **options):
         """
         Example usage: ./manage.py import_course_metadata --partner_code=edx --csv_path=test.csv
         """
         partner_short_code = options.get('partner_code')
-        csv_path = options.get('csv_path')
+        csv_path = options.get('csv_path') if options.get('csv_path') \
+            else CSVDataLoaderConfiguration.current().csv_file.path
```
azanbinzahid: error: working for disabled/deleted objects as well

course_discovery/apps/course_metadata/management/commands/tests/test_import_course_metadata.py
DawoudSheraz: Please add discussion details here for visibility.

course_discovery/apps/course_metadata/models.py
```diff
@@ -3230,6 +3231,18 @@ class Meta:
         )
 
 
+class CSVDataLoaderConfiguration(ConfigurationModel):
+    """
+    Configuration to store a csv file that will be used in import_course_metadata.
+    """
+    cache_timeout = 0
```
DawoudSheraz, line 3279: Add a comment to explain why this is needed.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -766,7 +766,7 @@ class JobOutlookItemAdmin(admin.ModelAdmin):
               DeletePersonDupsConfig, DrupalPublishUuidConfig, MigrateCommentsToSalesforce,
               MigratePublisherToCourseMetadataConfig, ProfileImageDownloadConfig, PersonAreaOfExpertise,
               TagCourseUuidsConfig, BackpopulateCourseTypeConfig, RemoveRedirectsConfig, BulkModifyProgramHookConfig,
-              BackfillCourseRunSlugsConfig, BulkUpdateImagesConfig):
+              BackfillCourseRunSlugsConfig, BulkUpdateImagesConfig, CSVDataLoaderConfiguration):
```
DawoudSheraz: Please create a separate model admin for this and add fields for list_display.

course_discovery/apps/course_metadata/models.py
```diff
@@ -3230,6 +3231,18 @@ class Meta:
         )
 
 
+class CSVDataLoaderConfiguration(ConfigurationModel):
```
DawoudSheraz, line 3274: Make sure this model is skipped in test_signals.py file https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/tests/test_signals.py#L48

DawoudSheraz (approved):
Left a few small comments to address before merge

course_discovery/apps/course_metadata/management/commands/tests/test_import_course_metadata.py
Ali-D-Akbar: I got an error something like `AttributeError: '_io.BufferedReader' object has no attribute` which was caused by temp file. I'm using `SimpleUploadedFile` which is compatible with FileField unit testing (inspiration taken from here: https://github.com/openedx/edx-platform/blob/master/common/djangoapps/student/management/tests/test_bulk_unenroll.py#L106)

I had to manually write json in csv as there is no proper or generic method to do that. Once the data was converted, I wrote it in file as a binary string.

azanbinzahid (approved):

PR #3500, opened 2022-06-27
feat: add location restriction models

bseverino:
Re-opening https://github.com/openedx/course-discovery/pull/3470 with a fork due to permissions changes.

https://2u-internal.atlassian.net/browse/WS-2986

- Location restriction models: for course or program, add list of countries and/or states, with the ability to specify allowlist or blocklist
- `location_restriction` field added to both REST API and Algolia models for Course and Program

Example response:
```
"location_restriction": {
  "restriction_type": "allowlist",
  "countries": ["US", "CA"],
  "states": []
}
```

Differences from previous PR:
- Removed home country/state fields as they were confusing and unnecessary for the functionality introduced here


noisecapella (approved):

course_discovery/apps/course_metadata/admin.py
```diff
@@ -283,6 +283,38 @@ def save_model(self, request, obj, form, change):
             messages.add_message(request, messages.ERROR, msg)
 
 
+class CourseInline(admin.TabularInline):
```
srwang, line 286: What is this class for?

srwang (approved):

course_discovery/apps/course_metadata/admin.py
bseverino: This is just a convenience to add course info to the admin page for `CourseLocationRestriction`:

Screen Shot 2022-06-27 at 8 24 39 PM

azanbinzahid: I guess this is view only. We can't add courses from here.

course_discovery/apps/course_metadata/models.py
```diff
@@ -834,6 +837,32 @@ def __str__(self):
...
 
 
+class AbstractLocationRestrictionModel(TimeStampedModel):
+    ALLOWLIST = 'allowlist'
+    BLOCKLIST = 'blocklist'
+    RESTRICTION_TYPE_CHOICES = (
+        (ALLOWLIST, _('Allowlist')),
+        (BLOCKLIST, _('Blocklist'))
+    )
+
```
azanbinzahid, line 847: [Not entirely related] We could add a field to define name/presets like `Baltic Countries`, `South East Asia` if we choose not to go with OneToOne relation.

azanbinzahid (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -1109,6 +1112,30 @@ def create(self, validated_data):
         return course_editor
 
 
+class AbstractLocationRestrictionSerializer(BaseModelSerializer):
+    restriction_type = serializers.ChoiceField(choices=AbstractLocationRestrictionModel.RESTRICTION_TYPE_CHOICES)
```
jesseday, line 1116: @bseverino can we change this structure for our algolia query.  It will work better for our filtering requirements if we can use something like the following.

Note that if we want to block a state we have this format of `Country-State`.  Does that seem reasonable?

```
allowed_in: ['US', 'IT', 'LT', 'DK'],
blocked_in: ['SP', 'RU', 'US-TX']
```

PR #3501, opened 2022-06-27
fix: use list filter in degree admin search

azanbinzahid:
### Description:
This PR fixes 500 - Internal server error while searching degree via django admin

### Linked Ticket:
[PROD-2875](https://2u-internal.atlassian.net/browse/PROD-2875)

### Testing:
1. Open admin and search degree on master branch [expected: error]
2. Checkout this branch and search again [expected: display results]

### TODO:
- [x] Discuss and add more fields for searching  
- [x] Alternate approach like `partner__name` in search list

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -716,7 +716,8 @@ class DegreeAdmin(admin.ModelAdmin):
     list_display = ('title', 'partner', 'status', 'hidden')
     ordering = ('title', 'status')
     readonly_fields = ('uuid', )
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/admin.py
```diff
@@ -716,7 +716,8 @@ class DegreeAdmin(admin.ModelAdmin):
     list_display = ('title', 'partner', 'status', 'hidden')
```
DawoudSheraz: suggestion: while we are at it, we can update the list display. Partner won't be changing much, so we can remove it and add uuid and slug.

DawoudSheraz (approved):

PR #3505, opened 2022-06-30
Run semgrep ourselves, not from an action

nedbat:
The action doesn't add any value, and pins semgrep to an old version.


nedbat:
This is superceded by #3507 .

PR #3508, opened 2022-07-01
fix: return filterable value for empty location restriction lists

bseverino:
https://2u-internal.atlassian.net/browse/WS-2861

Algolia can't filter on empty lists. For `allowed_in` and `blocked_in`, if the value is an empty list, index it as `['null']`. (the string can be any value here as long as it's consistent)

### DIFF
--- a/course_discovery/apps/course_metadata/algolia_models.py
+++ b/course_discovery/apps/course_metadata/algolia_models.py
@@ -12,6 +12,9 @@ from course_discovery.apps.course_metadata.models import (
     AbstractLocationRestrictionModel, Course, CourseType, Program, ProgramType
 )
 
+# Algolia can't filter on an empty list, provide a value we can still filter on
+EMPTY_LOCATION_RESTRICTION_LIST = ['null']
+
 
 # Utility methods used by both courses and programs
 def get_active_language_tag(course):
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -12,6 +12,8 @@
     AbstractLocationRestrictionModel, Course, CourseType, Program, ProgramType
 )
 
+EMPTY_LIST = ['null']
```
kalebjdavenport: Do you think there's a good place you can let other devs know that they should return this value if there are no matches in general or do you think having this constant here is enough documentation in itself? This might be enough! I'm just not as familiar with this section of the code so I thought I'd ask.

kalebjdavenport (approved):

course_discovery/apps/course_metadata/algolia_models.py
bseverino: I'll change this variable name so that it's clear it's just meant for location restriction (at least for now); I'm not sure where else we would document this, as noted below it's just due to restrictions with Algolia filtering

aheykamp2u (approved):

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -90,6 +92,11 @@ def get_course_availability(course):
 
 
 def get_location_restriction(location_restriction):
+    if (len(location_restriction.countries) == 0 and len(location_restriction.states) == 0):
+        # Algolia can't filter on an empty list, so in the case that there are no values for some reason,
+        # provide a value we can still filter on
```
DawoudSheraz: This comment will make better sense where the variable is defined. This variable has been returned on multiple locations in the code.

DawoudSheraz (approved):
A small nit to address before merge.

PR #3512, opened 2022-07-04
fix: use unicodecsv to read s3 csv file

Ali-D-Akbar:
Fixes the issue reading file from s3. Instead of reading the absolute path, we need to read the file directly through `unicodecsv`. 

### Common Questions:

- Did I try reading the file through `csv.DictReader(csv_file)`? Yes, but it doesn't work as it reads data through strings not bytes
- Did I try using `unicodecsv.DictReader(open(csv_path, 'r'))`? yes, it still doesn't work. Says `'str' object has no attribute 'decode'`

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -5,6 +5,7 @@ creating and updating related objects in Studio, and ecommerce, provided a csv c
 import csv
 import logging
 
+import unicodecsv
 from django.conf import settings
 from django.db.models import Q
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/management/commands/import_course_metadata.py
```diff
@@ -40,8 +40,8 @@ def handle(self, *args, **options):
         """
         partner_short_code = options.get('partner_code')
         csv_loader_config = CSVDataLoaderConfiguration.current()
-        csv_path = options.get('csv_path') if options.get('csv_path') \
-            else csv_loader_config.csv_file.path if csv_loader_config.is_enabled() else ''
+        csv_path = options.get('csv_path') if options.get('csv_path') else None
```
uzairr: else clause may be avoided: `options.get('csv_path', None)`

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -53,13 +54,13 @@ class CSVDataLoader(AbstractDataLoader):
...
 
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None):
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None, csv_file=None):
         super().__init__(partner, api_url, max_workers, is_threadsafe)
 
         self.messages_list = []  # to show failure/skipped ingestion message at the end
         self.course_uuids = {}  # to show the discovery course ids for each processed course
         try:
-            self.reader = csv.DictReader(open(csv_path, 'r'))  # lint-amnesty, pylint: disable=consider-using-with
+            self.reader = csv.DictReader(open(csv_path, 'r')) if csv_path else list(unicodecsv.DictReader(csv_file))  # lint-amnesty, pylint: disable=consider-using-with
```
uzairr: it takes too much attention to grab whats going on here!
`
if csv_path:
    serlf.reader = csv.DictReader(open(csv_path, 'r'))
else:
    self.reader = list(unicodecsv.DictReader(csv_file) if csv_file
`

DawoudSheraz: what happens if the list is not used? Does it not return an iterable object?

uzairr (approved):
few nits otherwise, LGTM

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
Ali-D-Akbar: Yes, it wasn't iterable. I think the data is in bytes-like structure before that. (Haven't tested it)

Ali-D-Akbar: Its the same thing but I will improve the readability here a bit.

azanbinzahid (approved):

PR #3516, opened 2022-07-05
feat: download and save svg image from url in csv data loader

AliAdnanSohail:
[Use logos provided by the Exec Ed team in the GEAG API](https://2u-internal.atlassian.net/browse/PROD-2818)

# Description:
For organization logo provided by the Exec Ed team in GEAG API we need to convert SVG logos to PNG. Because our database doesn't accept SVG file. So we are using [CairoSVG](https://pypi.org/project/CairoSVG/) for image conversion.

# Test:
To test it locally, update CSV organization_logo_url column to some svg image e.g https://www.getsmarter.com/disk/public/zn6BcUyaz77c8rzsSHLgjW9o/university_of_cape_town_sitewide_logo.svg

And run the `import_course_metadata` command for ingestion. It will convert that image into PNG and save in database for organization_logo_override


course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -152,15 +154,9 @@ def ingest(self):  # pylint: disable=too-many-statements
 
             if row.get('organization_logo_override'):
                 course.refresh_from_db()
-                is_logo_downloaded = download_and_save_course_image(
+                is_logo_downloaded = download_and_save_organization_logo_override_image(
```
DawoudSheraz: we expect PNGs links here for other products (bootcamps, for instance). I would suggest using the existing method but add util to convert SVG -> PNG if the downloaded image is SVG.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -706,6 +707,29 @@ def download_and_save_course_image(course, image_url, data_field='image', header
...
+        extension = 'png'
+        filename = '{uuid}.{extension}'.format(uuid=str(course.uuid), extension=extension)
+        image_file = convert_svg_to_png_from_url(image_url)
+        course.organization_logo_override.save(filename, image_file)
+        logger.info('Image for course [%s] successfully updated.', course.key)
+        return True
+    except Exception:  # pylint: disable=broad-except
+        logger.exception('An unknown exception occurred while downloading image for course [%s]', course.key)
+    return False
+
```
DawoudSheraz: As I have mentioned above, we can utilize the existing download and save method. A check to convert SVG -> PNG if the downloaded image is SVG would do the trick. This will also ensure the override behavior does not change for CSVs where the logo override image is PNG.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -706,6 +707,29 @@ def download_and_save_course_image(course, image_url, data_field='image', header
...
+
+def convert_svg_to_png_from_url(image_url):
+    temp_file = NamedTemporaryFile()
```
DawoudSheraz: Please add appropriate docstrings.

course_discovery/apps/course_metadata/constants.py
```diff
@@ -11,6 +11,7 @@
 IMAGE_TYPES = {
     'image/jpeg': 'jpg',
     'image/png': 'png',
+    'image/svg+xml': 'svg'
```
DawoudSheraz: nit: leave a comment here that svg will be converted into png and not stored as svg.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -705,3 +705,9 @@ def test_skill_data_transformation(self):
 
         output = transform_skills_data(input_data)
         assert output == expected_data
+
+
+class TestConvertSvgToPngFromUrl(TestCase):
+    @mock.patch('course_discovery.apps.course_metadata.utils.svg2png')
+    def test_convert_svg_to_png_from_url(self, _svg2png_mock):
+        assert convert_svg_to_png_from_url('https://www.svgimageurl.com') is not None
```
DawoudSheraz: nit: add new lines to make it more readable.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -717,6 +722,16 @@ def download_and_save_course_image(course, image_url, data_field='image', header
     return False
 
 
+def convert_svg_to_png_from_url(image_url):
+    """
+    Given an image file url of svg, it will convert that svg image to png
+    and save in temporary file
+    """
+    with NamedTemporaryFile() as temp_file:
+        svg2png(url=image_url, write_to=temp_file.name)
```
DawoudSheraz, line 736: nit: wrap this in try/catch, in case something goes wrong during conversion.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -701,7 +702,11 @@ def download_and_save_course_image(course, image_url, data_field='image', header
                 if data_field == 'image':
                     course.image.save(filename, ContentFile(response.content))
                 elif data_field == 'organization_logo_override':
-                    course.organization_logo_override.save(filename, ContentFile(response.content))
+                    image_file = ContentFile(response.content)
+                    if extension == 'svg':
+                        filename = '{uuid}.png'.format(uuid=str(course.uuid))
+                        image_file = convert_svg_to_png_from_url(image_url)
+                    course.organization_logo_override.save(filename, image_file)
```
DawoudSheraz: unit tests for this are missing.

DawoudSheraz (commented):
left a few nits and comments.

course_discovery/apps/course_metadata/utils.py
AliAdnanSohail: Currently we don't have any tests for this method, so we can create a follow up task to test this method with all possible parameters

uzairr: I know a test is added in `TestConvertSvgToPngFromUrl` but i suspect that there must be a use-case around, whose test is part of the existing codebase. Can you please double check, may be we can refactor that test to cover this change as well ?

uzairr:
A suggestion is added that may be ignored if there is future task planned around it. Approving this PR. 👍

uzairr (approved):

azanbinzahid (approved):

course_discovery/apps/course_metadata/utils.py
```diff
@@ -717,6 +722,20 @@ def download_and_save_course_image(course, image_url, data_field='image', header
...
 
+def convert_svg_to_png_from_url(image_url):
+    """
+    Given an image file url of svg, it will convert that svg image to png
+    and save in temporary file
+    """
+    try:
+        with NamedTemporaryFile() as temp_file:
+            svg2png(url=image_url, write_to=temp_file.name)
+            return temp_file
```
DawoudSheraz: The context is switched when the temp file is returned. Due to the context switch, the temp file is closed and removed. The following piece of code worked for me during the testing:
```
temp_file = NamedTemporaryFile()
svg2png(url=image_url, write_to=temp_file.name)
temp_file.seek(0)
return temp_file
```
seek(0) takes the file pointer to the start of temp file. Otherwise, the read starts from where the write pointer was left at.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -701,7 +702,11 @@ def download_and_save_course_image(course, image_url, data_field='image', header
                 if data_field == 'image':
                     course.image.save(filename, ContentFile(response.content))
                 elif data_field == 'organization_logo_override':
-                    course.organization_logo_override.save(filename, ContentFile(response.content))
+                    image_file = ContentFile(response.content)
+                    if extension == 'svg':
+                        filename = '{uuid}.png'.format(uuid=str(course.uuid))
+                        image_file = convert_svg_to_png_from_url(image_url)
```
DawoudSheraz, line 708: nit: Since None can be returned by convert method, a check will be needed here to avoid DB error against None.

DawoudSheraz (approved):
A comment to address before merge.

PR #3517, opened 2022-07-06
feat: use config model for csv file and update import degree metadata

Ali-D-Akbar:
Using a config model instead of using the csv file path in `import_degree_data` command. A file can be uploaded on the config model through django admin which will be used in the current version of the `import_degree_data` command. We want to be able to run this command on Jenkins. We opted to keep the config models separate so that we will be able to run all of the ingestions separately.

We will be adding this command in our stage/prod jenkins in the following PR: https://github.com/edx/edx-internal/pull/6828

### Testing Instructions
1. Run migrations
2. Add an entry against `http://localhost:18381/admin/course_metadata/degreedataloaderconfiguration/` with a sample degree data file
3. Run `./manage.py import_degree_data` in Discovery Shell
4. Expect it to run the same way it does with `./manage.py import_degree_data --csv_path=degree.csv`


course_discovery/apps/course_metadata/models.py
```diff
@@ -3284,6 +3284,19 @@ class CSVDataLoaderConfiguration(ConfigurationModel):
...
+    Configuration to store a csv file that will be used in import_degree_data.
+    """
+    # Timeout set to 0 so that the model does not read from cached config in case the config entry is deleted.
+    cache_timeout = 0
+    csv_file = models.FileField(
+        validators=[FileExtensionValidator(allowed_extensions=['csv'])],
+        help_text=_("It expects the data will be provided in a csv file format "
+                    "with first row containing all the headers.")
+    )
+
```
DawoudSheraz, line 3305: Do we really need a new model? Can't we use CSVloader config model for this too? Seems overkill to create a new model that isn't different from the existing?

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -20,17 +22,18 @@ class DegreeCSVDataLoader(AbstractDataLoader):
...
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None):
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, csv_path=None, csv_file=None):
         super().__init__(partner, api_url, max_workers, is_threadsafe)
 
         self.messages_list = []  # to show failure/skipped ingestion message at the end
         self.degree_uuids = {}  # to show the discovery degrees/program ids for each processed degree
         try:
-            self.reader = csv.DictReader(open(csv_path, 'r'))  # lint-amnesty, pylint: disable=consider-using-with
+            # Read file from the path if given. Otherwise, read from the file received from CSVDataLoaderConfiguration.
+            self.reader = csv.DictReader(open(csv_path, 'r')) if csv_path \
```
DawoudSheraz, line 36: nit: DegreeDataLoaderConfiguration

course_discovery/apps/course_metadata/management/commands/import_degree_data.py
```diff
@@ -10,13 +10,15 @@
...
+from course_discovery.apps.course_metadata.models import DegreeDataLoaderConfiguration
 from course_discovery.apps.course_metadata.signals import connect_api_change_receiver
 
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
-    help = 'Import degree information from a CSV available on a provided csv path.'
+    help = 'Import degree information from a CSV available either through a provided csv file path ' \
+           'or a CSV file uploaded through CSVDataLoaderConfiguration in django admin.'
```
DawoudSheraz: nit: DegreeDataLoaderConfiguration

DawoudSheraz (commented):
- nit comments
- Add info in PR to explain why we opted for a new model instead of using existing CSV loader configuration model

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Using same model as discussed. Added in PR description.

course_discovery/apps/course_metadata/models.py
```diff
@@ -3291,6 +3291,19 @@ class CSVDataLoaderConfiguration(ConfigurationModel):
     )
 
 
+class DegreeDataLoaderConfiguration(ConfigurationModel):
```
DawoudSheraz, line 3294: Remove this model from test_signals file's test, like it was done for CSVloader model

DawoudSheraz (approved):

AliAdnanSohail (approved):
LGTM

PR #3519, opened 2022-07-08
docs: Algolia start date ranking ADR

bseverino:
This ADR arises from a need to de-emphasize Exec Ed courses in search results when they start in more than 45 days. We need to come up with an approach to give Algolia the data it needs not only to accomplish this goal, but potentially to make the functionality extensible.

Internal ticket: https://2u-internal.atlassian.net/browse/WS-3108

### DIFF
--- a/docs/decisions/0013-ranking-in-algolia-by-start-date.rst
+++ b/docs/decisions/0013-ranking-in-algolia-by-start-date.rst
@@ -1,0 +1,39 @@
+Ranking in Algolia by Start Date
+================================
+
+Status
+------
+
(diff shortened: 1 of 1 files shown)


marekwrobel (approved):

marekwrobel (approved):

idegtiarov:
@bseverino could you please add a description to the PR instead of a link to the internal 2U issue which cannot be read outside 2U organization?

kalebjdavenport (approved):
Looks great to me!

DawoudSheraz (approved):

PR #3450, opened 2022-05-31
feat: add new program/degree types

azanbinzahid:
## Description:
This PR adds `bachelors, doctorate, license` ProgramType slug in models and excludes them from affiliate endpoints. 

## Linked Ticket:
[PROD-2804](https://2u-internal.atlassian.net/browse/PROD-2804)

## Checklist:
- [x] fix (any) failings tests
- [x] check if we need to add doctorate program or not
- [x] communicated `Available Now` change with WS team
- [ ] double check what else needs to be updated
- [ ] create admin entries in stage/prod
- [ ] research if any seat related changes are needed or not

PR #3522, opened 2022-07-11
feat: add cascade toggle logic to enterprise sub inclusion boolean

alex-sheehan-edx:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/boards/630?modal=detail&selectedIssue=ENT-5840

When an org subscription inclusion is toggled, all courses, course runs and programs included in the org should check if they should be toggled.

accounts for course/program types as well.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -398,6 +398,12 @@ class ProgramType(TranslatableModel, TimeStampedModel):
             program_type = program_model.objects.get(slug=slug)
         return program_type, name
 
+    @classmethod
+    def is_enterprise_catalog_program_type(cls, program_type):
+        return program_type.slug in [
+            cls.XSERIES, cls.MICROMASTERS, cls.PROFESSIONAL_CERTIFICATE, cls.PROFESSIONAL_PROGRAM_WL, cls.MICROBACHELORS
+        ]
+
 
 class ProgramTypeTranslation(TranslatedFieldsModel):
     master = models.ForeignKey(ProgramType, models.CASCADE, related_name='translations', null=True)
@@ -543,6 +549,12 @@ class CourseType(TimeStampedModel):
     def empty(self):
         """ Empty types are special - they are the default type used when we don't know a real type """
         return self.slug == self.EMPTY
+
+    @classmethod
+    def is_enterprise_catalog_course_type(cls, course_type):
+        return course_type.slug in [
+            cls.AUDIT, cls.VERIFIED_AUDIT, cls.PROFESSIONAL, cls.CREDIT_VERIFIED_AUDIT, cls.EMPTY
+        ]
 
 
 class Subject(TranslatableModel, TimeStampedModel):
@@ -1007,7 +1019,8 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
 
     def _check_enterprise_subscription_inclusion(self):
         # if false has been passed in, or it's already been set to false
-        if self.enterprise_subscription_inclusion is False:
+        if not CourseType.is_enterprise_catalog_course_type(self.type) or \
+                self.enterprise_subscription_inclusion is False:
             return False
         for org in self.authoring_organizations.all():
             if not org.enterprise_subscription_inclusion:
@@ -1023,6 +1036,12 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         kwargs['force_insert'] = False
         kwargs['force_update'] = True
         super().save(*args, **kwargs)
+
+        # Course runs calculate enterprise subscription inclusion based off of their parent's status, so we need to
+        # force a recalculation
+        course_runs = CourseRun.objects.filter(course=self)
+        for course_run in course_runs:
+            course_run.save()
 
     def __str__(self):
         return f'{self.key}: {self.title}'
@@ -2765,8 +2784,10 @@ class Program(PkSearchableMixin, TimeStampedModel):
         return self.status == ProgramStatus.Active
 
     def _check_enterprise_subscription_inclusion(self):
-        if self.type == 'micromasters':
+        # We exclude Bachelors, Masters, and Doctorate programs as the cost per user would be too high
+        if not ProgramType.is_enterprise_catalog_program_type(self.type):
             return False
+
         for course in self.courses.all():
             if not course.enterprise_subscription_inclusion:
                 return False
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -1,6 +1,7 @@
 import logging
 
 import waffle  # lint-amnesty, pylint: disable=invalid-django-waffle-import
+from celery import shared_task
 from django.apps import apps
 from django.core.exceptions import ValidationError
 from django.db.models import Q
@@ -17,6 +18,7 @@ from course_discovery.apps.course_metadata.publishers import ProgramMarketingSit
 from course_discovery.apps.course_metadata.salesforce import (
     populate_official_with_existing_draft, requires_salesforce_update
 )
+from course_discovery.apps.course_metadata.tasks import update_org_program_and_courses_ent_sub_inclusion
 from course_discovery.apps.course_metadata.utils import get_salesforce_util
 
 logger = logging.getLogger(__name__)
@@ -178,6 +180,14 @@ def update_or_create_salesforce_organization(instance, created, **kwargs):
             util.create_publisher_organization(instance)
         if not created and requires_salesforce_update('organization', instance):
             util.update_publisher_organization(instance)
+
+
+@shared_task
+@receiver(post_save, sender=Organization)
+def update_enterprise_inclusion_for_courses_and_programs(instance, created, **kwargs):  # pylint: disable=unused-argument
+    update_org_program_and_courses_ent_sub_inclusion.delay(
+        org_pk=instance.pk, org_sub_inclusion=instance.enterprise_subscription_inclusion
+    )
 
 
 @receiver(post_save, sender=Course)
--- a/course_discovery/apps/course_metadata/tasks.py
+++ b/course_discovery/apps/course_metadata/tasks.py
@@ -1,0 +1,49 @@
+"""
+Celery tasks for course metadata.
+"""
+import logging
+
+from celery import shared_task
+
+from course_discovery.apps.course_metadata.models import Course, CourseType, Program, ProgramType
+
+LOGGER = logging.getLogger(__name__)
+
+
+@shared_task()
+def update_org_program_and_courses_ent_sub_inclusion(org_pk, org_sub_inclusion):
+    """
+    Task to update an organization's child courses' and programs' enterprise subscription inclusion status upon saving
+    the org object.
+    Arguments:
+        org_pk (int): primary key of the organization
+        org_sub_inclusion (bool): whether or not the org is included in enterprise subscriptions
+    """
+    courses = Course.objects.filter(
+        authoring_organizations__pk=org_pk,
+        enterprise_subscription_inclusion__in=[None, True],
+        type__slug__in=[
+            CourseType.AUDIT,
+            CourseType.VERIFIED_AUDIT,
+            CourseType.PROFESSIONAL,
+            CourseType.CREDIT_VERIFIED_AUDIT,
+            CourseType.EMPTY
+        ]
+    )
+    course_ids = []
+    for course in courses:
+        course.enterprise_subscription_inclusion = org_sub_inclusion
+        course.save()
+        course_ids.append(course.id)
+    programs = Program.objects.filter(
+        courses__in=course_ids,
+        type__slug__in=[
+            ProgramType.XSERIES,
+            ProgramType.MICROMASTERS,
+            ProgramType.PROFESSIONAL_CERTIFICATE,
+            ProgramType.PROFESSIONAL_PROGRAM_WL,
+            ProgramType.MICROBACHELORS
+        ]
+    )
+    for program in programs:
+        program.save()
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -31,9 +31,9 @@ from course_discovery.apps.core.utils import SearchQuerySetWrapper
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
     FAQ, AbstractHeadingBlurbModel, AbstractMediaModel, AbstractNamedModel, AbstractTitleDescriptionModel,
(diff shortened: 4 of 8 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -2724,8 +2743,6 @@ def is_active(self):
         return self.status == ProgramStatus.Active
 
     def _check_enterprise_subscription_inclusion(self):
-        if self.type == 'micromasters':
-            return False
```
kiram15, line 2769: TY!

alex-sheehan-edx: of course!

course_discovery/apps/course_metadata/models.py
```diff
@@ -992,6 +1008,9 @@ def save(self, *args, **kwargs):
         kwargs['force_insert'] = False
         kwargs['force_update'] = True
         super().save(*args, **kwargs)
```
kiram15, line 1038: In the above method of `_check_enterprise_subscription_inclusion`, we now need to check if it is of the course types we allow. Instead of excluding, we should do an include so that if new weird ones are added, we don't have to change the code. Using the CourseTypes detailed in this file, this would probably look something like this.

```
enterprise_catalog_course_types = [CourseType.AUDIT, CourseType.VERIFIED_AUDIT, CourseType.PROFESSIONAL, 
CourseType.CREDIT_VERIFIED_AUDIT, CourseType.EMPTY] (probably in a constants file)

if self.type in enterprise_catalog_course_types or self.enterprise_subscription_inclusion is False:
...
```




.

course_discovery/apps/course_metadata/models.py
kiram15: Actually, replace this with

```
enterprise_catalog_program_types = [ProgramType.XSERIES, ProgramType.MICROMASTERS, ProgramType.PROFESSIONAL_CERTIFICATE, ProgramType.PROFESSIONAL_PROGRAM_WL, ProgramType.LICENSE, ProgramType.MICROBACHELORS]

#We exclude Bachelors, Masters, and Doctorate programs as the cost per user would be too high
if self.type not in enterprise_catalog_program_types:
  return false;
```

kiram15 (commented):
So we're hacking out some changed product requirements but I tried to type as much info as possible to give you context!

course_discovery/apps/course_metadata/models.py
```diff
@@ -267,8 +267,24 @@ def save(self, *args, **kwargs):
         We cache the key here before saving the record so that we can hit the correct
         endpoint in lms.
         """
+        old_copy = Organization.objects.filter(pk=self.pk).first()
```
DawoudSheraz: pk is unique, hence you can use `Organization.objects.get(pk=self.pk)`

course_discovery/apps/course_metadata/models.py
```diff
@@ -267,8 +267,24 @@ def save(self, *args, **kwargs):
...
+        if toggle_course_subscription_inclusion:
+            courses = Course.objects.filter(authoring_organizations__pk=self.pk)
+            for course in courses:
+                course.enterprise_subscription_inclusion = self.enterprise_subscription_inclusion
+                course.save()
+
+            programs = Program.objects.filter(courses__in=courses)
+            for program in programs:
+                program.save()
+
```
DawoudSheraz: This can be an expensive operation. There are post_save signal handles for Course model that then call other services for updating information. I would advocate making this part an async task. Also, keep draft vs non-draft distinction and impact in mind when doing this operation.

alex-sheehan-edx: thanks for the suggestion!

alex-sheehan-edx: @DawoudSheraz could you talk a little more about draft vs non-draft implications? Theoretically we'd want all courses associated with the org to be updated when the org is updated, no? I could potentially only save if there's a change in the `enterprise_subscription_inclusion` value but I'm not sure that's what you meant.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2731,8 +2748,10 @@ def is_active(self):
         return self.status == ProgramStatus.Active
 
     def _check_enterprise_subscription_inclusion(self):
-        if self.type == 'micromasters':
+        # We exclude Bachelors, Masters, and Doctorate programs as the cost per user would be too high
```
azanbinzahid, line 2787: Shouldn't we exclude `LICENSE` as well?

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    courses = Course.objects.filter(
+        authoring_organizations__pk=org_pk, draft=False, enterprise_subscription_inclusion__in=[None, True],
+    )
+    course_ids = []
+    for course in courses:
+        course.enterprise_subscription_inclusion = org_sub_inclusion
+        course.save()
+        course_ids.append(course.id)
+    programs = Program.objects.filter(courses__in=course_ids)
```
azanbinzahid: are we interested in 2U programs which does not have associated courses?

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+@shared_task()
+def update_org_program_and_courses(org_pk, org_sub_inclusion):
+    """
+    Task to update course skills.
+    Arguments:
+        org_pk (int): primary key of the organization
+        org_sub_inclusion (bool): whether or not the org is included in enterprise subscriptions
+    """
+    courses = Course.objects.filter(
+        authoring_organizations__pk=org_pk, draft=False, enterprise_subscription_inclusion__in=[None, True],
```
azanbinzahid: Can it have a `False` value? What if the value is prev set to `False` and we need to set it `True`?

course_discovery/apps/course_metadata/models.py
alex-sheehan-edx: will need @kiram15's feedback on this.

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: I don't believe so because those associated programs will default to false and won't need updating in this post org save logic. Basically we only want to toggle the subscription inclusion of programs that have OCM courses that belong to the org.

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: No- any course that is set to `False` we want to stay false. We only want to toggle courses that were previously `True` or were never set in the first place (all new courses will default have a value so these `None` cases only apply to courses generated before the sub inclusion logic was implemented.

course_discovery/apps/course_metadata/models.py
kiram15: Do we have any documentation on why License should be excluded? In talking with our PM to go over requirements that wasn't one that was mentioned.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -2033,18 +2126,17 @@ def test_staff_no_advertised_course_run(self):
...
+        program1 = factories.ProgramFactory(type=program_type, courses=course_list)
         assert program1.enterprise_subscription_inclusion is False
-
-        course3 = factories.CourseFactory(enterprise_subscription_inclusion=True)
-        course4 = factories.CourseFactory(enterprise_subscription_inclusion=True)
+        course3 = factories.CourseFactory(enterprise_subscription_inclusion=True, type=course_type)
+        course4 = factories.CourseFactory(enterprise_subscription_inclusion=True, type=course_type)
         course_list2 = [course3, course4]
-        program2 = factories.ProgramFactory(type=type1, courses=course_list2)
+        program2 = factories.ProgramFactory(courses=course_list2, type=program_type)
```
kiram15, line 2140: Can we add in a test that checks against a program that is not included in the `is_enterprise_catalog_program_type` list?

alex-sheehan-edx: yeah definitely

course_discovery/settings/test.py
```diff
@@ -72,7 +72,8 @@
 
 CELERY_TASK_IGNORE_RESULT = True
 
-CELERY_RESULT_BACKEND = f'file://{tempfile.TemporaryDirectory()}'
+results_dir = tempfile.TemporaryDirectory()
```
alex-sheehan-edx, line 75: seems like testing celery needed a small tweak

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -0,0 +1,31 @@
+"""
+Celery tasks for course metadata.
+"""
+import logging
+
+from celery import shared_task
+
+from .models import Course, Program
```
DawoudSheraz: nit: please use absolute import path.

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+@shared_task()
+def update_org_program_and_courses(org_pk, org_sub_inclusion):
+    """
+    Task to update course skills.
```
DawoudSheraz: Please rename the task to showcase what is actually being done. Same for the docstrings.

course_discovery/apps/course_metadata/tasks.py
DawoudSheraz: - Add an additional filter on course type, to filter only those courses whose types can be included in the enterprise catalog
- I suspect draft=False would not work here. What we should do is to filter the draft versions and if a draft version has an official version (non-draft), copy the changes from the draft to the non-draft version. This will make sure to keep draft and non-draft versions in sync. Otherwise, whenever any of the courses are edited by this task, their next edit run via the publisher, which performs edits on draft first will override these changes in DB when copying values from Draft -> Non-draft. Let me know if you need further clarification on it.

course_discovery/apps/course_metadata/tasks.py
DawoudSheraz: Add filtering against allowed program types here as well.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -180,6 +182,14 @@ def update_or_create_salesforce_organization(instance, created, **kwargs):
             util.update_publisher_organization(instance)
 
 
+@shared_task
+@receiver(post_save, sender=Organization)
+def hydrate_org_programs_and_courses(instance, created, **kwargs):  # pylint: disable=unused-argument
```
DawoudSheraz: nit: this can be renamed to a meaningful name.

course_discovery/apps/course_metadata/models.py
```diff
@@ -992,6 +1005,9 @@ def save(self, *args, **kwargs):
         kwargs['force_insert'] = False
         kwargs['force_update'] = True
         super().save(*args, **kwargs)
+        course_runs = CourseRun.objects.filter(course=self)
+        for course_run in course_runs:
+            course_run.save()
```
DawoudSheraz, line 1044: Why is this needed?

alex-sheehan-edx: The course run subscription tagging status is calculated and updated when the course run is saved. This is determined based on the ent sub tagging status of it's parent course. When a course is saved, we want to update the tagging status of each of its course runs.

course_discovery/apps/course_metadata/signals.py
alex-sheehan-edx: What were you thinking? To me this explains exactly what's going on. We have a parent org that has been updated and we need to ensure the children programs and courses of the are up to date with subscription tagging.

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: oops! my mistake for missing the doc string.

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: I'm not sure that's necessary. We already have allowed program type filter on the program save. Subscription inclusion for programs is calculated on save same as course runs and does the filtering for us for allowed types.

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: > Add an additional filter on course type, to filter only those courses whose types can be included in the enterprise catalog

this should be taken care of by the course's save `_check_enterprise_subscription_inclusion` logic

alex-sheehan-edx: > I suspect draft=False would not work here. What we should do is to filter the draft versions and if a draft version has an official version (non-draft), copy the changes from the draft to the non-draft version. This will make sure to keep draft and non-draft versions in sync. Otherwise, whenever any of the courses are edited by this task, their next edit run via the publisher, which performs edits on draft first will override these changes in DB when copying values from Draft -> Non-draft. Let me know if you need further clarification on it.

I think I need more clarifications on this. Would you be available to huddle and talk at some point?

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: Something along the lines of `update_enterprise_inclusion_for_courses_and_programs`.

course_discovery/apps/course_metadata/tasks.py
DawoudSheraz: Adding the filter on course type will only bring those results whose enterprise inclusion value needs to be updated.

course_discovery/apps/course_metadata/models.py
```diff
@@ -992,6 +1005,9 @@ def save(self, *args, **kwargs):
         kwargs['force_insert'] = False
         kwargs['force_update'] = True
         super().save(*args, **kwargs)
+        course_runs = CourseRun.objects.filter(course=self)
```
DawoudSheraz, line 1042: nit: a small comment on why this is done will be helpful for the future devs.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: License and Bachelors are new types added for new degrees import. Maybe they can be added here in a followup PR once confirmed internally.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1238,6 +1245,91 @@ def test_user_organizations(self):
 
         assert len(Organization.user_organizations(user)) == 1
 
+    def test_org_enterprise_subscription_inclusion_toggle_course(self):
```
DawoudSheraz, line 1249: 👍🏽 Nicely done.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/models.py
azanbinzahid: In essence, these were added together: https://github.com/openedx/course-discovery/pull/3450

course_discovery/apps/course_metadata/models.py
alex-sheehan-edx: happy to add a comment!

PR #3528, opened 2022-07-18
fix: convert google drive link image url to downloadable url

AliAdnanSohail:
# Description:
We are getting some google drive url for course images during ingestion. So we need to convert that google drive url to a downloadable url and then download the file to save course image


DawoudSheraz (approved):
once the build is green.

uzairr (approved):

PR #3529, opened 2022-07-18
feat: adding organization patch endpoint

kiram15:
My team needs to update over 100 organization's enterprise_subscription_inclusion variables, and since there is no way to do this through the current API, here's a PR to add a partial_update functionality to the Organization view. 

NOTE: If any additional security measures are needed, please describe them below and I will add them to this PR.

### DIFF
--- a/course_discovery/apps/api/v1/views/organizations.py
+++ b/course_discovery/apps/api/v1/views/organizations.py
@@ -10,7 +10,7 @@ from course_discovery.apps.publisher.models import OrganizationExtension
 
 
 # pylint: disable=useless-super-delegation
-class OrganizationViewSet(CompressedCacheResponseMixin, viewsets.ReadOnlyModelViewSet):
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/views/organizations.py
```diff
@@ -50,3 +50,7 @@ def list(self, request, *args, **kwargs):
     def retrieve(self, request, *args, **kwargs):
         """ Retrieve details for an organization. """
         return super().retrieve(request, *args, **kwargs)
+
+    def partial_update(self, request, *args, **kwargs):
+        """ Update one or more fields for an organization. """
+        return super().partial_update(request, *args, **kwargs)
```
DawoudSheraz, line 56: This would need a permission class to verify if the user making the request is allowed to do so. This is done for the course and course-run APIs. The organization model is a very high risk model and any unauthorized attempts, both intentional and unintentional, to change other org data should not be allowed.

kiram15:
Will instead use a management command

PR #3530, opened 2022-07-19
WS-2936/WS-3083: add in-year value fields to product models

bseverino:
### Description

Algolia allows for weighting of search results based on any field we send them. A useful application of this would be to elevate products based on their in-year value per lead and value per click.

To do this, we’ll need to pass Algolia a figure for each product – for both edX and 2U products.

I've added a `Value` model to store these fields. There is a follow up ticket to add these fields to Publisher.

### Acceptance Criteria

1. All product-level objects in Disco include fields for U.S. in-year value per lead and value per click
2. All product-level objects in Disco include fields for international in-year value per lead and value per click
3. The fields should accept an integer
4. For each product that has them, U.S. and international in-year value per lead and value per click should be indexed in Algolia

### Internal Tickets

https://2u-internal.atlassian.net/browse/WS-2936
https://2u-internal.atlassian.net/browse/WS-3083

PR #3531, opened 2022-07-20
feat: [WS-3083] add value per click to courses and programs

srwang:
This PR adds value per click to courses and programs to allow Algolia to weigh search results by these fields.  Value per click has been added as two fields: one which represents the in-year value for US visitors only and one that represents the value for international visitors.  Adding these fields will allow us to rank search results with greater granularity.

See https://2u-internal.atlassian.net/browse/WS-3083


course_discovery/apps/course_metadata/models.py
```diff
@@ -966,6 +966,7 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         null=True,
         help_text=_('This field signifies if this course is in the enterprise subscription catalog'),
     )
+    value_per_click = models.IntegerField(null=True, blank=True)
```
bseverino: based on the AC can you split these into two fields each, one for US and one for international?

course_discovery/apps/course_metadata/index.py
```diff
@@ -130,7 +131,8 @@ class SpanishProductIndex(BaseProductIndex):
         ],
         'attributesForFaceting': ['partner', 'availability', 'subject', 'level', 'language', 'product', 'program_type',
                                   'filterOnly(staff)', 'filterOnly(allowed_in)', 'filterOnly(blocked_in)'],
-        'customRanking': ['desc(promoted_in_spanish_index)', 'asc(availability_rank)', 'desc(recent_enrollment_count)']
+        'customRanking': ['desc(promoted_in_spanish_index)', 'asc(availability_rank)',
+                          'desc(recent_enrollment_count)', 'desc(value_per_click)']
```
bseverino: I think we want to hold off on using this field for custom ranking just yet - we just want to pass it in without doing anything with it until [WS-3137](https://2u-internal.atlassian.net/browse/WS-3137)

bseverino (changes requested):

bseverino (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -966,6 +966,8 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         null=True,
         help_text=_('This field signifies if this course is in the enterprise subscription catalog'),
     )
+    value_per_click_usa = models.IntegerField(null=True, blank=True, default=0)
+    value_per_click_international = models.IntegerField(null=True, blank=True, default=0)
```
DawoudSheraz: Both PR and ticket do not specify why is there a need for two fields? Add help text explaining these fields.

DawoudSheraz (commented):
- Added a question for more clarity on the changes being added in this PR
- Make sure the PR is only adding one migration file. Squash your migrations or recreate them in a single file.
- Make PR descriptive to explain why are the fields being added
- Is there no use of these fields in ElasticSearch documents and ES document APIs?

srwang:
> * Added a question for more clarity on the changes being added in this PR
> * Make sure the PR is only adding one migration file. Squash your migrations or recreate them in a single file.
> * Make PR descriptive to explain why are the fields being added
> * Is there no use of these fields in ElasticSearch documents and ES document APIs?

Thanks Darwoud. 

I've made these changes.

Nope, these fields will not be used in the ES APIs.

DawoudSheraz (commented):
The changes look good. A question: I was seeing another PR by @bseverino , https://github.com/openedx/course-discovery/pull/3530, which is adding value per lead fields to the products. Why can't the lead and value fields be transformed into a separate model and that model is then linked to Program and Courses? The intent of the fields seems similar to me and it would make sense to add these related fields in a separate model.

uzairr (approved):
looks pretty straight forward! approved from my side once @DawoudSheraz concern is addressed!

bseverino:
Thank you both! We've condensed these PRs into one, and split the fields into a separate model as per @DawoudSheraz's suggestion: https://github.com/openedx/course-discovery/pull/3530

PR #3536, opened 2022-07-25
feat: create backfill management command

kiram15:
Uses a provided set of production course keys and organization uuids that are currently participating in the enterprise subscription catalog to backfill the new boolean field. 
https://2u-internal.atlassian.net/browse/ENT-5839


course_discovery/apps/course_metadata/management/commands/constants.py
```diff
@@ -0,0 +1,1722 @@
+org_uuids = ['f5d1ff66002f43fa9d761b191d4a3272',
```
johnnagro, line 1: i like the simplicity of including the input data this way versus including a separate set of input files. less code (no file handling), easier to test, we still get git history of what data was used.

johnnagro (approved):

Ali-D-Akbar (commented):
Looks good. Just a couple of questions here:
1. Is this going to be a recurring or just a one-time task?
2. If this is just a one-time task, are we planning to remove the constants.py file once it has served its purpose in another commit? (if yes, we can add a TODO in the constants.py file just in case)
4. If this is a recurring task, did you consider using a config model with a filefield? (which I think, is a better option in this case)

kiram15:
> Looks good. Just a couple of questions here:
> 
> 1. Is this going to be a recurring or just a one-time task?
> 2. If this is just a one-time task, are we planning to remove the constants.py file once it has served its purpose in another commit? (if yes, we can add a TODO in the constants.py file just in case)
> 3. If this is a recurring task, did you consider using a config model with a filefield? (which I think, is a better option in this case)

This will be a one time task, I can remove the constants file if that would lead to a cleaner codebase. I'll put a PR up for it right after this is merged in!

Ali-D-Akbar (approved):

PR #3538, opened 2022-07-27
Small bug fix for course discovery 

kiram15:
I made an error where I was checking is a value was False, but not accounting for the fact that all of these booleans have null=True in their model definition. To combat this, I changed the logic to instead be `not boolean` because None is falsy in python.


Ali-D-Akbar (approved):

DawoudSheraz (approved):
Please squash your commits and add appropriate commit message to make it ready for merge.

PR #3539, opened 2022-07-28
feat: add a signal on update program

Ali-D-Akbar:
[PROD-2902](https://2u-internal.atlassian.net/browse/PROD-2902)
Adds a signal to be used in [taxonomy-connector](https://github.com/openedx/taxonomy-connector/pull/94) whenever the degree is updated based on overview.

Steps to verify:
1. set `FIRE_UPDATE_DEGREE_SKILLS_SIGNAL` to True
2. Publish a program through django admin
3. Notice a logged message similar to this: `Signal fired to update program skills. Program: [uuid]'`


course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -176,6 +179,10 @@ def _update_or_create_degree(
             "partner": self.partner,
 
         }
+        # We will be refreshing skills when overview is updated in degree.
+        degree = Degree.objects.filter(marketing_slug=data['slug'], overview=data['overview'])
```
AliAdnanSohail: Rather we trigger signal in data loader, can we move this to model level? So if Degree is changed from any other place like Admin or publisher. It will trigger this signal from there as well? 
We can use `post_save` callback on model level, so if this model is updated from anywhere it will trigger this signal

May be this link will help you
https://docs.djangoproject.com/en/4.0/ref/signals/#post-save

AliAdnanSohail (changes requested):

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
Ali-D-Akbar: Okay, I'll do that instead.

DawoudSheraz: Yes, I second Ali Adnan's comment here. The data loader is used for a limited set of data. This solution should be for Programs model as a whole.

AliAdnanSohail (approved):

AliAdnanSohail (approved):
Wait for the green build, then it seems good to me

course_discovery/apps/course_metadata/models.py
```diff
@@ -2761,6 +2762,12 @@ def save(self, *args, **kwargs):
                 kwargs['force_update'] = True
                 super().save(**kwargs)
                 publisher.publish_obj(self, previous_obj=previous_obj)
+            if settings.FIRE_UPDATE_PROGRAM_SKILLS_SIGNAL and self.is_active and \
+                    previous_obj and not previous_obj.is_active:
```
DawoudSheraz, line 2766: This condition can be simplified or converted into a util method for readability purposes.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -2105,6 +2106,25 @@ def test_publication_enabled(self):
             program.delete()
             assert mock_delete_obj.called
 
+    @override_switch('publish_program_to_marketing_site', True)
```
DawoudSheraz, line 2109: Why should this switch be on?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -2105,6 +2106,25 @@ def test_publication_enabled(self):
...
+        """
+        Verify that publishing the program fires UPDATE_PROGRAM_SKILLS signal.
+        """
+        program = factories.ProgramFactory(status=ProgramStatus.Unpublished)
+        program.status = ProgramStatus.Active
+        with mock.patch.object(ProgramMarketingSitePublisher, 'publish_obj', return_value=None) as mock_publish_obj:
+            program.save()
+            assert mock_publish_obj.called
+            assert mock_signal_send.called
+            assert mock_signal_send.call_count == 1
```
DawoudSheraz, line 2122: MarketingSitePublisher is not impacting the signal because the signal is sent in a different execution block. I am not sure why this has been done.

DawoudSheraz (commented):
A few questions to address.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: This is needed for publishing on marketing.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: https://github.com/openedx/course-discovery/blob/a2f571d2db3e858378b5b6c66df65b45a8ee98c3/course_discovery/apps/course_metadata/models.py#L2752

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Please let me know if there's a better way to write this condition.

DawoudSheraz, line 2766: Why is there a need to check on the old object? Isn't setting + is_active enough?

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: The skill update signal code block does not seem to come under `publish_program_to_marketing_site` flow? Why is this needed then? Can't it be kept as off?

DawoudSheraz: Oh, nevermind, got my answer in the upcoming comments.

DawoudSheraz (approved):
- nit comment/question
- Rebase before merge.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: We're checking it so that if the the program status is changed in the recent save. Otherwise, it will always trigger the signal in case program status is active/published. We want to trigger from unpublish to publish state only.

DawoudSheraz: So let's say the overview field of a program is updated. What will happen then?

Ali-D-Akbar: It won't send the signal. Aren't we just updating Program upon publish only according to the ADR? 
`When a program is published in Discovery, emit a signal indicating the program skills must be updated.`

DawoudSheraz: Well, this was not meant literally only on publish. For the courses, if the setting is on and if any skill-related field is changed, the course skill update signal is re-sent. Please check https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L399-L404. You will need a similar mechanism here.

PR #3540, opened 2022-07-29
feat: add uuid filtering on learner pathway API

sameenfatima78:
Allows filtering via uuid on learner-pathway list endpoint. Also supports comma-delimited uuids list for filtering. 

JIRA: https://2u-internal.atlassian.net/browse/ENT-6069

### DIFF
--- a/course_discovery/apps/learner_pathway/api/filters.py
+++ b/course_discovery/apps/learner_pathway/api/filters.py
@@ -1,0 +1,22 @@
+"""
+Filters for the Learner Pathway APIs.
+"""
+
+from django_filters import rest_framework as filters
+
+from course_discovery.apps.learner_pathway.models import LearnerPathway
+
+
+class PathwayUUIDFilter(filters.FilterSet):
+    """
+    Filter pathways by uuid. Supports filtering by comma-delimited list of uuids.
+    """
+    uuid = filters.CharFilter(method='filter_by_uuid')
+
+    class Meta:
+        model = LearnerPathway
+        fields = ['uuid']
+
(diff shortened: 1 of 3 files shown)


sameenfatima78 (commented):
@DawoudSheraz @Ali-D-Akbar Hey, can you please review these changes?

saleem-latif (approved):

course_discovery/apps/learner_pathway/api/filters.py
```diff
@@ -0,0 +1,22 @@
...
+from django_filters import rest_framework as filters
+
+from course_discovery.apps.learner_pathway.models import LearnerPathway
+
+
+class PathwayUUIDFilter(filters.FilterSet):
+    """
+    Filter pathways by uuid. Supports filtering by comma-delimited list of uuids.
+    """
+    uuid = filters.CharFilter(method='filter_by_uuid')
```
HammadAhmadWaqas, line 14: Can we look for [ModelChoiceFilter](https://django-filter.readthedocs.io/en/main/ref/filters.html#modelchoicefilter)
What will suit more in our case

DawoudSheraz (approved):

Ali-D-Akbar (approved):

course_discovery/apps/learner_pathway/api/filters.py
sameenfatima78: I think using `ModelChoiceFilter` makes more sense when the choices are well-defined. In our case, we are filtering based on uuid which could be any random value so sticking with `CharFiiter` here.

PR #3543, opened 2022-08-02
fix: fixing incorrect course keys

kiram15:
Our PM made a mistake when fetching the course keys, here are the updated ones along with error handling.


course_discovery/apps/course_metadata/management/commands/backfill_enterprise_inclusion.py
```diff
@@ -1,8 +1,13 @@
+import logging
+from ast import Or
```
Ali-D-Akbar: What is this import used for?

Ali-D-Akbar (changes requested):
As I said in the previous PR, rather than using the constant file, it may be better to use a config model to store the file through django admin. In my suggested approach, you wouldn't have to create another PR to update the course keys data.

This way, we'll need another PR to remove the constants file.

mphilbrick211:
@azanbinzahid would you mind taking a look at these changes? Thank you!

azanbinzahid (approved):

DawoudSheraz:
@kiram15 Hi, is this still needed? Thanks

mphilbrick211:
Hi @kiram15 - just following up on this. Please let us know if you plan to proceed with this pull request. Thanks!

PR #3546, opened 2022-08-03
feat: Add management command for consuming events.

dianakhuang:
https://github.com/openedx/event-bus-kafka/issues/4

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/consume_events.py
+++ b/course_discovery/apps/course_metadata/management/commands/consume_events.py
@@ -1,0 +1,83 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger(__name__)
+
+
+# FIXME: This is currently a direct copy of the Kafka implementation. In the future,
+# we would like to make this configurable to different implementations of the event bus.
+class Command(BaseCommand):
+    """
+    Consumes events off the event bus implementation and fires signals in response.
+    """
+
+    help = """
+    This starts a Kafka event consumer that listens to the specified topic and logs all messages it receives. Topic
+    is required.
+    example:
+        python3 manage.py cms consume_events -t user-event-debug -g user-event-consumers
+    # TODO (EventBus): Add pointer to relevant future docs around topics and consumer groups, and potentially
+    update example topic and group names to follow any future naming conventions.
+    """
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            '-t', '--topic',
+            nargs=1,
+            required=True,
+            help='Topic to consume'
+        )
+
+        parser.add_argument(
+            '-g', '--group_id',
+            nargs=1,
+            required=True,
+            help='Consumer group id'
+        )
+        parser.add_argument(
+            '-s', '--signal',
+            nargs=1,
+            required=True,
+            help='Type of signal to emit from consumed messages.'
+        )
+
+    # We are not testing the current version of this code because it is copied from tested code
+    # in event-bus-kafka, and this code will have to change once we have a better idea for the
+    # consumer API. (https://github.com/openedx/event-bus-kafka/issues/19)
+    def handle(self, *args, **options):  # pragma: no cover
+        try:
+            # FIXME: This needs to be replaced with a configurable loading
+            # mechanism built into openedx-events.
+            # See https://github.com/openedx/openedx-events/pull/85)
+            from edx_event_bus_kafka.consumer.event_consumer import \
+                KafkaEventConsumer  # pylint: disable=import-outside-toplevel
+            from edx_toggles.toggles import SettingToggle  # pylint: disable=import-outside-toplevel
+            from openedx_events.tooling import OpenEdxPublicSignal  # pylint: disable=import-outside-toplevel
+        except ImportError:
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -0,0 +1,16 @@
...
+    # mechanism built into openedx-events.
+    # See https://github.com/openedx/openedx-events/pull/85)
+    from edx_event_bus_kafka.consumer.event_consumer import ConsumeEventsCommand
+except ImportError:
+    pass
+
+
+# FIXME: This is currently a direct copy of the Kafka implementation. In the future,
+# we would like to make this configurable to different implementations of the event bus.
+class Command(ConsumeEventsCommand):
```
robrap: 1. [question] I'm not clear if we want a `BaseCommand` to be part of the ultimate API, or just functions that the local command can call? This is a non-blocking question.
2. Consider renaming `Command` to `DiscoveryConsumeEventsCommand`.

robrap (commented):
I'm not clear on the final design, but this seems ok as an iterative step towards tying everything together.

course_discovery/apps/course_metadata/management/commands/consume_events.py
dianakhuang: 1. I think this sort of `BaseCommand` to be part of the API doesn't make sense just yet. I think the way I am imagining this going forward is that we'll have the topic/group/events configured via settings and then this management command can read the settings instead of taking them as arguments. But I think for this first iteration, we can do it this way.
2. I did name it that way at first, but then trying to run the management command locally for testing, it complained that there wasn't a `Command` defined in this module, so I switched it to the generic name.

robrap (approved):

course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -0,0 +1,80 @@
...
+        parser.add_argument(
+            '-s', '--signal',
+            nargs=1,
+            required=True,
+            help='Type of signal to emit from consumed messages.'
+        )
+
+    def handle(self, *args, **options):
+        if not KAFKA_CONSUMERS_ENABLED.is_enabled():
+            logger.error("Kafka consumers not enabled")
```
robrap: Shouldn't this just be an info?

course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -0,0 +1,80 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+try:
```
timmc-edx: I think whole block (and the declaration of `KAFKA_CONSUMERS_ENABLED`) needs to move into the `handle` method.

timmc-edx: e.g. SettingToggle is used at the top level, so the module will fail to load if the import fails anyhow.

course_discovery/apps/course_metadata/management/commands/consume_events.py
dianakhuang: This is what it is in the originally code, but I don't mind switching it to use `info` instead.

course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        parser.add_argument(
+            '-s', '--signal',
+            nargs=1,
+            required=True,
+            help='Type of signal to emit from consumed messages.'
+        )
+
+    def handle(self, *args, **options):
+        if not KAFKA_CONSUMERS_ENABLED.is_enabled():
```
robrap, line 71: Can we also make test coverage more explicit, by either testing all or some and/or marking what we aren't testing and why? It would be nice to know if it is being left untested because we don't see a good way to test, or because we plan to refactor and don't want to over invest yet, etc. You can use pragmas with added comments.

course_discovery/apps/course_metadata/management/commands/consume_events.py
robrap: @timmc-edx: Are you saying "needs" because you think this won't work as-is, or because you want it in the other method?

timmc-edx: Yeah, I think it would break when the `edx_event_bus_kafka` package isn't installed.

course_discovery/apps/course_metadata/management/commands/consume_events.py
robrap: Thanks. I also add an AC for test coverage to https://github.com/openedx/event-bus-kafka/issues/19

robrap (approved):

course_discovery/apps/course_metadata/management/commands/consume_events.py
robrap: I see. Because of the new `return` in the `handle` method.
Note: This is how this was handled with monitoring and New Relic imports: https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/monitoring/internal/utils.py#L22-L25.
In any case, this should all be temporary.

PR #3548, opened 2022-08-04
feat: add coordinate fields

bseverino:
- Add `GeoLocation` model with `lat` and `lng` fields, and add corresponding `geolocation` field to product models
   - Gave this model a generic name in case we want to build more geo information on top of it
- Update serializers, including update method for course geolocation field
- Index coordinates as `geo_field` in Algolia ([see algoliasearch-django docs](https://github.com/algolia/algoliasearch-django#geo-search-1))

Internal ticket: https://2u-internal.atlassian.net/browse/WS-3236


course_discovery/apps/course_metadata/models.py
```diff
@@ -983,6 +983,8 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         validators=[FileExtensionValidator(['png'])]
     )
 
+    lat = models.FloatField(null=True, blank=True, verbose_name=_('Latitude'))
+    lng = models.FloatField(null=True, blank=True, verbose_name=_('Longitude'))
```
bseverino, line 987: Opted to just add these fields as opposed to creating a separate model, as I'm not sure it's worth the extra complexity, but I could use opinions there.

marekwrobel (approved):

srwang (approved):

ansabgillani (approved):
LGTM. Please take another review from the team.

Ali-D-Akbar (approved):

course_discovery/apps/api/serializers.py
```diff
@@ -1309,6 +1335,17 @@ def update_additional_metadata(self, instance, additional_metadata):
 
         # save() will be called by main update()
 
+    def update_geolocation(self, instance, geolocation):
+        # Allow setting geolocation field to null
+        if not geolocation:
+            instance.geolocation = None
```
DawoudSheraz: Under what circumstances the geolocation can be null? The update method is checking the existence of geolocation in dict.

bseverino: My thought here was that we would want a way to set this field to null through Publisher - however, thinking about it since you brought it up, it's going to be an edge case and not really worth implementing when we can just adjust manually through admin. I removed this.

DawoudSheraz:
@bseverino This is ready to merge, correct?

bseverino:
Hi @DawoudSheraz, sorry I was out of office - yes, this is ready to merge, I just rebased now. Thanks!

PR #3550, opened 2022-08-05
feat: add tag field for Program model

DawoudSheraz:
### Description

- Add taggable manager in Program model
- Expose the tags in Program serializer
- Add/update unit tests
- semgrep is disable on Taggable manager field because semgrep expected to add null=True on the non-text field when blank=Truee is added. However, the tag manager is handling null=True internally and it does not accept that as an argument.

The query count update in most tests is the addition of the following query:

```
SELECT DISTINCT "taggit_tag"."id", "taggit_tag"."name", "taggit_tag"."slug" FROM "taggit_tag" INNER JOIN "taggit_taggeditem" ON ("taggit_tag"."id" = "taggit_taggeditem"."tag_id") INNER JOIN "django_content_type" ON ("taggit_taggeditem"."content_type_id" = "django_content_type"."id") WHERE ("django_content_type"."app_label" = 'course_metadata' AND "django_content_type"."model" = 'program' AND "taggit_taggeditem"."object_id" = 4)
```

Only one test went from 77 -> 82. The query count increase by this PR is still 3. However, 2 query count has been increased in some prior PR but the tests didn't fail due to a threshold of 2 on FuzzyInt

### Testing
- Checkout the branch and migrate the changes
- Go to program admin and you will see labels field when editing a program. 
- Add various labels and save
- Go to http://localhost:18381/api/v1/programs/<Program_uuid> and verify the label names are present in the response.


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -1073,6 +1075,7 @@ def get_expected_data(cls, program, request):
             'primary_subject_override': SubjectSerializer(program.primary_subject_override).data,
             'level_type_override': LevelTypeSerializer(program.level_type_override).data,
             'language_override': program.language_override.code,
+            'labels': ['topic'] if include_labels else []
```
azanbinzahid, line 1078: (why) do we need quotes around topic?

DawoudSheraz: Because that is the topic name. 
https://github.com/openedx/course-discovery/pull/3550/files#diff-13102dcecb6b87aa0ee8cecd952a3d5ed06b344c6155ce2a4cf4fada8adfe8caR1359

AliAdnanSohail:
@DawoudSheraz a question, Do we need to add labels in API and serializer? As far as I remember user can update/create tags from django admin only

DawoudSheraz:
> @DawoudSheraz a question, Do we need to add labels in API and serializer? As far as I remember user can update/create tags from django admin only

The program API is only used for listing only.

AliAdnanSohail (approved):

uzairr (approved):

azanbinzahid (approved):
We need to keep track of increasing queries (hence time) in future.

PR #3555, opened 2022-08-10
fix: send signal when overview is updated

Ali-D-Akbar:
[PROD-2902](https://2u-internal.atlassian.net/browse/PROD-2902)
Sends signal whenever overview field changes instead of when the program is published.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -2820,9 +2820,9 @@ class Program(PkSearchableMixin, TimeStampedModel):
                 kwargs['force_update'] = True
                 super().save(**kwargs)
                 publisher.publish_obj(self, previous_obj=previous_obj)
-            if settings.FIRE_UPDATE_PROGRAM_SKILLS_SIGNAL and self.is_active and \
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -2816,9 +2816,8 @@ def save(self, *args, **kwargs):
                 kwargs['force_update'] = True
                 super().save(**kwargs)
                 publisher.publish_obj(self, previous_obj=previous_obj)
-            if settings.FIRE_UPDATE_PROGRAM_SKILLS_SIGNAL and self.is_active and \
-                    previous_obj and not previous_obj.is_active:
-                # If a Program is published then fire signal
+            if settings.FIRE_UPDATE_PROGRAM_SKILLS_SIGNAL and previous_obj and self.overview != previous_obj.overview:
+                # If a Program overview field is changed then fire signal
```
DawoudSheraz: We should keep the active check in place. It should be OR among (Degree is published), (Overview changes)

DawoudSheraz (approved):

ansabgillani (approved):

PR #3556, opened 2022-08-11
feat: Add logging to event handling.

dianakhuang:
Add the ability to log information when we consume
course catalog events from the event bus.

Eventually, this code will be replaced by code that processes the course data in accordance with what refresh_course_metadata does currently.

Testing:
- We are planning on using this code to test the end-to-end infrastructure of the event bus on Stage. When an event is consumed by course-discovery, we want to verify that the correct signal is sent and that we can deserialize the event data correctly.

https://github.com/edx/edx-arch-experiments/issues/48

### DIFF
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -325,3 +325,45 @@ def check_course_runs_within_course_for_duplicate_external_key(course, specific_
         if external_key == specific_course_run.external_key and course_run != specific_course_run:
             message = _duplicate_external_key_message([course_run])
             raise ValidationError(message)
+
+
+# This is code to handle signals emitted from the event bus. This is currently test code and not to be used
+# for production use cases.
+try:  # pragma: no cover
+    from edx_toggles.toggles import SettingToggle  # pylint: disable=import-outside-toplevel
+    from openedx_events.content_authoring.data import CourseCatalogData  # pylint: disable=import-outside-toplevel
+    from openedx_events.content_authoring.signals import \
+        COURSE_CATALOG_INFO_CHANGED  # pylint: disable=import-outside-toplevel
+
+    # .. toggle_name: LOG_KAFKA_EVENTS
+    # .. toggle_implementation: SettingToggle
+    # .. toggle_default: False
+    # .. toggle_description: Enables the logging of events from the Kafka event bus
+    # .. toggle_use_cases: opt_in
+    # .. toggle_creation_date: 2022-08-11
+    # .. toggle_tickets: https://github.com/edx/edx-arch-experiments/issues/48
+    LOG_KAFKA_EVENTS = SettingToggle('LOG_KAFKA_EVENTS', default=False)
+
+    @receiver(COURSE_CATALOG_INFO_CHANGED)
+    def log_event_from_event_bus(**kwargs):
+        """
+        Logs information in signals we should receive from the event bus.
+
+        This is test code and should be removed eventually.
+
+        Args:
+            kwargs: event data sent to signal
+        """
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/signals.py
```diff
@@ -325,3 +325,45 @@ def check_course_runs_within_course_for_duplicate_external_key(course, specific_
...
             raise ValidationError(message)
+
+
+# This is code to handle signals emitted from the event bus. This is currently test code and not to be used
+# for production use cases.
+try:  # pragma: no cover
+    from edx_toggles.toggles import SettingToggle  # pylint: disable=import-outside-toplevel
+    from openedx_events.content_authoring.data import CourseCatalogData  # pylint: disable=import-outside-toplevel
+    from openedx_events.content_authoring.signals import \
+        COURSE_CATALOG_INFO_CHANGED  # pylint: disable=import-outside-toplevel
```
timmc-edx, line 336: We could also just add the dependencies. Both of these are broadly used and we'll almost certainly be using them once this is productionized anyhow.

dianakhuang: Yeah, this is fair. I have no strong opinions on adding them in now vs. later. Were you concerned about circular dependencies here? I didn't quite follow which ones you were concerned about.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -325,3 +325,45 @@ def check_course_runs_within_course_for_duplicate_external_key(course, specific_
...
+    # .. toggle_implementation: SettingToggle
+    # .. toggle_default: False
+    # .. toggle_description: Enables the logging of events from the Kafka event bus
+    # .. toggle_use_cases: opt_in
+    # .. toggle_creation_date: 2022-08-11
+    # .. toggle_tickets: https://github.com/edx/edx-arch-experiments/issues/48
+    LOG_KAFKA_EVENTS = SettingToggle('LOG_KAFKA_EVENTS', default=False)
+
+    @receiver(COURSE_CATALOG_INFO_CHANGED)
+    def log_event_from_event_bus(**kwargs):
```
timmc-edx, line 348: Would there be any problem with doing this instead?

```suggestion
    def log_event_from_event_bus(*, catalog_info, **kwargs):
```

This would allow other kwargs but always extract the `catalog_info` param (and require it to be present, which should be a pretty safe bet).

timmc-edx: I'm OK with having just `**kwargs` if all of this is expected to be subject to change, just curious.

course_discovery/apps/course_metadata/signals.py
timmc-edx: No concern about circular dependencies. The thing with openedx-events and event-bus-kafka (and other implementations) was about two different kinds of dependencies -- one explicit (package dependency) and the other implicit (documented API contracts and such). Nothing that would cause a loop.

course_discovery/apps/course_metadata/signals.py
dianakhuang: Yeah, this was just the pattern from the previous logging code. This code is going to go away as soon as we confirm we're getting event data correctly.

timmc-edx (approved):

course_discovery/apps/course_metadata/signals.py
timmc-edx: Cool. I'm fine with taking it cautious at first.

DawoudSheraz (commented):
If this is a temporary addition, please change commit type to temp

dianakhuang:
Updated @DawoudSheraz

DawoudSheraz (approved):

PR #3562, opened 2022-08-15
feat: taxonomy-connector 1.20 | caches serialized course skills data

iloveagent57:
This will help performance related to jobs that have to fetch a lot of the
same course data multiple times during a single run.
https://2u-internal.atlassian.net/browse/ENT-6169
https://github.com/openedx/taxonomy-connector/pull/101

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -210,7 +210,7 @@ class CourseRunSearchViewSetTests(mixins.SerializationMixin, mixins.LoginMixin, 
         assert response_data == active_program.type.name
 
     @ddt.data(
-        ([{'title': 'Software Testing', 'excluded': True}], 3),
+        ([{'title': 'Software Testing', 'excluded': True}], 6),
         ([{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True}], 7),
         ([{'title': 'Software Testing', 'excluded': False}, {'title': 'Software Testing 2', 'excluded': False}], 7),
         ([{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True},
@@ -239,7 +239,7 @@ class CourseRunSearchViewSetTests(mixins.SerializationMixin, mixins.LoginMixin, 
         )
         self.reindex_courses(program)
 
-        with self.assertNumQueries(expected_queries):
+        with self.assertNumQueries(expected_queries, threshold=5):
             response = self.get_response('software', path=self.list_path)
 
         assert response.status_code == 200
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -623,7 +623,7 @@ def test_nondraft_does_not_collide_with_draft(self):
             )
 
     def test_collision_does_not_include_drafts(self):
-        with self.assertNumQueries(FuzzyInt(136, 1)):
```
long74100, line 626: neat trick, I'm wondering if the threshold should be > 1 so we don't have to deal with these changes.

long74100 (approved):

course_discovery/apps/course_metadata/tests/test_signals.py
iloveagent57: good call.  I updated to make this more lenient (although the tests were finally passing and they've been super finicky, so I hope I don't come to regret this decision :P )

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -566,7 +566,7 @@ def test_curriculum_repeats(self):
                 course_run.external_key = course_run_ba.external_key
                 course_run.save()
 
-        with self.assertNumQueries(FuzzyInt(67, 1)):
+        with self.assertNumQueries(FuzzyInt(65, 10)):
```
iloveagent57, line 569: I increased the threshold value here (it's like the margin of error) with the following reasoning:
We're already making 65 (or in some cases, 130) queries.  The goal of assertNumQueries should be that
we don't completely wreck our performance; having it be a strict count of queries is too low-level and hinders
our ability to make changes quickly.  With a higher threshold on the query count, the tests become less brittle - we can make improvements in the business logic of the system without worrying about underlying implementation details, while still having _some_ protection against introducing an egregious number of new database calls.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -585,8 +586,7 @@ def test_results_ordered_by_start_date(self, ordering):
         upcoming = CourseRunFactory(course__partner=self.partner, start=now + datetime.timedelta(weeks=4))
         course_run_keys = [course_run.key for course_run in [archived, current, starting_soon, upcoming]]
 
-        with self.assertNumQueries(12):
```
iloveagent57, line 588: Removed this particular assertion because it was extremely flaky, often failing with "9 != 12" even when providing a `FuzzyInt(12, 4)` as the expected value, which should pass the assertion for any query count between 8 and 16 (9 is in that range).

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -538,12 +537,12 @@ def test_mix_of_curriculums_with_and_without_programs(self):
         course_run = course_a.course_runs.first()
         course_run_ca = CourseRun.objects.get(external_key='ext-key-course-ca')
         message = _duplicate_external_key_message([course_run_ca])
-        with self.assertNumQueries(FuzzyInt(6, 1)):  # 3 Selects
+        with self.assertNumQueries(6):  # 3 Selects
```
iloveagent57, line 540: We don't need to use `FuzzyInt` directly, the APITestCase mixin abstracts that out and provides a `threshold` kwarg for `assertNumQueries`: https://github.com/openedx/course-discovery/blob/7c139fe081bcdcd80d80f772c74e8924e43613f0/course_discovery/apps/api/v1/tests/test_views/mixins.py#L209-L218

Ali-D-Akbar (approved):

DawoudSheraz (approved):

PR #3568, opened 2022-08-23
fix: populate exec ed data csv mgt command fixups

DawoudSheraz:
### Description

- Do not require auth_token in management command. Raise error if both token and json file is missing
- Add temporary headers that allow API call to product API, which started returning 403 for some reason. This is expected to remove once there is more info from Product API owners on why the script API call is not allowed

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
+++ b/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
@@ -54,7 +54,6 @@ class Command(BaseCommand):
             '--auth_token',
             dest='auth_token',
             type=str,
-            required=True,
             help='Bearer token for making API calls to Product API'
         )
         parser.add_argument(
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -168,7 +172,9 @@ def get_product_details(self, auth_token):
         """
         url = settings.PRODUCT_API_URL
         headers = {
-            "Authorization": f"Bearer {auth_token}"
+            "Authorization": f"Bearer {auth_token}",
+            'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 '
```
azanbinzahid, line 177: we can add an in-line comment as a reminder.

azanbinzahid (approved):

Ali-D-Akbar (approved):

PR #3572, opened 2022-08-24
feat: make created available for learner pathways in search/all

sameenfatima78:
As an enhancement to our current search page in Learner Portal, we want to sort Learner Pathways according to the their created date. For this we need to add this into the model and make it available in search/all endpoint. This is what this PR is basically about. 
Screenshot 2022-08-24 at 11 11 02 PM

JIRA: https://2u-internal.atlassian.net/browse/ENT-6173


sameenfatima78 (commented):
@DawoudSheraz Hi, can you please review these changes. Thanks

sameenfatima78:
@HammadAhmadWaqas @saleem-latif I can't seem to tag you for review so posting it here.

saleem-latif (approved):

course_discovery/apps/course_metadata/search_indexes/documents/learner_pathway.py
```diff
@@ -19,7 +19,7 @@ class LearnerPathwayDocument(BaseDocument, OrganizationsMixin):
     """
     LearnerPathway Elasticsearch document.
     """
-
+    created = fields.DateField()
```
HammadAhmadWaqas, line 22: I think we are not using this field, We are using created directly from object, So this line can be removed

HammadAhmadWaqas (changes requested):

course_discovery/apps/course_metadata/search_indexes/documents/learner_pathway.py
sameenfatima78: Actually we need this field otherwise we end up with this error on `search/all` endpoint: 
 
Screenshot 2022-08-25 at 6 19 03 PM

ansabgillani (approved):

Ali-D-Akbar (approved):
Not a blocker but Can we have a named migration instead?

DawoudSheraz (approved):

HammadAhmadWaqas (approved):

PR #3580, opened 2022-08-29
feat: custom validations on location restriction serializer

srwang:
This is a part of [WS-3120](https://2u-internal.atlassian.net/browse/WS-3120)

This PR customizes the validations on the location restriction serializer to facilitate creating and editing course location restriction from Publisher.

The validations will handle the following cases:
1. If all fields on the location restriction are null, it is assumed that the user does not wish add location restriction.  No error will occur, and no location restriction object will be created. 
2. If `countries` or `states` is set but the `restriction_type` is null, the user will receive an error.
3. If `restriction_type` is set but both `countries` and `states` is null, the user will not receive an error. The assumption is that the user wants to remove the location restriction by setting restrictions to null.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1145,13 +1145,33 @@ class CourseEditorSerializer(serializers.ModelSerializer):
 
 
 class AbstractLocationRestrictionSerializer(BaseModelSerializer):
-    restriction_type = serializers.ChoiceField(choices=AbstractLocationRestrictionModel.RESTRICTION_TYPE_CHOICES)
+    restriction_type = serializers.ChoiceField(
+        choices=AbstractLocationRestrictionModel.RESTRICTION_TYPE_CHOICES,
+        allow_null=True,
+        required=False
+    )
     countries = serializers.ListField(
-        child=CountryField(), allow_empty=True, required=False
+        child=CountryField(), allow_empty=True, allow_null=True, required=False
     )
     states = serializers.ListField(
-        child=serializers.ChoiceField(choices=CONTIGUOUS_STATES), allow_empty=True, required=False
+        child=serializers.ChoiceField(choices=CONTIGUOUS_STATES), allow_empty=True, allow_null=True, required=False
     )
+
+    def check_has_data(self, key, attrs):
+        return key in attrs and bool(attrs[key])
+
+    def validate(self, attrs):
+        has_restriction_type = self.check_has_data('restriction_type', attrs)
+        has_countries = self.check_has_data('countries', attrs)
+        has_states = self.check_has_data('states', attrs)
+
+        if (has_countries or has_states) and not has_restriction_type:
+            raise serializers.ValidationError({'Restriction Type': _('Restriction Type cannot be empty')})
+
+        countries = attrs.get('countries', [])
+        states = attrs.get('states', [])
+        attrs.update({'countries': countries, 'states': states})
+        return attrs
 
     class Meta:
         model = AbstractLocationRestrictionModel
@@ -1365,7 +1385,9 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
         if 'geolocation' in validated_data:
             self.update_geolocation(instance, validated_data.pop('geolocation'))
         if 'location_restriction' in validated_data:
-            self.update_location_restriction(instance, validated_data.pop('location_restriction'))
+            location_restriction_data = validated_data.pop('location_restriction')
+            if not all(bool(value) is False for value in location_restriction_data.values()):
+                self.update_location_restriction(instance, location_restriction_data)
         if 'in_year_value' in validated_data:
             self.update_in_year_value(instance, validated_data.pop('in_year_value'))
         return super().update(instance, validated_data)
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1145,14 +1145,34 @@ def create(self, validated_data):
...
+    def check_has_data(self, key, attrs):
+        return key in attrs and bool(attrs[key])
+
+    def validate(self, attrs):
+        if self.check_has_data('restriction_type', attrs):
+            if not self.check_has_data('countries', attrs) and not self.check_has_data('states', attrs):
+                raise serializers.ValidationError({'countries, states': _('Countries and States cannot both be empty')})
+        else:
+            if self.check_has_data('countries', attrs) or self.check_has_data('states', attrs):
+                raise serializers.ValidationError({'countries, states': _('Restriction Type cannot be empty')})
```
DawoudSheraz: For better readability, you can save:
-  responses of validate calls (has_restriction_type, ..)
- error message in a variable based on validation error
- if the message is present, raise a validation error

course_discovery/apps/api/serializers.py
```diff
@@ -1365,7 +1385,9 @@ def update(self, instance, validated_data):
         if 'geolocation' in validated_data:
             self.update_geolocation(instance, validated_data.pop('geolocation'))
         if 'location_restriction' in validated_data:
-            self.update_location_restriction(instance, validated_data.pop('location_restriction'))
+            location_restriction_data = validated_data.pop('location_restriction')
+            if not all(value is None for value in location_restriction_data.values()):
```
DawoudSheraz: Can't value be the empty list for states/countries? The "value is []" is False and it can impact update_location_restriction call.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -2915,6 +2915,80 @@ def test_program_data(self):
...
+                'restriction_type': None,
+                'countries': ['CA'],
+                'states': ['MI'],
+            }
+        }
+        location_restriction = CourseLocationRestrictionFactory()
+        course = CourseFactory(location_restriction=location_restriction)
+        serializer = CourseSerializer(course, context={'request': request}, data=data)
+        serializer.is_valid()
+        assert 'location_restriction' in serializer.errors
```
DawoudSheraz, line 2946: Please verify that the correct error message is present for each case.

course_discovery/apps/api/serializers.py
```diff
@@ -1145,14 +1145,34 @@ def create(self, validated_data):
...
+    def check_has_data(self, key, attrs):
+        return key in attrs and bool(attrs[key])
+
+    def validate(self, attrs):
+        has_restriction_type = self.check_has_data('restriction_type', attrs)
+        has_countries = self.check_has_data('countries', attrs)
+        has_states = self.check_has_data('states', attrs)
+
+        if (has_countries or has_states) and not has_restriction_type:
+            raise serializers.ValidationError({'Restriction Type': _('Restriction Type cannot be empty')})
```
DawoudSheraz, line 1169: nit: Even though countries and states values' presence is checked before calling `update_location_restriction`, I will advocate for adding the presence of countries/states in validate function. This will be called whenever location serializer is used and is central in creating the correct location information.

azanbinzahid:
We can avoid a lot of changes if [this](https://edx-internal.slack.com/archives/C032DFXHV7W/p1661854989856629?thread_ts=1661809918.900169&cid=C032DFXHV7W) works. Any blockers in this approach?

course_discovery/apps/api/serializers.py
srwang: thanks @DawoudSheraz - I didn't quite understand what is being suggested. Could you write sample code to describe it? Thank you!

srwang:
> We can avoid a lot of changes if [this](https://edx-internal.slack.com/archives/C032DFXHV7W/p1661854989856629?thread_ts=1661809918.900169&cid=C032DFXHV7W) works. Any blockers in this approach?
@azanbinzahid 

I ended up popping the location_restriction item in the [update](https://github.com/openedx/course-discovery/blob/4afc8b66461825b269b0c6100edbe4437ea6430d/course_discovery/apps/api/serializers.py#L1378-L1393) method because it looks like that's where similar logic was already being handled. 

are you recommending take all of the items inside this method and move it to `update_course` on the view?

azanbinzahid:
@srwang  I was suggesting that we can pop it much earlier, even before the serializer validations are being checked - but I see you have already customized the validator so this approach is looking as well.

course_discovery/apps/api/serializers.py
DawoudSheraz: Hello. I noticed my above comment had a typo which made the sentence weird. What I was suggesting is that:
- if restriction type is present
- country information is missing

Since states' information will only be needed for US, the states' value checks can be ignored. In the above function, the absence of restriction type in presence of states/countries is checked. I am suggesting adding a case for the opposite scenario. This is a non-blocking change that can be added in a followup PR.

srwang: oh gotcha! we decided to allow for that functionality to make countries and states both blank. this is how someone would "remove" a location restriction via Publisher.

the reasoning was that it's hard to delete a record using publisher.  lmk how you feel.

DawoudSheraz (approved):

azanbinzahid (approved):

PR #3583, opened 2022-09-01
chore: adding loggin to org ent sub inclusion update cascade task

alex-sheehan-edx:
related ticket: https://2u-internal.atlassian.net/browse/ENT-6205

Context: 
We are noticing that when organizations are saved, the signal/task that are responsible for re-saving all the org's courses (in order to update the courses' subscription tagging to match the org's if the org's value has changed) are only affecting published courses under the org, while draft courses are never changed when an org is saved. To bring clarity to the situation, I've added logs to the task to see the number of courses fetched by the task. This should let us know if it is the task that is not fetching the draft courses or if there is an issue further down when saving said draft courses.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -9,6 +9,7 @@ import pytz
 import responses
 from django.contrib.auth.models import Group
 from django.db.models.functions import Lower
+from freezegun import freeze_time
 from rest_framework.reverse import reverse
 from rest_framework.test import APIRequestFactory
 
@@ -130,6 +131,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         assert response.data == \
             self.serialize_course_run(self.course_run, extra_context={'include_unpublished_programs': True})
 
+    @freeze_time("2022-01-14 12:00:01")
     @responses.activate
     def test_create_minimum(self):
         """ Verify the endpoint supports creating a course_run with the least info. """
@@ -267,6 +269,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         assert new_course_run.language == self.draft_course_run.language
         assert list(new_course_run.transcript_languages.all()) == list(self.draft_course_run.transcript_languages.all())
 
+    @freeze_time("2022-01-14 12:00:01")
     @ddt.data(True, False, "bogus")
     @responses.activate
     def test_create_draft_ignored(self, draft):
@@ -290,6 +293,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         self.assertDictEqual(response.data, self.serialize_course_run(new_course_run))
         assert new_course_run.draft
 
+    @freeze_time("2022-01-14 12:00:01")
     @responses.activate
     def test_create_using_type_with_price(self):
         """ Verify the endpoint supports creating a course_run and sets the seats price to the given price """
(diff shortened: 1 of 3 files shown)


ansabgillani (approved):

alex-sheehan-edx:
@ansabgillani, in talking with @DawoudSheraz I've changed our filtering of courses under orgs and course runs under courses to now use `.everything` instead of `.objects` as advised to pick up draft content as well as published content. Would love a quick rereview to make sure it makes sense!

ansabgillani:
This makes much more sense now 👍🏻

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -30,20 +30,29 @@ def update_org_program_and_courses_ent_sub_inclusion(org_pk, org_sub_inclusion):
...
-            ProgramType.XSERIES,
-            ProgramType.MICROMASTERS,
-            ProgramType.PROFESSIONAL_CERTIFICATE,
-            ProgramType.PROFESSIONAL_PROGRAM_WL,
-            ProgramType.MICROBACHELORS
-        ]
-    )
-    for program in programs:
-        program.save()
+    if courses:
```
uzairr: no need to enclose loop processing of `courses` and `programs` inside this `if` clause

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -30,20 +30,29 @@ def update_org_program_and_courses_ent_sub_inclusion(org_pk, org_sub_inclusion):
...
+        LOGGER.info(
+            f'Org: {org_pk} has been saved. Updating enterprise sub tagging logic for {len(courses)} courses.'
+        )
```
uzairr: logging may be rewritten as follows: 
`
sub_tag_log = "Org: {org_pk} has been saved. Updating enterprise sub tagging logic for {count} {product}"
`
`LOGGER.info(sub_tag_log.format(org_pk=org_pk, count=len(courses), product='courses'`
`LOGGER.info(sub_tag_log.format(org_pk=org_pk, count=len(programs), product='programs'`
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L297

uzairr (commented):
suggestions are added, it may reduce the change-set alongwith redundancy!

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: the intention was to avoid unnecessary logging but I supposed you're right!

course_discovery/apps/course_metadata/tasks.py
alex-sheehan-edx: thanks for the suggestion!

course_discovery/apps/course_metadata/models.py
```diff
@@ -1076,7 +1076,7 @@ def save(self, *args, **kwargs):
 
         # Course runs calculate enterprise subscription inclusion based off of their parent's status, so we need to
         # force a recalculation
-        course_runs = CourseRun.objects.filter(course=self)
+        course_runs = CourseRun.everything.filter(course=self)
```
DawoudSheraz: nit: Reverse lookup on course can be used to get thee associated runs.

alex-sheehan-edx: done!

uzairr (approved):
ensure the green build otherwise LGTM 👍, @DawoudSheraz please have a look on it whenever you are available before its merge!

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -9,6 +9,7 @@
 import responses
 from django.contrib.auth.models import Group
 from django.db.models.functions import Lower
+from freezegun import freeze_time
```
DawoudSheraz, line 12: out of curiosity, what is this doing?

DawoudSheraz (approved):
- Please update PR description and title since there are additions other than logging

PR #3585, opened 2022-09-02
fix: use different logging settings for k8s deployment

rgraber:
Creates a different environment file for use with a kubernetes deployment of course-discovery, and a new target in the Dockerfile that uses this new environment. The environment file uses different logging settings to account for a kubernetes deployment of course-discovery. In particular, this removes the local syslog handler, which the container does not have access to.

### DIFF
--- a/course_discovery/settings/kubernetes.py
+++ b/course_discovery/settings/kubernetes.py
@@ -1,0 +1,72 @@
+"""
+Specific overrides to the base prod settings for a docker production deployment.
+"""
+
+import platform
+import sys
+from .production import *  # pylint: disable=wildcard-import, unused-wildcard-import
+
+
+def get_docker_logger_config():
+    """
+    Return the appropriate logging config dictionary. You should assign the
+    result of this to the LOGGING var in your settings.
+    """
+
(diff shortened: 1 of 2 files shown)


christopappas (approved):

ansabgillani (approved):

course_discovery/settings/kubernetes.py
```diff
@@ -0,0 +1,73 @@
...
+            'django.request': {
+                'handlers': handlers,
+                'propagate': True,
+                'level': 'WARNING'
+            },
+            '': {
+                'handlers': handlers,
+                'level': 'DEBUG',
+                'propagate': False
+            },
```
DawoudSheraz, line 67: out of curiosity; why are different loggres needed?

DawoudSheraz (commented):
- Squash commits
- Add PR description

rgraber:
> * Squash commits
> 
>     * Add PR description

My git got into a weird state locally and I'm having trouble squashing commits, is it ok if I just squash on merge?

DawoudSheraz (approved):

PR #3588, opened 2022-09-06
feat: Update course information from signal.

dianakhuang:
As part of the event bus work, an event for course information changing
will trigger a signal to be sent. This adds a receiver for the signal
that will then update the course information and refactors existing data loader code to accommodate the changes.

https://github.com/edx/edx-arch-experiments/issues/36

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/__init__.py
+++ b/course_discovery/apps/course_metadata/data_loaders/__init__.py
@@ -18,18 +18,22 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
     LOADER_MAX_RETRY = 3
     PAGE_SIZE = 50
 
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False):
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, enable_api=True):
         """
         Arguments:
             partner (Partner): Partner which owns the APIs and data being loaded
             api_url (str): URL of the API from which data is loaded
             max_workers (int): Number of worker threads to use when traversing paginated responses.
             is_threadsafe (bool): True if multiple threads can be used to write data.
+            enable_api (bool): True if we want to use the api functionalities and clients with the dataloader.
+                This will most likely only be turned off for event bus use cases.
         """
         self.partner = partner
-        self.api_url = api_url.strip('/') if api_url else api_url
-        self.api_client = self.partner.oauth_api_client
-        self.username = self.get_username_from_client(self.api_client)
+        self.enable_api = enable_api
+        if self.enable_api:
+            self.api_url = api_url.strip('/') if api_url else api_url
+            self.api_client = self.partner.oauth_api_client
+            self.username = self.get_username_from_client(self.api_client)
 
         self.max_workers = max_workers
         self.is_threadsafe = is_threadsafe
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -3,14 +3,18 @@ import logging
 import waffle  # lint-amnesty, pylint: disable=invalid-django-waffle-import
 from celery import shared_task
 from django.apps import apps
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django.db.models.signals import m2m_changed, post_delete, post_save, pre_delete, pre_save
 from django.dispatch import receiver
+from openedx_events.content_authoring.data import CourseCatalogData
+from openedx_events.content_authoring.signals import COURSE_CATALOG_INFO_CHANGED
 
 from course_discovery.apps.api.cache import api_change_receiver
 from course_discovery.apps.core.models import Partner
 from course_discovery.apps.course_metadata.constants import MASTERS_PROGRAM_TYPE_SLUG
+from course_discovery.apps.course_metadata.data_loaders.api import CoursesApiDataLoader
 from course_discovery.apps.course_metadata.models import (
     Course, CourseRun, Curriculum, CurriculumCourseMembership, CurriculumProgramMembership, Organization, Program
 )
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -717,3 +723,33 @@ def test_authoring_organizations_changed(self):
 
             course.authoring_organizations.add(organization)
             mock_salesforce_util().create_course.assert_called()
+
+
+class TestCourseDataUpdateSignal(TestCase):
+    def setUp(self):
```
rgraber, line 727: I think we could use some additional tests:
1. Processing the same message 2x in a row. We're stressing idempotency pretty hard so I think it'd be worthwhile to test it.
2. What happens when the various optional fields of CourseCatalogData are None
3. What happens when publisher is/not enabled (although I'm kind of ambivalent on this one since theoretically that should already be tested in any tests of _process_response)
4. Creating a new course from a signal
5. Error case (we can just check that we catch any exceptions and log them)
This might mean a little bit of overlapping with existing data loader tests but I do think it's worth making sure we can do all these things from a signal in addition to the old version

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -717,3 +723,33 @@ def test_authoring_organizations_changed(self):
...
+            effort='a lot',
+            hidden=False,
+        )
+        self.partner = PartnerFactory()
+
+    def test_event_handled(self):
+        update_course_data_from_event(catalog_info=self.catalog_data)
+        course_run = CourseRun.objects.get(key=self.course_key)
+
+        assert course_run.title == self.catalog_data.name
```
rgraber, line 750: Why are we only checking these fields (and not short_description or pacing, for example)?

course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -18,18 +18,20 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
...
 
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False):
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, enable_api=True):
         """
         Arguments:
             partner (Partner): Partner which owns the APIs and data being loaded
             api_url (str): URL of the API from which data is loaded
             max_workers (int): Number of worker threads to use when traversing paginated responses.
             is_threadsafe (bool): True if multiple threads can be used to write data.
+            enable_api (bool): True if we want to use the api functionalities and clients with the dataloader.
```
rgraber, line 28: Since this is kind of an odd case, maybe we should specify that this should probably only be used for things like the event bus.

course_discovery/apps/course_metadata/tests/test_signals.py
dianakhuang: Sounds good. I was definitely wrestling with this one test long enough that my mind was blanking on what other tests I could do. 😂

course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -18,18 +18,22 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
...
+                This will most likely only be used with event bus use cases.
```
rgraber: [nit]
```suggestion
                This will most likely only be turned off for event bus use cases.
```

course_discovery/apps/course_metadata/signals.py
```diff
@@ -327,12 +331,48 @@ def check_course_runs_within_course_for_duplicate_external_key(course, specific_
...
+        kwargs: event data sent to signal
+    """
+    course_data = kwargs.get('catalog_info', None)
+    if not course_data or not isinstance(course_data, CourseCatalogData):
+        logger.error('Received null or incorrect data from COURSE_CATALOG_INFO_CHANGED.')
+        return
+
+    # Handle optional fields.
+    schedule_data = course_data.schedule_data
+    end = str(schedule_data.end) if schedule_data.end else None
```
rgraber, line 350: [question] Do we need to specify that it's an isoformat string?

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -717,3 +721,108 @@ def test_authoring_organizations_changed(self):
...
+            schedule_data=scheduling_data,
+        )
+        update_course_data_from_event(catalog_info=catalog_data)
+        course_run = CourseRun.objects.get(key=self.course_key)
+
+        assert course_run.title == self.catalog_data.name
+        assert course_run.start == self.catalog_data.schedule_data.start
+        assert course_run.pacing_type == 'self_paced'
+
+    def test_update_existing_course(self):
```
rgraber, line 803: [question/request] What is the expected behavior when we update a course that has a start date from an event that didn't have a start date (or other optional field)? Can we test that?

rgraber (commented):
Couple more test thoughts

course_discovery/apps/course_metadata/signals.py
dianakhuang: The string that's generated from this code, you mean?

course_discovery/apps/course_metadata/tests/test_signals.py
dianakhuang: I am assuming that there would be no update to the existing values, but I will add a test for that now

dianakhuang: It seems like this is not what happens, so I guess we should talk about this at parking lot.

course_discovery/apps/course_metadata/signals.py
dianakhuang: Clarification that happened synchronously: any string that can be handled by `dateutil.parser` is valid, so it's fine to leave it.

course_discovery/apps/course_metadata/tests/test_signals.py
dianakhuang: We are okay with the default behavior where if the optional fields are blank, they get cleared on the course. I have added a test to display this behavior.

rgraber (approved):
LGTM!

course_discovery/apps/course_metadata/signals.py
```diff
@@ -327,12 +331,48 @@ def check_course_runs_within_course_for_duplicate_external_key(course, specific_
             raise ValidationError(message)
 
 
+@receiver(COURSE_CATALOG_INFO_CHANGED)
```
Ali-D-Akbar, line 334: Where will this event be triggered?

Ali-D-Akbar (approved):
One nit otherwise LGTM.

course_discovery/apps/course_metadata/signals.py
dianakhuang: This event will be triggered from the `consume_events` management command: https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/consume_events.py

The consumer will be set to listen for events from the event bus that match `COURSE_CATALOG_INFO_CHANGED`, and then emit the matching signal.

dianakhuang: The event that ends up on the event bus will be sent from Studio when course information changes there: https://github.com/openedx/edx-platform/blob/322c22a4eb396347e8c521c0f8a54dd2b35cb37c/cms/djangoapps/contentstore/signals/handlers.py#L154-L162

PR #3592, opened 2022-09-08
feat: add gspread client to communicate with google sheets and images

AliAdnanSohail:
Ticket: [Add utility to read data from google sheets via gspread](https://2u-internal.atlassian.net/browse/PROD-2922)

## Description:
To automate our ingestion process, we need to communicate with google APIs to get data of google drive, So in this PR a new package [Gspread](https://docs.gspread.org/en/v5.4.0/) is installed and used to communicate with google sheets using their link and tab_id.
Note: test cases will be further improved in the next ticket

### DIFF
--- a/course_discovery/apps/course_metadata/gspread_client.py
+++ b/course_discovery/apps/course_metadata/gspread_client.py
@@ -1,0 +1,44 @@
+import logging
+
+import gspread
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+class GspreadClient:
+    """
+    API Client for GSpread to communicate with google spread sheets and drive images
+    """
+
+    def __init__(self):
+        try:
+            self.client = gspread.service_account_from_dict(settings.GOOGLE_SERVICE_ACCOUNT_CREDENTIALS)
+            logger.info('[Connection Successful]: Successful connection with google service account')
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.exception(f'[Connection Failed]: Failed to connect with google service account error_message: {ex}')
+
+    def get_spread_sheet_by_url(self, url):
+        try:
+            spread_sheet = self.client.open_by_url(url)
+            logger.info('[Spread Sheet Found]: Opening google sheet')
+            return spread_sheet
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.exception(f'[Spread Sheet Not Found]: No spreadsheet found for url: {url} error_message: {ex}')
+        return None
+
+    @staticmethod
+    def get_worksheet_data_by_tab_id(spread_sheet, tab_id):
+        try:
+            worksheet_title = [ws.title for ws in spread_sheet.worksheets() if ws.id == tab_id]
+            if not worksheet_title:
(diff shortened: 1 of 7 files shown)


azanbinzahid:
Can we add a test call in one of the import metadata commands so that we get the log in jenkins?

AliAdnanSohail:
@azanbinzahid why we need test call? I think it can create confusion, If it get merged then we can start using it in management commands. 
Or do you want to test connection in management commands before using it?

course_discovery/apps/course_metadata/gspread_client.py
```diff
@@ -0,0 +1,44 @@
...
+            logger.info('[Spread Sheet Found]: Opening google sheet')
+            return spread_sheet
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.error(f'[Spread Sheet Not Found]: No spreadsheet found for url: {url} error_message: {ex}')
+        return None
+
+    @staticmethod
+    def get_worksheet_data_by_tab_id(spread_sheet, tab_id):
+        try:
+            worksheet_title = [ws.title for ws in spread_sheet.worksheets() if ws.id == tab_id]
```
uzairr, line 33: any specific reason to create a list here if we are looking for singular worksheet ?

AliAdnanSohail: Actually `spread_sheet.worksheets()` returns a list of sheets with id and title, so here we need to filter and get the title of the sheet matching the id.

DawoudSheraz (commented):
I second the comment by Azan that there should be a mechanism to verify GSpread and service account setup. That can be a temporary change.

uzairr (approved):

azanbinzahid:
> @azanbinzahid why we need test call? I think it can create confusion, If it get merged then we can start using it in management commands. Or do you want to test connection in management commands before using it?

May be just add a call [here](https://github.com/openedx/course-discovery/blob/f94e85149872c75c365b3d13035a79502728631d/course_discovery/apps/course_metadata/management/commands/import_degree_data.py#L60) (and log) so that we can see (via jenkins) if its able to connect with sheets.

course_discovery/apps/course_metadata/gspread_client.py
```diff
@@ -0,0 +1,44 @@
...
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.error(f'[Connection Failed]: Failed to connect with google service account error_message: {ex}')
+
+    def get_spread_sheet_by_url(self, url):
+        try:
+            spread_sheet = self.client.open_by_url(url)
+            logger.info('[Spread Sheet Found]: Opening google sheet')
+            return spread_sheet
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.error(f'[Spread Sheet Not Found]: No spreadsheet found for url: {url} error_message: {ex}')
```
DawoudSheraz: nit: logger.exception would be enough as it logs the exception traceback automatically.

DawoudSheraz (approved):

azanbinzahid (approved):

PR #3594, opened 2022-09-09
fix: downgrade edx-drf-extension to fix disable user issue

DawoudSheraz:
### Description
- Downgrade edx-drf-extensions to 8.2.0 to fix disabled user issues introduced in https://github.com/openedx/edx-drf-extensions/pull/268/files


azanbinzahid (approved):
Once the build is green.

PR #3595, opened 2022-09-09
chore: update edx-event-bus-kafka

rgraber:
Upgrade the edx-event-bus-kafka library and import the correct classes. Also, only check that confluent-kafka is correctly installed since edx-event-bus-kafka is now in the normal requirements.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/consume_events.py
+++ b/course_discovery/apps/course_metadata/management/commands/consume_events.py
@@ -2,7 +2,7 @@ import logging
 
 from django.core.management.base import BaseCommand
 from edx_event_bus_kafka import KafkaEventConsumer
-from edx_toggles.toggles import SettingToggle  #
+from edx_toggles.toggles import SettingToggle
 from openedx_events.tooling import OpenEdxPublicSignal
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


dianakhuang (approved):

course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -48,15 +51,9 @@ def add_arguments(self, parser):
...
     def handle(self, *args, **options):  # pragma: no cover
         try:
-            # FIXME: This needs to be replaced with a configurable loading
-            # mechanism built into openedx-events.
-            # See https://github.com/openedx/openedx-events/pull/85)
-            from edx_event_bus_kafka.consumer.event_consumer import \
-                KafkaEventConsumer  # pylint: disable=import-outside-toplevel
-            from edx_toggles.toggles import SettingToggle  # pylint: disable=import-outside-toplevel
-            from openedx_events.tooling import OpenEdxPublicSignal  # pylint: disable=import-outside-toplevel
+            import confluent_kafka  # pylint: disable=import-outside-toplevel,unused-import
```
Ali-D-Akbar, line 54: where does this `confluent_kafka` come from? Is this installed somewhere along with the edx event bus kafta?

Ali-D-Akbar (approved):
Just a question otherwise looks good.

course_discovery/apps/course_metadata/management/commands/consume_events.py
rgraber: This is installed in the private dockerfile. See https://github.com/edx/internal-dockerfiles/pull/64 for the whole story

Ali-D-Akbar: Ohh, right so basically it's already installed in the image already. Got it.

DawoudSheraz (approved):

PR #3604, opened 2022-09-14
feat: add settings variable for service account key

azanbinzahid:
[PROD-2962](https://2u-internal.atlassian.net/browse/PROD-2962)
re-attempting for #3593

### DIFF
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -638,3 +638,6 @@ FIRE_UPDATE_PROGRAM_SKILLS_SIGNAL = False
 DISCOVERY_BASE_URL = "http://localhost:18381"
 
 PRODUCT_API_URL = ''
+
+# Required for communicating with Google Service Account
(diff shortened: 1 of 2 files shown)


DawoudSheraz (approved):

AliAdnanSohail (approved):

course_discovery/settings/base.py
```diff
@@ -638,3 +638,6 @@
 DISCOVERY_BASE_URL = "http://localhost:18381"
 
 PRODUCT_API_URL = ''
+
+# Required for communicating with Google Service Account
+GOOGLE_SERVICE_ACCOUNT_CREDENTIALS = '{"key":"value"}'
```
AliAdnanSohail:
```suggestion
GOOGLE_SERVICE_ACCOUNT_CREDENTIALS = {"key":"value"}
```

In production.py we are getting value from vault so we are parsing fine in production.py but locally we are expecting value as a dict. So we can replace this variable to dict value

PR #3630, opened 2022-09-30
feat: Modified POST handler to take params from body of POST

irfanuddinahmad:
**Description:** Modified POST handler to take params from body of POST, if present, otherwise use the existing flow with GET query params

**JIRA:** [ENT-6295](https://2u-internal.atlassian.net/browse/ENT-6295)

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes/serializers/course.py
+++ b/course_discovery/apps/course_metadata/search_indexes/serializers/course.py
@@ -68,6 +68,9 @@ class CourseSearchDocumentSerializer(ModelObjectDocumentSerializerMixin, DateTim
         exclude_expired = request.GET.get('exclude_expired_course_run')
         detail_fields = request.GET.get('detail_fields')
         query_params = dict(request.query_params)
+        if request.method == 'POST':
+            exclude_expired = request.POST.get('exclude_expired_course_run') or exclude_expired
+            detail_fields = request.POST.get('detail_fields') or detail_fields
 
         def should_include_course_run(course_run, params, exclude_expired):
             matches_parameter = False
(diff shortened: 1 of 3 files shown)


bilaltahir21 (approved):

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -68,6 +68,11 @@ def get_course_runs(self, result):
         exclude_expired = request.GET.get('exclude_expired_course_run')
         detail_fields = request.GET.get('detail_fields')
         query_params = dict(request.query_params)
+        if request.method == 'POST':
+            if request.POST.get('exclude_expired_course_run'):
```
bilaltahir21: Can we remove these `if` statements because `get()` would automatically return `None` in case a field is not there.

irfanuddinahmad: can you elaborate?

bilaltahir21: If we remove the statements `if request.POST.get('exclude_expired_course_run'):` and `if request.POST.get('detail_fields'):` inside `if request.method == 'POST':` then I think the result would be the same.

irfanuddinahmad: The flow is such that we use the POST values, if they are present, otherwise the GET ones remain assigned to the variables.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -68,6 +68,11 @@ def get_course_runs(self, result):
...
+                exclude_expired = request.POST.get('exclude_expired_course_run')
+            if request.POST.get('detail_fields'):
```
DawoudSheraz: you can do something along the lines of `exclude_expired = request.POST.get('exclude_expired_course_run') or exclude_expired`. Please add relevant unit tests for it.

DawoudSheraz (approved):

PR #3631, opened 2022-09-30
feat: course_length field in Course model to store actual experienced duration of course

bilaltahir21:
## Description 
This PR adds a new field `course_length` in the `Course` model in the app [course_metadata](https://github.com/openedx/course-discovery/blob/86c194dc342f5b82daf70eacda3d38a9cc180208/course_discovery/apps/course_metadata/models.py#L939). This new field helps to store the actual duration of a course as experienced by learners who complete that course. 

## Context
This is an effort to run an A/B experiment in [frontend-app-learner-portal-enterprise](https://github.com/openedx/frontend-app-learner-portal-enterprise) where we would add tags on some of the course cards saying something like "Short Course". This experiment would measure the effect of this new tag on learners viewing the course about page. This field would be populated for each course using a regular interval management command and we would not be giving an option to edit it in publisher. 

How do we calculate the value of this field? This is calculated in this [management command](https://github.com/openedx/course-discovery/pull/3637) by querying and calculating the time it took for learners to complete a course and calculating the thresholds upon which we categorise these courses. 

NOTE: We would add this field in the Algolia index in another PR which will be used on the frontend mentioned above. 

JIRA: [ENT-6266](https://2u-internal.atlassian.net/browse/ENT-6266)


course_discovery/apps/course_metadata/models.py
```diff
@@ -962,6 +962,10 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
         related_name='related_courses',
         default=None, null=True, blank=True
     )
+    course_length = models.CharField(
+        max_length=64, choices=CourseLength, blank=True,
+        help_text=_('The actual duration of this course as experienced by learners who complete it.'),
```
uzairr, line 980: `The actual duration of this course...`
Just a nit! otherwise LGTM

uzairr, line 967: Just a nit! otherwise LGTM

uzairr (approved):

course_discovery/apps/course_metadata/models.py
DawoudSheraz, line 980: I have a couple of questions on this because the usage and intent are not very clear.

- What is the purpose of the field?
- How/When will it be populated in a course's life cycle?
- What impact does it bring that distinguishes it from CourseRun's duration/length field?

DawoudSheraz (commented):
left some questions as inline comments.

course_discovery/apps/course_metadata/models.py
bilaltahir21: The answers are in the below points

- The purpose of this field is to store a string describing course length
- Currently, it would be populated using a management command which categorises all courses according to their length and the time it took for learners to complete those courses
- This string would be shown directly on the course cards instead of the duration fields in course runs (currently an experiment if it would increase the users viewing course about page)

bilaltahir21: Can you elaborate some more? I think this would emphasise that this is the duration experienced by learners.

DawoudSheraz: Alright, some followups:

- Why is there a need to create a new field for this? Why can't the existing duration field be transformed to a string during page build?
- Who decides the value of this duration? When will it be populated in a course's life cycle (not talking about the method of setting the value here)?

bilaltahir21: This ticket might be helpful to answer these questions [ENT-6237](https://2u-internal.atlassian.net/browse/ENT-6237).

bilaltahir21:
@uzairr @DawoudSheraz can we merge this PR?

bilaltahir21:
[reminder] @uzairr @DawoudSheraz 
Can we merge this PR today? It is a pre-requisite for another work in [this PR](https://github.com/openedx/course-discovery/pull/3637) ([ENT-6267](https://2u-internal.atlassian.net/browse/ENT-6267)).

course_discovery/apps/course_metadata/models.py
ansabgillani: @bilaltahir21 can you add all of this context to the request description? Almost missed the point of the whole PR.

ansabgillani:
@bilaltahir21 We have a lot of course data coming in from various edX partners that we either ingest via loaders. Even though the PR looks good, do you think adding this field will affect any of the data we are adding? And if so, do we need to cater to those loaders as well? 
Thanks.

course_discovery/apps/course_metadata/models.py
```diff
@@ -975,6 +975,10 @@ class Course(DraftModelMixin, PkSearchableMixin, CachedMixin, TimeStampedModel):
...
```
azanbinzahid, line 980: I don't see any migration errors but what would be the default value? I get the context from the ticket -- will it stay null/blank in start and we only populate as needed?

Also, I don't see any change in API/serializer -- we ain't expecting any change from publisher or APIs in longer run?

How course staff will perceive it? or it won't be set manually at all.

course_discovery/apps/course_metadata/models.py
bilaltahir21: @ansabgillani I have added the context in the description. I am working on other reviews on this PR. Meanwhile, please let me know if there is anything else that we need to cater.

course_discovery/apps/course_metadata/models.py
bilaltahir21: I have confirmed that `blank=True` would be the right option for the default value of this field because then Django would not enforce filling out this field in its forms. 

We would not be adding it to the publisher for changing it on frontend because it would be populated using the management command. But we will add it to the Algolia Index. 

It would not be set manually at all.

bilaltahir21:
@ansabgillani I have discussed it with @azanbinzahid and we think there is no need to change data loaders.

ansabgillani (approved):

course_discovery/apps/course_metadata/models.py
azanbinzahid: Will you create a follow-up PR for algolia or make a commit in this one? @bilaltahir21

bilaltahir21: That will be a follow-up PR.

PR #3632, opened 2022-09-30
docs: Add ADR for event bus use case.

dianakhuang:
This adds an ADR to explain how we are using the event bus along with refresh_course_metadata for our first event bus use case.

https://github.com/edx/edx-arch-experiments/issues/61

### DIFF
--- a/docs/decisions/0015-event-bus-with-rcm.rst
+++ b/docs/decisions/0015-event-bus-with-rcm.rst
@@ -1,0 +1,40 @@
+15. Use Event Bus to Replace Refresh Course Metadata
+----------------------------------------------------
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+In an effort to update and progress our overall system architecture, we have `made the decision to use an event bus`_ for communicating between services. Discovery
+is a service that synchronizes data between itself and several other services, relying on a management command run on Jenkins, ``refresh_course_metadata``, that makes
+REST api requests to both ecommerce and LMS. This work is being done by a non-owning team of discovery to prove out a use case for the event bus by utilizing an
+already existing pattern.
+
+.. _made the decision to use an event bus: https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0052-arch-event-bus-architecture.html
+
+Decision
+--------
+
+We will be duplicating an existing ``refresh_course_metadata`` workflow with a event that gets sent to and received from the event bus, replicating course information
+published in Studio. While we would like to see all of ``refresh_course_metadata`` replaced with events, that work is outside the scope of this ADR.
+
+Consequences
+------------
+
+- Changes in Studio will be reflected on Discovery in the Discovery database in a much more timely manner instead of being updated once a day with the run of RCM.
+- The current implementation of this work with event bus will not replace, merely duplicate, the work done by ``refresh_course_metadata``. We are not removing the
+  RCM course update code at this time, and it will act as a backstop for the event workflow until we are certain of event bus stability. Future work to remove
+  ``refresh_course_metadata`` will have to be more careful without the backstop.
+- This may be a small part of a larger effort to replace all of RCM with events, but this ADR only covers the Studio use case.
+- We may see race conditions between events, and we will be relying on workarounds (making a new change to issue a new event) or another RCM run to fix the issue.
+- An event consumer will be running on new infrastructure and will have to be maintained separately from the Discovery application deployment. See the `Managing Kafka Consumers ADR`_
+  for more details.
+- The event consumer will convert any instances of an event it receives into `a Django signal`_ that will indicate that course data has been updated.
+  While we are implementing only one listener for this signal at the moment, other parts of the system could make use of it in the future.
+
+.. _a Django signal: https://github.com/openedx/openedx-events/blob/7620775586f2746c77ffb391162094de901fb4b0/openedx_events/content_authoring/signals.py#L18
+.. _Managing Kafka Consumers ADR: https://github.com/openedx/event-bus-kafka/blob/main/docs/decisions/0003-managing-kafka-consumers.rst


docs/decisions/0015-event-bus-with-rcm.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Consequences
+------------
+
+- Changes in Studio will be reflected on Discovery in the Discovery database in a much more timely manner instead of being updated once a day with the run of RCM.
+- The current implementation of this work with event bus will not replace, merely duplicate, the work done by ``refresh_course_metadata``. We are not removing the
+  RCM course update code at this time.
+- This may be a small part of a larger effort to replace all of RCM with events, but this ADR only covers the Studio use case.
+- We may see race conditions between events, and we will be relying on workarounds (making a new change to issue a new event) or another RCM run to fix the issue.
+- A new Django signal will be sent to indicate course data has been updated. While we are implementing only one receiver of this signal at the moment, other parts of
```
robrap: 1. I'm not clear on this last point. Are you talking about the event Production, or event Consumption, or both?
2. Related, we could mention the new event consumer to handle these events, and point to the event consumer ADR (which you should find in the OEP).

docs/decisions/0015-event-bus-with-rcm.rst
```diff
@@ -0,0 +1,33 @@
...
+--------
+
+We will be duplicating existing ``refresh_course_metadata`` workflow with events that get sent to and received from the event bus, starting with course information
+published in Studio.
+
+Consequences
+------------
+
+- Changes in Studio will be reflected on Discovery in the Discovery database in a much more timely manner instead of being updated once a day with the run of RCM.
+- The current implementation of this work with event bus will not replace, merely duplicate, the work done by ``refresh_course_metadata``. We are not removing the
```
robrap, line 29: You might add/clarify that RCM will continue to provide a backup to the events, and that additional attention may be required in this area when and if the events no longer have this backup.

docs/decisions/0015-event-bus-with-rcm.rst
```diff
@@ -0,0 +1,33 @@
...
+In an effort to update and progress our overall system architecture, we have `made the decision to use an event bus`_ for communicating between services. Discovery
+is a service that synchronizes data between itself and several other services, relying on a management command run on Jenkins, ``refresh_course_metadata``, that makes
+REST api requests to both ecommerce and LMS.
+
+.. _made the decision to use an event bus: https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0052-arch-event-bus-architecture.html
+
+Decision
+--------
+
+We will be duplicating existing ``refresh_course_metadata`` workflow with events that get sent to and received from the event bus, starting with course information
```
robrap: 1. Under context, I'd clarify that the owning team of the initial event bus work is different from the owning team of Discovery.
2. I'd either scope this decision to proving out the event bus to get a benefit for Discovery without making any commitment to replacing the rest of RCM. Or, clarifying that there is no commitment to the completion of this work, although it is intended direction.

robrap (commented):
When you get this settled, maybe mention it to Simon, since he was asking for this context and it might be better to get a review from outside our team.

docs/decisions/0015-event-bus-with-rcm.rst
robrap: [I mistakenly added this comment outside of a review.]

docs/decisions/0015-event-bus-with-rcm.rst
```diff
@@ -0,0 +1,39 @@
...
+Consequences
+------------
+
+- Changes in Studio will be reflected on Discovery in the Discovery database in a much more timely manner instead of being updated once a day with the run of RCM.
+- The current implementation of this work with event bus will not replace, merely duplicate, the work done by ``refresh_course_metadata``. We are not removing the
+  RCM course update code at this time, and it will act as a backstop for the event workflow until we are certain of event bus stability. Future work to remove
+  ``refresh_course_metadata`` will have to be more careful without the backstop.
+- This may be a small part of a larger effort to replace all of RCM with events, but this ADR only covers the Studio use case.
+- We may see race conditions between events, and we will be relying on workarounds (making a new change to issue a new event) or another RCM run to fix the issue.
+- A new Django signal will be present in Discovery to indicate course data has been updated. While we are implementing only one receiver of this signal at the
```
robrap: * Can we move this after the next bullet, and maybe update to read something like: "The event consumer will convert the new event into a Django signal...". The rest of the sentence may change, depending on how you ultimately want to phrase this.
* We could also add a link to https://github.com/openedx/openedx-events/blob/7620775586f2746c77ffb391162094de901fb4b0/openedx_events/content_authoring/signals.py#L18 since it is already implemented.

robrap (commented):
Looks good. After this minor adjustment, can we let Simon give a quick pass to see if this would have answered his questions? Thanks!

robrap (approved):

schenedx (commented):
Reviewed and now I understand this much better. I am surprised that this jenkins job is still running.

docs/decisions/0015-event-bus-with-rcm.rst
```diff
@@ -0,0 +1,40 @@
...
+- The current implementation of this work with event bus will not replace, merely duplicate, the work done by ``refresh_course_metadata``. We are not removing the
+  RCM course update code at this time, and it will act as a backstop for the event workflow until we are certain of event bus stability. Future work to remove
+  ``refresh_course_metadata`` will have to be more careful without the backstop.
+- This may be a small part of a larger effort to replace all of RCM with events, but this ADR only covers the Studio use case.
+- We may see race conditions between events, and we will be relying on workarounds (making a new change to issue a new event) or another RCM run to fix the issue.
+- An event consumer will be running on new infrastructure and will have to be maintained separately from the Discovery application deployment. See the `Managing Kafka Consumers ADR`_
+  for more details.
+- The event consumer will convert any instances of an event it receives into `a Django signal`_ that will indicate that course data has been updated.
```
DawoudSheraz, line 36: nit: The _ after Django Signal is appearing as it is in the preview and not applying markdown rendering.

DawoudSheraz (approved):
one doc nit to address before merge

PR #3635, opened 2022-10-04
feat: add course title and short_description field text for skill tagging

muhammad-ammar:
**JIRA:** https://2u-internal.atlassian.net/browse/ENT-6293

**Description:** We currently only use full_description field of a course for skill tagging but we believe that including course `title` and `short_description` fields can generate and improve course skill tagging and cover more use cases.

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -40,7 +40,7 @@ from course_discovery.apps.publisher.utils import is_publisher_user
 
 logger = logging.getLogger(__name__)
 
-COURSE_FIELDS_FOR_SKILLS = ['full_description']
+COURSE_FIELDS_FOR_SKILLS = ['title', 'short_description', 'full_description']
 
 
 def writable_request_wrapper(method):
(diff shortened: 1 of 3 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -40,7 +40,7 @@
 
 logger = logging.getLogger(__name__)
 
-COURSE_FIELDS_FOR_SKILLS = ['full_description']
+COURSE_FIELDS_FOR_SKILLS = ['title', 'short_description', 'full_description']
```
DawoudSheraz, line 43: Shouldn't CourseMetaProvider be updated to cater this?

muhammad-ammar: I don't think we need any changes in `DiscoveryCourseMetadataProvider`. It already extracts all the relevant course fields that we need. `COURSE_FIELDS_FOR_SKILLS` is only used when a course author updates any of these fields and then we fire a Django signal and after that a celery task jump in to extract the skills. 
Am I missing something here?

DawoudSheraz (approved):

muhammad-ammar:
@DawoudSheraz Can you please rerun the build?

PR #3642, opened 2022-10-12
fix: country codes capitalization in geotargeting loader

aheykamp2u:
When we initially created the data loader, we were creating the country codes in lowercase. While this doesn't effect performance in prospectus, it does make it so that the location restrictions do not show up in discovery's django admin. The checkbox for the country codes will only show as checked if the country code is capitalized. This PR makes it so that those country codes are entered into the database capitalized.


marekwrobel (approved):

DawoudSheraz (approved):
- Add PR description
- Fix the failing build -- incorrect commit type. It should be `fix`.

PR #3643, opened 2022-10-13
feat: add new title override.

zxcs001:
Adding new course metadata that allows people to override title pronouns, instead of "course", we can use another term such as "Programme", "Program", "Short Course" etc.

https://2u-internal.atlassian.net/browse/WS-3010


DawoudSheraz (commented):
Copying comments/questions from internal
- This is only required for 2U product lines, correct?
- The field name course_title_override seems to be misleading. A better name can be thought of to convey that this is overriding Course” term and not the complete title
- Will this field be edited on a need basis? Or do you intend to add it to CSV loader flow

zxcs001:
> Copying comments/questions from internal
> 
> * This is only required for 2U product lines, correct?
> * The field name course_title_override seems to be misleading. A better name can be thought of to convey that this is overriding Course” term and not the complete title
> * Will this field be edited on a need basis? Or do you intend to add it to CSV loader flow

* 1. This one, currently we will only apply for 2U product lines.
* 2. I will change the name
* 3. This field should be editable and only apply to a few specific courses, so it won't be added to the CSV loader flow.

DawoudSheraz (approved):
- Please squash the commits

uzairr (approved):

PR #3652, opened 2022-10-21
[WS-3261] add geolocation csv loader

danielkalen:
[WS-3261](https://2u-internal.atlassian.net/browse/WS-3261) requests that a CSV loader to be implemented (similar to previous loaders) in order to quickly load geolocation data in bulk. While working on this ticket Steve Tiszenkel had expressed how the current implementation of geolocation is not really allowing these coords to be easily reused, because right now the dropdown to select geolocation displays the actual coords in the selection dropdown which means to reuse coords the manager of the program/course needs to know his/her exact coords and visually match it in the list, instead of simply typing something like "Harvard University" and using its coords. Therefore, a new location_name field was added to the GeoLocation model.


course_discovery/apps/course_metadata/models.py
```diff
@@ -905,6 +905,12 @@ class GeoLocation(TimeStampedModel):
     LAT_MAX_DIGITS = 9
     LNG_MAX_DIGITS = 10
 
+    location_name = models.CharField(
```
DawoudSheraz, line 978: How it is going to be populated? Manually or via some API call to a 3rd party service to get name against coords?

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,210 @@
...
+
+import unicodecsv
+
+from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoader
+from course_discovery.apps.course_metadata.models import Course, GeoLocation, Program
+
+logger = logging.getLogger(__name__)
+
+
+class GeolocationCSVDataLoader(AbstractDataLoader):
```
DawoudSheraz, line 16: Will this be run one-time only?

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,210 @@
...
+        Get an object from the database by its key and value
+        """
+        model_name = model._meta.object_name
+        kwrags = {key: value}
+        try:
+            return (
+                model.everything.filter(**kwrags).exists() if model_name == 'Course'
+                else model.objects.filter(**kwrags).exists()
+            )
+        except model.DoesNotExist:
```
DawoudSheraz: filter() does not raise this exception. The return of filter is QuerySet and it is empty if the objects are not located.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,210 @@
...
+            return False
+    
+    def check_for_potential_orphans(self, model):
+        for geolocation_id in self.updated_geolocations:
+            related_product = model.everything.filter(geolocation__id=geolocation_id).first()
```
DawoudSheraz: Program model does not have draft/non-draft differences and there is no everything ObjectManager on it.

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -217,6 +217,54 @@ def _write_csv(self, csv, lines_dict_list, headers=None):
         return csv
 
 
+class GeolocationCSVLoaderMixin:
```
DawoudSheraz: This Mixin is same as GeotargetingCSVLoaderMixin, with the only difference being CSV_DATA_KEYS_ORDER. You can inherit that mixin and override the data keys list.

course_discovery/apps/course_metadata/data_loaders/tests/test_geolocation_csv_loader.py
```diff
@@ -0,0 +1,252 @@
...
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
+from course_discovery.apps.course_metadata.data_loaders.geolocation_loader import GeolocationCSVDataLoader
+from course_discovery.apps.course_metadata.data_loaders.tests import mock_data
+from course_discovery.apps.course_metadata.data_loaders.tests.mixins import GeolocationCSVLoaderMixin
+from course_discovery.apps.course_metadata.models import GeoLocation
+
+LOGGER_PATH = 'course_discovery.apps.course_metadata.data_loaders.geolocation_loader'
+INITIATED_LOG_MESSAGE = (LOGGER_PATH, 'INFO', 'Initiating Geolocation CSV data loader flow.')
+COMPLETED_LOG_MESSAGE = (LOGGER_PATH, 'INFO', 'Geolocation CSV loader ingest pipeline has completed.')
+SKIPPED_ITEMS_LOG_MESSAGE = (LOGGER_PATH, 'INFO', 'Skipped items:')
```
DawoudSheraz: nit: These can be moved to TestSuite's class-level variables.

course_discovery/apps/course_metadata/data_loaders/tests/test_geolocation_csv_loader.py
```diff
@@ -0,0 +1,252 @@
...
+                    ),
+                    (
+                        LOGGER_PATH,
+                        'INFO',
+                        'Created geolocation data for course with UUID: {}'.format(valid_uuid)
+                    )
+                )
+                assert GeoLocation.objects.count() == 1
+
+    def test_program_success_flow(self, jwt_decode_patch):  # pylint: disable=unused-argument
```
DawoudSheraz: these 2 tests can be combined into a DDT test.

DawoudSheraz (commented):
Left a few comments to address.

course_discovery/apps/course_metadata/models.py
danielkalen: Manually

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
danielkalen: It would be run at least one, the way I understood it is that this might be used again in the future, hence this change.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
danielkalen: makes sense, do you think this should be converted into a catch-all exception handler? Or should we remove the try/except block all together?

DawoudSheraz: try/except can be removed. The queryset can be converted into a list and then any existence checks can be performed on the list.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,208 @@
...
+                )
+                self.skipped_items.append("Skipped {} with UUID {}. Errors: {}".format(product_type, row_uuid, err_message))  # pylint: disable=line-too-long
+                continue
+
+            logger.info('Starting data import flow for {}: {}'.format(product_type, row_uuid))  # lint-amnesty, pylint: disable=logging-format-interpolation
+
+            if product_type == 'course':
+                self.injest_entry(model, row_uuid, geolocation, self.processed_courses)
+            else:  # product_type == 'progarm'
+                self.injest_entry(model, row_uuid, geolocation, self.processed_programs)
```
DawoudSheraz: nit: You can create a map/dict of product_type -> processed_product_ids and get the appropriate list from dict instead of having this if-else block.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,208 @@
...
+        self.check_for_potential_orphans(model)
+
+        logger.info("Geolocation CSV loader ingest pipeline has completed.")
+
+        self.log_skipped_items()
+
+        self.log_processed_products()
+
+    def injest_entry(self, model, row_uuid, geolocation, processed_entries):
+        for product_obj in model.objects.filter(uuid=row_uuid):
```
DawoudSheraz: For Course, model.objects will only return non-draft objects. The draft objects will get out of sync with non-draft with regards to GeoLocation data.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,208 @@
...
+        model_name = model._meta.object_name
+        kwrags = {key: value}
+
+        return (
+            model.everything.filter(**kwrags).exists() if model_name == 'Course'
+            else model.objects.filter(**kwrags).exists()
+        )
+
+    def check_for_potential_orphans(self, model):
+        for geolocation_id in self.updated_geolocations:
```
DawoudSheraz, line 191: There is a util method in discovery core app, delete_orphans https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/core/utils.py#L130. Check it out, it might already be doing what you are attempting to do.

course_discovery/apps/course_metadata/data_loaders/tests/test_geolocation_csv_loader.py
```diff
@@ -0,0 +1,223 @@
...
+                            INVALID_GEOLOCATION_CSV_DICT['UUID'],
+                            INVALID_GEOLOCATION_CSV_DICT['UUID']
+                        )
+                    )
+                )
+                assert GeoLocation.objects.count() == initial_geolocation_obj_count
+
+    @ddt.data('course', 'program')
+    def test_missing_product_validation_failure(self, product_type, jwt_decode_patch):  # pylint: disable=unused-argument
+        initial_geolocation_obj_count = GeoLocation.objects.count()
```
DawoudSheraz, line 102: Out of curiosity; shouldn't it be 0 by default?

DawoudSheraz (commented):
left some followup comments

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
danielkalen: I checked this thoroughly and it looks like even though these methods are similarly named they have a slightly diff nature. I can explain (or at least try to) if necessary...

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,211 @@
...
+                    'Skipping ingestion for this item.'
+                    'Details: {}'
+                    .format(row_uuid, err_message)
+                )
+                self.skipped_items.append("Skipped {} with UUID {}. Errors: {}".format(product_type, row_uuid, err_message))  # pylint: disable=line-too-long
+                continue
+
+            logger.info('Starting data import flow for {}: {}'.format(product_type, row_uuid))  # lint-amnesty, pylint: disable=logging-format-interpolation
+
+            self.injest_entry(model, row_uuid, geolocation, processed_products[product_type])
```
DawoudSheraz: the method is ingest_entry. The typo `j` is causing the new failures.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
DawoudSheraz: Yes, I will be interested in learning about that.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,211 @@
...
+        logger.info("Initiating Geolocation CSV data loader flow.")
+        for row in self.reader:
+            row = self.transform_dict_keys(row)
+            row_uuid = row['uuid']
+            product_type = row['product_type']
+            model = Course if product_type == 'course' else Program
+            processed_products = {
+                'course': self.processed_courses,
+                'program': self.processed_programs,
+            }
```
DawoudSheraz: nit: This should be moved at the top of ingest to avoid re-assignment in each loop iteration.

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
```diff
@@ -0,0 +1,211 @@
...
+        for product_obj in products:
+            action_taken = 'Created'
+            existing_geolocation_id = product_obj.geolocation.id if product_obj.geolocation else None  # lint-amnesty
+
+            product_obj.geolocation = GeoLocation.objects.create(**geolocation)
+            product_obj.save()
+
+            if existing_geolocation_id:
+                self.updated_geolocations.append(existing_geolocation_id)
+                action_taken = 'Updated'
```
DawoudSheraz, line 97: Did you verify this new flow for Course, especially for a course that has both draft and non-draft versions? I suspect this will result in the creation of 2 separate GeoLocation objects for both draft and non-draft versions and that is not desirable. Both draft and non-draft typically point to the same related object.

DawoudSheraz (commented):
There is just one comment/question to answer. Once answered, I will give a 👍🏽

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
danielkalen: 😞 I am embarrassed that I missed this

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
danielkalen: Yes that's what the check_for_potential_orphans takes care of if I'm not mistaken. This logic was knocked off of geotargeting_loader.py as the business requirements of this loader is extremely similar to the one I'm creating here, and here is the justification for it as laid out my teammate Marek 
CleanShot 2022-11-17 at 18 01 12@2x

DawoudSheraz: Alright, that helps. Although it would have been ok to keep the same object for both draft and non-draft courses. I would ask you to verify this on stage after merge/local before merge for the following criteria:

- Course that only has a draft version (Create a new one from the publisher)
- Course that has both draft and non-draft versions (complete review process from Publisher)

Apart from these, do the following 2 separate runs as well:

- Run loader on the course that has a draft version, complete its review process and verify if the Geolocation object is copied over to the non-draft version
- Run loader on the course that has both draft and non-draft versions. Re-run the loader but change some values for the same course. Verify the values are updated for Geolocation for both draft and non-draft versions.

I am sorry I am taking some time with this review. The draft and non-draft distinction in Discovery can easily spill over if not handled correctly. Any discrepancy in the data will affect the services that utilize the data from Discovery.

DawoudSheraz (approved):
- Some testing to be done before merge
- Squash the commits

AliAdnanSohail (approved):

course_discovery/apps/course_metadata/data_loaders/geolocation_loader.py
danielkalen: Hey Dawoud, sorry it took me so long to take care of this this completely blew past my mind! I have ran the manual tests you suggested to do above and haven't found any issues, everything looks as expected. I will update from master and then merge this once everything is green.

PR #3659, opened 2022-10-28
feat: Added weeks_to_complete in MinimalCourseRunSerializer

zamanafzal:
Ticket: https://2u-internal.atlassian.net/browse/ENT-6407
Description:
We need the `weeks_to_complete` data for every course run in the program's API. I've added it to `MinimalCourseRunSerializer`
http://localhost:18381/api/v1/programs/{program_uuid}/

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -868,7 +868,7 @@ class MinimalCourseRunSerializer(FlexFieldsSerializerMixin, TimestampModelSerial
     class Meta:
         model = CourseRun
         fields = ('key', 'uuid', 'title', 'external_key', 'image', 'short_description', 'marketing_url',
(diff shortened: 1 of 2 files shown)


ansabgillani (approved):

Ali-D-Akbar (approved):
Out of curiosity, how do we use this `weeks_to_complete` field to fix the problem stated in the ticket? I'm seeing something like `b2c` method... just wondering what this is.

zamanafzal:
@Ali-D-Akbar on the b2c side whole page is rendered as HTML and this `weeks_to_complete` field was provided by the view. 
On the B2B side, we are using programs API and the field we were using is deprecated and the data in that field is not correct.

zamanafzal:
@ansabgillani @Ali-D-Akbar Can you guys please merge this PR?

PR #3660, opened 2022-10-28
feat: add model for meta/seo fields for external courses

DawoudSheraz:
### Description

- Add model for storing SEO/Meta fields for external product lines, ExecEd specifically
- Link the model with AdditionalMetadata.


course_discovery/apps/course_metadata/models.py
```diff
@@ -657,9 +657,31 @@ class CertificateInfo(AbstractHeadingBlurbModel):
...
+    """
+    title = models.CharField(
+        max_length=100, default='', null=True, blank=True,
+        help_text="Product title that will appear in meta tag for search engine ranking"
+    )
+    description = models.CharField(
+        max_length=255, default='', null=True, blank=True,
+        help_text="Product description that will appear in meta tag for search engine ranking"
+    )
+    keywords = TaggableManager(
```
Ali-D-Akbar, line 672: Does this field contains all the SEO attributes? And only `AdditionalMetadata` uses this model, right?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -552,6 +552,12 @@ def facts_list(self, obj):
         ])
 
 
+@admin.register(ProductMeta)
+class ProductMetaAdmin(admin.ModelAdmin):
+    list_display = ['title', 'description']
+    search_fields = ['title']
```
Ali-D-Akbar, line 558: do we not want to search with keywords?

Ali-D-Akbar (approved):
Just some questions otherwise, looks good.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: No, this field is only for SEO keywords. For example, a web dev course can contain tags such as HTML,web-dev, web-design, etc. These keywords/tags are meant for search engines. And yes, the model will be linked within AdditionalMetadata because it is for external courses only.

course_discovery/apps/course_metadata/admin.py
DawoudSheraz: Not needed. The product title would be enough for searching.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -191,6 +191,18 @@ def test_course_ends(self, end, expected):
         factories.CourseRunFactory(course=course, end=end)
         self.assertEqual(course.course_ends, expected)
 
+    def test_product_meta(self):
+        """
+        Verify the ProductMeta has expected fields' data.
+        """
+        keywords_list = ['keyword_1', 'keyword_2', 'keyword_3']
```
azanbinzahid, line 198: will keywords have spaces, or special chars b/w them?

azanbinzahid (approved):

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: Yes, they can have spaces.

course_discovery/apps/course_metadata/models.py
```diff
@@ -697,6 +719,14 @@ class AdditionalMetadata(TimeStampedModel):
         help_text=_('This field allows for override the default course term'),
         blank=True,
     )
+    product_meta = models.OneToOneField(
+        ProductMeta,
+        on_delete=models.DO_NOTHING,
```
uzairr, line 724: is there is a specific reason to not `cascading` here like above in the FK of `CertificateInfo` ?

uzairr (approved):
Just one comment otherwise 👍

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Because it is not required for all external courses (at this moment) and deleting the SEO info of a product should not be deleting its related additional metadata.

PR #3669, opened 2022-11-02
feat: add explicit status field for external courses

DawoudSheraz:
### Description

- Add field in AdditionalMetadata to reflect the status of external course. The status can either be Published or Archived
- Determine the availability of ExecEd courses using the newly added status field


course_discovery/apps/course_metadata/models.py
```diff
@@ -1986,6 +1991,13 @@ def availability(self):
         now = datetime.datetime.now(pytz.UTC)
         upcoming_cutoff = now + datetime.timedelta(days=60)
 
+        # Use external course availability for ExecEd course run types
+        # Check additional_metadata existence as a fail safe for some unit tests
+        # where this flow is triggered via signals before AdditionalMetadata object can be persisted
+        if self.type.slug in [CourseRunType.UNPAID_EXECUTIVE_EDUCATION, CourseRunType.PAID_EXECUTIVE_EDUCATION] and \
+                self.course.additional_metadata:
+            return self.external_course_availability
```
azanbinzahid, line 1999: What about date specific checks that are later in order? this if condition will ignore those, won't it?

DawoudSheraz: Yes, since we are not catering for dates at the moment. With the current implementation, the dates can be handled in `external_course_availability ` when needed.

azanbinzahid (approved):

course_discovery/apps/course_metadata/choices.py
```diff
@@ -63,3 +63,11 @@ class CourseLength(DjangoChoices):
...
     Medium = ChoiceItem('medium', _('Medium'))
     Long = ChoiceItem('long', _('Long'))
+
+
+class ExternalProductStatus(DjangoChoices):
+    """
+    The statuses for external product lines.
+    """
+    Archived = ChoiceItem('archived', _('Archived'))
+    Published = ChoiceItem('published', _('Published'))
```
uzairr, line 73: normally choice items are UPPERCASE, please validate!

uzairr (approved):
Just a small nit! otherwise 👍

course_discovery/apps/course_metadata/choices.py
DawoudSheraz: https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/choices.py The convention is different in Discovery.

PR #3670, opened 2022-11-02
fix: add debug logging for event bus caching

robrap:
Add debug logging and monitoring for api caching

In order to help understand why caching is not being busted
properly for the event consumer, we are temporarily add
logging and monitoring to gather more information.


DawoudSheraz (approved):

DawoudSheraz:
Commit type should be temp instead of fix

PR #3672, opened 2022-11-03
temp: more logging around course_metadata cache

rgraber:
Be more specific about logging exactly what is getting set in and then gotten from the cache.

### DIFF
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -59,11 +59,10 @@ def set_api_timestamp():
     #   This code uses/exposes a toggle that was only intended for the consumer code in:
     #      course_discovery/apps/course_metadata/management/commands/consume_events.py
     KAFKA_CONSUMERS_ENABLED = SettingToggle('EVENT_BUS_KAFKA_CONSUMERS_ENABLED', default=False)
-    if KAFKA_CONSUMERS_ENABLED.is_enabled():
-        logger.info(f"Updating api timestamp in api_change_receiver to {timestamp}")
     cache.set(API_TIMESTAMP_KEY, timestamp, None)
     if KAFKA_CONSUMERS_ENABLED.is_enabled():
-        logger.info(f"Api timestamp in cache is now set to to {cache.get(API_TIMESTAMP_KEY)}")
+        logger.info(f"Set api timestamp in set_api_timestamp to {timestamp}. "
+                    f"Api timestamp in cache is now set to {cache.get(API_TIMESTAMP_KEY)}.")
 
 
 def api_change_receiver(sender, **kwargs):  # pylint: disable=unused-argument
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -54,7 +54,16 @@ def timestamped_object_key_constructor(*args, **kwargs):
 
 def set_api_timestamp():
     timestamp = time.time()
+    # 2022-11: Temporary logging to help debug a caching issue related to the new event bus event.
+    #   If this code hasn't been cleaned up by 2023, feel free to remove it.
+    #   This code uses/exposes a toggle that was only intended for the consumer code in:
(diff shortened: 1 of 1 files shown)


course_discovery/apps/api/cache.py
```diff
@@ -54,7 +54,16 @@ def timestamped_object_key_constructor(*args, **kwargs):
...
+    # 2022-11: Temporary logging to help debug a caching issue related to the new event bus event.
+    #   If this code hasn't been cleaned up by 2023, feel free to remove it.
+    #   This code uses/exposes a toggle that was only intended for the consumer code in:
+    #      course_discovery/apps/course_metadata/management/commands/consume_events.py
+    KAFKA_CONSUMERS_ENABLED = SettingToggle('EVENT_BUS_KAFKA_CONSUMERS_ENABLED', default=False)
+    if KAFKA_CONSUMERS_ENABLED.is_enabled():
+        logger.info(f"Updating api timestamp in api_change_receiver to {timestamp}")
     cache.set(API_TIMESTAMP_KEY, timestamp, None)
+    if KAFKA_CONSUMERS_ENABLED.is_enabled():
+        logger.info(f"Api timestamp in cache is now set to to {cache.get(API_TIMESTAMP_KEY)}")
```
robrap: You have "to to" in the message.

course_discovery/apps/api/cache.py
```diff
@@ -54,7 +54,16 @@ def timestamped_object_key_constructor(*args, **kwargs):
 
 def set_api_timestamp():
     timestamp = time.time()
+    # 2022-11: Temporary logging to help debug a caching issue related to the new event bus event.
+    #   If this code hasn't been cleaned up by 2023, feel free to remove it.
+    #   This code uses/exposes a toggle that was only intended for the consumer code in:
+    #      course_discovery/apps/course_metadata/management/commands/consume_events.py
+    KAFKA_CONSUMERS_ENABLED = SettingToggle('EVENT_BUS_KAFKA_CONSUMERS_ENABLED', default=False)
+    if KAFKA_CONSUMERS_ENABLED.is_enabled():
+        logger.info(f"Updating api timestamp in api_change_receiver to {timestamp}")
```
robrap: 1. The message should now mention `set_api_timestamp` instead of `api_change_receiver`.
2. Nit: I'd move all this code after the cache set call and have a single log line that includes all the details you want to record.

course_discovery/apps/api/cache.py
```diff
@@ -98,6 +99,8 @@ def process_cache_response(self, view_instance, view_method, request, args, kwar
                 kwargs=kwargs
             )
             response_triple = self.cache.get(key)
+            api_timestamp = self.cache.get(API_TIMESTAMP_KEY)
+            set_custom_attribute('api_cache_timestamp', api_timestamp)
```
robrap, line 102: Nit: Could be on one line like was done for the log message, but fine as-is as well.

robrap (approved):
Great. Minor changes noted, but no blockers.

course_discovery/apps/api/cache.py
rgraber: I think it looks neater in 2 lines. [wontfix] unless you really want to argue about it

course_discovery/apps/api/cache.py
rgraber: [fixed]

course_discovery/apps/api/cache.py
rgraber: [fixed]

Ali-D-Akbar (approved):
Looks good but why did we remove the older logs? Are they not needed?

rgraber:
> Looks good but why did we remove the older logs? Are they not needed?
We just moved them to a more useful place so we could see the actual timestamp getting set, not just the fact that we were setting something. The effect should be basically the same.

robrap:
Thanks for the very late night merge @Ali-D-Akbar.

PR #3673, opened 2022-11-03
[DO NOT MERGE] [POC] Index Contentful data in Algolia Spike

marekwrobel:
This is a simple example of how we can fetch specific data (such as degrees) from Contentful and create a dictionary where each key is a degree UUID (and value contains the fields to be indexed for that degree), that can then be matched with degrees from our DB. 

This way we can add additional fields to `AlgoliaProxyProduct`, which results in these Contentful-based fields being indexed in Algolia together with disco-based fields.

This is a very simple/raw example. Please read the [spike description and summary available here](https://2u-internal.atlassian.net/wiki/spaces/WS/pages/216269142/SPIKE+Indexing+Contentful+data+in+Algolia) for more in-depth tips on actual implementation. 

Also, ignore all package-related changes that came from me upgrading all packages locally. We only need to install the `contentful` package for this implementation.

### DIFF
--- a/course_discovery/apps/course_metadata/algolia_models.py
+++ b/course_discovery/apps/course_metadata/algolia_models.py
@@ -116,9 +116,9 @@ class AlgoliaProxyProduct(Program):
         self.product = product
         self.product.language = language
         product_uuid = str(product.uuid)
-        contentful_product = contentful_data[product_uuid] if contentful_data else None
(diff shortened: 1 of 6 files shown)


DawoudSheraz:
Closing this PR as this work has been complete in other PRs.

PR #3678, opened 2022-11-07
feat: remove logging around cache

rgraber:
Remove some temporary logging that was set up to debug eks access to memcached.


robrap (approved):
Hopefully we don't see the same issue in Prod. But if we do, I guess we can bring back any logs if we need them.

Ali-D-Akbar (approved):

PR #3680, opened 2022-11-08
feat: add handling of slug in CSV for ExecEd

DawoudSheraz:
### Description

- Add the handling of the newly added slug field in detail=2 endpoint of product API in data transformation command
- Use the slug in CSV in CSV loader, falling back to default course slug if CSV slug is not present
- **Unrelated to Ticket AC** Fix some logging behavior involving exceptions. Instead of logging the exception message, log the response message from exception as that contain the actual useful information

### Testing

- Check out this branch on your local
- Get auth token from post collection
- Go to discovery shell
- Run `./manage.py populate_executive_education_data_csv --auth_token=<token> --output_csv=test.csv`
- Make sure the output csv has slug column
- You can choose to keep small number of rows. Change 1-2 slugs to values that are present in the DB `/admin/course_metadata/courseurlslug/`
- Run the CSV loader `./manage.py import_course_metadata --csv_path=test.csv`
- Verify the slug is updated for valid columns and that the error is raised for the slug already present in the DB.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/constants.py
+++ b/course_discovery/apps/course_metadata/data_loaders/constants.py
@@ -34,10 +34,10 @@ class CSVIngestionErrorMessages:
     MISSING_REQUIRED_DATA = '[MISSING_REQUIRED_DATA] Course {course_title} is missing the required data for ' \
                             'ingestion. The missing data elements are "{missing_data}"'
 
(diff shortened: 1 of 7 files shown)


ansabgillani (approved):
Looks good to me

AliAdnanSohail (approved):

azanbinzahid (approved):

PR #3706, opened 2022-11-28
fix: hotfix for catering a specific org on cybermonday

uzairr:
ES is behaving weird in case of filtering the courses against an org.This PR introduces a hotfix which will mask a particular org to not been checked against ES.

### DIFF
--- a/course_discovery/apps/api/v1/views/catalog_queries.py
+++ b/course_discovery/apps/api/v1/views/catalog_queries.py
@@ -46,6 +46,16 @@ class CatalogQueryContainsViewSet(ValidElasticSearchQueryRequiredMixin, GenericA
             if course_uuids:
                 course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
                 specified_course_ids += course_uuids
+                # hotfix for Cyber Monday starting (28th Nov 2022)
+                cyber_query_string = 'number: (-CSE6040x AND -ISYE6501x AND -MGT6203x) AND org: (-GTx)'
+                if query == cyber_query_string:
+                    courses_orgs = [
+                        Course.objects.get(uuid=course_uuid).key.split('+')[0] for course_uuid in specified_course_ids
+                    ]
+                    if 'GTx' not in courses_orgs:
+                        contains = {str(course_uuid): bool(course_uuid) for course_uuid in specified_course_ids}
+                        return Response(contains)
+
                 log.info("Specified course ids: %s", specified_course_ids)
                 identified_course_ids.update(
                     Course.search(query).filter(partner=partner, uuid__in=course_uuids).values_list('uuid', flat=True)


course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -46,6 +46,14 @@ def get(self, request):
...
                 course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
                 specified_course_ids += course_uuids
+                # hotfix for Cyber Monday starting (28th Nov 2022)
+                courses_orgs = [
+                    Course.objects.get(uuid=course_uuid).key.split('+')[0] for course_uuid in specified_course_ids
+                ]
+                if 'GTx' not in courses_orgs:
+                    contains = {str(course_uuid): bool(course_uuid) for course_uuid in specified_course_ids}
+                    return Response(contains)
+
```
colinbrash, line 58: This endpoint is used for a lot more than just for this coupon.  So we need to make sure this logic doesn't get applied other API queries.  (If it did, **every single** API query, including `org:(GTx)` would return `true` if the courses were not GTx!

colinbrash: So -- you need to check the `query`.  Only apply this logic if it is equal to `number:(-CSE6040x AND -ISYE6501x AND -MGT6203x) AND org:(-GTx)`

course_discovery/apps/api/v1/views/catalog_queries.py
```diff
@@ -46,6 +46,14 @@ def get(self, request):
             if course_uuids:
                 course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
                 specified_course_ids += course_uuids
+                # hotfix for Cyber Monday starting (28th Nov 2022)
+                courses_orgs = [
+                    Course.objects.get(uuid=course_uuid).key.split('+')[0] for course_uuid in specified_course_ids
+                ]
+                if 'GTx' not in courses_orgs:
+                    contains = {str(course_uuid): bool(course_uuid) for course_uuid in specified_course_ids}
```
colinbrash: Instead of `bool(course_uuid)` you just want `true`.

colinbrash (changes requested):

course_discovery/apps/api/v1/views/catalog_queries.py
colinbrash: Also would be a good idea to add logs, so we can be more confident it is working as intended!

uzairr: it will evaluate to `true`, putting it directly is raising error.

colinbrash (approved):
Looks good.  Will Slack you about testing!

ansabgillani (approved):

PR #3714, opened 2022-12-05
feat: transform contentful fields and index in algolia

Ali-D-Akbar:
[PROD-3024](https://2u-internal.atlassian.net/browse/PROD-3024)
This PR adds a util functions which are used to transform bootcamp data from Contentful. The transformed fields are going to be sent to algolia for indexing and to lightcast for skill extraction. 

Note: Unit tests for data transformation does sufficient job but please be sure to run this on local too.

### Testing Instructions for Contentful data transformation
1. Set the following variables in `private.py` file:
```
CONTENTFUL_SPACE_ID = <space_id>
CONTENTFUL_CONTENT_DELIVERY_API_KEY = <api_key>
BOOTCAMP_CONTENTFUL_CONTENT_TYPE = 'bootCampPage'
```
2. Run python shell in discovery shell: `make discovery-shell` `./manage.py shell`
3. Run the following code in shell to read the data from contentful:
```
from course_discovery.apps.course_metadata.utils import transform_bootcamp_contentful_data
transform_bootcamp_contentful_data()
```

### Testing Instructions for Algolia Reindexing on local
1. Set the following variables in `private.py` file (`MyAppID` is the id for EdX_Staging application and `MyApiKey` is the admin key of algolia):
```
ALGOLIA = {
    'APPLICATION_ID': 'MyAppID',
    'API_KEY': 'MyApiKey'
}
```
2. Apply this patch to your current github branch: [algolia_indexing_test_data.patch.gz](https://github.com/openedx/course-discovery/files/10205688/algolia_indexing_test_data.patch.gz) using this command `gunzip < /path/to/algolia_indexing_test_data.patch.gz | git apply -`
3. Edit the uuid on L126 with your local devstack `Demo_Course` uuid
4. Run discovery shell: `make discovery-shell`
5. Run the following: `./manage.py algolia_reindex`
6. See the new fields being indexed here for your demo course: https://www.algolia.com/apps/testing2MLYZ9WWU0/explorer/browse/product?searchMode=search

### Pre-merge:
- [x] Make an edx-internal [PR](https://github.com/edx/edx-internal/pull/7514) to add `CONTENTFUL_SPACE_ID` and `CONTENTFUL_CONTENT_DELIVERY_API_KEY` settings for stage and prod.

### Post-merge:
- [ ] Run Algolia-reindexing on staging and verify bootcamps data properly updated on EdX-Staging


course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -113,10 +114,18 @@ class AlgoliaProxyProduct(Program):
...
         super().__init__()
         self.product = product
         self.product.language = language
+        product_uuid = str(product.uuid)
+
+        if not contentful_data or product_uuid not in contentful_data:
+            return
+        contentful_product = contentful_data[product_uuid]
+
+        for key, value in contentful_product:
```
AliAdnanSohail: Shouldn't we move all these fields in a separate dict in Algolia? Or there is any requirement or use case to keep each key separately?
If we keep all fields separate in Algolia not in a specific dict, then there will be a big difference between Algolia Bootcamp product and Algolia non-bootcamp product

AliAdnanSohail (changes requested):

course_discovery/apps/course_metadata/algolia_models.py
DawoudSheraz: +1 on this.

Ali-D-Akbar: Hmmm, I'm not sure if this was their requirement to keep separate values. We can keep this in a dict for now...

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+    total_entries.extend(contentful_entries.items)
+    total_count = contentful_entries.total
+
+    while skip < total_count:
+        contentful_entries = client.entries({
+            'content_type': content_type,
+            'limit': limit,
+            'skip': skip,
+            'include': include
+        })
```
DawoudSheraz: Please add some logs here to keep track of the fetch process.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
-    return client.entries({
-        'content_type': content_type
+    limit = 10
+    skip = 10
+    include = 5  # the depth of linked entries to be fetched from contentful
+    total_entries = []
+    contentful_entries = client.entries({
+        'content_type': content_type,
+        'limit': limit,
+        'include': include
```
DawoudSheraz: Please update docstring to showcase these changes.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
     """
     client = Client(
         settings.CONTENTFUL_SPACE_ID,
-        settings.CONTENTFUL_CONTENT_DELIVERY_API_KEY
+        settings.CONTENTFUL_CONTENT_DELIVERY_API_KEY,
+        timeout_s=30  # increases read timeout
     )
-    return client.entries({
-        'content_type': content_type
+    limit = 10
```
DawoudSheraz: Why is the limit so small here?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+            'include': include
+        })
+        if contentful_entries:
+            total_entries.extend(contentful_entries.items)
+            skip += limit
+
+    return total_entries
+
+
+def traverse_plain_text(dictt, key):
```
DawoudSheraz: nit: rename dictt.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+    """
+    Recursive function to extract a list of values of all the keys containing plain text.
```
DawoudSheraz: Please add info in the docstring showcasing the structure of dict and list.

course_discovery/apps/course_metadata/utils.py
DawoudSheraz: nit: Since this is only needed for contentful data, update the method name.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+                text_list += traverse_plain_text(ob, key)
+    return text_list
+
+
+def rich_text_to_plain_text(rich_text):
+    """
+    Converts rich text from Contentful to plain text.
+    Plain text resides in the dict with 'value' as dict key name.
+    """
+    text_key = 'value'
```
DawoudSheraz: nit: no need to declare this variable.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+    return ''.join(traverse_plain_text(rich_text, text_key))
+
+
+def transform_bootcamp_contentful_data():
```
DawoudSheraz: nit: fetch and transform bootcamp contentful data. 

Also, it would be a good idea to move contentful-related methods into a separate utils file. I can expect we would add more utils once the code reaches production and contentful is expanded upon.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+            bootcamp_curriculum_faq_module = bootcamp_curriculum_module.items
+
+            for bootcamp_curriculum_faq_item in bootcamp_curriculum_faq_module:
+                faq_question = bootcamp_curriculum_faq_item.question
+                faq_answer = rich_text_to_plain_text(bootcamp_curriculum_faq_item.answer_rich_text)
+                faq_items.append({
+                    'question': faq_question,
+                    'answer': faq_answer,
+                })
+
```
DawoudSheraz: nit: create small utils methods for each data key and call them here. It will make code readable and easy to follow.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -859,8 +859,145 @@ def get_data_from_contentful(content_type):
...
+
+def transform_bootcamp_contentful_data():
+    """
+    Transforms incoming bootcamp data from contentful to algolia-usable form.
+
+    Each Contentful entry has seo, hero and modules list.
+    Each rich text content field has been transformed into plain text using `rich_text_to_plain_text`.
+    """
+    contentful_bootcamp_page_entries = get_data_from_contentful('bootCampPage')
```
DawoudSheraz: nit: These contentful specific keys can be read from settings. It will make it easier to override them for dev and testing environment when needed. The current flow makes it very specific to internal use.

course_discovery/apps/course_metadata/tests/contentful_bootcamp_mock_data.py
```diff
@@ -0,0 +1,143 @@
...
+})
+
+about_the_program_entry = create_contentful_entry('aboutTheProgramModule', {
+    'internalName': 'Data Analytics Boot Camp | The University of Western Australia',
+    'heading': 'The support you need to transform your career.',
```
DawoudSheraz: - Run formatter on this file to organize dicts
- Use lorem ipsum or random text. The current content has similarities to edx.org public data which might impact search engines (Github being public).

course_discovery/apps/course_metadata/index.py
```diff
@@ -81,7 +84,7 @@ class EnglishProductIndex(BaseProductIndex):
                      ('product_organization_logo_override', 'organization_logo_override'),
                      ('product_meta_title', 'meta_title'),
                      'active_run_key', 'active_run_start', 'active_run_type', 'owners', 'course_titles', 'tags',
-                     'skills', )
+                     'skills', 'bootcamp_contentful_fields',)
```
DawoudSheraz: instead of bootcamp_contentful_fields, lets use contentful_fields. Bootcamp becomes specific to a particular product.

course_discovery/apps/course_metadata/index.py
```diff
@@ -16,7 +17,9 @@ def get_queryset(self):  # pragma: no cover
         if not self.language:
             raise Exception('Cannot update Algolia index \'{index_name}\'. No language set'.format(
                 index_name=self.index_name))
-        qs1 = [AlgoliaProxyProduct(course, self.language) for course in AlgoliaProxyCourse.objects.all()]
+        contentful_bootcamp_data = transform_bootcamp_contentful_data()
```
DawoudSheraz: nit: rename to contentful_data.

DawoudSheraz (commented):
Left a few followup comments.

course_discovery/apps/course_metadata/utils.py
Ali-D-Akbar: I tried to fetch larger count number like 50 but that sometimes gives an error like "data is too big, cannot be fetched".

Ali-D-Akbar: `Response size too big. Maximum allowed response size: 7340032B.`

course_discovery/apps/course_metadata/utils.py
Ali-D-Akbar: logging like this:
Screenshot 2022-12-14 at 12 05 13 PM

course_discovery/apps/course_metadata/tests/contentful_bootcamp_mock_data.py
Ali-D-Akbar: Looks uglier when formatter is run, specifically the rich_text dicts:
Screenshot 2022-12-14 at 2 03 39 PM

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -113,10 +113,17 @@ class AlgoliaProxyProduct(Program):
...
         proxy = True
 
-    def __init__(self, product, language='en'):
+    def __init__(self, product, language='en', contentful_data=None):
         super().__init__()
         self.product = product
         self.product.language = language
+        product_uuid = str(product.uuid)
+
+        if not contentful_data or product_uuid not in contentful_data:
```
ansabgillani: Just use 
`if contentful_data and product_uuid in contentful_data:`
`self.product.contentful_fields = contentful_data[product_uuid]`

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+                ]
+            }
+        ]
+    }
+
+    This utility will returns this plain text:
+    "Lorem ipsum:  dolor sit amet, consectetur adipiscing elit. Excepteur sint: occaecat cupidatat non proident."
+    """
+    text_list = []
+    for k, v in rich_text_dict.items():
```
ansabgillani: Use proper variable names.

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+        if k == 'value':
+            text_list.append(v)
+        elif isinstance(v, dict):
```
ansabgillani: Add inline comment `# checking if the value is a nested dict`

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+            text_list += extract_plain_text_from_rich_text(v)
+        elif isinstance(v, list):
```
ansabgillani: Same as above

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+"""
+import logging
+
+from contentful import Client
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+def get_data_from_contentful(content_type):
```
ansabgillani, line 20: Can you convert all these functions in a class like BootcampContentful utils with a bunch of private functions? It might be confusing when Degrees functions are added here.

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -190,7 +190,7 @@ def test_exclude_unavailable_program_types(self, path, serializer, result_locati
         ProgramFactory(courses=[course_run.course], status=program_status)
         self.reindex_courses(active_program)
 
-        with self.assertNumQueries(expected_queries, threshold=1):  # CI sometimes adds a query
+        with self.assertNumQueries(expected_queries, threshold=2):  # CI sometimes adds a bunch of query
```
DawoudSheraz: nit: a bunch of queries

course_discovery/apps/course_metadata/algolia_models.py
DawoudSheraz: +1 to this. return in init also seems weird.

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+            'skip': skip,
+            'include': include
+        })
+        if contentful_entries:
+            total_entries.extend(contentful_entries.items)
+            logger.info(f'Successfully fetched Contentful Entries from {skip + 1} to '
+                        f'{skip + limit if skip + limit < total_count else total_count}')
+            skip += limit
+
+    return total_entries
```
DawoudSheraz: We should think about caching this response. There are quite a bunch of API calls here that can overwhelm the Contentful system if there is ever a need to index in a short period of time.

course_discovery/apps/course_metadata/contentful_utils.py
DawoudSheraz: and add \n after adding the value.

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+                text_list += extract_plain_text_from_rich_text(ob)
+    return text_list
+
+
+def rich_text_to_plain_text(rich_text):
+    """
+    Converts rich text from Contentful to plain text.
+    Plain text resides in the dict with 'value' as dict key name.
+    """
+    return ''.join(extract_plain_text_from_rich_text(rich_text))
```
DawoudSheraz, line 161: oh, ignore my \n comment above. Do this here instead.

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+            transformed_bootcamp_data[product_uuid].update(bootcamp_curriculum_module)
+
+        if 'partnershipModule' in module_list:
+            partnership_module = get_partnership_module(bootcamp_entry.modules[module_list.index('partnershipModule')])
+            transformed_bootcamp_data[product_uuid].update(partnership_module)
+
+        if 'faqModule' in module_list:
+            faq_module = get_faq_module(bootcamp_entry.modules[module_list.index('faqModule')].faqs)
+            faq_items += faq_module
+            print(faq_items)
```
DawoudSheraz: remove this print.

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def fetch_and_transform_bootcamp_contentful_data():
+    """
+    Transforms incoming bootcamp data from contentful to algolia-usable form.
+
+    Each Contentful entry has seo, hero and modules list.
+    Each rich text content field has been transformed into plain text using `rich_text_to_plain_text`.
+    """
+    contentful_bootcamp_page_entries = get_data_from_contentful(settings.BOOTCAMP_CONTENTFUL_CONTENT_TYPE)
```
DawoudSheraz, line 253: BOOTCAMP_CONTENTFUL_CONTENT_TYPE -- has this been added in a prior PR?

DawoudSheraz (approved):
I have not tested it locally. Overall, looks good.

course_discovery/apps/course_metadata/contentful_utils.py
Ali-D-Akbar: I'm not adding `\n` because it breaks sentences in the middle.

course_discovery/apps/course_metadata/contentful_utils.py
Ali-D-Akbar: I think these functions will eventually be reused with Degree work so keeping them public will aid us in degree work.

course_discovery/apps/course_metadata/contentful_utils.py
Ali-D-Akbar: Breaks sentences. I don't think we'd need something like this here.

course_discovery/apps/course_metadata/contentful_utils.py
Ali-D-Akbar: https://github.com/edx/edx-internal/pull/7514

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -67,7 +67,7 @@ def delegate_attributes(cls):
                      'product_max_effort', 'product_min_effort', 'active_run_key', 'active_run_start',
                      'active_run_type', 'owners', 'program_types', 'course_titles', 'tags',
                      'product_organization_short_code_override', 'product_organization_logo_override', 'skills',
-                     'product_meta_title',
+                     'product_meta_title', 'contentful_fields'
```
AliAdnanSohail, line 70: Will query be searched in `contentful_fields`? If not, then we should add this in search_fields

AliAdnanSohail (changes requested):

course_discovery/apps/course_metadata/algolia_models.py
Ali-D-Akbar: The decision for now is to do it like this. Ideally, we'd first like to check how the data looks like and what kind of effects it will have on the searches.
Besides, I don't think that the search field should be a nested object.

AliAdnanSohail (approved):

PR #3727, opened 2022-12-19
feat: Add ability to replay events to event consumer.

dianakhuang:
This adds an optional argument of a timestamp to
start event consumption from.

https://github.com/edx/edx-arch-experiments/issues/115

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/consume_events.py
+++ b/course_discovery/apps/course_metadata/management/commands/consume_events.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime
 
 from django.core.management.base import BaseCommand
 from edx_event_bus_kafka import KafkaEventConsumer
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -70,11 +79,20 @@ def handle(self, *args, **options):  # pragma: no cover
             return
         try:
             signal = OpenEdxPublicSignal.get_signal_by_type(options['signal'][0])
+            if options['offset_time'] and options['offset_time'][0] is not None:
```
robrap, line 82: Checking `options['offset_time']` wouldn't hurt, but may be unnecessary since it has a default?

robrap (approved):
I guess we decided earlier not to test this? I added one nit comment separately.

course_discovery/apps/course_metadata/management/commands/consume_events.py
dianakhuang: Hmm, what do you mean by this? If I don't check `options['offset_time']` then I can't check `options['offset_time'][0]`.

robrap: I meant that since you set the default to None, I would assume (but haven't tested), that `options['offset_time'][0]` will always have a value, either the default of None, or what it has been set to. But, I haven't tested this theory.

dianakhuang: I also thought that, but apparently that isn't the real behavior. I've at least tested to make sure that this code runs without the option set locally.

robrap: Got it. This was just a nit/question anyway. Please ignore at this point.

PR #3721, opened 2022-12-14
feat: send contentful data to lightcast to get EMSI skills

AliAdnanSohail:
[Add ability to read data from Contentful and send to Lightcast for skills for Degrees](https://2u-internal.atlassian.net/browse/PROD-3025)
[Add ability to read data from Contentful and send to Lightcast for skills for Bootcamps](https://2u-internal.atlassian.net/browse/PROD-3061)

## Description
We need to send Contentful data to Lightcast API to get skills tagging, So in this PR we are sending the data we are getting from contenful to Taxonomy Connector which will send that data to EMSI API and save skills for that product

## Test instruction
To test this we need to go on Course page from discovery admin, and click on refresh_skills button for the desired course. 
It will get data from contentful and extract skills from that data

PR #3733, opened 2022-12-22
feat: send degrees contentful data to taxonomy for skills extraction

AliAdnanSohail:
## Description:

In the follow up this PR https://github.com/openedx/course-discovery/pull/3721. We need to send degrees contentful data to Lightcast to extract skills as well. In this PR we will get aggregated data from contentful and send to lightcast for skills extraction

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_contentful_utils.py
+++ b/course_discovery/apps/course_metadata/tests/test_contentful_utils.py
@@ -124,7 +124,7 @@ class TestContentfulUtils(TestCase):
         self.assertDictEqual(
             transformed_data, mock_bootcamp_response.bootcamp_transformed_data)
 
-    def test_get_aggregated_data_from_contentful_data(self):
+    def test_get_aggregated_data_from_contentful_data__bootcamp(self):
         mock_bootcamp_response = MockContentfulBootcampResponse()
         expected_data = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit Lorem ipsum dolor sit amet, ' \
                         'consectetur adipiscing elitLorem ipsum dolor sit amet, consectetur adipiscing elitLorem ' \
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/tests/test_contentful_utils.py
```diff
@@ -124,7 +124,7 @@ def test_transform_bootcamp_contentful_data(self, *args):
         self.assertDictEqual(
             transformed_data, mock_bootcamp_response.bootcamp_transformed_data)
 
-    def test_get_aggregated_data_from_contentful_data(self):
+    def test_get_aggregated_data_from_contentful_data__bootcamp(self):
```
ansabgillani, line 127: Can we use `ddt` here instead?

course_discovery/apps/course_metadata/tests/test_contentful_utils.py
```diff
@@ -145,6 +145,24 @@ def test_get_aggregated_data_from_contentful_data(self):
         assert get_aggregated_data_from_contentful_data(
             mock_bootcamp_response.bootcamp_transformed_data, 'test-uuid') == expected_data
 
+    def test_get_aggregated_data_from_contentful_data__degree(self):
```
ansabgillani, line 148: Same as above, can we use `ddt` to differentiate degrees and bootcamps?

ansabgillani (approved):

course_discovery/apps/course_metadata/tests/test_contentful_utils.py
AliAdnanSohail: Actually IMO ddt should be used where we are testing a method for some kind of same type but with different input values,
But if type is different then a separate test make more sense to me and make test simpler and readable. Although it make code lengthy but looks simpler, we can get rid of if conditions in test cases

course_discovery/apps/course_metadata/tests/test_contentful_utils.py
AliAdnanSohail: same as above

DawoudSheraz (approved):

Ali-D-Akbar (approved):

PR #3749, opened 2023-01-04
feat: add display_on_org_page field

bseverino:
Now that degree about pages live on edX, we should have the ability to display degrees on organization pages. However, not all organizations want to link their degree brand to their edX brand. Add a field at the degree level in order to determine whether it should be displayed on the organization page.

Internal ticket: [WS-3648](https://2u-internal.atlassian.net/browse/WS-3648)


Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -341,6 +341,11 @@ def product_external_status(self):
             return getattr(self.additional_metadata, 'product_status', None)
         return None
 
+    @property
+    def product_display_on_org_page(self):
+        # Only courses display on organization pages
```
Ali-D-Akbar, line 346: Does this also include bootcamp and exec-ed?

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -507,6 +512,14 @@ def product_blocked_in(self):
             return get_location_restriction(self.location_restriction)
         return ALGOLIA_EMPTY_LIST
 
+    @property
+    def product_display_on_org_page(self):
```
Ali-D-Akbar, line 516: Why do we need to add this in Program Proxy model? Isn't this field for courses only?

Ali-D-Akbar (commented):
Left a bunch of questions.

course_discovery/apps/course_metadata/algolia_models.py
bseverino: Nope, currently only courses (not sure if that will change in the future)

course_discovery/apps/course_metadata/algolia_models.py
bseverino: The fields need to be consistent between both models - essentially this is only going to be used for degrees for now, but I want to make sure that the field makes sense for all products in this case

bseverino:
@Ali-D-Akbar Aside from rebasing, is there anything else needed from me here?

course_discovery/apps/course_metadata/algolia_models.py
Ali-D-Akbar: (Oh my, I questioned the wrong product. I wanted to ask regarding courses since this field is only added in degrees only but your response also answers this question too 👍)

Ali-D-Akbar:
> @Ali-D-Akbar Aside from rebasing, is there anything else needed from me here?

@bseverino I think you're good to go. I'd ask a team member as it is better to have an extra pair of 👀.

DawoudSheraz (approved):
Please ensure CI is green to make this ready for merge

bseverino:
@DawoudSheraz Still a couple failing tests but everything passes locally. Are these known to fail intermittently?

bseverino:
Can this one be rerun? https://github.com/openedx/course-discovery/actions/runs/3988359032/jobs/6839407233

PR #3758, opened 2023-01-09
docs: add ADR for FieldTracker

ansabgillani:
[PROD-3067](https://2u-internal.atlassian.net/browse/PROD-3067)

### DIFF
--- a/docs/decisions/0021-exclude-from-search-and-seo-fields.rst
+++ b/docs/decisions/0021-exclude-from-search-and-seo-fields.rst
@@ -1,4 +1,4 @@
-18. Exclude From Search And SEO Fields
+21. Exclude From Search And SEO Fields
 ------------------------------------------------------------------
 
 Status
--- a/docs/decisions/0022-introducing-field-trackers.rst
+++ b/docs/decisions/0022-introducing-field-trackers.rst
@@ -1,0 +1,91 @@
+22. Tracking data changes using Field Trackers in Discovery
+=============================================================
+
+Status
+=======
+
+Accepted
+
+Context
+========
+
+Discovery provides data to various services in Open edX ecosystem, including but not limited to Studio, E-commerce, and frontend marketing site. The current APIs in Discovery provide all the products with some limited filtering capabilities.
+With a large data set in Discovery, the time to fetch the data is large and can cause bottleneck in many different areas of Open edX infrastructure.
+There is no way to identify which products have been changed over a period of time. Therefore, the services have to fetch all the information from Discovery to utilize data or keep data in sync.
+
+Decision
+=========
+
+There are two capabilities being added in Discovery to overcome the above challenges:
+
+* Add timestamp filter capabilities in courses and programs APIs.
+* Introduce a new field, **data_modified_timestamp**, in Course and Program models that contains the timestamp a product was changed.
+
+timestamp filter
+-----------------
+
+The timestamp filter will be available as query parameter **timestamp**. The timestamp will take date or datetime in isoformat **YYYY-MM-DDTHH:MM:SS**. The timestamp filter will list the products that have been modified since the provided timestamp.
+
+With the introduction of timestamp-based query, the services will have capability to fetch data changed in a specified duration. Resultantly, there will be lesser data to fetch, return and consume for every discovery API call, reducing the processing and time cost considerably.
+
+Field trackers
+----------------
+
+A challenge with data_modified_timestamp field is identifying how to track changes in model and its related objects. During the investigation, a utility class `FieldTracker` built inside `django-model-utils` was discovered. When added to a model, it will track all modification to it's values until `save()` is called. It has capabilities of tracking all modifications, such as old and new values, and if a value has been changed or not.
+
+FieldTracker will be added to select models inside various course-discovery apps. The `save()` method of Course and Program models will be overridden, checking if any change has been observed as well as editable `ForeignKeys`. If there is a change, the `data_modified_timestamp` will be updated to the current timestamp.
+
+Lets consider the following example. In our `couse_discovery.apps.course_metadata.models` we have a `Course` Model. The changes mentioned above will look similar to following:
+
+
+.. code-block::
+
+    from datetime import datetime
+    from model_utils import FieldTracker
+
+
+    class Course(TimeStampedModel):
+
+        field_tracker = FieldTracker()
+        data_modified_timestamp = models.DateTimeField()
+
+        @property
+        def has_changed(self, external_fields):
+            change = field_tracker.has_changed() and len(field_tracker.changed())
+            for field in external_fields:  # Foreign keys to check
+                change = change and field.has_changed()
+            return change
+
+        def save(self, *args, **kwargs):
+            if self.has_changed([self.partner, self.course_run]):
+                self.data_modified_timestamp = datetime.now()
+            super().save(*args, **kwargs)
+
+
+Implications
+=============
+
+* Diamond Problem - Consider a Program that has two courses. If a Program's data is altered by a course query, should the modifications be reflected in other courses from the same program?
+* Child-Parent Relationship - Consider a Program having courses. If a Program's data is altered, should it reflect it in course? Or vice versa?
+
+Future Improvements
+=====================
+
+* The field tracker solution does not work well with Django ORM update() method because the method does not trigger save(). Discovery course API use update method on various linked objects. That would mean updating API to use save() instead of update to trigger last modified update flow
+* Directly changing the related objects of Course or Program (via admin or ORM) does not update last modified timestamp. There will be a need to add pre_save signals for desired models to allow changing last modified timestamp for Courses and Programs when their related objects are changed directly.
+
+Alternates Considered
+======================
+
+* All of this functionality can be added inside a `pre_save` signal. However, overriding the `save()` method seems like a straightforward way.
+* Most models in the discovery service inherit a `TimeStampedModel`, which include a modified field. However, this field is updated everytime the `save()` function is called, even if there is no change.
+* django-simple-history provides a capability to query Django object history. However, the history is only for main object and does not apply to changes in related objects. To use django-simple-history for tracking changes, there would be a need to add history objects for all of the related models for Course and Program. Also, the history query makes an additional DB call to get object which can have its performance aspects in the long run.
+
+
+References
+============
+
+* https://django-model-utils.readthedocs.io/en/latest/utilities.html#field-tracker
+* https://ilovedjango.com/django/models-and-databases/tips/field-tracker/
+* https://stackoverflow.com/questions/36600293/django-tracking-if-a-field-in-the-model-is-changed
+* https://django-simple-history.readthedocs.io/en/latest/querying_history.html


docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,37 @@
+18. Introducing Field Trackers for Programs and Courses in Discovery
```
DawoudSheraz: nit: Tracking data changes using Field Trackers in Discovery

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,37 @@
...
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Discovery provides data to various services in openedX ecosystem and  prospectus is one of those services. The current build flow of prospectus fetches all the information when deploying to production, regardless if the data has changed or not. With increasing data in discovery, the time to fetch the data is increasing and causes a bottleneck in prospectus deployment.
```
DawoudSheraz: We can mention ecommerce and Studio here. Rather than prospectus, explain the importance of discovery data for marketing. The prospectus is private/internal and not accessible to a larger audience.

DawoudSheraz: Additionally, the pain point here is to be able to identify which products have changed. And with large data in Discovery, fetching all the data is expensive.

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,37 @@
...
+
+It was decided mutually by Phoenix and Discover squad to add a timestamp-based query that will only fetch data changed in a specified duration. Resultantly, we will have lesser data to fetch, return and consume for every discovery API call, reducing the processing and time cost considerably.
```
DawoudSheraz: - These are internal details and won't be needed for ADR
- Make sure to add modified date filtering in Alternatives considered section as well

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,37 @@
...
+
+Potential Issues to be catered
```
DawoudSheraz: Rename this to Impact and move it after Decision.

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,37 @@
...
+----------------------------
+
+* Diamond Problem - Consider a Program that has two courses. If we change the Program's data by changing it's course's data, should the modifications be reflected in the other course?
+* Child-Parent Relationship - Consider a Program having courses. If we change the program data, should it reflect it in course? Or vice versa?
```
DawoudSheraz: nit: Rather than using `we`, explain it differently. For instance, if the program data is changed, should it be reflected under course changes too?

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,37 @@
...
+
+Decision
+--------
+
+During our investigation, we found that we have a `FieldTracker` class built inside `django-model-utils` (configured for edx-platform, course-discovery, etc.). When added to a model, it will track all modification to it's values until `save()` is called. It has capabilities of tracking all modifications, such as old and new values, and if a value has been changed or not.
+
+It, along with `last_modified_timestamp` will be added to all models inside course-discovery repository. We will override the `save()` method of the model checking if any change has been observed as well as editable `ForeignKeys`. If there is a change, we will update the `last_modified_timestamp` to the current timestamp.
```
DawoudSheraz: This can use some more context, specially w.r.t M2M fields

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,66 @@
...
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Discovery provides data to various services in openedX ecosystem and  prospectus is one of those services. The current build flow of prospectus fetches all the information when deploying to production, regardless if the data has changed or not. With increasing data in discovery, the time to fetch the data is increasing and causes bottleneck in many different areas of edx infrastructure. Moreoever, it is becoming increasingly hard to identify the products undergoing changes. The services that will feel the impact by these trackings include prospectus, ecommerce and studio.
```
robrap: Prospectus is a private project. You should either make this more generic, referring to any marketing service, and/or clarify when you are providing a 2U specific example.

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,66 @@
...
+                change = change and field.has_changed()
+        return change
+
+    def save(self, *args, **kwargs):
+        if self.has_changed():
+            self.last_modified_field = datetime.now()
+        super().save(*args, **kwargs)
+```
+
+Impact
```
robrap: I don't know if what I'm about to say would be related to a different column or not, but while you are thinking about this, I wonder if we should consider the ability to store a timestamp coming from another system? For example, the catalog changed events coming from Studio will soon have a timestamp that will match the published timestamp from Mongo in Studio. Would discovery ever want to capture/use this timestamp in addition to or in place of the time it happened to process the event?

robrap (commented):
In general, improving performance sounds great. I don't have enough experience with these tools to recommend the best implementation. I did note a couple of things. Thanks.

docs/decisions/0018-introducing-field-trackers
ansabgillani: Agreed to it. Fixed it accordingly.

robrap: I don't see the fix, but maybe that will come soon (i.e. tomorrow or later).

docs/decisions/0018-introducing-field-trackers
```diff
@@ -0,0 +1,66 @@
...
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Discovery provides data to various services in openedX ecosystem and  prospectus is one of those services. The current build flow of prospectus fetches all the information when deploying to production, regardless if the data has changed or not. With increasing data in discovery, the time to fetch the data is increasing and causes bottleneck in many different areas of edx infrastructure. Moreoever, it is becoming increasingly hard to identify the products undergoing changes. The services that will feel the impact by these trackings ecommerce and studio, as well as any marketing services.
```
DawoudSheraz: remove prospectus info

DawoudSheraz: The context should be re-written to focus on the benefits of adding the time tracker filtering capability and that how the services external to Discovery can benefit from it.

docs/decisions/0018-introducing-field-trackers
DawoudSheraz: We can discuss the benefits of adding that field too, but that would be a different action item altogether.

docs/decisions/0018-introducing-field-trackers.rst
```diff
@@ -0,0 +1,89 @@
...
+=====================
+
+* The field tracker solution does not work well with Django ORM update() method because the method does not trigger save(). Discovery course API use update method on various linked objects. That would mean updating API to use save() instead of update to trigger last modified update flow
+* Directly changing the related objects of Course or Program (via admin or ORM) does not update last modified timestamp. There will be a need to add pre_save signals for desired models to allow changing last modified timestamp for Courses and Programs when their related objects are changed directly.
+
+Alternates Considered
+======================
+
+* All of this functionality can be added inside a `pre_save` signal. However, overriding the `save()` method seems like a straightforward way.
+* Most models in the discovery service inherit a `TimeStampedModel`, which include a modified field. However, this field is updated everytime the `save()` function is called, even if there is no change.
```
AliAdnanSohail: Shouldn't we consider this option? we can avoid from many manual changes. If there is no change and object is calling its `save()` method, that means any activity is performed on that object so if modified is changed we can use that as well. WDYT? @DawoudSheraz

zawan-ila (approved):

docs/decisions/0018-introducing-field-trackers.rst
DawoudSheraz: Not quite. For instance, during RCM flow, the course and course run objects are saved regardless they are changed or not.

AliAdnanSohail (approved):

docs/decisions/0018-introducing-field-trackers.rst
```diff
@@ -0,0 +1,89 @@
...
+
+* Add timestamp filter capabilities in courses and programs APIs.
+* Introduce a new field, **last_modified_timestamp**, in Course and Program models that contains the timestamp a product was changed.
+
+timestamp filter
+-----------------
+
+The timestamp filter will be available as query parameter **timestamp**. The timestamp will take date or datetime in isoformat **YYYY-MM-DDTHH:MM:SS**. The timestamp filter will list the products that have been modified since the provided timestamp.
+
+With the introduction of timestamp-based query, the services will have capability to fetch data changed in a specified duration. Resultantly, there will be lesser data to fetch, return and consume for every discovery API call, reducing the processing and time cost considerably.
```
ashultz0: since this is resting on the performance of field tracker we should validate that processing and time actually go down before leaping. Field tracker is the sort of object about objects whose performance could be arbitrarily bad.

docs/decisions/0018-introducing-field-trackers.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Field trackers
+----------------
+
+A challenge with last_modified_timestamp field is identifying how to track changes in model and its related objects. During the investigation, a utility class `FieldTracker` built inside `django-model-utils` was discovered. When added to a model, it will track all modification to it's values until `save()` is called. It has capabilities of tracking all modifications, such as old and new values, and if a value has been changed or not.
+
+FieldTracker will be added to select models inside various course-discovery apps. The `save()` method of Course and Program models will be overridden, checking if any change has been observed as well as editable `ForeignKeys`. If there is a change, the `last_modified_timestamp` will be updated to the current timestamp.
```
ashultz0: The documentation for field tracker is silent about what happens when models change. Would field tracker cause migration headaches? If you use field tracker across a migration, does it break?

docs/decisions/0018-introducing-field-trackers.rst
```diff
@@ -0,0 +1,89 @@
...
+With a large data set in Discovery, the time to fetch the data is large and can cause bottleneck in many different areas of Open edX infrastructure.
+There is no way to identify which products have been changed over a period of time. Therefore, the services have to fetch all the information from Discovery to utilize data or keep data in sync.
+
+Decision
+=========
+
+There are two capabilities being added in Discovery to overcome the above challenges:
+
+* Add timestamp filter capabilities in courses and programs APIs.
+* Introduce a new field, **last_modified_timestamp**, in Course and Program models that contains the timestamp a product was changed.
```
robrap: It sounds like this won't be capturing every type of modification if you are only tracking certain models. Also, I imagine you already have a generically named field for modified timestamp. Maybe you could make this name a bit more specific for what is actually being tracked? Based on the rest of the sentence, even something like `product_modified_timestamp`, or whatever makes the most sense. You should also add clear documentation of what is being tracked (and maybe what isn't) in this field.

docs/decisions/0018-introducing-field-trackers.rst
robrap: [inform] It looks like there was one previous usage, but unfortunately it is in Teams, which probably doesn't get a ton of usage. New Relic could confirm that for you. See https://github.com/openedx/edx-platform/blob/97a1fee393dd4236339cbe4ae8c8fff8aa1c0ecc/lms/djangoapps/teams/models.py#L148

docs/decisions/0018-introducing-field-trackers.rst
DawoudSheraz: Right. We will do the performance testing. We have not seen any issues with save performance so far.

docs/decisions/0018-introducing-field-trackers.rst
DawoudSheraz: FieldTracker won't be a model/DB field. It is an object in a model that tracks the changes from last save and resets its context when save() is called.

docs/decisions/0018-introducing-field-trackers.rst
DawoudSheraz: Right. The last_modified_timestamp field is used Course & Program models because the APIs for these models are used by prospectus to fetch the content information. I will update ADR to be more specific to using this field for Course and Program.

robrap: Ah. Sorry, I was confused. I thought there was something tracking across models.

Course is already a TimeStampedModel: https://github.com/openedx/course-discovery/blob/47864da2c2c0072cc1d5c48f92ec4f64ee61ee9f/course_discovery/apps/course_metadata/models.py#L1086. I think you were saying that the TimeStampedModel modified date will change even when the course does not change, correct, or otherwise you wouldn't need a new field. I think `last_modified_timestamp` is a longer, but still very generic name, and can't easily be differentiated from the TimeStampedModel field. What would a better name be? Even `data_modified_timestamp` could be documented to note that sometimes the row is updated without the data actually changing, and this field would differentiate between the two.

You would know better than I do what the right name should be. I'm just pointing out a flaw in the current name.

DawoudSheraz: 👍🏽

docs/decisions/0018-introducing-field-trackers.rst
ashultz0: ah got it, it only tracks in the current session. No concerns then.

docs/decisions/0018-introducing-field-trackers.rst
robrap: @DawoudSheraz: 
1. Is it possible to simply not save unless the data changes, and avoid the new field, so the `TimeStampedModel` modified field is enough as it is?
2. Separate but related, you may want to consider using the timestamp from the event when the data changes due to an event. This was Dave O's plan for these timestamps, but this is up to you.

DawoudSheraz: > Separate but related, you may want to consider using the timestamp from the event when the data changes due to an event. This was Dave O's plan for these timestamps, but this is up to you.

The modified timestamp applies to API, Loader, and RCM flow altogether. The event you are talking about is only available during RCM flow.

> Is it possible to simply not save unless the data changes, and avoid the new field, so the TimeStampedModel modified field is enough as it is?

As stated before, the timestamp update will be done in all different data loading flows. Not performing the save operation can have unintended consequence across different channels (like handling API responses).

robrap: > The modified timestamp applies to API, Loader, and RCM flow altogether. The event you are talking about is only available during RCM flow.

Yes. But it would be possible for the RCM flow to send the timestamp as an optional argument to be used for storage. However, it seems this isn't urgent, so I'll leave it be for now. Just know that it is an option here, and for future events, so the modified date could match the real timestamp, rather than the time that the event gets delivered, which may be long afterwards if it is delayed or re-run.

> As stated before, the timestamp update will be done in all different data loading flows. Not performing the save operation can have unintended consequence across different channels (like handling API responses).

Thanks for repeating this, and sorry about that. Just double checking.

PR #3760, opened 2023-01-11
fix: add spaces in each word separately using rich_text_to_plain_text

Ali-D-Akbar:
Fixes words concatenated together without spaces using the `rich_text_to_plain_text`.

### DIFF
--- a/course_discovery/apps/course_metadata/contentful_utils.py
+++ b/course_discovery/apps/course_metadata/contentful_utils.py
@@ -150,7 +150,7 @@ def extract_plain_text_from_rich_text(rich_text_dict):
     text_list = []
     for key, value in rich_text_dict.items():
         if key == 'value':
-            text_list.append(value)
+            text_list.append(value.strip())
         elif isinstance(value, dict):  # recursive call if the value is a nested dict
             text_list += extract_plain_text_from_rich_text(value)
         elif isinstance(value, list):  # recursive call if the value is a nested list
@@ -164,7 +164,7 @@ def rich_text_to_plain_text(rich_text):
     Converts rich text from Contentful to plain text.
     Plain text resides in the dict with 'value' as dict key name.
     """
-    return ''.join(extract_plain_text_from_rich_text(rich_text))
+    return ' '.join(extract_plain_text_from_rich_text(rich_text))
 
 
 def get_modules_list(entry):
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -164,7 +164,7 @@ def rich_text_to_plain_text(rich_text):
     Converts rich text from Contentful to plain text.
     Plain text resides in the dict with 'value' as dict key name.
     """
-    return ''.join(extract_plain_text_from_rich_text(rich_text))
+    return ' '.join(' '.join(extract_plain_text_from_rich_text(rich_text)).split())
```
DawoudSheraz: Why do the split and then join afterward?

AliAdnanSohail: Also can't we do the join thing in `extract_plain_text_from_rich_text` method?

Ali-D-Akbar: @DawoudSheraz This is the before vs after comparison:
```
>>> ''.join(['hello', ' world', 'something', '  yess      ', 'no'])
'hello worldsomething  yess      no'
>>> ' '.join(' '.join(['hello', ' world', 'something', '  yess      ', 'no']).split())
'hello world something yess no'
```

Ali-D-Akbar: @AliAdnanSohail yes but requires more than one-line changes since I have to revamp the approach to extract plain text.

AliAdnanSohail: You can do that change IMO, Otherwise each time developer have to add `<space>` where he is using this method. So better approach is, this method should return value with spaces

DawoudSheraz (approved):

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -351,7 +351,7 @@ def fetch_and_transform_bootcamp_contentful_data():
     return transformed_bootcamp_data
 
 
-def get_aggregated_data_from_contentful_data(data, product_uuid):
+def get_aggregated_data_from_contentful(data, product_uuid):
```
AliAdnanSohail: Actually I choose this name for this method. The reason was, this method is not actually getting data from contentful service, this method is just using contentful data to get aggregated data from that. 
Name `get_aggregated_data_from_contentful` seems like we are getting aggregated data from contentful service directly.

DawoudSheraz: In that case, the name can be transform_contentful_data.

AliAdnanSohail: I think better is `aggregate_contentful_data`

ansabgillani (approved):

AliAdnanSohail (approved):

PR #3764, opened 2023-01-17
fix: fix google drive download url

AliAdnanSohail:
## Description
Before google drive API implementation we were converting google drive URL to downloadable URL, but now we don't need that functionality as we can directly download image from google drive URL.


DawoudSheraz (approved):
nit: remove double underscore from test names

AfaqShuaib09 (approved):

AliAdnanSohail:
@DawoudSheraz actually double underscore added intentionally to distinguish between method name and scenario we are testing in the test case

PR #3766, opened 2023-01-18
feat: add product source API for publisher access

ansabgillani:
[PROD-3106](https://2u-internal.atlassian.net/browse/PROD-3106)

This PR covers updates courseViewSet create method to make product_source required for creating a course. We are passing the `product_souce.slug` in the request data to create a course.
- This PR also updates the CSVLoader to pass the `product_source.slug` in the request data to create a course if the course is not already created else it will update the course product_source using `add_product_source` method.

- This PR also updates the course and course_run factories to create a product_source for the course and course_run and update the tests accordingly.
- This PR also update CSV loader tests to also add product source to courses

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1282,7 +1282,16 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
     geolocation = GeoLocationSerializer(required=False, allow_null=True)
     location_restriction = CourseLocationRestrictionSerializer(required=False)
     in_year_value = ProductValueSerializer(required=False)
-    product_source = SourceSerializer(required=False)
+    product_source = serializers.SlugRelatedField(required=False, slug_field='slug', queryset=Source.objects.all())
+
+    def to_representation(self, instance):
+        """
+        Conversion of the source slug to the source serializer data
+        """
+        representation = super().to_representation(instance)
+        if 'product_source' in representation:
+            representation['product_source'] = SourceSerializer(Source.objects.get(slug=representation['product_source'])).data  # pylint: disable=line-too-long
+        return representation
 
     def get_organization_logo_override_url(self, obj):
         logo_image_override = getattr(obj, 'organization_logo_override', None)
@@ -1342,7 +1351,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
             'enterprise_subscription_inclusion', 'geolocation', 'location_restriction', 'in_year_value',
             'product_source', 'data_modified_timestamp', 'excluded_from_search', 'excluded_from_seo'
         )
-        read_only_fields = ('data_modified_timestamp',)
+        read_only_fields = ('enterprise_subscription_inclusion', 'product_source', 'data_modified_timestamp')
         extra_kwargs = {
             'partner': {'write_only': True}
         }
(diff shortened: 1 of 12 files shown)


Ali-D-Akbar (approved):

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -228,6 +233,11 @@ def create(self, request, *args, **kwargs):
         organization = Organization.objects.get(key=course_creation_fields['org'])
         course.authoring_organizations.add(organization)
 
+        if course_creation_fields['product_source']:
+            source = Source.objects.get(slug=course_creation_fields['product_source'])
+            course.product_source = source
+            course.save()
```
DawoudSheraz: Can't we assign the source before the serializer is saved? Adding an additional save call will have its own consequences.

DawoudSheraz: Also, the source is now required and this check does not mandate this.

course_discovery/apps/api/v1/views/sources.py
```diff
@@ -0,0 +1,17 @@
...
+
+from course_discovery.apps.api import serializers
+from course_discovery.apps.api.cache import CompressedCacheResponseMixin
+from course_discovery.apps.course_metadata.models import Source
+
+
+class SourceViewSet(CompressedCacheResponseMixin, viewsets.ReadOnlyModelViewSet):
+    """ Source resource. """
+
+    permission_classes = (IsAuthenticated,)
```
DawoudSheraz: Maybe add a new permission to check if a user is staff or editor, to avoid leaking source information because this is only checking authentication and not AuthZ. Also, add proper authentication_classes unless already present in Mixin.

course_discovery/apps/api/v1/views/sources.py
```diff
@@ -0,0 +1,17 @@
...
+    serializer_class = serializers.SourceSerializer
+    lookup_field = 'slug'
+
+    def get_queryset(self):
+        return Source.objects.all()
```
DawoudSheraz: This can be converted to `queryset = Source.objects.all()`

course_discovery/apps/api/serializers.py
```diff
@@ -1282,7 +1282,15 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
...
-    product_source = SourceSerializer(required=False)
+    product_source = serializers.SlugRelatedField(required=False, slug_field='slug', queryset=Source.objects.all())
+
+    def to_representation(self, instance):
+        """
+        Conversion of the source slug to the source serializer data
+        """
+        representation = super().to_representation(instance)
+        representation['product_source'] = SourceSerializer(Source.objects.get(slug=representation['product_source'])).data  # pylint: disable=line-too-long
+        return representation
```
DawoudSheraz: This will hit DB for an additional call per course and can have long-term performance consequences.

DawoudSheraz (approved):
A small comment, otherwise looks good.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -349,6 +349,7 @@ def _create_course_api_request_data(self, data, course_type, course_run_type_uui
         which will be used as input for course creation via course api.
         """
         pricing = self.get_pricing_representation(data['verified_price'], course_type)
+        product_source = self.product_source.slug if self.product_source else None
```
DawoudSheraz, line 357: Isn't product source present by default on CSV loader?

AfaqShuaib09: nope. We are previously adding it to the course using `add_product_source` method

PR #3769, opened 2023-01-20
feat: add waffle switch to bypass end date updated check in LMS data loader

DawoudSheraz:
### Description
- Add a waffle switch `course_metadata.bypass_lms_data_loader__end_date_updated_check` that overrides the end date changed check in LMS Data loader and ensures that the upgrade deadline is updated in Discovery and info is pushed to eCommerce if applicable. This flag will be turned on if there are course runs whose end date information is up to date in Discovery but eCommerce does not have updated dates. This usually happens if:
- End date is changed via admin to made same as Studio's End date
- Data unrelated Ecommerce API call failure (rate limit, service unavailable, configuration mismatch)

### DIFF
--- a/course_discovery/apps/course_metadata/toggles.py
+++ b/course_discovery/apps/course_metadata/toggles.py
@@ -1,0 +1,21 @@
+"""
+Toggles for course metadata app.
+"""
+
+from edx_toggles.toggles import WaffleSwitch
+
+# .. toggle_name: course_metadata.bypass_lms_data_loader__end_date_updated_check
+# .. toggle_implementation: WaffleSwitch
+# .. toggle_default: False
+# .. toggle_description: Enable to bypass end date updated checks in LMS Data loader and make sure that
+#     upgrade deadline information is updated in Discovery and then pushed to ecommerce if applicable.
+# .. toggle_use_cases: open_edx
+# .. toggle_creation_date: 2022-01-20
+# .. toggle_target_removal_date: None
+# .. toggle_tickets: PROD-3121
+# .. toggle_warning: This flag should only be turned on if either calls to ecommerce failed or end dates were updated
+#     via admin, leaving the upgrade deadline information inconsistent across Discovery and ecommerce.  Once LMS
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/toggles.py
```diff
@@ -0,0 +1,21 @@
...
+# .. toggle_default: False
+# .. toggle_description: Enable to bypass end date updated checks in LMS Data loader and make sure that
+# upgrade deadline information is updated in Discovery and then pushed to ecommerce if applicable.
+# .. toggle_use_cases: open_edx
+# .. toggle_creation_date: 2022-01-20
+# .. toggle_target_removal_date: None
+# .. toggle_tickets: PROD-3121
+# .. toggle_warning: This flag should only be turned on if either calls to ecommerce failed or end dates were updated
+# via admin, leaving the upgrade deadline information inconsistent across Discovery and ecommerce.  Once RCM's LMS
+# data loader runs and dates are in sync, turn off this flag.
```
timmc-edx: I think you'll need this additional whitespace at the beginnings of hard-wrapped lines in order to keep all the annotations intact:
```suggestion
# .. toggle_description: Enable to bypass end date updated checks in LMS Data loader and make sure that
#   upgrade deadline information is updated in Discovery and then pushed to ecommerce if applicable.
# .. toggle_use_cases: open_edx
# .. toggle_creation_date: 2022-01-20
# .. toggle_target_removal_date: None
# .. toggle_tickets: PROD-3121
# .. toggle_warning: This flag should only be turned on if either calls to ecommerce failed or end dates were updated
#   via admin, leaving the upgrade deadline information inconsistent across Discovery and ecommerce.  Once RCM's LMS
#   data loader runs and dates are in sync, turn off this flag.
```
ref: https://github.com/openedx/edx-toggles/blob/master/docs/how_to/documenting_new_feature_toggles.rst#multi-line-annotations

Ali-D-Akbar (approved):

AliAdnanSohail (approved):

PR #3771, opened 2023-01-22
feat: introduce field trackers in Course and Program models

ansabgillani:
#### Add Field Tracker for various important models

FieldTracker is a model utility offered in django-model-utils package (already configured for most services in edX-infrastructure), keeping track of all changes prior to database commit. For timestamp queries, we are using FieldTracker as a reference to identify if the values in the model are changing or not.

#### Add has_model_changed

We have further introduced a has_model_changed utility in course_metadata that essentially checks if the model and the provided Foreign objects have seen any changes. We also have added excluded_fields to identify any fields that we do not want to identify as a potential change in the model object.

#### Adding DateTimeFilter in Program/Course Viewsets

We are using DateTimeFilter in Program and Course filters, which essentially use the timestamp parameter to check if any course exists that has been modified after that timestamp. Both courses (/api/v1/courses) and programs (/api/v1/programs) have new query param `timestamp` to filter based on modified timestamp (YYYY-MM-DD or YYYY-MM-DDTHH:MM:SS)

#### Todo in a follow-up PR
- Add model methods and signals to cater to Foreign keys reverse updates. For instance, RCM does not update the timestamp because CourseRun object is saved before Course. By the time course is saved, there is nothing different on course run level
- There is a weird behavior on Publisher where updating course run data is updating data modified timestamp only on Non-draft version of the course. This is by-design because if the course run is published, course's official version is updated (https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/utils.py#L61)from draft version. That means the modified timestamp changes are reflected in non-draft/official version. That will be tricky to handle but it will be handled.
- Handle Program & Degree's M2M fields

#### Testing
- For courses, use publisher to edit and save the course. Verify the timestamp is updated for draft (and non-draft if applicable) and values are in sync. **For ExecEd & Bootcamps, Additionalmetadata changes will not update timestamp.** 
- For programs, edit the program via admin and verify timestamp updates
- Also verify that saving without any changes does not update the timestamp


course_discovery/apps/course_metadata/model_utils.py
```diff
@@ -0,0 +1,14 @@
+def has_model_changed(field_tracker, external_keys=[]):
```
Ali-D-Akbar: Docstring for the file

Ali-D-Akbar (changes requested):
Tests and description require some updated.
Also, please add some mechanism to test these changes on local and how this can be used to filter changed objects.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1196,6 +1283,11 @@ def _check_enterprise_subscription_inclusion(self):
     # subscription catalog, we need the id that is created on save to access the many-to-many fields
     # and then need to update the boolean in the record based on conditional logic
     def save(self, *args, **kwargs):
+        if self.last_modified_timestamp:
+            if self.has_changed:
+                self.last_modified_timestamp = datetime.datetime.now(pytz.UTC)
+        else:
```
DawoudSheraz: This needs a comment/docstring as to why double if is needed.

DawoudSheraz (commented):
Most changes make sense.  Although PR can use some extra unit tests for Fieldtracker and last modified testing

Ali-D-Akbar (approved):
Great work. Looks very promising to fix our underlying issue. 👍

AliAdnanSohail (approved):

zawan-ila (approved):

PR #3775, opened 2023-01-26
feat: add source model in admin and add new column for ofac restricted program types in Source model

AliAdnanSohail:
[Mark non-Bachelor Degree products as [OFAC restricted], and directly publish them on edx.org.](https://2u-internal.atlassian.net/browse/PROD-3022)


course_discovery/apps/course_metadata/admin.py
```diff
@@ -923,3 +922,11 @@ class GeolocationDataLoaderConfigurationAdmin(admin.ModelAdmin):
...
     """
     list_display = ('id', 'enabled', 'changed_by', 'change_date')
+
+
+@admin.register(Source)
+class SourceAdmin(admin.ModelAdmin):
+    """
+    Admin for GeolocationDataLoaderConfiguration model.
+    """
+    list_display = ('id', 'name', 'slug', 'description', 'ofac_restricted_program_types')
```
DawoudSheraz: Isn't ofac_restricted_program_types an M2M relation?

course_discovery/apps/course_metadata/models.py
```diff
@@ -432,6 +420,24 @@ def is_enterprise_catalog_program_type(cls, program_type):
...
+    """
+    Source Model to find where a course or program originated from.
+    """
+    name = models.CharField(max_length=255, help_text=_('Name of the external source.'))
+    slug = AutoSlugField(
+        populate_from='name', editable=True, slugify_function=uslugify, overwrite_on_add=False,
+        help_text=_('Leave this field blank to have the value generated automatically.')
+    )
+    description = models.CharField(max_length=255, blank=True, help_text=_('Description of the external source.'))
+    ofac_restricted_program_types = SortedManyToManyField(
```
Ali-D-Akbar, line 433: What does this field tells us in Source model?
Also just a nit: A helptext for this field can be useful.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -125,7 +125,6 @@ class ProductValueAdmin(admin.ModelAdmin):
 @admin.register(Course)
 class CourseAdmin(DjangoObjectActions, admin.ModelAdmin):
     form = CourseAdminForm
-    inline = (SourceInline,)
```
Ali-D-Akbar, line 128: Do we not need source in courseadmin anymore?

Ali-D-Akbar (commented):
Do we not need to update the SourceSerializer?
Just a bunch of comments before approval.

course_discovery/apps/course_metadata/admin.py
AliAdnanSohail: We don't need in the form of inline, simple foreign key will work fine to select any source from dropdown

AliAdnanSohail:
@Ali-D-Akbar - I don't think we need to update SourceSerializer, because we don't need to update these program types value from API

DawoudSheraz (approved):

Ali-D-Akbar (approved):

PR #3776, opened 2023-01-26
fix: geolocation serializer fixups

DawoudSheraz:
### Description
- Allow GeoLocation fields to be null. This is to accommodate GeoLocation data coming from Publisher. GeoLoc info on Publisher is optional. Related PR: https://github.com/openedx/frontend-app-publisher/pull/819
- Update courses serializer to skip geolocation handling if all fields are not present. All of the geoloc fields should be provided to keep the data consistent. 
- Fix the update geoloc behavior where it would always create a new geoloc object if any of the fields is edited on the publisher. This results in having a lot of orphan GeoLoc objects in DB. 
- UUID filtering flaky test fixup

### Testing
You would need publisher running for proper testing.
- Checkout this PR and publisher's PR https://github.com/openedx/frontend-app-publisher/pull/819 on your local
- Go to any course on the publisher. You should see Geolocation section on Course Edit section
- Do not add any values in any of the field. Update any other field of Course/Course run so that backend call is made. Ensure no geoLoc object is created
- Add value for all 3 geoloc fields and save/publish again. Ensure geoloc is created. If the course is already published, ensure that geoloc object is linked with both draft and non-draft versions of the course.
- Update any of geoloc fields and save again. Ensure that no new GeoLoc object is created and that existing GeoLoc object related to the course is updated.


marekwrobel:
Generally looks good, but I noticed something about the way you update `location_restriction` on the course instance. 

If I remember correctly, due to the nature of courses (as opposed to the programs), the way they are linked with `location_restiction` objects is that multiple courses may use the same `location_restriction`. 

That means, if on every update you create a new instance of  `location_restriction`, that's OK (it will only be updated for that one course). Then you check if the old `location_restriction` has any relations to any other courses, if not, its an orphan and it can be removed (that's how it works in the geotargeting bulk data loader).

But looking at this PR I noticed that the way `location_restriction` gets updated is that it updates it directly on that existing `location_restiction`. So I am concerned that if you have 2 courses pointing to the same `location_restriction`, and then update it for one of the courses, it will get updated for both of them. 

**Could your verify/check this?** 

Either way, since this PR is about geolocation (as opposed to `location_restriction`) this problem does not occur and you can update it directly on the course instance you are updating, so this change looks good to me.

marekwrobel (approved):

srwang (approved):

DawoudSheraz:
> Generally looks good, but I noticed something about the way you update `location_restriction` on the course instance.
> 
> If I remember correctly, due to the nature of courses (as opposed to the programs), the way they are linked with `location_restiction` objects is that multiple courses may use the same `location_restriction`.
> 
> That means, if on every update you create a new instance of `location_restriction`, that's OK (it will only be updated for that one course). Then you check if the old `location_restriction` has any relations to any other courses, if not, its an orphan and it can be removed (that's how it works in the geotargeting bulk data loader).
> 
> But looking at this PR I noticed that the way `location_restriction` gets updated is that it updates it directly on that existing `location_restiction`. So I am concerned that if you have 2 courses pointing to the same `location_restriction`, and then update it for one of the courses, it will get updated for both of them.
> 
> **Could your verify/check this?**
> 
> Either way, since this PR is about geolocation (as opposed to `location_restriction`) this problem does not occur and you can update it directly on the course instance you are updating, so this change looks good to me.

Judging from how the logic for create/update is for other Nested serializers, you are right that location_restriction gets updated instead of creating a new one. While I get your point, I am not sure how one location restriction can be linked with two different courses. When I talk about different courses, I am referring to courses with different keys and not draft and non-draft versions of same course. When course API is hit, if location restriction is not present on a course, a new object is created and associated with draft and non-draft versions of the same course. Now, if a different course has same restriction values, still a new object is created for that course (https://github.com/openedx/course-discovery/blob/6702719829f1d0cf4da7e877aa24288be750936a/course_discovery/apps/api/serializers.py#L1453-L1457).

AliAdnanSohail (approved):

Ali-D-Akbar (approved):

PR #3779, opened 2023-01-27
feat: update additionalMetadata serializer to include external_course_marketing_type

zawan-ila:
# [PROD-3125](https://2u-internal.atlassian.net/browse/PROD-3125)

## Description

This PR updates the additionalMetadata serializer to include external_course_marketing_type

### DIFF
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -9,7 +9,7 @@ from taxonomy.models import CourseSkills, ProgramSkill, Skill
 
 from course_discovery.apps.core.tests.factories import PartnerFactory, UserFactory, add_m2m_data
 from course_discovery.apps.core.tests.utils import FuzzyURL
-from course_discovery.apps.course_metadata.choices import ExternalProductStatus
+from course_discovery.apps.course_metadata.choices import ExternalCourseMarketingType, ExternalProductStatus
 from course_discovery.apps.course_metadata.models import *  # pylint: disable=wildcard-import
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
 
@@ -119,7 +119,7 @@ class AdditionalMetadataFactory(factory.django.DjangoModelFactory):
     course_term_override = FuzzyText()
     product_meta = factory.SubFactory(ProductMetaFactory, keywords=['test', 'test2'])
     product_status = ExternalProductStatus.Published
-    external_course_marketing_type = None
+    external_course_marketing_type = FuzzyChoice([name for name, __ in ExternalCourseMarketingType.choices])
 
     @factory.post_generation
     def facts(self, create, extracted, **kwargs):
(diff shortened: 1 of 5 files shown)


Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -1,4 +1,5 @@
-from random import randint
+from random import randint, choice
+from course_discovery.apps.course_metadata.choices import ExternalCourseMarketingType
```
DawoudSheraz: incorrect import order.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -120,6 +121,7 @@ class Meta:
     product_meta = factory.SubFactory(ProductMetaFactory, keywords=['test', 'test2'])
     product_status = ExternalProductStatus.Published
     external_course_marketing_type = None
+    external_course_marketing_type = choice(list(ExternalCourseMarketingType.values.keys()))
```
DawoudSheraz: Use factory boy ChoiceField to make this random on different objects rather than creating the randomization here. Plus, this is the incorrect way of creating random values. factory_boy has a special syntax for calling custom methods for values assignment https://factoryboy.readthedocs.io/en/stable/introduction.html#lazyfunction

DawoudSheraz (commented):
Left a few comments.

course_discovery/apps/course_metadata/tests/factories.py
zawan-ila: fixed

course_discovery/apps/course_metadata/tests/factories.py
zawan-ila: done

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -120,6 +120,7 @@ class Meta:
     product_meta = factory.SubFactory(ProductMetaFactory, keywords=['test', 'test2'])
     product_status = ExternalProductStatus.Published
     external_course_marketing_type = None
+    external_course_marketing_type = FuzzyChoice([name for name, __ in ExternalCourseMarketingType.choices])
```
DawoudSheraz, line 122: you don't need to loop on this. Just provide the choices.

zawan-ila: essentially we need a string value to assign to external_course_marketing_type but then it'd be a tuple....the tests would still pass i think but the semantics would be off....

zawan-ila: to handle the tuple properly the docs pass in a getter function to retrieve first element of tuple

zawan-ila: also this is the same pattern used in some earlier factories in this file

DawoudSheraz: The tuple is handled automatically. Just notice how other choices are also using tuple in Discovery.

AliAdnanSohail (approved):

DawoudSheraz (approved):

PR #3783, opened 2023-01-30
feat: Update to event-bus-kafka 3.7.1; use its management command directly

timmc-edx:
This removes the implementation of the `consume_events` management command
that we copied into course-discovery from event-bus-kafka, and now instead
imports the Command from the library.

As of event-bus-kafka 3.7.1, the functionality of the commands is
identical, with one difference: `EVENT_BUS_KAFKA_CONSUMERS_ENABLED` now
defaults to True.

Relevant changes in the library upgrade:

- A fix for the DB reconnection issue.
- Prepares for a breaking change that will remove the `offset_timestamp`
  parameter from `consume_indefinitely`. This is no longer as important
  since that is only called from the management command, but if this
  deletion is rolled back, the management command will need to be re-copied
  from event-bus-kafka to start using the new replay method
  `reset_offsets_and_sleep_indefinitely`.


course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -93,6 +94,9 @@ def handle(self, *args, **options):  # pragma: no cover
                 group_id=options['group_id'][0],
                 signal=signal,
             )
-            event_consumer.consume_indefinitely(offset_timestamp=offset_timestamp)
+            if offset_timestamp is None:
+                event_consumer.consume_indefinitely()
+            else:
+                event_consumer.reset_offsets_and_sleep_indefinitely(offset_timestamp=offset_timestamp)
```
timmc-edx: Author's note: The changes here are mostly from the following PRs:

- https://github.com/openedx/event-bus-kafka/pull/118
- https://github.com/openedx/event-bus-kafka/pull/119 (pending review at this moment)

...or we could just go ahead and import the Command directly from the library, which would remove our need to keep this synchronized. But maybe that's for a later PR.

course_discovery/apps/course_metadata/management/commands/consume_events.py
```diff
@@ -1,98 +1,8 @@
...
-        except Exception:  # pylint: disable=broad-except
-            logger.exception("Error consuming Kafka events")
+# Import the consumer Command from event-bus-kafka in order to expose it here.
+#
+# FIXME: This currently hardcodes our use of the Kafka implementation.
+#   https://github.com/openedx/openedx-events/issues/147 discusses ways
+#   we might make this configurable.
+
+# pylint: disable=unused-import
+from edx_event_bus_kafka.management.commands.consume_events import Command
```
robrap, line 8: I know it isn't being introduced here, but this makes the name `Command` stick out. It seems odd that we don't have a more specific name. Let's discuss for a moment before this gets locked in. Thoughts?

robrap (commented):
Looks good. Just one question.

course_discovery/apps/course_metadata/management/commands/consume_events.py
timmc-edx: That's actually the name we have to use for management commands: https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/

robrap: I forgot it is the module name that matters. :) Thanks.

robrap (approved):

robrap:
@timmc-edx: Looks like the branch needs to be updated. Let me know when this is ready to merge. Can you ensure this works on Stage and then ensure it gets pushed to Prod? Thank you!

robrap:
@timmc-edx: Also, can you update the how-to in Confluence: https://openedx.atlassian.net/wiki/spaces/AC/pages/edit-v2/3508699151? You could point to discovery after merge, but better yet, just add the import and fixme comment into the how-to.

timmc-edx:
Ready to merge, and I've updated the public wiki page.

PR #3787, opened 2023-02-02
feat: add management command to add course tags from csv

zawan-ila:
# [PROD-3126](https://2u-internal.atlassian.net/browse/PROD-3126)

## Description

This PR adds a management command to upload course tags from csv

## Testing

1. Add a BulkUploadCourseTagsConfig object in Admin using the csv. Also mark the enabled flag.
2. Run `./manage.py bulk_upload_tags`
3. Verify in admin or publisher that the tags have been updated

You can also pass the path of csv as argument to the command like `./manage.py bulk_upload_tags --csv_path=test.csv`


zawan-ila:
This work will be catered in a different PR

PR #3784, opened 2023-02-01
chore: update django-elasticsearch-dsl to latest commit to fix make upgrade

DawoudSheraz:
### Description
- Using django-elasticsearch-dsl from PyPI to fix make upgrade
- This package was previously added via a direct commit hash (https://github.com/django-es/django-elasticsearch-dsl/commits/0e92e01c6ef74d2fe329965deee5f4b25da7ec87). The hash included support for Django 32. The same change was released in [v7.2.1](https://github.com/django-es/django-elasticsearch-dsl/releases/tag/7.2.1) a few days later. This PR is updating to the latest version to get a dependency syntax error fix from https://github.com/django-es/django-elasticsearch-dsl/pull/435.
- This error started appearing because setuptools was updated to 67.0.0 and that version included some restricted requirements parsing https://setuptools.pypa.io/en/latest/history.html#v67-0-0
- Fix quality failures due to pylint update

PR #3789, opened 2023-02-03
feat: manually upgrade event-bus-kafka

rgraber:
Brings in an update to edx-event-bus-kafka that will emit the original metadata from the course_catalog_info_changed signal event.

Done manually while `make upgrade` is busted, pending https://github.com/openedx/course-discovery/pull/3784

### DIFF
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -363,7 +363,7 @@ edx-drf-extensions==8.2.0
     # via
     #   -c requirements/constraints.txt
     #   -r requirements/base.in
-edx-event-bus-kafka==3.7.1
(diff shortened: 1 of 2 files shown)


timmc-edx (approved):
Looks fine. Only transitive dependency difference is openedx-events >= 4.2.0, but that's already satisfied in this repo.

https://github.com/openedx/event-bus-kafka/compare/v3.7.1..v3.8.0#diff-ffb9a8aa35f8e58dd818a724713720da25a2847e19d36b4bbf5da7d5744415e0

dianakhuang (approved):

PR #3799, opened 2023-02-14
feat: remove empty tags form html content which were converted to hr tags

AliAdnanSohail:
## Description
We are cleaning html before adding it to our database, during cleaning it was converting `<p><b></b></p>` to `<hr />`. And we don't allow this tag in our html so we need to remove these empty tags


Ali-D-Akbar (approved):

DawoudSheraz (commented):
How does this render on Publisher? Can you verify if the new tag is rendering as expected?

AliAdnanSohail:
@DawoudSheraz I've checked this on Publisher and its showing a straight line where this tag is placed

DawoudSheraz (dismissed):

DawoudSheraz (approved):

DawoudSheraz (commented):
Please update PR title & description before merge.

PR #3803, opened 2023-02-16
feat: Exclude objects from search and SEO fields

marekwrobel:
Please read the ADR attached as part of this PR for more context. 

Two fields have been added to the Course and Program models:
- `excluded_from_search`
- `excluded_from_seo`

The first one takes part in the Algolia indexing logic. The second one will be passed to the frontend (Prospectus) during build, and pages that have this field set to TRUE will have a `noindex` meta tag added (Prospectus PR is still in-progress for this). 

More importantly, these fields can be edited through Django Admin for each `course-discovery` object, but also can be added in Contentful (for `Boot Camp Page` and `Degree Detail Page`). The Contentful value will override course-discovery one if its set to TRUE, but if its set to FALSE (or not defined), we will run our regular additional logic to determine whether we should index/hide this object.

If a field is set to False (or undefined), nothing really changes. But if it's set to true, it will be as easy way to force non-indexing (or hiding from SEO).


course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -360,7 +366,8 @@ def should_index(self):
                 self.partner.name == 'edX' and
                 self.availability_level and
                 bool(self.advertised_course_run) and
-                not self.advertised_course_run.hidden)
+                not self.advertised_course_run.hidden and
+                not self.excluded_from_search)
```
DawoudSheraz, line 380: shouldn't this be self.excluded_from_seo? or to be more clear, where would be the usage of exlcude_from_seo? Would it be on the frontend side?

docs/decisions/0018-exclude-from-search-and-seo-fields.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Status
+------
+
+In-Progress (February 2023)
+
+Context
+-------
+
+It is important to easily toggle items on and off from appearing in our internal search results. Currently, there have been a lot of conditions that would need to be met for a particular course (or program, item) to be indexed (in Algolia), but a simple interface for this functionality was needed.
```
DawoudSheraz: nit: toggle items from appearing in internal search results.

DawoudSheraz: nit: for a particular course or program to be indexed in Algolia

docs/decisions/0018-exclude-from-search-and-seo-fields.rst
```diff
@@ -0,0 +1,35 @@
...
+-------
+
+It is important to easily toggle items on and off from appearing in our internal search results. Currently, there have been a lot of conditions that would need to be met for a particular course (or program, item) to be indexed (in Algolia), but a simple interface for this functionality was needed. 
+Additionally, for some of the About Pages, it was important to easily hide them from SEO (add noindex meta tag).
+
+Decision
+--------
+
+Two fields have been added on Course and Program models:
+- excluded_from_search
```
DawoudSheraz, line 20: nit: An empty line is needed to render this list correctly in markdown.

course_discovery/apps/course_metadata/algolia_models.py
marekwrobel: Exactly - `excluded_from_seo` is only really needed on the frontend side for each template (to put the noindex meta tag in html), while `excluded_from_search` is for algolia, so in this case we want to use excluded_from_search.

DawoudSheraz (commented):
Changes look good from the code point of view. I will add 👍🏽 once CI is green.

PR #3807, opened 2023-02-20
feat: add delay in event bus event processing

DawoudSheraz:
### Description

This PR adds the capability to artificially delay a Kafka event bus event received from Studio if the event was created within last minute. This delay is needed to avoid concurrency issues that happen when:

- A course run is created on Publisher
- As part of atomic transaction that is used by Discovery to create a new run, Studio's course API is called to create course run in Studio
- Once the studio's course run is created, it emits Kakfa event
- Kafka event is consumed by Discovery workers before the original atomic transaction commit
- This results in creation of 2 versions of course and course runs with same keys but different UUIDs

More details are on https://github.com/edx/edx-arch-experiments/issues/159


course_discovery/apps/course_metadata/signals.py
```diff
@@ -345,6 +347,15 @@ def update_course_data_from_event(**kwargs):
         logger.error('Received null or incorrect data from COURSE_CATALOG_INFO_CHANGED.')
         return
 
+    event_metadata = kwargs.get('metadata')
+    if event_metadata:
+        logger.info(f"metadata located for COURSE_CATALOG_INFO_CHANGED for course run {course_data.course_key}")
```
rgraber: This is going to be pretty noisy. Maybe make it a `logger.debug`?

course_discovery/apps/course_metadata/signals.py
```diff
@@ -345,6 +347,15 @@ def update_course_data_from_event(**kwargs):
...
+        event_timestamp = event_metadata.time
+        time_diff = datetime.now(tz=timezone.utc) - event_timestamp
+        if time_diff.seconds < settings.EVENT_BUS_KAFKA_MESSAGE_DELAY_THRESHOLD_SECONDS:
+            logger.info("COURSE_CATALOG_INFO_CHANGED event received within the delay applicable window.")
```
rgraber: Same as above (re: info vs debug)

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -856,3 +857,23 @@ def test_error_cases(self):
...
             assert 'An error occurred while updating' in captured_logs.output[1]
+
+    @mock.patch('time.sleep')
+    def test_event_processing_delay(self, sleep_patch):
+        """
+        Verify that event processing is artificially delayed if the event is received within delay
+        applicable time window.
+        """
+        metadata = EventsMetadata(event_type='catalog-data-changed', minorversion=0)
+        with override_settings(EVENT_BUS_KAFKA_MESSAGE_DELAY_THRESHOLD_SECONDS=120):
```
rgraber, line 868: We should probably also override EVENT_BUS_KAFKA_PROCESSING_DELAY_SECONDS to make the test a little more robust against environment changes

rgraber (commented):
One thought about logging. We should also make sure we announce this to partner support so they can let partners know that the sync from studio to discovery will be often be delayed by a minute rather than instantaneous. Do we have a plan for announcing?

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: I can remove this line altogether.

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: ok

course_discovery/apps/course_metadata/tests/test_signals.py
DawoudSheraz: I don't see any benefit in overriding that setting variable, though I can refer that variable in assertions instead of hardcoded 60.

DawoudSheraz:
> One thought about logging. We should also make sure we announce this to partner support so they can let partners know that the sync from studio to discovery will be often be delayed by a minute rather than instantaneous. Do we have a plan for announcing?

We plan to test this change on stage first. Once done, this will be shared with relevant team.

rgraber (approved):
LGTM
I have a hunch we'll want to tweak the configured delay times but we can do that in edx-internal when the time comes

AliAdnanSohail (approved):

Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/signals.py
```diff
@@ -345,6 +347,15 @@ def update_course_data_from_event(**kwargs):
         logger.error('Received null or incorrect data from COURSE_CATALOG_INFO_CHANGED.')
         return
 
+    event_metadata = kwargs.get('metadata')
```
robrap, line 350: @DawoudSheraz: Since my understanding is that this was meant to be a temporary hack, it would be great if there were an ADR and/or comment that explains some of this. Either way, you could probably point to https://github.com/edx/edx-arch-experiments/issues/159 for more details. Thank you.

PR #3820, opened 2023-03-01
fix: remove unique constraint from GeoLocation in serializer

Ali-D-Akbar:
[PROD-3188](https://2u-internal.atlassian.net/browse/PROD-3188)
Removes unique constraints on Lat/Long in GeoLocation which fixes the publisher workflow and allows same lat/long pair to be assigned to another product.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -346,6 +346,9 @@ class GeoLocationSerializer(BaseModelSerializer):
     class Meta:
         model = GeoLocation
         fields = ('lat', 'lng', 'location_name')
+        # set validators to empty to override the Unique Constraint Validator
+        # to handle this in update_geolocation in api/serializer`
+        validators = []
 
 
 class PositionSerializer(BaseModelSerializer):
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1022,15 +1025,12 @@ class GeoLocation(TimeStampedModel):
         null=True
     )
 
-    class Meta:
-        unique_together = (('lat', 'lng'),)
-
```
DawoudSheraz: Rather than removing the constraint, it would be better if we can modify the validation behavior in API or serializer. This constraint is important to ensure there are no two locations with the same coordinates but different names. This can result in the possible creation of duplicates.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -2005,6 +2005,27 @@ def test_update_geolocation__valid_data(self, lat, lng, location_name, expected_
...
+        )
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
+        self.course.save()
+        assert self.course.draft_version is None
+        course_data = {'geolocation': {'lat': lat, 'lng': lng, 'location_name': location_name}, 'draft': False}
+        response = self.client.patch(url, course_data, format='json')
+        content = response.json()
+        self.course.refresh_from_db()
+        assert response.status_code == 200
+        assert content['geolocation'] == {'lat': '-47.97350000000', 'lng': '23.95000000000', 'location_name': 'Alps'}
```
DawoudSheraz, line 2029: Also assert on GeoLoc count to verify that no new objects are created.

course_discovery/apps/api/serializers.py
```diff
@@ -346,6 +346,7 @@ class GeoLocationSerializer(BaseModelSerializer):
     class Meta:
         model = GeoLocation
         fields = ('lat', 'lng', 'location_name')
+        validators = []
```
DawoudSheraz, line 351: TODO comment to explain why this is needed.

DawoudSheraz (approved):
A few nits before merge

AliAdnanSohail (approved):

PR #3822, opened 2023-03-01
feat: Exclude objects from search and SEO fields

marekwrobel:
Previous PR with some review feedback/discussion: https://github.com/openedx/course-discovery/pull/3803

Please read the ADR attached as part of this PR for more context. 

Two fields have been added to the Course and Program models:
- `excluded_from_search`
- `excluded_from_seo`

The first one takes part in the Algolia indexing logic. The second one will be passed to the frontend (Prospectus) during build, and pages that have this field set to TRUE will have a `noindex` meta tag added (Prospectus PR is still in-progress for this). 

More importantly, these fields can be edited through Django Admin for each `course-discovery` object, but also can be added in Contentful (for `Boot Camp Page` and `Degree Detail Page`). The Contentful value will override course-discovery one if its set to TRUE, but if its set to FALSE (or not defined), we will run our regular additional logic to determine whether we should index/hide this object.

If a field is set to False (or undefined), nothing really changes. But if it's set to true, it will be as easy way to force non-indexing (or hiding from SEO).


DawoudSheraz (approved):
please add the previous PR link to ensure the conversations on the PR are reachable to others.

AliAdnanSohail (approved):

PR #3833, opened 2023-03-10
feat: Add subscriptions eligibility models

aht007:
Models for Program Subscription Pricing.

Ticket: https://2u-internal.atlassian.net/browse/PON-4 (2u Internal)


christopappas (approved):
seems reasonable to me 👍

shafqatfarhan (approved):

aht007:
Table structures and data population examples:


**Program Table**
Screenshot 2023-03-13 at 5 28 20 PM


**Program Subscription Table**
Screenshot 2023-03-13 at 5 28 24 PM


**Program Subscription Pricing Table**
Screenshot 2023-03-13 at 5 28 28 PM

julianajlk (commented):
Can you modify the title to follow https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html#specification?

aht007:
> Can you modify the title to follow https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html#specification?

Done

course_discovery/apps/course_metadata/models.py
```diff
@@ -3092,6 +3092,40 @@ def trigger_program_skills_update_flow(self):
             UPDATE_PROGRAM_SKILLS.send(self.__class__, program_uuid=self.uuid)
 
 
+class ProgramSubscription(PkSearchableMixin, TimeStampedModel):
+    """Model for storing program subscription eligibility"""
+    id = models.UUIDField(primary_key=True, default=uuid4, editable=False, unique=True, verbose_name=_('ID'))
```
DawoudSheraz: Why do we need uuid to be the primary key as opposed to id field provided by Django? Also, it would be better to rename this field to `<>_uuid` instead of id to not cause conflicts with the existing id field provided by Django. It

course_discovery/apps/course_metadata/models.py
```diff
@@ -3092,6 +3092,40 @@ def trigger_program_skills_update_flow(self):
...
+    price = models.DecimalField(**PRICE_FIELD_CONFIG)
+    currency = models.ForeignKey(Currency, on_delete=models.CASCADE, default="USD")
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(fields=['program_subscription', 'currency'], name='unique_subscription_currency')
+        ]
+
+    def __str__(self):
+        return f"{self.price} {self.currency}"
```
DawoudSheraz: nit: should include program title or slug to differentiate easily.

DawoudSheraz (commented):
I have left a few comments.

course_discovery/apps/course_metadata/models.py
aht007: For most of the other models in course-discovery like Pathways, Organizations etc we were using uuid fields so to maintain consistency I used it as a key and also it wouldn't conflict with Django's own field because there would be no id field when we have declared our own field with unique and primary_key attributes.

aht007:
Inform: The CI failure looks like a flaky test, since its not connected to the changes in the PR

course_discovery/apps/course_metadata/models.py
DawoudSheraz: But what's the benefit of this field that the default id field does not provide? I dont see any benefit in creating a UUID field when it is not doing anything special.

aht007: Renamed the field to uuid as discussed on call.

DawoudSheraz (approved):

aht007:
@DawoudSheraz I have resolved the migration conflicts. Feel free to merge

course_discovery/apps/course_metadata/migrations/0319_auto_20230317_1424.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 3.2.18 on 2023-03-17 14:24
```
adzuci, line 1: I wonder if we could add a linter to encourage engineers to name migrations with human readable names?

PR #3841, opened 2023-03-15
fix: ignore empty fact objects in AdditionalMetadata

zawan-ila:
# [PROD-3215](https://2u-internal.atlassian.net/browse/PROD-3215)

## Description

This PR ignores empty fact objects in AdditionalMetadataSerializer i.e if an empty fact is passed in additionalMetadata from the frontend it will no longer be created in the database

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1381,6 +1381,9 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
         else:
             instance.facts.clear()
             for fact in facts_data:
+                # skip objects that do not have a truthy heading
+                if 'heading' not in fact or not fact['heading']:
+                    continue
                 # we can query all orphan facts and use one of them but that is more computation
                 # therefore, just checking if a fact with same data exists
                 facts_queryset = Fact.objects.filter(**fact)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1380,6 +1380,9 @@ def update_facts(self, instance, facts_data):
         else:
             instance.facts.clear()
             for fact in facts_data:
+                # skip empty fact objects coming from the frontend
+                if not fact:
```
DawoudSheraz: Each fact will be `{blurb: "", heading: ""}`. I don't think this condition would fix this. Also, please add a related unit test to capture this use case

zawan-ila: Facts reach the backend as {} The frontend appears to be sending them as {blurb: undefined, heading: undefined} Axios then appears to be dropping the keys with undefined values probably because that is standard JSON stringify behavior.

Regardless I have updated this to check for a non-empty heading as I think it would be more robust. Tests have also been added

DawoudSheraz (approved):

Ali-D-Akbar (approved):

AfaqShuaib09 (approved):

PR #3842, opened 2023-03-16
feat: update populate_default_product_source to populate program's value as well

AliAdnanSohail:
[Populate legacy/OCM Degrees/Programs edX product_source via management command.](https://2u-internal.atlassian.net/browse/PROD-3210)

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
+++ b/course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
@@ -1,5 +1,5 @@
 """
-Management command to populate the default identifier to the courses having no identifier
+Management command to populate the default identifier to the products having no identifier
 """
 import logging
 
@@ -7,30 +7,59 @@ from django.conf import settings
 from django.core.management import BaseCommand, CommandError
 from django.db.transaction import atomic
 
-from course_discovery.apps.course_metadata.models import Course, Source
+from course_discovery.apps.course_metadata.models import Course, Program, Source
 
 logger = logging.getLogger(__name__)
 
 
+def get_products_with_no_product_source(product_type):
+    """
+    If product type is course then filtering the courses with no product_source and no additional_metadata because
+    legacy edX courses do not have additional_metadata, only the courses from external sources have additional_metadata
+    Else filtering the programs with no product_source and are not external degree programs
+    """
+    if product_type == 'course':
+        return Course.everything.filter(product_source=None, additional_metadata=None)
+    products = [
+        program.id for program in Program.objects.filter(product_source=None)
+        if not program.is_2u_degree_program
+    ]
+    return Program.objects.filter(id__in=products)
+
+
 class Command(BaseCommand):
-    help = 'Populate the default product_source to the courses having no product_source'
+    help = 'Populate the default product_source to the products having no product_source'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--product_type',
+            help='The product type we want to populate with default product source.'
+                 'It can only be either course or program',
+            default='course',
+            required=False,
+            type=str,
+        )
 
     def handle(self, *args, **kwargs):
+        product_type = kwargs.get('product_type', 'course').lower()
         try:
+            if product_type not in ['course', 'program']:
+                raise Exception(f"Product Type {product_type} is invalid")  # pylint: disable=broad-exception-raised
             default_product_source = Source.objects.get(slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG)
             with atomic():
-                # filtering the courses with no product_source and no additional_metadata because legacy edX courses
-                # do not have additional_metadata, only the courses from external sources have additional_metadata
-                courses = Course.everything.filter(product_source=None, additional_metadata=None)
-                # tranforming the queryset to list because after updating the courses, the queryset becomes empty
-                courses_list = list(courses)
-                courses.update(product_source=default_product_source)
-                logger.info(f'Updated {len(courses_list)} courses with default product_source')
-                if len(courses_list) > 0:
+                products = get_products_with_no_product_source(product_type)
+                # tranforming the queryset to list because after updating the products, the queryset becomes empty
+                product_list = list(products)
+                products.update(product_source=default_product_source)
+                logger.info(f'Updated {len(product_list)} {product_type}s with default product_source')
+                if len(product_list) > 0:
                     # pylint: disable=logging-not-lazy
-                    logger.info('Updated courses:\n' +
-                                '\n'.join(f"{course.key} - {'draft' if course.draft else 'non-draft'}"
-                                          for course in courses_list))
+                    if product_type == 'course':
+                        logger.info('Updated courses:\n' +
+                                    '\n'.join(f"{course.key} - {'draft' if course.draft else 'non-draft'}"
+                                              for course in product_list))
+                    else:
+                        logger.info('Updated programs:\n' + '\n'.join(program.title for program in product_list))
 
         except Source.DoesNotExist as ex:
             logging.exception(f'Default product_source {settings.DEFAULT_PRODUCT_SOURCE_SLUG} does not exist')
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
```diff
@@ -1,36 +1,60 @@
...
 
 class Command(BaseCommand):
-    help = 'Populate the default product_source to the courses having no product_source'
+    help = 'Populate the default product_source to the products having no product_source'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--product_type',
+            help='The product type we want to populate with default product source.'
+                 'It can only either course or program',
```
Ali-D-Akbar: incomplete sentence

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
```diff
@@ -1,36 +1,60 @@
...
                     # pylint: disable=logging-not-lazy
-                    logger.info('Updated courses:\n' +
-                                '\n'.join(f"{course.key} - {'draft' if course.draft else 'non-draft'}"
-                                          for course in courses_list))
+                    if product_type == 'course':
+                        logger.info('Updated courses:\n' +
+                                    '\n'.join(f"{course.key} - {'draft' if course.draft else 'non-draft'}"
+                                              for course in product_list))
+                    else:
+                        logger.info('Updated programs:\n' + '\n'.join(program.title for program in product_list))
```
Ali-D-Akbar, line 62: Why not just do the logging in the above if condition? I think this can added in the above condition.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_default_product_source.py
```diff
@@ -50,7 +50,24 @@ def test_populate_default_product_source_with_no_existing_product_source(self):
                 )
             ])
 
-    def test_populate_default_product_source_with_existing_product_source(self):
+    def test_populate_default_product_source__programs_with_no_existing_product_source(self):
```
Ali-D-Akbar, line 56: This test can be reduced by using ddt or by making a function as it replicates the above test.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_default_product_source.py
```diff
@@ -65,6 +82,27 @@ def test_populate_default_product_source_with_existing_product_source(self):
                 assert Course.everything.filter(product_source=self.default_product_source).count() == 0
                 mock_logger.info.assert_called_once_with('Updated 0 courses with default product_source')
 
+    def test_populate_default_product_source__programs_with_existing_product_source(self):
```
Ali-D-Akbar, line 88: same concern as the previous test.

Ali-D-Akbar (commented):
A bunch of comments to address

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
DawoudSheraz: How do we plan to toggle these params? We would need to change the command in Jenkins to re-create and re-seed the job.

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
```diff
@@ -1,36 +1,60 @@
...
-                courses_list = list(courses)
-                courses.update(product_source=default_product_source)
-                logger.info(f'Updated {len(courses_list)} courses with default product_source')
-                if len(courses_list) > 0:
+                if product_type == 'course':
+                    # filtering the courses with no product_source and no additional_metadata because legacy edX courses
+                    # do not have additional_metadata, only the courses from external sources have additional_metadata
+                    products = Course.everything.filter(product_source=None, additional_metadata=None)
+                else:
+                    # filtering the programs with no product_source and are not 2u degree programs
```
DawoudSheraz: Like course, we should use additional metadata on degree to check if a program is external program, instead of hardcoding 2u related strings.

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
DawoudSheraz: Please separate out course and program handling into different util methods. The current code is hard to read.

DawoudSheraz (commented):
- Left a few comments

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
AliAdnanSohail: program don't have `key` or `draft` attribute which is using to log course information

course_discovery/apps/course_metadata/management/commands/tests/test_populate_default_product_source.py
AliAdnanSohail: using ddt was making test case complex, like we have to change logging string, factory object, model etc. Due to which separating these tests were more readable and simple

course_discovery/apps/course_metadata/management/commands/tests/test_populate_default_product_source.py
AliAdnanSohail: same as above

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
AliAdnanSohail: We have to create 2 jobs in Jenkins, it will not effect the previous one because default product type is course. For programs we have to create a new

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
AliAdnanSohail: Actually additional metadata is for Degree model only, not for Program model. Degree is child class of Program, so if I search for additional_metadata of any Program it will not throw error. So we already have a method which tells this program is 2U program or not that's why I'm using that method here.

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
AliAdnanSohail: I think rather than creating utils method, I can separate the functions here because I don't think we will use this util method any where else

DawoudSheraz: Yeah, I meant that as well. Bad choice of word.

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
DawoudSheraz: Ok, maybe we can rename that method then.

AliAdnanSohail: Renaming this method requires some changes and testing which can be done in a separate task.

DawoudSheraz: We can use PyCharm's refactoring to do this.

AliAdnanSohail: I just checked, the same thing is being used in serializers as well. So I think we should do this is a separate follow up task, because after this change we need to do proper testing

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
```diff
@@ -1,36 +1,63 @@
...
-from course_discovery.apps.course_metadata.models import Course, Source
+from course_discovery.apps.course_metadata.models import Course, Program, Source
 
 logger = logging.getLogger(__name__)
 
 
+def get_products_with_no_product_source(product_type):
+    if product_type == 'course':
+        # filtering the courses with no product_source and no additional_metadata because legacy edX courses
+        # do not have additional_metadata, only the courses from external sources have additional_metadata
```
DawoudSheraz: Let's move these comments to function docstring

DawoudSheraz (approved):

course_discovery/apps/course_metadata/management/commands/populate_default_product_source.py
Ali-D-Akbar: nit: only be either course or program*

Ali-D-Akbar (approved):

PR #3847, opened 2023-03-20
fix: allow blank on excluded from search and seo fields

marekwrobel:
[This recent PR ](https://github.com/openedx/course-discovery/pull/3822) has added two fields on Course and Program models: `excluded_from_search` and `excluded_from_seo`.

It was reported that in Publisher there is an error when you try to create a new Course. The error is:
`Failed to set data. Field 'excluded_from_search' does not have a default value. `

After a small investigation, it looks like we need to add `blank=True` to the model (and switch `Null=True` as well), which will cause `required=False` for the form field in Publisher. 

More can be read [here](https://stackoverflow.com/questions/17199046/blank-null-and-required-in-django-models-and-modelforms), as well as [here](https://stackoverflow.com/questions/71352387/django-form-validation-error-not-showing-on-boolean-field) and [here](https://stackoverflow.com/questions/16955256/booleanfield-default-value-not-set-when-creating-a-model-instance)

Screenshots from local Publisher (having created a new course) and django admin showing the default state for both fields right after creation.

Screenshot 2023-03-21 at 10 43 20 AM

Screenshot 2023-03-21 at 10 49 25 AM


DawoudSheraz (commented):
Was this fix verified on local publisher?

marekwrobel:
@DawoudSheraz Yes, I created a new course in local publisher successfully and verified the fields were set to `False` by default.

DawoudSheraz (approved):

zawan-ila (approved):

PR #3848, opened 2023-03-21
feat: serialize subscription fields

aht007:
[Jira Ticket](https://2u-internal.atlassian.net/browse/PON-3)(2u-Internal)

## Description

The new program subscription model “program_subscription“ which were added in this [PR](https://github.com/openedx/course-discovery/pull/3833) should be serialized into the Program serializer(s) and its fields should be synced in the Algolia connector for Algolia search indexing and Contentful data.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -41,8 +41,8 @@ from course_discovery.apps.course_metadata.models import (
     CourseType, Curriculum, CurriculumCourseMembership, CurriculumProgramMembership, Degree, DegreeAdditionalMetadata,
     DegreeCost, DegreeDeadline, Endorsement, Fact, GeoLocation, IconTextPairing, Image, LevelType, Mode, Organization,
     Pathway, Person, PersonAreaOfExpertise, PersonSocialNetwork, Position, Prerequisite, ProductMeta, ProductValue,
-    Program, ProgramLocationRestriction, ProgramType, Ranking, Seat, SeatType, Source, Specialization, Subject,
-    TaxiForm, Topic, Track, Video
+    Program, ProgramLocationRestriction, ProgramSubscription, ProgramSubscriptionPrice, ProgramType, Ranking, Seat,
+    SeatType, Source, Specialization, Subject, TaxiForm, Topic, Track, Video
 )
 from course_discovery.apps.course_metadata.utils import get_course_run_estimated_hours, parse_course_key_fragment
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
@@ -1869,6 +1869,38 @@ class DegreeSerializer(BaseModelSerializer):
         return list(degree.specializations.values_list('value', flat=True))
 
 
+class ProgramSubscriptionPriceSerializer(BaseModelSerializer):
+    """
+    Serializer for the ``ProgramSubscriptionPrice`` model.
+    """
+    currency = serializers.SlugRelatedField(read_only=True, slug_field='code')
+
+    class Meta:
+        model = ProgramSubscriptionPrice
+        fields = ['price', 'currency']
+
+    @classmethod
+    def prefetch_queryset(cls, queryset):
+        return queryset.select_related('currency')
+
+
+class ProgramSubscriptionSerializer(BaseModelSerializer):
+    """
+    Serializer for the ``ProgramSubscription`` model.
+    """
+    prices = ProgramSubscriptionPriceSerializer(many=True)
+
+    class Meta:
+        model = ProgramSubscription
+        fields = ['subscription_eligible', 'prices']
+
+    @classmethod
+    def prefetch_queryset(cls, queryset):
+        return queryset.prefetch_related(
+            'prices',
+        )
+
+
 class MinimalProgramSerializer(TaggitSerializer, FlexFieldsSerializerMixin, BaseModelSerializer):
     """
     Basic program serializer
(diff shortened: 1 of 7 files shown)


aht007:
This is how the data will appear on our programs api (api/v1/programs/):

For a program with subscription eligible True
Screenshot 2023-03-24 at 11 21 03 AM

For a program with subscription eligible False or N/A

Screenshot 2023-03-24 at 11 21 07 AM

Also on Alogolia we will have the below new fields as a result of these changes:
Screenshot 2023-03-24 at 9 56 04 PM

And these facets will also be added:

Screenshot 2023-03-24 at 9 55 53 PM

course_discovery/apps/api/serializers.py
```diff
@@ -2025,6 +2060,27 @@ def get_card_image_url(self, obj):
...
+            return data
+
+        subscription_eligible = subscription.get('subscription_eligible', False)
+        subscription_prices = subscription.get('prices', [])
+
+        # Add subscription properties to data object
+        data['subscription_eligible'] = subscription_eligible
+        data['subscription_prices'] = subscription_prices
+
+        return data
```
DawoudSheraz: Why is this needed? Can't this be kept under `subscription_eligible`?

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -258,7 +258,7 @@ def test_filter_by_type(self):
         program_type_name = 'foo'
         program = ProgramFactory(type__name_t=program_type_name, partner=self.partner)
         url = self.list_path + '?type=' + program_type_name
-        self.assert_list_results(url, [program], 18)
+        self.assert_list_results(url, [program], 21)
```
DawoudSheraz, line 261: 3 counts? Is it 3 because of new serializer? Can you double check if that leads to N+1 problem?

course_discovery/apps/course_metadata/index.py
```diff
@@ -79,7 +79,8 @@ class EnglishProductIndex(BaseProductIndex):
     facet_fields = (('availability_level', 'availability'), ('subject_names', 'subject'), ('levels', 'level'),
                     ('active_languages', 'language'), ('product_type', 'product'), ('program_types', 'program_type'),
                     ('staff_slugs', 'staff'), ('product_allowed_in', 'allowed_in'),
-                    ('product_blocked_in', 'blocked_in'))
+                    ('product_blocked_in', 'blocked_in'), 'subscription_eligible',
+                    'subscription_prices')
```
DawoudSheraz, line 83: Is there a need to facet on prices? Eligibility makes sense. But price is a list of dict, not sure what benefits will be there.

DawoudSheraz (commented):
I have left a few comments on different parts of this PR.

course_discovery/apps/api/serializers.py
aht007: The original structure of the data would have a subscription dictionary which contained the field nested in it. We just removed the subscription dictionary here and instead placed the fields at the top level so that parsing the data is easy.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
aht007: I don't think it is leading to N+1 problems. In our program serializer we are prefetching the program subscription which prefetches all the subscription prices which in turn prefetches the currencies so an overall increase of 3 in the count.

course_discovery/apps/course_metadata/index.py
aht007: Prices is a list of dict but I wanted to add it so if someone wants to segregate records based on the subscription currency then they should be able to. Price field is not of much interest here for us.

course_discovery/apps/api/serializers.py
```diff
@@ -2025,6 +2060,27 @@ def get_card_image_url(self, obj):
...
+
+        subscription_eligible = subscription.get('subscription_eligible', None)
+        subscription_prices = subscription.get('prices', None)
+
+        # Add subscription properties to data object
+        data['subscription_eligible'] = subscription_eligible
+        data['subscription_prices'] = subscription_prices
+
+        return data
+
```
uzairr, line 2070: i think this code could be refactored:
`
data['subscription_prices'] = None if not subscription else subscription.get('subscription_prices', None)
`
the above same pattern can be used to extract the eligibility. After that the function size will be reduced to just 3,4 lines

course_discovery/apps/api/serializers.py
DawoudSheraz: Ok.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
DawoudSheraz: Can you list down which 3 queries are executing? Cause if we are adding 3 queries for each program, it can lead to  N+1.

course_discovery/apps/course_metadata/index.py
DawoudSheraz: Ah, Ok. I personally don't see any value in adding this field as facet field. Maybe I am missing some context here.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
aht007: These are the queries that are getting added, plus there is one for fetching the currencies too which is also being prefetched. So it wouldn't do three queries for every single program but 3 queries overall for all programs.

Screenshot 2023-03-29 at 12 08 49 PM

DawoudSheraz (approved):

uzairr (approved):

PR #3852, opened 2023-03-24
build: update docker publish workflow

UsamaSadiq:
### Description
- Copied the workflow fix from the PR https://github.com/openedx/edx-analytics-dashboard/pull/1408. 
- The existing workflow failed to trigger build on every open-release branch commit so the workflow is updated to create a docker file on every commit and publish the image. 

### Follow Up Changes
- Once this PR is merged, a follow up back-port PR will be created against `open-release/olive.master` branch as well to resolve the blocker on the issue https://github.com/openedx/devstack/issues/1000


iamsobanjaved (approved):

.github/workflows/docker-publish.yml
```diff
@@ -4,32 +4,31 @@ on:
...
+      # Examples: if we're building 'open-release/olive.master', tag the image as 'olive.master'.
       # Otherwise, we must be building from a push to master, so use 'latest'.
       - name: Get tag name
         id: get-tag-name
         uses: actions/github-script@v5
         with:
           script: |
-            const releasePrefix = 'refs/tags/open-release/';
-            const tagName = context.ref.split(releasePrefix)[1] || 'latest';
+            const branchName = context.ref.split('/').slice(-1)[0];
```
DawoudSheraz, line 24: What does this do? Like what is present in context ref?

iamsobanjaved: It is getting the name of the branch from which this workflow is triggered. `context.ref` contains the value of the default env variable of github actions `GITHUB_REF`, which contains the branch name.

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

DawoudSheraz: What happens if the branch name contains `/`, like this PR? Will it only contain the part after `/`?

iamsobanjaved: It is configured for master and open release branches which don't have this pattern.

UsamaSadiq: We only want to trigger this workflow on the merge commits in the master or open-release branches so we don't need to handle such cases.

DawoudSheraz (approved):

PR #3867, opened 2023-03-31
feat: use pre_save signal to determine data modified timestamp from related models

DawoudSheraz:
### Description

1. Fix an issue with FieldTracker based updates where even saving a Course with no changes would result in timestamp update for Non-draft version. This happened because of how draft is converted to non-draft in Discovery using [set_official_state](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/utils.py#L61) . The pk of non-draft is assigned to draft object to make it non-draft, with the original draft obj assigned to non-draft. This resulted in the field tracker picking non-draft course/course run as updated, resulting in timestamp update for non-draft version. This is also the reason that whenever RCM LMS data loader was run, all the courses had updated timestamps even if course run was not updated.
2. Add pre-save signal for following models to update data modified timestamp for Course if the related/reverse object. is changed but the course object is not changed. This also fixes the cases where CourseRun only updates on publisher or RCM were not updating timestamps
  - AdditionalMetadata
  - CertificateInfo
  - CourseRun
  - CourseLocationRestriction
  - CourseEntitlement
   - GeoLocation
   - ProductMeta
   - ProductValue
3. add m2m_changed signal for CourseRun's transcript_languages to update timestamp if transcript languages are updated for a Course Run (https://docs.djangoproject.com/en/4.2/ref/signals/#m2m-changed)
4. For pre_save signal specifically on CourseRun, the method in CourseRun uses `Course.everything` to update both timestamp for draft and non-draft versions of the Course. Ideally, based on draft status of CourseRun, the related Course version (draft/non-draft) should have been updated. I Implemented it that way but the timestamp was not updating for non-draft courses in RCM flow.  This is because RCM does not use [set_official_state](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/utils.py#L61) to update non-draft version of CourseRun. Instead,  the data dict is manually updated for both draft and non-draft versions separately. 
### New Identified Issues / Todos

During the implementation of this work, a few more issues were identified:
- For M2M, there are 2 issues.
     - m2m_changed signal was being fired even if the object was not changing. This could be due to the behavior of the SortedManyToMany field. This results in data-modified timestamp updating even if there is no change in M2M.
    - Taggable manager’s m2m_changed signal is being sent by models other than Course. This is new and would require more digging.
- The handling of AdditionalMetadata and its related objects needs an update in course API. The current approach to updating AdditionalMeetadata uses .filter().update() which does not trigger pre_save signal.
- Create internal tickets test_signals update and to update draft & non-draft handling logic in RCM LMS data loader

#### RCM

- Go to Studio and update the dates from Schedule & Details page (http://localhost:18010/settings/details/course-v1:testOrg+APSPM+1T2023)
- Go to Discovery shell and run refresh_course_metadata management command (`./manage.py refresh_course_metadata`)
- Make sure to notice CourseRun has updated in the logs
- verify the timestamp has been updated for Course in Discovery. Redo this by changing dates again from Studio and run RCM. Also, verify the course is listed when filtering via timestamp on Course API (http://localhost:18381/api/v1/courses/?timestamp=2023-04-06)
- Now, run RCM again without changing anything. Ensure no timestamp update happens

#### Publisher
- Open any course on publisher. Edit the data on course or course run form. Apart from the following fields, changing any data will result in updating timestamp:
   - Topics/Tags (Taggable manager)
   - Collaborators (Sorted M2M)
   - Subjects (Sorted M2M)
   - AdditionalMetadata (only for external courses, due to how it is handled in course API)
   - Staff (CourseRun, Sorted M2M)

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -39,6 +39,9 @@ from course_discovery.apps.course_metadata.models import (
 from course_discovery.apps.course_metadata.publishers import (
     CourseRunMarketingSitePublisher, ProgramMarketingSitePublisher
 )
+from course_discovery.apps.course_metadata.signals import (
+    connect_course_data_modified_timestamp_related_models, disconnect_course_data_modified_timestamp_related_models
+)
 from course_discovery.apps.course_metadata.tests import factories
 from course_discovery.apps.course_metadata.tests.factories import (
     AdditionalMetadataFactory, CourseFactory, CourseRunFactory, ImageFactory, ProgramFactory, SeatFactory,
@@ -55,6 +58,17 @@ from course_discovery.apps.publisher.tests.factories import OrganizationExtensio
 @pytest.mark.usefixtures('elasticsearch_dsl_default_connection')
 @ddt.ddt
 class TestCourse(TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        disconnect_course_data_modified_timestamp_related_models()
+
+    @classmethod
+    def tearDownClass(cls):
+        connect_course_data_modified_timestamp_related_models()
+        super().tearDownClass()
+
     def test_str(self):
         course = factories.CourseFactory()
         assert str(course), f'{course.key}: {course.title}'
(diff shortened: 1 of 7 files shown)


Ali-D-Akbar (approved):
Really good PR. self-explains a lot of stuff.

AfaqShuaib09 (approved):
I've verified changes on local, LGTM

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -80,21 +94,32 @@ def test_data_modified_timestamp_model_field_change(self):
...
-        course = factories.CourseFactory()
+        course = factories.CourseFactory(draft=True)
         data_modified_timestamp = course.data_modified_timestamp
         course.short_description = 'Testing change'
         course.save()
         assert data_modified_timestamp < course.data_modified_timestamp
 
+    def test_data_modified_timestamp_model_non_draft(self):
+        """
+        Verify data modified timestamp does not change for non-draft course change.
```
zawan-ila, line 105: Why do we want no changes for non-drafts?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1021,6 +1056,22 @@ def test_old_mongo_not_marketable(self, key, expected):
         assert course_run.is_marketable == expected
         assert course_run.could_be_marketable == expected
 
+    def test_update_product_data_modified_timestamp(self):
```
zawan-ila, line 1059: In general the approach in these tests appears to be directly calling update_data_timestamp. Would it not be better to enable the signals and call save and then test the behavior?

zawan-ila (approved):

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: This is mentioned in PR description, the first point. Due to how draft is converted into non-draft, FieldTracker considers this a valid change, and timestamp is updated for non-draft object (https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/utils.py#L61).  That is why timestamp is updated for draft version only. Whenever set_official_state is called, it will just copy the modified timestamp from draft to non-draft.

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: That can be debatable.  The methods have been added in models.py, so it is better to test them in isolation in test_models.py. Although I can look into adding some behavioral tests in test_signals.py in a followup PR.

PR #3877, opened 2023-04-06
feat: handle logic to ofac restrict programs

AliAdnanSohail:
[Create a model that handles select program types based on select product Source to be automatically moved in Legal Review upon creation via CSV Ingestion or Django](https://2u-internal.atlassian.net/browse/PROD-3022)

## Description
We need to OFAC restrict specific program types of specific product source. So this PR will add logic to check if current program/degree object's program type exist in restricted program types, then it will be marked as OFAC restricted

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -364,10 +364,10 @@ class ProgramAdmin(DjangoObjectActions, admin.ModelAdmin):
 
     # ordering the field display on admin page.
     fields = (
-        'uuid', 'title', 'subtitle', 'marketing_hook', 'status', 'type', 'partner', 'banner_image', 'banner_image_url',
-        'card_image', 'marketing_slug', 'overview', 'credit_redemption_overview', 'video', 'total_hours_of_effort',
-        'weeks_to_complete', 'min_hours_effort_per_week', 'max_hours_effort_per_week', 'courses',
-        'order_courses_by_start_date', 'custom_course_runs_display', 'excluded_course_runs', 'product_source',
+        'uuid', 'title', 'subtitle', 'marketing_hook', 'product_source', 'type', 'status', 'partner', 'banner_image',
+        'banner_image_url', 'card_image', 'marketing_slug', 'overview', 'credit_redemption_overview', 'video',
+        'total_hours_of_effort', 'weeks_to_complete', 'min_hours_effort_per_week', 'max_hours_effort_per_week',
+        'courses', 'order_courses_by_start_date', 'custom_course_runs_display', 'excluded_course_runs',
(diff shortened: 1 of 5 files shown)


course_discovery/apps/course_metadata/admin.py
```diff
@@ -447,6 +447,9 @@ def response_change(self, request, obj):
 
     def save_model(self, request, obj, form, change):
         try:
+            if obj.type in obj.product_source.ofac_restricted_program_types.all():
+                obj.has_ofac_restrictions = True
+                obj.ofac_comment = f"Program type {obj.type.slug} is OFAC restricted for {obj.product_source.name}"
```
DawoudSheraz: nit: I would suggest moving this message to some constant/settings.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -447,6 +447,9 @@ def response_change(self, request, obj):
 
     def save_model(self, request, obj, form, change):
         try:
+            if obj.type in obj.product_source.ofac_restricted_program_types.all():
```
DawoudSheraz: perf/nit: We are fetching all types from DB. Instead, we can check the existence of obj.type in source's restricted program types.

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -283,6 +283,10 @@ def _update_or_create_degree(
             defaults=data_dict
         )
 
+        if degree.product_source and program_type in degree.product_source.ofac_restricted_program_types.all():
+            degree.has_ofac_restrictions = True
+            degree.save()
+
```
DawoudSheraz: Move ofac handling logic to models.py and use that method both here and in admin.

zawan-ila (approved):

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -283,6 +283,10 @@ def _update_or_create_degree(
             defaults=data_dict
         )
 
+        if degree.product_source and program_type in degree.product_source.ofac_restricted_program_types.all():
+            degree.has_ofac_restrictions = True
+            degree.save()
```
Ali-D-Akbar: Are we not adding comments in case of degree?

Ali-D-Akbar (commented):
A PR description would be nice to have which explains the approach and changes done in the PR. Approval once the comments are addressed.

course_discovery/apps/course_metadata/admin.py
AliAdnanSohail: As we are using this text just here only, so I don't think we need to save it into constant

Ali-D-Akbar (approved):

DawoudSheraz (approved):

PR #3879, opened 2023-04-07
feat: indexing external_url to algolia

srwang:
Ensuring that prospectus search results can reference the external_url. For Emeritus products, because we will not be building pages for them, we will always use the external url and direct the user off site.

[WS-3714](https://2u-internal.atlassian.net/browse/WS-3714)

### DIFF
--- a/course_discovery/apps/course_metadata/algolia_models.py
+++ b/course_discovery/apps/course_metadata/algolia_models.py
@@ -150,10 +150,6 @@ class AlgoliaBasicModelFieldsMixin(models.Model):
         abstract = True
 
     @property
-    def product_external_url(self):
-        return self.additional_metadata.external_url if self.additional_metadata else None
-
-    @property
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -208,6 +212,10 @@ class AlgoliaProxyCourse(Course, AlgoliaBasicModelFieldsMixin):
 
     class Meta:
         proxy = True
+    
+    @property
+    def product_external_url(self):
+        return self.additional_metadata.external_url if self.additional_metadata else None
```
DawoudSheraz: Why define it again for the course if it is already present on ProxyProduct?

DawoudSheraz (commented):
Please add an appropriate unit test

DawoudSheraz (approved):

PR #3883, opened 2023-04-10
feat: modify data loader to support degree variants

uzairr:
In order to make compliant the existing data loader with external degrees ingestion, changes have been made to ingest the degrees based on its product source.

[PROD-3253](https://2u-internal.atlassian.net/browse/PROD-3253)


course_discovery/settings/base.py
```diff
@@ -713,3 +713,16 @@
...
+        'title', 'card_image_url', 'product_type', 'organization_key', 'organization_short_code_override',
+        'slug', 'primary_subject', 'content_language', 'course_level', 'paid_landing_page_url', 'organic_url',
+        'identifier', 'overview',
+    ],
+    'degree_required_data_fields': [
+        'title', 'card_image_url', 'product_type', 'organization_key', 'organization_short_code_override',
+        'slug', 'primary_subject', 'content_language', 'course_level', 'paid_landing_page_url', 'organic_url',
+        'identifier', 'overview', 'courses',
+    ],
+}
```
AfaqShuaib09: I think that instead of declaring this setting variable here, it would be more appropriate to include it in the `edx-internal` repo. Additionally, we could define a set of base mapping fields that are common between `degree_required_data_fields` and `shorelight` degree fields, because all of the fields are identical in both except for the `courses` field,  which we can add to the mapping fields based on the `product_source` criteria.

uzairr: yeah it could be added on `yaml` but in case we still need to add these mappings in the settings for our local environment. 
In the light of above comment, i am refactoring field map list in this PR 👍

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -190,8 +184,9 @@ def validate_degree_data(self, data):
         data dictionary and return a comma separated string of missing data fields.
         """
         missing_fields = []
-
-        for field in self.DEGREE_REQUIRED_DATA_FIELDS:
+        degree_variants = settings.DEGREE_VARIANTS_FIELD_MAP
```
DawoudSheraz: Use .copy() to create a copy, otherwise it will cause tests to fail (base list updated via ref)

course_discovery/apps/course_metadata/data_loaders/tests/mixins.py
```diff
@@ -91,7 +91,7 @@ def setUp(self):
         super().setUp()
         self.program_type = ProgramType.objects.get(slug=ProgramType.MASTERS)
         self.marketing_text = "<ul><li>ABC</li><li>D&E</li><li>Harvard CS50</li></ul>"
-        self.product_source = SourceFactory(slug='text-source')
+        self.product_source = SourceFactory(slug='2u')
```
DawoudSheraz: please use generic source here.

course_discovery/settings/base.py
```diff
@@ -713,3 +713,13 @@
 EVENT_BUS_KAFKA_MESSAGE_DELAY_THRESHOLD_SECONDS = 60
 
 ALGOLIA_INDEX_EXCLUDED_SOURCES = []
+
+DEGREE_REQUIRED_DATA_FIELDS = [
+    'title', 'card_image_url', 'product_type', 'organization_key', 'organization_short_code_override',
+    'slug', 'primary_subject', 'content_language', 'course_level', 'paid_landing_page_url', 'organic_url',
+    'identifier', 'overview',
+]
```
DawoudSheraz: Keep this list in Degree loader but name it as bases required fields.

course_discovery/settings/base.py
```diff
@@ -713,3 +713,13 @@
...
+
+DEGREE_VARIANTS_FIELD_MAP = {
+    '2u': [ 'courses', ],
+}
```
DawoudSheraz: please avoid 2u specific details here. This should be moved to edx-internal.

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -190,8 +191,9 @@ def validate_degree_data(self, data):
         data dictionary and return a comma separated string of missing data fields.
         """
         missing_fields = []
-
-        for field in self.DEGREE_REQUIRED_DATA_FIELDS:
+        degree_variants = copy.copy(settings.DEGREE_VARIANTS_FIELD_MAP)
```
DawoudSheraz: nit: you can use settings.DEGREE_VARIANTS_FIELD_MAP.copy()

course_discovery/apps/course_metadata/data_loaders/tests/test_degree_csv_loader.py
```diff
@@ -54,7 +54,7 @@ def _assert_default_logs(self, log_capture):
 
         )
 
-    @ddt.data('identifier', 'card_image_url', 'title', 'paid_landing_page_url', 'organic_url', 'courses')
+    @ddt.data('identifier', 'card_image_url', 'title', 'paid_landing_page_url', 'organic_url')
```
DawoudSheraz: I would suggest adding source based checks here to test the behavior.

course_discovery/apps/course_metadata/data_loaders/degrees_loader.py
```diff
@@ -137,11 +137,15 @@ def ingest(self):
...
-            degree = Degree.objects.filter(
-                marketing_slug=degree_slug, partner=self.partner,
-                additional_metadata__external_identifier=row['identifier']
-            ).first()
+            degree_params = {
+                'marketing_slug': degree_slug,
+                'partner': self.partner,
+            }
+            identifier = row.get('identifier', None)
+            if identifier:
```
DawoudSheraz: Identifier is used in the ingestion email. The absence of an identifier will cause the email to show nothing for newly created Degrees. Please refer to email template for this.

uzairr: In that case, we should rethink around the email template to make it customized based on product source. Identifier filed has no utility in shorelight context and even it is not been provided in the ingestion data. 🤔

DawoudSheraz: The email template is generic enough. External identifier is important for external products and provides a way to link entry within Discovery with external products.

uzairr: hmm, then we should think a work around for the external products that are not provided by `identifier` value.

course_discovery/apps/course_metadata/data_loaders/tests/test_degree_csv_loader.py
```diff
@@ -54,7 +55,8 @@ def _assert_default_logs(self, log_capture):
 
         )
 
-    @ddt.data('identifier', 'card_image_url', 'title', 'paid_landing_page_url', 'organic_url', 'courses')
+    @override_settings(DEGREE_VARIANTS_FIELD_MAP={'text-source': ['courses']})
```
Ali-D-Akbar, line 58: nit: test-source

Ali-D-Akbar (commented):
One comment and handling identifier as discussed before final approval.

Ali-D-Akbar (approved):

DawoudSheraz (approved):

course_discovery/apps/course_metadata/data_loaders/tests/test_degree_csv_loader.py
uzairr: addressed this small nit and identifier support is implemented as per discussed in Today's standup. @Ali-D-Akbar

PR #3885, opened 2023-04-10
fix: fix the reviewable_data_has_changed to properly handle dictionaries

zawan-ila:
[PROD-3202](https://2u-internal.atlassian.net/browse/PROD-3202)

## Description

In this PR, we fix the reviewable_data_has_changed function to properly handle dictionaries.

## Context
As things stand now, every request to update a course is considered as having changed the course data regardless of whether any data has actually changed or not. This is because the geolocation and location_restriction fields (which are dicts) are not properly handled by reviewable_data_has_changed. This causes the issue mentioned in the ticket above. That is, it is not possible to have multiple course runs in scheduled state. As soon as you finish the review of a course run(to move it into scheduled state) the reviewable_data_has_changed decides that course data has changed and removes all runs from scheduled state to unsubmitted. This PR attempts to fix this issue.

## Testing

1. Create a course on publisher
2. Complete its review process
3. Make multiple(>1) runs of this course with publish dates in future
4. Complete their review process.
5. Ensure that moving one run into scheduled state does not change status of another run
6. Ensure that at the end all runs are in scheduled state.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -325,7 +325,7 @@ class VideoSerializer(MediaSerializer):
 
 class GeoLocationSerializer(BaseModelSerializer):
     """Serializer for the ``GeoLocation`` model."""
-    location_name = serializers.CharField(allow_null=True, max_length=128)
+    location_name = serializers.CharField(allow_blank=True, allow_null=True, max_length=128)
     lat = serializers.DecimalField(
         max_digits=GeoLocation.LAT_MAX_DIGITS,
         decimal_places=GeoLocation.DECIMAL_PLACES,
@@ -686,7 +686,7 @@ class AdditionalMetadataSerializer(BaseModelSerializer):
     """Serializer for the ``AdditionalMetadata`` model."""
 
     facts = FactSerializer(many=True)
-    certificate_info = CertificateInfoSerializer()
+    certificate_info = CertificateInfoSerializer(allow_null=True)
     product_meta = ProductMetaSerializer(required=False, allow_null=True)
     start_date = serializers.DateTimeField()
     end_date = serializers.DateTimeField()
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -325,7 +325,7 @@ class Meta:
 
 class GeoLocationSerializer(BaseModelSerializer):
     """Serializer for the ``GeoLocation`` model."""
-    location_name = serializers.CharField(allow_null=True, max_length=128)
+    location_name = serializers.CharField(allow_blank=True, allow_null=True, max_length=128)
```
DawoudSheraz, line 328: this is not allowed.

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -57,7 +57,7 @@ def inner(*args, **kwargs):
             content = str(e)
             if hasattr(e, 'content'):
                 content = e.content.decode('utf8') if isinstance(e.content, bytes) else e.content
-            msg = _('Failed to set data: {}').format(content)
+            msg = _('Faileding to set data: {} {}').format(content, type(e))
```
DawoudSheraz, line 60: Faileding?

course_discovery/apps/api/utils.py
```diff
@@ -103,8 +103,34 @@ def reviewable_data_has_changed(obj, new_key_vals, exempt_fields=None):
...
+            if old_dict_serializer.is_valid():
+                old_dict_value = old_dict_serializer.validated_data
+            else:
+                continue
+
+            for prop in new_value:
+                if new_value[prop] != old_dict_value[prop]:
+                    changed = True
+                    break
+        elif isinstance(new_value, list):
```
DawoudSheraz, line 133: Why can't these fields be declared exempt_fields? Also, geoloc and location restriction are on Course level. How are they affecting course run? OR is it happening from course api?

zawan-ila: Happening from [course api](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L420)

PR #3888, opened 2023-04-13
feat: handle introRichText in degree contentful data

DawoudSheraz:
### Description

In Degree contentful data transformation
- Add default value for introduction if it is not provided
- Add intro_rich_text field and give priority to introduction if present.

### DIFF
--- a/course_discovery/apps/course_metadata/contentful_utils.py
+++ b/course_discovery/apps/course_metadata/contentful_utils.py
@@ -251,10 +251,12 @@ def get_partnership_module(partnership_module):
 
 def get_featured_products_module(featured_products_module):
     """
-        Given a Contentful fetaured products module, extracts related data and returns them in the form of a dict.
+    Given a Contentful featured products module, extracts related data and returns them in the form of a dict.
     """
     featured_products_heading = featured_products_module.heading
-    featured_products_introduction = featured_products_module.introduction
+    featured_products_introduction_rich_text = rich_text_to_plain_text(
+        getattr(featured_products_module, 'intro_rich_text', {})
+    )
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -264,7 +266,8 @@ def get_featured_products_module(featured_products_module):
             })
     return {
         'heading': featured_products_heading,
-        'introduction': featured_products_introduction,
+        'introduction': getattr(featured_products_module, 'introduction', ''),
+        'introduction_rich_text': featured_products_introduction_rich_text,
```
Ali-D-Akbar: You can simply go for a single `introduction` field here. Just prefer the rich_text field over the plain introduction.

Ali-D-Akbar (commented):
one comment before approval.

course_discovery/apps/course_metadata/contentful_utils.py
DawoudSheraz: Both fields have values in Contentful at this moment, introduction being required and richText optional. So that's why I went for separate fields instead of giving pref to introRichText.

GPKyte: Be sure to test that the expected rich or plain content reaches the introduction field both in the presence and in the absence of rich data. One test broke while expecting the plain introduction. https://github.com/openedx/course-discovery/actions/runs/4700819930/jobs/8336001450?pr=3888#step:4:1775

Ali-D-Akbar (approved):

AliAdnanSohail (approved):

PR #3894, opened 2023-04-17
feat: move ingested courses to legal review state directly

AliAdnanSohail:
[Newly created EE and Boot Camp offerings should start in the "legal review"](https://2u-internal.atlassian.net/browse/PROD-2822)

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -252,6 +252,11 @@ class CSVDataLoader(AbstractDataLoader):
                     self._register_ingestion_error(CSVIngestionErrors.COURSE_RUN_UPDATE_ERROR, error_message)
                     continue
 
+            if course_run.status == CourseRunStatus.Unpublished:
+                course_run.refresh_from_db()
+                course_run.status = CourseRunStatus.LegalReview
+                course_run.save(update_fields=['status'])
+
             logger.info("Course and course run updated successfully for course key {}".format(course_key))  # lint-amnesty, pylint: disable=logging-format-interpolation
             self.course_uuids[str(course.uuid)] = course_title
             self._register_successful_ingestion(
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -252,6 +252,12 @@ def ingest(self):  # pylint: disable=too-many-statements
                     self._register_ingestion_error(CSVIngestionErrors.COURSE_RUN_UPDATE_ERROR, error_message)
                     continue
 
+            # Update course run status to In Legal Review
```
DawoudSheraz: nit: Either add more details to this comment or remove it.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -252,6 +252,12 @@ def ingest(self):  # pylint: disable=too-many-statements
...
+            if course_run.status == CourseRunStatus.Unpublished:
+                course_run.refresh_from_db()
+                course_run.status = CourseRunStatus.LegalReview
+                course_run.save()
```
AfaqShuaib09: can we use `update` over here instead of `save` call?

DawoudSheraz: use update_fields param to only update status for performance benefits.

AliAdnanSohail: `update` can not be called for a model instance. that is called on a Queryset. And we can't use queryset here that can decrease performance

AliAdnanSohail: There is some logic for sending email if status is in review in `save` method. So we may need that logic

DawoudSheraz: I dont think that will be effected, because the emails are sent based on status changes in save().

AfaqShuaib09 (approved):

PR #3901, opened 2023-04-20
feat: add program loader to ingest external program products

uzairr:
In order to add external program products to our system, a management command is added so that programs can be incorporated to the system based on their product source.

PROD-3209


DawoudSheraz:
This has been done in a different PR.

PR #3902, opened 2023-04-20
fix: give value per click default of 5

bseverino:
We’re updating product values (i.e. value per click USA) to be more granular. As part of this effort, we want new products added to the catalog to automatically take on the median value: 5. This way, they won’t get buried before we can add a value for them.

Internal ticket: https://2u-internal.atlassian.net/browse/WS-3860


course_discovery/apps/course_metadata/models.py
```diff
@@ -1056,12 +1056,12 @@ class ProductValue(TimeStampedModel):
...
     )
     per_click_usa = models.IntegerField(
-        null=True, blank=True, default=0, verbose_name=_('U.S. Value Per Click'), help_text=_(
+        null=True, blank=True, default=5, verbose_name=_('U.S. Value Per Click'), help_text=_(
             'U.S. value per click in U.S. dollars.'
         )
     )
     per_click_international = models.IntegerField(
-        null=True, blank=True, default=0, verbose_name=_('International Value Per Click'), help_text=_(
+        null=True, blank=True, default=5, verbose_name=_('International Value Per Click'), help_text=_(
```
DawoudSheraz, line 1064: This won't be impacting existing entries. Will they be updated manually?

DawoudSheraz (approved):
- To add related unit test before merge

course_discovery/apps/course_metadata/models.py
bseverino: Yes existing entries will be updated, we've got a separate ticket to handle that

bseverino:
@DawoudSheraz Unit test added

PR #3909, opened 2023-04-26
feat: Add a command to manage program subscription

uzairr:
In order to add subscriptions for the existing programs, a management command is added.It will process the programs based on their uuids and attach it to its subscription.

PROD-3209


course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,117 @@
...
+            type=str,
+            default='PROGRAMS',
+            choices=['PROGRAMS']
+        )
+        parser.add_argument(
+            '--product_source',
+            help='Slug of product source with which ingested degrees are to be linked.',
+            type=str,
+            required=True
+        )
```
DawoudSheraz: How/why are these arguments used here?  Both are irrelevant.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def handle(self, *args, **options):
+        """
+        Example usage: ./manage.py manage_program_subscription --csv_path=test.csv
+        """
+        csv_path = options.get('csv_path', None)
+        program_loader_config = ProgramDataLoaderConfiguration.current()
```
DawoudSheraz: The name of the loader config can be better. It should be ProgramSubscriptionDataLoader or something along these lines. ProgramDataLoaderConfiguration feel likes a complete program is being loaded.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,117 @@
...
+        # repeatedly invalidating the cache.
+        for model in apps.get_app_config('course_metadata').get_models():
+            for signal in (post_save, post_delete):
+                signal.disconnect(receiver=api_change_receiver, sender=model)
+
+        try:
+            if args_from_env:
+                product_config = settings.PRODUCT_METADATA_MAPPING[product_type][source.slug]
+                gspread_client = GspreadClient()
+                reader = gspread_client.read_data(product_config)
```
DawoudSheraz: PRODUCT_METADATA_MAPPING is used for external products' ingestion. Please add some other setting variable for this.

DawoudSheraz: Also, why is gspread flow needed here? Shouldn't the data loader config be enough here?

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,117 @@
...
+            program_uuid = row.get('uuid', None)
+            price = float(row.get('subscription_price', None).strip('$'))
+            try:
+                program = Program.objects.get(uuid=program_uuid)
+                subscription, _ = ProgramSubscription.objects.update_or_create(
+                    program=program, subscription_eligible=True
+                )
+                subscription_price, _ = ProgramSubscriptionPrice.objects.update_or_create(
+                    program_subscription=subscription, price=price
+                )
```
DawoudSheraz: Please add in logs if Subscription and Price are created/updated.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,117 @@
...
+    help = 'Import programs from a CSV available either through a provided csv file path ' \
+           'or a CSV file uploaded through ProgramDataLoaderConfiguration in django admin.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--csv_path',
+            help='Path to the CSV file',
+            type=str,
+        )
+
```
DawoudSheraz: If we are using LoaderConfig, this argument wont be of much help in this mgt. command.

course_discovery/settings/base.py
```diff
@@ -695,6 +695,14 @@
...
     },
+    'PROGRAMS': {
+        'ext_source': {
+            'SHEET_ID': '',
+            'INPUT_TAB_ID': '',
+            'ERROR_TAB_ID': '',
+            'EMAIL_NOTIFICATION_LIST': []
+        }
+    },
 }
```
DawoudSheraz: This is not needed here. If we are taking inspiration from other loaders, we should only use the items that are applicable in this case and not ditto copy every different path.

DawoudSheraz (changes requested):
- Unit tests are missing
- Questions and feedback on some of the implementation approaches.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
uzairr: gspread is needed as the data is needed to be imported from Spreadsheet.It is specifically communicated.Based on this requirement, product type and product source is naturally desired and i think it will make it flexible for the future in the context of sources other than ponderosa.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
uzairr: they are needed in terms of having a generic implementation for not only ponderosa but for all future subscription ingestion.

course_discovery/settings/base.py
uzairr: its not about ditto copy thing, the only thing which is debatable here is `ERROR_TAB_ID`. Other than this, nothing is irrelevant.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
DawoudSheraz: How? Subscription is related to Product and product is already linked to a source.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
DawoudSheraz: - Did we discuss the usage of spreadsheets for subscription addition? I am under the impression that the team will be using admin + Jenkins for this. I don't know how this will be useful for them if this is going to be limited run operation.

uzairr: Initially i was also under the same impression but later on, an explicit request came to fetch data from Google sheet.Maybe, later on down the road, they will make changes at their own.

Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,99 @@
...
+                product_config = settings.SUBSCRIPTION_METADATA_MAPPING
+                gspread_client = GspreadClient()
+                reader = gspread_client.read_data(product_config)
+
+            else:
+                reader = list(unicodecsv.DictReader(csv_file))
+
+        except Exception:
+            logger.exception("Error reading the input data source")
+            raise  # re-raising exception to avoid moving the code flow
```
DawoudSheraz: nit: raise CommandError here.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -0,0 +1,99 @@
...
+                subscription_price, created = ProgramSubscriptionPrice.objects.update_or_create(
+                    program_subscription=subscription, price=price
+                )
+                created_or_updated = 'created' if created else 'updated'
+                logger.info('Program located with slug: %s.Its subscription with price: %s USD is %s',
+                            program.marketing_slug, subscription_price.price, created_or_updated)
+            except Program.DoesNotExist:
+                raise CommandError(  # pylint: disable=raise-missing-from
+                    "Unable to locate Program instance with code {}".format(program_uuid)
+                )
```
DawoudSheraz: raise will stop the command execution. The error should be logged so that execution can keep on happening for other rows.

DawoudSheraz (approved):
A few nits/comments to address before merge.

uzairr:
All comments and suggestions are addressed.i ll going to merge it as the build will pass

PR #3914, opened 2023-04-28
feat: return default product values in algolia

bseverino:
Follow up to https://github.com/openedx/course-discovery/pull/3902. Although default values were added in that PR, `in_year_value` is still a foreign key field, so entries won't be created automatically on course/program creation.

This is mainly important for search. Instead of creating unnecessary `ProductValue` entries when courses/programs are created (which can get messy), this simply returns the default values in Algolia when `in_year_value` is null, in order to ensure that new products don't get lost in the ranking before we get the chance to add product values for them.

Internal ticket: https://2u-internal.atlassian.net/browse/WS-3859


DawoudSheraz (approved):

Ali-D-Akbar (approved):
To be merged as soon as the branch is rebased

PR #3920, opened 2023-05-03
chore: upgrade taxonomy-connector to 1.37.3

muhammad-ammar:


requirements/local.txt
```diff
@@ -772,7 +772,7 @@ requests-oauthlib==1.3.1
     #   getsmarter-api-clients
     #   google-auth-oauthlib
     #   social-auth-core
-requests-toolbelt==0.10.1
+requests-toolbelt==1.0.0
```
saleem-latif: Can we make sure this will not cause any issues? otherwise, maybe pin this version?

saleem-latif (approved):
Just one comment, once it is sorted out 👍🏼

requirements/local.txt
muhammad-ammar: There is a [breaking change](https://github.com/requests/toolbelt/blob/master/HISTORY.rst#100----2023-05-01) in this version but I don't think it will break anything. Any thoughts?

jajjibhai008 (approved):

sameenfatima78 (approved):

Ali-D-Akbar (approved):

PR #3921, opened 2023-05-03
chore: upgrade taxonomy connector to v1.38.0

justinhynes:


requirements/local.txt
```diff
@@ -772,7 +772,7 @@ requests-oauthlib==1.3.1
     #   getsmarter-api-clients
     #   google-auth-oauthlib
     #   social-auth-core
-requests-toolbelt==0.10.1
+requests-toolbelt==1.0.0
```
justinhynes: changes: https://pypi.org/project/requests-toolbelt/

Breaking Change: Removed Google App Engine support to allow using urllib3 2.0. 

I think this is fine for us.

requirements/local.txt
```diff
@@ -876,7 +876,7 @@ stevedore==5.0.0
     #   code-annotations
     #   edx-django-utils
     #   edx-opaque-keys
-taxonomy-connector==1.37.2
+taxonomy-connector==1.38.0
```
MaxFrank13, line 881: Just so I understand correctly, this PR was created to update this plugin here. Everything else is just a normal dependency update, right?

MaxFrank13 (approved):

requirements/local.txt
justinhynes: Yep! I just ran `make upgrade` to create this PR, which would pull the latest version of the taxonomy connector (as well as all the other dependencies for the IDA).

requirements/local.txt
justinhynes: Apparently this _was_ an issue and after I made this PR there was an update to the constraints. This version update has been reverted.

Ali-D-Akbar (approved):

PR #3925, opened 2023-05-05
feat:  Adding video clip flag

zxcs001:
Add new course type video clip

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -555,6 +555,7 @@ class CourseRunType(TimeStampedModel):
     - "Verified and Audit"
     - "Audit only"
     - "Masters only"
+    - "Video clip"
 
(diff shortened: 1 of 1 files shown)


DawoudSheraz (commented):
- The PR does not explain why is this needed and how it is a course type
- What are the expected use-cases for this?

zxcs001:
> * The PR does not explain why is this needed and how it is a course type
> * What are the expected use-cases for this?

The content team is starting to publish brief video clips taken from boot camps, via course about pages. The problem with this solution, though, is that the course about pages include statistics that aren’t really relevant to the video clips。

image

If the course is a video clip, the statistics bar should not appear (in the screenshot above, the bar with 1 weeks, self-paced and free)

DawoudSheraz:
Closing for now, feel free to re-open if this is needed.

PR #3934, opened 2023-05-09
feat: Default coordinates in Algolia

aheykamp2u:
Adds a default set of coordinates to courses and programs into Algolia during indexing.
The default coordinates are the Langham, MD Headquarters

### DIFF
--- a/course_discovery/apps/course_metadata/algolia_models.py
+++ b/course_discovery/apps/course_metadata/algolia_models.py
@@ -19,6 +19,10 @@ from course_discovery.apps.course_metadata.utils import transform_skills_data
 # Algolia can't filter on an empty list, provide a value we can still filter on
 ALGOLIA_EMPTY_LIST = ['null']
 
+# Every record needs geolocation data to show up in results after we turn on georanking.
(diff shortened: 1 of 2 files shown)


Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -19,6 +19,10 @@
 # Algolia can't filter on an empty list, provide a value we can still filter on
 ALGOLIA_EMPTY_LIST = ['null']
 
+# Every record needs geolocation data to show up in results after we turn on georanking.
+# These are the coordinates of the HQ office in Maryland
+ALGOLIA_DEFAULT_GEO_COORDINATES = 38.951302, -76.873100
```
DawoudSheraz, line 24: nit: not a blocker for now but this can be converted into settings and provided via edx-internal so that we don't need to add a PR in Discovery if we ever need to change this,

DawoudSheraz (approved):

DawoudSheraz (approved):

DawoudSheraz (approved):

PR #3940, opened 2023-05-12
fix: remove infinite loop in algolia indexing job

aheykamp2u:
Fixes the infinite loop that was introduced in my previous commit.

### DIFF
--- a/course_discovery/apps/course_metadata/algolia_models.py
+++ b/course_discovery/apps/course_metadata/algolia_models.py
@@ -134,9 +134,8 @@ class AlgoliaProxyProduct(Program):
             contentful_product = contentful_data[product_uuid]
             self.product.contentful_fields = contentful_product
 
-    @property
     def coordinates(self):
-        return self.coordinates
+        return self.product.coordinates()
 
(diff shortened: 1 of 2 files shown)


AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -134,9 +134,12 @@ def __init__(self, product, language='en', contentful_data=None):
             contentful_product = contentful_data[product_uuid]
             self.product.contentful_fields = contentful_product
 
-    @property
     def coordinates(self):
-        return self.coordinates
+        geolocation = getattr(self.product, 'geolocation', None)
+        if geolocation:
+            return getattr(geolocation, 'coordinates', ALGOLIA_DEFAULT_GEO_COORDINATES)
```
DawoudSheraz: Can there be a case where coordinates on geolocation are None?

aheykamp2u: Reading through the model, this is a good point. I'm going to make the model return none if either lat or lon are null

DawoudSheraz (approved):

PR #3947, opened 2023-05-17
feat: use simple history excluded_fields to always skip modified fields in diff check

DawoudSheraz:
### Description
Use clean_duplicate_history's excluded_fields mechanism to skip `modified` and provided excluded_fields to in history diff_check. Most models are inherited from Timestamped model, which updates modified whenever obj is saved, regardless of change. Previously, this was happening manually. 

- https://github.com/jazzband/django-simple-history/blob/master/simple_history/management/commands/clean_duplicate_history.py


uzairr (approved):
LGTM 👍

zawan-ila (approved):

PR #3953, opened 2023-05-23
feat: add subscription flag

uzairr:
Process the subscription eligible column to determine the eligiblity state of the program subscription.

PON-224

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
@@ -3114,12 +3114,21 @@ VALID_PROGRAM_SUBSCRIPTION_DICT = {
     'uuid': '0668c6f0-5dcf-41dc-80cd-6222d2de4df0',
     'title': 'Test Program',
     'subscription price': '$79.89',
+    'subscription_eligible': 'TRUE',
 }
 
 INVALID_PROGRAM_SUBSCRIPTION_DICT = {
     'uuid': '0668eef0-5dcf-41dc-80cd-6222d2de4d22',
     'title': 'Test Program 2',
     'subscription price': '$79.99',
+    'subscription_eligible': 'FALSE',
+}
+
+PROGRAM_WITH_INVALID_SUBSCRIPTION_ELIGIBILITY_DICT = {
+    'uuid': '0668eef0-5dcf-41dc-80cd-6222d2de4d22',
+    'title': 'Test Program 2',
+    'subscription price': '$79.99',
+    'subscription_eligible': 'test-flag',
 }
 
 VALID_MINIMAL_DEGREE_CSV_DICT = VALID_DEGREE_CSV_DICT
--- a/course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
+++ b/course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
@@ -67,19 +67,29 @@ class Command(BaseCommand):
             program_uuid = row.get('uuid', None)
 
             price = Decimal(row.get('subscription_price', None).strip('$'))
+            subscription_eligible = row.get('subscription_eligible', None)
+
+            if subscription_eligible not in ['TRUE', 'FALSE']:
+                logger.info(f"Skipped record: {program_uuid} because of invalid subscription eligibility value")
+                continue
+
             try:
+                subscription_eligible = subscription_eligible == 'TRUE'
                 program = Program.objects.get(uuid=program_uuid)
+                default_params = {
+                    'subscription_eligible': subscription_eligible
+                }
                 subscription, _ = ProgramSubscription.objects.update_or_create(
-                    program=program, subscription_eligible=True
+                    program=program, defaults=default_params
                 )
                 subscription_price, created = ProgramSubscriptionPrice.objects.update_or_create(
                     program_subscription=subscription, price=price
                 )
                 created_or_updated = 'created' if created else 'updated'
-                logger.info('Program located with slug: %s.Its subscription with price: %s USD is %s',
-                            program.marketing_slug, subscription_price.price, created_or_updated)
+                logger.info(f"Program located with slug: {program.marketing_slug}."
+                            f"Its subscription with price: {subscription_price.price} USD is {created_or_updated}")
             except Program.DoesNotExist:
-                logger.exception("Unable to locate Program instance with code %s", program_uuid)
+                logger.exception(f"Unable to locate Program instance with code {program_uuid}")
 
     def transform_dict_keys(self, data):
         """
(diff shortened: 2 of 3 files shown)


course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -96,3 +97,18 @@ def transform_dict_keys(self, data):
...
+        Evaluates a provided string to its bool representation
+        """
+        val = val.lower()
+
+        if val in ['false', 'f', '0']:
+            return False
+
+        elif val in ['true', 't', '1']:
+            return True
+
```
DawoudSheraz: nit: We should define what is expected in subscription_eligible column and skip the row if the value is not correct.  This. seems like a weird way that we have to check each option, in my opinion.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3114,12 +3114,14 @@
     'uuid': '0668c6f0-5dcf-41dc-80cd-6222d2de4df0',
     'title': 'Test Program',
     'subscription price': '$79.89',
+    'subscription_eligible': 'TRUE',
```
AfaqShuaib09, line 3117: Shouldn't `TRUE` be in lowercase like it is in the `to_bool` method below?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3114,12 +3114,14 @@
...
 }
 
 INVALID_PROGRAM_SUBSCRIPTION_DICT = {
     'uuid': '0668eef0-5dcf-41dc-80cd-6222d2de4d22',
     'title': 'Test Program 2',
     'subscription price': '$79.99',
+    'subscription_eligible': 'FALSE',
```
AfaqShuaib09, line 3124: same here

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
uzairr: after fetching the CSV all the keys are transformed to lowercase! so not needed to include the uppercase string.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -67,10 +67,16 @@ def handle(self, *args, **options):
             program_uuid = row.get('uuid', None)
 
             price = Decimal(row.get('subscription_price', None).strip('$'))
+            subscription_eligible = row.get('subscription_eligible', None)
+
+            if subscription_eligible not in ['TRUE', 'FALSE']:
+                continue
```
AfaqShuaib09, line 74: why we are using `continue` here, Don't we need to update the program's `subscription_eligible` value to `False` if `subscription_eligible` is `none`?
I guess we can add some `logging` here instead of just using `continue`

uzairr: `continue` is used to skip the row which don't has a boolean representation, it will cover the `None` case that you mentioned.
Having a log is a good idea! 👍

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
DawoudSheraz: Only keys are transformed, not values?

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
DawoudSheraz: logging is +1.

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -67,10 +67,16 @@ def handle(self, *args, **options):
...
+
             try:
+                subscription_eligible = subscription_eligible == 'TRUE'
                 program = Program.objects.get(uuid=program_uuid)
                 subscription, _ = ProgramSubscription.objects.update_or_create(
-                    program=program, subscription_eligible=True
+                    program=program, subscription_eligible=subscription_eligible
```
DawoudSheraz: Let's say we toggle the subscription, this line would result in the creation of a new object. We should use defaults to handle subscription changes.

DawoudSheraz: Verify and add appropriate unit tests for this.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
uzairr: yes, just keys

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -67,10 +67,20 @@ def handle(self, *args, **options):
...
+
             try:
+                subscription_eligible = subscription_eligible == 'TRUE'
                 program = Program.objects.get(uuid=program_uuid)
+                defaults_param = {
+                    'subscription_eligible': subscription_eligible
+                }
                 subscription, _ = ProgramSubscription.objects.update_or_create(
-                    program=program, subscription_eligible=True
+                    program=program, defaults=defaults_param
```
AfaqShuaib09: Based on my understanding, the purpose of this piece of code is to update the fields of a `ProgramSubscription` object if a matching object is found against a specified `program`. This is done by using the values provided in the defaults_param dict
Am I correct over here?

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
```diff
@@ -67,10 +67,20 @@ def handle(self, *args, **options):
             program_uuid = row.get('uuid', None)
 
             price = Decimal(row.get('subscription_price', None).strip('$'))
+            subscription_eligible = row.get('subscription_eligible', None)
+
+            if subscription_eligible not in ['TRUE', 'FALSE']:
+                logger.info("Skipped record: %s because of invalid subscription eligibility value", program_uuid)
```
AfaqShuaib09: nit: use fstring

AfaqShuaib09 (approved):
Just one nit & a question, otherwise LGTM

AfaqShuaib09 (approved):
Just one nit & a question, otherwise LGTM

AfaqShuaib09 (approved):
Just one nit & a question, otherwise LGTM

course_discovery/apps/course_metadata/management/commands/manage_program_subscription.py
uzairr: yes, absolutely correct 👍

DawoudSheraz (approved):

PR #3956, opened 2023-05-25
chore: use get method for stats fields in populate exec ed data csv to avoid errors

AliAdnanSohail:
[Ingest GS stage data to Discovery stage for Exec Ed](https://2u-internal.atlassian.net/browse/PROD-3378)

## Description
Make stats field as optional in populate exec ed data csv command


DawoudSheraz (approved):
The commit type should be chore instead of fix.

PR #3976, opened 2023-06-09
feat: handle timestamp updates for additional metadata in course api

DawoudSheraz:
### Description
This PR is a followup to https://github.com/openedx/course-discovery/pull/3867. One of the todos in the PR mentioned that AdditionalMetadata and its related models in course api are using .update() to update the underlying objects. When using queryset update(), save() is not called by Django internally, and thus no signals are sent. 

This PR changes AdditionalMetadata & related objects handling in course api to ensure the data modified timestamp is updated for Course when in course api:
- Any of field on AdditionalMetadata is updated in course
- CertificateInfo, one of FK on AdditionalMetadata, has one of its fields change
- Facts, a M2M field, has either m2m _relationship change or any of underlying Fact objects in the relationship is updated
- ProductMeta, one of FK on AdditionalMetadata, has its fields or **keywords** Taggable manager(M2M) change

### Testing
- Make sure you have one ExecEd course in your local discovery
- Checkout this branch.
- Run latest publisher on your local. 
- Go to ExecEd course. Change any of the main fields for AdditionalMetadata. Verify data modified timestamp for the course is updated
- Change any of CertificateInfo field and verify data modified timestamp for the course is updated
- Update product meta's title or description.  Verify data modified timestamp for the course is updated
- Add keywords on product meta.  Verify data modified timestamp for the course is updated
- Update product meta's title but do not change keyword. Verify from logs that m2m_changed is not called for ProductMeta
- Edit any fact object and verify data modified timestamp for the course is updated

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -33,6 +33,7 @@ from course_discovery.apps.api.fields import (
 from course_discovery.apps.api.utils import StudioAPI
 from course_discovery.apps.catalogs.models import Catalog
 from course_discovery.apps.core.api_client.lms import LMSAPIClient
+from course_discovery.apps.core.utils import update_instance
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.fields import HtmlField as MetadataHtmlField
 from course_discovery.apps.course_metadata.models import (
@@ -1389,8 +1390,11 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
 
         existing_facts = instance.facts.all()
         if len(existing_facts) == len(facts_data):
+            changed = False
             for i, fact in enumerate(existing_facts):
-                Fact.objects.filter(id=fact.id).update(**facts_data[i])
+                _, fact_changed = update_instance(fact, facts_data[i], True)
+                changed = changed or fact_changed
+            return changed
         else:
             instance.facts.clear()
             for fact in facts_data:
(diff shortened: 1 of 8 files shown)


course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -984,14 +985,77 @@ def test_remove_orphan_editors(self):
...
+        course = factories.CourseFactory(
+            draft=True,
+            additional_metadata=factories.AdditionalMetadataFactory(
+                product_meta=product_meta
+            )
+        )
+        course_timestamp = course.data_modified_timestamp
+
+        with LogCapture(LOGGER_NAME) as log:
+            product_meta.keywords.set(['keyword_1', 'keyword_2'], clear=True)
```
zawan-ila: Is a clear here necessary? The `pre_add` should fire without it too 🤔

course_discovery/apps/course_metadata/signals.py
```diff
@@ -402,6 +402,24 @@ def course_m2m_changed(sender, instance, action, **kwargs):
         instance.refresh_from_db()
 
 
+@receiver(m2m_changed, sender=ProductMeta.keywords.through)
+def course_taggable_manager_changed(sender, instance, action, **kwargs):
+    if action == 'pre_add' and not kwargs['reverse'] and instance._meta.label == 'course_metadata.ProductMeta':
```
zawan-ila: [confirm] This assertion on the label is because of not explicitly defining a through in TaggableManager?

zawan-ila: Why is `pre_remove` ignored?

course_discovery/apps/api/serializers.py
```diff
@@ -1408,52 +1412,68 @@ def update_facts(self, instance, facts_data):
...
 
         if instance.product_meta and keywords is not None:
-            instance.product_meta.keywords.set(keywords, clear=True)
-            instance.product_meta.save()
+            # Do not update M2M relation if the contents of keywords are unchanged.
+            # Since keywords is TaggableManager, clear=true will always trigger m2m_changed showing
+            # new relationships have been added. This can cause timestamp to update incorrectly even if there are
+            # no data updates.
+            if set(keyword.name for keyword in instance.product_meta.keywords.all()) != set(keywords):
+                instance.product_meta.keywords.set(keywords, clear=True)
```
zawan-ila: Why clear? To make the downstream handling easier? e.g I see you only care about pre_add on keywords later on. Without this, that'd need more work?

zawan-ila (approved):
Haven't tested on local. Leaving that for the second reviewer 😄

course_discovery/apps/api/serializers.py
DawoudSheraz: The tags can both be added or removed. Clear removes any existing tags and replace them with the incoming values. I think it might be possible that TaggableMgt is internally checking this, I will verify that.

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: - pre_remove -- mainly because API is handling everything via clear=True, but I will add check for that.
- The explicit label is needed because django-taggit uses [TaggedItem](https://github.com/jazzband/django-taggit/blob/master/taggit/models.py#L185) to store information on which tags are assigned to which Django model (via content type). TaggedItem is the [default](https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#L440) through value if it is not defined explicitly. In Discovery course_metadata app, the value for through is not defined whenever TaggableManager is used (be it Org, Course, ProductMeta, Program, etc). Now, to receive m2m_changed signal, we need to specify the sender. The sender is i[ntermediate or through model ](https://docs.djangoproject.com/en/4.2/ref/signals/#m2m-changed)(which is TaggedItem for tag manager). So, therefore, when m2m_changed is attempted to be set only for ProductMeta.keywords.through, since through=TaggedItem, whenever m2m for tags is changed on any other model, say Course, the m2m_changed handler is called because through is the same for Course & ProductMeta. (Hope it makes sense, Taken from https://2u-internal.atlassian.net/wiki/spaces/CAT/pages/356156060/Timestamp+Based+Data+Query#django-taggit-Taggable-Manager)

course_discovery/apps/api/serializers.py
DawoudSheraz: Yeah, it seems clear=True is not needed. Taggable manager handles that internally.

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: There might not be a need for pre_remove check. It seems django-taggit does not emit a m2m_changed signal against that event. m2m_changed is fired only with pre and post add. Ref https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#L152-L194

course_discovery/apps/api/serializers.py
DawoudSheraz: @zawan-ila updated.

AfaqShuaib09 (approved):
Tested changes on local, LGTM.
I noticed an issue on publisher. When updating the fields in the `Certificate Info` section, the submit button does not change its state on save. We can create a follow-up task to address this.

course_discovery/apps/core/tests/test_utils.py
```diff
@@ -59,6 +63,44 @@ def test_delete_orphans_with_exclusions(self):
...
+        assert changed is True
+        assert instance.description != updated_data['description']
+
+    def test_update_instance__no_instance(self):
+        """
+        Verify the default values if no instance is provided to update_instance.
+        """
+        instance, changed = update_instance(None, {})
+        assert instance is None
+        assert changed is False
```
Ali-D-Akbar: nit: `assert not changed`

course_discovery/apps/core/tests/test_utils.py
```diff
@@ -59,6 +63,44 @@ def test_delete_orphans_with_exclusions(self):
...
+    )
+    @unpack
+    def test_update_instance(self, instance_factory, updated_data):
+        """
+        Verify update_instance commit changes to DB if commit flag is True.
+        """
+        instance = instance_factory()
+        _, changed = update_instance(instance, updated_data, True)
+        instance.refresh_from_db()
+        assert changed is True
```
Ali-D-Akbar: nit: `assert changed`

course_discovery/apps/course_metadata/signals.py
```diff
@@ -402,6 +402,25 @@ def course_m2m_changed(sender, instance, action, **kwargs):
         instance.refresh_from_db()
 
 
+@receiver(m2m_changed, sender=ProductMeta.keywords.through)
+def course_taggable_manager_changed(sender, instance, action, **kwargs):
```
Ali-D-Akbar, line 406: Docstring missing

Ali-D-Akbar (approved):
Comments are not blockers.

PR #3977, opened 2023-06-09
Restructure course url slug

AliAdnanSohail:
## Testing instructions

- Take the pull of this branch
- Run migrations
- Add `NOTIFY_SLUG_UPDATE_RECIPIENTS` value in private.py
- Run command `./manage.py migrate_course_slugs` in discovery shell with following parameters
   There are 3 inputs of this command
   1. `--course_uuids <uuid_1> --course_uuids <uuid_2>` It will take course_uuid as input and run slug updates for those courses
   2. `--args_from_database True` upload a CSV in `MigrateCourseSlugConfiguration` model from django admin with a row containing `course_uuids` as header and its values. It will run slug updates for those courses
   3. `--limit <any number>` It will get first n records of courses from DB and run slug updates on those
   4. `--dry_run True` It will just dry run the slug update flow, will not commit in DB
   5. Admin can also give these command line arguments using MigrateCourseSlugConfiguration model.
   
Screen Shot 2023-06-22 at 1 04 12 AM


course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    def update_course_slug(self, course):
+        current_slug = course.active_url_slug
+        if validate_slug_format(current_slug):
+            self.log_slug_change(course, error='slug is already in correct format')
+        else:
+            new_slug = get_slug_for_course(course)
+            self.log_slug_change(course, new_slug)
+            course.set_active_url_slug(new_slug)
```
DawoudSheraz: We might want to check/add draft version handling as well. Verify the slug response from DB for Draft versions after updating slug and see if the special check is needed.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,76 @@
...
+            courses = Course.objects.filter(uuid__in=course_uuids)
+
+        for course in courses:
+            if dry_run:
+                new_slug = get_slug_for_course(course)
+                self.log_slug_change(course, new_slug)
+            else:
+                self.update_course_slug(course)
+
+        print(self.slug_update_report)
```
DawoudSheraz: Loop each element of report, and print in row form separated by comma so that it can be converted to CSV quickly if needed.

course_discovery/apps/course_metadata/models.py
```diff
@@ -3933,12 +3935,45 @@ class Meta:
...
+        )
+
+
 class CourseUrlSlug(TimeStampedModel):
     course = models.ForeignKey(Course, models.CASCADE, related_name='url_slug_history')
     # need to have these on the model separately for unique_together to work, but it should always match course.partner
     partner = models.ForeignKey(Partner, models.CASCADE)
-    url_slug = AutoSlugField(populate_from='course__title', editable=True, slugify_function=uslugify,
-                             overwrite_on_add=False, max_length=255)
+    url_slug = AutoSlugWithSlashesField(populate_from='course__title', editable=True,
```
DawoudSheraz, line 3970: populate_from is still needed here?

course_discovery/apps/course_metadata/models.py
```diff
@@ -3933,12 +3935,45 @@ class Meta:
         verbose_name_plural = 'Person Areas of Expertise'
 
 
+SLUG_DISALLOWED_CHARS = re.compile(r'[^-a-zA-Z0-9/]+')
+SLUG_ALLOWED_CHARS = re.compile(r'^[-a-zA-Z0-9_/]+\Z')
```
DawoudSheraz: some comments on these would be helpful.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,17 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
+def validate_slug_format(slug):
+    valid_slug_pattern = r"learn\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+$"
+    return bool(re.match(valid_slug_pattern, slug))
+
+
+def get_slug_for_course(course):
+    # TODO: error handling in case of missing data
+    primary_subject_slug = course.subjects.first().slug
+    organization_slug = slugify(course.authoring_organizations.first().name)
+    course_title_slug = slugify(course.title)
```
DawoudSheraz: we should have appropriate checks in place, in case any data element is missing.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,76 @@
...
+        dry_run = options.get('dry_run', False)
+        course_uuids = options.get('course_uuids', None)
+        csv_from_config = options.get('args_from_database', None)
+
+        if csv_from_config:
+            csv_loader_config = MigrateCourseSlugConfiguration.current()
+            csv_file = csv_loader_config.csv_file if csv_loader_config.is_enabled() else None
+            rows = list(unicodecsv.DictReader(csv_file))
+            course_uuids = [row['course_uuids'] for row in rows]
+            courses = Course.objects.filter(uuid__in=course_uuids)
```
DawoudSheraz: Since we are focusing on OCM courses only, we should fetch edX source objs only. or skip courses that have additional metadata assigned to them.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,17 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
         instance.update_product_data_modified_timestamp()
+
+
+def validate_slug_format(slug):
+    valid_slug_pattern = r"learn\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+$"
+    return bool(re.match(valid_slug_pattern, slug))
+
+
+def get_slug_for_course(course):
+    # TODO: error handling in case of missing data
```
DawoudSheraz: This format is only for OCM courses. Add that in docstring to clarity.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
AliAdnanSohail: I think Draft/Non-Draft thing is handled in `set_active_url_slug`

course_discovery/apps/course_metadata/models.py
AliAdnanSohail: This will be updated in other task https://2u-internal.atlassian.net/browse/PROD-3395, in this PR we are just allowing `/` in DB level

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if csv_from_config:
+            courses = self.get_courses_from_csv_config()
+
+        if course_uuids:
+            courses = self.get_courses_from_uuids(course_uuids)
+
+        if limit:
+            logger.info(f"Getting first {limit} open course records")
+            courses = Course.everything.filter(product_source__slug='edx')[:limit]
```
DawoudSheraz: everything will bring in both draft & non-draft courses. Is that ok?

DawoudSheraz: Maybe only get drafts and if non-draft exists on draft, update url in that case?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,37 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
     """
     if hasattr(instance, 'field_tracker') and hasattr(instance, 'update_product_data_modified_timestamp'):
         instance.update_product_data_modified_timestamp()
+
+
+def is_valid_slug_format(slug):
+    valid_slug_pattern = r"learn\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+$"
+    return bool(re.match(valid_slug_pattern, slug))
```
DawoudSheraz: There have been some format changes. learn/ would be present in all cases but the slug after that can differ.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -347,3 +347,14 @@ def send_ingestion_email(partner, subject, to_users, product_type, ingestion_det
...
+
+def send_email_for_slug_updates(stats):
+    subject = 'Slugs Update Summary'
+    to_users = settings.NOTIFY_SLUG_UPDATE_RECIPIENTS
+    import pdb
+    pdb.set_trace()
+    email_msg = EmailMultiAlternatives(
+        subject, str(stats), settings.PUBLISHER_FROM_EMAIL, to_users
+    )
+    email_msg.send()
```
DawoudSheraz: No email body?

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,131 @@
...
+            self.add_to_slug_update_report(course, error=error_msg)
+        else:
+            logger.info(f"Updating slug for course with uuid {course.uuid} and title {course.title}, "
+                        f"current slug is '{current_slug}'")
+            new_slug, error = get_slug_for_course(course)
+            self.add_to_slug_update_report(course, new_slug, error)
+            if not dry_run:
+                course.set_active_url_slug(new_slug)
+                if course.official_version:
+                    course.official_version.set_active_url_slug(new_slug)
```
DawoudSheraz: nit: to add log when draft version is updated.

DawoudSheraz (commented):
Some notes from my local testing:
- if an Integrity error occurs during slug set, the command stops execution. Instead, it should log the error and move forward with other slugs
- courses created in Studio and synced via LMS data loader do not have slug values (unless added from publisher). When checking for slug validation, verify if the active slug actually has a value.
- If a course has the same title, same org(?), and the same subject, it can fail during the slug setting. I had a few old test courses in my local where this caused Integrity error. Ideally, we would need to make course title slug unique (like it already happens) and use that to curate new slug instead of simple slugification of title

course_discovery/apps/course_metadata/emails.py
```diff
@@ -347,3 +347,11 @@ def send_ingestion_email(partner, subject, to_users, product_type, ingestion_det
...
             context
         )
+
+
+def send_email_for_slug_updates(stats, to_users):
+    subject = 'Slugs Update Summary'
+    email_msg = EmailMultiAlternatives(
+        subject, str(stats), settings.PUBLISHER_FROM_EMAIL, to_users
+    )
+    email_msg.send()
```
DawoudSheraz, line 357: P.S. not sure if we would get much value from sending the email.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,140 @@
...
+        parser.add_argument(
+            '--args_from_database',
+            help='Get CSV from MigrateCourseSlugConfiguration model',
+            type=str,
+        )
+        parser.add_argument(
+            '--limit',
+            help='Limit of number of courses to update their slugs',
+            type=int,
+            default=1,
```
DawoudSheraz: are we planning to define this in edx-internal? This will also mean deployment whenever we want to change the value. Plus, we can't define a starting point around this limit.

course_discovery/apps/course_metadata/emails.py
AliAdnanSohail: `str(stats)` will be sent as email body

course_discovery/apps/course_metadata/emails.py
AliAdnanSohail: We can add anyone to the email group, no need to look into jenkins logs. Email will have formatted logs so it will be useful

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
AliAdnanSohail: Defining a starting point seems to be complex, because we can't add a check on DB level to check the format of slugs, it will be too complex or we have to loop every record in our python code to check if format is correct or not.

AliAdnanSohail: Also I'm looking for some option to give these params from jenkins

course_discovery/apps/course_metadata/emails.py
DawoudSheraz: Understood, but this is going to be a limited run effort and not something that will be run periodically.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,143 @@
...
+                self._add_to_slug_update_report(course, error=error_msg)
+            else:
+                logger.info(f"Updating slug for course with uuid {course.uuid} and title {course.title}, "
+                            f"current slug is '{current_slug}'")
+                new_slug, error = get_slug_for_course(course)
+                self._add_to_slug_update_report(course, new_slug, error)
+                if not dry_run and new_slug:
+                    course.set_active_url_slug(new_slug)
+                    if course.official_version:
+                        logger.info(f"Updating slug for non-draft course with uuid {course.official_version.uuid} and "
```
DawoudSheraz: uuid of draft and non-draft is the same.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _update_course_slug(self, course, dry_run=False):
+        """
+        It will update course slug to new format if its not already in new format and commit in DB only if dry run is
+        False
+        """
+        try:
+            current_slug = course.active_url_slug
+            if current_slug and is_valid_slug_format(current_slug):
+                error_msg = f"Course with uuid {course.uuid} and title {course.title} slug is already in correct " \
```
DawoudSheraz: nit: title {} has slug {} already in correct format

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,143 @@
...
+            type=int,
+            default=0,
+        )
+
+    def handle(self, *args, **options):
+        """
+        It will execute the command to update slugs to new format
+        'learn/<primary_subject>/<organization_name>-course_title'
+        """
+        self.slug_update_report = []  # pylint: disable=attribute-defined-outside-init
```
DawoudSheraz: can't we add init for this?

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
DawoudSheraz: This is still unresolved, considering that the code below is especially handling official_versions on a draft. We are essentially fetching all the courses and repeating an operation on non-drafts.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,143 @@
...
+                'old_slug': course.active_url_slug,
+                'new_slug': new_slug,
+                'error': error
+            }
+        )
+
+    def _update_course_slug(self, course, dry_run=False):
+        """
+        It will update course slug to new format if its not already in new format and commit in DB only if dry run is
+        False
```
DawoudSheraz: lets avoid using `new` word and use instead focus on the actual slug format.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,143 @@
...
+        return self._get_courses_from_uuids(course_uuids)
+
+    @staticmethod
+    def _get_courses_from_uuids(course_uuids):
+        valid_course_uuids = []
+        for course_uuid in course_uuids:
+            if is_valid_uuid(course_uuid):
+                valid_course_uuids.append(course_uuid)
+            else:
+                logger.info(f"Skipping uuid {course_uuid} because of incorrect format")
```
DawoudSheraz: shouldn't this be added to report too?

course_discovery/apps/course_metadata/management/commands/tests/test_migrate_course_slugs.py
```diff
@@ -0,0 +1,154 @@
...
+        product_source = SourceFactory(slug='edx')
+        partner = PartnerFactory()
+        self.course1 = CourseFactory(draft=False, product_source=product_source, partner=partner)
+        self.course2 = CourseFactory(draft=False, product_source=product_source, partner=partner)
+        self.organization = OrganizationFactory(name='test-organization')
+        self.subject = SubjectFactory(name='business')
+
+        self.course1.authoring_organizations.add(self.organization)
+        self.course1.subjects.add(self.subject)
+        self.course1.save()
```
DawoudSheraz: do we need save() here? We are just adding m2m_relationships above.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -998,3 +1001,90 @@ def test_fetch_getsmarter_products__with_invalid_credentials(self, mock_getsmart
...
+    """
+
+    @ddt.data(
+        ('learn/primary-subject/organization-course-title', True),
+        ('learn/some-text/some-text-some-text', True),
+        ('learn/', False),
+        ('/learn/', False),
+        ('/media/', False),
+        ('media/primary-subject/organization_name-course_title', False),
+        ('learn2/primary-subject/organization-name-course-title', False),
```
DawoudSheraz, line 1019: Add one with format `learn/subject-org-title`

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,51 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
     """
     if hasattr(instance, 'field_tracker') and hasattr(instance, 'update_product_data_modified_timestamp'):
         instance.update_product_data_modified_timestamp()
+
+
+def is_valid_slug_format(slug):
+    valid_slug_pattern = r"learn\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+$"
```
DawoudSheraz, line 955: This pattern does not support `learn/<slug>`. It only supports learn/first-str/2nd-str

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,51 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
+    return bool(re.match(valid_slug_pattern, slug))
+
+
+def get_slug_for_course(course):
+    error = None
```
DawoudSheraz, line 970: Missing docstring.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,51 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
+    primary_subject_slug = course_subjects[0].slug
+    organization_slug = slugify(organizations[0].name.replace('\'', ''))
+    course_slug = course.active_url_slug
+    # course slug is None for courses which are created from studio
+    if not course_slug:
+        course_slug = course.title
+        slug = f"learn/{primary_subject_slug}/{organization_slug}-{course_slug}"
+        if is_slug_already_exists(slug, course):
+            logger.info(f"Slug '{slug}' already exist in DB, recreating slug by adding course id in course_title")
+            course_slug = f"{course.title}-{course.id}"
```
DawoudSheraz: I dont think we should add course.id in here. Rather, we can get count of courses with same title as course.title and estimate the count value to append at end of title in slug (computer-science-<n>)

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,51 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
+    slug = f"learn/{primary_subject_slug}/{organization_slug}-{course_slug}"
+    return slug, error
+
+
+def is_slug_already_exists(slug, course):
```
DawoudSheraz: nit: is_existing_slug

DawoudSheraz (commented):
Left some more followup comments.

course_discovery/apps/course_metadata/models.py
```diff
@@ -3933,12 +3935,45 @@ class Meta:
...
+    def formfield(self, **kwargs):
+        return super().formfield(
+            **{
+                "form_class": SlashSlugField,
+                "allow_unicode": self.allow_unicode,
+                **kwargs,
+            }
+        )
+
+
```
AfaqShuaib09: nit: move these custom fields to `fields.py`

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,51 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
+    # to avoid circular dependency
+    from course_discovery.apps.course_metadata.models import CourseUrlSlug  # pylint: disable=import-outside-toplevel
+    all_course_historical_slugs_excluding_current = CourseUrlSlug.objects.filter(
+        url_slug=slug, partner=course.partner).exclude(course__uuid=course.uuid)
+    return all_course_historical_slugs_excluding_current.exists()
+
+
+def is_valid_uuid(val):
```
AfaqShuaib09, line 1007: nit: missing docstring

AfaqShuaib09 (commented):
I have tested the changes locally, and everything seems to be working fine with all the options. Just found a case it is not ignoring the slug change if the slug contains single slash `/` after learn. Need to update the regular expression in the `is_valid_slug_format` method

There is a test failing on PR `TestCacheInvalidation::test_model_change` it can be resolved by adding `MigrateCourseSlugConfiguration` [here](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/tests/test_signals.py#L59-L65) in the model list

course_discovery/apps/course_metadata/tests/test_utils.py
AliAdnanSohail: As per discussed with Sidra, we are expecting two `/` in slugs. so this is not correct format

course_discovery/apps/course_metadata/utils.py
AliAdnanSohail: As per discussed with Sidra, we are expecting two `/` in slugs

course_discovery/apps/course_metadata/utils.py
AliAdnanSohail: Logically through SEO perspective this number doesn't matter much. So rather than looping over all records and then getting count and decide the next possible number, I decided to use course_id which will always be unique

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,153 @@
...
+from course_discovery.apps.course_metadata.emails import send_email_for_slug_updates
+from course_discovery.apps.course_metadata.models import Course, MigrateCourseSlugConfiguration
+from course_discovery.apps.course_metadata.utils import get_slug_for_course, is_valid_slug_format, is_valid_uuid
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """
+    It will update course url slugs to the new format i.e 'learn/<primary_subject>/<organization_name>-course_title'
```
DawoudSheraz: nit: to the format

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,153 @@
...
+        parser.add_argument(
+            '--limit',
+            help='Limit of number of courses to update their slugs',
+            type=int,
+            default=0,
+        )
+
+    def handle(self, *args, **options):
+        """
+        It will execute the command to update slugs to new format
```
DawoudSheraz: nit: omit "new" word

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,153 @@
...
+                    {
+                        'course_uuid': course_uuid,
+                        'old_slug': None,
+                        'new_slug': None,
+                        'error': error
+                    }
+                )
+                logger.info(error)
+
+        return Course.objects.filter(product_source__slug='edx', uuid__in=valid_course_uuids)
```
DawoudSheraz: Course.objects only returns non-draft versions. So, course.official_version will be None.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
```diff
@@ -0,0 +1,153 @@
...
+
+        send_email_for_slug_updates(self.slug_update_report, settings.NOTIFY_SLUG_UPDATE_RECIPIENTS)
+        self._log_report_in_csv_format()
+
+    def _add_to_slug_update_report(self, course, new_slug=None, error=None):
+        """
+        It will add course and slug information in slug_update_report to show stats
+        """
+        if error:
+            logger.warning(error)
```
DawoudSheraz: Do we need to log it here?

course_discovery/apps/course_metadata/management/commands/tests/test_migrate_course_slugs.py
```diff
@@ -0,0 +1,149 @@
...
+                    LOGGER_PATH,
+                    'INFO',
+                    f"course_uuid,old_slug,new_slug,error\n"
+                    f"{self.course1.uuid},{current_slug_course1},{self.course1.active_url_slug},None\n"
+                    f"{self.course2.uuid},{current_slug_course2},{self.course2.active_url_slug},None\n"
+                )
+            )
+
+            assert self.course1.active_url_slug == f"learn/business/test-organization-{current_slug_course1}"
+            assert self.course2.active_url_slug == f"learn/business/test-organization-{current_slug_course2}"
```
DawoudSheraz, line 149: Verify the slug on draft version as well.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
AliAdnanSohail: It will help for unit testing. I can find all logs of this file in a single LOGGER

course_discovery/apps/course_metadata/management/commands/tests/test_migrate_course_slugs.py
AliAdnanSohail: Since we are using .objects draft versions will be skipped.

course_discovery/apps/course_metadata/management/commands/migrate_course_slugs.py
AliAdnanSohail: So I'm now getting only Draft courses and if it has any official version, it will update that as well

course_discovery/apps/course_metadata/utils.py
AfaqShuaib09: Regex pattern need to be updated to `learn(?:\/[a-zA-Z0-9-]+){1,2}$`

course_discovery/apps/course_metadata/tests/test_utils.py
AfaqShuaib09: +1 here

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/management/commands/tests/test_migrate_course_slugs.py
```diff
@@ -0,0 +1,244 @@
...
+        self.course2.authoring_organizations.add(self.organization)
+        self.course2.subjects.add(self.subject)
+
+        self.course3_draft = CourseFactory(
+            draft=True, product_source=product_source, partner=partner, url_slug='test_course'
+        )
+        course_url_slug = CourseUrlSlugFactory(
+            url_slug='test_slug', is_active=True, partner=partner, course=self.course3_draft
+        )
+        self.course3_draft.url_slug_history.add(course_url_slug)
```
DawoudSheraz, line 36: nit: why is the addition to url_slug_history necessary?

course_discovery/apps/course_metadata/management/commands/tests/test_migrate_course_slugs.py
```diff
@@ -0,0 +1,244 @@
...
+            log_capture.check_present(
+                (
+                    LOGGER_PATH,
+                    'INFO',
+                    f"course_uuid,old_slug,new_slug,error\n"
+                    f"{self.course1.uuid},{current_slug_course1},{self.course1.active_url_slug},None\n"
+                    f"{self.course2.uuid},{current_slug_course2},{self.course2.active_url_slug},None\n"
+                )
+            )
+
```
DawoudSheraz: nit: you can combine these check_present calls -> check_present((tuple1), (tuple2))

course_discovery/apps/course_metadata/migrations/0327_auto_20230621_1959.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 3.2.17 on 2023-06-21 19:59
```
DawoudSheraz: nit: please convert this migration into a named migration.

course_discovery/apps/course_metadata/tests/test_utils.py
DawoudSheraz: To be addressed.

course_discovery/apps/course_metadata/utils.py
DawoudSheraz: +1.

DawoudSheraz (commented):
Nearly there, some more follow-up nits/comments. Also, can you share how is the email rendered? Since the email is sending stats as it is and not transforming, I suspect it is going to be a big JSON. We will still need to convert that, something we will need to do with Mgt command logs also.

course_discovery/apps/course_metadata/management/commands/tests/test_migrate_course_slugs.py
AliAdnanSohail: Course factory was creating 2 courses (draft and non-draft) with different title and different course_url_slug. So I've created a single course_url entry and add that to both models

AliAdnanSohail:
> Nearly there, some more follow-up nits/comments. Also, can you share how is the email rendered? Since the email is sending stats as it is and not transforming, I suspect it is going to be a big JSON. We will still need to convert that, something we will need to do with Mgt command logs also.

Locally emails are displayed in shell. And it will show stats dict in email. Since this is not the email we need periodically so json of the stats will work for testing purpose. We can format these stats to csv file or any other format if needed. But that will be done in follow up PR if required

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -998,3 +1001,90 @@ def test_fetch_getsmarter_products__with_invalid_credentials(self, mock_getsmart
...
+        assert slug == f"learn/{subject.slug}/{organization.name}-{course.title}"
+
+    def test_get_slug_for_course__with_existing_url_slug(self):
+        partner = PartnerFactory()
+        course1 = CourseFactory(title='test-title')
+        subject = SubjectFactory(name='business')
+        course1.subjects.add(subject)
+        organization = OrganizationFactory(name='test-organization')
+        course1.authoring_organizations.add(organization)
+        course1.partner = partner
```
DawoudSheraz, line 1073: nit: this partner can be defined when creating course via CourseFactory

course_discovery/apps/course_metadata/utils.py
```diff
@@ -944,3 +946,70 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
...
+    primary_subject_slug = course_subjects[0].slug
+    organization_slug = slugify(organizations[0].name.replace('\'', ''))
+    course_slug = course.active_url_slug
+    # course slug is None for courses which are created from studio
+    if not course_slug:
+        course_slug = course.title
+        slug = f"learn/{primary_subject_slug}/{organization_slug}-{course_slug}"
+        if is_existing_slug(slug, course):
+            logger.info(f"Slug '{slug}' already exist in DB, recreating slug by adding course id in course_title")
+            course_slug = f"{course.title}-{course.id}"
```
DawoudSheraz, line 991: We can raise error or do some count on Courses with the same title to estimate <n> to append after course title. Course ID is not appropriate here. We should remove or update this, in followup PR.

DawoudSheraz (approved):
Left 1-2 comments.

PR #3987, opened 2023-06-21
WS-4069: add learning_type facet to algolia index

bseverino:
Context: we currently have a "program type" filter in the product index, which is limited to just OC program types. We would like to introduce a new filter that includes all types of products, including courses, degree programs, etc.

This adds `learning_type` as a facet field, which, in most cases, returns a list with a single item representing the product type (`['Masters']`, `['Professional Certificate']`, `['Boot Camp']`, etc).

The exceptions to this are OC courses, which can also return the types of programs they belong to. For example, if a course belongs to both a Masters and Professional Certificate program, the return value would be `['Course', 'Masters', 'Professional Certificate']`.

Internal ticket: https://2u-internal.atlassian.net/browse/WS-4069


DawoudSheraz (approved):

bseverino:
@TheEarlyNerd Do you still want to take a look at this?

bseverino:
@kalebjdavenport Last chance 🙂

kalebjdavenport (approved):

AfaqShuaib09 (approved):

PR #3988, opened 2023-06-22
feat: automate url slug generation

uzairr:
To make a consistent SEO experience, url slug generation strategy is modified.It will introduce its hook right after the course is transitioned into its Review.

PROD-3395

### DIFF
--- a/course_discovery/apps/api/utils.py
+++ b/course_discovery/apps/api/utils.py
@@ -2,8 +2,10 @@ import base64
 import functools
 import logging
 import math
+import re
 from urllib.parse import parse_qsl, urlencode, urljoin
 
+from django.conf import settings
 from django.core.files.base import ContentFile
 from django.db.models.fields.related import ManyToManyField
 from django.utils.translation import gettext as _
@@ -13,9 +15,31 @@ from sortedm2m.fields import SortedManyToManyField
 
 from course_discovery.apps.core.api_client.lms import LMSAPIClient
 from course_discovery.apps.core.utils import serialize_datetime
-from course_discovery.apps.course_metadata.models import CourseRun
+from course_discovery.apps.course_metadata.constants import SUBDIRECTORY_SLUG_FORMAT_REGEX
+from course_discovery.apps.course_metadata.models import Course, CourseRun
+from course_discovery.apps.course_metadata.utils import get_slug_for_course
 
 logger = logging.getLogger(__name__)
+
+
+def set_subdirectory_slug_for_course(course_run):
+    """
+    Sets the active url slug for draft and non-draft courses if the current
+    slug is not validated as per the new format.
+    """
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
+    is_slug_in_subdirectory_format = bool(re.match(SUBDIRECTORY_SLUG_FORMAT_REGEX, draft_course.active_url_slug))
+    if not is_slug_in_subdirectory_format and draft_course.product_source.slug == settings.DEFAULT_PRODUCT_SOURCE_SLUG:
+        slug, error = get_slug_for_course(draft_course)
+        if slug:
+            draft_course.set_active_url_slug(slug)
+            if draft_course.official_version:
+                draft_course.official_version.set_active_url_slug(slug)
+        else:
+            raise Exception(  # pylint: disable=broad-exception-raised
+                f"Slug generation Failed: unable to set active url slug for course: {draft_course.key} "
+                f"with error {error}"
+            )
 
 
 def cast2int(value, name):
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -9,6 +9,8 @@ import pytz
 import responses
 from django.contrib.auth.models import Group
 from django.db.models.functions import Lower
+from django.test import override_settings
+from edx_toggles.toggles.testutils import override_waffle_switch
 from freezegun import freeze_time
 from rest_framework.reverse import reverse
 from rest_framework.test import APIRequestFactory
@@ -21,10 +23,14 @@ from course_discovery.apps.course_metadata.choices import CourseRunStatus, Progr
 from course_discovery.apps.course_metadata.models import CourseRun, CourseRunType, Seat, SeatType
 from course_discovery.apps.course_metadata.tests.factories import (
     CourseEditorFactory, CourseFactory, CourseRunFactory, CourseRunTypeFactory, CourseTypeFactory, OrganizationFactory,
-    PersonFactory, ProgramFactory, SeatFactory, SourceFactory, TrackFactory
+    PersonFactory, ProgramFactory, SeatFactory, SourceFactory, SubjectFactory, TrackFactory
 )
+from course_discovery.apps.course_metadata.toggles import IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED
+from course_discovery.apps.course_metadata.utils import is_valid_slug_format
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
 from course_discovery.apps.publisher.tests.factories import OrganizationExtensionFactory
+
+LOGGER_NAME = 'course_discovery.apps.api.utils'
 
 
 @ddt.ddt
@@ -32,11 +38,11 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
     def setUp(self):
         super().setUp()
         self.user = UserFactory(is_staff=True)
-        self.product_source = SourceFactory(name='test source')
+        self.product_source = SourceFactory(name='test source', slug='edx')
         self.client.force_authenticate(self.user)
         self.course_run = CourseRunFactory(course__partner=self.partner)
         self.course_run_2 = CourseRunFactory(course__key='Test+Course', course__partner=self.partner)
-        self.draft_course = CourseFactory(partner=self.partner, draft=True)
+        self.draft_course = CourseFactory(partner=self.partner, draft=True, product_source=self.product_source)
         self.draft_course_run = CourseRunFactory(course=self.draft_course, draft=True)
         self.draft_course_run.course.authoring_organizations.add(OrganizationFactory(key='course-id'))
         self.course_run_type = CourseRunTypeFactory(tracks=[TrackFactory()])
@@ -1279,3 +1285,43 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         response = self.client.patch(url, patch_transaction['body'], format='json')
 
         assert response.status_code == patch_transaction['status_code']
+
+    @override_settings(DEFAULT_PRODUCT_SOURCE_SLUG='edx')
+    @ddt.data(
+        (IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED, True),
+        (IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED, False),
+    )
+    @ddt.unpack
+    def test_url_restructuring_with_feature_flag_state(self, flag, state):
+        """
+        Tests url slug restructuring must work under its relevant feature flag
+        """
+        self.mock_patch_to_studio(self.draft_course_run.key)
+        course = self.draft_course_run.course
+        course.subjects.add(SubjectFactory(name='Subject1'))
+
+        self.draft_course_run.status = CourseRunStatus.Unpublished
+        self.draft_course_run.save()
+        url = reverse('api:v1:course_run-detail', kwargs={'key': self.draft_course_run.key})
+        body = {
+            'course': self.draft_course_run.course.key,  # required, so we need for a put
+            'start': self.draft_course_run.start,  # required, so we need for a put
+            'end': self.draft_course_run.end,  # required, so we need for a put
+            'run_type': str(self.draft_course_run.type.uuid),  # required, so we need for a put
+            'draft': False,
+        }
+
+        with override_waffle_switch(flag, active=state):
+            if state:
+                response = self.client.put(url, body, format='json')
+                draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
+                assert draft_course_run.course.active_url_slug.startswith('learn/')
+            else:
+                response = self.client.put(url, body, format='json')
+                draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
+                assert not draft_course_run.course.active_url_slug.startswith('learn/')
+
+            assert is_valid_slug_format(draft_course_run.course.active_url_slug)
+
+        assert response.status_code == 200, f"Status {response.status_code}: {response.content}"
+        assert draft_course_run.status == CourseRunStatus.LegalReview
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.conf import settings
 from django.db import models, transaction
 from django.db.models.functions import Lower
 from django.http.response import Http404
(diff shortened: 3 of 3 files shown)


course_discovery/apps/api/utils.py
```diff
@@ -199,6 +201,21 @@ def increment_character(character):
...
 
 
+def validate_or_create_course_slug(course_run):
+    course = course_run.course
+    slug_regex = r"learn\/+[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+$"
+    is_slug_compliant = bool(re.search(slug_regex, course.active_url_slug))
+    non_ocm_variants = [CourseType.EXECUTIVE_EDUCATION_2U, CourseType.BOOTCAMP_2U]
+    is_ocm = course.type.slug not in non_ocm_variants
+    if is_ocm and not is_slug_compliant:
+        primary_subject = course.subjects.first()
```
DawoudSheraz: Instead of using OCM/Non-OCM, just verify the source to be default(edX). That is much simpler now.

course_discovery/apps/api/utils.py
```diff
@@ -199,6 +201,21 @@ def increment_character(character):
...
+        org = course.authoring_organizations.first()
+        new_slug = f"learn/{slugify(primary_subject.slug)}/{slugify(org.name)}-{slugify(course.title)}"
+        course.set_active_url_slug(new_slug)
```
DawoudSheraz: - How are we dealing draft and non-draft here?
- What happens if slug setting fails due to uniqueness constraint? Does frontend receive appropriate error?

uzairr: 1. Only non-draft course runs are changed to `LegalReview` state.I am extracting course from that run.Can you elaborate what would be the expectation other than this ?
2. `set_active_url_slug` is raising the `IntegrityError` which ultimately hits the frontend, i believe.

DawoudSheraz: 1. Why are we only changing non-draft? We should follow the pattern from course api (draft, and then non-draft if applicable). 
2. Please confirm this.

course_discovery/apps/course_metadata/migrations/0328_auto_20230623_1635.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('course_metadata', '0327_create_migrate_course_slug_config_model'),
+        ('waffle', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_switch, delete_switch),
```
DawoudSheraz: That is not the correct way of creating and documenting flags. Please see:
-  OEP https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0017-bp-feature-toggles.html
- doc https://edx.readthedocs.io/projects/edx-toggles/en/latest/how_to/documenting_new_feature_toggles.html

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +13,26 @@
...
 from course_discovery.apps.core.api_client.lms import LMSAPIClient
 from course_discovery.apps.core.utils import serialize_datetime
-from course_discovery.apps.course_metadata.models import CourseRun
+from course_discovery.apps.course_metadata.models import Course, CourseRun
+from course_discovery.apps.course_metadata.utils import get_slug_for_course, is_valid_slug_format
 
 logger = logging.getLogger(__name__)
 
 
+def validate_or_create_course_slug(course_run):
```
DawoudSheraz: the name can be updated, a bit confusing because we are not sending any indicator of set vs valid.

course_discovery/apps/api/utils.py
uzairr: The PR code is updated as per the above suggestion.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -1279,3 +1281,56 @@ def test_change_status_and_ofac_info(self, patch_transaction):
...
+
+    @override_switch('is_new_slug_format_enabled', False)
+    def test_url_slug_not_created_if_switch_disabled(self):
+        """
+        Tests url slug generation does not happen if the relevant feature
+        flag is disabled.
+        """
+        self.mock_patch_to_studio(self.draft_course_run.key)
+        course = self.draft_course_run.course
+        course.subjects.add(SubjectFactory(name='Subject1'))
```
DawoudSheraz: These two tests can be combined in one, using ddt to define switch and expected outcomes.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -1,5 +1,6 @@
 import logging
 
+import waffle  # lint-amnesty, pylint: disable=invalid-django-waffle-import
```
DawoudSheraz: Instead of waffe, please use edx-toggles. That is go-to toggle package for Open edX.

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +13,26 @@
...
+    """
+    Validates or creates the active url slug for draft and non-draft course, relevant to the
+    provided course run
+    """
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
+    is_slug_valid = is_valid_slug_format(draft_course.active_url_slug)
+    if not is_slug_valid and draft_course.product_source.slug == 'edx':
+        slug, _ = get_slug_for_course(draft_course)
```
AliAdnanSohail: We need to do something with error as well. May be we can log error here as well, Also in case of error slug will be None, so need to check what is the behavior of `set_active_url_slug` if slug is None. Otherwise have to handle None explicitly

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +13,26 @@
...
+        draft_course.set_active_url_slug(slug)
```
AliAdnanSohail, line 31: We need to explicitly call this method for non-draft course as well. something like this
```
if draft_course.official_version:
  draft_course.official_version.set_active_url_slug(slug)
```

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +13,29 @@
...
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
+    is_slug_valid = is_valid_slug_format(draft_course.active_url_slug)
+    if not is_slug_valid and draft_course.product_source.slug == 'edx':
+        slug, _ = get_slug_for_course(draft_course)
+        if slug:
+            draft_course.set_active_url_slug(slug)
+            if draft_course.official_version:
+                draft_course.official_version.set_active_url_slug(slug)
+        else:
+            logger.error(f"Unable to set the slug for course: {draft_course.key}")
```
AliAdnanSohail: nit: we should log the error message as well

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +13,29 @@
...
 
 
+def set_course_slug(course_run):
+    """
+    Sets the active url slug for draft and non-draft courses if the current
+    slug is not validated as per the new format.
+    """
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
+    is_slug_valid = is_valid_slug_format(draft_course.active_url_slug)
+    if not is_slug_valid and draft_course.product_source.slug == 'edx':
```
DawoudSheraz: not valid? Should not this be "valid and not updated to new format"?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -1279,3 +1282,42 @@ def test_change_status_and_ofac_info(self, patch_transaction):
...
+                draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
+                assert draft_course_run.status == CourseRunStatus.LegalReview
+                assert is_valid_slug_format(draft_course_run.course.active_url_slug)
+            else:
+                response = self.client.put(url, body, format='json')
+                assert response.status_code == 200, f"Status {response.status_code}: {response.content}"
+                draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
+                assert draft_course_run.status == CourseRunStatus.LegalReview
+                assert bool(draft_course_run.course.active_url_slug) is True
+                assert not is_valid_slug_format(draft_course_run.course.active_url_slug)
```
DawoudSheraz: There are repeated code lines here -- the ones that are not effected by switch state. Move them out of if-else. Or better, add expected values in ddt data and remove if-else altogether.

course_discovery/apps/api/utils.py
DawoudSheraz: +1

course_discovery/apps/api/utils.py
DawoudSheraz: If the slug is not set, the flow should be blocked.

course_discovery/apps/api/utils.py
uzairr: if its valid then no need to create the new slug, if its not valid only then new slug will be generated!

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -286,6 +287,8 @@ def _update_course_run(self, course_run, draft, changed, serializer, request, pr
         # back into legal review if a non exempt field was changed (expected_program_name and expected_program_type)
         if not draft and (course_run.status == CourseRunStatus.Unpublished or non_exempt_update):
             save_kwargs['status'] = CourseRunStatus.LegalReview
+            if IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED.is_enabled():
```
DawoudSheraz: You can add source check here as well, to avoid calling the function if source is not 'edX'.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
uzairr: we can't mitigate the if-else check because of uncommon checks between two states, refactoring it out wherever possible 👍

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -1279,3 +1285,73 @@ def test_change_status_and_ofac_info(self, patch_transaction):
...
+                response = self.client.put(url, body, format='json')
+                draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
+                assert not draft_course_run.course.active_url_slug.startswith('learn/')
+
+            assert is_valid_slug_format(draft_course_run.course.active_url_slug)
+
+        assert response.status_code == 200, f"Status {response.status_code}: {response.content}"
+        assert draft_course_run.status == CourseRunStatus.LegalReview
+
+    # @mock.patch('course_discovery.apps.api.utils.get_slug_for_course', return_value=None)
```
AliAdnanSohail: nit: uncomment or remove this code

course_discovery/apps/api/utils.py
DawoudSheraz: Hmm, alright.

DawoudSheraz: The confusion here is that is_valid_slug_format checks either the new format(if the flag is enabled) or the old format(if the flag is not enabled). Ideally, for OCM courses, we should check if the current slug follows the sub-directory format. If not, generate the slug. The method `is_valid_slug_format` is meant for different purpose. It will work here but leaves a room for error.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -286,6 +288,8 @@ def _update_course_run(self, course_run, draft, changed, serializer, request, pr
         # back into legal review if a non exempt field was changed (expected_program_name and expected_program_type)
         if not draft and (course_run.status == CourseRunStatus.Unpublished or non_exempt_update):
             save_kwargs['status'] = CourseRunStatus.LegalReview
+            if IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED.is_enabled() and product_source.slug == 'edx':
```
DawoudSheraz: Use DEFAULT_PRODUCT_SOURCE_SLUG setting for the default(edx) source slug instead of hardcoding the value.

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +15,31 @@
...
 
 
+def set_course_slug(course_run):
+    """
+    Sets the active url slug for draft and non-draft courses if the current
+    slug is not validated as per the new format.
+    """
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
+    is_slug_valid = bool(re.match(settings.SUBDIRECTORY_SLUG_FORMAT, draft_course.active_url_slug))
+    if not is_slug_valid and draft_course.product_source.slug == 'edx':
```
DawoudSheraz: nit: use DEFAULT_PRODUCT_SOURCE_SLUG instead of hard-code
instead of slug_valid, use name like is_slug_in_subdirectory format, for consistency,

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -1279,3 +1285,43 @@ def test_change_status_and_ofac_info(self, patch_transaction):
...
+            'start': self.draft_course_run.start,  # required, so we need for a put
+            'end': self.draft_course_run.end,  # required, so we need for a put
+            'run_type': str(self.draft_course_run.type.uuid),  # required, so we need for a put
+            'draft': False,
+        }
+
+        with override_waffle_switch(flag, active=state):
+            if state:
+                response = self.client.put(url, body, format='json')
+                draft_course_run = CourseRun.everything.get(key=self.draft_course_run.key, draft=True)
```
DawoudSheraz, line 1317: these 2 lines can be moved outside. Only assertion is differing based on if-else.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -286,6 +289,9 @@ def _update_course_run(self, course_run, draft, changed, serializer, request, pr
         # back into legal review if a non exempt field was changed (expected_program_name and expected_program_type)
         if not draft and (course_run.status == CourseRunStatus.Unpublished or non_exempt_update):
             save_kwargs['status'] = CourseRunStatus.LegalReview
+            if IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED.is_enabled() and \
+                    product_source.slug == settings.DEFAULT_PRODUCT_SOURCE_SLUG:
+                set_course_slug(course_run)
```
DawoudSheraz: nit: set_course_slug can be refactored to set_subdirectory_slug_for_course

DawoudSheraz (approved):
Some small nits/refactors, otherwise good.

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +15,31 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+def set_course_slug(course_run):
+    """
+    Sets the active url slug for draft and non-draft courses if the current
+    slug is not validated as per the new format.
+    """
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
```
AliAdnanSohail, line 30: sorry I missed this before, can you please tell me why we can't get the course directly by using
`course_run.course`?

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +15,31 @@
...
+    is_slug_valid = bool(re.match(settings.SUBDIRECTORY_SLUG_FORMAT, draft_course.active_url_slug))
+    if not is_slug_valid and draft_course.product_source.slug == 'edx':
+        slug, _ = get_slug_for_course(draft_course)
+        if slug:
+            draft_course.set_active_url_slug(slug)
+            if draft_course.official_version:
+                draft_course.official_version.set_active_url_slug(slug)
+        else:
+            raise Exception(  # pylint: disable=broad-exception-raised
```
AliAdnanSohail, line 39: can we log error returned from `get_slug_for_course(draft_course)` as well? It will be easy to track why slug is not returned

course_discovery/apps/api/utils.py
```diff
@@ -13,11 +15,31 @@
...
 logger = logging.getLogger(__name__)
 
 
+def set_course_slug(course_run):
+    """
+    Sets the active url slug for draft and non-draft courses if the current
+    slug is not validated as per the new format.
+    """
+    draft_course = Course.everything.get(key=course_run.course.key, draft=True)
+    is_slug_valid = bool(re.match(settings.SUBDIRECTORY_SLUG_FORMAT, draft_course.active_url_slug))
```
AliAdnanSohail: can we use util method? we have added waffle switch check in that, so if switch is off then old format will be acceptable

course_discovery/settings/base.py
```diff
@@ -706,6 +706,8 @@
 
 DEFAULT_PRODUCT_SOURCE_SLUG = ''
 
+SUBDIRECTORY_SLUG_FORMAT = r"learn\/[a-zA-Z0-9-]+\/[a-zA-Z0-9-]+$"
```
AliAdnanSohail: I've added this regex in constants, either you can use that or if there is any specific reason to add in base.py, then I'll update in my code
https://github.com/openedx/course-discovery/pull/3999/files#diff-7c965dbd177fa3ab2b36a1eecebf53710226751f02b8bc5ba52bc8ad7fcc1151R9

course_discovery/apps/api/utils.py
uzairr: it has the non-draft variant thats why retrieving the draft here!

course_discovery/apps/api/utils.py
uzairr: raising the Exception here will serve the purpose because it is intentional to disrupt the flow here in case of error from `get_slug_for_course`

course_discovery/apps/api/utils.py
AliAdnanSohail: can't we do something like this `course_run.course.draft_version`? we can avoid looping over all courses

course_discovery/apps/api/utils.py
uzairr: this utility function is invoked only for subdirectory flow and its execution is controlled by the feature flag so thats why i modified it to not depend on `is_valid_format`

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
uzairr: yeah it seems, but put call is invoking different flow in both of the cases above which doesn't allow me to move it to common place!

AliAdnanSohail (approved):

PR #3996, opened 2023-06-27
fix: remove extra port

aht007:
The extra port is overriding the actual port in kubernetes which is causing Issues. Removing the extra port on the request of SRE team

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,9 @@ COPY package-lock.json package-lock.json
 COPY bower.json bower.json
 RUN npm install --production && ./node_modules/.bin/bower install --allow-root --production
 
+# Run webpack
+RUN webpack --config webpack.config.js
+
 # Expose canonical Discovery port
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -54,7 +54,6 @@ RUN npm install --production && ./node_modules/.bin/bower install --allow-root -
 
 # Expose canonical Discovery port
 EXPOSE 18381
-EXPOSE 8381
```
DawoudSheraz, line 57: - Will this impact local devstack?
- The commit type should be `build`.

nadeemshahzad (approved):

Dockerfile
aht007: No, this would't impact devstack as we are using the 18381 port as I mentioned. Reworded the commit message.

DawoudSheraz (approved):

DawoudSheraz (approved):

PR #4000, opened 2023-07-02
feat: modify course api update method for subdirectory slug format

AfaqShuaib09:
Description:
----------------

This PR modifies course API `course_update` method to accept the subdirectory slug format for OCM courses. The subdirectory slug format will be
`learn/<primary-subject>/<org-name>-<course-title>.

If any of CourseRun of the course is in IN_REVIEW_STATUS or in POST_REVIEW_STATUS, then only subdirectory slug format will be used for OCM courses. Otherwise, the both slug formats will be used for OCM courses. Currently, I'm allowing both slug formats for OCM courses for any case once all OCM courses are migrated to subdirectory slug format, then we can remove the old slug format by removing check:
```python
  # TODO: remove this check once all OCM courses are migrated to subdirectory slug format
        if not valid_slug_flag:
            validators.validate_slug(url_slug)
            valid_slug_flag = True
```

https://github.com/openedx/course-discovery/pull/4000/files#diff-eadad00c3534feef2ef1587465eec30387a58cd5b0b71fecc61793e7ee8e4127R1060-R1063

Testing Instructions:
- Run the containers (discovery, frontend-app-publisher, studio, ecommerce (optional))
- Go to the `frontend-app-publisher` repo and move to `.env.development` file and set the `IS_NEW_SLUG_FORMAT_ENABLED='true'`
- Try to set new url slug format for OCM course using publisher.
- Make sure it is accepting both slug formats for OCM courses.
- Try to set course url slug format for non-OCM course using publisher.
- Make sure it is not accepting subdirectory slug format for non-OCM courses.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -7,9 +7,11 @@ import ddt
 import pytest
 import pytz
 import responses
+from django.conf import settings
 from django.db import IntegrityError
 from django.db.models.functions import Lower
 from django.db.models.query import Prefetch
+from edx_toggles.toggles.testutils import override_waffle_switch
 from rest_framework.reverse import reverse
 from testfixtures import LogCapture
 
@@ -22,13 +24,14 @@ from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
 from course_discovery.apps.course_metadata.models import (
     AbstractLocationRestrictionModel, Course, CourseEditor, CourseEntitlement, CourseRun, CourseRunType, CourseType,
-    Fact, GeoLocation, ProductMeta, Seat
+    Fact, GeoLocation, ProductMeta, Seat, Source
 )
 from course_discovery.apps.course_metadata.tests.factories import (
     CourseEditorFactory, CourseEntitlementFactory, CourseFactory, CourseLocationRestrictionFactory, CourseRunFactory,
     CourseTypeFactory, GeoLocationFactory, LevelTypeFactory, OrganizationFactory, ProductValueFactory, ProgramFactory,
     SeatFactory, SeatTypeFactory, SourceFactory, SubjectFactory
 )
+from course_discovery.apps.course_metadata.toggles import IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED
 from course_discovery.apps.course_metadata.utils import ensure_draft_world
 from course_discovery.apps.publisher.tests.factories import OrganizationExtensionFactory
 
@@ -51,6 +54,14 @@ class CourseViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin
         self.org = OrganizationFactory(key='edX', partner=self.partner)
         self.course.authoring_organizations.add(self.org)
         self.refresh_index()
+
+        self.course_data = {
+            'title': 'Course title',
+            'url_slug': 'learn/physics/harvardx-applied-physics',
+            'partner': self.partner.id,
+            'key': self.course.key,
+            'type': str(self.audit_type.uuid),
+        }
 
     def tearDown(self):
         super().tearDown()
@@ -1062,6 +1073,25 @@ class CourseViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin
         course = Course.everything.get(uuid=self.course.uuid, draft=True)
         assert course.title == 'Course title'
         assert course.active_url_slug == 'manual'
+        assert course.organization_logo_override.url is not None
+        self.assertDictEqual(response.data, self.serialize_course(course))
+
+    @ddt.data('put', 'patch')
+    @responses.activate
+    def test_update_success_with_subdirectory_slug_format(self, method):
+        url = reverse('api:v1:course-detail', kwargs={'key': self.course.uuid})
+
+        source = Source.objects.get(slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG)
+        self.course.product_source = source
+        self.course.save()
+
+        with override_waffle_switch(IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED, active=True):
+            response = getattr(self.client, method)(url, self.course_data, format='json')
+            assert response.status_code == 200
+
+        course = Course.everything.get(uuid=self.course.uuid, draft=True)
+        assert course.title == 'Course title'
+        assert course.active_url_slug == 'learn/physics/harvardx-applied-physics'
         assert course.organization_logo_override.url is not None
         self.assertDictEqual(response.data, self.serialize_course(course))
 
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -34,7 +34,7 @@ from course_discovery.apps.course_metadata.models import (
     Seat, Source, Video
 )
 from course_discovery.apps.course_metadata.utils import (
-    create_missing_entitlement, ensure_draft_world, validate_course_number
+    create_missing_entitlement, ensure_draft_world, validate_course_number, validate_slug_format
 )
 from course_discovery.apps.publisher.utils import is_publisher_user
 
@@ -314,7 +314,6 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         """ Updates an existing course from incoming data. """
         # logging to help debug error around course url slugs incrementing
         logger.info('The raw course data coming from publisher is {}.'.format(data))  # lint-amnesty, pylint: disable=logging-format-interpolation
-
         changed = False
         # Sending draft=False means the course data is live and updates should be pushed out immediately
         draft = data.pop('draft', True)
@@ -387,7 +386,8 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         changed = changed or bool(changed_fields)
 
         if url_slug:
-            validators.validate_slug(url_slug)
+            validate_slug_format(url_slug, course)
+
             all_course_historical_slugs_excluding_present = CourseUrlSlug.objects.filter(
                 url_slug=url_slug, partner=course.partner).exclude(course__uuid=course.uuid)
             if all_course_historical_slugs_excluding_present.exists():
--- a/course_discovery/apps/course_metadata/tests/test_utils.py
+++ b/course_discovery/apps/course_metadata/tests/test_utils.py
@@ -10,6 +10,7 @@ import pytz
 import requests
 import responses
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.test import TestCase
 from edx_toggles.toggles.testutils import override_waffle_switch
 
@@ -18,6 +19,7 @@ from course_discovery.apps.api.v1.tests.test_views.mixins import OAuth2Mixin
 from course_discovery.apps.core.models import Currency
 from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata import utils
+from course_discovery.apps.course_metadata.choices import CourseRunStatus
 from course_discovery.apps.course_metadata.data_loaders.utils import map_external_org_code_to_internal_org_code
 from course_discovery.apps.course_metadata.exceptions import (
     EcommerceSiteAPIClientException, MarketingSiteAPIClientException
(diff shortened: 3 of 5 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -387,7 +386,12 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
         changed = changed or bool(changed_fields)
 
         if url_slug:
-            validators.validate_slug(url_slug)
+            if not validate_slug_format(url_slug, course):
+                # pylint: disable=line-too-long
+                raise Exception(  # pylint: disable=broad-exception-raised
+                    (f'Course edit was unsuccessful. The course URL slug ‘[{url_slug}]’ is an invalid format. '
+                      'Please ensure that the slug is in the format ‘learn/<primary_subject>/<organization_name>-<course_title>’ ')
```
DawoudSheraz: Shouldn't the error show both slug formats until the complete rollout?

course_discovery/apps/course_metadata/constants.py
```diff
@@ -30,6 +32,9 @@
 
 GOOGLE_CLIENT_API_SCOPE = ['https://www.googleapis.com/auth/drive.readonly']
 
+IN_REVIEW_STATUS = [CourseRunStatus.InternalReview, CourseRunStatus.LegalReview]
+POST_REVIEW_STATUS = [CourseRunStatus.Reviewed, CourseRunStatus.Published]
```
DawoudSheraz: Move them as CourseRun model static methods. These statuses are specific to CourseRun.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1035,3 +1037,34 @@ def is_valid_uuid(val):
...
+            if not valid_slug_flag:
+                validators.validate_slug(url_slug)
+                valid_slug_flag = True
+        else:
+            valid_slug_flag = is_valid_slug_format(url_slug) or validators.validate_slug(url_slug)
+            valid_slug_flag = True
+    else:
+        validators.validate_slug(url_slug)
+        valid_slug_flag = True
+    return valid_slug_flag
```
DawoudSheraz: There is a bit of repeated code here. It can be improved.
1. If source is edX, check status subset/intersection (loop is avoidable). If valid, give priority to learn slug checks, otherwise regular. If status is unpublished, give preference to old format.
2. If flag is not enabled, the regular checks are ok.

DawoudSheraz (commented):
Missing appropriate unit tests.

course_discovery/apps/api/v1/views/courses.py
AfaqShuaib09: `validators.validate_slug` already takes care of that. It automatically raise an exception if slug didn't match the old format.
image

course_discovery/apps/course_metadata/utils.py
AfaqShuaib09: 1. Ok I'm replacing loop with intersection, otherwise it is working as expected. I've intentionally add these lines it will be removed after all the OCM courses have been migrated.
```python
# TODO: remove this check once all OCM courses are migrated to subdirectory slug format
            if not valid_slug_flag:
                validators.validate_slug(url_slug)
                valid_slug_flag = True
```
3. If flag is not enabled, only else part of code executes (which is checking for old format)

course_discovery/apps/course_metadata/constants.py
AfaqShuaib09: static methods or variables?. I'm assuming you're referring to static variables. 
Or you want both of the statuses list to return in a static method

DawoudSheraz: There might already be methods. Either method or var will work. Or if you want to keep it constant, Add course_run prefix to the added vars.

course_discovery/apps/api/v1/views/courses.py
DawoudSheraz: But that will cause confusion because there will be 2 different messages showcasing 2 different slug formats.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -1125,3 +1127,90 @@ def test_get_existing_slug_count(self):
...
+            is_valid_format = validate_slug_format(slug, self.test_course_2)
+            assert is_valid_format is expected_result
+
+    @ddt.data(
+        ('learn/physics/applied-physics', False),
+        ('learn/', True),
+        ('learn/', False),
+        ('learn/123', True),
+        ('learn/123', False),
+        ('learn/physics/applied-physics/', True),
```
DawoudSheraz, line 1195: why does this raise validation error? The ending `/`?

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -1125,3 +1127,90 @@ def test_get_existing_slug_count(self):
...
+    """
+    Test suite for validate_slug_format method
+    """
+    def setUp(self):
+        self.product_source = SourceFactory(slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG)
+        self.test_course_1 = CourseFactory(title='test-title', product_source=self.product_source)
+        self.test_course_2 = CourseFactory(title='test-title-2')
+
+        CourseRunFactory(course=self.test_course_1, status=CourseRunStatus.Published)
+        CourseRunFactory(course=self.test_course_2, status=CourseRunStatus.InternalReview)
```
DawoudSheraz, line 1144: Have another course & course run where the status is unpublished (OCM) and verify both slug formats are working as expected.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1035,3 +1036,36 @@ def is_valid_uuid(val):
...
+
+    valid_slug_flag = False
+
+    # pylint: disable=line-too-long
+    if course.product_source.slug == settings.DEFAULT_PRODUCT_SOURCE_SLUG and IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED.is_enabled():
+        if set(course.course_run_statuses).intersection(set(CourseRun.IN_REVIEW_STATUS + CourseRun.POST_REVIEW_STATUS)):
+            valid_slug_flag = is_valid_slug_format(url_slug)
+            # TODO: remove this check once all OCM courses are migrated to subdirectory slug format
+            if not valid_slug_flag:
+                validators.validate_slug(url_slug)
```
DawoudSheraz: Can't this be written with or?

course_discovery/apps/course_metadata/tests/test_utils.py
AfaqShuaib09: yes

course_discovery/apps/course_metadata/utils.py
AfaqShuaib09: yup, I've intentionally written it separately to make it clear.
Will update it

course_discovery/apps/api/v1/views/courses.py
AfaqShuaib09: Yup, another approach would be to display a custom message by catching the exception, depending on the slug format that failed.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1035,3 +1036,35 @@ def is_valid_uuid(val):
...
+    Given a url_slug and subdirectory_slug_flag it will check if url_slug is valid or not based on
+    subdirectory_slug_flag and course_run_statuses
+
+    Args:
+        url_slug: url_slug to be validated
+        course: course object
+
+    Returns:
+        valid_slug_flag: True if url_slug is valid else False
+    """
```
zawan-ila: [think] The way this method is written, it'll never return False. It either returns True, or raises in case of invalid format. The docstring ought to be updated to reflect this. Secondly, keeping this^ in mind, would it be wise to do away with all the assignments to `valid_slug_flag` and simply return True at the end?

zawan-ila (approved):
A comment, mostly for my own enlightenment. Other than that, looks good!

AliAdnanSohail (approved):
We will further refactor this in a followup PR

course_discovery/apps/course_metadata/utils.py
AfaqShuaib09: docstring updated & method will no longer return a boolean value it will simply returns `None` if there are no errors and raise exception if any error occurs in slug format

course_discovery/apps/api/v1/views/courses.py
AfaqShuaib09: I will be doing it in the Followup PR to show custom message

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1035,3 +1036,41 @@ def is_valid_uuid(val):
...
+
+    Returns:
+        valid_slug_flag: None if url_slug is valid else raise ValidationError
+    """
+    from course_discovery.apps.course_metadata.models import CourseRun  # pylint: disable=import-outside-toplevel
+
+    valid_slug_flag = False
+
+    # pylint: disable=line-too-long
+    if course.product_source.slug == settings.DEFAULT_PRODUCT_SOURCE_SLUG and IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED.is_enabled():
```
DawoudSheraz: nit: instead of disabling the linting here, you can breakdown the line after `and`

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1035,3 +1036,41 @@ def is_valid_uuid(val):
...
+    else:
+        validators.validate_slug(url_slug)
+        valid_slug_flag = True
+
+    if not valid_slug_flag:
+        # pylint: disable=line-too-long
+        raise Exception(  # pylint: disable=broad-exception-raised
+            (f'Course edit was unsuccessful. The course URL slug ‘[{url_slug}]’ is an invalid format. '
+                'Please ensure that the slug is in the format ‘learn/<primary_subject>/<organization_name>-<course_title>’ ')
+        )
```
DawoudSheraz, line 1076: In which case will this exception be raised? Doesn't validate_slug raise the exception internally? Cause we are setting valid_slug_flag=True after validation calls.

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -1125,3 +1127,104 @@ def test_get_existing_slug_count(self):
...
+        ('learn/physics/applied-physics', None, True),
+        # will make this false once we will migrate all the OCM courses to new slug format
+        ('learn-course', None, True),
+        ('learn-course', None, False),
+        ('learn-123', None, False),
+    )
+    @ddt.unpack
+    def test_validate_slug_format__for_ocm_course(self, slug, expected_result, is_subdirectory_slug_format_active):
+        """
+        Test that validate_slug_format to check if the slug is in the correct format for OCM course
```
DawoudSheraz, line 1157: nit: OCM course with course runs past/in their review phase.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1065,6 +1068,38 @@ def test_update_success(self, method):
...
+            'key': self.course.key,
+            'type': str(self.audit_type.uuid),
+            # The API is expecting the image to be base64 encoded. We are simulating that here.
+            'image': 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY'
+                     '42YAAAAASUVORK5CYII=',
+            'organization_logo_override': 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0l'
+                                          'EQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=',
+            'video': {'src': 'https://link.to.video.for.testing/watch?t_s=5'},
+        }
+
```
DawoudSheraz: refactor: you can move this dict to constants/class level dict and override slug/relevant fields where needed.

DawoudSheraz (commented):
Some followup/nits before thumbsup.

course_discovery/apps/course_metadata/utils.py
AfaqShuaib09: Currently no case. Once, we remove or condition from below, this exception raises in case if subdirectory slug format not followed
```python
# TODO: remove the or condition once all the OCM courses are migrated to subdirectory slug format
            valid_slug_flag = is_valid_slug_format(url_slug) or validators.validate_slug(url_slug)
```

DawoudSheraz (approved):
I have not tested these out on the local, it looks good. I will give it a go on staging.

PR #4007, opened 2023-07-06
[WIP] feat: add management command to check 404s

uzairr:
PROD-3424


course_discovery/apps/course_metadata/models.py
```diff
@@ -4133,6 +4133,19 @@ class ProgramSubscriptionConfiguration(ConfigurationModel):
     )
 
 
+class SlugRedirectionDataLoaderConfiguration(ConfigurationModel):
+    """
+    Configuration to store a csv file that will be used in validate_slug_redirection.
+    """
```
DawoudSheraz, line 4142: We probably don't need this. I suspect this will be used in limited capacity and adding a model is overkill.

PR #4008, opened 2023-07-07
feat: add missing pieces for Course data modified timestamp workflow

DawoudSheraz:
### Description
This PR is a followup work from the PR https://github.com/openedx/course-discovery/pull/3976. This PR adds the necessary elements to complete course data modified timestamp update flow.

#### Seat Model
Seat model is linked with CourseRun and contains information about seat pricing and upgrade deadline override (this field is available on publisher in CourseRunForm section). Appropriate field tracker  based signal handler is added to update course data modified timestamp when seat related attrs are changed.

#### Course url slug
Course url slug and its history are maintained in CourseUrlSlug model. The slug handling also differs if the course is draft or non-draft. The draft->non-draft changes are not managed via set_official_state util. Instead, set_active_url_slug method is responsible for appropriate slug and history updates. Therefore,  in the method,  the mechanism to update course data modified timestamp has been added in the method set_active_url_slug. **That method did not have any unit tests, so I had to add them as well to verify the timestamp updates are working as expected.**

#### M2M relationships 

Following two categories of M2M relationships were not covered before:
- SortedM2M Field
- Taggable Manager

The details of the M2M relationship complexity can be seen in the internal doc https://2u-internal.atlassian.net/wiki/spaces/CAT/pages/356156060/Timestamp+Based+Data+Query. However, I am adding some concise context here.

Rest framework’s serializer [update() method use M2M field’s .set()](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L1010) method to add M2M relationship after updating the instance object The set() method in [M2MField in Django](https://github.com/django/django/blob/main/django/db/models/fields/related_descriptors.py#L895) varies the m2m setting behavior based on clear flag. However, for SortedM2M, the clear is[ always set to True](https://github.com/jazzband/django-sortedm2m/blob/e646aafa81416c7888a7d358c7293bd4f707a25f/sortedm2m/fields.py#L47-L50), to maintain the ordering of the objects in M2M relationship.  **The workaround here is to get M2M relationship pks from non-draft versions of Course/CourseRun and then compare those Pks with values being added in m2m_relationship.**

Taggable manager in django-taggit uses [TaggedItem](https://github.com/jazzband/django-taggit/blob/master/taggit/models.py#L185) to store information on which tags are assigned to which Django model (via content type). TaggedItem is the [default](https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#L440) through value if it is not defined explicitly. In Discovery course_metadata app, the value for through is not defined whenever TaggableManager is used (be it Org, Course, ProductMeta, Program, etc). Now, to receive m2m_changed signal, we need to specify the sender. The sender is i[ntermediate or through model ](https://docs.djangoproject.com/en/4.2/ref/signals/#m2m-changed)(which is TaggedItem for tag manager). So, therefore, when m2m_changed is attempted to be set only for Course.topics.through, since through=TaggedItem, whenever m2m for tags is changed on any other model, say ProductMeta, the m2m_changed handler is called because through is the same for Course & ProductMeta. **Explicit model label checks helped curate the proper handler for each taggable manager**

#### Product Value & Location Restriction
In Course api, the handling for product value & location restriction is updated to rely on calling save() instead of .filter().update() to enforce pre_save signal handler execution.

### Testing

- Make sure you have one ExecEd course in your local discovery for complete flow testinng
- Check out this branch.
- Run the latest publisher on your local. 
- Go to any course. 
- Edit its url slug. Ensure the slug is changed and the timestamp is updated.  Verify for both draft & non-draft versions
- Edit the course collaborators (both add/remove) and verify the timestamp updates
- Edit the course subjects (both add/remove) and verify the timestamp updates
- Edit product value and verify data timestamp updates
- Add/Edit location restriction and verify data timestamp updates
- In course run form, update verified upgrade deadline (only for Verified types) and verify the course timestamp is updated
- Edit the course run staff (both add/remove) and verify the timestamp updates
- For last step in testing, edit any field in Course/CourseRun form and ensure the data timestamp is updated. If any field change does not do that, flag that during review.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1465,10 +1465,7 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
             _, certification_info_changed = self.update_certificate_info(instance.additional_metadata, certificate_info)
             changed = changed or certification_info_changed
 
-        if changed:
-            # If any of the related model has updated timestamp for Course, refresh the timestamp
-            # value on the instance to represent correct state of field.
-            instance.refresh_from_db(fields=('data_modified_timestamp',))
+        return changed
         # save() will be called by main update()
 
     def update_geolocation(self, instance, geolocation):
@@ -1494,23 +1491,29 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
 
     def update_location_restriction(self, instance, location_restriction):
         if instance.location_restriction:
-            CourseLocationRestriction.objects.filter(id=instance.location_restriction.id).update(**location_restriction)
+            _, location_restriction_changed = update_instance(instance.location_restriction, location_restriction, True)
+            return location_restriction_changed
         else:
             instance.location_restriction = CourseLocationRestriction.objects.create(**location_restriction)
+            return True
 
     def update_in_year_value(self, instance, in_year_value):
         if instance.in_year_value:
-            ProductValue.objects.filter(id=instance.in_year_value.id).update(**in_year_value)
+            _, in_year_value_changed = update_instance(instance.in_year_value, in_year_value, True)
+            return in_year_value_changed
         else:
             instance.in_year_value = ProductValue.objects.create(**in_year_value)
+            return True
 
     def update(self, instance, validated_data):
+        changed = False
         # Handle writing nested fields separately
         if 'additional_metadata' in validated_data:
             # Handle additional metadata only for external courses else just pop
             additional_metadata_data = validated_data.pop('additional_metadata')
             if instance.is_external_course:
-                self.update_additional_metadata(instance, additional_metadata_data)
+                additional_metadata_changed = self.update_additional_metadata(instance, additional_metadata_data)
+                changed = changed or additional_metadata_changed
         if 'geolocation' in validated_data:
             geolocation = validated_data.pop('geolocation')
             if all(bool(value) is True for value in geolocation.values()):
@@ -1518,9 +1521,16 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
         if 'location_restriction' in validated_data:
             location_restriction_data = validated_data.pop('location_restriction')
             if not all(bool(value) is False for value in location_restriction_data.values()):
-                self.update_location_restriction(instance, location_restriction_data)
+                location_restriction_changed = self.update_location_restriction(instance, location_restriction_data)
+                changed = changed or location_restriction_changed
         if 'in_year_value' in validated_data:
-            self.update_in_year_value(instance, validated_data.pop('in_year_value'))
+            in_year_value_changed = self.update_in_year_value(instance, validated_data.pop('in_year_value'))
+            changed = changed or in_year_value_changed
+
+        if changed:
+            # If any of the related model has updated timestamp for Course, refresh the timestamp
+            # value on the instance to represent correct state of field.
+            instance.refresh_from_db(fields=('data_modified_timestamp',))
         return super().update(instance, validated_data)
 
 
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -2,7 +2,6 @@ import logging
 import time
 from datetime import datetime, timezone
 
-import pytz
 import waffle  # lint-amnesty, pylint: disable=invalid-django-waffle-import
 from celery import shared_task
 from django.apps import apps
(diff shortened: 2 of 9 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1494,33 +1491,46 @@ def update_geolocation(self, instance, geolocation):
...
+        changed = False
         # Handle writing nested fields separately
         if 'additional_metadata' in validated_data:
             # Handle additional metadata only for external courses else just pop
             additional_metadata_data = validated_data.pop('additional_metadata')
             if instance.is_external_course:
-                self.update_additional_metadata(instance, additional_metadata_data)
+                additional_metadata_changed = self.update_additional_metadata(instance, additional_metadata_data)
+                changed = changed or additional_metadata_changed
         if 'geolocation' in validated_data:
```
zawan-ila, line 1517: Are we handling geoloc elsewhere?

zawan-ila, line 1517: Are we handling geoloc elsewhere?

DawoudSheraz: For geolocation, the code does not create new geoloc object for each course. The code gets the geolocation object based on provided coords and assigns FK to course. FieldTracker treats that as FK change on Course object. 
For Product Value and some other models, the object linked with Course is updated, which then causes appropriate signal to trigger.

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/signals.py
```diff
@@ -453,9 +504,9 @@ def connect_course_data_modified_timestamp_related_models():
         CourseLocationRestriction,
         CourseEntitlement,
         GeoLocation,
-        Organization,
```
Ali-D-Akbar, line 456: Why is this being removed?

Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: There is no timestamp handling against Organization. It might have been added mistakenly.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -385,34 +384,70 @@ def update_course_data_from_event(**kwargs):
...
+    To determine if the course collaborators have been changed, get the collab pk from non-draft/official version,
+    compare that set with values coming in from m2m_relation's pre_add or pre_remove, and if the set is not the same,
+    then the collaborators have been changed.
+    https://github.com/jazzband/django-sortedm2m/blob/e646aafa81416c7888a7d358c7293bd4f707a25f/sortedm2m/fields.py#L47-L50
     """
-    if action in ['pre_add', 'pre_remove'] and instance.draft:
-        logger.info(f"{sender} has been updated for course {instance.key}.")
-        Course.everything.filter(key=instance.key).update(
-            data_modified_timestamp=datetime.now(pytz.UTC)
+    if action in ['pre_add', 'pre_remove'] and instance.draft and instance.official_version and not kwargs['reverse']:
```
zawan-ila, line 399: Do we really need a pre_remove here? Will it ever fire?

course_discovery/apps/course_metadata/signals.py
```diff
@@ -385,34 +384,70 @@ def update_course_data_from_event(**kwargs):
...
         instance.update_product_data_modified_timestamp(bypass_has_changed=True)
 
 
+@receiver(m2m_changed, sender=Course.topics.through)
+def course_topics_taggable_changed(sender, instance, action, **kwargs):
+    """
+    Signal handler to handle Taggable manager changes for the course tag field or course related models' tag fields.
+    """
+    if action in ['pre_add', 'pre_remove'] and not kwargs['reverse'] \
+            and kwargs['pk_set'] and instance._meta.label == 'course_metadata.Course' and instance.draft:
```
zawan-ila, line 446: Do we really need `kwargs['pk_set']` here? Can the signal fire with it being empty?

zawan-ila (approved):

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: Yes, if a course team wants to remove a collaborator, then yes.

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: Yes. Taggit does not check if the set is true when sending the signal (https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#L171).  When updating the course via API, the handler will be called because of how DRF handles M2M in create/update (https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L1045).

course_discovery/apps/course_metadata/signals.py
zawan-ila: Hmm. Wont a clear first be fired and then an add? Since this is a SortedM2M?

DawoudSheraz: Hmm, valid point. I will make a note of this and remove this in a followup PR if needed.

PR #4015, opened 2023-07-12
feat: allow uuid in Contentful to be null

Ali-D-Akbar:
[PROD-3476](https://2u-internal.atlassian.net/browse/PROD-3476)

Fixing errors while Algolia is fetching Contentful. A degree entry can have a uuid-list and the uuid field can be empty. In that  case, we should take all the uuids from uuid-list.

### DIFF
--- a/course_discovery/apps/course_metadata/contentful_utils.py
+++ b/course_discovery/apps/course_metadata/contentful_utils.py
@@ -401,10 +401,13 @@ def fetch_and_transform_degree_contentful_data():
     transformed_degree_data = {}
 
     for degree_entry in contentful_degree_page_entries:
-        product_uuid = degree_entry.uuid
+        product_uuid = getattr(degree_entry, 'uuid', None)
         if product_uuid in transformed_degree_data:
             continue
         uuid_list = getattr(degree_entry, 'uuid_list', [])
+        if product_uuid is None:
+            # add all the transformed data against the first uuid in uuid_list.
+            product_uuid = uuid_list[0]
         excluded_from_search = getattr(degree_entry, 'excluded_from_search', False)
         excluded_from_seo = getattr(degree_entry, 'excluded_from_seo', False)
         page_title = degree_entry.seo.page_title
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -401,10 +401,12 @@ def fetch_and_transform_degree_contentful_data():
     transformed_degree_data = {}
 
     for degree_entry in contentful_degree_page_entries:
-        product_uuid = degree_entry.uuid
+        product_uuid = getattr(degree_entry, 'uuid', None)
         if product_uuid in transformed_degree_data:
             continue
         uuid_list = getattr(degree_entry, 'uuid_list', [])
+        if product_uuid is None:
+            product_uuid = uuid_list[0]
```
DawoudSheraz: Why is only the first entry assigned? Shouldn't all UUIDs have the contentful record added in Algolia?

Ali-D-Akbar: I'm merely adding all the transformed data in the first one then adding it in all the uuid-list. see https://github.com/openedx/course-discovery/pull/4015/files#diff-1e525eb080b8f41c59a26f5a6975ff5d180a04d6cbf0aad2da718cb81136b06eR444

DawoudSheraz: Ok, I guess you can mention this explicitly via comment. It will be confusing to read why is this done.

uzairr (approved):

course_discovery/apps/course_metadata/contentful_utils.py
```diff
@@ -401,10 +401,13 @@ def fetch_and_transform_degree_contentful_data():
     transformed_degree_data = {}
 
     for degree_entry in contentful_degree_page_entries:
-        product_uuid = degree_entry.uuid
+        product_uuid = getattr(degree_entry, 'uuid', None)
         if product_uuid in transformed_degree_data:
             continue
         uuid_list = getattr(degree_entry, 'uuid_list', [])
+        if product_uuid is None:
+            # add all the transformed data against the first uuid in uuid_list.
```
DawoudSheraz, line 409: nit: Also explain why is this addition necessary.

DawoudSheraz (approved):

PR #4024, opened 2023-07-17
feat: add admin action to toggle display_on_org_page for degrees

zawan-ila:
## Testing Instructions

1. Checkout this branch
2. Go to Admin
3. Trigger the action `Display on org page` for some degrees and verify that `display_on_org_page` is set to True
4. Trigger the action `Hide on org page` for some degrees and verify that `display_on_org_page` is set to False

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -798,46 +798,13 @@ class SpecializationAdmin(admin.ModelAdmin):
     list_display = ('value', )
 
 
-def change_degree_status(modeladmin, request, queryset, status):
-    """
-    Changes the status of a degree.
-    """
-    count = queryset.count()
-    if count:
-        for obj in queryset:
-            obj.status = status
-            obj.save()
-
-        modeladmin.message_user(request, _("Successfully %(status)s %(count)d %(items)s.") % {
-            "status": "published" if status == ProgramStatus.Active else "unpublished",
-            "count": count,
-            "items": model_ngettext(modeladmin.opts, count),
-        }, messages.SUCCESS)
-
-
-@admin.action(permissions=['change'], description='Publish selected Degrees')
-def publish_degrees(modeladmin, request, queryset):
-    """
-    Django admin action to bulk publish degrees.
-    """
-    change_degree_status(modeladmin, request, queryset, ProgramStatus.Active)
-
-
-@admin.action(permissions=['change'], description='Unpublish selected Degrees')
-def unpublish_degrees(modeladmin, request, queryset):
-    """
-    Django admin action to bulk unpublish degrees.
-    """
-    change_degree_status(modeladmin, request, queryset, ProgramStatus.Unpublished)
-
-
 @admin.register(Degree)
 class DegreeAdmin(admin.ModelAdmin):
     """
     This is an inheritance model from Program
 
     """
-    list_display = ('uuid', 'title', 'marketing_slug', 'status', 'hidden')
+    list_display = ('uuid', 'title', 'marketing_slug', 'status', 'hidden', 'display_on_org_page')
     ordering = ('title', 'status')
     readonly_fields = ('uuid', )
     list_filter = ('partner', 'status',)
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -490,6 +492,23 @@ def test_publish_degree_actions(self, before_status, after_status, admin_action,
...
+        self.degree.save()
+        response = self.client.post(
+            reverse('admin:course_metadata_degree_changelist'),
+            {'action': admin_action, '_selected_action': [self.degree.id, ]},
+            follow=True
+        )
+        assert response.status_code == HTTP_200_OK
+        assert success_message in response.content
+        updated_degree = Degree.objects.get(id=self.degree.id)
+        assert updated_degree.display_on_org_page == after_value
```
AfaqShuaib09: Can we add testcase for multiple degree updates against these actions?

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/admin.py
```diff
@@ -860,7 +861,35 @@ class DegreeAdmin(admin.ModelAdmin):
...
+    def display_degrees_on_org_page(self, request, queryset):
+        updated = queryset.update(display_on_org_page=True)
+        self.message_user(
+            request,
+            ngettext(
+                "%d degree was successfully set to display on org page.",
+                "%d degrees were successfully set to display on org page.",
+                updated,
+            )
+            % updated,
```
DawoudSheraz: Can't we use f-string here?

course_discovery/apps/course_metadata/tests/test_admin.py
DawoudSheraz: +1

course_discovery/apps/course_metadata/tests/test_admin.py
```diff
@@ -490,6 +492,23 @@ def test_publish_degree_actions(self, before_status, after_status, admin_action,
         updated_degree = Degree.objects.get(id=self.degree.id)
         assert updated_degree.status == after_status
 
+    @ddt.data(
+        (False, True, 'display_degrees_on_org_page', b'1 degree was successfully set to display on org page.'),
+        (True, False, 'hide_degrees_on_org_page', b'1 degree was successfully set to be hidden on org page.')
```
DawoudSheraz: Why is string conversion to bytes necessary?

course_discovery/apps/course_metadata/tests/test_admin.py
zawan-ila: Done

course_discovery/apps/course_metadata/admin.py
zawan-ila: Done

course_discovery/apps/course_metadata/tests/test_admin.py
zawan-ila: It was necessary since we were checking for its presence in a bytes string. However, I've updated that byte string to utf-8 and comparing with that now.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -860,7 +860,25 @@ class DegreeAdmin(admin.ModelAdmin):
         'micromasters_org_name_override', 'faq', 'costs_fine_print', 'deadlines_fine_print', 'specializations',
         'program_duration_override', 'display_on_org_page',
     )
-    actions = [publish_degrees, unpublish_degrees]
+    actions = [publish_degrees, unpublish_degrees, "display_degrees_on_org_page", "hide_degrees_on_org_page"]
```
DawoudSheraz: nit: remove "" from actions to keep the code consistent

DawoudSheraz (approved):

course_discovery/apps/course_metadata/admin.py
zawan-ila: Don't think it'd work like that. The issue is that the earlier actions are "functions" defined outside the admin class and the newly added actions are "methods" of the admin class. https://docs.djangoproject.com/en/4.2/ref/contrib/admin/actions/#advanced-action-techniques.

I think we should move the earlier actions inside the admin class too?

DawoudSheraz: ok

PR #4025, opened 2023-07-19
feat: Support multiple verified seats for a single course run

jawad-khan:
## Problem Statement
Mobile team has started adding additional two verified seats(One for Android and second for iOS) on ecommerce for a single run of a course. When these changes are pushed into discovery through publish api they are producing [errors](https://splunk.edx.org/en-US/app/search/search?q=search%20index%3D%22prod-edx%22%20course_discovery.apps.course_metadata.models.Seat.MultipleObjectsReturned%3A%20get()%20returned%20more%20than%20one%20Seat%20--%20it%20returned%202!&display.page.search.mode=smart&dispatch.sample_ratio=1&earliest=1684627200&latest=1684886400&sid=1686037847.1378796) and creating problems while publishing the course.

## Solution
This PR is adjusting places where discovery is expecting only a single verified seat for a course run. 
We are planning to add third variant either through ecommerce api or using discovery django admin.

## Testing instructions

Go to http://localhost:18381/admin/course_metadata/seat/ and add a third seat for any course(make sure to add draft version too). Now with master branch if you try to change price of the course in publisher it will give you error and you will not be able to publish it. Move to this branch and it course will be published successfully.

[Learner-9425](https://2u-internal.atlassian.net/browse/LEARNER-9425)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -601,15 +601,19 @@ def update_seat(self, course_run, product_body):
...
         )
-
-        if created:
+        if not seats:
+            course_run.seats.create(type=seat_type, credit_provider=credit_provider, currency=currency, **defaults)
             logger.info('Created seat for course with key [%s] and sku [%s].', course_run.key, sku)
+        else:
+            for seat in seats:
+                for key, value in defaults.items():
+                    setattr(seat, key, value)
```
Ali-D-Akbar: Why can't we do this with the update method in seat?

Ali-D-Akbar (approved):
Just an improvement otherwise, looks good.

DawoudSheraz:
General questions:
1. Why can't the existing seat be used for this?
2. How are new seats dedicated for mobile distinguished from regular seats? The uniqueness is determined using ('course_run', 'type', 'currency', 'credit_provider', 'draft'). You cant have two seats of the same type at the same price. 
3. What are its data reporting implications? How will ecommerce be affected by this?

course_discovery/apps/course_metadata/data_loaders/api.py
jawad-khan: I tried update method earlier but it doesn't fire post or pre save signals which we need [here](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/signals.py#L509).

jawad-khan:
Answers to general questions:

1: Mobile skus will have different prices and sku numbers which forced us to add another seat.
2. Price can be same in many cases but skus will surely be different in every case, that is how we were differentiating them in ecommerce.
3. Are you talking about sales data(oredrs, refund, payments)? If yes, we are already in touch with BI and DE teams for this and they have started ingesting this data in their reports.

DawoudSheraz:
> Answers to general questions:
> 
> 1: Mobile skus will have different prices and sku numbers which forced us to add another seat. 2. Price can be same in many cases but skus will surely be different in every case, that is how we were differentiating them in ecommerce. 3. Are you talking about sales data(oredrs, refund, payments)? If yes, we are already in touch with BI and DE teams for this and they have started ingesting this data in their reports.

1. How/why can a same course have a different price? If so, where will this be configured? I would say to add an ADR for this because this is a significant change and warrants an ADR.
2. But how will the system know which sku belongs to which type?

jawad-khan:
> > Answers to general questions:
> > 1: Mobile skus will have different prices and sku numbers which forced us to add another seat. 2. Price can be same in many cases but skus will surely be different in every case, that is how we were differentiating them in ecommerce. 3. Are you talking about sales data(oredrs, refund, payments)? If yes, we are already in touch with BI and DE teams for this and they have started ingesting this data in their reports.
> 
>     1. How/why can a same course have a different price? If so, where will this be configured? I would say to add an ADR for this because this is a significant change and warrants an ADR.
> 
>     2. But how will the system know which sku belongs to which type?

1. Sometimes Android or iOS  force us to quote 100 as 99 or doesn't allow price greater than a threshold, so we have to adjust for those cases. Currently we have a sheet maintaining all these prices and we sync it with BI team, Here is a link to all the [documentation](https://2u-internal.atlassian.net/wiki/spaces/LEARNER/pages/7156620/2021+Mobile+Payments) we did. Let me know if you need more context we can have a call on this.

2. We add Android or iOS keyword in mobile seat title and sku also contains mobile keyword. Also when making a payment we process it through mobile processors which eventually get recorded in data.

jawad-khan:
Is PR ok to be merged?

jawad-khan:
@DawoudSheraz @Ali-D-Akbar Please merge this PR as I don't have write access.

DawoudSheraz:
1. Please have an ADR that explains this new addition and its technical implications. The internal docs are not visible on public repositories. This is important because this change is significant and will have implications across the eco-system. We might need some product input on this
2. Can you give an example of how the new Seat will look? You mentioned that sku will have android or ios keyword but the code you added does not seem to do so.
3. Is this meant for Verified track only? The additions do not seem to care about that.
4. Add some testing instructions so that this be tested & verified locally.

course_discovery/apps/course_metadata/data_loaders/api.py
DawoudSheraz: There is a util method in discovery codebase, update_instance. That does the same thing you are doing in this for-loop.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -742,20 +742,22 @@ def test_enrollable_seats(self):
         professional_seat_type = factories.SeatTypeFactory.professional()
         honor_seat_type = factories.SeatTypeFactory.honor()
         verified_seat = factories.SeatFactory(course_run=course_run, type=verified_seat_type, upgrade_deadline=None)
+        second_verified_seat = factories.SeatFactory(course_run=course_run, type=verified_seat_type,
+                                                     upgrade_deadline=None)
```
DawoudSheraz: Ideally, you should verify with same and different price because it will result in uniquess error. The factories will use random value if not provided explicitly.

jawad-khan:
> 1. Please have an ADR that explains this new addition and its technical implications. The internal docs are not visible on public repositories. This is important because this change is significant and will have implications across the eco-system. We might need some product input on this
> 
>     2. Can you give an example of how the new Seat will look? You mentioned that sku will have android or ios keyword but the code you added does not seem to do so.
> 
>     3. Is this meant for Verified track only? The additions do not seem to care about that.
> 
>     4. Add some testing instructions so that this be tested & verified locally.

1. I'll add an ADR for this, any particular point you want me to add in ADR?
2. We'll have Android or ios keyword in title of seat, as you can see it on [production](https://ecommerce.edx.org/dashboard/catalogue/products/100131/#product_stock).
3. Yes it is meant for verified tracks only. Which additions are you talking about?
4. Sure.

course_discovery/apps/course_metadata/tests/test_models.py
jawad-khan: Okay.

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
```diff
@@ -0,0 +1,30 @@
...
+Discovery will support multiple verified seats for a single course run.
+All the places will be identified and fixed where we are expecting only a single verified seat for a course run.
+We'll keep monitoring discovery and will fix any future issue regarding multiple verified skus.
+
+Consequences
+------------
+
+Ecommerce can sync data having multiple verified skus without any error.
+
+Course run can be changed without any issue on publisher.
```
DawoudSheraz: In ADR, it would be better to avoid using first person pronouns and instead use third-person. The ADR being technical documents should focus on the context (✅ ), the decision taken (can be elaborated more), the consequence of the decision (needs more elaboration), and any alternatives considered (if any) -- among other things. 
ADR is missing the impact on discovery/publisher while allowing multiple seats, how the price will change when updated from publisher, how will discovery identify various seat categories, etc.

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Context
+-------
+
+While implementing in app payments mobile team decided to use current ecommerce implementation as its backend. This way all the analytics, sales and refunds record will stay at a single place.
+Mobile platforms e.g. play store and app store apply some restrictions on product price and sku names. Therefore a separate verified seat was needed for mobile.
+The seat was added on ecommerce side and it worked perfectly with django oscars on ecommerce. But when discovery consumed this data, course publishers started having problems on publishing a course run.
+During spike it turned out discovery was expecting only a single verified seat for a course run and code was written according to this logic.
+Discovery repo code needs to be adjusted to handle multiple verified seats for a single course run.
+
```
DawoudSheraz, line 17: While the context here makes sense, I will suggest not referencing mobile team, spike, etc. The context can emphasize on mobile app payments were added, how it caused issues with discovery, and what investigation results showed about discovery not being able to handle multiple seats.

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
```diff
@@ -0,0 +1,38 @@
...
+Decision
+--------
+
+To support in app payments Discovery will support multiple verified seats for a single course run. In addition all services where discovery publishes course run changes(e.g. ecommerce, lms) are also required to expect multiple verified seats.
+All the places will be identified and fixed where Discovery is expecting only a single verified seat for a course run.
```
DawoudSheraz, line 22: instead of fixed, it should be updated. It is new feature-ish work.

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
```diff
@@ -0,0 +1,38 @@
...
+Every mobile sku will have an identifier is sku name(e.g. android, ios). That is how discovery can differentiate between mobile and web skus of verified seat category.
+Discovery service will be monitored and any future issue regarding multiple verified skus will be fixed.
+Mobile team will do a discovery on how to push a mobile sku in discovery(whether to give an interface on publisher or push it through ecommerce api).
```
DawoudSheraz, line 25: This is internal only details, in my opinion. Instead, focus on the large term consequences of this (different prices for same product, etc.).

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Consequences
+------------
+
+Discovery will have multiple verified skus having same or different price for a single course run. Discovery can differentiate between mobile and web skus of verified seat category by mobile keywords(android, ios) in sku value or title.
+
+When there is a price change on publisher, course run will be updated and discovery pushes updated data to ecommerce, Ecommerce will detect this change and will update skus price on relevant mobile platforms(android, ios) through apis.
+
+Ecommerce can sync data having multiple verified skus without any error.
```
DawoudSheraz, line 34: Ah, this will cause unnecessary operations. Lets say seats are created in ecomm. They get synced in discovery. Someone changes price from Publisher for that course and all seats have their prices updated (including mobile). Once data is pushed to ecommerce, the price will be reverted back manually(?) for mobile skus. If price changes in publisher again, there will be a new cycle of changes. This slight operation can easily cause data discrepancies as this is something that has been observed in past when it comes to Dates. 
What if we can block updates to mobile skus when publisher price changes? Or is that not desirable?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -601,15 +602,17 @@ def update_seat(self, course_run, product_body):
...
+        seats = course_run.seats.filter(
             credit_provider=credit_provider,
             currency=currency,
-            defaults=defaults
+            type=seat_type,
         )
-
-        if created:
+        if not seats:
+            course_run.seats.create(type=seat_type, credit_provider=credit_provider, currency=currency, **defaults)
```
DawoudSheraz, line 611: If multiple seats are coming from ecomm, won't this create only one seat?

course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -601,15 +602,17 @@ def update_seat(self, course_run, product_body):
...
             logger.info('Created seat for course with key [%s] and sku [%s].', course_run.key, sku)
+        else:
+            for seat in seats:
+                update_instance(seat, defaults, should_commit=True)
```
DawoudSheraz, line 615: The method is update_seat and is supposed to handle one seat for a product. This makes it different. Isn't it possible to do some of this processing in update_Seats method https://github.com/openedx/course-discovery/blob/8c31bd03bf715c9ce4742c474f747a2e233cde6e/course_discovery/apps/course_metadata/data_loaders/api.py#L533 and leave update_seat as it is?

DawoudSheraz:
Ok, I got a chance to test it out locally. Here are my observations:

- To mimic how things are happening, I created the seat manually in ecommerce. However, when I ran RCM ecommerce loader on discovery shell, it did not create the seat in Discovery. I could see seat in ecommerce response but ecommerce loader did not handle it apparently(?). I suspect it has to do with update_seat vs update_seats method that I mentioned above.
image
image

- When I created seats directly from admin (with custom SKU), they were appearing in Discovery API response. However, when I updated the price from publisher, it overrode the price that I added from admin (as expected). The seat was pushed to ecommerce. But since ecommerce did not have that seat before, it generated an SKU on its side and synced that SKU with discovery. So, once the price save operation was completed, the new seat that I had created had its SKU changed. 

There is a need to revisit how things are being handled currently and do appropriate changes to ensure data integrity.

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
jawad-khan: Currently we have already pushed code on ecommerce to alert mobile team for any change in course run through mail. In future we will update any price change through apis on relevant platforms and alert mobile team for a review on that change. 
Mobile team will make a decision on that alert to keep the price change or not depending on the course and price value.

course_discovery/apps/course_metadata/data_loaders/api.py
jawad-khan: Every mobile seat will have a different sku, Also we are doing a spike on how to add skus on discovery too. Maybe through an interface or django admin.

course_discovery/apps/course_metadata/data_loaders/api.py
jawad-khan: sure.

jawad-khan:
> * errode the price that I added from admin (as expected). The seat was pushed to ecommerce. But since ecommerce did not have that seat before, it generated an SKU on its side and synced that SKU with discovery. So, once the pri

On your first comment, this ticket is fixing this issue of update_seat. 
On your second comment, we have already identified places where we need to make changes and will be working on those changes. Not only sku, ecommerce also override title of the seat. The code on ecommerce overrides any sku or title coming from any service. We will be adjusting that code to consider mobile skus a special case and assign them a different title. And on sku we are already defining skus on ecommerce therefore we weren't having that issue.
Maybe we can have a sync on this where I can walk you through that code and will discuss our solution with you.

docs/decisions/0025-support-multiple-verified-seats-for-a-course-run.rst
DawoudSheraz: Ah, this is not a scalable solution. Plus, as I mentioned above, it creates a back & forth in the pricing. The addition of different seats of same type (and in same price will cause also confusion / breakages on downward consumers of the discovery.

course_discovery/apps/course_metadata/data_loaders/api.py
DawoudSheraz: But you are creating based on the type. sku is in defaults. The fields added in defaults dict are not used as a criteria of filtering.

PR #4034, opened 2023-07-26
docs: ADR for bringing programs in Publisher

Ali-D-Akbar:
[PROD-3495](https://2u-internal.atlassian.net/browse/PROD-3495)

ADR for adding programs on Publisher for SEO teams. Teams will be able to see programs and edit program slugs on Publisher.

### DIFF
--- a/docs/decisions/0025-program-slugs-in-publisher.rst
+++ b/docs/decisions/0025-program-slugs-in-publisher.rst
@@ -1,0 +1,31 @@
+25. Program Slugs in Publisher
+-------------------------------
+
+Status
+------
+
+Accepted (July 2023)
+
+Context
+-------
+
+SEO teams want to be able change the program slugs. These slugs are used in generating url for each type of programs.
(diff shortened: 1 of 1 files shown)


DawoudSheraz:
shouldn't this ADR be in publisher?

Ali-D-Akbar:
> shouldn't this ADR be in publisher?

I intend to add this in both disco and publisher once the questions are done bcz we're adding functionalities in disco apis for that as well.

DawoudSheraz:
> > shouldn't this ADR be in publisher?
> 
> I intend to add this in both disco and publisher once the questions are done bcz we're adding functionalities in disco apis for that as well.

The publisher ADR should be enough, as this change is more driven towards publisher core and does not impact much on Discovery side.

Ali-D-Akbar:
Closing PR and opening new one in Publisher: https://github.com/openedx/frontend-app-publisher/pull/895/files

PR #4054, opened 2023-08-08
feat: migrate DjangoChoices to native django choices (models.TextChoices)

AfaqShuaib09:
[PROD-3521](https://2u-internal.atlassian.net/browse/PROD-3521)
----------------

This PR migrates dj-choices `Django Choices` to native django choices `models.TextChoices`. 

Django docs regarding native django choices:
  - https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types

Django-choices contains documentation for the migration [Migrating to native Django Choices — Django-Choices 2.0.0 documentation](https://django-choices.readthedocs.io/en/latest/migrating.html) .

Testing:
- Check from discovery django admin from each of choices field working correctly.
- Go to publisher Course Edit Page and try to change the value of each choice field.
- Change the choices values through `api/v1/` endpoints


course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -347,7 +347,7 @@ def update(self, request, **kwargs):
 
         # Handle staff update on course run in review with valid status transition
         if (request.user.is_staff and course_run.in_review and 'status' in request.data and
-                request.data['status'] in CourseRunStatus.INTERNAL_STATUS_TRANSITIONS):
+                request.data['status'] in CourseRunStatus.REVIEW_STATES()):
```
DawoudSheraz: Internal status transition had internal & reviewed. Review states has legal & internal review state.

course_discovery/apps/course_metadata/models.py
```diff
@@ -883,7 +883,7 @@ class AdditionalMetadata(TimeStampedModel):
     )
     product_status = models.CharField(
         default=ExternalProductStatus.Published, max_length=50, null=False, blank=False,
-        choices=ExternalProductStatus.choices, validators=[ExternalProductStatus.validator]
+        choices=ExternalProductStatus.choices
```
DawoudSheraz, line 886: Is there no need for validators with django choices?

AfaqShuaib09: Yes, I've tried to pass invalid value through the enpoint and it is giving me invalid choice error.

course_discovery/apps/api/v1/views/course_runs.py
AfaqShuaib09: Ah.. I missed that, I've added it again as `class method` because we can't keep tuple in the models.TextChoices class.

DawoudSheraz (approved):
I have not tested it on local but will give it a shot on stage. CI is green, so 👍🏽

course_discovery/apps/course_metadata/choices.py
```diff
@@ -1,82 +1,81 @@
...
+    Reviewed = 'reviewed', _('Reviewed')
+    Published = 'published', _('Published')
 
-    INTERNAL_STATUS_TRANSITIONS = (
-        InternalReview.value,
-        Reviewed.value,
-    )
+    @classmethod
+    def INTERNAL_STATUS_TRANSITIONS(cls):
+        return (cls.InternalReview.value, cls.Reviewed.value)
```
Ali-D-Akbar, line 14: Why are we not returning a list here?

Ali-D-Akbar (approved):
You can add what this PR generally adds (using `.value` with choice items) and removes (removing djchoices references and not using the validators) in the description so adding this in another repo may be helpful. We may have a lot of repos in edX that may be using djchoices package so your work can be a standard PR to updated such changes.

course_discovery/apps/course_metadata/choices.py
AfaqShuaib09: because it is previously a tuple, therefore I added it here as a tuple
https://github.com/openedx/course-discovery/pull/4054/files#diff-0847e48b70fe28d50327a9b35def1b63d59655518d84bc54536434901ebb87d8L12-L15

PR #4055, opened 2023-08-09
perf: improving database queries for CourseRecommendations endpoint

zawan-ila:
[PROD-3561](https://2u-internal.atlassian.net/browse/PROD-3561)
------------------------

This PR fixes most of the N+1 issues on CourseRecommendations endpoint. 

### Screenshots
These are particular changes from an endpoint on my machine. The effects on prod will probably be significantly different, and most likely not this significant, due to the type and size of the data over there.
#### Before
Screenshot 2023-08-10 at 12 21 26 PM

#### After
Screenshot 2023-08-10 at 12 23 31 PM

### Testing
- Go to /api/v1/course_recommendations/<course_key> for a particular course on local
- Note the query count and Time in Django Debug Toolbar
- Pull these changes
- Go to the the same course_recommendations endpoint
- Note the query count and Time again. They should both decrease.

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -525,4 +525,4 @@ class CourseRecommendationViewSet(RetrieveModelMixin, viewsets.GenericViewSet):
     lookup_value_regex = COURSE_ID_REGEX
     permission_classes = (IsAuthenticated, IsCourseEditorOrReadOnly,)
     serializer_class = serializers.CourseWithRecommendationsSerializer
-    queryset = serializers.MinimalCourseSerializer.prefetch_queryset()
+    queryset = Course.objects.all()
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -2410,6 +2410,21 @@ class CourseViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin
 
     @responses.activate
     def test_recommendations(self):
-        url = reverse('api:v1:course_recommendations-detail', kwargs={'key': self.course.key})
-        response = self.client.get(url)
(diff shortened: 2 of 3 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -525,4 +525,4 @@ class CourseRecommendationViewSet(RetrieveModelMixin, viewsets.GenericViewSet):
     lookup_value_regex = COURSE_ID_REGEX
     permission_classes = (IsAuthenticated, IsCourseEditorOrReadOnly,)
     serializer_class = serializers.CourseWithRecommendationsSerializer
-    queryset = serializers.MinimalCourseSerializer.prefetch_queryset()
+    queryset = Course.objects.all()
```
DawoudSheraz, line 528: This might cause N+1 based on serializer. You can use https://github.com/jmcarp/nplusone on your local to identify this.

zawan-ila: I have checked with debug toolbar. It does not increase the query count, as the prefetched_information is not being used in the response.

Ali-D-Akbar (commented):
Can you add steps to test how this improves the performance?

course_discovery/apps/api/v1/views/courses.py
DawoudSheraz: Ok. Follow-up question: Do the SQL queries after fixing N+1 use DB indices? Or is there no DB index present yet?

DawoudSheraz (approved):
Before merge, if you can add a unit test on SQL query count (+- some threshold), that would be great. It will give devs heads-up that the change they are making will increase DB round-trip.

course_discovery/apps/api/v1/views/courses.py
zawan-ila: Yes. The queries are using indexes.

Ali-D-Akbar (approved):

PR #4060, opened 2023-08-17
temp: add and handle waffle switch to enable sub directory url format for exec ed courses

AliAdnanSohail:
## Description
This PR will handle add and handle waffle switch to enable sub directory url format for exec ed courses


DawoudSheraz (approved):
Commit type should be temp as this flag will be removed.

AfaqShuaib09 (approved):

PR #4063, opened 2023-08-21
feat: include all PCs in review emails

DawoudSheraz:
### Description
- Send emails to all PCs when PCs are notified via email
- Replace contact us email of PCs with contact_us_emails for all PCs.
- Update PC email template to remove the name and add default "Project Coordinator team"

### Testing
In send_email method in emails.py, add `    open(f"{str(datetime.datetime.now())}.html", 'w').write(html_content)
` before email_msg.send() call. On local, the emails are not sent. However, this will render the email content in an HTML file. 

- Ensure you have two or more PCs for an organization in http://localhost:18381/admin/publisher/organizationuserrole/
- Create a new course in publisher and send it to legal review after adding the data. Notice the test html. Ensure the content includes email for both PCs
- Move to PC review and ensure the PC email does not include pc name
- Move to scheduled/published and verify the pc emails are present in the content

### Screenshots
image

image

image

image

### DIFF
--- a/course_discovery/apps/course_metadata/emails.py
+++ b/course_discovery/apps/course_metadata/emails.py
@@ -44,7 +44,7 @@ def log_missing_project_coordinator(key, org, template_name):
 
 
 def get_project_coordinator(org):
-    """ Get the registered project coordinator for an organization.
+    """ Get the registered project coordinators for an organization.
 
         Only returns the first one. Technically the database supports multiple. But in practice, we only use one.
         Requires a OrganizationUserRole to be set up first.
@@ -53,7 +53,7 @@ def get_project_coordinator(org):
             org (Object): Organization object
 
         Returns:
-            Object: a User object or None if no project coordinator is registered
+            list(Object): a list of User objects or None if no project coordinator is registered
     """
     # Model imports here to avoid a circular import
     from course_discovery.apps.publisher.models import OrganizationUserRole  # pylint: disable=import-outside-toplevel
@@ -61,13 +61,20 @@ def get_project_coordinator(org):
     if not org:
         return None
 
-    role = OrganizationUserRole.objects.filter(organization=org,
-                                               role=InternalUserRole.ProjectCoordinator.value).first()
-    return role.user if role else None
+    are_project_coordinators_present = OrganizationUserRole.objects.filter(
+        organization=org, role=InternalUserRole.ProjectCoordinator.value
+    ).exists()
+
+    if not are_project_coordinators_present:
+        return None
+    project_coordinators = [pc.user for pc in OrganizationUserRole.objects.filter(
+        organization=org, role=InternalUserRole.ProjectCoordinator.value
+    ).select_related('user')]
+    return project_coordinators
 
 
 def send_email(template_name, subject, to_users, recipient_name,
-               course_run=None, course=None, context=None, project_coordinator=None):
+               course_run=None, course=None, context=None, project_coordinators=None):
     """ Send an email template out to the given users with some standard context variables.
 
         Arguments:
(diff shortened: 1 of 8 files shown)


course_discovery/apps/course_metadata/emails.py
```diff
@@ -53,21 +53,28 @@ def get_project_coordinator(org):
...
-    role = OrganizationUserRole.objects.filter(organization=org,
-                                               role=InternalUserRole.ProjectCoordinator.value).first()
-    return role.user if role else None
+    are_project_coordinators_present = OrganizationUserRole.objects.filter(
+        organization=org, role=InternalUserRole.ProjectCoordinator.value
+    ).exists()
+
+    if not are_project_coordinators_present:
+        return None
+    project_coordinators = [pc.user for pc in OrganizationUserRole.objects.filter(
```
AliAdnanSohail, line 64: Seems like same query is called above, can we combine these 2 queries?

course_discovery/apps/course_metadata/templates/course_metadata/email/reviewed.txt
```diff
@@ -23,5 +23,5 @@
 {% endblocktrans %}
 
 {% blocktrans trimmed %}
-    Note: This email address is unable to receive replies. For questions or comments, please contact your Project Coordinator at {{ contact_us_email }}.
-{% endblocktrans %}
+    Note: This email address is unable to receive replies. For questions or comments, please contact your Project Coordinator(s):
+{% endblocktrans %}{% for contact_us_email in contact_us_emails %}{% if forloop.last %}{{ contact_us_email }}{% else %}{{ contact_us_email }},{% endif %}{% endfor %}
```
AliAdnanSohail, line 27: are we rendering `contact_us_email` in both if and else?

AliAdnanSohail (approved):
just a few nits

course_discovery/apps/course_metadata/emails.py
zawan-ila: Can we do `return project_coordinators or None` and get rid of the first query?

zawan-ila (approved):

course_discovery/apps/course_metadata/emails.py
DawoudSheraz: The first query is just checking the existence of PCs (https://docs.djangoproject.com/en/4.2/ref/models/querysets/#exists). If PCs exist, then fetch the values from DB.

course_discovery/apps/course_metadata/templates/course_metadata/email/reviewed.txt
DawoudSheraz: Yes. The only difference is that for the last item in list, do not display `,` at the end.

course_discovery/apps/course_metadata/emails.py
DawoudSheraz: But I will try to make it better and do in one query.

course_discovery/apps/course_metadata/templates/course_metadata/email/reviewed.txt
AliAdnanSohail: can't we add `if` condition for `,` only?
something like this
```
{% if forloop.last %},{% endif %}
```

DawoudSheraz: Sure, I will double check this.

DawoudSheraz: Adding the tag within interpolate_html is causing argument issue on the interpolate html tag.

PR #4071, opened 2023-08-31
feat: skip the history object creation

uzairr:
This change will add the modifications in which the history object is created only if the model instance is changed not the otherwise.

PROD-3525


course_discovery/apps/course_metadata/models.py
```diff
@@ -1458,18 +1458,25 @@ def save(self, *args, **kwargs):
...
+            super().save(*args, **kwargs)
+
+            # Course runs calculate enterprise subscription inclusion based off of their parent's status, so we need to
+            # force a recalculation
+            course_runs = self.course_runs.all()
+            for course_run in course_runs:
+                course_run.save()
+        finally:
+            if hasattr(self, 'skip_history_when_saving'):
+                delattr(self, 'skip_history_when_saving')
```
DawoudSheraz: Rather than doing this, add a new method that is called in place of super. save. Inside that method, we can decide that if self.has_changed is true, call super.save as it is. Otherwise, set skip history attribute to true. This way, the history logic will be contained in a single place. Also, since the new method is being called inside the overridden save, we wont need any changes across the codebase.

AliAdnanSohail (commented):
Just a suggestion: can we create a Mixin which will have save() method in it. And we will add 
logic of `should_history_be_skipped_on_save` in there. By doing this we don't have to change logic in model level. 
We just have to inherit that mixin for that model and rest will be handled in overridden `save` method. 

WDYT?

uzairr:
by and large, it is the same thing that you are suggesting, right now `should_history_be_skipped_on_save` is part of model utils and is independent of any particular model. it is written in this way by taking the reference from the edx-platform which is also implemented as a stand-alone method.
https://github.com/openedx/edx-platform/blob/705ff1ba7cebc01728d66a8ab556ba372ecbab54/openedx/core/djangoapps/site_configuration/models.py#L144

AliAdnanSohail:
But we have to update all `save()` method calling in specific model. If we create something like mixin, we just have to Inherit that mixin and it will override all `save()` methods

uzairr:
this change is aimed to target `super().save()` instead of default `save()`.Maybe i am not able to follow the scenario that you are trying to build here. we can discuss it over a call about it.

course_discovery/apps/course_metadata/model_utils.py
```diff
@@ -24,3 +24,20 @@ def has_model_changed(field_tracker, external_keys=None, excluded_fields=None):
...
+
+def should_history_be_skipped_on_save(obj, *args, **kwargs):
+    """
+    Sets the parameter 'skip_history_on_save' if the object is not changed
+
+    Args:
+        obj: Any Model instance
+        parent_obj: parent object of the instance obj
+    """
+    if not obj.has_changed:
```
AfaqShuaib09: Aren't we planning to use a field tracker here instead of simple `has_changed` method to check for changes in attributes and associated models?

uzairr: `FieldTracker()` is already implemented there! `has_changed` is working on top of it.
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L1407

course_discovery/apps/course_metadata/management/commands/tests/test_deduplicate_course_metadata_history.py
```diff
@@ -30,13 +30,17 @@ def _assert_normal_case_pre_command(self):
         """
         Verify the history count before running the clean-up command.
         """
-        # Induce a few history records:
-        # - 2 updates for courserun1
-        # - 0 updates for courserun2
-        # - 3 updates for courserun3
+        # In PROD-3525, history is only created if the instance is changed rather than saving it multiple times
```
DawoudSheraz: There is no need to mention an internal ticket in the comments.

course_discovery/apps/course_metadata/management/commands/tests/test_deduplicate_course_metadata_history.py
```diff
@@ -61,9 +65,9 @@ def _assert_normal_case_post_command(self):
         courserun3_count_final = CourseRun.history.filter(id=self.courserun3.id).count()  # pylint: disable=no-member
 
         # Ensure that the only history records left are the 3 original creates.
-        assert courserun1_count_final == 1
+        assert courserun1_count_final == 3
         assert courserun2_count_final == 1
-        assert courserun3_count_final == 1
+        assert courserun3_count_final == 3
```
DawoudSheraz, line 70: Please ensure that inline comments are updated to reflect the changes.

course_discovery/apps/course_metadata/model_utils.py
```diff
@@ -24,3 +24,19 @@ def has_model_changed(field_tracker, external_keys=None, excluded_fields=None):
     return len(changed) or any(
         item.has_changed for item in external_keys if hasattr(item, 'has_changed')
     )
+
+
+def should_history_be_skipped_on_save(mro, obj, *args, **kwargs):
+    """
+    Sets the parameter 'skip_history_on_save' if the object is not changed
+    Args:
+        mro: method resolution order of the model
```
DawoudSheraz: nit: Why is the first argument named mro? It is evaluated to super(). Wouldn't super() take care of MRO based on the inheritance structure?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -92,6 +92,40 @@ def test_image_url(self):
         course.image = None
         assert course.image_url == course.card_image_url
 
+    def test_validate_history_created_only_on_change(self):
+        """
+        Validate that course history object would be created if the object is changed otherwise not.
+        """
+        course = factories.CourseFactory()
+        course_run = factories.CourseRunFactory()
+        program = factories.ProgramFactory()
```
DawoudSheraz, line 105: Missing Seat & CourseEntitlement

course_discovery/apps/course_metadata/management/commands/tests/test_deduplicate_course_metadata_history.py
uzairr: yes, a comment was added just above the assert statement(s).

course_discovery/apps/course_metadata/model_utils.py
uzairr: This code is refactored and now it is much more readable!

course_discovery/apps/course_metadata/admin.py
```diff
@@ -122,7 +123,7 @@ class ProductValueAdmin(admin.ModelAdmin):
 
 
 @admin.register(Course)
-class CourseAdmin(DjangoObjectActions, admin.ModelAdmin):
+class CourseAdmin(DjangoObjectActions, SimpleHistoryAdmin):
```
DawoudSheraz, line 126: Can this increase loading time on admin? There are million of rows on prod, just be careful here.

DawoudSheraz (approved):
While I have not tested it out on the local, CI being green gives a confidence boost on the changes here. Love the mixin approach, much cleaner now.

AliAdnanSohail (approved):
LGTM! thanks for considering the suggestion

course_discovery/apps/course_metadata/admin.py
uzairr: I have to add this change because of the difference in working nature of django model admin and simple history admin. Changes in this PR can better be observed via `SimpleHistoryAdmin`. 

Your concern is valid.After its deployment, i ll monitor it and will take any followup action accordingly.

PR #4087, opened 2023-09-12
fix: strip off the spaces in mktg url

uzairr:
Marketing url of the program pages must not contain the leading or trailing spaces.

PROD-3470

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -3303,8 +3303,8 @@ class Program(ManageHistoryMixin, PkSearchableMixin, TimeStampedModel):
     @property
     def marketing_url(self):
         if self.marketing_slug:
-            path = f'{self.type.slug.lower()}/{self.marketing_slug}'
-            return urljoin(self.partner.marketing_site_url_root, path)
+            path = f'{self.type.slug.lower().strip()}/{self.marketing_slug.strip()}'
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -3303,8 +3303,8 @@ def weeks_to_complete_max(self):
     @property
     def marketing_url(self):
         if self.marketing_slug:
-            path = f'{self.type.slug.lower()}/{self.marketing_slug}'
-            return urljoin(self.partner.marketing_site_url_root, path)
+            path = f'{self.type.slug.lower().strip()}/{self.marketing_slug.strip()}'
+            return urljoin(self.partner.marketing_site_url_root.strip(), path)
```
AfaqShuaib09, line 3307: nit: we can skip `strip()` method usage on `marketing_site_url_root` and `type.slug` here.
[suggestion]:  we can add regex validator to the `marketing_slug` field in Program model.

uzairr: yeah, there are multiple workarounds of this problem. this is specifically happened at the time of data input in `marketing_slug` field. To make the error proofing more robust, both of them can be considered! i.e. adding a validator along with having a strip while creating the url

zawan-ila (approved):

AfaqShuaib09 (approved):

PR #4093, opened 2023-09-15
fix: update html_parser for <ul> and <b>/<strong>

AfaqShuaib09:
[PROD-3616](https://2u-internal.atlassian.net/browse/PROD-3616)
--------------------
This PR updates `clean_html` to check for the rtl direction for any of list related tag.

#### Approach:

In order to check for direction of list in html, I've implemented a solution by introducing a flag. This flag helps identify list elements (ul or ol) that have a direction attribute. When such a list element is detected, it is transformed into a standard ul or ol element without any specific direction. This adjustment is necessary because the HTML2TextWithLangSpans.handle function converts HTML text into Markdown text, and Markdown lacks a representation for lists with right-to-left (RTL) direction. Consequently, the code replaces these specialized list elements with standard list items (li) as needed.


course_discovery/apps/course_metadata/utils.py
```diff
@@ -720,6 +745,8 @@ def clean_html(content):
     html_converter.wrap_links = False
     cleaned = html_converter.handle(cleaned).strip()
     cleaned = markdown.markdown(cleaned)
+    for tag in LIST_TAGS:
+        cleaned = cleaned.replace(f'<{tag}>', f'<{tag} dir="rtl">') if is_list_with_dir_attr_present else cleaned
```
DawoudSheraz, line 758: Let's update the docstring to include this information.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -693,7 +693,21 @@ def whitelist_html_tags_attribute(self, tag, dict_attrs, start):
...
+            super().handle_tag(tag, attrs, start)
+        elif tag != 'p':
+            if start:
+                self.outtextf(f'<{tag}')
+                if attrs:
+                    self.outtextf(' ')
+                    self.outtextf(' '.join(f'{attr}="{value}"' for attr, value in attrs))
+                self.outtextf('>')
+            else:
+                self.outtextf(f'</{tag}>')
```
DawoudSheraz: Why is this needed?

AfaqShuaib09: Due to the presence of a paragraph tag with a "direction" attribute, our previous approach involved invoking `super().handle_tag` for tags located within <p dir='rtl'>. This action converted the inner HTML tags into markdown text. However, when we subsequently used `markdown.markdown` on this text, it did not convert the inner markdown text back into HTML. This was because the outer tag was already in HTML format, leading to the assumption that the inner text should also be in HTML format. Therefore, I overrode the super function call in this context and retain the text in its original form if it was enclosed within a <p dir="rtl"> tag.

AfaqShuaib09: Due to the presence of a paragraph tag with a "direction" attribute, our previous approach involved invoking `super().handle_tag` for tags located within `<p dir='rtl'>`. This action converted the inner HTML tags into markdown text. However, when we subsequently used `markdown.markdown` on this text, it did not convert the inner markdown text back into HTML. This was because the outer tag was already in HTML format, leading to the assumption that the inner text should also be in HTML format. Therefore, I overrode the super function call in this context and retain the text in its original form if it was enclosed within a `<p dir="rtl">` tag.

DawoudSheraz: >  it did not convert the inner markdown text back into HTML. This was because the outer tag was already in HTML format, leading to the assumption that the inner text should also be in HTML format.

Please elaborate on this, I am unable to follow this correctly.

AfaqShuaib09: i.e; if the htm text like
```python
<p dir="rtl">hello there, <strong>Welcome to edx</strong></p>
```
When passed through the `html2text.handler`, it will be converted to:
```python
<p dir="rtl">hello there, **Welcome to edx**</p>
```
Since we've previously overridden the method for the `<p> tag with the dir attribute`, it doesn't convert it to Markdown notation. So when we pass the following 
```python
<p dir="rtl">hello there, **Welcome to edx**</p>
```
to the markdown.markdown method, it will consider it already in HTML because its outer tag, `<p>`, is in HTML format instead of Markdown notation.

DawoudSheraz: Ok, summarize this information in docstring then. The current docstring is not clear on this.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/utils.py
```diff
@@ -693,7 +693,21 @@ def whitelist_html_tags_attribute(self, tag, dict_attrs, start):
...
+        for <span> tags and to include the dir attribute for <p> tags
+        """
+        if not self.is_p_tag_with_dir:
+            super().handle_tag(tag, attrs, start)
+        elif tag != 'p':
+            if start:
+                self.outtextf(f'<{tag}')
+                if attrs:
+                    self.outtextf(' ')
+                    self.outtextf(' '.join(f'{attr}="{value}"' for attr, value in attrs))
```
zawan-ila, line 710: This seems to allow **all** tags and **attributes** inside a `p-tag-with-dir-attribute` whereas at other places we seem to be restricting in some way e.g in the `whitelist_html_tags_attribute` method. Does not appear to pose any issues but just wanted to point it out.

zawan-ila (approved):

course_discovery/apps/course_metadata/utils.py
AfaqShuaib09: Let me update the check for this

PR #4096, opened 2023-09-19
perf: N+1 fixes

zawan-ila:
## [PROD-3626](https://2u-internal.atlassian.net/browse/PROD-3626)

This PR fixes most of the N+1 issues on CourseRun.List endpoint.

### Comparison
Screenshot 2023-09-19 at 3 31 16 PM
Screenshot 2023-09-19 at 3 36 45 PM

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1023,10 +1023,17 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
     def prefetch_queryset(cls, queryset=None):
         queryset = super().prefetch_queryset(queryset=queryset)
 
-        return queryset.select_related('language', 'video', 'expected_program_type').prefetch_related(
+        return queryset.select_related(
             'course__level_type',
+            'course__video__image',
+            'language',
+            'video',
+            'expected_program_type'
+        ).prefetch_related(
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1023,10 +1023,17 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
...
+            'course__video__image',
+            'language',
+            'video',
+            'expected_program_type'
+        ).prefetch_related(
+            'course__level_type__translations',
             'transcript_languages',
             'video__image',
+            'language__translations',
             Prefetch('staff', queryset=MinimalPersonSerializer.prefetch_queryset()),
```
DawoudSheraz, line 1037: I am assuming this is fixing all of N+1 issues in this listing endpoint, correct? Can you double check with nplusone package, if not done already?

DawoudSheraz (approved):

course_discovery/apps/api/serializers.py
zawan-ila: Earlier I was using Debug Toolbar only. Now I have verified through nplusone package too that there are no N+1s.

AfaqShuaib09 (approved):

PR #4100, opened 2023-09-20
feat: update program/degree slugs

uzairr:
To improve seo experience, program slugs are updated for all product sources so that the url becomes more meaningful.

PROD-3634


course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+from course_discovery.apps.course_metadata.models import MigrateProgramSlugConfiguration, Program
+from course_discovery.apps.course_metadata.utils import is_valid_slug_format
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """
+    It will update program url slugs to the format 'category/subcategory/org-title' or 'category/custom-slug' for
+    all degrees and programs irrelevant to the product source.
```
DawoudSheraz: nit: no need to mention product source here

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+        parser.add_argument(
+            '--args_from_database',
+            action='store_true',
+            help='Use arguments from the MigrateProgramSlugConfiguration model instead of the command line.',
+        )
+        parser.add_argument(
+            '--args_from_env',
+            help='Link to the Data Spreadsheet',
+            type=bool,
+        )
```
DawoudSheraz: suggestion: since you have added MigrateProgramSlugConfiguration already, the reading from the sheet is unnecessary. It only adds testing complications.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.slug_update_report = []
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dry_run',
+            help='Just display updated slugs, do not commit in DB',
+            type=bool,
+        )
```
DawoudSheraz: The command is setting url slugs against provided uuids. There is no use for a dry run here. The dry run would have been beneficial if the slugs were auto-generated. 
(Looking to what others have to say).

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+        args_from_env = options.get('args_from_env', None)
+
+        # The signal disconnect has been taken from refresh_course_metadata management command.
+        # We only want to invalidate the API response cache once data loading
+        # completes. Disconnecting the api_change_receiver function from post_save
+        # and post_delete signals prevents model changes during data loading from
+        # repeatedly invalidating the cache.
+        for model in apps.get_app_config('course_metadata').get_models():
+            for signal in (post_save, post_delete):
+                signal.disconnect(receiver=api_change_receiver, sender=model)
```
DawoudSheraz: There might be a util method for that in signals.py. If not, you can add the util and call that here.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+
+        for record in self.slug_update_report:
+            report_in_csv_format = report_in_csv_format + f"{record['program_uuid']},{record['old_slug']}," \
+                                                          f"{record['new_slug']},{record['error']}\n"
+
+        return report_in_csv_format
+
+    def validate_program_fields(self, program_uuid, new_url_slug):
+        """
+        Checks whether a given program uuid and slug is in valid format, also updates slug report.
```
DawoudSheraz, line 115: Also, validate that the slug is not in use. Otherwise, the uniqueness error can come.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Returns:
+            True if uuid and slug is in valid format
+        """
+        is_uuid_valid = is_valid_uuid(program_uuid)
+        is_slug_valid = is_valid_slug_format(new_url_slug)
+
+        if not (is_uuid_valid or is_slug_valid):
+            error_field = self.get_error_field(is_uuid_valid, is_slug_valid)
+            error = f"Skipping uuid {program_uuid} because of incorrect {error_field} format"
```
DawoudSheraz: Rather than adding error fields, please add the actual errors in the string. Otherwise, it wont be clear what is the error.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+        Each key is stripped of whitespaces around the edges, converted to lower case,
+        and has internal spaces converted to _. This convention removes the dependency on CSV
+        headers format(Enrollment Track vs Enrollment track) and makes code flexible to ignore
+        any case sensitivity, among other things.
+        """
+        transformed_dict = {}
+        for key, value in data.items():
+            updated_key = key.strip().lower().replace(' ', '_')
+            transformed_dict[updated_key] = value
+        return transformed_dict
```
DawoudSheraz: nit: please convert this into a util. I have seen this being used across many mgt commands and loaders. Better to convert it to util.

course_discovery/apps/course_metadata/constants.py
```diff
@@ -9,7 +9,8 @@
 SUBDIRECTORY_SLUG_FORMAT_REGEX = (
     r'learn\/[a-zA-Z0-9-_]+\/[a-zA-Z0-9-_]+$|'
     r'executive-education\/[a-zA-Z0-9-_]+$|'
-    r'boot-camps\/[a-zA-Z0-9-_]+\/[a-zA-Z0-9-_]+$'
+    r'boot-camps\/[a-zA-Z0-9-_]+\/[a-zA-Z0-9-_]+$|'
+    r'[a-zA-Z0-9-_]+\/[a-zA-Z0-9-_\/]+$'
```
AliAdnanSohail: I think we should create a separate regex for Programs/Degrees, otherwise it will validate wrong slug format for courses. i.e it will allow `any_text/any_text` in course slug

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
AliAdnanSohail: +1 to @DawoudSheraz since we are assigning slug as it is in this command, so we don't need to add dry run

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+        # We only want to invalidate the API response cache once data loading
+        # completes. Disconnecting the api_change_receiver function from post_save
+        # and post_delete signals prevents model changes during data loading from
+        # repeatedly invalidating the cache.
+        for model in apps.get_app_config('course_metadata').get_models():
+            for signal in (post_save, post_delete):
+                signal.disconnect(receiver=api_change_receiver, sender=model)
+
+        try:
+            if args_from_env:
```
AliAdnanSohail: we don't need this argument because we can't give input to Jenkins or ArgoCd. So configuration model is enough for this command

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
AliAdnanSohail: In case of duplicate slug it will throw error which will be logged in CSV. I don't think we need to add any check in this command because we are not auto generating slug in this case and appending any number in slug to avoid duplication will not a proper solution in this phase

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,190 @@
...
+
+            if not self.validate_program_fields(program_uuid, new_url_slug):
+                continue
+
+            try:
+                logger.info(f'Updating Program ({program_uuid}) with slug: {old_slug} to new url slug: {new_url_slug}')
+                if not dry_run:
+                    program = Program.objects.get(uuid=program_uuid)
+                    old_slug = program.marketing_slug
+                    program.marketing_slug = new_url_slug
```
AliAdnanSohail: We need to catch errors in this part of line as well. Because we can get error of duplicate slug or any other while saving the object. And we have to log that error in CSV as well

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1177,3 +1177,20 @@ def validate_slug_format(url_slug, course):
...
+    snake case. For example, Enrollment Track becomes enrollment_track.
+
+    Each key is stripped of whitespaces around the edges, converted to lower case,
+    and has internal spaces converted to _. This convention removes the dependency on CSV
+    headers format(Enrollment Track vs Enrollment track) and makes code flexible to ignore
+    any case sensitivity, among other things.
+    """
+    transformed_dict = {}
+    for key, value in data.items():
+        updated_key = key.strip().lower().replace(' ', '_')
```
zawan-ila, line 1193: [nit] It might be better to write this as a dict comprehension `return {k.strip.lower.replace:v for k,v in data.items}`

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,156 @@
...
+
+            except Program.DoesNotExist:
+                error = f'Unable to locate Program instance with code {program_uuid}'
+                self.update_slug_report(program_uuid, error)
+
+            except Exception as ex:  # pylint: disable=broad-except
+                error = str(ex)
+                self.update_slug_report(program_uuid, error)
+
+            csv_report = self._get_report_in_csv_format()
```
zawan-ila: this block seems to be indented more than it should? Could be something on github's end too

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,146 @@
...
+                continue
+
+            try:
+                program = Program.objects.get(uuid=program_uuid)
+                old_slug = program.marketing_slug
+                program.marketing_slug = new_url_slug
+                program.save()
+                logger.info(f'Updated Program ({program_uuid}) with slug: {old_slug} '
+                            f'to new url slug: {program.marketing_slug}')
+
```
DawoudSheraz: update_slug_report should be executed in success cases as well so that it is easy to compare old and new slugs. It will also be helpful with redirections.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,146 @@
...
+            except Program.DoesNotExist:
+                error = f'Unable to locate Program instance with code {program_uuid}'
+                self.update_slug_report(program_uuid, error)
+
+            except Exception as ex:  # pylint: disable=broad-except
+                error = str(ex)
+                self.update_slug_report(program_uuid, error)
+
+        csv_report = self._get_report_in_csv_format()
+        subject = 'Migrate Program Slugs Summary Report'
```
DawoudSheraz: nit: there is no need to have a separate variable for subject here.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,146 @@
...
+            self.update_slug_report(program_uuid, error.format(program_uuid, 'incorrect slug format'))
+            return False
+
+        if not is_valid_uuid(program_uuid):
+            self.update_slug_report(program_uuid, error.format(program_uuid, 'incorrect uuid'))
+            return False
+
+        if self.is_program_exists(new_url_slug):
+            self.update_slug_report(program_uuid, error.format(program_uuid, 'program already exists'))
+            return False
```
DawoudSheraz: the method should be along the lines of does_program_slug_exist or is_duplicate_slug. The current name misleads that it is checking program object existence.

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,146 @@
...
+
+        return True
+
+    def is_program_exists(self, slug):
+        try:
+            program = Program.objects.get(marketing_slug=slug)
+            return bool(program)
```
DawoudSheraz: Instead of try-except, use .exists() to check the presence of an object against a slug  https://docs.djangoproject.com/en/4.2/ref/models/querysets/#exists

course_discovery/apps/course_metadata/management/commands/tests/test_update_program_url_slugs.py
```diff
@@ -0,0 +1,242 @@
...
+                ),
+                (
+                    LOGGER_PATH,
+                    'INFO',
+                    f'Updated Program ({self.program3.uuid}) with slug: {current_slug_program3} '
+                    f'to new url slug: {self.test_active_url_slugs[2]}'
+                ),
+            )
+
+        assert mock_send_email_for_slug_updates.call_count == 1
```
DawoudSheraz: Also, add assertions against the CSV report/summary

DawoudSheraz (commented):
Small followup comments to address before 👍🏽

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
AliAdnanSohail: Just a suggestion: we don't need this method because in case of duplicate slug it will throw exception and we are already handling that exception and logging the error

course_discovery/apps/course_metadata/management/commands/tests/test_update_program_url_slugs.py
```diff
@@ -0,0 +1,241 @@
...
+                    ),
+                    (
+                        LOGGER_PATH,
+                        'INFO',
+                        f'Updated Program ({self.program3.uuid}) with slug: {current_slug_program3} '
+                        f'to new url slug: {self.test_active_url_slugs[2]}'
+                    ),
+                )
+
+            assert mock_send_email_for_slug_updates.call_count == 1
```
AliAdnanSohail, line 181: Add assertion to check slug is updated or not

AliAdnanSohail (approved):
just small nits and suggestions

AliAdnanSohail:
@uzairr We also need to update this logic, otherwise it will append duplicate program_type in the start of marketing_slug
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L3306

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,137 @@
...
+        send_email_for_slug_updates(
+            csv_report, settings.NOTIFY_SLUG_UPDATE_RECIPIENTS, 'Migrate Program Slugs Summary Report'
+        )
+        logger.info(csv_report)
+
+    def _get_report_in_csv_format(self):
+        report_in_csv_format = "program_uuid,old_slug,new_slug,error\n"
+
+        for record in self.slug_update_report:
+            report_in_csv_format = report_in_csv_format + f"{record['program_uuid']},{record['old_slug']}," \
```
DawoudSheraz: nit: We can remove + here and do everything via f-string (f"{report_in_csv_format}....")

course_discovery/apps/course_metadata/management/commands/update_program_url_slugs.py
```diff
@@ -0,0 +1,137 @@
...
+            self.update_slug_report(program_uuid, error.format(program_uuid, 'incorrect slug format'))
+            return False
+
+        if not is_valid_uuid(program_uuid):
+            self.update_slug_report(program_uuid, error.format(program_uuid, 'incorrect uuid'))
+            return False
+
+        programs_qs = Program.objects.filter(marketing_slug=new_url_slug)
+        if programs_qs.exists():
+            self.update_slug_report(program_uuid, error.format(program_uuid, 'program already exists'))
```
DawoudSheraz: Again, the error does not reflect the actual issue. A program with same slug already exists is the correct error here.

course_discovery/apps/course_metadata/management/commands/tests/test_update_program_url_slugs.py
```diff
@@ -0,0 +1,262 @@
...
+            name='test.csv',
+            content=self.csv_file_content.encode('utf-8'),
+            content_type='text/csv'
+        )
+
+    def write_csv_file_content(self):
+        """
+        Write the csv file content to a file.
+        """
+        # add csv headers
```
DawoudSheraz: nit: no need for this comment.

course_discovery/apps/course_metadata/management/commands/tests/test_update_program_url_slugs.py
DawoudSheraz: +1 on this.

course_discovery/apps/course_metadata/management/commands/tests/test_update_program_url_slugs.py
```diff
@@ -0,0 +1,262 @@
...
+            content=self.csv_file_content.encode('utf-8'),
+            content_type='text/csv'
+        )
+
+        _ = MigrateProgramSlugConfigurationFactory.create(csv_file=self.csv_file, enabled=True)
+
+        call_command(
+            'update_program_url_slugs', args_from_database=True
+        )
+        # Assert that the email is sent and assert the email content
```
DawoudSheraz: nit: no need for this comment, it is self-explanatory.

DawoudSheraz (approved):

PR #4101, opened 2023-09-20
build: add django42 in CI

UsamaSadiq:
## Description
- Before upgrading to Django 4.2, creating an intermediate PR to enable CI tests to run with `django 4.2`.
- It'll help us in identifying potential blockers from upgrading to `Django 4.2`.
- Some dependencies didn’t have support for both Django32 and Django42 so we had to create separate test requirements files for Django 4.2 support


course_discovery/apps/course_metadata/models.py
```diff
@@ -1256,8 +1256,8 @@ class AbstractLocationRestrictionModel(TimeStampedModel):
         (BLOCKLIST, _('Blocklist'))
     )
 
-    countries = MultiSelectField(choices=COUNTRIES, null=True, blank=True)
-    states = MultiSelectField(choices=CONTIGUOUS_STATES, null=True, blank=True)
+    countries = MultiSelectField(choices=COUNTRIES, null=True, blank=True, max_length=len(COUNTRIES))
```
UsamaSadiq, line 1259: `multiselectfield` does not have Django>4 support and raises IndexError. An [issue](https://github.com/goinnn/django-multiselectfield/issues/134) has been created against upstream repo and fix [PR](https://github.com/goinnn/django-multiselectfield/pull/136/files) has also been created. 
Since there is no response from the author so attempting the fix suggested in the issue discussion locally to fix the test failures.

course_discovery/apps/course_metadata/migrations/0334_auto_20230921_1101.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 3.2.20 on 2023-09-21 11:01
```
UsamaSadiq, line 1: New migration created because model was updated to handle multiselectfield issue.

Makefile
```diff
@@ -41,6 +41,11 @@ requirements.python: ## Install Python requirements for local development.
 
 requirements: requirements.js requirements.python ## Install Python and JS requirements for local development
 
+requirements.python_dj42: ## Install Python requirements for Django 4.2 env
```
UsamaSadiq, line 44: Added temporary targets to provide support for Django 4.2. These can be removed once we upgrade to Django 4.2

awais786 (approved):

Ali-D-Akbar (approved):

PR #4109, opened 2023-09-22
chore:  Add CompressedCacheResponseMixin to CourseRun Viewset

AfaqShuaib09:
[PROD-3627](https://2u-internal.atlassian.net/browse/PROD-3627)
-------
This PR adds `CompressedCacheResponseMixin` to `CourseRunViewSet`. `CompressedCacheResponseMixin` adds the functionality of key-value based caching.

Testing:
- On master branch, hit course_run endpoint after commenting the following line
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/cache.py#L101 
- Look into query count from django toolbar. Hit course_run endpoint again. Query count will not decrease.
- Checkout to `afaq/prod-3627` and hit course_run endpoint. You will notice Query count decreases to (1-3) queries (until you update or edit some model instance)

1st Call:
image

2nd Call:
image

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -38,6 +38,7 @@ LOGGER_NAME = 'course_discovery.apps.api.utils'
 
 
 @ddt.ddt
+@pytest.mark.usefixtures('django_cache')
 class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin, APITestCase):
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -38,6 +38,7 @@
 
 
 @ddt.ddt
+@pytest.mark.usefixtures('django_cache')
```
DawoudSheraz, line 41: why is this needed?

AfaqShuaib09: To set up caching for the tests, I've added it to ensure that no tests will break after its inclusion.
It is also present in previous viewsets tests which are inherited with CompressedCacheResponseMixin.
i.e; https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/tests/test_views/test_courses.py#L46

DawoudSheraz (approved):
I have not tested it locally but since I have seen the working of the mixin before, it should not cause any issues.

zawan-ila (approved):

PR #4130, opened 2023-09-27
Upgrading django-simple-history package

irtazaakram:
Upgrading history package to its latest version. To minimize the django42 scope make this separate PR.

https://django-simple-history.readthedocs.io/en/latest/#section-1
Upgrading the package.
https://github.com/jazzband/django-simple-history/pull/862 ( this pr shows the get_latest_by usage )

Similars PR merged in credentials https://github.com/openedx/credentials/pull/1787
and platform https://github.com/openedx/edx-platform/pull/32880


requirements/local.txt
```diff
@@ -292,11 +292,8 @@ django-parler==2.3
     # via -r requirements/base.in
 django-ses==3.5.0
     # via taxonomy-connector
-django-simple-history==3.0.0
-    # via
-    #   -c requirements/common_constraints.txt
-    #   -c requirements/constraints.txt
-    #   -r requirements/base.in
+django-simple-history==3.4.0
```
awais786, line 295: verfieid

awais786 (approved):

requirements/constraints.txt
```diff
@@ -26,9 +26,6 @@ authlib==1.0.0rc1
 # The latest version require Django 4.0, remove once we have upgraded to Django 4.0
 django-admin-sortable2<2.0.0
 
-# The latest versions of django-simple-history & django-stdimage have breaking changes
-# Remove this pin after fixing them.
-django-simple-history<=3.0.0
 django-stdimage<=5.3.0
```
UsamaSadiq, line 31: Both of these constraints were added together so we may also be able to remove `django-stdimage` constraint. 
If not, then we should add some description about the constraint for future reference.

course_discovery/settings/base.py
```diff
@@ -769,4 +769,7 @@
 
 NOTIFY_SLUG_UPDATE_RECIPIENTS = []
 
+# disable indexing on history_date
+SIMPLE_HISTORY_DATE_INDEX = False
```
DawoudSheraz, line 773: Why is this done?

requirements/common_constraints.txt
```diff
@@ -20,7 +20,6 @@ Django<4.0
 elasticsearch<7.14.0
 
 # django-simple-history>3.0.0 adds indexing and causes a lot of migrations to be affected
```
DawoudSheraz, line 22: since the pin is removed, the comment should be removed as well.

DawoudSheraz (approved):

course_discovery/settings/base.py
iamsobanjaved: If we don't do this, it will create indexing on the history date and that can bring down site if the table size is huge

requirements/common_constraints.txt
iamsobanjaved: This is the copy of [global constraint file](https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt) and we update this on each `make upgrade` run by `wget`  command, and removing constraint through `sed` command here: https://github.com/edx/video-encode-manager/blob/6cf71026b8c5964d980dcd57abf41a46ee1976a6/Makefile#L106

wget: https://github.com/edx/video-encode-manager/blob/6cf71026b8c5964d980dcd57abf41a46ee1976a6/Makefile#L15

Ali-D-Akbar (approved):

PR #4138, opened 2023-10-02
feat: organization field tracker

AfaqShuaib09:
[PROD-3582](https://2u-internal.atlassian.net/browse/PROD-3582)
---------
This PR adds data modified timestamp field for Organization model


course_discovery/apps/core/models.py
```diff
@@ -111,6 +113,15 @@ def oauth_api_client(self):
...
         )
 
+    @property
+    def has_changed(self):
+        """
+        Returns True if any of the fields tracked by field tracker have changed.
+        """
+        if not self.pk:
+            return False
+        return self.field_tracker.changed()
```
DawoudSheraz, line 123: There is a history model mixin where this check is done, alongwith some additional actions. Use that.

course_discovery/apps/core/models.py
```diff
@@ -111,6 +113,15 @@ def oauth_api_client(self):
...
+
```
DawoudSheraz, line 124: Personally, I dont see much value in doing so in Partner model. This model rarely changes.

course_discovery/apps/course_metadata/models.py
```diff
@@ -301,7 +307,9 @@ class Organization(ManageHistoryMixin, CachedMixin, TimeStampedModel):
     def has_changed(self):
         if not self.pk:
             return False
-        return self.has_model_changed()
+        excluded_fields = ['data_modified_timestamp']
+        external_keys = [self.partner]
```
DawoudSheraz: you can add the lists directly instead of creating these vars.

course_discovery/apps/core/models.py
AfaqShuaib09: I added it just in case to check it.

course_discovery/apps/course_metadata/models.py
```diff
@@ -301,7 +307,7 @@ class Organization(ManageHistoryMixin, CachedMixin, TimeStampedModel):
     def has_changed(self):
         if not self.pk:
             return False
-        return self.has_model_changed()
+        return self.has_model_changed(external_keys=[self.partner], excluded_fields=['data_modified_timestamp'])
```
DawoudSheraz: nit: you wont need external keys here. This will only be effective if the partner is updated in the same session as Organization. Both Org and Partners are handled differently in admin. This is just adding an unnecessary computation. For Partner changes to update timestamp on Organization, we would need a pre_Save signal on Partner save, which is not needed at this point as Partner object rarely changes.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1778,6 +1778,26 @@ def test_user_organizations(self):
...
+    def test_data_modified_timestamp_model_related_field_change(self):
+        """
+        Verify data modified timestamp changes on related field changes in Organization Model
+        """
+        org = factories.OrganizationFactory(partner=PartnerFactory(name='test partner', short_code='test_ptr'))
+        data_modified_timestamp = org.data_modified_timestamp
+        org.partner.short_code = 't_code'
+        org.save()
+        assert data_modified_timestamp < org.data_modified_timestamp
+
```
DawoudSheraz: If you address the external keys comment above, this test will fail. In this test, the partner is accessed from org object and is applicable for external_keys check. But in actual use-case, the org handling is happening via Admin.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -301,7 +307,7 @@ class Organization(ManageHistoryMixin, CachedMixin, TimeStampedModel):
     def has_changed(self):
         if not self.pk:
             return False
-        return self.has_model_changed()
+        return self.has_model_changed(excluded_fields=['data_modified_timestamp'])
```
zawan-ila, line 310: Any particular reason we need `excluded_fields` here, or is it just for safety?

course_discovery/apps/course_metadata/models.py
```diff
@@ -332,12 +338,22 @@ def marketing_url(self):
     def user_organizations(cls, user):
         return cls.objects.filter(organization_extension__group__in=user.groups.all())
 
+    def update_data_modified_timestamp(self):
+        """
+        Update the data_modified_timestamp field to the current time if the organization data has changed.
+        """
+        if not self.data_modified_timestamp:
+            self.data_modified_timestamp = datetime.datetime.now(pytz.UTC)
```
zawan-ila: Can we combine the `if` and `elif`, say with an `or`?

zawan-ila (approved):

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: Just for safety, it is being used previously in other models as well which are inherited with `ManageHistoryMixin`
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L1451-L1457

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: yup we can, lemme do that.

PR #4149, opened 2023-10-16
chore: add remaining language tags

AfaqShuaib09:
[PROD-3675](https://2u-internal.atlassian.net/browse/PROD-3675)
-------
This PR adds remaining language tags to discovery from following lists. Currently, most of following languages available in discovery with variants instead of simple one.
This PR adds simple variant of below mentioned language.

List of Missing Languages:
Arabic
French
German
Portuguese
Spanish


zawan-ila (approved):
[for myself] How do you generate this migration file? I would assume manually but it has the `Generate by django ` on top so part of it is automated?

AliAdnanSohail (approved):

AfaqShuaib09:
> [for myself] How do you generate this migration file? I would assume manually but it has the `Generate by django ` on top so part of it is automated?

I've created it using following command
```python3 
./manage.py makemigration {app-name} --emtpy
```
--empty option will an empty migration file

PR #4162, opened 2023-10-24
feat: send email for course publishing to marketing group

AliAdnanSohail:
[Update email group for Publisher emails with SEO team members](https://2u-internal.atlassian.net/browse/PROD-3632)

## Description
We need to send email when a course is reviewed or published to organic marketing email group. This PR will append that email to watchers list so they can be notified about new course addition


course_discovery/apps/course_metadata/emails.py
```diff
@@ -141,6 +142,8 @@ def send_email(template_name, subject, to_users, recipient_name,
     html_content = template.render(base_context)
 
     to_addresses = [u.email for u in to_users if is_email_notification_enabled(u)]
+    if email_group:
+        to_addresses = to_addresses.append(email_group)
```
DawoudSheraz: email_group is going to be list and this append will add this list as an element of to_addresses. It should be `.extend`

course_discovery/apps/course_metadata/emails.py
```diff
@@ -183,9 +186,10 @@ def send_email_to_project_coordinator(course_run, template_name, subject, contex
     if not project_coordinators:
         log_missing_project_coordinator(course_run.course.key, org, template_name)
         return
-
-    send_email(template_name, subject, project_coordinators, _('Project Coordinator team'), context=context,
-               project_coordinators=project_coordinators, course_run=course_run)
+    send_email(
+        template_name, subject, project_coordinators, _('Project Coordinator team'), context=context,
+        project_coordinators=project_coordinators, course_run=course_run, email_group=settings.ORGANIC_MARKETING_EMAIL
+    )
```
DawoudSheraz: Why is this being added to PC emails? This does not make sense. PC emails are specific, related to PC review.

AfaqShuaib09: +1, We can add seo email group to watcher email method

course_discovery/apps/course_metadata/emails.py
AliAdnanSohail: I'm updating approach and moving this to watcher email method.

course_discovery/apps/course_metadata/emails.py
AliAdnanSohail: Good point, we can add this email group to watcher email method

course_discovery/apps/course_metadata/emails.py
```diff
@@ -239,7 +239,10 @@ def send_email_to_notify_course_watchers(course, course_run_publish_date, course
         'course_marketing_url': course.marketing_url,
         'course_preview_url': course.preview_url,
     }
-    to_users = course.watchers
+    to_users = []
+    if course.watchers:
+        to_users.extend(course.watchers)
+    to_users.append(settings.ORGANIC_MARKETING_EMAIL)
```
DawoudSheraz: Only add if the value is available.
If to_users list is empty after addition of both watchers and marketing email, the flow should halt.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -239,7 +239,10 @@ def send_email_to_notify_course_watchers(course, course_run_publish_date, course
         'course_marketing_url': course.marketing_url,
         'course_preview_url': course.preview_url,
     }
```
DawoudSheraz, line 241: The method is not restricted to course_watchers only now. Please rename it to reflect the changes.

AliAdnanSohail: Marketing people seems like watchers not for a specific course but for all courses. So I think if we don't change the name it still keep sense to send email all watchers including marketing people

course_discovery/apps/course_metadata/emails.py
AliAdnanSohail: It will be some additional safe side check, because this value is coming from settings variable and there will be not any case when this variable is empty

course_discovery/apps/course_metadata/emails.py
DawoudSheraz: Still, watchers provide a special connotation and that there is a dedicated field for that in model. Therefore, it is better to rename the method.

course_discovery/apps/course_metadata/emails.py
DawoudSheraz: Since this repository is open-source, it is possible many wont be using this setting. The default value is also None. Therefore, it is better to make code defensive here.

course_discovery/apps/course_metadata/emails.py
AliAdnanSohail: I've renamed the method, please let me know if you have something more meaningful in your mind

DawoudSheraz: Please update docstring of the method as well.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -239,7 +239,13 @@ def send_email_to_notify_course_watchers(course, course_run_publish_date, course
...
-    to_users = course.watchers
+    to_users = []
+    if course.watchers:
+        to_users.extend(course.watchers)
+    if settings.ORGANIC_MARKETING_EMAIL:
+        to_users.append(settings.ORGANIC_MARKETING_EMAIL)
+    if not to_users:
```
DawoudSheraz, line 247: Add a log statement if the email is not sent.

DawoudSheraz (approved):
2 small comments to address before merge.

AfaqShuaib09 (approved):

PR #4176, opened 2023-11-08
fix: use appropriate OC course type checks for enterprise sub inclusion calculation

DawoudSheraz:
### Description
"Verified Only" course type was not included in the enterprise subscription inclusion checks for Course. Toggling the flag from publisher did not set the value for Course. The old code was checking against the following types:
- audit
- verified-audit
- professional
- empty
- credit-verified-audit
However, the type are the types that have their slugs in discovery codebase. Other types, like honor and verified only do not have slugs in discovery codebase and were setup directly on stage/prod. Secondly, as documented in https://github.com/openedx/course-discovery/blob/master/docs/decisions/0012-enterprise-program-inclusion-boolean.rst, OCs are eligible for enterprise inclusion. OC are all course types except ExecEd and Bootcamp. This PR updates the inclusion check for Course to take care of this details and ensures that all OC courses are unblock for setting the enterprise sub inclusion flag.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/factories.py
+++ b/course_discovery/apps/course_metadata/tests/factories.py
@@ -306,6 +306,7 @@ class CourseTypeFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = CourseType
+        django_get_or_create = ('slug',)
 
     @factory.post_generation
     def entitlement_types(self, create, extracted, **kwargs):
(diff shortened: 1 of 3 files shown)


course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -306,6 +306,7 @@ class CourseTypeFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = CourseType
+        django_get_or_create = ('slug',)
```
zawan-ila, line 309: [confirm] The tests don't break without it. It's just nice to have?

zawan-ila (approved):

course_discovery/apps/course_metadata/tests/factories.py
DawoudSheraz: Yeah, it is a built-in util within factory boy that allows using the mentioned fields as criteria in get_or_create. Some course types are already setup in Discovery and factory boy fails in creating such types. This method ensures that get the existing object if present, else proceed to creating a new one.

Ali-D-Akbar (approved):

PR #4201, opened 2023-12-11
feat: autogenerate program slugs

uzairr:
This change will add the modifications in which the marketing slug of a program will be autogenerated if not submitted in subdirectory format.This slug can be updated without triggering the autogeneration if it is in the sub-directory format else it will be autogenerated.

PROD-3290


course_discovery/apps/publisher/signals.py
```diff
@@ -12,3 +13,27 @@ def add_permissions_to_organization_group(sender, instance, created, **kwargs):
...
+
+        if category_type == 'xseries':
+            slug_head = category_type
+        elif category_type == 'license':
+            slug_head = 'certificates'
+        else:
+            slug_head = f'{category_map[category_type]}/{category_type}' if category_map.get(category_type) \
+                else f'{category_type}'
+
+        instance.marketing_slug = slugify_with_slashes(f'{slug_head}/{org_name}-{org_title}')
```
DawoudSheraz: org_name-org_title? Shouldn't this be org_slug-<program_slug / sluggified program title>?

course_discovery/apps/publisher/signals.py
```diff
@@ -12,3 +13,27 @@ def add_permissions_to_organization_group(sender, instance, created, **kwargs):
             codename__in=['add_person', 'change_person', 'delete_person']
         )
         instance.group.permissions.add(*target_permissions)
+
+
+@receiver(post_save, sender=Program)
+def autogenerate_program_slug(sender, instance, *args, **kwargs):  # pylint: disable=unused-argument
```
DawoudSheraz: Instead of signal, we can handle this better in save(). And we need to skip generation if slug is already in the correct /expected format.

course_discovery/apps/publisher/signals.py
uzairr: we are following this guideline to generate the program slugs
https://docs.google.com/document/d/1O7ddGVuYgo9maGrtX1L3Vz5zr6us1Ce537_uNWyUojg/edit#heading=h.vacmxajjuvrj

course_discovery/apps/publisher/signals.py
uzairr: there is an m2m field named `authoring_organizations` which is needed to be there to generate the slug.If the autogeneration logic is applied inside `save()` then organization objects are not been saved at that moment so autogeneration can't be done in it.

Ali-D-Akbar (approved):

course_discovery/apps/publisher/signals.py
AfaqShuaib09:
```python
slugify_with_slashes(f'{slug_head}/{org_name}-{org_title}')
```
Why are both `org_name` and `org_title` used in the marketing slug? Shouldn't it be `{org-slug}-{program-slug}` instead?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -2888,6 +2889,57 @@ def test_languages(self):
...
+        """
+        Verify that program slug can be autogenerated if not provided in sub-directory format.
+        """
+        program_type = factories.ProgramTypeFactory(slug='doctorate')
+        organizations = [factories.OrganizationFactory(name='test org')]
+        slug = 'test-slug'
+        program = ProgramFactory(
+            title='test-program', authoring_organizations=organizations, type=program_type, marketing_slug=slug
+        )
+        assert program.marketing_slug != slug
```
AfaqShuaib09, line 2940: instead of having it here not equal, can we compare it with the exact expected slug?

course_discovery/apps/publisher/signals.py
AfaqShuaib09: Ah.. seems like you're using program.title but named it org_title which is causing confusion

AfaqShuaib09: Instead of using simple program_title it should be slugify(program_title)

course_discovery/apps/course_metadata/tests/test_models.py
uzairr: slug format is designed based on the requirements of this document: https://docs.google.com/document/d/1O7ddGVuYgo9maGrtX1L3Vz5zr6us1Ce537_uNWyUojg/edit#heading=h.vacmxajjuvrj

please validate any inconsistency based on it.

AfaqShuaib09: You can skip the assertion for the exact slug, but it would be better if you compare it with exact expected slug instead of just checking format.

course_discovery/apps/publisher/signals.py
```diff
@@ -12,3 +16,27 @@ def add_permissions_to_organization_group(sender, instance, created, **kwargs):
...
+@receiver(m2m_changed, sender=Program.authoring_organizations.through)
+def autogenerate_program_slug(sender, instance, *args, **kwargs):  # pylint: disable=unused-argument
+    if instance.authoring_organizations.count() > 0 and not bool(re.fullmatch(slug_format, instance.marketing_slug)):
+        category_map = {
+            'micromasters': 'masters',
+            'microbachelors': 'bachelors',
+            'professional-certificate': 'certificates'
+        }
+        org_name = instance.authoring_organizations.first().slug
+        org_title = instance.title
```
AfaqShuaib09: nit: change variable name (it should be `program_title` instead of `org_title`)

course_discovery/apps/publisher/signals.py
```diff
@@ -12,3 +16,27 @@ def add_permissions_to_organization_group(sender, instance, created, **kwargs):
...
+
+        if category_type == 'xseries':
+            slug_head = category_type
+        elif category_type == 'license':
+            slug_head = 'certificates'
+        else:
+            slug_head = f'{category_map[category_type]}/{category_type}' if category_map.get(category_type) \
+                else f'{category_type}'
+
+        instance.marketing_slug = slugify_with_slashes(f'{slug_head}/{org_name}-{org_title}')
```
AfaqShuaib09: Instead of using simple title it should be `slugify(program_title)`. Because there can be whitespace characters in program title which can't be added to URL slug

AfaqShuaib09 (approved):

PR #4206, opened 2023-12-15
perf: fix N+1 and some duplicate queries issues across CatalogSerializer

DawoudSheraz:
### Description
- Fix N+1 issues in Catalog serializer
- Use courseurlslug to filter active url slug to avoid altering prefetch results
- Use SlugRelatedFieldwithReadSerializer for product source to avoid duplicate source calls in to_representation
- Add caching in  active_url_slug so that multiple calls to property in same call does not lead to repeated DB calls
- Change course run status handling to get statuses in Python Code instead of ORM to avoid disrupting Prefetch query results
- Fix the tests that have their query count decreased

### Testing
- Checkout master
- Go to admin http://localhost:18381/admin/catalogs/catalog/ and add 1-2 catalogs (In one query, use ""{
"content_type":"course", \n
"partner":"edx",\n
}"" and in another catalog query, use *). 
- Go to http://localhost:18381/api/v1/catalogs/2/courses/ or http://localhost:18381/api/v1/catalogs/1/courses/ (1, 2 are PK of catalog) and note the query count.
- Checkout this branch and hit the urls again. Note the new query count. Hit the url again and there should be small deduction in query count (due to caching). 

#### Master
image

#### N+1 fixups
image


course_discovery/apps/course_metadata/models.py
```diff
@@ -1767,8 +1783,14 @@ def set_active_url_slug(self, slug):
         logger.info('The current slug is {}; The slug to be set is {}; Current course is a draft: {}'  # lint-amnesty, pylint: disable=logging-format-interpolation
                     .format(self.url_slug, slug, self.draft))
 
+        # Clear cache for both draft and non-draft entries to ensure data consistency. There are too many branches
+        # in this method, and it is ok to clear cache than to serve stale data.
+        active_url_cache = RequestCache("active_url_cache")
+        active_url_cache.delete(get_cache_key(course_uuid=self.uuid, draft=True))
+        active_url_cache.delete(get_cache_key(course_uuid=self.uuid, draft=False))
+
```
AfaqShuaib09, line 1789: nit: can we wrap this in a method?

course_discovery/apps/course_metadata/models.py
```diff
@@ -1684,15 +1696,19 @@ def course_run_statuses(self):
...
-        runs = self.course_runs.exclude(hidden=True)
-        if runs.filter(status=CourseRunStatus.Unpublished, end__lt=now).exists():
-            statuses = ['archived']
-            runs = runs.exclude(status=CourseRunStatus.Unpublished, end__lt=now)
-        statuses += list(runs.values_list('status', flat=True).distinct().order_by('status'))
-        return statuses
+        for course_run in self.course_runs.all():
+            if course_run.hidden:
+                continue
+            elif course_run.end and course_run.end < now and course_run.status == CourseRunStatus.Unpublished:
```
AfaqShuaib09: why are we using `elif` after `continue`? switching it to simple `if` or using pass instead of continue i believe would be better.

AfaqShuaib09 (approved):
Tested it out locally, SQL query count reduced from 237 (master) to 63 (dsheraz/PROD-3826) for catalog courses endpoint.

Ali-D-Akbar (approved):
Looks good to me. In my local testing case, I only had 1 course in my catalog (0 courses in the other). 
Query count before: 36
Query count after: 28

DawoudSheraz:
> Looks good to me. In my local testing case, I only had 1 course in my catalog (0 courses in the other). Query count before: 36 Query count after: 28

If you can create a few more courses, it can provide some better insights.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Sure.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Sure.

PR #4215, opened 2023-12-29
feat: add anon user rate limiting for taxonomy recommendation endpoint

DawoudSheraz:
### Description

This PR adds rate limiting for unauthenticated requests against Taxonomy Support Recommendation endpoint. This endpoint is being hit with unauth requests aggressively and is causing issues. The throttling is aggressive for a while and might be increased in the future if needed (via settings variable).

Due to how DRF works, the anon throttle does not work for APIs that require authentication (See https://github.com/encode/django-rest-framework/issues/5234). Therefore, this PR adds a Mixin that overrides dispatch and then handle throttling there. 

- https://docs.djangoproject.com/en/4.2/topics/testing/tools/#urlconf-configuration
- https://github.com/encode/django-rest-framework/issues/5234
- https://stackoverflow.com/questions/75365158/django-unable-to-apply-function-view-decorator-to-class-based-view

### Testing
- Checkout this branch and start discovery
- Go to /taxonomy/api/v1/course_recommendations/edX+DemoX/ in an incognito tab/ logged out session
- Hit it 10 times, the 11th should show 429 error and also showcasing the retry-after/wait point

### DIFF
--- a/course_discovery/apps/api/mixins.py
+++ b/course_discovery/apps/api/mixins.py
@@ -2,10 +2,15 @@
 Mixins for the API application.
 """
 # pylint: disable=not-callable
+import math
 
+from django.contrib.auth.models import AnonymousUser
 from elasticsearch.exceptions import RequestError
+from rest_framework import status
 from rest_framework.decorators import action
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
+from rest_framework.throttling import AnonRateThrottle
 
 from course_discovery.apps.edx_elasticsearch_dsl_extensions.backends import (
     FacetedFieldSearchFilterBackend, FacetedQueryFilterBackend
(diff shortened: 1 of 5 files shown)


AfaqShuaib09 (approved):

zawan-ila:
Does it rate limit if invalid auth credentials are provided?

DawoudSheraz:
> Does it rate limit if invalid auth credentials are provided?

I believe it would, though I did not test that route (I tested the browser session route).

Edit: DRF defaults to Anonymous user if a user is not authenticated successfully (https://www.django-rest-framework.org/api-guide/settings/#authentication-settings). The mixin will throttle them as well.

zawan-ila (approved):
Really nice work on this 👍

PR #4219, opened 2024-01-02
refactor: refactor Course.retrieve to prevent calling get_object twice

zawan-ila:
## Overview
Right now, the [retrieve endpoint](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L516-L520) is written such that there are two calls to `get_object`. Once at line 516, and a [second one](https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L55-L58) inside of `super.retrieve` by DRF. This duplicates the database queries inside `get_object`. This PR attempts to fix that.

### [PROD-3833](https://2u-internal.atlassian.net/browse/PROD-3833)

### DIFF
--- a/course_discovery/apps/api/v1/views/courses.py
+++ b/course_discovery/apps/api/v1/views/courses.py
@@ -516,6 +516,7 @@ class CourseViewSet(CompressedCacheResponseMixin, viewsets.ModelViewSet):
         course = self.get_object()
         if get_query_param(request, 'editable') and not course.entitlements.exists():
             create_missing_entitlement(course)
+            course.refresh_from_db(fields=['entitlements'])
         # Rather than call super().retrieve, we instantiate the serializer and return its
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -516,8 +516,8 @@ def retrieve(self, request, *args, **kwargs):
         course = self.get_object()
         if get_query_param(request, 'editable') and not course.entitlements.exists():
             create_missing_entitlement(course)
-
-        return super().retrieve(request, *args, **kwargs)
+        serializer = self.get_serializer(course)
+        return Response(serializer.data)
```
DawoudSheraz: Add an inline comment to explain why this is being done. This is an exceptional case and having a comment is worth it here.

AfaqShuaib09 (approved):

DawoudSheraz (approved):

PR #4223, opened 2024-01-09
perf: database improvements for Course.list

zawan-ila:
## Overview
Some performance improvements for the course listing endpoint

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -12,6 +12,7 @@ from uuid import uuid4
 import pytz
 import waffle  # lint-amnesty, pylint: disable=invalid-django-waffle-import
 from django.contrib.auth import get_user_model
+from django.db.models import Count
 from django.db.models.query import Prefetch
 from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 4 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1344,12 +1345,16 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None):  # lint-am
             'in_year_value',
             'location_restriction'
         ).prefetch_related(
+            # The level_type is here instead of inside select_related to
+            # prevent the optimizer from choosing a poor query plan
```
DawoudSheraz, line 1355: What poor query plan? Can you explain?

course_discovery/apps/api/serializers.py
```diff
@@ -1344,12 +1345,16 @@ def prefetch_queryset(cls, partner, queryset=None, course_runs=None):  # lint-am
...
+            'level_type',
             'expected_learning_items',
             'level_type__translations',
             'prerequisites',
             'collaborators',
             'topics',
             'url_slug_history',
+            Prefetch('url_slug_history', queryset=CourseUrlSlug.objects.filter(is_active=True), to_attr='_prefetched_active_slug'),
```
DawoudSheraz: Should there be filter alongside course=self as well?

course_discovery/apps/api/v1/views/courses.py
```diff
@@ -156,9 +155,9 @@ def get_queryset(self):
                 programs=programs,
             )
         if pub_q and edit_mode:
-            return queryset.filter(Q(key__icontains=pub_q) | Q(title__icontains=pub_q)).order_by(Lower('key'))
+            return queryset.filter(Q(key__icontains=pub_q) | Q(title__icontains=pub_q)).order_by('key')
```
DawoudSheraz, line 158: Does Lower result in extra calls? Any reference?

course_discovery/apps/api/serializers.py
zawan-ila: Right now, here is the query plan that's being chosen.
Screenshot 2024-01-11 at 2 56 59 PM
Note how the sort is done at the end, and that too with a temp table which affects the execution time significantly. Doing the sort without a temp table and earlier in the execution path should decrease the runtime(imo). After playing around with the query, I got rid of the temp table by removing the join with the level_type table. After this, the query plan looks like
Screenshot 2024-01-11 at 2 59 47 PM
Note how the runtime goes from ~300ms to ~40 ms.

course_discovery/apps/api/serializers.py
zawan-ila: I don't think so.

course_discovery/apps/api/v1/views/courses.py
zawan-ila: No extra calls. The purpose is to allow the optimizer to benefit from the `partner_id, key, draft` index. Without it, it chooses the `partner_id, uuid, draft` index. After this change, the query plan looks like
Screenshot 2024-01-11 at 3 01 24 PM
Note how the runtime goes from ~40ms(see my comment above) to ~1 ms.

** Note how the plan still processes `offset+limit` rows even when it looks like we only need `limit` rows. Hence, with large offsets the performance does decrease (I have seen ~100ms too) but still much better than current worst (~400ms). This can also be [fixed](https://explainextended.com/2009/10/23/mysql-order-by-limit-performance-late-row-lookups/) too but would require a change to either our pagination strategy(e.g use CursorPagination) or override some methods in our current code (maybe paginate_queryset?). I have chosen not to do so yet as I'd like to see the impact of these changes first. We can always go for more optimizations later.

course_discovery/apps/api/serializers.py
DawoudSheraz: Right. Any ideas what is so special about level type that it is causing such a difference? Furthermore, are there any other tables in select_related that can impact this?
One thing about LevelType is that it is using TranslatableModel, maybe it is doing something here?

course_discovery/apps/api/serializers.py
DawoudSheraz: Can you double check this, please?

course_discovery/apps/api/v1/views/courses.py
DawoudSheraz: How does the usage of Lower change the index being used? This is new to me. How did you stumble upon this?

course_discovery/apps/api/serializers.py
zawan-ila: I don't know about this. I doubt that it has something to do with TranslatableModel stuff.

course_discovery/apps/api/v1/views/courses.py
zawan-ila: Say you have a column col, and you have an index on it then it is possible to use the index to retrieve the rows in sorted order. Since we have an index on key but we order by Lower(key), mySQL can not benefit from the index on key, and hence does not see any benefit in using the `partner_id, key, draft` index.
Note that this is a high level view of things and I am not familiar with the intricacies of the optimizer. You can find much about this issue online e.g [here](https://stackoverflow.com/questions/73311399/mysql-query-lower-method-in-sql-disables-my-index)

DawoudSheraz (approved):
While I still have some questions on stuff like level_type impact, I tested it out on local and saw the reduction in time and queries. Please update PR description with related links for future ref.

AfaqShuaib09 (approved):

PR #4226, opened 2024-01-09
feat: Improve DB performance by changing ES logic

Ali-D-Akbar:
[PROD-3873](https://2u-internal.atlassian.net/browse/PROD-3873)
Updates the underlying ES logic for AggregateSearchSerializer which uses CourseSearchDocumentSerializer and LearnerPathwaySearchDocumentSerializers. These serializers used to fetch each item from the DB one by one. This PR will stay backward compatible along with changing only the problematic serializers to fetch all objects in one DB call. 

This results in significant reduction of DB queryes of Course and LearnerPathway along with it reducing calls to their related models (courserun, courseruntype, seattype etc). Ultimately, this will greatly improve performance for the ES endpoints like `search/all/` and `search/courses/`.

### Testing Instruction

- Update ES index on local: `./manage.py update_index --disable-change-limit`
- Verify all the following URLs and check the decreased amount of queries:

```
router.register(r'search/limited', search_views.LimitedAggregateSearchView, basename='search-limited')
router.register(r'search/all', search_views.AggregateSearchViewSet, basename='search-all')
router.register(r'search/courses', search_views.CourseSearchViewSet, basename='search-courses')
router.register(r'search/course_runs', search_views.CourseRunSearchViewSet, basename='search-course_runs')
router.register(r'search/programs', search_views.ProgramSearchViewSet, basename='search-programs')
router.register(r'search/people', search_views.PersonSearchViewSet, basename='search-people')
```

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
+++ b/course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
@@ -1,4 +1,6 @@
+from django.db import models
 from django_elasticsearch_dsl_drf.serializers import DocumentSerializer
+from rest_framework.serializers import ListSerializer
 
 from course_discovery.apps.api.serializers import COMMON_IGNORED_FIELDS
 from course_discovery.apps.course_metadata.search_indexes import documents
@@ -74,7 +76,41 @@ class LimitedAggregateSearchSerializer(MultiDocumentSerializerMixin, DocumentSer
         }
 
 
-class AggregateSearchSerializer(MultiDocumentSerializerMixin, DocumentSerializer):
+class AggregateSearchListSerializer(MultiDocumentSerializerMixin, ListSerializer):
+    """
+    Custom List Serializer for AggregateSearch to fetch all instances at once.
+    """
+    def to_representation(self, data):
+        """
+        Custom representation for all the grouped multi-serializer instances.
+        This will invoke the respective custom list serializers for all the documents.
+        """
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        grouped_instances = self.group_multi_serializer_instances(iterable)
+
+        representations = []
+
+        for serializer_class, instances_for_serializer in grouped_instances.items():
+            representations += serializer_class(
+                context=self._context, many=True
+            ).to_representation(instances_for_serializer)
+
+        return representations
+
+    class Meta:
+        """
+        Meta options.
+        """
+        serializers = {
+            documents.CourseRunDocument: CourseRunSearchDocumentSerializer,
+            documents.CourseDocument: CourseSearchDocumentSerializer,
+            documents.ProgramDocument: ProgramSearchDocumentSerializer,
+            documents.LearnerPathwayDocument: LearnerPathwaySearchDocumentSerializer,
+            documents.PersonDocument: PersonSearchDocumentSerializer,
+        }
+
+
+class AggregateSearchSerializer(DocumentSerializer):
     """
     Serializer for aggregated elasticsearch documents.
     """
@@ -83,13 +119,6 @@ class AggregateSearchSerializer(MultiDocumentSerializerMixin, DocumentSerializer
         """
         Meta options.
         """
-
+        list_serializer_class = AggregateSearchListSerializer
         document = DummyDocument
         ignore_fields = COMMON_IGNORED_FIELDS
-        serializers = {
-            documents.CourseRunDocument: CourseRunSearchDocumentSerializer,
-            documents.CourseDocument: CourseSearchDocumentSerializer,
-            documents.ProgramDocument: ProgramSearchDocumentSerializer,
-            documents.LearnerPathwayDocument: LearnerPathwaySearchDocumentSerializer,
-            documents.PersonDocument: PersonSearchDocumentSerializer,
-        }
--- a/course_discovery/apps/course_metadata/search_indexes/serializers/common.py
+++ b/course_discovery/apps/course_metadata/search_indexes/serializers/common.py
@@ -22,29 +22,37 @@ class ModelObjectDocumentSerializerMixin:
     Model object document serializer mixin.
 
     This mixin can be added to provide db model interface for elasticsearch response.
-    There is method `get_model_object_by_instance` to fetch Model object.
+    There is method `get_model_object_by_instances` to fetch Model object.
     """
 
-    def get_model_object_by_instance(self, instance):
+    def get_model_object_by_instances(self, instances):
         """
-        Provide Model object by elasticsearch response instance.
+        Provide Model objects by elasticsearch response instances.
+        Fetches all the incoming instances at once and returns model queryset.
         """
+        if not isinstance(instances, list):
+            instances = [instances]
         document = None
-        _object = None
-        index_or_alias_name = ElasticsearchUtils.get_alias_by_index_name(instance.meta.index)
+        _objects = []
+        index_or_alias_name = ElasticsearchUtils.get_alias_by_index_name(instances[0].meta.index)
         for doc in registry.get_documents():
             if index_or_alias_name == doc._index._name:  # pylint: disable=protected-access
                 document = doc
                 break
-        hit = self._build_hit(instance)
-        es_pk = hit['_source'].get('pk')
-        if document and es_pk:
+
+        es_pks = []
+        for instance in instances:
+            es_pks.append(instance.pk)
+
+        hit = self._build_hit(instances[0])
+
+        if document and es_pks:
             try:
-                _object = document(hit).get_queryset().get(pk=es_pk)
+                _objects = document(hit).get_queryset().filter(pk__in=es_pks)
             except ObjectDoesNotExist:
                 log.error("Object could not be found in database for SearchResult '%r'.", self)
 
-        return _object
+        return _objects
 
     @staticmethod
     def _build_hit(instance):
@@ -69,5 +77,5 @@ class DocumentDSLSerializerMixin(ModelObjectDocumentSerializerMixin):
     """
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
+        _object = self.get_model_object_by_instances(instance).get()
         return super().to_representation(_object)
--- a/course_discovery/apps/course_metadata/search_indexes/serializers/course.py
+++ b/course_discovery/apps/course_metadata/search_indexes/serializers/course.py
@@ -1,8 +1,11 @@
+# pylint: disable=W0223
 import datetime
 
 import pytz
+from django.db import models
 from django_elasticsearch_dsl_drf.serializers import DocumentSerializer
 from rest_framework import serializers
+from rest_framework.serializers import ListSerializer
 from taxonomy.choices import ProductTypes
 from taxonomy.utils import get_whitelisted_serialized_skills
 
@@ -16,6 +19,24 @@ from ..documents import CourseDocument
 from .common import DateTimeSerializerMixin, DocumentDSLSerializerMixin, ModelObjectDocumentSerializerMixin
 
 __all__ = ('CourseSearchDocumentSerializer',)
+
+
+class CourseSearchDocumentListSerializer(ModelObjectDocumentSerializerMixin, ListSerializer):
+    """
+    Custom List Serializer for CourseSearchDocument to fetch all instances at once.
+    """
+
+    def to_representation(self, data):
+        """
+        Custom list representation to fetch all the course instances at once.
+        """
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        _objects = list(self.get_model_object_by_instances(iterable))
+
+        object_dict = {obj.pk: obj for obj in _objects}
+        result_tuples = [(item, object_dict.get(item.pk)) for item in iterable]
+
+        return super().to_representation(result_tuples)
 
 
 class CourseSearchDocumentSerializer(ModelObjectDocumentSerializerMixin, DateTimeSerializerMixin, DocumentSerializer):
@@ -120,15 +141,28 @@ class CourseSearchDocumentSerializer(ModelObjectDocumentSerializerMixin, DateTim
             self.fields.pop('outcome', None)
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
-        setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
-        return super().to_representation(instance)
+        """
+        Custom instance representation.
+        The instance needs to be handled differently and can be either of the two:
+
+        1. A tuple consistent of an ES Hit object and a model object to be assigned to the hit object.
+        2. A single ES Hit object
+        """
+        if isinstance(instance, tuple):
+            setattr(instance[0], 'object', instance[1])  # pylint: disable=literal-used-as-attribute
+            prepared_instance = instance[0]
+        else:
+            _object = self.get_model_object_by_instances(instance).get()
+            setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
+            prepared_instance = instance
+        return super().to_representation(prepared_instance)
 
     class Meta:
         """
         Meta options.
         """
 
+        list_serializer_class = CourseSearchDocumentListSerializer
         document = CourseDocument
         ignore_fields = COMMON_IGNORED_FIELDS
         fields = BASE_SEARCH_INDEX_FIELDS + (
--- a/course_discovery/apps/course_metadata/search_indexes/serializers/learner_pathway.py
+++ b/course_discovery/apps/course_metadata/search_indexes/serializers/learner_pathway.py
@@ -1,5 +1,8 @@
+# pylint: disable=W0223
+from django.db import models
 from django_elasticsearch_dsl_drf.serializers import DocumentSerializer
 from rest_framework import serializers
+from rest_framework.serializers import ListSerializer
 
 from course_discovery.apps.api.fields import StdImageSerializerField
 from course_discovery.apps.api.serializers import ContentTypeSerializer
@@ -10,6 +13,20 @@ from ..documents import LearnerPathwayDocument
 from .common import DateTimeSerializerMixin, DocumentDSLSerializerMixin, ModelObjectDocumentSerializerMixin
 
 __all__ = ('LearnerPathwaySearchDocumentSerializer',)
+
+
+class LearnerPathwaySearchDocumentListSerializer(ModelObjectDocumentSerializerMixin, ListSerializer):
+    def to_representation(self, data):
+        """
+        Custom list representation to fetch all the learner_pathway instances at once.
+        """
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        _objects = list(self.get_model_object_by_instances(iterable))
+
+        object_dict = {obj.pk: obj for obj in _objects}
+        result_tuples = [(item, object_dict.get(item.pk)) for item in iterable]
+
+        return super().to_representation(result_tuples)
 
 
 class LearnerPathwaySearchDocumentSerializer(
@@ -34,6 +51,7 @@ class LearnerPathwaySearchDocumentSerializer(
         Meta options.
         """
 
+        list_serializer_class = LearnerPathwaySearchDocumentListSerializer
         document = LearnerPathwayDocument
         ignore_fields = COMMON_IGNORED_FIELDS
         fields = (
@@ -44,9 +62,21 @@ class LearnerPathwaySearchDocumentSerializer(
         )
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
-        setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
-        return super().to_representation(instance)
+        """
+        Custom instance representation.
+        The instance needs to be handled differently and can be either of the two:
+
+        1. A tuple consistent of an ES Hit object and a model object to be assigned to the hit object.
+        2. A single ES Hit object
+        """
+        if isinstance(instance, tuple):
+            setattr(instance[0], 'object', instance[1])  # pylint: disable=literal-used-as-attribute
+            prepared_instance = instance[0]
+        else:
+            _object = self.get_model_object_by_instances(instance).get()
+            setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
+            prepared_instance = instance
+        return super().to_representation(prepared_instance)
 
 
 class LearnerPathwaySearchModelSerializer(DocumentDSLSerializerMixin, ContentTypeSerializer, LearnerPathwaySerializer):
--- a/course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
+++ b/course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
@@ -1,5 +1,5 @@
 import datetime
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
 
 from django.core.exceptions import ImproperlyConfigured
@@ -285,7 +285,7 @@ class MultiDocumentSerializerMixin:
     several indexes are used at the same time.
     """
 
-    def to_representation(self, instance):
+    def to_representation(self, data):
         """
         If we have a serializer mapping, use that.  Otherwise, use standard serializer behavior
         Since we might be dealing with multiple indexes, some fields might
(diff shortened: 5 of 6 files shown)


course_discovery/apps/course_metadata/search_indexes/serializers/common.py
```diff
@@ -25,26 +25,33 @@ class ModelObjectDocumentSerializerMixin:
     There is method `get_model_object_by_instance` to fetch Model object.
     """
 
-    def get_model_object_by_instance(self, instance):
+    def get_model_object_by_instance(self, instances):
```
DawoudSheraz: nit: I would suggest having a separate util for this, for sake of backwards compatibility.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -120,15 +139,19 @@ def __init__(self, *args, **kwargs):
...
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
-        setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
-        return super().to_representation(instance)
+        if isinstance(instance, tuple):
+            setattr(instance[0], 'object', instance[1])  # pylint: disable=literal-used-as-attribute
+            prepared_instance = instance[0]
+        else:
+            prepared_instance = instance
```
DawoudSheraz: Dont we need to set object to instance in any case?

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
```diff
@@ -346,15 +351,19 @@ def patch_multi_serializer_context(self, serializer_class):
         serializer_class.Meta = Meta_origin
         serializer_class._declared_fields = origin_declared_fields  # pylint: disable=protected-access
 
-    def multi_serializer_representation(self, instance):
+    def multi_serializer_representation(self, instances):
```
DawoudSheraz: Again, it would be better to have a new util method that represents what is being done here. This is no longer returning the representation, only serializers grouping.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -74,22 +76,51 @@ class Meta:
...
+        """
+        # Dealing with nested relationships, data can be a Manager,
+        # so, first get a queryset from the Manager if needed
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        grouped_instances = self.multi_serializer_representation(iterable)
+
+        representation = []
+
+        for serializer_class, instances_for_serializer in dict(grouped_instances).items():
+            # Instantiate the serializer class
```
DawoudSheraz: no need for this comment, it is obvious.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -74,22 +76,51 @@ class Meta:
...
 
-class AggregateSearchSerializer(MultiDocumentSerializerMixin, DocumentSerializer):
+class AggregateSearchListSerializer(MultiDocumentSerializerMixin, ListSerializer):
     """
-    Serializer for aggregated elasticsearch documents.
+    Custom List Serializer for AggregateSearch to fetch all instances at once.
     """
+    def to_representation(self, data):
+        """
+        List of object instances -> List of dicts of primitive datatypes.
```
DawoudSheraz: nit: update docstring to include proper details of what is happening here.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -74,22 +76,51 @@ class Meta:
...
+        # so, first get a queryset from the Manager if needed
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        grouped_instances = self.multi_serializer_representation(iterable)
+
+        representation = []
+
+        for serializer_class, instances_for_serializer in dict(grouped_instances).items():
+            # Instantiate the serializer class
+            representation += serializer_class(
+                context=self._context, many=True
```
DawoudSheraz, line 95: do we really need many=True here, considering the individual serializers already have list_serializer_class defined?

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -74,22 +76,51 @@ class Meta:
...
+    def to_representation(self, data):
+        """
+        List of object instances -> List of dicts of primitive datatypes.
+        """
+        # Dealing with nested relationships, data can be a Manager,
+        # so, first get a queryset from the Manager if needed
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        grouped_instances = self.multi_serializer_representation(iterable)
+
+        representation = []
```
DawoudSheraz: nit: representations

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -1,8 +1,11 @@
+# pylint: disable=W0223
```
DawoudSheraz, line 1: why is this needed?

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -18,7 +21,23 @@
...
 
 
-class CourseSearchDocumentSerializer(ModelObjectDocumentSerializerMixin, DateTimeSerializerMixin, DocumentSerializer):
+class CourseSearchDocumentListSerializer(ModelObjectDocumentSerializerMixin, ListSerializer):
+    """
+    Custom List Serializer for CourseSearchDocument to fetch all instances at once.
+    """
+    def to_representation(self, data):
+        """
+        List of object instances -> List of dicts of primitive datatypes.
```
DawoudSheraz: docstring update.

course_discovery/apps/course_metadata/search_indexes/serializers/program.py
```diff
@@ -16,6 +19,20 @@
...
+    """
+    Custom List Serializer for ProgramSearchDocument to fetch all instances at once.
+    """
+    def to_representation(self, data):
+        """
+        List of object instances -> List of dicts of primitive datatypes.
+        """
+        # Dealing with nested relationships, data can be a Manager,
+        # so, first get a queryset from the Manager if needed
+        iterable = data.all() if isinstance(data, models.Manager) else data
```
DawoudSheraz: This to_representation is doing what DRF's default to_representation in ListSerializer does. It is not needed here it seems.

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
```diff
@@ -299,21 +299,26 @@ def to_representation(self, instance):
...
-            index_name = instance.meta['index']
-            index_alias = ElasticsearchUtils.get_alias_by_index_name(index_name)
-            serializer_class = instance_serializers.get(index_alias, None)
-            if not serializer_class:
-                raise ImproperlyConfigured('Could not find serializer for %s in mapping' % index_name)
-        return serializer_class
+    def get_serializer_classes_by_instances(self, instances):
+        if not isinstance(instances, list):
+            instances = [instances]
+        serializer_classes = []
```
DawoudSheraz: Add a comment explaining the situation where `instances` might not be a list.

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
Ali-D-Akbar: I have made the util backwards compatible by adding this

```
        if not isinstance(instances, list):
            instances = [instances]
```

Ali-D-Akbar: I'll update the name for this.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
Ali-D-Akbar: Just in case the custom list serializer isn't invoked, we won't have an instance with object so it handles that case.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
Ali-D-Akbar: In this instance, yes. This will invoke the respective custom list serializers for all the documents.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
Ali-D-Akbar: Pylint fails. Says`serializer_class` is not callable.

course_discovery/apps/course_metadata/search_indexes/serializers/program.py
Ali-D-Akbar: No, it is needed as well. It sends the whole queryset rather than iterating over it.

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
Ali-D-Akbar: Nevermind I've separated the util functions for them.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
DawoudSheraz: But the docserializers already have list_serializer_class defined, shouldn't that do the job? For instance, the same approach is used for AggregateSearch serializer and it works with many=true when called.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
DawoudSheraz: Right, but don't we need to do this in "else"?
```
_object = self.get_model_object_by_instance(instance)
  setattr(instance, 'object', _object)
```

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -74,22 +76,51 @@ class Meta:
...
-    Serializer for aggregated elasticsearch documents.
+    Custom List Serializer for AggregateSearch to fetch all instances at once.
     """
+    def to_representation(self, data):
+        """
+        Custom representation for all the grouped multi-serializer instances.
+        This will invoke the respective custom list serializers for all the documents.
+        """
+        # Dealing with nested relationships, data can be a Manager,
+        # so, first get a queryset from the Manager if needed
```
DawoudSheraz: no need for this comment as it is a default comment.

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
```diff
@@ -69,5 +77,5 @@ class DocumentDSLSerializerMixin(ModelObjectDocumentSerializerMixin):
     """
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
+        _object = self.get_model_object_by_instances(instance).get()
         return super().to_representation(_object)
```
DawoudSheraz, line 81: Since this is in common.py, don't we need to handle tuple usecase here as well?

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
```diff
@@ -358,3 +383,15 @@ def multi_serializer_representation(self, instance):
             representation = serializer_class(context=self._context).to_representation(instance)
 
         return representation
+
+    def group_multi_serializer_instances(self, instances):
+        """
+        Groups Multi serializer instances in a list of dicts.
+        """
+        serializer_classes = self.get_serializer_classes_by_instances(instances)
```
DawoudSheraz: Can't this be consolidated in here? We are getting serializers list first and then looping over it again via zip (to create dict mapping).

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
Ali-D-Akbar: This actually invokes the custom serializers in `list_serializer_class` which means each serializer will handle it's own serializing.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
Ali-D-Akbar: We're not fetching objects from DB in case of programs and courseruns. I've added this custom serializer to keep consistency with the **Grouped Serializers** approach.

DawoudSheraz: But this is course.py, not courseruns.

course_discovery/apps/course_metadata/search_indexes/serializers/program.py
DawoudSheraz: But the queryset is being passed to super().to_representation, which essentially does the same iteration over queryset.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
Ali-D-Akbar: Adding this in the instance as discussed

course_discovery/apps/course_metadata/search_indexes/serializers/program.py
Ali-D-Akbar: Ah right. It makes sense. Removing them for program, courserun and person docs.

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
Ali-D-Akbar: I'm keeping the behavior intact for DSL search variants of documents like `CourseSearchModelSerializer`, `CourseRunSearchModelSerializer` bcz I honestly don't know how `DocumentDSLSerializerMixin` work and don't want to disrupt their working/performance.

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
Ali-D-Akbar: Right. I'm on it.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -74,22 +76,49 @@ class Meta:
...
+        """
+        Custom representation for all the grouped multi-serializer instances.
+        This will invoke the respective custom list serializers for all the documents.
+        """
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        grouped_instances = self.group_multi_serializer_instances(iterable)
+
+        representations = []
+
+        for serializer_class, instances_for_serializer in dict(grouped_instances).items():
```
DawoudSheraz: grouped_instances is already a dict, so why call dict() on it again?

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
```diff
@@ -69,5 +77,5 @@ class DocumentDSLSerializerMixin(ModelObjectDocumentSerializerMixin):
     """
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
+        _object = self.get_model_object_by_instances(instance).get()
```
DawoudSheraz, line 80: nit: don't we need to set the object to instance?

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
DawoudSheraz: It is better to make the code defensive then.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -18,6 +21,20 @@
 __all__ = ('CourseSearchDocumentSerializer',)
 
 
+class CourseSearchDocumentListSerializer(ModelObjectDocumentSerializerMixin, ListSerializer):
+    """
+    Custom List Serializer for CourseSearchDocument to fetch all instances at once.
+    """
+    def to_representation(self, data):
+        """
+        Custom representation for all the course instances.
```
DawoudSheraz: docstring should specify why do we need this custom representation.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -18,6 +21,20 @@
...
+        """
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        _objects = self.get_model_object_by_instances(iterable)
+        objects_list = list(_objects)
```
DawoudSheraz: nit: both can be combined

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -120,15 +137,21 @@ def __init__(self, *args, **kwargs):
             self.fields.pop('outcome', None)
 
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
-        setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
-        return super().to_representation(instance)
+        if isinstance(instance, tuple):
```
DawoudSheraz: update docstring to differentiate b/w instance being a tuple and instance being a Hit.

course_discovery/apps/course_metadata/search_indexes/serializers/learner_pathway.py
```diff
@@ -12,6 +15,17 @@
 __all__ = ('LearnerPathwaySearchDocumentSerializer',)
 
 
+class LearnerPathwaySearchDocumentListSerializer(ModelObjectDocumentSerializerMixin, ListSerializer):
+    def to_representation(self, data):
+        """
+        Custom representation for all the learner_pathway instances.
```
DawoudSheraz: same docstring comment as above to be applied here.

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
```diff
@@ -22,29 +22,37 @@ class ModelObjectDocumentSerializerMixin:
...
-        Provide Model object by elasticsearch response instance.
+        Provide Model objects by elasticsearch response instances.
+        Fetches all the incoming instances at once and returns model queryset.
         """
+        if not isinstance(instances, list):
+            instances = [instances]
         document = None
-        _object = None
-        index_or_alias_name = ElasticsearchUtils.get_alias_by_index_name(instance.meta.index)
+        _object = []
```
DawoudSheraz: nit: `_objects`

DawoudSheraz (commented):
Tested locally. There are two N+1 issues:
1. additional_metadata on course
2. Various models on LearnerPathway
Otherwise, the code seems to be working. I have already left some nit/changes that are to be catered before merge.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
Ali-D-Akbar: Updated.

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
Ali-D-Akbar: Not in this case for some reason 🤔

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
```diff
@@ -358,3 +361,14 @@ def multi_serializer_representation(self, instance):
             representation = serializer_class(context=self._context).to_representation(instance)
 
         return representation
+
+    def group_multi_serializer_instances(self, instances):
+        """
+        Groups Multi serializer instances in a list of dicts.
+        """
+        grouped_instances = defaultdict(list)
```
AfaqShuaib09, line 369: question: any reason behind using defaultdict here?

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -18,6 +21,19 @@
...
+    """
+    Custom List Serializer for CourseSearchDocument to fetch all instances at once.
+    """
+    def to_representation(self, data):
+        """
+        Custom list representation to fetch all the course instances at once.
+        """
+        iterable = data.all() if isinstance(data, models.Manager) else data
+        _objects = list(self.get_model_object_by_instances(iterable))
+        return super().to_representation(list(zip(iterable, _objects)))
```
zawan-ila: I do not think this is correct. You can not assume that the order of `_objects` will match that of iterable. Hence you'll end up zipping wrong objects. i.e you'll end up zipping Course1ESObject with Course2SQLObject and when it will be serialized, you'll have mixed i.e incorrect info.

zawan-ila: Can you verify if you get the correct results here?

zawan-ila: To illustrate my point, here's an object I get in the output of `/search/all`
Screenshot 2024-01-22 at 5 37 01 PM
Note How the ES info and the SQL info belong to different courses.

*I hope it's not just because my local is messed up :-)

zawan-ila: Also you seem to be doing similar zip for `learner_pathway`. You'll need to fix it there too.

DawoudSheraz: Guess this will need to be handled manually via iteration then(?). Both queryset and instances are lists, so the code would need to search for each item individually (whether the hits or queryset is picked). 
However, that brings the question, of the primary keys that are provided to model via list. That list is maintained via the order of instances. That should bring queryset in same order as the provided keys.

DawoudSheraz: If the queryset is indeed in the order of pks provided, this should work out of box (provided the item length is same)
```
>>> Course.everything.filter(id__in=([1, 50, 56, 90, 100])).values_list('pk')
<CourseQuerySet [(1,), (50,), (56,), (90,), (100,)]>
```

DawoudSheraz: Ok, never mind. I verified it locally, this does mess up the results.
image

DawoudSheraz: The ordering messes up because the hit instances are not ordered by pks. When the pks list is sent to queryset, it returns the items ordered by pks (by default). This causes an incorrect tuple setting in zip().

course_discovery/apps/edx_elasticsearch_dsl_extensions/serializers.py
Ali-D-Akbar: Just to define an empty dict.

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
Ali-D-Akbar: Added `iterable.sort(key=lambda x: x.pk) if iterable else None`

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -120,15 +141,28 @@ def __init__(self, *args, **kwargs):
...
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
-        setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
-        return super().to_representation(instance)
+        """
+        Custom instance representation.
+        The instance needs to be handled differently and can be either of the two:
+
+        1. A tuple consistent of a Hit object and a model object to be assigned to the hit object.
+        2. A single model object
```
DawoudSheraz: 2. Isn't it Hit/doc instance, against which the DB object is fetched?

Ali-D-Akbar: Oh.. wait.

course_discovery/apps/course_metadata/search_indexes/serializers/common.py
DawoudSheraz: What is that "some reason" then?

Ali-D-Akbar: (Found the reason after this comment) `DocumentDSLSerializerMixin` is used in `CourseSearchModelSerializer` which creates document from model only. (without using index)

course_discovery/apps/course_metadata/search_indexes/serializers/learner_pathway.py
```diff
@@ -44,9 +62,21 @@ class Meta:
...
     def to_representation(self, instance):
-        _object = self.get_model_object_by_instance(instance)
-        setattr(instance, 'object', _object)  # pylint: disable=literal-used-as-attribute
-        return super().to_representation(instance)
+        """
+        Custom instance representation.
+        The instance needs to be handled differently and can be either of the two:
+
+        1. A tuple consistent of a Hit object and a model object to be assigned to the hit object.
+        2. A single model object
```
DawoudSheraz: nit: 2 is single hit object.

zawan-ila (approved):

PR #4227, opened 2024-01-09
chore: update remaining local configs for Django v4.2

AfaqShuaib09:
[PROD-3864](https://2u-internal.atlassian.net/browse/PROD-3864)
-------------
This PR adds the local configs for `CSRF_TRUSTED_ORIGINS` to whitelist them and update the `STUDIO URL` from to `http://edx.devstack.cms:18010/` (from `http://edx.devstack.studio:18010/`) for the edX partner provisioned in devstack.

### DIFF
--- a/course_discovery/settings/local.py
+++ b/course_discovery/settings/local.py
@@ -57,14 +57,6 @@ JWT_AUTH.update({
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
 })
 
-CSRF_TRUSTED_ORIGINS = (
-    'http://localhost:8734',  # frontend-app-learner-portal-enterprise
-    'http://localhost:1991',  # frontend-app-admin-portal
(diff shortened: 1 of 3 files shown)


DawoudSheraz (approved):

course_discovery/settings/local.py
```diff
@@ -57,6 +57,14 @@
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
 })
 
+CSRF_TRUSTED_ORIGINS = (
```
DawoudSheraz: shouldn't this be in base.py?

AfaqShuaib09: Yes, we can include it in the base.py file. I initially added it here as it is local level config.
I believe placing it in `base.py` would be better.

zawan-ila (approved):

PR #4230, opened 2024-01-11
feat: prefetch course run type

Ali-D-Akbar:
[PROD-3841](https://2u-internal.atlassian.net/browse/PROD-3841)
The PR fixes the only problem it has currently i.e. prefetching course run type for each run beforehand.

Updates the query from:
SELECT ••• FROM `course_metadata_courseruntype` WHERE `course_metadata_courseruntype`.`id` = 1
to
SELECT ••• FROM `course_metadata_courseruntype` WHERE `course_metadata_courseruntype`.`id` IN (1)

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -214,7 +214,7 @@ class CourseRunSearchViewSetTests(mixins.SerializationMixin, mixins.LoginMixin, 
         ([{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True}], 7),
         ([{'title': 'Software Testing', 'excluded': False}, {'title': 'Software Testing 2', 'excluded': False}], 7),
         ([{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True},
-          {'title': 'Software Testing 3', 'excluded': False}], 9),
+          {'title': 'Software Testing 3', 'excluded': False}], 8),
     )
     @ddt.unpack
     def test_excluded_course_run(self, course_runs, expected_queries):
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -214,7 +214,7 @@ def test_exclude_unavailable_program_types(self, path, serializer, result_locati
         ([{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True}], 7),
         ([{'title': 'Software Testing', 'excluded': False}, {'title': 'Software Testing 2', 'excluded': False}], 7),
         ([{'title': 'Software Testing', 'excluded': True}, {'title': 'Software Testing 2', 'excluded': True},
-          {'title': 'Software Testing 3', 'excluded': False}], 9),
+          {'title': 'Software Testing 3', 'excluded': False}], 3),
```
DawoudSheraz: I dont think this one is correct. Can you check locally? There were no course fetch related SQLs for this test in logs.

Ali-D-Akbar: Strange that it is failing on the local for some reason...

Ali-D-Akbar: It does have a threshold of 5 so I believe it does decrease the count a little bit. Changing this to 8 or 7.

DawoudSheraz (approved):

course_discovery/apps/api/v1/tests/test_views/test_search.py
DawoudSheraz: Ok, please double check the queries executed before/after, on local, by adding the wrong value 🙂 .

zawan-ila (approved):

PR #4248, opened 2024-01-26
feat: adding course run statuses to programs/pathways

kiram15:
Adding in the course_run_status variable for both programs and pathways as both can be added under highlights (along with courses). 

https://2u-internal.atlassian.net/browse/ENT-8236
Here’s the epic that this work is under. Essentially, we are using the variable that already existed under the course model course_run_statuses which gives the information about if a course is archived to also apply to programs and pathways. It loops through all the courses and associated course runs in a program/pathway to get this information. We then go on to use this surfaced info in enterprise’s admin and learner portal to communicate to people that those courses they’ve selected for their highlight are archived.

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -2673,7 +2673,6 @@ class TestProgramSearchDocumentSerializer(TestCase):
                                  credit_backing_organizations=[crediting_organization])
         serializer = self.serialize_program(program, self.request)
         expected = self.get_expected_data(program, self.request)
-        # expected['course_run_statuses'] = []
         assert serializer.data == expected
 
     def test_data_without_organizations(self):
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -148,7 +148,6 @@ class MinimalCourseSerializerTests(SiteMixin, TestCase):
         }
 
     def test_data(self):
-        self.maxDiff = None
         request = make_request()
         organizations = OrganizationFactory(partner=self.partner)
         course = CourseFactory(authoring_organizations=[organizations], partner=self.partner)
@@ -1166,6 +1165,7 @@ class ProgramSerializerTests(MinimalProgramSerializerTests):
             'skills': [],
             'subscription_eligible': None,
             'subscription_prices': [],
+            'course_run_statuses': ['published'],
         })
         return expected
 
@@ -1424,6 +1424,7 @@ class ProgramSerializerTests(MinimalProgramSerializerTests):
         expected_micromasters_path = url.sub('', degree.micromasters_url)
 
         # Tack in degree data
+        expected['course_run_statuses'] = []
         expected['curricula'] = [expected_curriculum]
         expected['degree'] = {
             'application_requirements': degree.application_requirements,
@@ -1488,7 +1489,7 @@ class ProgramSerializerTests(MinimalProgramSerializerTests):
         self.assertIsNotNone(serializer.data['subscription_prices'])
 
 
-class PathwaySerialzerTests(TestCase):
+class PathwaySerializerTest(TestCase):
     def test_data(self):
         pathway = PathwayFactory()
         serializer = PathwaySerializer(pathway)
@@ -1503,6 +1504,7 @@ class PathwaySerialzerTests(TestCase):
             'description': pathway.description,
             'destination_url': pathway.destination_url,
             'pathway_type': pathway.pathway_type,
+            'course_run_statuses': [],
         }
         self.assertDictEqual(serializer.data, expected)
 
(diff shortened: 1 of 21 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -4121,6 +4142,28 @@ def validate_partner_programs(cls, partner, programs):
             msg = _('These programs are for a different partner than the pathway itself: {}')
             raise ValidationError(msg.format(', '.join(bad_programs)))
 
+    @property
+    def course_run_statuses(self):
```
marlonkeating, line 4131: Code looks identical to Program.course_run_statuses, can we have it in a helper method?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -2671,6 +2673,7 @@ def test_data(self):
                                  credit_backing_organizations=[crediting_organization])
         serializer = self.serialize_program(program, self.request)
         expected = self.get_expected_data(program, self.request)
+        # expected['course_run_statuses'] = []
```
marlonkeating: Commented code needed?

marlonkeating (changes requested):

course_discovery/apps/course_metadata/models.py
kiram15: The three look similar but they have different things that they loop through depending on the original model, and I feel like showing what its looping through instead of abstracting it makes the code more readable

marlonkeating: As far as I can see, the only line that really differs between the 3 methods is
`for course_run in some_list_of_course_runs:`

In my view that shared structure, along with the common business logic (that we would ideally like to change in as few places as possible should it change in the future) makes this a prime candidate for using a single helper method.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1239,3 +1241,18 @@ def get_product_skill_names(product_identifier, product_type):
...
+    """
+    now = datetime.datetime.now(pytz.UTC)
+    for course_run in course_runs:
+        if course_run.hidden:
+            continue
+        if course_run.end and course_run.end < now and course_run.status == CourseRunStatus.Unpublished:
+            statuses.add('archived')
+        else:
+            statuses.add(course_run.status)
+    return sorted(list(statuses))
```
marlonkeating: We're also sorting `statuses` in the functions that use this helper.  We can do it either place I think, but we only need to do it once.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1239,3 +1241,18 @@ def get_product_skill_names(product_identifier, product_type):
     """
     product_skills = get_whitelisted_serialized_skills(product_identifier, product_type=product_type)
     return list({product_skill['name'] for product_skill in product_skills})
+
+
+def get_course_run_statuses(statuses, course_runs):
```
marlonkeating, line 1246: nit: The `statuses` Set is always empty in all the places this is called.  Do we expect to pass in non-empty Sets in the future, or could we just create the empty set in this method?

marlonkeating (changes requested):

course_discovery/apps/course_metadata/models.py
```diff
@@ -1705,15 +1705,7 @@ def course_run_statuses(self):
...
-        now = datetime.datetime.now(pytz.UTC)
-        for course_run in self.course_runs.all():
-            if course_run.hidden:
-                continue
-            if course_run.end and course_run.end < now and course_run.status == CourseRunStatus.Unpublished:
-                statuses.add('archived')
-            else:
-                statuses.add(course_run.status)
-        return sorted(list(statuses))
+        return get_course_run_statuses(statuses, self.course_runs.all())
```
marlonkeating: Should we be sorting here like we are in the other places?

marlonkeating (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -4121,6 +4127,21 @@ def validate_partner_programs(cls, partner, programs):
...
+
+        Note that it skips hidden courses - this list is typically used for presentational purposes.
+        The code performs the filtering on Python level instead of ORM/SQL because filtering on course_runs
+        invalidates the prefetch on API level.
+        """
+        statuses = set()
+        for program in self.programs.all():
+            for course in program.courses.all():
+                get_course_run_statuses(statuses, course.course_runs.all())
+        return sorted(list(statuses))
```
Ali-D-Akbar, line 4143: In this example, we may have a lot of DB queries for statuses of each course_run in each course in each program.

A better approach would be to have `programs = self.programs.prefetch_related('courses__course_runs')` before all of this hence all of this is done in one single DB call for each program.

Ali-D-Akbar (commented):
Just a suggestion before approval.

Ali-D-Akbar (approved):

PR #4265, opened 2024-02-09
perf: optimise queryset of ES

uzairr:
After making the above changes query count is reduced to its half i.e. Before this change query count was around ~85 queries and after this change it is dropped to ~45 queries.
Following are the screenshots:

**Before this change:**
Screenshot 2024-02-15 at 4 00 50 PM

**After this change:**
Screenshot 2024-02-15 at 3 57 20 PM

The above statistics can be observed by hitting the rest api of the course runs with query param so that it will go the Elasticsearch.

**Steps to Reproduce:**

- Go to http://localhost:18381/api/v1/course_runs/?q=*
- Refresh 2 to 3 times so that caches are populated to make a stable response.
- Observe the Django tool bar, specifically Time and SQL queries.

Follow the above steps first on master branch and then on this PR to notice the drop in query count.

**Note:** One can also observe a drop in the response time of upto ~400ms. Also set the `edit_mode` to `False` directly in the codebase so that rest api with ES query can be loaded efficiently.

- file path: **course_discovery/apps/api/v1/views/course_runs.py**
- after line 85, explicitly set `edit_mode=False`

### DIFF
--- a/course_discovery/apps/api/v1/views/course_runs.py
+++ b/course_discovery/apps/api/v1/views/course_runs.py
@@ -97,12 +97,13 @@ class CourseRunViewSet(CompressedCacheResponseMixin, ValidElasticSearchQueryRequ
             queryset = self.queryset
 
         if q:
-            qs = SearchQuerySetWrapper(CourseRun.search(q).filter('term', partner=partner.short_code))
-            # This is necessary to avoid issues with the filter backend.
-            qs.model = self.queryset.model
-            return qs
+            queryset = SearchQuerySetWrapper(
+                CourseRun.search(q).filter('term', partner=partner.short_code),
+                model=queryset.model
+            )
+        else:
+            queryset = queryset.filter(course__partner=partner)
 
-        queryset = queryset.filter(course__partner=partner)
         return self.get_serializer_class().prefetch_queryset(queryset=queryset)
 
     def get_serializer_context(self):
--- a/course_discovery/apps/core/utils.py
+++ b/course_discovery/apps/core/utils.py
@@ -164,7 +164,7 @@ class SearchQuerySetWrapper:
     def prefetch_related(self, *lookups):
         """Same as QuerySet.prefetch_related()"""
         clone = self._chain()
-        if lookups == (None,):
+        if not lookups or lookups == (None,):
             clone._prefetch_related_lookups = ()  # pylint: disable=protected-access
         else:
             clone._prefetch_related_lookups += lookups
@@ -173,7 +173,7 @@ class SearchQuerySetWrapper:
     def select_related(self, *lookups):
         """Will work same as .prefetch_related()"""
         clone = self._chain()
-        if lookups == (None,):
+        if not lookups or lookups == (None,):
             clone._select_related_lookups = ()  # pylint: disable=protected-access
         else:
             clone._select_related_lookups += lookups
--- a/course_discovery/apps/core/utils.py
+++ b/course_discovery/apps/core/utils.py
@@ -4,6 +4,7 @@ import re
 from collections import namedtuple
 
 from django.conf import settings
+from django.db.models import prefetch_related_objects
 from django_elasticsearch_dsl import Index
 
 IndexMeta = namedtuple("IndexMeta", "name alias")
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
     Decorates a SearchQuerySet object using a generator for efficient iteration
     """
 
-    def __init__(self, qs):
-        self.qs = qs
+    def __init__(self, queryset, model):
+        # This is necessary to act like Django ORM Queryset
+        self.model = model
+
+        self.queryset = queryset
+        self._select_related_lookups = ()
+        self._prefetch_related_lookups = ()
+
+    def prefetch_related(self, *lookups):
+        """Same as QuerySet.prefetch_related()"""
+        clone = self._chain()
+        if lookups == (None,):
+            clone._prefetch_related_lookups = ()  # pylint: disable=protected-access
+        else:
+            clone._prefetch_related_lookups += lookups
+        return clone
+
+    def select_related(self, *lookups):
+        """Will work same as .prefetch_related()"""
+        clone = self._chain()
+        if lookups == (None,):
+            clone._select_related_lookups = ()  # pylint: disable=protected-access
+        else:
+            clone._select_related_lookups += lookups
+        return clone
+
+    def _chain(self):
+        clone = self.__class__(queryset=self.queryset, model=self.model)
+        clone._select_related_lookups = self._select_related_lookups  # pylint: disable=protected-access
+        clone._prefetch_related_lookups = self._prefetch_related_lookups  # pylint: disable=protected-access
+        return clone
 
     def __getattr__(self, item):
-        try:
-            return super().__getattr__(item)
-        except AttributeError:
-            # If the attribute is not found on this class,
-            # proxy the request to the SearchQuerySet.
-            return getattr(self.qs, item)
+        # If the attribute is not found on this class,
+        # proxy the request to the SearchQuerySet.
+        return getattr(self.queryset, item)
 
     def __iter__(self):
-        for result in self.qs:
-            yield result.object
+        results = [r.object for r in self.queryset]
+
+        # Both select_related & prefetch_related will act as prefetch_related
+        prefetch_lookups = set(self._select_related_lookups) | set(self._prefetch_related_lookups)
+        if prefetch_lookups:
+            prefetch_related_objects(results, *prefetch_lookups)
+
+        yield from results
 
     def __getitem__(self, key):
-        if isinstance(key, int) and (key >= 0 or key < self.count()):
-            # return the object at the specified position
-            return self.qs[key].execute()[0].object
-        # Pass the slice/range on to the delegate
-        return SearchQuerySetWrapper(self.qs[key])
+        single_value = isinstance(key, int)
+
+        clone = self._chain()
+        clone.queryset = self.queryset[slice(key, key + 1) if single_value else key]
+
+        if single_value:
+            return list(clone)[0]
+
+        return clone
 
 
 def use_read_replica_if_available(queryset):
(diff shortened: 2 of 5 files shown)


course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
+            clone._select_related_lookups = ()
+        else:
+            clone._select_related_lookups += lookups
+        return clone
+
+    def _chain(self):
+        clone = self.__class__(queryset=self.queryset, model=self.model)
+        clone._select_related_lookups = self._select_related_lookups
+        clone._prefetch_related_lookups = self._prefetch_related_lookups
+        return clone
```
DawoudSheraz: Why are these additions needed?

course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
+        # If the attribute is not found on this class,
+        # proxy the request to the SearchQuerySet.
+        return getattr(self.queryset, item)
 
     def __iter__(self):
-        for result in self.qs:
-            yield result.object
+        results = [r.object for r in self.queryset]
+
+        # Both select_related & prefetch_related will act as prefetch_related
```
DawoudSheraz, line 196: Why should both select and prefetch be treated as prefetch?

course_discovery/apps/core/utils.py
uzairr: To make `SearchQuerySetWrapper` coherent so that it can be used in all the use cases(in case of ES query and without it).

DawoudSheraz: Please provide some links/references as this is a specific use-case and the changes need some further explanation.

uzairr: Idea of the implemented logic here is taken from the django's source code i.e. from its ORM Queryset class.By following this approach `select_related` and `prefetch_related` logic is not only encapsulated inside the `SearchQuerySetWrapper` class but also equally applied on the search results from Elasticsearch.

https://github.com/django/django/blob/977d25416954a72ad100b01762078bf1ceb89a63/django/db/models/query.py#L1599

course_discovery/apps/core/utils.py
uzairr: In case of `select_related` query is being needed to execute on the db but in this particular case(search results are already gathered from ES). So, the query is not needed to be executed that is why `prefetch_related` and `select_related` should act as `prefetch_related`.

https://github.com/django/django/blob/977d25416954a72ad100b01762078bf1ceb89a63/django/db/models/query.py#L523

course_discovery/apps/api/serializers.py
```diff
@@ -2002,6 +2002,7 @@ def prefetch_queryset(cls, partner, queryset=None):
             'degree__costs',
             'degree__deadlines',
             'curricula',
+            'labels',
```
DawoudSheraz: I added that in https://github.com/openedx/course-discovery/pull/4276, alongside other N+1 fixups.

course_discovery/apps/core/utils.py
DawoudSheraz: But this means every relation that can be fetch via join is now fetched by IN and joined internally by Django -- this will increase DB calls overall.

course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
-            # proxy the request to the SearchQuerySet.
-            return getattr(self.qs, item)
+        # If the attribute is not found on this class,
+        # proxy the request to the SearchQuerySet.
+        return getattr(self.queryset, item)
 
     def __iter__(self):
-        for result in self.qs:
-            yield result.object
+        results = [r.object for r in self.queryset]
```
DawoudSheraz, line 194: This changes queryset to be not an iterator, thus losing the perf benefit. The object will be evaluated and will result in hitting the DB.

course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
+
+    def _chain(self):
+        clone = self.__class__(queryset=self.queryset, model=self.model)
+        clone._select_related_lookups = self._select_related_lookups
+        clone._prefetch_related_lookups = self._prefetch_related_lookups
+        return clone
 
     def __getattr__(self, item):
-        try:
-            return super().__getattr__(item)
```
AfaqShuaib09, line 160: Why was the above check, which gets attribute value from the class instead of the queryset, removed?

course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
+        """Same as QuerySet.prefetch_related()"""
+        clone = self._chain()
+        if lookups == (None,):
+            clone._prefetch_related_lookups = ()
+        else:
+            clone._prefetch_related_lookups += lookups
+        return clone
+
+    def select_related(self, *lookups):
+        """Will work same as .prefetch_related()"""
```
AfaqShuaib09, line 174: Why is the overridden method of `select_related` working similarly to `prefetch_related` and having the same logic?

course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
+        if lookups == (None,):
+            clone._select_related_lookups = ()
+        else:
+            clone._select_related_lookups += lookups
+        return clone
+
+    def _chain(self):
+        clone = self.__class__(queryset=self.queryset, model=self.model)
+        clone._select_related_lookups = self._select_related_lookups
+        clone._prefetch_related_lookups = self._prefetch_related_lookups
```
AfaqShuaib09: Are these lookup variables different from those above in the `__init__` method? And how they are working?

course_discovery/apps/core/utils.py
uzairr: i don't think so, it will target only `SearchQuerySetWrapper` queryset which would be created only in case of ES query.

course_discovery/apps/core/utils.py
uzairr: In the subsequent lines below, we need to `prefetch_related_objects` that demands to pass the results ultimately gives the performance benefit. 

https://github.com/django/django/blob/977d25416954a72ad100b01762078bf1ceb89a63/django/db/models/query.py#L524

course_discovery/apps/core/utils.py
uzairr: In case of select_related query is being needed to execute on the db but in this particular case(search results are already gathered from ES). So, the query is not needed to be executed that is why prefetch_related and select_related should act as prefetch_related.

https://github.com/django/django/blob/977d25416954a72ad100b01762078bf1ceb89a63/django/db/models/query.py#L523

course_discovery/apps/core/utils.py
uzairr: I think `return getattr(self.queryset, item)` will serve the purpose so thats why, removed,

course_discovery/apps/core/utils.py
uzairr: No, they are the same, here a clone object is created with `prefetch_queryset`, nothing special.

course_discovery/apps/core/utils.py
```diff
@@ -152,27 +153,63 @@ class SearchQuerySetWrapper:
...
-        return SearchQuerySetWrapper(self.qs[key])
+        single_value = isinstance(key, int)
+
+        clone = self._chain()
+        clone.queryset = self.queryset[slice(key, key + 1) if single_value else key]
+
+        if single_value:
+            return list(clone)[0]
+
+        return clone
```
DawoudSheraz, line 212: what's going on here?

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -97,12 +97,13 @@ def get_queryset(self):
...
 
         if q:
-            qs = SearchQuerySetWrapper(CourseRun.search(q).filter('term', partner=partner.short_code))
-            # This is necessary to avoid issues with the filter backend.
-            qs.model = self.queryset.model
-            return qs
+            queryset = SearchQuerySetWrapper(
+                CourseRun.search(q).filter('term', partner=partner.short_code),
+                model=queryset.model
+            )
```
DawoudSheraz, line 103: Previously, SearchQuerySetWrapper was being returned but now, it is serializer prefetched_queryset. Any unexpected consequences for this?

course_discovery/apps/core/utils.py
DawoudSheraz: Are we certain about this, just in case?

course_discovery/apps/api/v1/views/course_runs.py
uzairr: I am not foreseeing any consequence(s) about this implementation.Technically, it is more semantically related to the Django ORM.
By incorporating this approach, additional benefits like `prefetch_related` and `select_related` can be implemented for ES queries.

course_discovery/apps/core/utils.py
uzairr: yes it will not break, as model object is treated separately now so no need to call `super()`

course_discovery/apps/core/utils.py
uzairr: retrieval of single val and slice object, similar to the python list.

DawoudSheraz: Why is it so different from prior implementation?

uzairr: it has a minor difference in terms of retrieval of a single value.Previously it was returning an object and now it is returning just a single value.

DawoudSheraz: Is it single value of an object or is it still an object?

uzairr: Previously `__getitem__` was returning either an object from a queryset or an object of `SearchQuerySetWrapper`. but now there is no need to cater both of them separately as this refactoring has introduced a coherence. Now, queryset is taking care of it independently.

DawoudSheraz (approved):

AfaqShuaib09 (approved):

PR #4285, opened 2024-03-08
WS-4468: add new field for algolia experiment

bseverino:
https://2u-internal.atlassian.net/browse/WS-4468

Adding an alternative field for `learning_type`, with different product types grouped together into certain categories. We'll toggle which field to use on the frontend using Optimizely. Tested indexing with Algolia stage environment.

**A/C**
1. exec ed, xseries, licenses, certificates, and professional certificates bucketed under "Certificate courses"
2. micro masters and micro bachelors bucketed under "Paths to degrees"
3. masters, bachelors, doctorates bucketed under "Degrees"

**Course example**
Screenshot 2024-03-08 at 10 56 02 AM

### DIFF
--- a/course_discovery/apps/course_metadata/algolia_models.py
+++ b/course_discovery/apps/course_metadata/algolia_models.py
@@ -63,7 +63,7 @@ def delegate_attributes(cls):
     search_fields = ['partner_names', 'partner_keys', 'product_title', 'product_source', 'primary_description',
                      'secondary_description', 'tertiary_description']
     facet_fields = ['availability_level', 'subject_names', 'levels', 'active_languages', 'staff_slugs',
-                    'product_allowed_in', 'product_blocked_in', 'learning_type']
+                    'product_allowed_in', 'product_blocked_in', 'learning_type', 'learning_type_exp']
     ranking_fields = ['availability_rank', 'product_recent_enrollment_count', 'promoted_in_spanish_index',
                       'product_value_per_click_usa', 'product_value_per_click_international',
                       'product_value_per_lead_usa', 'product_value_per_lead_international']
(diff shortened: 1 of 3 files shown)


Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -294,6 +294,29 @@ def program_types(self):
     def learning_type(self):
         return [self.product_type, *self.program_types]
 
+    @property
+    def learning_type_exp(self):
```
DawoudSheraz, line 298: _exp does not signify what the field is supposed to be.

bseverino: I've added comments explaining the purpose of the field

course_discovery/apps/course_metadata/algolia_models.py
```diff
@@ -294,6 +294,33 @@ def program_types(self):
...
+        """
+        Temporary field used as a variant of `learning_type` for an experiment. If the experiment is successful,
+        this will replace `learning_type`.
+        """
+        if self.type.slug == CourseType.EXECUTIVE_EDUCATION_2U:
+            return [_('Certificate courses')]
```
DawoudSheraz, line 304: why is this single element but the others are (product_type, *experiement_types)?

bseverino: Courses are the only product type that can return multiple learning types because they can be part of multiple programs (for example an example response can be `['Course', 'MicroMasters', 'Professional Certificate']` etc). We return the rest of the product types as lists with single items to keep it consistent.

*I didn't single out Boot Camps here because we're not changing the value in the experiment, but it will also only return ['Boot Camp']

DawoudSheraz (approved):

PR #4293, opened 2024-03-18
build: add atlas support in discovery docker

Ali-D-Akbar:
[PROD-3932](https://2u-internal.atlassian.net/browse/PROD-3932)
Adds the pull translation logic during the creation of docker image. Adds a minimal.yml to be used for running the `python manage.py compilemessages` command during the creation of docker image.
## Testing Instructions

1. Create docker image: `docker build . -t temp -f Dockerfile`
2. Run `docker images` and note the IMAGE ID.
3. Run `docker run -it <IMAGE_ID> /bin/bash`
4. Verify that the translation configs are created exactly how we want in [`edx/app/discovery/discovery/course_discovery/conf/locale`](https://github.com/openedx/openedx-translations/tree/main/translations/course-discovery/course_discovery/conf/locale)

Screenshot 2024-03-19 at 8 01 04 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ ARG DISCOVERY_CODE_DIR="${DISCOVERY_APP_DIR}/${DISCOVERY_SERVICE_NAME}"
 ARG DISCOVERY_NODEENV_DIR="${DISCOVERY_APP_DIR}/nodeenvs/${DISCOVERY_SERVICE_NAME}"
 
 ENV PATH "${DISCOVERY_VENV_DIR}/bin:${DISCOVERY_NODEENV_DIR}/bin:$PATH"
-ENV DISCOVERY_CFG "/edx/etc/discovery.yml"
+ENV DISCOVERY_CFG "minimal.yml"
 ENV DISCOVERY_CODE_DIR "${DISCOVERY_CODE_DIR}"
 ENV DISCOVERY_APP_DIR "${DISCOVERY_APP_DIR}"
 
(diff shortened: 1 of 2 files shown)


timmc-edx (approved):

Dockerfile
```diff
@@ -63,6 +64,8 @@ ENV DJANGO_SETTINGS_MODULE "course_discovery.settings.production"
 
 RUN pip install -r ${DISCOVERY_CODE_DIR}/requirements/production.txt
 
+RUN make pull_translations
```
OmarIthawi: I'd move this temporary variable here:

```suggestion
RUN make OPENEDX_ATLAS_PULL=true pull_translations
```

Dockerfile
```diff
@@ -72,6 +75,8 @@ ENV DJANGO_SETTINGS_MODULE "course_discovery.settings.devstack"
 RUN pip install -r ${DISCOVERY_CODE_DIR}/requirements/django.txt
 RUN pip install -r ${DISCOVERY_CODE_DIR}/requirements/local.txt
 
+RUN make pull_translations
```
OmarIthawi: We're pulling translations twice, could you please elaborate?

OmarIthawi (approved):
Overall, this should work pretty well!

I've left few optional comments.

Dockerfile
AfaqShuaib09: +1 here

timmc-edx: Two different layers -- `dev` and `prod` both build from `app`. Python requirements are installed after the split, and translations need to get pulled after requirements are installed, so that gets duplicated as well.

AfaqShuaib09 (approved):

PR #4295, opened 2024-03-21
feat: skill validation configuration to disable validation for course or organization

muhammad-ammar:
**Description:** Add a new model and related taxonomy provider method to disable skill validation for a course or an organization.

**JIRA:** https://2u-internal.atlassian.net/browse/ENT-8457

**Dependency:** https://github.com/openedx/taxonomy-connector/pull/193


saleem-latif (approved):
@muhammad-ammar Please add test for the new provider method here https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/taxonomy_support/tests/test_providers.py 

Once that is done 👍🏼

course_discovery/apps/taxonomy_support/models.py
```diff
@@ -60,3 +66,103 @@ class UpdateCourseRecommendationsConfig(SingletonModel):
 
     def __str__(self):
         return 'Configuration for the update_course_recommendations management command'
+
+
+class SkillValidationConfiguration(TimeStampedModel):
+    """
+    Model to store the configuration for disabling skill validation for a course or organization.
+    """
```
DawoudSheraz, line 75: why do we need this model in here as opposed to taxonomy-connector?

muhammad-ammar: Adding the new model in taxonomy_support app in course discovery will give us below benefits

- Make taxonomy-connector independent of skill validation disable logic
   - This will make integration smoother in case we want to install taxonomy-connector in any other service
- Better user experience and constraints validation in django admin due to foreign key relationships with `Course` and `Organization` models
   - This eliminates the need for users to manually copy and paste the course key and organization key, reducing the likelihood of errors

DawoudSheraz: > Make taxonomy-connector independent of skill validation disable logic

How so? Isn't this linked to taxonomy-connector?

> Better user experience and constraints validation in django admin due to foreign key relationships with Course and Organization models

While I get this point, it does not alone warrant adding this model to discovery. At this point, we are trying to avoid adding many unrelated items in Discovery. Let's see what any other team members have to say.

muhammad-ammar: > How so? Isn't this linked to taxonomy-connector?

This is not directly related to the taxonomy-connector. The logic to disable skill validation lies with the service that intends to utilize the taxonomy. Specifically, in this case, the data (Course and Organization models) required by the taxonomy-connector are part of course-discovery. This is where the [providers](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/taxonomy_support/providers.py) come into play. Providers act as intermediaries between the taxonomy-connector and consumers, providing the required data. Moving this model into the taxonomy would force consumers of the taxonomy-connector to disable skill validation using course or organization, which would be undesirable for them.

course_discovery/apps/taxonomy_support/models.py
```diff
@@ -60,3 +67,107 @@ class UpdateCourseRecommendationsConfig(SingletonModel):
...
+
+    course = models.ForeignKey(
+        Course,
```
DawoudSheraz, line 78: How does it differ across draft and non-draft? I see the filtering is applied on admin but should it not be on field level as well?

course_discovery/apps/taxonomy_support/models.py
```diff
@@ -60,3 +67,107 @@ class UpdateCourseRecommendationsConfig(SingletonModel):
...
+        """
+        try:
+            return True, CourseKey.from_string(course_run_key)
+        except InvalidKeyError:
+            LOGGER.error('[TAXONOMY_SKILL_VALIDATION_CONFIGURATION] Invalid course_run key: [%s]', course_run_key)
+
+        return False, None
+
+    @classmethod
+    def is_disabled(cls, course_run_key) -> bool:
```
DawoudSheraz, line 146: Why is course_run_key coming in here? It should be Course key as Taxonomy skills are calculated on Course level. The validation model is linked with Course, so it makes sense to provide course key here.

course_discovery/apps/taxonomy_support/tests/test_providers.py
```diff
@@ -138,3 +144,61 @@ def test_get_video_xblocks_content(self):
         provider = get_xblock_metadata_provider()
         xblocks = provider.get_xblocks(block_ids)
         assert 'Should not be included in tagging content.' not in xblocks[0].content
+
+
+@pytest.mark.django_db
```
DawoudSheraz, line 149: p.s. is this really needed -- considering we are using Django testcase already?

course_discovery/apps/taxonomy_support/tests/test_providers.py
```diff
@@ -138,3 +144,61 @@ def test_get_video_xblocks_content(self):
...
+            self.course_with_skill_validation_enabled.key
+        ) is False
+
+        # skill validation should be disabled for all courses runs of an orgranization
+        assert get_course_metadata_provider().skill_validation_disabled(
+            self.courserun_with_org_skill_validation_disabled_1.key
+        ) is True
+        assert get_course_metadata_provider().skill_validation_disabled(
+            self.courserun_with_org_skill_validation_disabled_2.key
+        ) is True
```
DawoudSheraz, line 204: nit: it would be better to have these in separate unit tests, even though they will be one-liner.

course_discovery/apps/taxonomy_support/models.py
Ali-D-Akbar: Rather than keeping a tightly coupled model with Course and Org, can't we just keep something like a uuid in the model and keep it in Taxonomy instead? Making a specific model in the consuming app kinda kills the purpose of an add-on service like Taxonomy i.e. we keep a lot of outside info in Discovery. Having this model in Taxonomy should make more sense if we can make it more generic by using an identifier like uuid.

muhammad-ammar: @Ali-D-Akbar Can you add more details regarding the uuid approach? Any prior art?

Ali-D-Akbar: A model that keeps track of uuids of different products (Course, Programs etc) for which we want to disable validation. Something like this:

```
class DisableSkillValidationConfiguration(TimeStampedModel):
    product_uuid = models.UUIDField(blank=False, null=False, default=uuid4, editable=True, verbose_name=_('Product UUID'))
```

I renamed it `DisableSkillValidationConfiguration` so the name shows the model's exact purpose. 
Since you're not using the organization as shown in the PR, I haven't added any config in the model but feel free to add some identifier for that and make a logic around this accordingly.

muhammad-ammar: In our case the product_uuid will be course uuid, right? So the admins have to copy and paste a uuid in django admin, correct? And to find if the skill validation is disable for a course/courseun/org, we will find, lets say courserun  against a courserun key and then see if uuid of the courserun exists in DisableSkillValidationConfiguration model, correct?

course_discovery/apps/taxonomy_support/models.py
muhammad-ammar: This will be a courserun key. This key will be passed via an api from a lms xblock. We find course key of a courserun key and then see if the skill validation is disabled for that course?

DawoudSheraz: Yes, I would advocate getting course key in Taxonomy connector or whatever the service using this. This method should be taking course key in here -- for sake of consistency.

muhammad-ammar:
@DawoudSheraz @Ali-D-Akbar Closing this in favor of https://github.com/openedx/course-discovery/pull/4300

PR #4296, opened 2024-03-21
ENT-8400: Added transcript_languages_search_facet_names to CourseRunSerializer

mahamakifdar19:
### Description
To incorporate the "Transcript Languages" search facet into Algolia, I've added the `transcript_languages_search_facet_names` in the `CourseRunSerializer` to ensure it is included in the API response. 
It contains a list of available human-readable video transcript language names parsed from a language code.

### Jira Ticket
[ENT-8400](https://2u-internal.atlassian.net/browse/ENT-8400)

Screenshot 2024-03-21 at 4 01 20 PM

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -1018,6 +1018,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
         required=False, many=True, slug_field='code',
         queryset=LanguageTag.objects.prefetch_related('translations').order_by('name')
     )
+    transcript_languages_search_facet_names = serializers.SerializerMethodField()
     video = VideoSerializer(required=False, allow_null=True, source='get_video')
     instructors = serializers.SerializerMethodField(help_text='This field is deprecated. Use staff.')
     staff = SlugRelatedFieldWithReadSerializer(slug_field='uuid', required=False, many=True,
@@ -1066,10 +1067,11 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
             'level_type', 'mobile_available', 'hidden', 'reporting_type', 'eligible_for_financial_aid',
             'first_enrollable_paid_seat_price', 'has_ofac_restrictions', 'ofac_comment',
             'enrollment_count', 'recent_enrollment_count', 'expected_program_type', 'expected_program_name',
-            'course_uuid', 'estimated_hours', 'content_language_search_facet_name', 'enterprise_subscription_inclusion'
+            'course_uuid', 'estimated_hours', 'content_language_search_facet_name', 'enterprise_subscription_inclusion',
+            'transcript_languages_search_facet_names'
         )
         read_only_fields = ('enrollment_count', 'recent_enrollment_count', 'content_language_search_facet_name',
-                            'enterprise_subscription_inclusion')
+                            'enterprise_subscription_inclusion', 'transcript_languages_search_facet_names')
 
     def get_instructors(self, obj):  # pylint: disable=unused-argument
         # This field is deprecated. Use the staff field.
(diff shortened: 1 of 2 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -1084,6 +1086,17 @@ def get_content_language_search_facet_name(self, obj):
             return None
         return language.get_search_facet_display(translate=True)
 
+    def get_transcript_languages_search_facet_names(self, obj):
+        transcript_languages = obj.transcript_languages.all()
+        if not transcript_languages:
+            return None
```
irfanuddinahmad, line 1092: Should we not return an empty list []?

mahamakifdar19: @irfanuddinahmad we are storing None when no transcript language is present for a course.

irfanuddinahmad (approved):

jajjibhai008 (approved):

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -705,6 +705,7 @@ def get_expected_data(cls, course_run, request):
             'ofac_comment': course_run.ofac_comment,
             'estimated_hours': get_course_run_estimated_hours(course_run),
             'enterprise_subscription_inclusion': course_run.enterprise_subscription_inclusion,
+            'transcript_languages_search_facet_names': None
```
Ali-D-Akbar, line 708: nit: maybe we can add a more specific value like `["Arabic", "English"]`

Ali-D-Akbar (approved):

DawoudSheraz:
@Ali-D-Akbar I hope you checked that this does not cause N+1 issues.

course_discovery/apps/api/serializers.py
```diff
@@ -1018,6 +1018,7 @@ class CourseRunSerializer(MinimalCourseRunSerializer):
         required=False, many=True, slug_field='code',
         queryset=LanguageTag.objects.prefetch_related('translations').order_by('name')
     )
+    transcript_languages_search_facet_names = serializers.SerializerMethodField()
```
DawoudSheraz, line 1021: the field name is not good/descriptive -- if you want language names only, it should be transcript_language_names. Though I don't know why we need a new field for this -- when we have language info already available.

mahamakifdar19: @DawoudSheraz The field name `transcript_languages_search_facet_names` suggests that it's specifically for Algolia search facets. We already have language information (stored in the transcript_languages attribute) in the form of language codes. Now, we need to extract language names from these codes to store in Algolia. 
That's why we introduced this new variable, which indicates that it contains language names mapped to each language code, specifically for use in Algolia search facets.

DawoudSheraz: If it is only needed for Algolia(and for enterprise usage), the API is not the right place to add this information. It should be done on Enterprise algolia code level.

mahamakifdar19: @DawoudSheraz In past, this approach has been explored [here](https://github.com/openedx/course-discovery/pull/2990) where the catalog can use the langcodes method to convert language codes into human-readable names. However, this method didn't align perfectly with the `LanguageTag` model data stored in the `course-discovery` database and in order to sync it, we would need to replicate the LanguageTag model in the catalog.

A similar strategy has been used previously for similar tasks [(here).](https://github.com/openedx/course-discovery/pull/2990) Language names were included in the CourseRunSerializer within course-discovery. This setup allows the enterprise-catalog to retrieve language names seamlessly through the API endpoint, eliminating the need to parse language codes or replicate the LanguageTag model from course-discovery.

DawoudSheraz: > In past, this approach has been explored https://github.com/openedx/course-discovery/pull/2990 

And

> A similar strategy has been used previously for similar tasks https://github.com/openedx/course-discovery/pull/2990

both seem to be referring to same PR? Anyways, while I get the context, I would suggest finding a better way for doing this. I personally do not agree with how it has been done in the past 🙂 . Discovery APIs already carry too much information and adding an new field for a very specific consumer is not worth it.

mahamakifdar19: @DawoudSheraz Alternative approaches that have been considered in the past are listed in the PR.
image

mahamakifdar19: @DawoudSheraz I will discuss with team markhors, and going forward, we will explore more effective methods of obtaining this information. Currently, replicating such information in the catalog seems like a substantial effort. Could you please confirm if these changes are ready for deployment at this time, and we can work on refining the approach in the future?

PR #4297, opened 2024-03-22
fix: update default database settings

justinhynes:
This fixes an invalid `ENGINE` setting in the default database settings in the `base.py` file.

Sets the default `ENGINE` value to `django.db.backends.mysql`, which matches default values for other IDAs.

This resolves an issue with migrations not running in development environments.


justinhynes:
Closed -- at this time it doesn't look like we're going to update the default in source.

PR #4298, opened 2024-03-22
fix: use different config for different envs

Ali-D-Akbar:
Fixes local devstack issues by adding the prod config file name.


Dockerfile
```diff
@@ -60,6 +60,7 @@ EXPOSE 8381
 FROM app as prod
 
 ENV DJANGO_SETTINGS_MODULE "course_discovery.settings.production"
+ENV DISCOVERY_CFG "minimal.yml"
```
timmc-edx: It looks like this would use minimal settings in prod, which I don't think is what we want.

timmc-edx (approved):
Seems about right!

DawoudSheraz (approved):

AfaqShuaib09 (approved):

PR #4300, opened 2024-03-26
feat: provider utils for skill validation exemption

muhammad-ammar:
**Description:** Add provider utils for skill validation exemption.

**JIRA:** https://2u-internal.atlassian.net/browse/ENT-8457

**Dependency:** https://github.com/openedx/taxonomy-connector/pull/195

### DIFF
--- a/course_discovery/apps/taxonomy_support/providers.py
+++ b/course_discovery/apps/taxonomy_support/providers.py
@@ -27,7 +27,7 @@ from course_discovery.apps.course_metadata.choices import CourseRunStatus
 from course_discovery.apps.course_metadata.contentful_utils import (
     aggregate_contentful_data, fetch_and_transform_bootcamp_contentful_data, fetch_and_transform_degree_contentful_data
 )
-from course_discovery.apps.course_metadata.models import Course, CourseRun, Program
+from course_discovery.apps.course_metadata.models import Course, CourseRun, Organization, Program
 
 
 class DiscoveryCourseMetadataProvider(CourseMetadataProvider):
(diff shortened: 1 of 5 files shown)


course_discovery/apps/taxonomy_support/providers.py
```diff
@@ -72,6 +72,46 @@ def get_all_courses():  # lint-amnesty, pylint: disable=arguments-differ
...
+
+        Arguments:
+            course_run_key(str): course run key
+
+        Returns:
+            str: course key if course run exists, None otherwise
+        """
+        try:
+            course_run = CourseRun.objects.select_related('course').get(draft=False, key=course_run_key)
+            return course_run.course.key
```
DawoudSheraz: p.s. in docstring, we should mention why we need to fetch the object (difference of old and new keys). 
Furthermore, if we need key only, maybe we should fetch that field only.

course_discovery/apps/taxonomy_support/providers.py
```diff
@@ -72,6 +72,46 @@ def get_all_courses():  # lint-amnesty, pylint: disable=arguments-differ
...
+        """
+        Validate that a course with given `course_key` exists.
+
+        Arguments:
+          course_key(str): course key
+
+        Returns:
+          bool: True if course is valid, False otherwise
+        """
+        return Course.objects.filter(draft=False, key=course_key).exists()
```
DawoudSheraz: since you are using .objects, there is no need to add draft=False. objects manager filters out draft objects by default.

course_discovery/apps/taxonomy_support/providers.py
muhammad-ammar: can you explain more about `in docstring, we should mention why we need to fetch the object (difference of old and new keys).`

DawoudSheraz: I meant to explain why the fetch of object is needed to get the course key as opposed to using CourseKey.from_string to get course key.

DawoudSheraz (approved):

Ali-D-Akbar (approved):

PR #4303, opened 2024-03-26
fix: update limit for es response

Ali-D-Akbar:
[PROD-3986](https://2u-internal.atlassian.net/browse/PROD-3986)
Adds a setting to `update_index`. This will increase the `max_result_window` for all existing indexes and subsequently update any newly created index as well. 

~Also updates the unit test that previously didn't support a custom value of `max_result_window` in ES related logic.~
Keeping the same limit count for unit tests as we won't reach this limit while testing.

### DIFF
--- a/course_discovery/apps/api/tests/test_serializers.py
+++ b/course_discovery/apps/api/tests/test_serializers.py
@@ -5,7 +5,6 @@ import re
 from urllib.parse import urlencode
 
 import ddt
-import pytest
 import responses
 from django.test import TestCase
 from django.utils.text import slugify
@@ -2623,9 +2622,7 @@ class PersonSearchModelSerializerTests(PersonSearchDocumentSerializerTest):
         }
 
 
-@pytest.mark.django_db
-@pytest.mark.usefixtures('elasticsearch_dsl_default_connection')
-class TestProgramSearchDocumentSerializer(TestCase):
+class TestProgramSearchDocumentSerializer(ElasticsearchTestMixin, TestCase):
     serializer_class = ProgramSearchDocumentSerializer
 
     def setUp(self):
@@ -2722,9 +2719,7 @@ class ProgramSearchModelSerializerTest(TestProgramSearchDocumentSerializer):
         return expected
 
 
-@pytest.mark.django_db
-@pytest.mark.usefixtures('elasticsearch_dsl_default_connection')
-class TestLearnerPathwaySearchDocumentSerializer(TestCase):
+class TestLearnerPathwaySearchDocumentSerializer(ElasticsearchTestMixin, TestCase):
     serializer_class = LearnerPathwaySearchDocumentSerializer
 
     def setUp(self):
@@ -2782,9 +2777,7 @@ class LearnerPathwaySearchModelSerializerTest(TestLearnerPathwaySearchDocumentSe
         return expected
 
 
-@pytest.mark.django_db
-@pytest.mark.usefixtures('elasticsearch_dsl_default_connection')
-class TestTypeaheadCourseRunSearchSerializer:
+class TestTypeaheadCourseRunSearchSerializer(ElasticsearchTestMixin, TestCase):
     serializer_class = TypeaheadCourseRunSearchSerializer
 
     @classmethod
@@ -2809,9 +2802,7 @@ class TestTypeaheadCourseRunSearchSerializer:
         return serializer
 
 
-@pytest.mark.django_db
-@pytest.mark.usefixtures('elasticsearch_dsl_default_connection')
-class TestTypeaheadProgramSearchSerializer:
+class TestTypeaheadProgramSearchSerializer(ElasticsearchTestMixin, TestCase):
     serializer_class = TypeaheadProgramSearchSerializer
 
     @classmethod
--- a/course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
+++ b/course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
@@ -117,11 +117,13 @@ class Command(DjangoESDSLCommand):
                     )
                     if record_count_is_sane:
                         ElasticsearchUtils.set_alias(conn, alias, new_index_name)
+                        ElasticsearchUtils.update_max_result_window(conn, settings.MAX_RESULT_WINDOW, new_index_name)
                         indexes_pending.pop(new_index_name, None)
                     else:
                         indexes_pending[new_index_name] = index_info_string
                 else:
                     ElasticsearchUtils.set_alias(conn, alias, new_index_name)
+                    ElasticsearchUtils.update_max_result_window(conn, settings.MAX_RESULT_WINDOW, new_index_name)
                     indexes_pending.pop(new_index_name, None)
 
         for index_alias_mapper in alias_mappings:
--- a/course_discovery/apps/core/utils.py
+++ b/course_discovery/apps/core/utils.py
@@ -72,6 +72,11 @@ class ElasticsearchUtils:
         connection.indices.update_aliases(body)
 
     @classmethod
+    def update_max_result_window(cls, connection, max_result_window, index):
+        if connection.indices.exists(index=index):
+            connection.indices.put_settings(body={"index": {"max_result_window": max_result_window}})
+
+    @classmethod
     def create_index(cls, index, conn_name='default'):
         """
         Creates a new index whose name is prefixed with the specified value.
--- a/course_discovery/settings/base.py
+++ b/course_discovery/settings/base.py
@@ -461,11 +461,13 @@ SYNONYMS_MODULE = 'course_discovery.settings.synonyms'
 # (by default it uses the database driver's default setting)
 # https://docs.djangoproject.com/en/3.1/ref/models/querysets/#iterator
 # Thus set the 'chunk_size'
-ELASTICSEARCH_DSL_QUERYSET_PAGINATION = 10000
+ELASTICSEARCH_DSL_QUERYSET_PAGINATION = 15000
 
 # Defining default pagination for all requests to ElasticSearch,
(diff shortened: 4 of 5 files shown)


course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
```diff
@@ -127,6 +127,8 @@ def _update(self, models, options):
         for index_alias_mapper in alias_mappings:
             index_alias_mapper.registered_index._name = index_alias_mapper.alias  # pylint: disable=protected-access
 
+        conn.indices.put_settings(body={"index": {"max_result_window": settings.MAX_RESULT_WINDOW}})
+
```
DawoudSheraz: If this is global setting, it should be set only once. it should not be set in a periodically running management command.

Ali-D-Akbar: This setting is added to each index separately. Having it in this place specifically works as all the new/updated indexes will have this setting.

zawan-ila (commented):
Why do we need it in 2 management commands? Wouldn't the settings update when only one runs? Would it be nicer to extract this into a separate management command?

Ali-D-Akbar:
> Why do we need it in 2 management commands? 

To update this setting for both cases where we're either updating (through update_index) or creating new/rebuilding indexes (through search_index).

> Wouldn't the settings update when only one runs? 

In case when we create new index, it won't have the settings. These settings work for pre-existing indexes.

>Would it be nicer to extract this into a separate management command?

That's debatable although the same command will be needed to run in both of the commands above.

course_discovery/apps/core/tests/mixins.py
```diff
@@ -17,6 +18,10 @@ class ElasticsearchTestMixin:
     def setUp(self):
         super().setUp()
         self.es = get_connection()
+        try:
+            self.es.indices.put_settings(body={"index": {"max_result_window": settings.MAX_RESULT_WINDOW}})
+        except NotFoundError:
+            pass
```
DawoudSheraz: I personally am not a fan of this, specially in unit tests. Its a recipe for weirdness and unexpected behavior.

course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
```diff
@@ -8,6 +9,7 @@
 from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseRunFactory
 
 
+@pytest.mark.django_db
```
DawoudSheraz: why is this needed? I am assuming APITestCase is child of django or DRF's TestCase which takes care of Django DB ops.

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/search_index.py
```diff
@@ -60,6 +60,11 @@ def _create(self, models, aliases, options):
                     )
                 )
                 ElasticsearchUtils.set_alias(es_connection, created_index_info.alias, created_index_info.name)
+        try:
+            es_connection = get_connection()
+            es_connection.indices.put_settings(body={"index": {"max_result_window": settings.MAX_RESULT_WINDOW}})
+        except NotFoundError:
+            pass
```
DawoudSheraz: In what cases will NotFoundError raise?

Ali-D-Akbar: Just in case if no index is created, we won't be adding this setting to any index.

course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
Ali-D-Akbar: True, it can be removed. 👍

course_discovery/apps/core/tests/mixins.py
Ali-D-Akbar: Basically, no setting is added when there is no index.

course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py
DawoudSheraz: Well, go ahead then.

course_discovery/apps/core/tests/mixins.py
DawoudSheraz: under what Testcases is this valid?

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/search_index.py
DawoudSheraz: How so? This is being done inside _create. Under what circumstances will we not create an index despite calling create?

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
DawoudSheraz: Is the operation idempotent?

course_discovery/apps/core/tests/mixins.py
zawan-ila: I don't really see any tests for this. Not sure why this should be required in the first place?

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/search_index.py
zawan-ila: It'd be nice to put this logic inside the `ElasticSearchUtils` class. Then you might be able to get by by calling it within `ElasticsearchUtils.create_index`.

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
```diff
@@ -127,6 +127,8 @@ def _update(self, models, options):
         for index_alias_mapper in alias_mappings:
             index_alias_mapper.registered_index._name = index_alias_mapper.alias  # pylint: disable=protected-access
 
+        conn.indices.put_settings(body={"index": {"max_result_window": settings.MAX_RESULT_WINDOW}})
```
zawan-ila: Similar to my comment above, can we place this inside `ElasticsearchUtils`?

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
Ali-D-Akbar: Yes, it is.

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/search_index.py
Ali-D-Akbar: @zawan-ila can you elaborate on where I should put these settings in the utils?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -2624,8 +2624,7 @@ def get_expected_data(cls, person, request):
 
 
 @pytest.mark.django_db
```
DawoudSheraz: You might not need django_db mark if you are already using Django testcase.

course_discovery/apps/core/tests/mixins.py
DawoudSheraz: For some testing, can you try this in setUpclass? I still dont understand why this is needed for tests as there are not any items in unit tests thaa can cross the threshold.

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/search_index.py
DawoudSheraz: This is what @zawan-ila is referring to https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/core/utils.py#L28. When calling create_index, this result_window update can be done there (possibly without try/catch?).

course_discovery/apps/core/tests/mixins.py
Ali-D-Akbar: The settings needs to be updated bcz I changed `ELASTICSEARCH_DSL_QUERYSET_PAGINATION`, `ELASTICSEARCH_DSL_LOAD_PER_QUERY` to 15000 in `base.py`.
Should I set this to 10k for tests only?

DawoudSheraz: Hmmm, yeah I think we can keep the settings as it is for tests and not complicate this unnecessarily.

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/search_index.py
Ali-D-Akbar: I tried adding this settings isn't being added for all indexes. Seems like it needs to be added after all the indexes and aliases are defined.

DawoudSheraz (approved):

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
```diff
@@ -117,11 +117,13 @@ def _update(self, models, options):
                     )
                     if record_count_is_sane:
                         ElasticsearchUtils.set_alias(conn, alias, new_index_name)
+                        ElasticsearchUtils.update_max_result_window(conn, settings.MAX_RESULT_WINDOW, new_index_name)
```
zawan-ila, line 120: Would be really nice if we could do this just once at the start of the loop.

zawan-ila (approved):

course_discovery/apps/edx_elasticsearch_dsl_extensions/management/commands/update_index.py
Ali-D-Akbar: Should be done only after `set_alias`.

PR #4306, opened 2024-03-28
fix: use correct request type in anon throttle mixin

DawoudSheraz:
### Description

Moves request_initialization at start of dispatch so that WSGIRequest --> DRFRequest conversion can take place and request.user does DRF authentication

### Testing
- You need Postman to test this out (for JWT Authentication ease)
- Go to LMS /admin/oauth2_provider/application/ and setup a new OAuth app. It should be set to Client Credentials public app.
- Go to postman and make POST request against http://localhost:18000/oauth2/access_token. See pic below for body of request. It should return JWT token
image

- Go to discovery and ensure there are recommendation entries at http://localhost:18381/admin/taxonomy_support/courserecommendation/
- In a new Postman tab, make GET request against http://localhost:18381/taxonomy/api/v1/course_recommendations/<key>. Add Authorization: JWT <token> header. 
- For easy testing, you can set CourseRecommendationsViewAnonymousUserThrottle rate to 2/h to ensure that more than 2 requests by JWT Auth does not get throttled.

### DIFF
--- a/course_discovery/apps/api/mixins.py
+++ b/course_discovery/apps/api/mixins.py
@@ -182,7 +182,8 @@ class AnonymousUserThrottleAuthenticatedEndpointMixin:
     See https://github.com/encode/django-rest-framework/issues/5234 for more context.
     """
     def dispatch(self, request, *args, **kwargs):
-        user = request.user
+        initialized_request = self.initialize_request(request, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


zawan-ila (approved):
Haven't tested it out, but looks good!

AfaqShuaib09 (approved):

PR #4318, opened 2024-04-08
add RestrictedCourseRun model

zawan-ila:
### [PROD-4003](https://2u-internal.atlassian.net/browse/PROD-4003)

Add RestrictedCourseRun model


course_discovery/apps/course_metadata/models.py
```diff
@@ -4537,3 +4537,20 @@ class Meta:
...
+    Model to hold information for restricted Course Runs. Consult
+    ADR-27 for more details
+    """
+    course_run = models.OneToOneField(
+        CourseRun, models.CASCADE, related_name='restricted_run',
+        help_text='Course Run that will be restricted'
+    )
+    restricted_to = models.CharField(
+        max_length=255, null=True, blank=True,
+        help_text='Query Param that will expose this run in the APIs'
```
DawoudSheraz: nit: should we have this in Choices? Or how do you anticipate this particular field will be used?

zawan-ila: I ws thinking of this as the value to be passed in the API to include these runs `?include_restricted=<value of restricted_to>` I am not sure how many different values are possible here. In fact, for our current usecase, I dont think we need it at all. 

Happy to shift this to Choices if you can let me know what values are possible here.

DawoudSheraz: Choices will allow the code to be explicit and to specific point. 
For now, as being added in ADR, the two spectrums of APIs in Discovery: ES and non-ES. We can start from there for now. Open to discussion around this.

zawan-ila: I think we should not really make a distinction about the APIs here. My understanding is that we'll add `include_restricted` on all APIs. Hence, I would prefer the values to be something more representative of the runs themselves e.g `custom-enterprise`, `custom-b2c` etc.

DawoudSheraz: Agree, there are much better distinguishing terms.

course_discovery/apps/course_metadata/choices.py
```diff
@@ -79,3 +79,7 @@ class ExternalCourseMarketingType(models.TextChoices):
     ShortCourse = 'short_course', _('Short Course')
     Sprint = 'sprint', _('Sprint')
     CourseStack = 'course_stack', _('Course Stack')
+
+class CourseRunRestriction(models.TextChoices):
+    CustomEnterprise = 'custom-enterprise', _('Custom Enterprise')
```
DawoudSheraz: nit: for sake if consistency, we can rename it to custom-b2b-enterprise

course_discovery/apps/course_metadata/models.py
```diff
@@ -4537,3 +4537,21 @@ class Meta:
...
+    Model to hold information for restricted Course Runs. Consult
+    ADR-27 for more details
+    """
+    course_run = models.OneToOneField(
+        CourseRun, models.CASCADE, related_name='restricted_run',
+        help_text='Course Run that will be restricted'
+    )
+    restricted_to = models.CharField(
+        max_length=255, null=True, blank=True,
+        help_text='Query Param that will expose this run in the APIs',
```
DawoudSheraz: help_text needs an update

DawoudSheraz: Also, instead of restricted_to, it can be renamed to restriction_type(?).

course_discovery/apps/course_metadata/models.py
```diff
@@ -4537,3 +4537,21 @@ class Meta:
 
     def __str__(self):
         return self.arguments
+
+class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
+    """
+    Model to hold information for restricted Course Runs. Consult
```
Ali-D-Akbar: maybe add a little more than what it restricts here.

course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -1,3 +1,4 @@
+import string
```
Ali-D-Akbar: why is this imported?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -3982,3 +3982,16 @@ def test_recommendation_ordering(self):
...
         assert course2_recs[1].key == 'course6'
         assert course2_recs[2].key == 'course4'
+
+class RestrictedCourseRunTests(TestCase):
+    """ Tests for the `RestrictedCourseRun` model. """
+
+    def test_basic(self):
+        """ Some sanity checks """
+        course_run = factories.CourseRunFactory(key="course-v1:SC+BreadX+3T2015")
+        restricted_cr = factories.RestrictedCourseRunFactory(course_run = course_run, restriction_type=CourseRunRestrictionType.CustomB2BEnterprise.value)
```
Ali-D-Akbar: nit: restricted_course_run

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -3982,3 +3982,16 @@ def test_recommendation_ordering(self):
...
+        restricted_cr.refresh_from_db()
```
Ali-D-Akbar: hmmm, why do we need refreshing?

Ali-D-Akbar (approved):
A bunch of comments to address before merging

course_discovery/apps/course_metadata/admin.py
```diff
@@ -1044,6 +1044,10 @@ class ProgramSubscriptionPriceAdmin(admin.ModelAdmin):
                      "program_subscription__program__uuid",
                      "price", "currency__name")
 
+@admin.register(RestrictedCourseRun)
```
DawoudSheraz, line 1048: need empty lines before and after.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4537,3 +4537,21 @@ class Meta:
...
+    ADR-27 for more details
+    """
+    course_run = models.OneToOneField(
+        CourseRun, models.CASCADE, related_name='restricted_run',
+        help_text='Course Run that will be restricted'
+    )
+    restriction_type = models.CharField(
+        max_length=255, null=True, blank=True,
+        choices=CourseRunRestrictionType.choices,
+        help_text='The type of restriction for the above course run'
```
DawoudSheraz: nit: no need to have "above" in there.

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: this should be course_run object, not restricted_course_run

DawoudSheraz (approved):
some small comments, otherwise looks alright.

course_discovery/apps/course_metadata/tests/factories.py
zawan-ila: I was using it earlier. Removed now

course_discovery/apps/course_metadata/tests/test_models.py
zawan-ila: We dont need it. I had it just for fun :-) Removing it now

PR #4330, opened 2024-04-18
fix: avoid span tags overlapping

Ali-D-Akbar:
[PROD-4016](https://2u-internal.atlassian.net/browse/PROD-4016)
Updates the logic to avoid adding multiple span tags around span tags after each time `handle_tag` is called for one of the richtext fields on Publisher. Filters out extra span tags as well.

It used to add extra span tags for a richtext having `dir="rtl"` like this:
`'<p dir="rtl"><span lang="en"><span lang="en">lang and class</span></span></p>'`
And after any subsequent save call to any of the fields using it, it will multiply the `span` tags to something like:
`'<p dir="rtl"><span lang="en"><span lang="en"><span lang="en">lang and class</span></span></span></p>'`

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_utils.py
+++ b/course_discovery/apps/course_metadata/tests/test_utils.py
@@ -807,6 +807,8 @@ class UtilsTests(TestCase):
         ('<p><span lang="en">with lang</span></p>', '<p><span lang="en">with lang</span></p>'),
         ('<p><span class="body" lang="en">lang and class</span></p>', '<p><span lang="en">lang and class</span></p>'),
         ('<p><span class="body">class only</span></p>', '<p>class only</p>'),
+        ('<p dir="rtl"><span lang="en">1 span tag inside RTL p</span></p>', '<p dir="rtl"><span lang="en">1 span tag inside RTL p</span></p>'),
+        ('<p dir="rtl"><span lang="en"><span lang="en"><span lang="en">multiple span tags inside RTL p</span></span></span></p>', '<p dir="rtl"><span lang="en">multiple span tags inside RTL p</span></p>'),
 
         # A sample text from real life when pasting from Microsoft Word into a rich text editor
         # pylint: disable=line-too-long
--- a/course_discovery/apps/course_metadata/utils.py
+++ b/course_discovery/apps/course_metadata/utils.py
@@ -732,7 +732,7 @@ class HTML2TextWithLangSpans(html2text.HTML2Text):
         if not self.is_p_tag_with_dir:
             super().handle_tag(tag, attrs, start)
 
-        elif tag not in HTML_TAGS_ATTRIBUTE_WHITELIST:
+        elif tag not in HTML_TAGS_ATTRIBUTE_WHITELIST and tag != 'span':
             if start:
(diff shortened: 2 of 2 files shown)


AfaqShuaib09 (commented):
would be good if we can add test for this case

course_discovery/apps/course_metadata/utils.py
```diff
@@ -732,7 +732,7 @@ def handle_tag(self, tag, attrs, start):
         if not self.is_p_tag_with_dir:
             super().handle_tag(tag, attrs, start)
 
-        elif tag not in HTML_TAGS_ATTRIBUTE_WHITELIST:
+        elif tag not in HTML_TAGS_ATTRIBUTE_WHITELIST and not tag == 'span':
```
DawoudSheraz: Please add unit tests for this.

Ali-D-Akbar: added

course_discovery/apps/course_metadata/tests/test_utils.py
```diff
@@ -807,6 +807,7 @@ class UtilsTests(TestCase):
         ('<p><span lang="en">with lang</span></p>', '<p><span lang="en">with lang</span></p>'),
         ('<p><span class="body" lang="en">lang and class</span></p>', '<p><span lang="en">lang and class</span></p>'),
         ('<p><span class="body">class only</span></p>', '<p>class only</p>'),
+        ('<p dir="rtl"><span lang="en"><span lang="en">lang and class</span></span></p>', '<p dir="rtl"><span lang="en"><span lang="en">lang and class</span></p>'),
```
DawoudSheraz: Why are there two span opening tags in `<p dir="rtl"><span lang="en"><span lang="en">lang and class</span></p>`? If the tests are passing, that means the functionality is not correct here.

Ali-D-Akbar: Before my fix, this scenario would've multiplied the span tags upto 3 tags.

Ali-D-Akbar: Please look at the description first.

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/tests/test_utils.py
DawoudSheraz: Against the input `'<p dir="rtl"><span lang="en"><span lang="en">lang and class</span></span></p>`, one should expect `<p dir="rtl"><span lang="en">lang and class</span></p>'` as an output. However, the value you have is `<p dir="rtl"><span lang="en"><span lang="en">lang and class</span></p>'`. See how there are 2 opening spans but only one closing span. The HTML is incorrect.

Ali-D-Akbar: Ohh that's a great catch. Lemme look at it again.

Ali-D-Akbar: Update the logic. This logic should also filter out extra span tags as demonstrated in the unit test.

DawoudSheraz (approved):
Once CI is green.

PR #4331, opened 2024-04-18
feat: add CRUD support for RestrictedCourseRun in CourseRun Api

zawan-ila:
[PROD-4004](https://2u-internal.atlassian.net/browse/PROD-4004)
-------
This PR adds support for marking CourseRuns as restricted through the CourseRun Api

Testing Instructions:
---------
You can test it by hitting the API directly with postman or any other tool and provide the `restriction_type` parameter. Then verify that the RestrictedCourseRun objects are added/updated appropriately in the database.

Here's how I have been doing it though.
- Apply the attached patch to Publisher [restriction.patch](https://github.com/openedx/course-discovery/files/15028761/restriction.patch)
- Create/Update runs as you would normally.
- However, just before clicking the Create/Update button, set `window.restriction_type` to whatever restriction_type you want to set. 
- Observe that the objects are created/updated properly in the backend.

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -922,6 +922,7 @@ class MinimalCourseRunSerializer(FlexFieldsSerializerMixin, TimestampModelSerial
                                             queryset=CourseRunType.objects.all())
     term = serializers.CharField(required=False, write_only=True)
     variant_id = serializers.UUIDField(allow_null=True, required=False)
+    restriction_type = serializers.CharField(source='restricted_run.restriction_type', read_only=True)
 
     @classmethod
     def prefetch_queryset(cls, queryset=None):
@@ -932,6 +933,7 @@ class MinimalCourseRunSerializer(FlexFieldsSerializerMixin, TimestampModelSerial
         return queryset.select_related('course', 'type').prefetch_related(
             '_official_version',
             'course__partner',
+            'restricted_run',
             Prefetch('seats', queryset=SeatSerializer.prefetch_queryset()),
         )
 
@@ -939,8 +941,8 @@ class MinimalCourseRunSerializer(FlexFieldsSerializerMixin, TimestampModelSerial
         model = CourseRun
         fields = ('key', 'uuid', 'title', 'external_key', 'image', 'short_description', 'marketing_url',
                   'seats', 'start', 'end', 'go_live_date', 'enrollment_start', 'enrollment_end', 'weeks_to_complete',
-                  'pacing_type', 'type', 'run_type', 'status', 'is_enrollable', 'is_marketable', 'term', 'availability',
-                  'variant_id')
+                  'pacing_type', 'type', 'restriction_type', 'run_type', 'status', 'is_enrollable', 'is_marketable',
+                  'term', 'availability', 'variant_id')
 
     def get_marketing_url(self, obj):
         include_archived = self.context.get('include_archived')
--- a/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
@@ -21,14 +21,14 @@ from course_discovery.apps.api.v1.exceptions import EditableAndQUnsupported
 from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase, OAuth2Mixin, SerializationMixin
 from course_discovery.apps.core.tests.factories import UserFactory
 from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
-from course_discovery.apps.course_metadata.choices import CourseRunStatus, ProgramStatus
-from course_discovery.apps.course_metadata.models import CourseRun, CourseRunType, Seat, SeatType
+from course_discovery.apps.course_metadata.choices import CourseRunRestrictionType, CourseRunStatus, ProgramStatus
+from course_discovery.apps.course_metadata.models import CourseRun, CourseRunType, RestrictedCourseRun, Seat, SeatType
 from course_discovery.apps.course_metadata.signals import (
     connect_course_data_modified_timestamp_signal_handlers, disconnect_course_data_modified_timestamp_signal_handlers
 )
 from course_discovery.apps.course_metadata.tests.factories import (
     CourseEditorFactory, CourseFactory, CourseRunFactory, CourseRunTypeFactory, CourseTypeFactory, OrganizationFactory,
-    PersonFactory, ProgramFactory, SeatFactory, SourceFactory, SubjectFactory, TrackFactory
+    PersonFactory, ProgramFactory, RestrictedCourseRunFactory, SeatFactory, SourceFactory, SubjectFactory, TrackFactory
 )
 from course_discovery.apps.course_metadata.toggles import (
     IS_COURSE_RUN_VARIANT_ID_EDITABLE, IS_SUBDIRECTORY_SLUG_FORMAT_ENABLED
@@ -182,6 +182,7 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
         }, format='json')
         assert response.status_code == 201
         new_course_run = CourseRun.everything.get(key=new_key)
+        assert RestrictedCourseRun.everything.count() == 0
         self.assertDictEqual(response.data, self.serialize_course_run(new_course_run))
         assert new_course_run.pacing_type == 'instructor_paced'
         # default we provide
@@ -266,11 +267,16 @@ class CourseRunViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mi
 
     @responses.activate
     def test_create_sets_additional_fields(self):
-        """ Verify that instructors, languages, min & max effort, and weeks to complete are set on a rerun. """
-        self.draft_course_run.staff.add(PersonFactory())
+        """
+            Verify that instructors, languages, min & max effort, and weeks to complete are set on a rerun.
+            Verify that the course run restriction is not copied to a rerun
+        """
         self.draft_course_run.transcript_languages.add(self.draft_course_run.language)
         self.draft_course_run.save()
-
+        RestrictedCourseRun.everything.create(
+            course_run=self.draft_course_run,
+            restriction_type=CourseRunRestrictionType.CustomB2BEnterprise.value
+        )
         # Create rerun based on draft course
         course = self.draft_course_run.course
         new_key = f'course-v1:{course.key_for_reruns}+1T2000'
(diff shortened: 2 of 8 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -932,14 +933,15 @@ def prefetch_queryset(cls, queryset=None):
         return queryset.select_related('course', 'type').prefetch_related(
             '_official_version',
             'course__partner',
+            'restricted_run',
```
DawoudSheraz, line 936: nit: I think we should rename this related_name for the field to restriction_type it will set correctly when accessing from course_run object.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -320,6 +324,40 @@ def test_create_draft_ignored(self, draft):
...
+
+        post_data = {
+            'course': course.key,
+            'start': '2000-01-01T00:00:00Z',
+            'end': '2001-01-01T00:00:00Z',
+            'run_type': str(self.course_run_type.uuid),
+        }
+
+        if is_restricted:
+            post_data['restriction_type'] = 'custom-b2c'
```
DawoudSheraz, line 351: instead of boolean, we can add the strings directly in ddt data (custom-b2b, custom-b2c,'')

course_discovery/apps/course_metadata/utils.py
```diff
@@ -102,6 +102,11 @@ def set_official_state(obj, model, attrs=None):
         obj.draft_version = draft_version
         if isinstance(obj, Course):
             obj.canonical_course_run = official_obj.canonical_course_run if official_obj else None
+        # If an object has a one-to-one field, it is necessary to change that field
+        # before officiating(saving) it here, else we'll get an IntegrityError
+        # because of two different objects "being" 1-1 connected with the same object
+        if isinstance(obj, RestrictedCourseRun):
+            obj.course_run = attrs['course_run']
```
DawoudSheraz, line 109: Cool, more draft/non-draft shenanigans, I guess.

DawoudSheraz:
Overall, the PR looks good. I have not had a chance to test it out on the local yet.

course_discovery/apps/api/serializers.py
zawan-ila: Then we will have to do something like `run.restriction_type.restriction_type` to get the restriction type. And I find that to be more confusing than the current `run.restricted_run.restriction_type`.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
zawan-ila: I don't want to set the `restriction_type` key at all. i.e not even set to '' or None. That's why I am using this approach.

course_discovery/apps/api/serializers.py
DawoudSheraz: Right. the related name can be `restricted` or `restriction`. Then, we can use `run.restriction.restriction_type`. That makes the relation access semantically correct.

zawan-ila: This definitely sounds good. I am not up for doing it here though. Maybe we can do that in followup work.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -266,11 +267,13 @@ def test_create_sets_canonical_course_run(self, has_canonical_run):
...
-        """ Verify that instructors, languages, min & max effort, and weeks to complete are set on a rerun. """
-        self.draft_course_run.staff.add(PersonFactory())
+        """ 
+            Verify that instructors, languages, min & max effort, and weeks to complete are set on a rerun.
+            Verify that the course run restriction is not copied to a rerun
+        """
         self.draft_course_run.transcript_languages.add(self.draft_course_run.language)
         self.draft_course_run.save()
-
+        RestrictedCourseRun.everything.create(course_run=self.draft_course_run, restriction_type='custom-b2c')
```
DawoudSheraz: nit: it would be good if we can use RestrictionType choices to reference the value, instead of repeating the string across various files.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2656,13 +2656,43 @@ def update_or_create_seats(self, run_type=None, prices=None, upgrade_deadline_ov
...
+    def update_or_create_restriction(self, restriction_type):
+        """
+        Updates or creates a CourseRunRestriction object for a draft course run
+        """
+
+        if restriction_type and restriction_type not in CourseRunRestrictionType.values:
+            raise Exception('Not a valid choice for restriction_type')
+
+        if restriction_type is None:
+            return
```
DawoudSheraz, line 2665: comment: Ideally, none/existence checks should be first in the method. Sequentially, it makes sense to stop the flow if the value is not there.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2656,13 +2656,43 @@ def update_or_create_seats(self, run_type=None, prices=None, upgrade_deadline_ov
...
+
+        if not restriction_type and hasattr(self, "restricted_run"):
+            self.restricted_run.delete()
+            official_obj = self.official_version
+            if hasattr(official_obj, "restricted_run"):
+                official_obj.restricted_run.delete()
+        elif restriction_type:
+            RestrictedCourseRun.everything.update_or_create(
```
DawoudSheraz, line 2676: q: do we need data_modified_timestamp related pre_save signal against this model?

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -902,6 +947,66 @@ def test_patch_published(self):
...
+        else:
+            assert RestrictedCourseRun.everything.count() == 0
+
+        # Publish the course run and verify that official versions
+        # of RestrictedCourseRuns are created
+        self.draft_course_run.status = CourseRunStatus.InternalReview
+        self.draft_course_run.save()
+        assert CourseRun.objects.filter(key=self.draft_course_run.key, draft=False).count() == 0
+        response = self.client.patch(url, {'status': 'reviewed'}, format='json')
+        assert response.status_code == 200, f"Status {response.status_code}: {response.content}"
```
DawoudSheraz: Generally speaking, this unit test is way too bulky and has many conditionals in it. It makes it difficult to figure out the behavior. It would be good to split this test into two tests (Whilst maintaining ddt behavior) to capture the behavior in a clean way.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -784,6 +799,12 @@ def test_create_makes_course_and_course_run(self):
         assert course_run.draft
         assert course_run.course == course
 
+        # Check Restricted Run
```
DawoudSheraz: nit: no need for this comment, the code is obvious here.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
AfaqShuaib09: +1 here
You can use `CourseRunRestrictionType.CustomB2BEnterprise.value`

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -526,6 +564,7 @@ def test_partial_update(self):
             'max_effort': expected_max_effort,
             'min_effort': expected_min_effort,
             'variant_id': variant_id,
+            'restriction_type': 'custom-b2b-enterprise'
```
AfaqShuaib09: here you can also refer this from `CourseRunRestrictionType` choices

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
```diff
@@ -725,9 +769,10 @@ def test_patch_put_restrict_when_reviewing(self, status):
             'start': self.draft_course_run.start,  # required, so we need for a put
             'end': self.draft_course_run.end,  # required, so we need for a put
             'run_type': str(self.draft_course_run.type.uuid),  # required, so we need for a put
+            'restriction_type': 'custom-b2c',
         }, format='json')
         assert response.status_code == 403
-
+        assert RestrictedCourseRun.everything.count() == 0
```
AfaqShuaib09, line 778: why `RestrictedCourseRun` count is 0 here?

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -769,12 +769,27 @@ def test_create_makes_editor(self):
...
+    @ddt.data(
+            ({'restriction_type': 'custom-b2c'}, True),
+            ({}, False),
+    )
+    @ddt.unpack
+    def test_create_makes_course_and_course_run(self, restriction_data, is_run_restricted):
         """
         When creating a course and supplying a course_run, it should create both the course
         and course run as drafts. When mode = 'audit', an audit seat should also be created.
         """
```
AfaqShuaib09, line 784: nit: update the test docstring to explain the `is_run_restricted`

AfaqShuaib09 (commented):
Overall, looks good to me, really nice work on this 👏🏻 
Make sure to replace the hardcoded `restriction_type` strings with `CourseRunRestrictionType` choices.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
AfaqShuaib09: Ah, just check the test name, ignore this

course_discovery/apps/course_metadata/models.py
zawan-ila: I am not sure. Mostly(exclusively?) we would want this to be set during creation so later modifications are not expected. Hence, this is effectively a "read only field". Thus, I decided against adding this.

I am happy to add this if you think it is important/useful.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
zawan-ila: The test is indeed bulky. I sorta like it though. I had a quick look above and below this, and found multiple similarly bulky tests. I have added a docstring to make the intent of the test clear. Let me know if it is still confusing.

DawoudSheraz: Even if there are existing bulky tests, that does not mean we should follow the convention. Ideally, the slimmer the tests, the better.

AfaqShuaib09 (approved):
Just a suggestion
You can also use the `CourseRunRestrictionType` choice values in test-cases assertions.

course_discovery/apps/api/v1/tests/test_views/test_course_runs.py
zawan-ila: > Even if there are existing bulky tests, that does not mean we should follow the convention

What I wanted to hint at is that perhaps it's not "bad" to have somewhat large tests. 

Splitting this out into multiple tests will need setting the state up (draft and non-draft versions of the restrictions etc.) in each of these and would lead to 3-4 tests instead of 1. I am unconvinced that trying to parse those 3-4 tests instead of this 1 (especially after the docstring update) would be any less mentally taxing than it is currently.

DawoudSheraz: Hmmm. While I still think the test is doing much more than it should be, the draft/non-draft weirdness can be a vouch for having this bulky test. Ideally, if we plan to test a behavior end-to-end, it is better to have segmented/separate unit tests for each part of interaction/behavior.

DawoudSheraz (approved):

PR #4332, opened 2024-04-19
build: fix ci-workflow migration issue

AfaqShuaib09:
`python-xmlsec`'s current version 1.3.14 is causing some issues during the dependency installation stage of `mysql8-migrations-check`. This PR is to pin it to 1.3.13 for the time being until the issue can be resolved.

https://github.com/xmlsec/python-xmlsec/issues/314

Thread Link:
https://twou.slack.com/archives/C030CC8T40N/p1713436115873339

### DIFF
--- a/.github/workflows/migrations-check-mysql8.yml
+++ b/.github/workflows/migrations-check-mysql8.yml
@@ -47,6 +47,8 @@ jobs:
         lsb_release -a
         mysql -V
 
+    # pinning xmlsec to version 1.3.13 to avoid the CI error, migration checks are failing due to an issue in the latest release of python-xmlsec
+    # https://github.com/xmlsec/python-xmlsec/issues/314
     - name: Install Python dependencies
       run: |
         pip install -r requirements/pip_tools.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/migrations-check-mysql8.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         pip uninstall -y mysqlclient
         pip install --no-binary mysqlclient mysqlclient
         pip uninstall -y xmlsec
-        pip install --no-binary xmlsec xmlsec
+        pip install --no-binary xmlsec xmlsec==1.3.13
```
DawoudSheraz, line 59: nit: we should add the comment here to that it is clear instantly why this is pinned.

AfaqShuaib09: done

DawoudSheraz: nit: move the comment above, to follow the conventions present in the codebase

DawoudSheraz (approved):
commit type should be build, not fix.

zawan-ila (approved):

PR #4343, opened 2024-04-30
style: add padding to the attributes of inner tags of table

AfaqShuaib09:
This PR adds the padding to the table data `<td>` tag of the table in the email_ingestion template.
image


zawan-ila (approved):

DawoudSheraz (approved):
nit: could have used css class (defined in template) but ok for now.

PR #4385, opened 2024-07-09
feat: add CRUD operations for taxi form

AfaqShuaib09:
[PROD-4058](https://2u-internal.atlassian.net/browse/PROD-4058)
---------
This pull request adds the taxi form to the `additional_metadata` serializer and manages the creation and updating of the `taxi_form`. Additionally, it includes the `taxi_form` field in the `discovery` admin.

image
---
image

### DIFF
--- a/course_discovery/apps/api/serializers.py
+++ b/course_discovery/apps/api/serializers.py
@@ -688,6 +688,13 @@ class ProductMetaSerializer(TaggitSerializer, BaseModelSerializer):
         fields = ('title', 'description', 'keywords')
 
 
+class TaxiFormSerializer(BaseModelSerializer):
+    """Serializer for the ``TaxiForm`` model."""
+    class Meta:
+        model = TaxiForm
+        fields = ('form_id', 'grouping', 'title', 'subtitle', 'post_submit_url')
+
+
 class AdditionalMetadataSerializer(BaseModelSerializer):
     """Serializer for the ``AdditionalMetadata`` model."""
 
@@ -698,10 +705,11 @@ class AdditionalMetadataSerializer(BaseModelSerializer):
     end_date = serializers.DateTimeField()
     registration_deadline = serializers.DateTimeField(allow_null=True)
     variant_id = serializers.UUIDField(allow_null=True)
+    taxi_form = TaxiFormSerializer(required=False, allow_null=True)
 
     @classmethod
     def prefetch_queryset(cls):
-        return AdditionalMetadata.objects.select_related('facts', 'certificate_info', 'product_meta')
+        return AdditionalMetadata.objects.select_related('facts', 'certificate_info', 'product_meta', 'taxi_form')
 
     class Meta:
         model = AdditionalMetadata
@@ -709,8 +717,19 @@ class AdditionalMetadataSerializer(BaseModelSerializer):
             'external_identifier', 'external_url', 'lead_capture_form_url',
             'facts', 'certificate_info', 'organic_url', 'start_date', 'end_date',
             'registration_deadline', 'variant_id', 'course_term_override', 'product_status',
-            'product_meta', 'external_course_marketing_type', 'display_on_org_page',
+            'product_meta', 'external_course_marketing_type', 'display_on_org_page', 'taxi_form',
         )
+
+    def update_taxi_form(self, instance, taxi_form):
+        if instance.taxi_form:
+            if not taxi_form:
+                instance.taxi_form = None
+            else:
+                TaxiForm.objects.filter(
+                    id=instance.taxi_form.id
+                ).update(**taxi_form)
+        elif taxi_form:
+            instance.taxi_form = TaxiForm.objects.create(**taxi_form)
 
     def _update_product_meta(self, instance, product_meta):
         if instance.product_meta:
@@ -721,22 +740,32 @@ class AdditionalMetadataSerializer(BaseModelSerializer):
                 **product_meta)
 
     def update(self, instance, validated_data):
-        # Handle writing nested fields separately
-        if 'product_meta' in validated_data:
+        """
+        Update the AdditionalMetadata instance with the validated data along with handling
+        of the nested fields separately.
+        """
+        def handle_product_meta(instance, validated_data):
             # Handle product metadata only for 2U Executive Education courses else just pop
-            product_meta_data = validated_data.pop(
-                'product_meta')
-            if instance:
-                self._update_product_meta(
-                    instance, product_meta_data)
+            if 'product_meta' in validated_data:
+                product_meta_data = validated_data.pop('product_meta')
+
+                if instance:
+                    self._update_product_meta(instance, product_meta_data)
+
+            return validated_data
+
+        def handle_taxi_form(instance, validated_data):
+            if 'taxi_form' in validated_data:
+                taxi_form_data = validated_data.pop('taxi_form')
+
+                if instance:
+                    self.update_taxi_form(instance, taxi_form_data)
+
+            return validated_data
+
+        validated_data = handle_product_meta(instance, validated_data)
+        validated_data = handle_taxi_form(instance, validated_data)
         return super().update(instance, validated_data)
-
-
-class TaxiFormSerializer(BaseModelSerializer):
-    """Serializer for the ``TaxiForm`` model."""
-    class Meta:
-        model = TaxiForm
-        fields = ('form_id', 'grouping', 'title', 'subtitle', 'post_submit_url')
 
 
 class DegreeAdditionalMetadataSerializer(BaseModelSerializer):
@@ -1488,12 +1517,24 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
 
         return instance.product_meta, changed
 
+    def update_taxi_form(self, instance, taxi_form):
+        changed = False
+        if instance.taxi_form:
+            _, changed = update_instance(instance.taxi_form, taxi_form, True)
+        elif taxi_form:
+            instance.taxi_form = TaxiForm.objects.create(**taxi_form)
+            instance.save()
+            changed = True
+
+        return instance.taxi_form, changed
+
     def update_additional_metadata(self, instance, additional_metadata):
 
         changed = False
         facts = additional_metadata.pop('facts', None)
         certificate_info = additional_metadata.pop('certificate_info', None)
         product_meta = additional_metadata.pop('product_meta', None)
+        taxi_form = additional_metadata.pop('taxi_form', None)
 
         if instance.additional_metadata:
             _, additional_metadata_changed = update_instance(instance.additional_metadata, additional_metadata, True)
@@ -1512,6 +1553,10 @@ class CourseSerializer(TaggitSerializer, MinimalCourseSerializer):
         if certificate_info:
             _, certification_info_changed = self.update_certificate_info(instance.additional_metadata, certificate_info)
             changed = changed or certification_info_changed
+
+        if taxi_form:
+            _, taxi_form_changed = self.update_taxi_form(instance.additional_metadata, taxi_form)
+            changed = changed or taxi_form_changed
 
         return changed
         # save() will be called by main update()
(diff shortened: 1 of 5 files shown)


course_discovery/apps/api/serializers.py
```diff
@@ -726,19 +741,17 @@ def update(self, instance, validated_data):
...
             product_meta_data = validated_data.pop(
                 'product_meta')
+
+        if 'taxi_form' in validated_data:
+            taxi_form_data = validated_data.pop('taxi_form')
+
             if instance:
                 self._update_product_meta(
                     instance, product_meta_data)
+                self.update_taxi_form(instance, taxi_form_data)
```
DawoudSheraz, line 751: Taxi form should not be handled in product_meta. Product_meta is for a specific DB model. Taxi form handling should be on same level as product_meta

course_discovery/apps/api/serializers.py
```diff
@@ -1488,12 +1501,21 @@ def update_product_meta(self, instance, product_meta_data):
...
+            instance.save()
+            return instance.taxi_form, False
+
     def update_additional_metadata(self, instance, additional_metadata):
 
         changed = False
         facts = additional_metadata.pop('facts', None)
         certificate_info = additional_metadata.pop('certificate_info', None)
         product_meta = additional_metadata.pop('product_meta', None)
+        taxi_form = additional_metadata.pop('taxi_form', None)
```
DawoudSheraz, line 1537: what fields will be coming against taxi_form dict? If talking about ExecEd. the only relevant fields are uuid and post_submit URL. I think we can default to using course title if the title field is not present in request POST. subtitle and grouping would likely be irrelevant altogether.

course_discovery/apps/api/serializers.py
```diff
@@ -1513,6 +1535,10 @@ def update_additional_metadata(self, instance, additional_metadata):
             _, certification_info_changed = self.update_certificate_info(instance.additional_metadata, certificate_info)
             changed = changed or certification_info_changed
 
+        if taxi_form:
+            _, taxi_form_changed = self.update_taxi_form(instance.additional_metadata, taxi_form)
+            changed = changed or taxi_form_changed
+
```
DawoudSheraz, line 1560: add a unit test to verify the taxi form CRUD handling

course_discovery/apps/api/serializers.py
```diff
@@ -1488,12 +1501,21 @@ def update_product_meta(self, instance, product_meta_data):
 
         return instance.product_meta, changed
 
+    def update_taxi_form(self, instance, taxi_form):
+        if instance.taxi_form:
+            return update_instance(instance.taxi_form, taxi_form, True)
+        else:
+            instance.taxi_form = TaxiForm.objects.create(**taxi_form)
+            instance.save()
+            return instance.taxi_form, False
```
DawoudSheraz: how will this handle the cases where we have taxiform data before and then we remove it?

AfaqShuaib09: this case will be handled by below condition, if `instance.taxi_form` exists, I will update it using values obtained from the API by passing the existing instance taxi_form (which can be null)

```python
if instance.taxi_form:
            return update_instance(instance.taxi_form, taxi_form, True)
```

course_discovery/apps/api/serializers.py
AfaqShuaib09: sure

course_discovery/apps/api/serializers.py
AfaqShuaib09: image
These fields are coming in `taxi_form` dict.

> I think we can default to using course title if the title field is not present in request POST. subtitle and grouping would likely be irrelevant altogether.

Yeah, we can handle this logic in the create and update methods of the `viewset` instead of implementing it directly in the serializer. Currently, I'm waiting for the representation format of taxi_form from GEAG API so that we can make a decision on this. Alternatively, we could create a separate `minimal serializer` for taxi_form for this purpose.

course_discovery/apps/api/serializers.py
```diff
@@ -1488,12 +1516,32 @@ def update_product_meta(self, instance, product_meta_data):
...
+                changed = True
+            else:
+                _, changed = update_instance(instance.taxi_form, taxi_form, True)
+        else:
+            if taxi_form:
+                instance.taxi_form = TaxiForm.objects.create(**taxi_form)
+                instance.save()
+                changed = True
+            else:
+                changed = False
```
DawoudSheraz: you don't need this else here. changed has False by default and this else block is just doing that.

course_discovery/apps/api/serializers.py
```diff
@@ -1488,12 +1516,32 @@ def update_product_meta(self, instance, product_meta_data):
...
+
+        return instance.taxi_form, changed
+
     def update_additional_metadata(self, instance, additional_metadata):
 
         changed = False
         facts = additional_metadata.pop('facts', None)
         certificate_info = additional_metadata.pop('certificate_info', None)
         product_meta = additional_metadata.pop('product_meta', None)
+        taxi_form = additional_metadata.pop('taxi_form', 'None')
```
DawoudSheraz: Why is None a string?

course_discovery/apps/api/serializers.py
```diff
@@ -1513,6 +1561,10 @@ def update_additional_metadata(self, instance, additional_metadata):
             _, certification_info_changed = self.update_certificate_info(instance.additional_metadata, certificate_info)
             changed = changed or certification_info_changed
 
+        if taxi_form != 'None':
```
DawoudSheraz: if taxi_form is None and not 'None', then `if taxi_form` should do the trick.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1632,6 +1634,153 @@ def test_update_product_meta_with_additional_metadata(self):
         pre_save.disconnect(data_modified_timestamp_update, ProductMeta)
         m2m_changed.disconnect(product_meta_taggable_changed, ProductMeta.keywords.through)
 
+    @responses.activate
+    def test_update_taxi_form_with_additional_metadata(self):
+        """ Verify that the taxi_form is updated when additional_metadata is updated. """
+        self.maxDiff = None
```
DawoudSheraz: you dont need maxDiff = None here.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1632,6 +1634,153 @@ def test_update_product_meta_with_additional_metadata(self):
...
+            'display_on_org_page': True,
+            'taxi_form': taxi_form,
+        }
+
+        url = reverse('api:v1:course-detail', kwargs={'key': course.uuid})
+        response = self.client.patch(url, {'additional_metadata': additional_metadata}, format='json')
+        assert response.status_code == 200
+
+        # The TaxiForm objects are being created 2 times. Once when the CourseViewSetTests create a Factory Course
+        # and once while running this testcase.
```
DawoudSheraz: set Factory value as None explicitly in the test to avoid creating the extra object.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1632,6 +1634,153 @@ def test_update_product_meta_with_additional_metadata(self):
...
+        )
+        additional_metadata['taxi_form'] = taxi_form
+        assert response.status_code == 200
+        course = Course.everything.get(uuid=course.uuid, draft=True)
+        assert previous_data_modified_timestamp == course.data_modified_timestamp
+
+        pre_save.disconnect(data_modified_timestamp_update, TaxiForm)
+
+    @responses.activate
+    def test_update_taxi_form_with_additional_metadata_and_then_make_it_none(self):
```
DawoudSheraz: the test name is not intuitive. Something along the lines of. test_update_additional_metadata__taxi_form_removal will be better.

course_discovery/apps/course_metadata/models.py
```diff
@@ -880,7 +880,7 @@ class TaxiForm(ManageHistoryMixin, TimeStampedModel):
     def has_changed(self):
         if not self.pk:
             return False
-        return self.has_model_changed()
+        return self.field_tracker.changed()
```
DawoudSheraz: has_model_changed should be taking care of field_Tracker.changed(), we don't need to call it explicitly.

course_discovery/apps/api/serializers.py
AfaqShuaib09: There are two scenarios, either we include `taxi_form` in the response to update the data, or we do not include it at all. If taxi_form is absent during the update call, I will set it to `"None"`, which means the existing value will remain unchanged. Otherwise it will update existing value to none.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
AfaqShuaib09: Added it for the sake of testing, removing it now.

course_discovery/apps/api/serializers.py
DawoudSheraz: ok. We can use None instead of "None" to signify missing/non-existent values. String none feels out of place.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
AfaqShuaib09: image

Factory value is already null in this testcase.

- One taxi-form is created during the course creation (along with additional metadata) in the setUp method of the test suite.
- Another one is created during this test case (during creation api call)

I'm hesitant to pass the None value for `additional_metadata` as it is used in various test cases.

DawoudSheraz: > One taxi-form is created during the course creation (along with additional metadata) in the setUp method of the test suite. 

You can do it in setUp.

course_discovery/apps/api/serializers.py
AfaqShuaib09: What if we actually want a null value for taxi_form removal?

DawoudSheraz: handle that usecase in the specific testcase.

AfaqShuaib09: Actually, we need to handle two different states for this scenario

- If taxi_form (with a null value) is present in the API, we want to remove it from the Course's additional metadata.
- If taxi_form is not present in the API (e.g., in a patch request), we do not want it to be null because it is not present in the API. 

To handle this, I've set its value to the string "None" if `taxi_form` is not present in the request, so we can differentiate b/w both use cases.

DawoudSheraz: - taxi_form is going to be dict, not a single value argument
- I don't see why taxiform is that special. We handle dict for other cases, like product_meta. Are those working as expected?

AfaqShuaib09: > taxi_form is going to be dict, not a single value argument

Okay, if that is the case for removal as well, then it's fine. I was assuming it would be null in case of removal.
updating it

Ali-D-Akbar (approved):

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -1632,6 +1634,152 @@ def test_update_product_meta_with_additional_metadata(self):
...
+        additional_metadata['taxi_form'] = taxi_form
+        assert response.status_code == 200
+        course = Course.everything.get(uuid=course.uuid, draft=True)
+        assert previous_data_modified_timestamp == course.data_modified_timestamp
+
+        pre_save.disconnect(data_modified_timestamp_update, TaxiForm)
+
+    @responses.activate
+    def test_update_additional_metadata__taxi_form_removal(self):
+        """ Verify that the taxi_form is updated and made None when additional_metadata is None in the API response """
```
DawoudSheraz: The docstring needs an update. The test is only checking taxi_form None, not entire additional_metadata dict.

DawoudSheraz (approved):

PR #4389, opened 2024-07-18
feat: Skipping price/couse updates in Course Entitlements for restricted runs

AfaqShuaib09:
[PROD-4116](https://2u-internal.atlassian.net/browse/PROD-4116)
--------
This PR handles the Skipping of price/course updates in Course Entitlements for restricted runs (specifically b2b-custom presentation) to resolve a pricing discrepancy issue.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -12,7 +12,7 @@ from django.urls import reverse
 
 from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.choices import (
-    CourseRunStatus, ExternalCourseMarketingType, ExternalProductStatus
+    CourseRunRestrictionType, CourseRunStatus, ExternalCourseMarketingType, ExternalProductStatus
 )
 from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoader
 from course_discovery.apps.course_metadata.data_loaders.constants import (
@@ -525,6 +525,11 @@ class CSVDataLoader(AbstractDataLoader):
         Create and return the request data for making a patch call to update the course.
         """
         collaborator_uuids = self.process_collaborators(data.get('collaborators', ''), course.key)
+        price = (
+            self.get_pricing_representation(data['verified_price'], course.type)
+            if data.get('restriction_type', 'None') != CourseRunRestrictionType.CustomB2BEnterprise.value else {}
+        )
+
         subjects = self.get_subject_slugs(
             data.get('primary_subject'),
             data.get('secondary_subject'),
@@ -539,7 +544,7 @@ class CSVDataLoader(AbstractDataLoader):
             'type': str(course.type.uuid),
             'subjects': subjects,
             'collaborators': collaborator_uuids,
-            'prices': self.get_pricing_representation(data['verified_price'], course.type),
+            'prices': price,
 
             'title': data['title'],
             'syllabus_raw': data.get('syllabus', ''),
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
@@ -821,6 +821,55 @@ class TestCSVDataLoader(CSVLoaderMixin, OAuth2Mixin, APITestCase):
                     assert course.subjects.first().slug == "computer-science"
                     assert course_run.staff.exists() is False
 
+    @data(True, False)
+    @responses.activate
+    def test_entitlement_price_update_for_custom_presentation(self, reverse_order, jwt_decode_patch):  # pylint: disable=unused-argument
+        """
+        Verify that the loader does not update price for custom-b2b-enterprise in Course's Entitlement.
+        """
+        self._setup_prerequisites(self.partner)
+        self.mock_ecommerce_publication(self.partner)
+        self.mock_studio_calls(self.partner)
+        self.mock_image_response()
+        csv_key_order = list(self.CSV_DATA_KEYS_ORDER)
+        csv_key_order.append('restriction_type')
+
+        csv_data = [
+            {
+                **mock_data.VALID_COURSE_AND_COURSE_RUN_CSV_DICT,
+                "restriction_type": None,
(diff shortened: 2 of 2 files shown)


course_discovery/apps/api/v1/views/courses.py
```diff
@@ -327,6 +327,7 @@ def update_course(self, data, partial=False):  # pylint: disable=too-many-statem
         org_logo_override_image = data.pop('organization_logo_override', None)
         video_data = data.pop('video', None)
         url_slug = data.pop('url_slug', '')
+        is_restricted = data.pop('is_restricted', False)
```
DawoudSheraz: Hmm, it does not look correct semantically. The restriction is on course run level. We are handling this on course level. One way can be to handle when course API is called from csv loader or publisher i.e. don't pass price information from restricted run. 
Also, this looks way too broad to check. We can't rely on the argument alone to determine if the call is coming from restricted run.

AfaqShuaib09: Fixed. I've refactored the logic

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -525,6 +525,11 @@ def _update_course_api_request_data(self, data, course, is_draft):
         Create and return the request data for making a patch call to update the course.
         """
         collaborator_uuids = self.process_collaborators(data.get('collaborators', ''), course.key)
+        price = (
+            self.get_pricing_representation(data['verified_price'], course.type)
+            if data.get('restriction_type', 'None') != 'custom-b2b-enterprise' else None
+        )
+
```
DawoudSheraz: If price is None, it can potentially remove the existing entitlements (in course API). Please verify this.

AfaqShuaib09: Already verified this here with this unittest
https://github.com/openedx/course-discovery/pull/4389/files#diff-ddc39d69fa6624e59c34cbdcaf2057b91ef62eaca715d30f1f94f4302ca50a1cR824-R868

Running the command locally to double-check this.

AfaqShuaib09: Verified on local as well

image


- CSV Price dummy entry - 1st one is of public run and 2nd one is of private restricted run

image

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -525,6 +525,11 @@ def _update_course_api_request_data(self, data, course, is_draft):
         Create and return the request data for making a patch call to update the course.
         """
         collaborator_uuids = self.process_collaborators(data.get('collaborators', ''), course.key)
+        price = (
+            self.get_pricing_representation(data['verified_price'], course.type)
+            if data.get('restriction_type', 'None') != 'custom-b2b-enterprise' else None
```
DawoudSheraz: nit: instead of using string `custom-b2b-enterprise`, you can import RestrictionType choices and reference B2B here.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -821,6 +821,51 @@ def test_ingest_flow_for_minimal_course_data(self, jwt_decode_patch):  # pylint:
...
+                    {
+                        **mock_data.VALID_COURSE_AND_COURSE_RUN_CSV_DICT,
+                        "restriction_type": None,
+                    },
+                    {
+                        **mock_data.VALID_COURSE_AND_COURSE_RUN_CSV_DICT,
+                        "restriction_type": "custom-b2b-enterprise",
+                        "verified_price": "250",
+                        "variant_id": "11111111-1111-1111-1111-111111111111",
+                        "start_date": "04/25/2020",
```
zawan-ila: [curious]: Why do we need the start_date here?

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -821,6 +821,51 @@ def test_ingest_flow_for_minimal_course_data(self, jwt_decode_patch):  # pylint:
...
+        self.mock_studio_calls(self.partner)
+        self.mock_image_response()
+        csv_key_order = list(self.CSV_DATA_KEYS_ORDER)
+        csv_key_order.append('restriction_type')
+
+        with NamedTemporaryFile() as csv:
+            csv = self._write_csv(
+                csv,
+                [
+                    {
```
zawan-ila: Can you also add a test where the order of data is reversed? i.e the restricted run comes first and non-restricted second. You can parametrize this test and depending on parameter, reverse the list of data.

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
AfaqShuaib09: nothing special, just for the sake of testing. I'll remove it from here

zawan-ila: No need to remove. I just wanted to see if I was missing something.

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
AfaqShuaib09: done, added

zawan-ila: I do not think that calling the course update API with a None price field "really works". In particular, I would expect [this](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L343-L347) to throw and the call to not return a 200. 

To test this^, I modified this test to the attached version, and it started failing. Could you please confirm if this is an issue, or the test failure is indeed expected :-)

[temp.patch](https://github.com/user-attachments/files/16366438/temp.patch)

AfaqShuaib09: Nice catch, figured out the issue.
Replacing `None` with `{}` resolved the issue

zawan-ila (approved):

PR #4394, opened 2024-07-24
fix: increase the default throttle rate for enterprise users (POC)

adamstankiewicz:
The `frontend-app-learner-portal-enterprise` MFE makes requests to several API endpoints within `course-discovery`, e.g. to serve metadata about courses, programs, course recommendations, and course reviews. However, enterprise users are frequently rate limited by the default throttle rate of `100/hour` (i.e., N = 2.45k requests rate limited in the past 1 week from this MFE).

This PR proposes a (POC) solution for increasing the default throttle rate if the authenticated user is an enterprise user, as determined by the authenticated user's `roles` in their decoded JWT cookie.

### DIFF
--- a/course_discovery/apps/core/throttles.py
+++ b/course_discovery/apps/core/throttles.py
@@ -49,7 +49,7 @@ class OverridableUserRateThrottle(UserRateThrottle):
 
                 # If the user is not a privileged user, increase throttling rate if they are an enterprise user
                 if is_enterprise_user(request):
-                    self.rate = '300/hour'
+                    self.rate = '600/hour'
 
         self.num_requests, self.duration = self.parse_rate(self.rate)
 
--- a/course_discovery/apps/core/throttles.py
+++ b/course_discovery/apps/core/throttles.py
@@ -1,5 +1,6 @@
 """Custom API throttles."""
 from django.core.cache import InvalidCacheBackendError, caches
+from edx_rest_framework_extensions.auth.jwt.decoder import configured_jwt_decode_handler
 from rest_framework.throttling import UserRateThrottle
 
 from course_discovery.apps.core.models import UserThrottleRate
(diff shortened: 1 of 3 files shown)


course_discovery/apps/core/throttles.py
```diff
@@ -28,10 +42,15 @@ def allow_request(self, request, view):
...
                 self.rate = user_throttle.rate
-                self.num_requests, self.duration = self.parse_rate(self.rate)
             except UserThrottleRate.DoesNotExist:
                 # If we don't have a custom user override, skip throttling if they are a privileged user
                 if user.is_superuser or user.is_staff or is_publisher_user(user):
                     return True
 
+                # If the user is not a privileged user, increase throttling rate if they are an enterprise user
+                if is_enterprise_user(request):
+                    self.rate = '300/hour'
```
adamstankiewicz: [inform/context] This specific rate is still up in the air; it might need to be higher than this (e.g., `600/hour`) for enterprise users. This POC PR intends to demonstrate the approach, rather than a specific rate at the moment.

course_discovery/apps/core/throttles.py
```diff
@@ -28,10 +42,15 @@ def allow_request(self, request, view):
                 # Override this throttle's rate if applicable
                 user_throttle = UserThrottleRate.objects.get(user=user)
                 self.rate = user_throttle.rate
-                self.num_requests, self.duration = self.parse_rate(self.rate)
             except UserThrottleRate.DoesNotExist:
                 # If we don't have a custom user override, skip throttling if they are a privileged user
                 if user.is_superuser or user.is_staff or is_publisher_user(user):
```
adamstankiewicz, line 47: [inform/context] We already remove the throttling based on whether the authenticated user is a "publisher user" (aka is a member of a Django `Group`). This POC is largely extending this to account for an `is_enterprise_user` check as well.

DawoudSheraz:
This has been added as part of https://github.com/openedx/course-discovery/pull/4506, closing this PR.

PR #4395, opened 2024-07-24
docs: add ADR for fixed USD pricing

zawan-ila:
### [PROD-4124](https://2u-internal.atlassian.net/browse/PROD-4124)

ADR for fixed USD pricing in discovery.

### DIFF
--- a/docs/decisions/0028-fixed-usd-pricing.rst
+++ b/docs/decisions/0028-fixed-usd-pricing.rst
@@ -1,0 +1,48 @@
+28. Fixed USD Pricing
+========================
+
+Status
+--------
+Accepted (Jul 2024)
+
+Context
+---------
+Course Discovery is equipped with the capability to ingest and update courses from external sources (through the csv_loader).
+These external courses may be priced in various currencies, necessitating conversion to a standardized currency (USD) before
+their introduction into discovery. Given the volatility of exchange rates, the product prices coming into discovery may fluctuate with
+each data ingestion.
+
+The regular fluctuation in prices poses challenges for customers, especially those purchasing courses in bulk (B2B). These fluctuations can lead to
+situations where a customer plans to make a purchase within budget, only to find that prices have changed unfavorably when
+they proceed with the transaction. This inconsistency may disrupt budgetary planning and affect decision-making processes for
+customers. This is particularly problematic for B2B customers, who may have predefined contracts and can not exceed their budget constraints. 
+
+Decision
+----------
+A new field, fixed_price_usd, will be introduced to the course run model within Course Discovery. This field will be accessible
+through the standard course and search APIs. Consumers will have the ability to retrieve and utilize this field from Discovery APIs,
+enabling them to price their offerings based on this stable fixed price in USD. It is assumed that the fixed price will remain
+relatively stable over time, minimizing the impact of currency fluctuations for B2B customers.
+
+The existing pricing fields and their associated logic will remain unchanged within Course Discovery. It will be the responsibility
+of consumers, such as enterprise, to appropriately manage any potential implications of using the fixed price field within their own
+systems and processes.
+
+Consequences
+--------------
+- The current Pricing flow in Discovery will remain as is.
+- There will be no impact on Seats, Entitlements, and their associated logic.
+- The current flow of propagation of pricing information to Ecommerce will remain as is.
+- Analytics/Data teams will need to ensure that they track Enterprise Customers' purchases against the fixed price appropriately.
+
+Alternatives Considered
+-------------------------
+
+- It would have been preferable to add this field at the Seat level somehow, and not clutter the CourseRun model unnecessarily.
+  However, due to some historical reasons, Executive Education courses (for which we need this functionality), only have unpaid
+  seats. It would be confusing to add pricing information to an unpaid seat.
+
+- The field could have been integrated at the course level, aligning with the current discovery flows where the course price (CourseEntitlement)
+  is applied to all course runs. However, this approach does not accommodate pricing at the course run level, which has been a significant
+  issue for discovery users in the past. Although there are some workarounds, such as manually adjusting prices in the database, these solutions
+  are both cumbersome and inefficient.


docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+Status
+--------
+Accepted (Jul 2024)
+
+Context
+---------
+Course Discovery is equipped with the capability to ingest and update courses from external sources (through the csv_loader).
+These external courses may be priced in various currencies, necessitating conversion to a standardized currency (USD) before
+their introduction into discovery. Given the volatility of exchange rates, the prices coming into discovery may fluctuate with
+each data ingestion.
```
DawoudSheraz: nit: the prices of same products coming into discovery

docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+
+The regular fluctuation in prices poses challenges for enterprises purchasing courses in bulk. These fluctuations can lead to
+situations where an enterprise plans to make a purchase within budget, only to find that prices have changed unfavorably when
+they proceed with the transaction. This inconsistency may disrupt budgetary planning and affect decision-making processes for
+enterprise customers.
```
DawoudSheraz: rather than making it all about enterprise, we can use B2B terminology here. Furthermore, we can also add the impact on downstream reporting because of price changes with every ingestion. The current context is a bit too focused on Enterprise. 
We should also focus on that because of predefined contracts, the price change is much more critical for businesses. This will also help in explanation of the solution of adding new field instead of updating the existing prices.

docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Decision
+----------
+A new field, fixed_price_usd, will be introduced to the course run model within Course Discovery. This field will be accessible
+through our standard course and search APIs. Enterprises will have the ability to retrieve and utilize this field from our APIs,
```
DawoudSheraz: nit: Enterprises --> Consumers

DawoudSheraz: from our APIs --> from discovery APIs.

docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+- The current flow of propagation of pricing information to Ecommerce will remain as is.
+- Analytics/Data teams will need to ensure that they track Enterprise Customers' purchases against the fixed price appropriately.
+
+
+Alternatives Considered
+-------------------------
+
+It would have been preferable to add this field at the Seat level somehow, and not clutter the CourseRun model unnecessarily.
+However, due to some historical reasons, Executive Education courses (for which we need this functionality), only have unpaid
+seats. It would be confusing to add pricing information to an unpaid seat.
```
DawoudSheraz: We should also mention that in discovery, the pricing SoT is entitlement. Having a separate price field in Seat for CR would have been confusing and would have a major impact on the reporting.

docs/decisions/0028-fixed-usd-pricing.rst
zawan-ila: I have changed it to `product prices` as I feel it's better that way. Let me know if you think otherwise.

docs/decisions/0028-fixed-usd-pricing.rst
zawan-ila: Tried to use B2B terminology. Also added that this is more critical for business customers.

I am not sure about the impacts on reporting part. I have not seen it being discussed as an issue in any of the docs I've come across. Could you provide more details here?

docs/decisions/0028-fixed-usd-pricing.rst
zawan-ila: I don't follow the impacts on reporting here. Yes, in discovery the "normal" flow is to push entitlement prices to seats. However, I recall having seen PCs create course runs with different prices somehow 🤔 . I have not heard of that having a negative impact on the reporting.

DawoudSheraz: it would be better to verify this then.

docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+Decision
+----------
+A new field, fixed_price_usd, will be introduced to the course run model within Course Discovery. This field will be accessible
+through the standard course and search APIs. Consumers will have the ability to retrieve and utilize this field from Discovery APIs,
+enabling them to price their offerings based on this stable fixed price in USD. It is assumed that the fixed price will remain
+relatively stable over time, minimizing the impact of currency fluctuations for B2B customers.
+
+The existing pricing fields and their associated logic will remain unchanged within Course Discovery. It will be the responsibility
+of consumers, such as enterprise, to appropriately manage any potential implications of using the fixed price field within their own
+systems and processes.
```
Ali-D-Akbar, line 29: is it worth mentioning that this field is specific to course runs?

Ali-D-Akbar (approved):
Looks good to me overall.

docs/decisions/0028-fixed-usd-pricing.rst
DawoudSheraz: Adding to that, given the pricing SoT is Entitlement, we might want to add here why the field is not on Course level.

docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Consequences
+--------------
+- Current Pricing flow in Discovery will remain as is.
```
DawoudSheraz: nit: The current pricing flow

docs/decisions/0028-fixed-usd-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+- The current flow of propagation of pricing information to Ecommerce will remain as is.
+- Analytics/Data teams will need to ensure that they track Enterprise Customers' purchases against the fixed price appropriately.
+
+
+Alternatives Considered
+-------------------------
+
+It would have been preferable to add this field at the Seat level somehow, and not clutter the CourseRun model unnecessarily.
+However, due to some historical reasons, Executive Education courses (for which we need this functionality), only have unpaid
+seats. It would be confusing to add pricing information to an unpaid seat.
```
DawoudSheraz: Many parts of this information are 2U specific. We can re-word to explain the implication of seat level field and its impact on downstream consumers (LMS, Studio, E-comm) who all would need to cater to the field.

DawoudSheraz (commented):
I have left some small followup comments. Overall, this is good.

docs/decisions/0028-fixed-usd-pricing.rst
zawan-ila: [confused] The alternative I'm discussing here is to add a separate `fixed_price_usd` field at the seat level. The sync with different services for that field should not make it any worse as even our proposed solution has no sync flows.

docs/decisions/0028-fixed-usd-pricing.rst
zawan-ila: > is it worth mentioning that this field is specific to course runs?

At line 22, I have mentioned that the field will be added to CourseRun model.

> we might want to add here why the field is not on Course level

Can you suggest a sentence or two here? My understanding is that in the "SoT is Entitlement" model, it surely could have been. It is just that our business needs necessitate run level "pricing". I don't understand why the SoT was chosen to be the Course level in the first place.

DawoudSheraz: >  it surely could have been. It is just that our business needs necessitate run level "pricing".

Rather than saying "our" business needs, we can mention that ability to have prices different on course runs was not directly allowed and we needed to use coupons in e-commerce to change pricing for runs, etc. I understand not all context might be relevant here but considering how the price flow is currently, the information would be helpful in future for others learning more about this.

docs/decisions/0028-fixed-usd-pricing.rst
DawoudSheraz: > The sync with different services for that field should not make it any worse as even our proposed solution has no sync flows.

Its not just sync, its ensuring the consumers are using that field appropriately. Ecomm and LMS have their versions of Seat models and they might require changes if the disco seat model is changed.

zawan-ila: > consumers are using that field appropriately

Even the proposed approach leaves it to the consumers to pull the fixed_price information and make appropriate changes on their end. And does not force the fixed_price information upon anybody. We could have taken a similar approach where we add the fixed_price_information to Seat level but do not push it downstream.

docs/decisions/0028-fixed-usd-pricing.rst
zawan-ila: I have added some context to this effect^ in the Alternatives section.

DawoudSheraz (approved):

PR #4408, opened 2024-08-12
chore: update course_run_detail method in CourseSearchDocument serializer

AfaqShuaib09:
This PR adds `fixed_price_usd` field to `CourseSearchDocument` serializer

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes/serializers/course.py
+++ b/course_discovery/apps/course_metadata/search_indexes/serializers/course.py
@@ -74,7 +74,7 @@ class CourseSearchDocumentSerializer(ModelObjectDocumentSerializerMixin, DateTim
             'restriction_type': (
                 course_run.restricted_run.restriction_type if hasattr(course_run, 'restricted_run') else None
             ),
-            'fixed_price_usd': float(course_run.fixed_price_usd) if course_run.fixed_price_usd else None,
+            'fixed_price_usd': course_run.fixed_price_usd,
         }
         if detail_fields:
             course_run_detail.update(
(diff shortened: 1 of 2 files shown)


zawan-ila (approved):

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
```diff
@@ -74,6 +74,7 @@ def course_run_detail(self, request, detail_fields, course_run):
             'restriction_type': (
                 course_run.restricted_run.restriction_type if hasattr(course_run, 'restricted_run') else None
             ),
+            'fixed_price_usd': float(course_run.fixed_price_usd) if course_run.fixed_price_usd else None,
```
DawoudSheraz, line 77: why do we need to convert it to float? Better to round the value to avoid unexpected results.

Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/search_indexes/serializers/course.py
AfaqShuaib09: Removed float-type casting; `fixed_price_usd` is itself a float field, so it is no longer required.

```python
fixed_price_usd = fields.FloatField()
```

AfaqShuaib09: Spoke too soon; removing the check results in some test failures.

`course_run.fixed_price_usd` returns a Decimal type i.e; `Decimal(32.23)`, and we need to cast it to float for comparison with the expected data.

DawoudSheraz (approved):

PR #4409, opened 2024-08-13
Django ORM POC for SFMC Catalog

AfaqShuaib09:
[PROD-4137](https://2u-internal.atlassian.net/browse/PROD-4137)
---------
This PR adds management command to populate the course-catalog for SFMC

### Testing:

**Testing Instructions:**

1. Run the following command:
   ```bash
   ./manage.py populate_product_catalog --product_type={product_type} --output_csv={csv-file-name} --product_source={product_source} --use_gspread_client=True

Verify that the course data has been populated successfully.

- To test the gspread integration:

    - Create a service account with your Google account.
   - Add the service account credentials to the `private.py` file.
   - Create a new Google Sheet from your account and add the sheet ID to the `PRODUCT_CATALOG_SHEET_ID` variable in 
   the private.py file.
   - Run the command again to ensure the changes are reflected.


course_discovery/apps/course_metadata/models.py
```diff
@@ -1708,6 +1708,13 @@ def languages(self, exclude_inactive_runs=False):
             serialize_language(course_run.language) for course_run in self.course_runs.all()
             if course_run.language is not None
         })
+    
+    @property
+    def languages_codes(self):
+        """
+        Returns a string of languages codes used in this course. The languages codes are separated by comma.
+        """
+        return ','.join(course_run.language.code for course_run in self.course_runs.all() if course_run.is_active and course_run.language.code is not None)
```
DawoudSheraz: - We would need restricted type filtering here
- It would be better to have the if checks as part of filter()

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,140 @@
...
+                                    translation.name
+                                    for subject in product.subjects.all()
+                                    for translation in subject.spanish_translations
+                                ),
+                                "Marketing_URL": product.marketing_url,
+                                "Marketing_Image": (
+                                    product.image.url if product.image else ""
+                                ),
+                            }
+                        )
```
DawoudSheraz: nit: it would be good if we could use dict writer. The code would be readable and can be made generic enough. We can use https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py#L256 as a ref.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,140 @@
...
+                'authoring_organizations',
+                Prefetch('subjects'),
+                subject_translations
+            )
+
+    def handle(self, *args, **options):  # pylint: disable=too-many-statements
+        product_type = options.get('product_type')
+        output_csv = options.get('output_csv')
+        product_source = options.get('product_source')
+        field_names = [
```
DawoudSheraz: These might remain constant, it is better to convert this to a class level variable

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: I'm already using `csv.DictWriter` 
https://github.com/openedx/course-discovery/pull/4409/files#diff-568322ed06768e3af85323c340d7cdfbf661355ebdc2da98226b48b15d18afd3R96 
However, I will further refactor this by moving it to a separate method, similar to the example you shared above

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,159 @@
...
+        return dict(
+            {
+                "UUID": str(product.uuid),
+                "Title": product.title,
+                "Organizations Name": ", ".join(org.name for org in product.authoring_organizations.all()),
+                "Organizations Logo": ", ".join(
+                    org.logo_image.url for org in product.authoring_organizations.all() if org.logo_image
+                ),
+                "Organizations Abbr": ", ".join(org.key for org in product.authoring_organizations.all()),
+                "Languages": product.languages_codes,
```
DawoudSheraz: Shouldn't this be lang names?

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,159 @@
...
+
+                    logger.info(f'Populated {products.count()} {product_type}s to {output_csv}')
+            
+            elif gspread_client_flag:
+                csv_data = [self.get_transformed_data(product) for product in products]
+                gspread_client.write_data(
+                    PRODUCT_CATALOG_CONFIG,
+                    self.CATALOG_CSV_HEADERS,
+                    csv_data,
+                    overwrite=True
```
DawoudSheraz: For clarity, do we need to overwrite the previous data? Wouldn't it be better that a new tab is created to add that new data?

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: verified from the Course Catalog v1.2 spreadsheet, there are the language codes instead of lang names

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: I'm not sure how frequently this job will be executed (I believe once a day). If that's the case, there are a lot of tabs in the sheet, which might cause confusion. I also believe there might be a limit on the number of tabs in a spreadsheet. Therefore, I kept it in overwrite mode.

However, I can add your suggestion into the PR if you'd like  (I'll rename the tab with the date on which job is executed + lob type)

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def get_transformed_data(self, product):
+        """
+        Transforms the product data for product's catalog
+        """
+        return dict(
+            {
+                "UUID": str(product.uuid),
+                "Title": product.title,
+                "Organizations Name": ", ".join(org.name for org in product.authoring_organizations.all()),
```
DawoudSheraz: did you check if this `,` is contained within the column? In CSV loader, I remember we had to wrap the string in additional `""` to ensure any escape character in the string does not cause CSV to format incorrectly.
Ref: https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py#L263

AfaqShuaib09: > did you check if this , is contained within the column? 

yes, I've verified this.

> I remember we had to wrap the string in additional "" to ensure any escape character in the string does not cause CSV to format incorrectly.

Lemme add this check real quick.

course_discovery/apps/course_metadata/gspread_client.py
```diff
@@ -37,6 +38,96 @@ def read_data(self, config):
...
         return None
 
+    def _get_or_create_worksheet(self, spread_sheet, tab_id, cols, rows):
+        """
+        Get or create a worksheet with the given tab_id in the given spread_sheet
+
+        Args:
+            spread_sheet: The spread sheet object
+            tab_id: The tab id of the worksheet
+            headers: The headers of the worksheet
```
DawoudSheraz: headers is present as cols on function, please update the function declaration.

course_discovery/apps/course_metadata/gspread_client.py
```diff
@@ -37,6 +38,96 @@ def read_data(self, config):
...
+    def _write_headers(self, sheet_tab, headers):
+        """
+        Write headers to the first row of the worksheet
+
+        Args:
+            sheet_tab: The worksheet object
+            headers: The headers of the worksheet
+        """
+        sheet_tab.append_row(headers)
+        end_column = ascii_uppercase[len(headers) - 1]
```
DawoudSheraz, line 69: why is there a need to call ascii_uppercase here?

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+    def handle(self, *args, **options):
+        product_type = options.get('product_type')
+        output_csv = options.get('output_csv')
+        product_source = options.get('product_source')
+        gspread_client_flag = options.get('gspread_client_flag')
+        overwrite = options.get('overwrite_flag')
+        PRODUCT_CATALOG_CONFIG = {
+            'SHEET_ID': settings.PRODUCT_CATALOG_SHEET_ID,
```
DawoudSheraz, line 145: In a followup PR, we can think of adding a config model to provide these arguments. It is possible we would need to run the job with slightly different config and it would be easy to do so via config model.

course_discovery/apps/course_metadata/gspread_client.py
```diff
@@ -37,6 +38,96 @@ def read_data(self, config):
...
+            config: The configuration for the google spread sheet
+            csv_headers: The headers of the data to be written in the worksheet
+            csv_data: The data to be written in the worksheet, as a list of dictionaries, where
+            each dictionary represents a row
+            overwrite: Whether to overwrite the existing data in the worksheet
+        """
+        try:
+            spread_sheet = self.get_spread_sheet_by_key(config["SHEET_ID"])
+            sheet_tab = self._get_or_create_worksheet(
+                spread_sheet, config["OUTPUT_TAB_ID"], len(csv_headers) + 1, len(csv_data) + 1
```
DawoudSheraz, line 109: The output tab should be provided from arguments and it should not be hardcoded. It will make it easier to verify data changes.

course_discovery/apps/course_metadata/gspread_client.py
AfaqShuaib09: to determine cell range on the basis of attributes count
i.e; `A1:H1`
the next line is doing this

course_discovery/apps/course_metadata/gspread_client.py
AfaqShuaib09: config is just a dict passing to this gspread method.
I'm currently setting `tab_id` based on the product_type, but we can add it as command parameter as well

DawoudSheraz: oh, right. I read the wrong ID variable. Makes sense.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+                "Subjects Spanish": ", ".join(
+                    translation.name for subject in product.subjects.all()
+                    for translation in subject.spanish_translations
+                ),
+                "Marketing URL": product.marketing_url,
+                "Marketing Image": (product.image.url if product.image else ""),
+            }
+        )
+
+    def handle(self, *args, **options):
```
Ali-D-Akbar, line 138: Please add an example usage for this command.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+                    PRODUCT_CATALOG_CONFIG,
+                    self.CATALOG_CSV_HEADERS,
+                    csv_data,
+                    overwrite=overwrite,
+                )
+                logger.info(f'Populated {products.count()} {product_type}s to Google Sheets')
+
+        except Exception as e:
+            logger.error(f'Error while populating product catalog: {str(e)}')
+            raise CommandError(f'Error while populating product catalog: {str(e)}') from e
```
Ali-D-Akbar: No need to use both logger and raising CommandError

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        gspread_client = GspreadClient()
+
+        try:
+            products = self.get_products(product_type, product_source)
+            if not products.exists():
+                raise CommandError('No products found for the given criteria.')
+
+            logger.info(f'Fetched {products.count()} courses from the database')
```
Ali-D-Akbar: nit: product_count = products.count() 
it has been used multiple times.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,178 @@
...
+                with open(output_csv, 'w', newline='') as output_file:
+                    output_writer = self.write_csv_header(output_file)
+                    for product in products:
+                        try:
+                            output_writer.writerow(self.get_transformed_data(product))
+                        except Exception as e:  # pylint: disable=broad-exception-caught
+                            logger.error(f"Error writing product {product.uuid} to CSV: {str(e)}")
+                            continue
+
+                    logger.info(f'Populated {products.count()} {product_type}s to {output_csv}')
```
Ali-D-Akbar: optional: You can move them to small methods like write_to_csv, write_to_gsheets to improve readability.

Ali-D-Akbar (approved):
Looks good overall. Bunch of suggestions here and there.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: added

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: I believe we can skip this for now because we need to add "degree" `product_type` to it. We can further refactor this after adding that.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        writer = csv.DictWriter(output_csv, fieldnames=self.CATALOG_CSV_HEADERS)
+        writer.writeheader()
+        return writer
+
+    def get_transformed_data(self, product):
+        """
+        Transforms the product data for product's catalog
+        """
+        return dict(
```
hamza-56: The `dict()` function is unnecessary when using curly braces `{}` for creating dictionaries.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,184 @@
...
+            {
+                "UUID": str(product.uuid),
+                "Title": product.title,
+                "Organizations Name": ", ".join(org.name for org in product.authoring_organizations.all()),
```
hamza-56: Store the result of `product.authoring_organizations.all()` in a variable and reuse it. It has been used multiple times.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,184 @@
...
+        if product_type.lower() in ['executive_education', 'bootcamp', 'ocm_course']:
+            queryset = Course.objects.available()
+
+            if product_type.lower() == 'ocm_course':
+                queryset = queryset.filter(type__slug__in=ocm_course_catalog_types)
+
+            elif product_type.lower() == 'executive_education':
+                queryset = queryset.filter(type__slug=CourseType.EXECUTIVE_EDUCATION_2U)
+
+            elif product_type.lower() == 'bootcamp':
```
hamza-56: Store the lowercase value of `product_type` in a variable and reuse it.
it has been used multiple times.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,184 @@
...
+                to_attr='spanish_translations'
+            )
+
+            return queryset.prefetch_related(
+                'authoring_organizations',
+                Prefetch('subjects'),
+                subject_translations
+            )
+        else:
+            return None
```
hamza-56:
```suggestion
            return Course.objects.none()
```

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: I believe it won't cause any improvement in queries count because I've already prefetched in the queryset here: https://github.com/openedx/course-discovery/pull/4409/files#diff-568322ed06768e3af85323c340d7cdfbf661355ebdc2da98226b48b15d18afd3R98-R101

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: Okay, however, this `lower()` is only going to be called once since they are if and else if conditions

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: I don't think we want to set an empty queryset here because there will come logic for fetching degrees.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
DawoudSheraz: No, it wont be called once. If product type is BootCamp, it will be called 4 times (for each if, elif block). You can use walrus `:=` operator in first if to define the variable and use that for if-elif blocks.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: Yeah, I can remove it. It is also present in other places in the codebase, so I added it as a convention here

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: Ah, updating it. Great suggestion for using the walrus operator.

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
hamza-56: When using a `product_type` other than 'executive_education', 'bootcamp', or 'ocm_course', the command fails and throws an error because queries like `.exists()` are attempted on `None`. Instead, we should handle all cases gracefully and return an empty queryset.

Screenshot 2024-08-28 at 1 14 20 PM

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
hamza-56: Storing the result of `product.authoring_organizations.all()` in a variable can still provide other benefits, particularly in terms of code **readability** and **maintainability**.

hamza-56: @AfaqShuaib09 can we please do the same in test file?

hamza-56 (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -1709,6 +1709,14 @@ def languages(self, exclude_inactive_runs=False):
             if course_run.language is not None
         })
 
+    @property
+    def languages_codes(self):
+        """
+        Returns a string of languages codes used in this course. The languages codes are separated by comma.
+        """
+        filtered_course_runs = self.active_course_runs.filter(language__isnull=False, restricted_run__isnull=True)
```
zawan-ila, line 1718: I feel like the method should have the signature `def languages_codes(self, exclude_restricted=True)`, or atleast we should make it more obvious that the method will ignore restricted runs (by adding it to the docstring or changing the function name maybe)

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -0,0 +1,184 @@
...
+            # Prefetch Spanish translations of subjects
+            subject_translations = Prefetch(
+                'subjects__translations',
+                queryset=SubjectTranslation.objects.filter(language_code='es'),
+                to_attr='spanish_translations'
+            )
+
+            return queryset.prefetch_related(
+                'authoring_organizations',
+                Prefetch('subjects'),
```
zawan-ila: [nit] Remove the `Prefetch`

zawan-ila (approved):

PR #4411, opened 2024-08-16
fix: use correct course_key in case of a key_for_rerun in RCM

Ali-D-Akbar:
[PROD-4133](https://2u-internal.atlassian.net/browse/PROD-4133)

Updates the logic behind bringing the courses from studio in case a course is created in Studio and not in Discovery. Previous logic used to create a course_run from Studio into Discovery with whatever was present in the key. In case a course has a value in `key_for_reruns`, it used to assign those course_runs to the key equal to `key_for_reruns`.

Now, it finds if a course has a `key_for_reruns`, it will assign this course_run to the correct course that has the custom `key_for_reruns` value.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/api.py
@@ -202,6 +202,7 @@ class CoursesApiDataLoader(AbstractDataLoader):
     def get_or_create_course(self, body):
         course_run_key = CourseKey.from_string(body['id'])
         course_key = self.get_course_key_from_course_run_key(course_run_key)
+
         defaults = self.format_course_data(body)
         # We need to add the key to the defaults because django ignores kwargs with __
         # separators when constructing the create request
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
@@ -534,6 +534,32 @@ class CoursesApiDataLoaderTests(DataLoaderTestMixin, TestCase):
         assert draft_run != official_run
         assert official_run.draft_version == draft_run
 
+    @responses.activate
+    def test_ingest_with_key_for_rerun(self):
+        """
+        Verify that we correctly stitch up a course run made in Studio but having key_for_rerun.
+        The test makes sure that the course_run isn't incorrectly assigned to course having key = key_for_rerun.
+        """
+        datum = mock_data.COURSES_API_BODY_ORIGINAL
+        self.mock_api([datum])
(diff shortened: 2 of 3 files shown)


course_discovery/apps/course_metadata/data_loaders/api.py
```diff
@@ -202,6 +202,8 @@ def create_course_run(self, course, body):
     def get_or_create_course(self, body):
         course_run_key = CourseKey.from_string(body['id'])
         course_key = self.get_course_key_from_course_run_key(course_run_key)
+        course_key = (Course.objects.filter(key_for_reruns=course_key).values_list('key', flat=True).first() or
```
DawoudSheraz: nit: There should be a comment explaining this scenario in the code. Or better, have an internal or util function to return the course key that considers both scenarios.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -534,6 +534,32 @@ def test_ingest_studio_made_run_with_existing_draft_course(self):
...
+    @responses.activate
+    def test_ingest_with_key_for_rerun(self):
+        """
+        Verify that we correctly stitch up a course run made in Studio but having key_for_rerun.
+        The test makes sure that the course_run isn't incorrectly assigned to course having key = key_for_rerun.
+        """
+        datum = mock_data.COURSES_API_BODY_ORIGINAL
+        self.mock_api([datum])
+        course_type = CourseType.objects.get(slug=CourseType.AUDIT)
+        course_key = '{org}+{number}'.format(org=datum['org'], number=datum['number'])
```
DawoudSheraz: you can use f-string, no need to use format anymore.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -534,6 +534,32 @@ def test_ingest_studio_made_run_with_existing_draft_course(self):
...
+        course_key_with_key_for_rerun = 'test+key_for_rerun'
+        Course.objects.create(partner=self.partner, key=course_key, title='Title', type=course_type)
```
DawoudSheraz, line 548: should the first course be created here? Shouldn't only the course with key_for_reruns be created only?

Ali-D-Akbar: The scenario we're testing should have a courseA and courseB having key_for_reruns = courseA.key

DawoudSheraz (approved):

AfaqShuaib09 (approved):

PR #4422, opened 2024-09-02
fix: decoding of google_service_account_credentials

AfaqShuaib09:
[PROD-4167](https://2u-internal.atlassian.net/browse/PROD-4167)
-------
This PR attempts to fix the deserialization issue with `GOOGLE_SERVICE_ACCOUNT_CREDENTIALS`. The same issue occurred with `ai_translations` IDA, and they added this script as a fix there: https://github.com/edx/ai-translations/pull/84/files.

### DIFF
--- a/course_discovery/settings/production.py
+++ b/course_discovery/settings/production.py
@@ -89,5 +89,8 @@ COMPRESS_OFFLINE = True
 # Have images and such that we upload be publicly readable
 AWS_DEFAULT_ACL = 'public-read'
 
-# Convert dict keys to lowercase
-GOOGLE_SERVICE_ACCOUNT_CREDENTIALS = {k.lower(): v for k, v in GOOGLE_SERVICE_ACCOUNT_CREDENTIALS.items()}
+# Convert dict keys to lowercase. The if condition replaces \\n in private_key value with \n.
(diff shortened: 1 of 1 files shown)


course_discovery/settings/production.py
```diff
@@ -89,5 +89,8 @@
 # Have images and such that we upload be publicly readable
 AWS_DEFAULT_ACL = 'public-read'
 
-# Convert dict keys to lowercase
-GOOGLE_SERVICE_ACCOUNT_CREDENTIALS = {k.lower(): v for k, v in GOOGLE_SERVICE_ACCOUNT_CREDENTIALS.items()}
+# Convert dict keys to lowercase. The if condition replaces \\n in private_key value with \n.
```
DawoudSheraz, line 92: any particular reasons this is needed?

DawoudSheraz (approved):

course_discovery/settings/production.py
AfaqShuaib09: I'm not 100% sure, Could be the reason mentioned here https://stackoverflow.com/questions/64694476/problem-with-k8s-reading-keys-valueerror-could-not-deserialize-key-data

Ali-D-Akbar (approved):

PR #4423, opened 2024-09-03
perf: reduce write requests call to google spreadsheet using gspread client via append_rows method

AfaqShuaib09:
This PR replaces the append_row method with the append_rows method to write all rows to the spreadsheet in a single call instead of making separate calls for each row.


DawoudSheraz (approved):
Once CI is green

PR #4426, opened 2024-09-04
feat: Add improvements to populate_product_catalog_command

AfaqShuaib09:
[PROD-4167](https://2u-internal.atlassian.net/browse/PROD-4167)
------

This PR adds following improvements to management command
- Skip record if required detail is empty.
- Remove duplicates from language_code method.
- Add flag to include inactive runs' language codes in the method.
- Fix N+1 query issue.
- Remove dashes from uuid.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
+++ b/course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
@@ -4,7 +4,7 @@ import logging
 
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
-from django.db.models import Prefetch
+from django.db.models import Count, Prefetch, Q
 
 from course_discovery.apps.course_metadata.gspread_client import GspreadClient
 from course_discovery.apps.course_metadata.models import Course, CourseType, Program, SubjectTranslation
@@ -74,7 +74,7 @@ class Command(BaseCommand):
         ]
 
         if product_type in ['executive_education', 'bootcamp', 'ocm_course']:
-            queryset = Course.objects.available()
+            queryset = Course.objects.available().select_related('partner', 'type')
 
             if product_type == 'ocm_course':
                 queryset = queryset.filter(type__slug__in=ocm_course_catalog_types)
@@ -87,6 +87,10 @@ class Command(BaseCommand):
 
             if product_source:
                 queryset = queryset.filter(product_source__slug=product_source)
+
+            queryset = queryset.annotate(
+                num_orgs=Count('authoring_organizations')
+            ).filter(Q(num_orgs__gt=0) & Q(image__isnull=False) & ~Q(image=''))
 
             # Prefetch Spanish translations of subjects
             subject_translations = Prefetch(
@@ -105,6 +109,10 @@ class Command(BaseCommand):
 
             if product_source:
                 queryset = queryset.filter(product_source__slug=product_source)
+
+            queryset = queryset.annotate(
+                num_orgs=Count('authoring_organizations')
+            ).filter(Q(num_orgs__gt=0) & Q(card_image__isnull=False) & ~Q(card_image=''))
 
             subject_translations = Prefetch(
                 'courses__subjects__translations',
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/tests/factories.py
```diff
@@ -669,6 +669,7 @@ class Meta:
     type = factory.SubFactory(ProgramTypeFactory)
     marketing_slug = factory.Sequence(lambda n: f'test-slug-{n}')
     card_image_url = FuzzyText(prefix='https://example.com/program/card')
+    card_image = factory.django.ImageField()
```
DawoudSheraz: why is this needed here?

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -88,6 +88,10 @@ def get_products(self, product_type, product_source):
             if product_source:
                 queryset = queryset.filter(product_source__slug=product_source)
 
+            queryset = queryset.annotate(
+                num_orgs=Count('authoring_organizations')
+            ).filter(Q(num_orgs__gt=0) & Q(image__isnull=False) & ~Q(image=''))
```
DawoudSheraz, line 93: is this image check against organization or against course? There are duplicate fields, so you might want to confirm which one is not deprecated.

course_discovery/apps/course_metadata/tests/factories.py
AfaqShuaib09: Previously, it was not added here. I added it now so I don't have to explicitly define the image value when creating a program or degree object through the factory

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: against course, I've verified it from admin. It's not deprecated

course_discovery/apps/course_metadata/tests/factories.py
AfaqShuaib09: However, it's failing other test cases, so I think I need to initialize it directly in the test case where it's required.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -100,6 +100,30 @@ def test_image_url(self):
...
+            language=LanguageTag.objects.get(code='en')
+        )
+        SeatFactory.create(course_run=course_run)
+        CourseRunFactory(
+            course=Course.objects.all()[0],
+            status=CourseRunStatus.Unpublished,
+            language=LanguageTag.objects.get(code='es'),
+            enrollment_end=datetime.datetime.now() - datetime.timedelta(days=5)
+        )
+        assert course.languages_codes() in ['en, es', 'es, en']
```
DawoudSheraz: isn't es, en and en, es the same thing?

DawoudSheraz: What you can do is convert the set back to list and sort in a particular order. The changing order of the data with every hit is not a consistent behavior

AfaqShuaib09: sure, lemme update this

DawoudSheraz (approved):

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
```diff
@@ -106,6 +110,10 @@ def get_products(self, product_type, product_source):
             if product_source:
                 queryset = queryset.filter(product_source__slug=product_source)
 
+            queryset = queryset.annotate(
+                num_orgs=Count('authoring_organizations')
+            ).filter(Q(num_orgs__gt=0) & Q(card_image__isnull=False) & ~Q(card_image=''))
```
Ali-D-Akbar, line 115: why are we filtering out based on `card_image` here while we used `image` before?

Ali-D-Akbar (approved):
one comment from me

course_discovery/apps/course_metadata/management/commands/populate_product_catalog.py
AfaqShuaib09: It's for degrees while the above one is for courses

PR #4434, opened 2024-09-09
feat: return override fields from properties when they are populated

AfaqShuaib09:
### [PROD-4176](https://2u-internal.atlassian.net/browse/PROD-4176)

Design methods for subjects and languages, giving preference to overrides rather than using the simple method values.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -3519,6 +3519,16 @@ class Program(ManageHistoryMixin, PkSearchableMixin, TimeStampedModel):
         return {course_run.language for course_run in self.course_runs if course_run.language is not None}
 
     @property
+    def active_languages(self):
+        """
+        :return: The list of languages; It gives preference to the language_override over the languages
+        extracted from the course runs.
+        """
+        if self.language_override:
+            return {self.language_override}
+        return self.languages
+
+    @property
     def transcript_languages(self):
         languages = [course_run.transcript_languages.all() for course_run in self.course_runs]
         languages = itertools.chain.from_iterable(languages)
@@ -3539,6 +3549,24 @@ class Program(ManageHistoryMixin, PkSearchableMixin, TimeStampedModel):
         common_primary = [s for s, _ in Counter(primary_subjects).most_common()][:1]
         common_others = [s for s, _ in Counter(course_subjects).most_common() if s not in common_primary]
         return common_primary + common_others
+
+    @property
+    def active_subjects(self):
+        """
+        :return: The list of subjects; the first subject should be the most common primary subjects of its courses,
+        other subjects should be collected and ranked by frequency among the courses.
+
+        Note: This method gives preference to the primary_subject_override over the primary subject of the courses.
+        """
+        subjects = self.subjects
+
+        if self.primary_subject_override:
+            if self.primary_subject_override not in subjects:
+                subjects = [self.primary_subject_override] + subjects
+            else:
+                subjects = [self.primary_subject_override] + [subject for subject in subjects if subject != self.primary_subject_override]
+
+        return subjects
 
     @property
     def topics(self):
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -3518,6 +3518,16 @@ def course_run_statuses(self):
...
         return {course_run.language for course_run in self.course_runs if course_run.language is not None}
 
+    @property
+    def active_languages(self):
+        """
+        :return: The list of languages; It gives preference to the language_override over the languages
+        extracted from the course runs.
+        """
+        if self.language_override:
+            return {self.language_override}
```
DawoudSheraz, line 3528: what's with dict operator?

DawoudSheraz: Shouldn't this be a list?

AfaqShuaib09: +1, it should be a list instead of a set

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -3519,6 +3519,65 @@ def test_program_duration_override(self):
...
+        """
+        Test that active_languages returns the language_override when it is set.
+        """
+
+        language_es = LanguageTag.objects.get(code='es')
+        language_de = LanguageTag.objects.get(code='de')
+        course_run = CourseRunFactory.create(language=language_de)
+        program = ProgramFactory.create(language_override=language_es, courses=[course_run.course])
+
+        expected_languages = {language_es}
```
AfaqShuaib09, line 3578: replace set with list

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -3519,6 +3519,65 @@ def test_program_duration_override(self):
...
+
+        language_en = LanguageTag.objects.create(code='en', name='English')
+        language_fr = LanguageTag.objects.get(code='fr')
+
+        course_run1 = CourseRunFactory.create(language=language_en)
+        course_run2 = CourseRunFactory.create(language=language_fr)
+
+        program = ProgramFactory.create(language_override=None, courses=[course_run1.course, course_run2.course])
+
+        expected_languages = {language_en, language_fr}
```
AfaqShuaib09, line 3565: replace set with list

course_discovery/apps/course_metadata/models.py
```diff
@@ -3540,6 +3550,25 @@ def subjects(self):
...
+        Note: This method gives preference to the primary_subject_override over the primary subject of the courses.
+        """
+        subjects = self.subjects
+
+        if self.primary_subject_override:
+            if self.primary_subject_override not in subjects:
+                subjects = [self.primary_subject_override] + subjects
+            else:
+                subjects = [self.primary_subject_override] + \
+                    [subject for subject in subjects if subject != self.primary_subject_override]
```
DawoudSheraz, line 3568: wont this code be simple if we make subjects list set, adding primary override (if present) and convert it back to list?

course_discovery/apps/course_metadata/models.py
hamza-56:
```
 @property
    def languages(self):
        return {course_run.language for course_run in self.course_runs if course_run.language is not None}
```

The return type of **languages** is a set, so we're returning a set in **active_languages** as well to maintain consistency in the code.

course_discovery/apps/course_metadata/models.py
hamza-56: While converting the subjects list to a set would simplify the code by automatically handling duplicates, we need to ensure the order of subjects is preserved, especially since the primary subject override should always appear first. 
The current approach ensures both:

1. The primary subject is explicitly placed at the front.
2. Order is maintained for the rest of the subjects.

DawoudSheraz: Makes sense.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Right, makes sense.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/models.py
zawan-ila: Why not remove the inner if else and swap it with the assignment in the else? That would be equivalent to the current implementation?

zawan-ila (approved):

PR #4443, opened 2024-09-11
feat: add capabilty in queryset method to exclude hidden runs

AfaqShuaib09:
[PROD-4177](https://2u-internal.atlassian.net/browse/PROD-4177)
--------
This PR adds the capability in `populate_product_catalog` command to exclude hidden available runs from the queryset.

Testing:
- Create a run from publisher and publish it. Make it hidden from django admin (both darft and non-draft)
- Run the following command from discovery shell.  
```bash
python manage.py populate_product_catalog --product_type={product_type} --output_csv=/path/to/output.csv --product_source={product_source}
```
- Verify Course having only hidden published runs are excluded from the catalog

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_populate_product_catalog.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_populate_product_catalog.py
@@ -41,6 +41,12 @@ class PopulateProductCatalogCommandTests(TestCase):
         self.course_run_2 = CourseRunFactory.create_batch(
             2, course=Course.objects.all()[1]
         )
+        self.course_run_3 = CourseRunFactory(
+            course=Course.objects.all()[1],
+            status=CourseRunStatus.Published,
+            hidden=True,
+        )
+        self.hidden_run_seat = SeatFactory.create(course_run=self.course_run_3)
         self.program_type = ProgramTypeFactory.create(slug=ProgramType.MICROMASTERS)
         self.degrees = DegreeFactory.create_batch(
             2,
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/management/commands/tests/test_populate_product_catalog.py
```diff
@@ -79,6 +85,65 @@ def test_populate_product_catalog(self):
                     self.assertIn("Marketing URL", row)
                     self.assertIn("Marketing Image", row)
 
+    def test_populate_product_catalog_for_courses_with_hidden_and_non_hidden_published_runs(self):
```
Ali-D-Akbar, line 112: Just a suggestion. We can maybe add this test in `test_populate_product_catalog`. 
Or maybe make a reusable method between the unit tests since it has some common block of code if I'm not wrong.

Ali-D-Akbar (approved):
Looks good to me overall.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_product_catalog.py
AfaqShuaib09: yup we can, lemme add all the use cases in that method

course_discovery/apps/course_metadata/query.py
```diff
@@ -54,7 +54,12 @@ def available(self):
         # By itself, the query performs a join across several tables and would return
         # the id of the same course multiple times (a separate copy for each available
         # seat in each available run).
-        ids = self.filter(enrollable & not_ended & marketable).values('id').distinct()
+        if exclude_hidden_runs:
+            # A CourseRun is "hidden" if it has a "hidden" attribute set to True.
```
DawoudSheraz: There is no need for the comment here, the code is fairly obvious here.

course_discovery/apps/course_metadata/query.py
```diff
@@ -54,7 +54,12 @@ def available(self):
...
+            hidden = ~Q(course_runs__hidden=True)
```
DawoudSheraz: the variable name is not reflecting the correct action. The value is ~Q i.e. not hidden. The code is including non-hidden runs only. Therefore, the variables should reflect that too.

course_discovery/apps/course_metadata/query.py
```diff
@@ -8,16 +8,16 @@
 
 
 class CourseQuerySet(models.QuerySet):
-    def available(self):
+    def available(self, exclude_hidden_runs=False):
```
DawoudSheraz, line 11: please update docstring to include the details of this argument.

course_discovery/apps/course_metadata/query.py
AfaqShuaib09: updated

course_discovery/apps/course_metadata/query.py
AfaqShuaib09: done

DawoudSheraz (approved):

PR #4453, opened 2024-09-26
feat: set publish date in future for scheduled state variants

AfaqShuaib09:
[PROD-4185](https://2u-internal.atlassian.net/browse/PROD-4185)
----------
This PR set publish date in future for scheduled state execed variants to support future variants feature.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
+++ b/course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
@@ -152,8 +152,10 @@ class Command(BaseCommand):
                 if getsmarter_flag:
                     product['organization'] = map_external_org_code_to_internal_org_code(
                         product['universityAbbreviation'], product_source)
-                if product.get('variants'):
+                if 'variants' in product:
                     variants = product.pop('variants')
+                    if not variants:
+                        logger.warning(f"Skipping product {product['name']} ingestion as it has no variants")
                     for variant in variants:
                         product.update({'variant': variant})
                         output_dict = self.get_transformed_data(row, product)
@@ -292,7 +294,6 @@ class Command(BaseCommand):
             'start_time': '00:00:00',
             'end_time': '00:00:00',
             'reg_close_time': '00:00:00',
-            'publish_date': date.today().isoformat(),
             'course_level': 'Introductory',
             'course_pacing': 'Instructor-Paced',
             'content_language': language,
@@ -313,6 +314,10 @@ class Command(BaseCommand):
             'organization_short_code_override': product_dict[
                 'altUniversityAbbreviation'
             ],
+            'publish_date': (
+                partially_filled_csv_dict.get('start_date') or product_dict['variant']['startDate']
+                if product_dict['variant'].get('status') == 'scheduled' else date.today().isoformat()
+            ),
             '2u_organization_code': product_dict['universityAbbreviation'],
             'number': product_dict['abbreviation'],
             'alternate_number': product_dict['altAbbreviation'],
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
@@ -7,6 +7,7 @@ import json
 from datetime import date
 from tempfile import NamedTemporaryFile
 
+import ddt
 import mock
 import responses
 from django.conf import settings
@@ -20,6 +21,7 @@ from course_discovery.apps.course_metadata.data_loaders.tests.mixins import CSVL
 LOGGER_PATH = 'course_discovery.apps.course_metadata.management.commands.populate_executive_education_data_csv'
 
 
+@ddt.ddt
 class TestPopulateExecutiveEducationDataCsv(CSVLoaderMixin, TestCase):
     """
     Test suite for populate_executive_education_data_csv management command.
@@ -191,6 +193,41 @@ class TestPopulateExecutiveEducationDataCsv(CSVLoaderMixin, TestCase):
                 ),
             )
 
+    @mock.patch("course_discovery.apps.course_metadata.utils.GetSmarterEnterpriseApiClient")
+    def test_skip_products_ingestion_if_variants_data_empty(self, mock_get_smarter_client):
+        """
+        Verify that the command skips the product ingestion if the variants data is empty
+        """
+        success_api_response = copy.deepcopy(self.SUCCESS_API_RESPONSE_V2)
+        success_api_response["products"][0]["variants"] = []
+        mock_get_smarter_client.return_value.request.return_value.json.return_value = (
+            self.mock_get_smarter_client_response(
+                override_get_smarter_client_response=success_api_response
+            )
+        )
+        with NamedTemporaryFile() as output_csv:
+            with LogCapture(LOGGER_PATH) as log_capture:
+                call_command(
+                    "populate_executive_education_data_csv",
+                    "--output_csv",
+                    output_csv.name,
+                    "--use_getsmarter_api_client",
+                    True,
+                )
+                log_capture.check_present(
+                    (
+                        LOGGER_PATH,
+                        "WARNING",
+                        f"Skipping product {success_api_response['products'][0]['name']} "
+                        f"ingestion as it has no variants",
+                    ),
+                )
+
+                output_csv.seek(0)
+                with open(output_csv.name, "r") as csv_file:
+                    reader = csv.DictReader(csv_file)
+                    assert not any(reader)
+
     @mock.patch('course_discovery.apps.course_metadata.utils.GetSmarterEnterpriseApiClient')
     def test_successful_file_data_population_with_getsmarter_flag_with_multiple_variants(self, mock_get_smarter_client):
         """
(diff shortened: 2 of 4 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -2600,13 +2600,6 @@ def availability(self):
         now = datetime.datetime.now(pytz.UTC)
         upcoming_cutoff = now + datetime.timedelta(days=60)
 
-        # Use external course availability for ExecEd course run types
-        # Check additional_metadata existence as a fail safe for some unit tests
-        # where this flow is triggered via signals before AdditionalMetadata object can be persisted
-        if self.type.slug in [CourseRunType.UNPAID_EXECUTIVE_EDUCATION, CourseRunType.PAID_EXECUTIVE_EDUCATION] and \
-                self.course.additional_metadata:
-            return self.external_course_availability
-
```
AfaqShuaib09, line 2620: This customized logic won't be required now. As we have moved Course Run's data for ExecEd from AdditionalMetadata to CourseRun model just like OCs

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -152,8 +152,10 @@ def handle(self, *args, **options):  # pylint: disable=too-many-statements
                 if getsmarter_flag:
                     product['organization'] = map_external_org_code_to_internal_org_code(
                         product['universityAbbreviation'], product_source)
-                if product.get('variants'):
+                if 'variants' in product:
                     variants = product.pop('variants')
+                    if not variants:
+                        logger.warning(f"Skipping product {product['name']} ingestion as it has no variants")
```
DawoudSheraz, line 158: It should be if-else because the code below will execute. This is just logging the warning.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
```diff
@@ -191,6 +191,36 @@ def test_successful_file_data_population_with_getsmarter_flag(self, mock_get_sma
...
+                    True,
+                )
+                log_capture.check_present(
+                    (
+                        LOGGER_PATH,
+                        "WARNING",
+                        f"Skipping product {success_api_response['products'][0]['name']} "
+                        f"ingestion as it has no variants",
+                    ),
+                )
```
DawoudSheraz, line 224: Checking log is not enough here, the CSV should be checked to verify no record is present there.

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -313,6 +314,10 @@ def get_transformed_data(self, partially_filled_csv_dict, product_dict):
             'organization_short_code_override': product_dict[
                 'altUniversityAbbreviation'
             ],
+            'publish_date': (
+                partially_filled_csv_dict.get('start_date') or product_dict['variant']['startDate']
+                if product_dict['variant'].get('status') == 'scheduled' else date.today().isoformat()
```
DawoudSheraz, line 319: Is the status always present? If yes, we can use [] instead of .get. If not, this condition would need an update to cater to that.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
```diff
@@ -247,6 +277,67 @@ def test_successful_file_data_population_with_getsmarter_flag_with_multiple_vari
...
 
+    @mock.patch('course_discovery.apps.course_metadata.utils.GetSmarterEnterpriseApiClient')
+    def test_successful_file_data_population_with_getsmarter_flag_with_future_variants(self, mock_get_smarter_client):
+        """
+        Verify the successful population has data from API response if getsmarter flag is provided and
+        the product can have multiple variants
+        """
+        success_api_response = copy.deepcopy(self.SUCCESS_API_RESPONSE_V2)
+        success_api_response['products'][0]['variants'][0]['status'] = 'scheduled'
+        success_api_response['products'][0]['variants'][1]['status'] = 'active'
```
DawoudSheraz: it would be good to do this via ddt. The test is unnecessarily fat and has repeating code.

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
AfaqShuaib09: The next line is a for loop, and if the condition if not variants is true, it means the variants list is empty, so the statements inside the loop won't execute.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
AfaqShuaib09: okay, updating it adding assertions statements here

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
AfaqShuaib09: > Is the status always present?

On stage, yes, it is present for all variants. However, I believe it's not available on prod yet.

> this condition would need an update to cater to that.

There is already a condition in place, like if the status is scheduled, the publish date is set to the start date. Otherwise, if the status has any other value or is missing, the publish date is set to the current date, just as it was previously.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
AfaqShuaib09: `SUCCESS_API_RESPONSE_V2` is a class variable, so we can't refer it in ddt. We need to make it a global variable in order to refer it for ddt.

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
DawoudSheraz: How so? How do we know variants is an empty list and not None?

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
DawoudSheraz: Global variable would be unnecessary. We don't need to change/access variable in ddt. We can set the necessary conditions as ddt arguments and work with them accordingly.

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
AfaqShuaib09: The variants attribute is a list in the response, as I have verified from the stage GEAG response. 

This condition is only to cater the stage response, as there are a few discontinued courses on stage that no longer have variants. This won’t be the case in prod.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
```diff
@@ -191,6 +193,42 @@ def test_successful_file_data_population_with_getsmarter_flag(self, mock_get_sma
...
+                        "WARNING",
+                        f"Skipping product {success_api_response['products'][0]['name']} "
+                        f"ingestion as it has no variants",
+                    ),
+                )
+
+                output_csv.seek(0)
+                with open(output_csv.name, "r") as csv_file:
+                    reader = csv.DictReader(csv_file)
+                    # check for empty csv file
```
DawoudSheraz: There is no need for this comment, the code is obvious.

course_discovery/apps/course_metadata/management/commands/tests/test_populate_executive_education_data_csv.py
```diff
@@ -247,6 +285,49 @@ def test_successful_file_data_population_with_getsmarter_flag_with_multiple_vari
                 ),
             )
 
+    @mock.patch("course_discovery.apps.course_metadata.utils.GetSmarterEnterpriseApiClient")
+    @ddt.data(("active", "scheduled", str(date.today().isoformat()), "2024-03-20"))
```
DawoudSheraz: If ddt is being used, please have at least 2 ddt cases. Also, Instead of having variant_1 and variant_2 variables, I would suggest only checking one use-case in one ddt case i.e. check variant publish date when its status is active, check when status is scheduled in another ddt case. The current implementation does not make sense from ddt point of view.

DawoudSheraz (approved):

hamza-56 (approved):

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
zawan-ila, line 319: [thought] This will make future variants not marketable until their start date, right? That is probably good for the case where you already have an ongoing variant. But what if you only have a single variant? I'd expect "marketing" to want to be able to market that beforehand?

Or maybe GEAG updates the `status` to something other than scheduled some time before the start date in that scenario?

AfaqShuaib09: Ah, just saw your comment after merge. 😶‍🌫️ 
Scheduled variants are specifically for enterprise use-cases

DawoudSheraz: Additionally, even if there is a single variant, we can expect it to be active. It has been echoed a few times that if a product is coming in GEAG, it will have one active variant.

PR #4464, opened 2024-10-14
build: pin workflows to use ubuntu 22.04

UsamaSadiq:
## Description
- `ubuntu-latest` is causing issues in workflows due to updated `binaries` and `dependencies`.
- Pinned all workflows to use `ubuntu-22.04` by default just like we did recently in `edx-platform`.
- Example PR in edx-platform: https://github.com/openedx/edx-platform/pull/35635.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   pytest:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


UsamaSadiq:
@AfaqShuaib09 this should resolve the issues with workflow failures.

AfaqShuaib09 (approved):

DawoudSheraz (approved):

PR #4474, opened 2024-10-24
fix: do not push runs without seats to ecommerce in data loader

zawan-ila:
[PROD-4171](https://2u-internal.atlassian.net/browse/PROD-4171)
--------

In the courses data loader, we push to ecommerce if the run has an updated end date. This is primarily to push seat upgrade deadline changes (which depend on run end date) to ecommerce. However, in case a run is newly created and has no seats so far, a push to ecommerce will convince ecommerce that the run does not have any seats. Hence, even if we add seats to the draft version of the run, the Ecommerce Data loader might [delete](https://github.com/openedx/course-discovery/blob/dba582d05b72afe3f9ca408626e0dfddf55e16d1/course_discovery/apps/course_metadata/data_loaders/api.py#L569-L571) the seats from discovery if the seats have not been pushed to the non-draft version and ecommerce before the time that the ecomm loader runs.

This PR attempts to circumvent the above issue by only pushing a run to ecommerce in the courses data loader if it has a non-zero number of associated seats.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
@@ -379,12 +379,14 @@ class CoursesApiDataLoaderTests(DataLoaderTestMixin, TestCase):
         self.assert_api_called(4)
         runs = CourseRun.objects.all()
 
+        # Change a run's end date to make it different from the one in studio. This
+        # is needed to trigger the push_to_ecommerce flow in the the loader, which is only
+        # done in case the end dates differ or a certain waffle flag is set.
         run = runs[0]
         run.end = datetime.datetime.now(pytz.UTC)
         run.save()
         assert run.seats.count() == 0
 
-        # Verify the CourseRuns were created correctly
         expected_num_course_runs = len(api_data)
         assert CourseRun.objects.count() == expected_num_course_runs
 
--- a/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
+++ b/course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
@@ -382,6 +382,7 @@ class CoursesApiDataLoaderTests(DataLoaderTestMixin, TestCase):
         run = runs[0]
         run.end = datetime.datetime.now(pytz.UTC)
         run.save()
+        assert run.seats.count() == 0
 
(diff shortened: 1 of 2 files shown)


AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -362,6 +362,34 @@ def test_ingest_verified_deadline_with_bypass_end_date_check(self, mock_push_to_
...
+        api_data = self.mock_api()
+        assert Course.objects.count() == 0
+        assert CourseRun.objects.count() == 0
+
+        self.loader.ingest()
+        self.assert_api_called(4)
+        runs = CourseRun.objects.all()
+
+        run = runs[0]
+        run.end = datetime.datetime.now(pytz.UTC)
```
DawoudSheraz, line 386: why is end date set to now here? Where are we checking the seat non-existence here?

zawan-ila: added an assert for the seat non-existence. the end date is set to now to make it different from the one in studio. this is needed to prevent the flow from short circuiting [here](https://github.com/openedx/course-discovery/pull/4474/files#diff-1f034729bd8cd6d768299365b653531b9274d6f0917b9fb07735c5a996a124a2R173)

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
```diff
@@ -362,6 +362,35 @@ def test_ingest_verified_deadline_with_bypass_end_date_check(self, mock_push_to_
...
+        run.save()
+        assert run.seats.count() == 0
+
+        # Verify the CourseRuns were created correctly
```
DawoudSheraz: nit: no need for this comment, it is obvious from the code ahead.

course_discovery/apps/course_metadata/data_loaders/tests/test_api.py
DawoudSheraz: this one warrants a comment on the code, considering lots of setup ops are happening in setup.

DawoudSheraz (approved):

PR #4484, opened 2024-11-07
feat: add future variant info to email

zawan-ila:
[PROD-4224](https://2u-internal.atlassian.net/browse/PROD-4224)
-----

This PR updates ExecEd Ingestion summary email to highlight future variants


course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -176,15 +176,15 @@ def get_variants(self, product):
...
-                else:
-                    variants.append(product[key])
+        if 'variants' in product:
+            return product['variants']
+
+        variants = (
+            [product['variant']] if product['variant'] else []
+        )
+        variants.extend([{**v, 'isFutureVariant': True} for v in product.get('futureVariants', [])])
+        variants.extend(product.get('customPresentations', []))
```
DawoudSheraz: we do not want to change the flow here. GEAG is experimenting on their end, so we want to check all keys and then build a product list.

zawan-ila: > experimenting

Can you provide an example here? This, to me, looks much cleaner than before.

DawoudSheraz: They might expose both variant and variants on their end for time being, both having valid product. The change you have done, you are returning if variants is found. Before that, we are checking all keys and if any key has product against it, it is added in the products list.

DawoudSheraz: We might not need to tweak this, as we know for future variants, the status will be scheduled https://github.com/openedx/course-discovery/pull/4484/files#diff-0c29ac162abebb2fa53ae347763849a283355d22f8a44ae94178734ea1ca6933R338. We can leverage that information to determine if the variant is a future variant.

zawan-ila: > They might expose both variant and variants on their end for time being, both having valid product

So, it might be the case that `variants != variant+custom+future`? Also, in that case, would the ingestion stats still be accurate if some product is repeated across multiple keys?

> as we know for future variants, the status will be scheduled

I thought about that. I feel that this is more robust.

DawoudSheraz: What will happen for time being is variant and variants will be there (variants will have custom presentations). When everything is ready on their end (as there are other consumers of GEAG besides discovery), they will be removing variants. The product wont be repeated across keys as variant key will always have the current/active variant

> I thought about that. I feel that this is more robust.

Yes, but it is adding unnecessary changes in multiple places. We know for certain the future variants will have scheduled status, so we can derive data from that instead of having to change the pipeline in multiple places. I am fine with having that here but I would recommend not tweaking the code that much.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -176,15 +176,15 @@ def get_variants(self, product):
...
-                    variants.extend(product[key])
-                else:
-                    variants.append(product[key])
+        if 'variants' in product:
+            return product['variants']
+
+        variants = (
+            [product['variant']] if product['variant'] else []
+        )
+        variants.extend([{**v, 'isFutureVariant': True} for v in product.get('futureVariants', [])])
```
AfaqShuaib09: No need to extend the dictionary here, we can identify future variants based on their status.

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
AfaqShuaib09: Do we still need these extra checks here, because now the flag on the GEAG side has been added?

course_discovery/apps/course_metadata/management/commands/populate_executive_education_data_csv.py
```diff
@@ -337,6 +337,7 @@ def get_transformed_data(self, partially_filled_csv_dict, product_dict):
                 partially_filled_csv_dict.get('start_date') or product_dict['variant']['startDate']
                 if product_dict['variant'].get('status') == 'scheduled' else date.today().isoformat()
             ),
+            'is_future_variant': product_dict['variant'].get('isFutureVariant', False),
```
AfaqShuaib09: update this value with ``` status == "scheduled" ``` check

AfaqShuaib09: Ah, my bad I was looking at the previous commit 🤦🏻

AfaqShuaib09 (approved):

PR #4488, opened 2024-11-12
identify and prefetch N+1 queries in search/all for learner pathways

hamza-56:
### TL;DR
This PR addresses the need for optimized query fetching in the `/api/v1/search/all/?include_learner_pathways=true` endpoint by implementing `prefetch_related` for `LearnerPathway` data.

1. Fixes N+1 issues in `LearnerPathwayViewSet`
2. Fixes N+1 issues in `LearnerPathwayStepViewSet`
3. Fixes N+1 issues in `LearnerPathwayCourseViewSet`
4. Fixed N+1 issues in `LearnerPathwayProgramViewSet`
5. Removes unnecessary eager loading of `learnerpathwayblock_set` 

### Details
The `get_linked_courses_and_course_runs` model method is only used in `LearnerPathwayProgramSerializer`.
The `LearnerPathwayProgramSerializer` is used by the `LearnerPathwayProgramViewSet` and the `LearnerPathwayStepSerializer`, which is then used in the `LearnerPathwaySearchDocumentSerializer`.
Even if we apply `prefetch_related` in the `LearnerPathwayDocument's` `get_queryset`, the filtering in `get_linked_courses_and_course_runs` bypasses the prefetched results and runs additional queries.

To solve this issue: I have moved the filtering logic to `LearnerPathwayProgramViewSet's` `get_queryset` and `LearnerPathwayDocument's``get_queryset` method

Since we are already overriding the `LearnerPathwayProgramViewSet's` `get_queryset`, I have also fixed the existing N+1 issue in the viewset.

Similarly for courses, `LearnerPathwayCourseMinimalSerializer's` `get_course_runs` method was causing additional queries because of filters.
`LearnerPathwayCourseMinimalSerializer` is used by `LearnerPathwayCourseSerializer`
which is used by `LearnerPathwayCourseViewSet` and `LearnerPathwayStepSerializer` (used in the `LearnerPathwaySearchDocumentSerializer`)
To solve this issue: I have moved the filtering logic to `LearnerPathwayCourseViewSet's` `get_queryset` and LearnerPathwayDocument's `get_queryset` method
Since we are already overriding the `LearnerPathwayCourseViewSet's` `get_queryset`, I have also fixed the existing N+1 issue in the viewset.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -763,30 +763,20 @@ class AggregateSearchViewSetTests(mixins.SerializationMixin, mixins.LoginMixin, 
         )
         assert expected == actual
 
-    @ddt.data(True, False)
-    def test_learner_pathway_feature_flag(self, include_learner_pathways):
+    @ddt.data((True, 10, 8), (False, 0, 4))
+    @ddt.unpack
+    def test_learner_pathway_feature_flag(self, include_learner_pathways, expected_result_count, expected_query_count):
         """ Verify the include_learner_pathways feature flag works as expected."""
-        LearnerPathwayStepFactory.create_batch(5, pathway__partner=self.partner)
+        LearnerPathwayStepFactory.create_batch(10, pathway__partner=self.partner)
         pathways = LearnerPathway.objects.all()
-        assert pathways.count() == 5
+        assert pathways.count() == 10
         query = {
             'include_learner_pathways': include_learner_pathways,
         }
 
-        if include_learner_pathways:
-            expected_result_count = pathways.count()
-            expected_query_count = 8
-        else:
-            expected_result_count = 0
-            expected_query_count = 4
-
         with self.assertNumQueries(expected_query_count):
             response = self.get_response(query, self.list_path)
 
-        response = self.get_response(
-            query,
-            self.list_path
-        )
         assert response.status_code == 200
         response_data = response.json()
 
--- a/course_discovery/apps/api/v1/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_search.py
@@ -766,12 +766,22 @@ class AggregateSearchViewSetTests(mixins.SerializationMixin, mixins.LoginMixin, 
     @ddt.data(True, False)
     def test_learner_pathway_feature_flag(self, include_learner_pathways):
         """ Verify the include_learner_pathways feature flag works as expected."""
-        LearnerPathwayStepFactory(pathway__partner=self.partner)
+        LearnerPathwayStepFactory.create_batch(5, pathway__partner=self.partner)
         pathways = LearnerPathway.objects.all()
-        assert pathways.count() == 1
+        assert pathways.count() == 5
         query = {
             'include_learner_pathways': include_learner_pathways,
         }
+
+        if include_learner_pathways:
+            expected_result_count = pathways.count()
+            expected_query_count = 8
+        else:
+            expected_result_count = 0
+            expected_query_count = 4
+
+        with self.assertNumQueries(expected_query_count):
+            response = self.get_response(query, self.list_path)
 
         response = self.get_response(
             query,
@@ -780,11 +790,11 @@ class AggregateSearchViewSetTests(mixins.SerializationMixin, mixins.LoginMixin, 
         assert response.status_code == 200
         response_data = response.json()
 
+        assert response_data['count'] == expected_result_count
+
         if include_learner_pathways:
-            assert response_data['count'] == 1
-            assert response_data['results'][0] == self.serialize_learner_pathway_search(pathways[0])
-        else:
-            assert response_data['count'] == 0
+            for pathway in pathways:
+                assert self.serialize_learner_pathway_search(pathway) in response.data['results']
 
 
 class LimitedAggregateSearchViewSetTests(
--- a/course_discovery/apps/learner_pathway/api/serializers.py
+++ b/course_discovery/apps/learner_pathway/api/serializers.py
@@ -3,8 +3,6 @@ Serializers for learner_pathway app.
 """
 from rest_framework import serializers
 
-from course_discovery.apps.api.utils import get_excluded_restriction_types
-from course_discovery.apps.course_metadata.choices import CourseRunStatus
 from course_discovery.apps.learner_pathway import models
 
 
@@ -20,12 +18,7 @@ class LearnerPathwayCourseMinimalSerializer(serializers.ModelSerializer):
         fields = ('key', 'course_runs')
 
     def get_course_runs(self, obj):
-        excluded_restriction_types = get_excluded_restriction_types(self.context['request'])
-        return list(obj.course.course_runs.filter(
-            status=CourseRunStatus.Published
-        ).exclude(
-            restricted_run__restriction_type__in=excluded_restriction_types
-        ).values('key'))
+        return [{'key': course_run.key} for course_run in obj.course.course_runs.all()]
 
 
 class LearnerPathwayCourseSerializer(LearnerPathwayCourseMinimalSerializer):
@@ -87,8 +80,7 @@ class LearnerPathwayProgramSerializer(LearnerPathwayProgramMinimalSerializer):
         return program.card_image_url
 
     def get_courses(self, obj):
-        excluded_restriction_types = get_excluded_restriction_types(self.context['request'])
-        return obj.get_linked_courses_and_course_runs(excluded_restriction_types=excluded_restriction_types)
+        return obj.get_linked_courses_and_course_runs()
 
 
 class LearnerPathwayBlockSerializer(serializers.ModelSerializer):
--- a/course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
+++ b/course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
@@ -6,7 +6,9 @@ from pytest import mark
 from rest_framework import status
 
 from course_discovery.apps.core.tests.factories import UserFactory
-from course_discovery.apps.course_metadata.tests.factories import CourseRunFactory, RestrictedCourseRunFactory
+from course_discovery.apps.course_metadata.tests.factories import (
+    CourseFactory, CourseRunFactory, RestrictedCourseRunFactory
+)
 from course_discovery.apps.learner_pathway.choices import PathwayStatus
 from course_discovery.apps.learner_pathway.tests.factories import (
     LearnerPathwayCourseFactory, LearnerPathwayFactory, LearnerPathwayProgramFactory, LearnerPathwayStepFactory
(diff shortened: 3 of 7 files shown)


Ali-D-Akbar (approved):

DawoudSheraz (commented):
- commit type should be perf, not chore
- Add a relevant unit test to showcase the decrease in query count

course_discovery/apps/learner_pathway/api/serializers.py
```diff
@@ -87,8 +82,7 @@ def get_card_image_url(self, step):
         return program.card_image_url
 
     def get_courses(self, obj):
-        excluded_restriction_types = get_excluded_restriction_types(self.context['request'])
-        return obj.get_linked_courses_and_course_runs(excluded_restriction_types=excluded_restriction_types)
```
DawoudSheraz, line 91: why are these removed from serializer? This should be independent from document changes.

course_discovery/apps/learner_pathway/api/v1/urls.py
```diff
@@ -6,8 +6,8 @@
 router = routers.SimpleRouter()
 router.register(r'learner-pathway', views.LearnerPathwayViewSet)
 router.register(r'learner-pathway-step', views.LearnerPathwayStepViewSet)
-router.register(r'learner-pathway-course', views.LearnerPathwayCourseViewSet)
-router.register(r'learner-pathway-program', views.LearnerPathwayProgramViewSet)
+router.register(r'learner-pathway-course', views.LearnerPathwayCourseViewSet, basename='learner-pathway-course')
+router.register(r'learner-pathway-program', views.LearnerPathwayProgramViewSet, basename='learner-pathway-program')
```
DawoudSheraz, line 10: why is this needed?

course_discovery/apps/learner_pathway/api/v1/views.py
```diff
@@ -81,12 +84,23 @@ class LearnerPathwayCourseViewSet(ReadOnlyModelViewSet):
...
     # versions of this API should only support the system default, PageNumberPagination.
     pagination_class = ProxiedPagination
 
+    def get_queryset(self):
+        excluded_restriction_types = get_excluded_restriction_types(self.request)
+        return models.LearnerPathwayCourse.objects.prefetch_related(
+            Prefetch(
+                'course__course_runs', 
+                queryset=CourseRun.objects.filter(
+                        status=CourseRunStatus.Published
```
DawoudSheraz: why are we using Published status as the only check here?

course_discovery/apps/learner_pathway/models.py
```diff
@@ -299,23 +299,15 @@ def get_skills(self) -> [str]:
 
         return program_skills
 
-    def get_linked_courses_and_course_runs(self, excluded_restriction_types=None) -> [dict]:
+    def get_linked_courses_and_course_runs(self):
```
DawoudSheraz, line 301: same question, why is the being removed from here considering this is a model method?

course_discovery/apps/learner_pathway/api/v1/urls.py
hamza-56: The basename is automatically generated from the `queryset` attribute of the viewset if it exists. However, in our case, we removed the `queryset` attribute and added a `get_queryset` method instead. Therefore, we need to explicitly set the basename to avoid any errors.

course_discovery/apps/learner_pathway/models.py
hamza-56: This model method is used exclusively in the `LearnerPathwayProgramSerializer`.

The `LearnerPathwayProgramSerializer` is used by the `LearnerPathwayProgramViewSet` and the `LearnerPathwayStepSerializer`, which is then used in the `LearnerPathwaySearchDocumentSerializer`.

In this model method, the use of `.filter` and `.exclude` bypasses the prefetch cache, causing additional database queries and leading to an N+1 query problem. To address this, the filtering logic has been moved to the `get_queryset` method of both `LearnerPathwayProgramViewSet` and `LearnerPathwayDocument`.

course_discovery/apps/learner_pathway/api/v1/views.py
hamza-56: This is the same filter that was being used in `get_linked_courses_and_course_runs` and `get_course_runs` https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/learner_pathway/models.py#L313

https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/learner_pathway/api/serializers.py#L25

zawan-ila:
The changes look good, but are quite dense. I'll have another look at them before approval. Can you please satisfy codecov in the meanwhile? Also, ref

> Excludes Learnerpathway course_runs based on excluded_restriction_types in /api/v1/search/all

Does the current implementation not take care of this? I'd expect the filtering at the serializer level (in `get_courses` and `get_course_runs`) to have taken care of this. I believe the search/all endpoint uses these serializers too.

hamza-56:
> The changes look good, but are quite dense. I'll have another look at them before approval. Can you please satisfy codecov in the meanwhile? Also, ref
> 
> > Excludes Learnerpathway course_runs based on excluded_restriction_types in /api/v1/search/all
> 
> Does the current implementation not take care of this? I'd expect the filtering at the serializer level (in `get_courses` and `get_course_runs`) to have taken care of this. I believe the search/all endpoint uses these serializers too.

@zawan-ila You're correct—the current implementation already handles this because we're using these serializers in /search/all. I've updated the PR description.

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
```diff
@@ -93,11 +105,20 @@ def setUp(self):
...
             step=learner_pathway_step,
             program__uuid=LEARNER_PATHWAY_DATA['steps'][0]['programs'][0]['uuid'],
             program__title=LEARNER_PATHWAY_DATA['steps'][0]['programs'][0]['title'],
             program__subtitle=LEARNER_PATHWAY_DATA['steps'][0]['programs'][0]['short_description'],
+            program__courses=[self.learner_pathway_program_course],
+        )
+        CourseRunFactory(
+            course=self.learner_pathway_program_course,
+            key=LEARNER_PATHWAY_DATA['steps'][0]['programs'][0]['courses'][0]['course_runs'][0]['key'],
+            status='published',
```
zawan-ila, line 120: None of the tests in this class are changed. Why is this needed?

hamza-56: Existing test doesn't check for program -> courses -> course runs.

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
```diff
@@ -245,3 +266,285 @@ def test_learner_pathway_uuids_endpoint(self, query_params, response):
...
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+
+        self.view_url = f'/api/v1/learner-pathway-program/{self.learner_pathway_program.uuid}/'
+
+    def _verify_learner_pathway_program_data(self, api_response, expected_data):
+        """
+        Verify that learner pathway program api response matches the expected data.
+        """
+        data = api_response.json()
+        print('_verify_learner_pathway_program_data', data)
```
zawan-ila: remove this print

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
```diff
@@ -245,3 +266,285 @@ def test_learner_pathway_uuids_endpoint(self, query_params, response):
...
+        # Verify program data
+        assert data['uuid'] == expected_data['uuid']
+        assert data['title'] == expected_data['title']
+        assert data['short_description'] == expected_data['short_description']
+
+        print('_verify_learner_pathway_program_data', data)
```
zawan-ila: Remove this print

zawan-ila (approved):
Great work on this  🎉

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
hamza-56: Removed ✅

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
hamza-56: Removed ✅

course_discovery/apps/api/v1/tests/test_views/test_search.py
```diff
@@ -766,25 +766,35 @@ def test_results_include_aggregation_key(self):
...
             'include_learner_pathways': include_learner_pathways,
         }
 
+        if include_learner_pathways:
+            expected_result_count = pathways.count()
+            expected_query_count = 8
+        else:
+            expected_result_count = 0
+            expected_query_count = 4
+
```
DawoudSheraz: nit: instead of if-else, you can move the counts to ddt as well. As for pathways.count(), it would be better to have static explicit values instead of comparing against DB count.

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
```diff
@@ -245,3 +266,283 @@ def test_learner_pathway_uuids_endpoint(self, query_params, response):
         learner_pathway_uuids_url = f'/api/v1/learner-pathway/uuids/?{urlencode(query_params)}'
         api_response = self.client.get(learner_pathway_uuids_url)
         assert api_response.json() == response
+
+
+@mark.django_db
```
DawoudSheraz: nit: wondering if this is still needed as the test suite is using Django's TestCase, not unittest TestCase

course_discovery/apps/learner_pathway/models.py
```diff
@@ -299,23 +298,15 @@ def get_skills(self) -> [str]:
...
         for course in self.program.courses.all():
-            course_runs = list(
-                course.course_runs.filter(
-                    status=CourseRunStatus.Published
-                ).exclude(
-                    restricted_run__restriction_type__in=excluded_restriction_types
-                ).values('key')
-            )
-            courses.append({"key": course.key, "course_runs": course_runs})
+            course_runs = [{'key': course_run.key} for course_run in course.course_runs.all()]
```
DawoudSheraz, line 308: nit: why can't we use .values() here like before?

course_discovery/apps/api/v1/tests/test_views/test_search.py
hamza-56: Updated ✅

course_discovery/apps/learner_pathway/api/v1/tests/test_views.py
hamza-56: Updated ✅

course_discovery/apps/learner_pathway/models.py
hamza-56: `.values()` causes additional queries

Screenshot 2024-11-21 at 1 00 35 AM
Screenshot 2024-11-21 at 12 59 57 AM

DawoudSheraz (approved):

PR #4489, opened 2024-11-14
perf: add CompressedCacheResponseMixin to CourseRecommendations ViewSet

AfaqShuaib09:
[PROD-3562](https://2u-internal.atlassian.net/browse/PROD-3562)
-------

This PR adds `CompressedCacheResponseMixin` to `CourseRecommendationsViewSet`. `CompressedCacheResponseMixin` adds the functionality of key-value based caching.

Testing:
- On master branch, hit `course-recommendations` endpoint after commenting the following line
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/cache.py#L101 
- Look into query count from django toolbar. Hit course_run endpoint again. Query count will not decrease.
- Checkout to `afaq/prod-3562` and hit `course-recommendations` endpoint. You will notice Query count decreases.

`http://localhost:18381/api/v1/course_recommendations/{course-key}`

1st Call:
image

2nd Call:
image

### DIFF
--- a/course_discovery/apps/api/cache.py
+++ b/course_discovery/apps/api/cache.py
@@ -154,14 +154,14 @@ class CompressedCacheResponseMixin():
     list_cache_timeout = settings.REST_FRAMEWORK_EXTENSIONS['DEFAULT_CACHE_RESPONSE_TIMEOUT']
 
     @conditional_decorator(
-        settings.USE_API_CACHING,
+        lambda: settings.USE_API_CACHING,
         compressed_cache_response(key_func=list_cache_key_func, timeout=list_cache_timeout),
     )
     def list(self, request, *args, **kwargs):
         return super().list(request, *args, **kwargs)
 
     @conditional_decorator(
-        settings.USE_API_CACHING,
+        lambda: settings.USE_API_CACHING,
         compressed_cache_response(key_func=object_cache_key_func, timeout=object_cache_timeout),
     )
     def retrieve(self, request, *args, **kwargs):
--- a/course_discovery/apps/api/v1/tests/test_views/test_courses.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_courses.py
@@ -12,6 +12,7 @@ from django.db import IntegrityError
 from django.db.models.functions import Lower
 from django.db.models.query import Prefetch
 from django.db.models.signals import m2m_changed, pre_save
+from django.test import override_settings
 from edx_toggles.toggles.testutils import override_waffle_switch
 from rest_framework.reverse import reverse
 from testfixtures import LogCapture
@@ -2666,6 +2667,7 @@ class CourseViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin
         self.assertContains(response, 'Invalid HTML received: h1 tag is not allowed', status_code=400)
 
     @responses.activate
+    @override_settings(USE_API_CACHING=True)
     def test_recommendations(self):
         courses_sharing_program = CourseFactory.create_batch(2)
         ProgramFactory(courses=[self.course, *courses_sharing_program])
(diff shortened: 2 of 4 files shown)


Ali-D-Akbar (approved):

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -2684,4 +2684,9 @@ def test_recommendations(self):
         with self.assertNumQueries(19, threshold=3):
             url = reverse('api:v1:course_recommendations-detail', kwargs={'key': self.course.key})
             response = self.client.get(url)
-        assert response.status_code == 200
+            assert response.status_code == 200
+
+        with self.assertNumQueries(15, threshold=3):
```
DawoudSheraz: why is the diff so low?

AfaqShuaib09: Because of this setting
https://github.com/openedx/course-discovery/blob/master/course_discovery/settings/test.py#L41-L42

AfaqShuaib09: I'll remove this test change from the PR

DawoudSheraz: You can override the setting and then verify the diff in count.

AfaqShuaib09: Tied overriding the setting, but the diff in the query count is still the same as before

DawoudSheraz: Where did you add the override? It would need to be added at very start, only then the first call would be cached.

AfaqShuaib09: I've tried it two ways, overriding it before the function definition and before the class definition. Both result in the same query count as before.

i.e;

```python
    @responses.activate
    def test_recommendations(self):
```
and

```python
@ddt.ddt
@override_settings(USE_API_CACHING=True)
@pytest.mark.usefixtures('django_cache')
class CourseViewSetTests(SerializationMixin, ElasticsearchTestMixin, OAuth2Mixin, APITestCase):
```

DawoudSheraz: Please debug it further by checking the queries before and after.

AfaqShuaib09: yup, looking into it

AfaqShuaib09: I've figured out the issue. The `@conditional_decorator` was not behaving as expected because it evaluates the `settings.USE_API_CACHING` value at definition time, not at runtime. Even if `@override_settings` was used in the test, it didn't affect the decorator because the decorator was already evaluated and applied during the initial definition of the method.

I've updated the logic to fix the above issue. Please have a look now.

course_discovery/apps/api/cache.py
```diff
@@ -154,15 +154,15 @@ class CompressedCacheResponseMixin():
...
         compressed_cache_response(key_func=list_cache_key_func, timeout=list_cache_timeout),
     )
     def list(self, request, *args, **kwargs):
         return super().list(request, *args, **kwargs)
 
     @conditional_decorator(
-        settings.USE_API_CACHING,
-        compressed_cache_response(key_func=object_cache_key_func, timeout=object_cache_timeout),
+        lambda: settings.USE_API_CACHING,
+        compressed_cache_response(key_func=list_cache_key_func, timeout=list_cache_timeout),
```
DawoudSheraz: no, this should not be changed. retrieve is to get a single object, that's why the key_func is using object key fn.

AfaqShuaib09: Ah, mistakenly updated this during refactoring. I've reverted it back now

zawan-ila: We should keep the timeout as is too

zawan-ila (approved):

DawoudSheraz (approved):

PR #4493, opened 2024-11-21
feat: introduce new marketable field

hamzawaleed01:
Introduce new marketable field that B2B will use for assigning courses early to their learners.


course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -636,6 +636,7 @@ def get_expected_data(cls, course_run, request):
             'external_key': course_run.external_key,
             'is_enrollable': course_run.is_enrollable,
             'is_marketable': course_run.is_marketable,
+            'is_marketable_for_enterprise': course_run.is_marketable_for_enterprise,
```
hamzawaleed01: Need to come up with a more suitable name for this.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,29 @@ def is_marketable(self):
...
+
+        is_exec_ed_course = self.course.type.slug == CourseType.EXECUTIVE_EDUCATION_2U
+
+        if is_exec_ed_course:
+            is_in_review = self.in_review
+            is_not_marketable = not self.is_marketable
+            has_future_start_date = self.is_upcoming()
+            return is_in_review and is_not_marketable and has_future_start_date
+        else:
+            return self.is_marketable
```
hamzawaleed01: @adamstankiewicz  is this what you were talking about? 🤔

course_discovery/apps/api/tests/test_serializers.py
hamzawaleed01: Tagging @adamstankiewicz @iloveagent57 @bcitro for more ideas 🧠 ✨ and I _just_ came up with followings; not sure if these make any sense: 

1. `is_assignable_to_learners`
2. `is_ready_for_assignment`

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,29 @@ def is_marketable(self):
...
+    def is_marketable_external(self):
+        """
+        Determines if the course_run is suitable for external marketing.
+
+        A course run is deemed suitable for enterprise marketing if it is an
+        executive education (EE) course, the discovery service status is
+        'in-review', it is not currently marketable, and the course start date is in the future.
+
+        For already marketable course runs, the suitability for enterprise marketing
+        refers to the existing marketable flag.
```
adamstankiewicz: nit: we may want to avoid mentioning "enterprise" in the docstring, now that the property name no longer indicates it's specific to enterprise.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +657,25 @@ def test_data(self):
         expected = self.get_expected_data(course_run, request)
         assert serializer.data == expected
 
+    def test_marketable_status(self):
```
adamstankiewicz: nit: should the test name reflect that it's testing the `is_marketable_external` property, e.g. `test_is_marketable_external`? Rationale: the current name _could_ be suggesting the original `is_marketable`.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +657,25 @@ def test_data(self):
...
+        # Test for executive education course with in-review status, not marketable, and future start date
+        exec_ed_type = CourseTypeFactory(slug=CourseType.EXECUTIVE_EDUCATION_2U)
+        course_run_exec_ed = CourseRunFactory(
+            course=CourseFactory(type=exec_ed_type),
+            status=CourseRunStatus.InternalReview,
+            start=datetime.datetime.now(tz=UTC) + datetime.timedelta(days=10)  # Future start date
+        )
+        course_run_exec_ed.seats.set([])  # No seats to ensure is_marketable is False
```
adamstankiewicz: nit: EE course runs _do_ have seats in practice; should the seats be mocked not as an empty array, but with a seat where `is_marketable=False` to better reflect the actual data?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +657,25 @@ def test_data(self):
         expected = self.get_expected_data(course_run, request)
         assert serializer.data == expected
 
+    def test_marketable_status(self):
+        # Test for executive education course with in-review status, not marketable, and future start date
```
adamstankiewicz: Are all the core permutations of `is_marketable_external` covered by this test? For example, if an EE course already has `is_marketable: True`, the property is a pass-thru even for EE, but I'm not sure this is reflected in this test?

Would it make sense to extend it with `@ddt.data` to account for the different permutations (e.g., different `status` of "in-review" vs. "unpublished", etc.)?

hamzawaleed01: oh yes my bad, after the recent changes `is_marketable_external` property; also need to update the tests.

hamzawaleed01: Pushed latest changes!

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +658,32 @@ def test_data(self):
         expected = self.get_expected_data(course_run, request)
         assert serializer.data == expected
 
+    @ddt.data(
+        # Exec-ed, in-review, not marketable, future start date, is_marketable_external
+        (CourseType.EXECUTIVE_EDUCATION_2U, CourseRunStatus.InternalReview, False, True, True),
+        # Exec-ed, in-review, marketable, future start date, is_marketable_external
+        (CourseType.EXECUTIVE_EDUCATION_2U, CourseRunStatus.InternalReview, False, True, True),
```
iloveagent57: This is exactly the same as the case above, I think you wanted marketable to be `True` here?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +658,32 @@ def test_data(self):
...
+        # Exec-ed, unpublished, not marketable, future start date, is_marketable_external
+        (CourseType.EXECUTIVE_EDUCATION_2U, CourseRunStatus.Unpublished, False, True, False),
+        # Non-exec ed, in-review, not marketable, future start date, is_marketable_external
+        (CourseType.PROFESSIONAL, CourseRunStatus.InternalReview, False, True, False),
+        # Non-exec ed, published, marketable, future start date, is_marketable_external
+        (CourseType.VERIFIED_AUDIT, CourseRunStatus.Published, True, True, True),
```
iloveagent57: Maybe have the last two values set to `False` just to "prove" that if `is_marketable` is true, it doesn't matter what the other values are.

course_discovery/apps/api/tests/test_serializers.py
hamzawaleed01: Good catch! 
Yes, as per the code comment it was marketable BUT it was failing because it's not possible to have `status = in-review` and `marketable = True`, so changing both now.

course_discovery/apps/api/tests/test_serializers.py
hamzawaleed01: These two `CourseType.VERIFIED_AUDIT, CourseRunStatus.Published` already should result in `is_marketable_external=False` but it's still True that proves it's working as expected.
But I think I can also set future date to `False` and it should still work.

Actually last two values (marketable and is_marketable_external) are evaluated by their definition in model so changing them from `ddt.data` won't actually update them - but will result in failure of assertions.

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +658,33 @@ def test_data(self):
         expected = self.get_expected_data(course_run, request)
         assert serializer.data == expected
 
+    @ddt.data(
+        # Exec-ed, in-review, not marketable, future start date, is_marketable_external
+        (CourseType.EXECUTIVE_EDUCATION_2U, CourseRunStatus.InternalReview, False, True, True),
+        # Exec-ed, in-review, not marketable, future start date, is_marketable_external
+        (CourseType.EXECUTIVE_EDUCATION_2U, CourseRunStatus.LegalReview, False, True, True),
```
adamstankiewicz: nit: both comments describing each test case mention the same "in-review", despite the first being `CourseRunStatus.InternalReview` and the second being `CourseRunStatus.LegalReview`. Should the comments differentiate between what is meant by "in-review" more concretely?

course_discovery/apps/api/tests/test_serializers.py
```diff
@@ -656,6 +658,33 @@ def test_data(self):
         expected = self.get_expected_data(course_run, request)
         assert serializer.data == expected
 
+    @ddt.data(
+        # Exec-ed, in-review, not marketable, future start date, is_marketable_external
```
adamstankiewicz: nit: just a suggestion, it might help to use dicts to define the test case data with key/value pairs vs. relying on specific ordering of args where the helper comments, e.g. something like:

```python
@ddt.data(
    # Exec-ed, in-review, not marketable, future start date, is_marketable_external
    {
        "course_type": CourseType.EXECUTIVE_EDUCATION_2U,
        "status": CourseRunStatus.InternalReview,
        "is_marketable": False,
        "has_future_start_date": True,
        "expected_is_marketable_external": True,
    },
)
@ddt.unpack
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,28 @@ def is_marketable(self):
...
+
+        A course run is deemed suitable for marketing if it is an
+        executive education (EE) course, the discovery service status is
+        'in-review', it is not currently marketable, and the course start date is in the future.
+
+        For already marketable course runs, the suitability for marketing
+        refers to the existing marketable flag.
+        """
+
+        is_exec_ed_course = self.course.type.slug == CourseType.EXECUTIVE_EDUCATION_2U
```
adamstankiewicz: nit: this line could technically come after the check/return for `if self.is_marketable:`, as it's not used before then.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,28 @@ def is_marketable(self):
...
+        if self.is_marketable:
+            # Course run is already marketable; pass-thru regardless of course type
+            return self.is_marketable
+        if is_exec_ed_course:
+            # Check whether external course run is marketable
+            return self.in_review and self.is_upcoming()
```
adamstankiewicz: [sanity check/clarification] Are we confident the requirements for this property call for whether the course run is "in_review" regardless of the review state? For example, per the tests, both `CourseRunStatus.InternalReview` and  `CourseRunStatus.LegalReview` are used as "in_review" statuses; will a future EE variant be considered externally marketable for _either_ review state, or is only 1 of these review states valid for this use case?

I tried to find any notes about the review state documented explicitly in any of the tickets/docs around this work, but wasn't able to find a mention of the exact review state. Perhaps this is something that should be verified by the team who worked on ingesting EE variants, if it hasn't been verified already?

adamstankiewicz (commented):
Left a few comments, mostly minor cleanup nits but also a clarifying question around whether the intent is to support all reviewable states (internal review vs. legal review), or if there's a specific/individual reviewable state that should be used (might be worth verifying).

course_discovery/apps/course_metadata/models.py
hamzawaleed01: As per my understanding any of those two review states will be considered "in_review" because `in_review` itself is a [property](https://github.com/openedx/course-discovery/blob/c54de1f02c99a64060f3f1f4fc33c154025f6016/course_discovery/apps/course_metadata/models.py#L2542) not a field
```
@property
def in_review(self):
    return self.status in CourseRunStatus.REVIEW_STATES()
```

```
@classmethod
    def REVIEW_STATES(cls):
        return [cls.LegalReview.value, cls.InternalReview.value]
```

hamzawaleed01: But I agree it's a good idea to confirm with the relevant team.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,27 @@ def is_marketable(self):
...
+        For already marketable course runs, the suitability for marketing
+        refers to the existing marketable flag.
+        """
+        if self.is_marketable:
+            # Course run is already marketable; pass-thru regardless of course type
+            return self.is_marketable
+        is_exec_ed_course = self.course.type.slug == CourseType.EXECUTIVE_EDUCATION_2U
+        if is_exec_ed_course:
+            # Check whether external course run is marketable
+            return self.in_review and self.is_upcoming()
```
AfaqShuaib09: Hey @hamzawaleed01, it should be `self.status == CourseRunStatus.Reviewed` instead of `self.in_review`

The aim of future variants is that they are scheduled ones, meaning that these are reviewed ones following the publishing lifecycle (scheduled as future runs)

In course-discovery, `CourseRunStatus.Reviewed` --> `Scheduled`

hamzawaleed01: Got it! Thanks @AfaqShuaib09. I've pushed the changes, see if it's good to go.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -299,7 +299,7 @@ def test_course_runs_restriction(self, include_restriction_param):
         url = reverse('api:v1:course-detail', kwargs={'key': self.course.key})
         if include_restriction_param:
             url += '?include_restricted=custom-b2c'
-        with self.assertNumQueries(36, threshold=3):
+        with self.assertNumQueries(40, threshold=3):
```
AfaqShuaib09: Make sure we don't run into any N+1 issues because of these changes. You can check for duplicate queries using the django debug toolbar.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,27 @@ def is_marketable(self):
...
+
+        A course run is deemed suitable for marketing if it is an
+        executive education (EE) course, the discovery service status is
+        'Reviewed', it is not currently marketable, and the course start date is in the future.
+
+        For already marketable course runs, the suitability for marketing
+        refers to the existing marketable flag.
+        """
+        if self.is_marketable:
+            # Course run is already marketable; pass-thru regardless of course type
```
AfaqShuaib09: nit: remove these inline comments, the docstring is already self explanatory.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,27 @@ def is_marketable(self):
         is_published = self.status == CourseRunStatus.Published
         return is_published and self.seats.exists() and bool(self.marketing_url)
 
+    @property
+    def is_marketable_external(self):
+        """
+        Determines if the course_run is suitable for marketing.
```
adamstankiewicz: nit: 
```suggestion
        Determines if the course_run is suitable for external marketing.
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,27 @@ def is_marketable(self):
...
+
+        A course run is deemed suitable for marketing if it is an
```
adamstankiewicz: nit: 
```suggestion
        A course run is deemed suitable for external marketing if it is an
```

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,27 @@ def is_marketable(self):
...
+        executive education (EE) course, the discovery service status is
+        'Reviewed', it is not currently marketable, and the course start date is in the future.
```
adamstankiewicz: nit: "it is not currently marketable"... If it is marketable, we return early with `self.is_marketable` (the first conditional check below); there are no other checks on `self.is_marketable` afterwards as it relates to Exec Ed courses. Should the docstring more closely reflect this logic?

course_discovery/apps/api/v1/tests/test_views/test_courses.py
hamzawaleed0102: I've pushed some updates to reduce duplicate queries and also had to do a +1 in numqueries assertions since fetching `course.type` (FK) adds a query.
Tested courses and course_runs endpoint with limited data after provisioning command. Let me know if you spot any other duplicates.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,25 @@ def is_marketable(self):
...
+    @property
+    def is_marketable_external(self):
+        """
+        Determines if the course_run is suitable for external marketing.
+
+        If already marketable, simply return self.is_marketable.
+
+        Else, a course run is deemed suitable for external marketing if it is an
+        executive education (EE) course, the discovery service status is
+        'Reviewed', and the course start date is in the future.
```
AfaqShuaib09: nit: go_live_date & start_date is in future.

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/models.py
hamzawaleed01: Are you referring to the docstring only, or the actual logic as well? In my understanding, checking the `start_date` should be sufficient. If the `start_date` is in the future, then the `go_live_date` will naturally also be in the future. ? 🤔

course_discovery/apps/api/serializers.py
```diff
@@ -961,7 +961,7 @@ def prefetch_queryset(cls, queryset=None):
         # queryset passed in happens to be empty.
         queryset = queryset if queryset is not None else CourseRun.objects.all()
 
-        return queryset.select_related('course', 'type').prefetch_related(
+        return queryset.select_related('course', 'type', 'course__type').prefetch_related(
```
DawoudSheraz, line 964: if the type has been added here (and below), then why do so many tests have their query counts increased?

hamzawaleed01: I hope [this](https://github.com/openedx/course-discovery/pull/4493#discussion_r1871489656) answers your question

course_discovery/apps/course_metadata/models.py
DawoudSheraz: @hamzawaleed01 Generally speaking, it is possible for go_live_date to be in past but start to be in future. go_live_date is used to determine if a course run can be set to "Published". It does not mean the run has started. 
However, for the current ExecEd issue, it is possible for go_live and start are in future. However, you should handle the code appropriately to avoid unexpected surprises.

hamzawaleed01: no worries,, just wanted to ensure it's not a unnecessary check.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
DawoudSheraz: I dont think we should need +1 if we are adding the type in select_related. If it was prefetch_related, then +1 would have been needed. select_related does join on DB level.

course_discovery/apps/api/serializers.py
hamzawaleed01: These changes also fixed the duplicate queries to DB (tested by me and Afaq locally)

course_discovery/apps/api/v1/tests/test_views/test_courses.py
hamzawaleed01: Done sir 🙂 no more +1s now.

course_discovery/apps/course_metadata/models.py
```diff
@@ -2969,6 +2969,27 @@ def is_marketable(self):
...
+
+        If already marketable, simply return self.is_marketable.
+
+        Else, a course run is deemed suitable for external marketing if it is an
+        executive education (EE) course, the discovery service status is
+        'Reviewed', and the course go_live_date & start_date is in the future.
+        """
+        if self.is_marketable:
+            return self.is_marketable
+        is_exec_ed_course = self.course.type.slug == CourseType.EXECUTIVE_EDUCATION_2U
```
DawoudSheraz, line 2985: As a caution, when filtering on your end, do check for product_source (equals `2u`). While it can be added here, it will make things way too specific.

DawoudSheraz (approved):

PR #4496, opened 2024-11-26
perf: admin perf optimization for course-discovery

AfaqShuaib09:
[PROD-4225](https://2u-internal.atlassian.net/browse/PROD-4225)
-------
This PR adds the `get_queryset` method to Django Admin registered models. The aim is to optimize database queries and reduce query counts wherever possible by customizing the way queries are fetched in the admin interface. It will avoid unnecessary data fetching and reduce overhead.

### Screenshots
Before:
image

After:
image

Before:
image

After:
image

Before: 
image

After:
image

and so on...

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -3,6 +3,7 @@ from dal import autocomplete
 from django.conf import settings
 from django.contrib import admin, messages
 from django.contrib.admin.utils import model_ngettext
+from django.db.models import Prefetch
 from django.db.utils import IntegrityError
 from django.forms import CheckboxSelectMultiple, ModelForm
 from django.http import HttpResponseRedirect
@@ -85,6 +86,9 @@ class SeatInline(admin.TabularInline):
     readonly_fields = ('_upgrade_deadline', )
     raw_id_fields = ('draft_version', 'currency')
 
+    def get_queryset(self, request):
+        return super().get_queryset(request).select_related('draft_version', 'currency')
+
 
 class PositionInline(admin.TabularInline):
     model = Position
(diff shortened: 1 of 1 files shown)


DawoudSheraz (approved):

course_discovery/apps/course_metadata/admin.py
```diff
@@ -391,6 +409,23 @@ class ProgramAdmin(DjangoObjectActions, SimpleHistoryAdmin):
...
+            'partner',
+            'type'
+        ).prefetch_related(
+            'courses',
+            'authoring_organizations',
+            'credit_backing_organizations',
+            Prefetch(
+                'type__translations',
+                queryset=ProgramTypeTranslation.objects.filter(language_code='en'),
+                to_attr='prefetched_translations'
```
zawan-ila: Is this Prefetch helpful? I don't see how the `prefetched_translations` attribute would be accessed?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -661,6 +739,14 @@ class PersonAdmin(admin.ModelAdmin):
     readonly_fields = ('uuid',)
     search_fields = ('uuid', 'salutation', 'family_name', 'given_name', 'slug',)
 
+    def get_queryset(self, request):
+        queryset = super().get_queryset(request)
+        queryset = queryset.select_related(
```
zawan-ila: Rather than assigning queryset multiple times, it would be better to chain things like `return super.get_queryset.select_related(...)...`

course_discovery/apps/course_metadata/admin.py
```diff
@@ -697,6 +783,16 @@ class LevelTypeAdmin(SortableAdminMixin, TranslatableAdmin):
     search_fields = ('name_t',)
     fields = ('name_t',)
 
+    def get_queryset(self, request):
+        queryset = super().get_queryset(request)
+        queryset = queryset.prefetch_related(
+            Prefetch('translations', queryset=LevelTypeTranslation.objects.all(), to_attr='translated_names')
+        )
```
zawan-ila, line 792: [style] Same comment as above

zawan-ila (approved):

course_discovery/apps/course_metadata/admin.py
AfaqShuaib09: It's because of the dropdown field, when I retrieve a single Program object from Django admin. If I remove this change, there will be 31 similar queries in django debug toolbar.

image

AfaqShuaib09: Removing this will cause extra queries to fetch options in dropdown on retrieving a single object
image

AfaqShuaib09: However, I've overridden the `formfield_for_foreignkey` method now to eliminate the similar queries

PR #4497, opened 2024-11-28
fix: add partner field to person documents

zawan-ila:
## [PROD-4240](https://2u-internal.atlassian.net/browse/PROD-4240)

This PR adds the partner field to person documents. This fixes the issue of person objects not being returned in GET requests against `search/all` due to the filtering in [CatalogDataFilterDataMixin](https://github.com/openedx/course-discovery/blob/dce153407d4e09f067ca28fc982e52573090dac7/course_discovery/apps/edx_elasticsearch_dsl_extensions/mixins.py#L28-L32)

### DIFF
--- a/course_discovery/apps/course_metadata/search_indexes/documents/learner_pathway.py
+++ b/course_discovery/apps/course_metadata/search_indexes/documents/learner_pathway.py
@@ -30,10 +30,6 @@ class LearnerPathwayDocument(BaseDocument, OrganizationsMixin):
             'edge_ngram_completion': fields.TextField(analyzer=edge_ngram_completion),
         },
     )
-    partner = fields.TextField(
-        analyzer=html_strip,
-        fields={'raw': fields.KeywordField(), 'lower': fields.TextField(analyzer=case_insensitive_keyword)}
-    )
     visible_via_association = fields.BooleanField()
     status = fields.TextField()
     overview = fields.TextField()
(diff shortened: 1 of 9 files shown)


course_discovery/apps/course_metadata/search_indexes/documents/learner_pathway.py
```diff
@@ -30,10 +30,6 @@ class LearnerPathwayDocument(BaseDocument, OrganizationsMixin):
             'edge_ngram_completion': fields.TextField(analyzer=edge_ngram_completion),
         },
     )
-    partner = fields.TextField(
-        analyzer=html_strip,
-        fields={'raw': fields.KeywordField(), 'lower': fields.TextField(analyzer=case_insensitive_keyword)}
-    )
```
DawoudSheraz, line 36: why is this being removed?

course_discovery/apps/course_metadata/search_indexes/documents/person.py
```diff
@@ -29,6 +29,9 @@ class PersonDocument(BaseDocument):
     def prepare_aggregation_key(self, obj):
         return 'person:{}'.format(obj.uuid)
 
+    def prepare_partner(self, obj):
+        return obj.partner.short_code if obj.partner else ''
+
```
DawoudSheraz: should this not be added in common too?

zawan-ila: We could but it wouldn't be as neat since some documents have "[different](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/search_indexes/documents/course_run.py#L118)" implementations for it.

DawoudSheraz: If only course run has a different implementation, it makes sense to override it there. But if all others have the same pattern, it would be good to move this in common as you have done for field itself.

AfaqShuaib09 (approved):

DawoudSheraz (approved):

PR #4506, opened 2024-12-03
Add JWT roles based throttling mechanism

zawan-ila:
## [PROD-4139](https://2u-internal.atlassian.net/browse/PROD-4139)

This PR adds the ADR for JWT-role based throttling. The code is mostly copied from [4394](https://github.com/openedx/course-discovery/pull/4394) Thanks to @adamstankiewicz for the idea and the POC!


course_discovery/apps/api/tests/jwt_utils.py
```diff
@@ -5,18 +5,21 @@
...
         'iss': settings.JWT_AUTH['JWT_ISSUER'],
         'aud': settings.JWT_AUTH['JWT_AUDIENCE'],
         'username': user.username,
         'email': user.email,
         'iat': now,
         'exp': now + ttl
     }
+    if payload:
+        jwt_payload.update(payload)
+    return jwt_payload
```
AfaqShuaib09, line 22:
```suggestion
return jwt_payload.update(payload) if payload else jwt_payload
```

course_discovery/apps/core/throttles.py
```diff
@@ -28,10 +43,15 @@ def allow_request(self, request, view):
...
                 self.rate = user_throttle.rate
-                self.num_requests, self.duration = self.parse_rate(self.rate)
             except UserThrottleRate.DoesNotExist:
                 # If we don't have a custom user override, skip throttling if they are a privileged user
                 if user.is_superuser or user.is_staff or is_publisher_user(user):
                     return True
 
+                # If the user is not a privileged user, increase throttling rate if they are an enterprise user
+                if is_enterprise_user(request):
+                    self.rate = '400/hour'
```
AfaqShuaib09: nit: maybe we can use constant variable for throttle limit and refer it here

AfaqShuaib09 (approved):

course_discovery/settings/base.py
```diff
@@ -790,3 +790,8 @@
 )
 
 ENABLE_COURSE_REVIEWS_ADMIN = False
+
+# Keywords that will be searched for inside the `roles` key of the JWT in case a user uses JWT authentication.
+# If the keyword is found, the user has more lenient throttling limits.
+ENHANCED_THROTTLE_JWT_ROLE_KEYWORDS = ['enterprise']
```
zawan-ila: Unsure if this should be a list or string.

course_discovery/apps/core/throttles.py
zawan-ila: Done

course_discovery/apps/api/tests/jwt_utils.py
zawan-ila: dict.update is an in-place update and returns nothing

AfaqShuaib09: ah missed this, thanks

course_discovery/settings/base.py
adamstankiewicz: I think keeping this as a list makes sense, as it's fairly generic as is and could theoretically be used for non-enterprise purposes, too.

course_discovery/apps/core/throttles.py
```diff
@@ -16,6 +18,23 @@ def throttling_cache():
         return caches['default']
 
 
+def is_enterprise_user(request):
```
adamstankiewicz: nit: if using settings (i.e., `settings.ENHANCED_THROTTLE_JWT_ROLE_KEYWORDS` and `settings.ENHANCED_THROTTLE_LIMIT`, where these settings could theoretically be configured for other non-enterprise use cases, too, might it make sense to rename this method more generically, e.g. `is_priviledged_user_with_enhanced_throttle` or similar?

course_discovery/settings/base.py
adamstankiewicz: Also worth considering: it might make sense to define the base setting as an empty list (i.e., no change for the Open edX platform), and configure the setting via edx-internal such that the enhanced throttle rate change is only applicable for the edX.org instance. The `ENHANCED_THROTTLE_LIMIT` setting could stay hardcoded, though.

docs/decisions/0029-jwt-based-throttling.rst
```diff
@@ -0,0 +1,37 @@
...
+learner portal is intended to be used by regular enterprise learners and admins i.e those without any staff or privileged access.
+
+As of late, our regular throttling limits have proven to be a little too aggressive for some of these learners. As this is clearly 
+non-malicious traffic, we would like to find a way to set more lenient throttling limits for these learners.
+
+Decision
+----------
+Since these learners authenticate with Discovery by using JWT tokens issued by the LMS, we have decided to use the `roles` key
+in the JWT to identify them. Enterprise customers are guaranteed to have one of a small number of fixed roles assigned to them. 
+Once we identify that an incoming request's user has one of these roles in their JWT, we will enable higher rate limits for them.
```
adamstankiewicz, line 22: It may make sense to expand on the decision to capture the configurable settings in the ADR related to the change, and while the immediate use case is for enterprise learners, the approach is generic for non-enterprise use cases, too.

adamstankiewicz (approved):
LGTM, with a few minor nits/suggestions to consider. Edit: nice job providing an ADR to go along with the change as well!

Note, it looks like CI is currently failing at the moment.

iloveagent57 (approved):

course_discovery/apps/core/throttles.py
zawan-ila: Done.

course_discovery/settings/base.py
zawan-ila: Done

docs/decisions/0029-jwt-based-throttling.rst
zawan-ila: Documented the two new settings and their usage.

DawoudSheraz (approved):

PR #4508, opened 2024-12-04
fix: CSV loader reverting future runs to in LegalReview

AfaqShuaib09:
[PROD-4243](https://2u-internal.atlassian.net/browse/PROD-4243)
----------
This PR fixes an issue in the CSV loader where future runs were being reverted to the LegalReview state.

We have added a check in the CSV loader to complete the review cycle if a course run is in LegalReview.

Previously, changing the enrollment_end date moved the run to the LegalReview state as it is a non-exempt field. To address this behavior, we have added the above condition

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -266,12 +266,14 @@ class CSVDataLoader(AbstractDataLoader):
                     self._register_ingestion_error(CSVIngestionErrors.COURSE_RUN_UPDATE_ERROR, error_message)
                     continue
 
-            if course_run.status == CourseRunStatus.Unpublished:
-                course_run.refresh_from_db()
-                # Pushing the run into LegalReview is necessary to ensure that the
-                # url slug is correctly generated in subdirectory format
-                course_run.status = CourseRunStatus.LegalReview
-                course_run.save(update_fields=['status'], send_emails=False)
+            course_run.refresh_from_db()
+
+            if course_run.status in [CourseRunStatus.Unpublished, CourseRunStatus.LegalReview]:
+                if course_run.status == CourseRunStatus.Unpublished:
+                    # Pushing the run into LegalReview is necessary to ensure that the
+                    # url slug is correctly generated in subdirectory format
+                    course_run.status = CourseRunStatus.LegalReview
+                    course_run.save(update_fields=['status'], send_emails=False)
                 self._complete_run_review(row, course_run)
 
             logger.info("Course and course run updated successfully for course key {}".format(course_key))  # lint-amnesty, pylint: disable=logging-format-interpolation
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -266,12 +266,14 @@ def ingest(self):  # pylint: disable=too-many-statements
                     self._register_ingestion_error(CSVIngestionErrors.COURSE_RUN_UPDATE_ERROR, error_message)
                     continue
 
-            if course_run.status == CourseRunStatus.Unpublished:
-                course_run.refresh_from_db()
-                # Pushing the run into LegalReview is necessary to ensure that the
-                # url slug is correctly generated in subdirectory format
-                course_run.status = CourseRunStatus.LegalReview
-                course_run.save(update_fields=['status'], send_emails=False)
+            course_run.refresh_from_db()
```
AfaqShuaib09, line 269: Moved this `refresh_from_db()` above to get the updated status and complete the review cycle in one go

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -721,6 +723,69 @@ def test_ingest_flow_for_preexisting_unpublished_course(self, jwt_decode_patch):
...
+
+                    # Verify the existence of draft and non-draft
+                    assert Course.everything.count() == 2
+                    assert CourseRun.everything.count() == 2
+
+                    course = Course.everything.get(key=self.COURSE_KEY, partner=self.partner, draft=True)
+                    course_run = CourseRun.everything.get(course=course, draft=True)
+
+                    self._assert_course_data(course, {**self.BASE_EXPECTED_COURSE_DATA})
+                    assert course_run.status == CourseRunStatus.Reviewed
```
DawoudSheraz: nit: why is the status reviewed and not published? I don't see any dates data to assert this is a future variant.

AfaqShuaib09: It's because of this [_complete_run_review](https://github.com/openedx/course-discovery/blob/e5584c0fb6e72efd870820558983f7209615b588/course_discovery/apps/course_metadata/data_loaders/csv_loader.py#L275) which is calling [complete_review_phase](complete_review_phase) which is moving the course_run to reviewed Status. However code changes works fine, we can add a `refresh_from_db` call after the `_complete_run_review` statement to refresh the course_run status.

AfaqShuaib09: 🤦🏻 Figured out the issue, `go_live_date` is empty for the run because the run is in the `review_in_legal state`, so the course_run data doesn't get updated—only the status changes. In the subsequent run, the course_run data will be updated.

AfaqShuaib09: Now, the code is written in such a way that it ends up moving the course to the review_by_legal status when the command is completed

DawoudSheraz (approved):

zawan-ila (approved):

PR #4516, opened 2024-12-10
fix: Drop cache@v2 because it will break soon.

feanil:
See the announcement here: https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/


DawoudSheraz (approved):
Thanks for this.

PR #4524, opened 2024-12-26
Management command to archive courses

zawan-ila:
[PROD-4257](https://2u-internal.atlassian.net/browse/PROD-4257)
Adds a management command to archive a bunch of courses. Archiving a **course** entails the following:-

For **course** and its **draft version** :-

1. Set the status of all associated runs to unpublished
2. Set the `product_status` in associated AdditionalMetadata to archived
3. Set the `end_date` in associated AdditionalMetadata to the past
4. (Optional) Set the `end` for associated runs to the past
5. (Optional) Prepend `DELETED` to the title


course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,129 @@
...
+from course_discovery.apps.api.utils import StudioAPI
+from course_discovery.apps.course_metadata.choices import ExternalProductStatus
+from course_discovery.apps.course_metadata.emails import send_email_for_course_archival
+from course_discovery.apps.course_metadata.models import ArchiveCoursesConfig, Course, CourseRunStatus, CourseType
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Archive a collection of courses"
```
Ali-D-Akbar, line 30: Please elaborate how it does it and add an example of this command in the class docstring. Also, add a file docstring.

course_discovery/settings/base.py
```diff
@@ -796,3 +796,5 @@
 # If the keyword is found, the user has more lenient throttling limits.
 ENHANCED_THROTTLE_JWT_ROLE_KEYWORDS = []
 ENHANCED_THROTTLE_LIMIT = '400/hour'
+
+COURSE_ARCHIVAL_MAIL_RECIPIENTS = ['phoenix@edx.org']
```
Ali-D-Akbar, line 802: Ideally, move this to edx-internal.

Ali-D-Akbar (approved):

course_discovery/apps/api/utils.py
```diff
@@ -362,6 +362,19 @@ def push_to_studio(self, course_run, create=False, old_course_run_key=None, user
         else:
             self.update_course_run_details_in_studio(course_run)
 
+    def _update_end_date_in_studio(self, course_run):
+        """
+        We do not update end dates in studio after creation. However, while archiving a course, we
+        may wish to set the end date in the past. This method serves that purpose.
```
DawoudSheraz, line 368: nit: should we change enrollment_end here alongside end?

course_discovery/settings/base.py
AfaqShuaib09: +1 and @2u.com

course_discovery/apps/course_metadata/templates/course_metadata/email/course_archival.html
```diff
@@ -0,0 +1,42 @@
...
+    </table>
+</div>
+
+
+{% if failure_count > 0 %}
+<div>
+    <h3>Archival Failures</h3>
+        <ul>
+            {% for failure in failures %}
+                <li>[{{failure.uuid}}] {{failure.reason}}</li>
```
AfaqShuaib09: a `:` or `-` would be good here

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,129 @@
...
+        mangle_title = options.get('mangle_title')
+
+        if from_db:
+            courses = Course.objects.filter(uuid__in=self.get_uuids_from_database())
+        elif course_type:
+            courses = Course.objects.filter(
+                type__in=[CourseType.objects.get(slug__in=[course_type, course_type.lower()])]
+            )
+        else:
+            raise CommandError("Please provide one of --type or --from-db")
```
AfaqShuaib09, line 84: nit: we can add the test case to cover this scenario

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,129 @@
...
+        self.report = {
+            'failures': [],
+            'successes': [],
+            'total_count': 0
+        }
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--from-db',
+            help='Query the db for list of uuids to archive',
```
DawoudSheraz: This should not be just uuid, it should be all the possible arguments to this command.

course_discovery/apps/course_metadata/emails.py
```diff
@@ -416,3 +417,29 @@ def send_email_for_slug_updates(stats, to_users, subject=None):
     attachment.add_header('Content-Disposition', 'attachment', filename='slugs_update_summary.csv')
     email_msg.attach(attachment)
     email_msg.send()
+
+
+def send_email_for_course_archival(report, to_users):
+    """
+    Send an overall report of a archive_courses mgmt command run
```
DawoudSheraz, line 424: Apart from success count, we should at least add the successful courses's uuids and title in CSV. It will make it easier to see what has been archived. The report would be easy to share with relevant parties.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: I have added a file docstring that includes a usage example too.

course_discovery/settings/base.py
zawan-ila: I added it as a dummy value (hence the `edx.org` domain). We will override it in `edx-internal`

course_discovery/apps/api/utils.py
zawan-ila: I have no clear opinions here. I have added it though

course_discovery/apps/course_metadata/templates/course_metadata/email/course_archival.html
zawan-ila: Added a `:`

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: We store a csv with the uuids in a config model `ArchiveCoursesConfig`. We could put other args in there too, but I do not think it would be very useful. There are only 2 other options (mangle-title and mangle-end-dates) and I don't  see why one would need to toggle them real fast (which is what adding all options to the db provides) In fact, any single installation would only really be needing one version of the args, depending on what their "definition" of "archival" is

course_discovery/apps/course_metadata/emails.py
zawan-ila: Added a csv attachment with the `uuid`, `title`, `status`(success or failure) fields

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: I will be adding them before merge. coverage is also complaining about these

course_discovery/apps/course_metadata/management/commands/archive_courses.py
DawoudSheraz: In the past, mostly we have had all the possible options configurable via Config model. It is just for ease. I suspect we might need it when we test this on stage. Otherwise, to be able to change one argument, we need to change the code where the command is called. Having the liberty to change args from DB makes it easy to test out things.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: Done

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: > we need to change the code where the command is called

This effort must be weighed against the effort of changing the code and related tests and any local testing  required for the restructure. I do not think it is worth it once you factor that.

> Having the liberty to change args from DB makes it easy to test out things.

Is this necessarily true in any deployment infrastructure? i.e considering a non `edx.org` scenario

DawoudSheraz: > This effort must be weighed against the effort of changing the code and related tests and any local testing required for the restructure. I do not think it is worth it once you factor that.

Factoring the args in DB is much simpler as you are just reading data from DB and updating args dict. The tests also wont need much update. 

> Is this necessarily true in any deployment infrastructure? i.e considering a non edx.org scenario

It depends on the size of installation and the ease of server access to perform management command + file upload actions. In all the previous instances, any command config model that was added was primarily driven by edX/2U usage. I mean the idea of having csv files uploadable from config model was to avoid any inter-teams comms.

zawan-ila: I am still unconvinced here. Regardless, I have made the suggested update.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,163 @@
...
+            raise CommandError('Configuration object is not enabled')
+
+        if not config.csv_file:
+            raise CommandError('Configuration object does not have any input csv')
+
+        reader = unicodecsv.DictReader(config.csv_file)
+        uuid_list = reduce(lambda uuid_list, row: uuid_list + list(row.values()), reader, [])
+        return uuid_list
+
+    def get_csv_report(self):
```
AfaqShuaib09, line 155: nit: docstring missing

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,163 @@
...
+                        'uuid': course.uuid,
+                        'title': course_title
+                    }
+                )
+                logger.info(f"Successfully archived course with uuid: {course.uuid}")
+
+        send_email_for_course_archival(self.report, self.get_csv_report(), settings.COURSE_ARCHIVAL_MAIL_RECIPIENTS)
+
+    @transaction.atomic
+    def archive(self, course, mangle_end_date, mangle_title):
```
AfaqShuaib09, line 116: nit: docstring missing

AfaqShuaib09 (approved):

course_discovery/apps/api/utils.py
```diff
@@ -362,6 +362,19 @@ def push_to_studio(self, course_run, create=False, old_course_run_key=None, user
...
+        """
+        data = self.generate_data_for_studio_api(course_run, creating=False)
+        data.setdefault('schedule', {})
+        for attr in ['end', 'enrollment_end']:
+            attr_val = getattr(course_run, attr)
+            if attr_val:
+                data['schedule'][attr] = serialize_datetime(attr_val)
```
DawoudSheraz, line 375: Last we discussed, did we not plan to add start dates check in here i.e. log the run details if the start date is in future?

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,163 @@
...
+        for course_run in course.course_runs.all():
+            course_run.status = CourseRunStatus.Unpublished
+            course_run.save(update_fields=['status'])
+
+            if mangle_end_date and course_run.end and course_run.end > timezone.now():
+                course_run.end = timezone.now()
+            if mangle_end_date and course_run.enrollment_end and course_run.enrollment_end > timezone.now():
+                course_run.enrollment_end = timezone.now() - timedelta(days=1)
+            course_run.save(update_fields=['end', 'enrollment_end'])
```
DawoudSheraz: Last we discussed, did we not plan to add start dates check in here i.e. log the run details if the start date is in future?

course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
```diff
@@ -0,0 +1,157 @@
...
+            content_type='text/csv'
+        )
+        ArchiveCoursesConfigFactory.create(csv_file=self.csv_file, enabled=True)
+
+    @ddt.data(
+        *list(product([0, 1], repeat=3))
+    )
+    @ddt.unpack
+    @responses.activate
+    def test(self, from_db, mangle_title, mangle_end_date):
```
DawoudSheraz: nit: test_success

course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
```diff
@@ -0,0 +1,157 @@
...
+            assert responses.assert_call_count(self.courserun1_studio_url, 1) is True
+        else:
+            assert responses.assert_call_count(self.courserun1_studio_url, 0) is True
+
+    @ddt.data(
+        True,
+        False
+    )
+    @responses.activate
+    def test_some_failures(self, is_course1_archived):
```
DawoudSheraz: nit: test_failures should suffice

course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
```diff
@@ -0,0 +1,157 @@
...
+            assert is_end_date_mangled if mangle_end_date else not is_end_date_mangled
+
+    def verify_not_archived(self, course):
+        for c in [course, course.draft_version]:
+            assert c.additional_metadata.product_status == ExternalProductStatus.Published
+            assert c.additional_metadata.end_date > timezone.now() + timedelta(minutes=1)
+            assert c.course_runs.last().status == CourseRunStatus.Published
+
+            assert not c.title.startswith('DELETED')
+            assert not c.course_runs.first().end < timezone.now() + timedelta(minutes=1)
```
DawoudSheraz: nit: usually, any util methods in test suite are written before the unit tests. It's not a standard but it would be good to have consistency with the majority of other files.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: I have added the checks that `start<=end` and `enrollment_start<=enrollment_end`. I am not sure if we should add any checks for the combinations of those?

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: I don't see anything I could add in a docstring that would convey any useful information beyond the name.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: I don't see anything I could add in a docstring that would convey any useful information beyond the name.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,175 @@
...
+
+            # Push to studio to prevent RCM rewrite
+            if mangle_end_date:
```
zawan-ila, line 129: Earlier, I had the non-draft check here. But since we may have runs with draft versions only, I have removed that.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,175 @@
...
+        for record in self.report['successes']:
+            writer.writerow([record['uuid'], record['title'], 'success'])
+        for record in self.report['failures']:
+            writer.writerow([record['uuid'], record['title'], 'failure'])
+
+        return output.getvalue()
+
+    def verify_date_order(self, run):
+        # Start date is <= End date
+        date_list = list(filter(bool, [run.start, run.end]))
```
DawoudSheraz, line 168: curious: isn't this over-engineering-ish for 2 fields? What's the issue with having a simple comparison here?

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,175 @@
...
+
+            if mangle_end_date and course_run.end and course_run.end > timezone.now():
+                course_run.end = timezone.now()
+            if mangle_end_date and course_run.enrollment_end and course_run.enrollment_end > timezone.now():
+                course_run.enrollment_end = timezone.now() - timedelta(days=1)
+            course_run.save(update_fields=['end', 'enrollment_end'])
+
+            # Push to studio to prevent RCM rewrite
+            if mangle_end_date:
+                self.verify_date_order(course_run)
```
DawoudSheraz: The ordering looks incorrect. L121-L125 are updating the enrollment dates and then we are "verifying" the date order on L129. If the dates are in incorrect order, we don't need to do any field updates. The flow should be something like:

```
if mangle_end_date:
  try:
     self.verify_date_order(course_run)
  except ValueError:
   # some sort of logs, or can ignore if parent fn is taking care of it

    if course_run.end and course_run.end > timezone.now():
                course_run.end = timezone.now()
    if course_run.enrollment_end and course_run.enrollment_end > timezone.now():
                course_run.enrollment_end = timezone.now() - timedelta(days=1)
     course_run.save(update_fields=['end', 'enrollment_end'])
     api = StudioAPI(course_run.course.partner)
     api._update_end_date_in_studio(course_run)  # pylint: disable=protected-access
```

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: Say `enrollment_start` is None. Then a simple comparison would throw.

course_discovery/apps/course_metadata/management/commands/archive_courses.py
zawan-ila: Moved the `verify` call to before `save`

course_discovery/apps/course_metadata/emails.py
```diff
@@ -416,3 +417,30 @@ def send_email_for_slug_updates(stats, to_users, subject=None):
     attachment.add_header('Content-Disposition', 'attachment', filename='slugs_update_summary.csv')
     email_msg.attach(attachment)
     email_msg.send()
+
+
+def send_email_for_course_archival(report, csv_report, to_users):
+    """
+    Send an overall report of a archive_courses mgmt command run
```
DawoudSheraz, line 424: nit: report of an archive_courses mgmt command run

course_discovery/apps/course_metadata/management/commands/archive_courses.py
```diff
@@ -0,0 +1,175 @@
...
+
+            # Push to studio to prevent RCM rewrite
+            if mangle_end_date:
+                api = StudioAPI(course_run.course.partner)
+                api._update_end_date_in_studio(course_run)  # pylint: disable=protected-access
+
+        if course.additional_metadata:
+            course.additional_metadata.product_status = ExternalProductStatus.Archived
+            if course.additional_metadata.end_date and course.additional_metadata.end_date > timezone.now():
+                course.additional_metadata.end_date = timezone.now()
```
DawoudSheraz, line 136: nit: additional_metadata has registration_deadline (equivalent to enrollment_end), we can set that here too.

DawoudSheraz (approved):

PR #4525, opened 2024-12-26
feat: exclude retired coursetypes and runtypes from api

zawan-ila:
[PROD-4259](https://2u-internal.atlassian.net/browse/PROD-4259)
-------

Exclude courses and courseruns belonging to retired types in API responses.


course_discovery/apps/course_metadata/toggles.py
```diff
@@ -90,3 +90,12 @@
 IS_DISABLE_PRICE_UPDATES_FOR_PUBLISHED_RUNS = WaffleSwitch(
     'course_metadata.is_disable_price_updates_for_published_runs', __name__
 )
+# .. toggle_name: course_metadata.is_disable_price_updates_for_published_runs
+# .. toggle_implementation: WaffleSwitch
+# .. toggle_default: False
+# .. toggle_description: Hide retired courses and courseruns from apis.
+# .. toggle_creation_date: 2024-12-27
+# .. toggle_tickets: PROD-4259
+HIDE_RETIRED_COURSE_AND_COURSE_RUNS = WaffleSwitch(
```
DawoudSheraz: Any particular reason for using the waffle switch here? Looking at the code above, the data won't be returned unless the query param `include_retired` is provided and `RETIRED_COURSE_TYPES` is set. The waffle flag does not achieve a lot.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -110,6 +112,9 @@ def get_queryset(self):
             )
         else:
             queryset = queryset.filter(course__partner=partner)
+            if self.request.method == "GET" and HIDE_RETIRED_COURSE_AND_COURSE_RUNS.is_enabled() and not get_query_param(self.request, 'include_retired'):
+                retired_type_ids = list(map(lambda ct: ct.id, CourseRunType.objects.filter(slug__in=settings.RETIRED_RUN_TYPES)))
```
DawoudSheraz: include_retired is semantically confusing. Is it referencing retired product, archived product, or something else? We can be explicit here, include_retired_course_types or include_retired_product_type, etc.

course_discovery/apps/course_metadata/toggles.py
AfaqShuaib09: +1 on this

zawan-ila: I thought that it might take us some time to "officially retire" bootcamps after this PR is merged. To keep the behavior as it currently is in the meanwhile, I added this switch.

course_discovery/apps/api/v1/views/course_runs.py
zawan-ila: Done

course_discovery/apps/course_metadata/toggles.py
AfaqShuaib09: I think adding config settings at that time would be more sufficient

zawan-ila: Done

course_discovery/apps/api/v1/tests/test_views/test_courses.py
```diff
@@ -2589,13 +2623,20 @@ def test_update_geolocation__validation_error(self):
         assert GeoLocation.objects.count() == 1
 
     @responses.activate
-    def test_options(self):
+    @ddt.data(
+        [['bootcamp-2u'], False],
+        [[], True]
+    )
+    @ddt.unpack
+    def test_options(self, retired_course_types, is_bootcamp_in_result):
```
DawoudSheraz: since the underlying code for skipping type is generic (with addition of new setting vars), the param `is_bootcamp_in_result` can be renamed to is_retired_type_in_results, etc.

course_discovery/apps/api/v1/views/course_runs.py
```diff
@@ -110,6 +111,9 @@ def get_queryset(self):
             )
         else:
             queryset = queryset.filter(course__partner=partner)
+            if self.request.method == "GET" and not get_query_param(self.request, 'include_retired_run_types'):
+                retired_type_ids = list(map(lambda ct: ct.id, CourseRunType.objects.filter(slug__in=settings.RETIRED_RUN_TYPES)))
```
DawoudSheraz: Any particular reason for using lambda and not using values() that Django provides?

zawan-ila: Because I like lambdas :-) Values does seem better here, and I have updated to make use of it.

course_discovery/apps/api/v1/tests/test_views/test_courses.py
zawan-ila: I feel it is ok as is. The ddt should make things obvious here imo.

course_discovery/apps/api/v1/views/course_runs.py
DawoudSheraz: lambda is good but considering ORM provides the way here what we intend to do with lambda, better to use it (as you have done it)

course_discovery/apps/api/v1/tests/test_views/test_courses.py
DawoudSheraz: Hmmm, the test is test_options, and the `RETIRED_COURSE_TYPES` does not explicitly include bootcamp ref, so you can see how `is_bootcamp_in_result` does not quite fit with the context.

zawan-ila: The first ddt case(line 2627 here) sets `retired_course_types` to `['bootcamp-2u']` and the second sets it to `[]`. That puts `is_bootcamp_in_result` in context

DawoudSheraz: When overridding on L2637 override_settings(RETIRED_COURSE_TYPES=retired_course_types), it is setting what types are "retired". What I am saying here is that the code we have in API and in test is generic. we can add a third ddt and have verified as retired there. Once overridden in settings, those types wont be returned in options. The test is generic, and the code is generic. If we want a BootCamp explicit, we can rename the test to signify this is for BootCamp type.

zawan-ila: I still do not understand. I have changed the test to make use of the `audit` type. Let me know if there are still any issues.

AfaqShuaib09 (approved):

course_discovery/apps/api/v1/tests/test_views/test_courses.py
DawoudSheraz: Well, I will try to explain it better. The options code is irrespective of a specific type. We can define whatever type's slugs in `RETIRED_COURSE_TYPES` and when we hit API, the types mentioned in RETIRED_COURSE_TYPES wont be returned. The test here is same as that. It does not care if the type is audit, BootCamp, prof-ed. We provide a type in ddt and that type is skipped. That's why the 2nd argument's name (only) should be independent of type slug/name.

zawan-ila: Hmm. I have made the change, but I am not convinced that it is in any way an improvement.

DawoudSheraz (approved):
please squash the commits and add PR description before merge.

PR #4527, opened 2024-12-31
Update next be a complete string url

hamza-56:
### [PROD-4266](https://2u-internal.atlassian.net/browse/PROD-4266)

Update `next` be a complete string url (like v1) rather than a list in /api/v2/search/all


course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
```diff
@@ -122,16 +124,20 @@ def paginate_queryset(self, queryset, request, view=None):
...
-        """
-        response = super().get_paginated_response(data)
-        last_item = data[-1] if data else None
-        search_after = last_item.get("sort") if last_item else None
-        next_link = response.data.pop("next", None)
-        response.data["next"] = search_after if next_link else None
-        return response
+    def get_next_link(self):
+        if not self.page.has_next():
+            return None
```
DawoudSheraz, line 129: how is has_next calculated here?

hamza-56: `SearchAfterPagination` extends `PageNumberPagination` from django_elasticsearch_dsl_drf.pagination 
`PageNumberPagination` extends `pagination.PageNumberPagination` from rest_framework


ref: https://github.com/barseghyanartur/django-elasticsearch-dsl-drf/blob/master/src/django_elasticsearch_dsl_drf/pagination.py

DawoudSheraz: Yes but that does not answer how it will fit in with SearchAfter pagination where the next is determined from sort values.

hamza-56:
```
def has_next(self):
    return self.number < self.paginator.num_pages
```

The `has_next` method checks if there are additional pages available by comparing the current page number (`self.number`) with the total number of pages (`self.paginator.num_pages`).

**num_pages:** Calculates the total number of pages based on the total count of objects (self.count), adjusted for orphans, and divided by per_page. It uses ceil to ensure partial pages are counted as full pages.
**count:** Determines the total number of objects. If the dataset has a callable count method (e.g., a QuerySet in Django), it uses that for efficiency. Otherwise, it defaults to the length of the object list.


References: 

1. https://github.com/django/django/blob/main/django/core/paginator.py#L205
2. https://github.com/django/django/blob/main/django/core/paginator.py#L114

The current implementation does not include anything that should break the `has_next` calculation for search_after pagination in Elasticsearch

DawoudSheraz (approved):

course_discovery/apps/api/v2/tests/test_views/test_search.py
```diff
@@ -96,20 +102,28 @@ def test_search_after_pagination(self):
...
+        while next_url:
+            # Parse the `search_after` value from the next_url query params
+            parsed_url = urlparse(next_url)
+            query_params = parse_qs(parsed_url.query)
+            search_after = query_params.get("search_after", [None])[0]
+            assert search_after is not None, "'search_after' parameter is missing in the next_url"
 
-        while next_token:
-            response_data = self.fetch_page_data(page_size, search_after=json.dumps(next_token))
+            # Validate that the `search_after` matches the `sort` of the last result
```
Ali-D-Akbar: The comment seems unnecessary. Can be removed.

Ali-D-Akbar (approved):

course_discovery/apps/api/v2/tests/test_views/test_search.py
hamza-56: Removed

PR #4529, opened 2024-12-31
refactor: csv loader code implementation

AfaqShuaib09:
[PROD-4238](https://2u-internal.atlassian.net/browse/PROD-4238)
-------
This PR refactors the CSV loader code to improve readability, making it cleaner and easier to follow.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -4,6 +4,7 @@ creating and updating related objects in Studio, and ecommerce, provided a csv c
 """
 import csv
 import logging
+from functools import cache
 
 import unicodecsv
 from django.conf import settings
@@ -75,8 +76,17 @@ class CSVDataLoader(AbstractDataLoader):
             * product_source: slug of the external source that actually owns the product.
         """
         super().__init__(partner, api_url, max_workers, is_threadsafe)
-        self.error_logs = {}
-        self.ingestion_summary = {
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.ingestion_summary = self._initialize_ingestion_summary()
+        self.course_uuids = {}  # to show the discovery course ids for each processed course
+        self.product_type = product_type
+        self.product_source = self._get_product_source(product_source)
+        self.reader = self._initialize_csv_reader(csv_path, csv_file, use_gspread_client)
+        self.ingestion_summary['total_products_count'] = len(self.reader)
+
+    def _initialize_ingestion_summary(self):
+        """Initialize the ingestion summary dictionary."""
+        return {
             'total_products_count': 0,
             'success_count': 0,
             'failure_count': 0,
@@ -84,36 +94,35 @@ class CSVDataLoader(AbstractDataLoader):
             'created_products': [],
             'archived_products': []
         }
-        self.course_uuids = {}  # to show the discovery course ids for each processed course
-        self.product_type = product_type
+
+    def _get_product_source(self, product_source):
+        """
+        Retrieve the product source or raise an exception if product source doesn't exist already
+        """
         try:
-            self.product_source = Source.objects.get(slug=product_source)
+            return Source.objects.get(slug=product_source)
         except Source.DoesNotExist:
-            logger.exception(f"Unable to locate source with slug {product_source}")
+            logger.exception(f"Unable to locate source with slug '{product_source}'")
             raise
 
-        for error_log_key in CSV_LOADER_ERROR_LOG_SEQUENCE:
-            self.error_logs.setdefault(error_log_key, [])
-
+    def _initialize_csv_reader(self, csv_path, csv_file, use_gspread_client):
+        """
+        Initialize the CSV reader based on the input source (csv_path, csv_file or gspread_client)
+        """
         try:
             if use_gspread_client:
-                # TODO: add unit tests
-                product_type_config = settings.PRODUCT_METADATA_MAPPING[product_type][self.product_source.slug]
+                product_type_config = settings.PRODUCT_METADATA_MAPPING[self.product_type][self.product_source.slug]
                 gspread_client = GspreadClient()
-                self.reader = gspread_client.read_data(product_type_config)
+                return list(gspread_client.read_data(product_type_config))
             else:
-                # Read file from the path if given. Otherwise, read from the file
-                # received from CSVDataLoaderConfiguration.
-                self.reader = csv.DictReader(open(csv_path, 'r')) if csv_path \
-                    else list(unicodecsv.DictReader(csv_file))  # lint-amnesty, pylint: disable=consider-using-with
+                # read the file from the provided path; otherwise, use the file received from CSVDataLoaderConfiguration
+                return list(csv.DictReader(open(csv_path, 'r'))) if csv_path else list(unicodecsv.DictReader(csv_file))
         except FileNotFoundError:
-            logger.exception("Error opening csv file at path {}".format(csv_path))  # lint-amnesty, pylint: disable=logging-format-interpolation
-            raise  # re-raising exception to avoid moving the code flow
-        except Exception:
-            logger.exception("Error reading the input data source")
-            raise  # re-raising exception to avoid moving the code flow
-        self.reader = list(self.reader)
-        self.ingestion_summary['total_products_count'] = len(self.reader)
+            logger.exception(f"Error opening CSV file at path: {csv_path}")
+            raise
+        except Exception as e:
+            logger.exception(f"Error reading input data source: {e}")
+            raise
 
     def ingest(self):  # pylint: disable=too-many-statements
         logger.info("Initiating CSV data loader flow.")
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -75,45 +75,56 @@ def __init__(
...
         """
         super().__init__(partner, api_url, max_workers, is_threadsafe)
-        self.error_logs = {}
-        self.ingestion_summary = {
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.ingestion_summary = self._initialize_ingestion_summary()
+        self.course_uuids = {}  # to show the discovery course ids for each processed course
+        self.org_cache = {}
+        self.course_type_cache = {}
+        self.course_run_type_cache = {}
```
DawoudSheraz: We can combine these caches into a single dict cache

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -318,6 +289,158 @@ def ingest(self):  # pylint: disable=too-many-statements
...
+                CSVIngestionErrors.MISSING_REQUIRED_DATA,
+                CSVIngestionErrorMessages.MISSING_REQUIRED_DATA.format(
+                    course_title=course_title, missing_data=missing_fields
+                )
+            )
+            return False, course_type, course_run_type
+
+        return True, course_type, course_run_type
+
+    def _validate_course_and_course_run_types(self, row, course_title):
```
DawoudSheraz: This can be made an internal function of _validate_and_process_row as it is only being used in that function.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: Done

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: done

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -75,45 +75,56 @@ def __init__(
             * product_source: slug of the external source that actually owns the product.
         """
         super().__init__(partner, api_url, max_workers, is_threadsafe)
-        self.error_logs = {}
-        self.ingestion_summary = {
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.ingestion_summary = self._initialize_ingestion_summary()
```
zawan-ila, line 80: why not define the dict here? the function does nothing really

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -75,45 +75,56 @@ def __init__(
...
+        self.course_uuids = {}  # to show the discovery course ids for each processed course
+        self.obj_cache = {
+            'org': {}, 'course_type': {}, 'course_run_type': {}
+        } # cache for organization, course type and course run type objects to avoid redundant queries
```
zawan-ila: The comment seems out of place. Maybe put it at the top?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -128,78 +139,43 @@ def ingest(self):  # pylint: disable=too-many-statements
...
             course = Course.objects.filter_drafts(key=course_key, partner=self.partner).select_related('type').first()
             is_course_already_ingested = bool(course) and str(course.uuid) in self.course_uuids
             is_course_created = False
             is_course_run_created = False
-            course_run_restriction = (
-                None
-                if row.get('restriction_type', None) == 'None'
-                else row.get('restriction_type', None)
-            )
+            course_run_restriction = self._get_course_run_restriction(row)
```
zawan-ila, line 150: Personally, I do not think one should be swapping simple statements for functions. It just causes one to lose their head trying to find the definition in the source file.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -318,6 +289,157 @@ def ingest(self):  # pylint: disable=too-many-statements
...
+
+        Args:
+            course_type_name (str): Course type name
+
+        Returns:
+            CourseType: CourseType object
+        """
+        if course_type_name not in self.obj_cache['course_type']:
+            try:
+                self.obj_cache['course_type'][course_type_name] = CourseType.objects.get(name=course_type_name)
```
zawan-ila: Better to use a decorator, if we can, for this purpose. `functools.cache` (or any of its siblings) may be useful here.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -318,6 +289,157 @@ def ingest(self):  # pylint: disable=too-many-statements
...
+                        course_title=course_title, course_type=row["course_enrollment_track"]
+                    ),
+                )
+                return False, None, None
+
+            course_run_type = self.get_course_run_type(row["course_run_enrollment_track"])
+            if not course_run_type:
+                self._log_and_register_ingestion_error(
+                    CSVIngestionErrors.MISSING_COURSE_RUN_TYPE,
+                    CSVIngestionErrorMessages.MISSING_COURSE_RUN_TYPE.format(
```
zawan-ila, line 419: [nit] Can we shorten the function name a bit? Is is possible to use the attributes directly i.e `MISSING_COURSE_RUN_TYPE` instead of `CSVIngestionErrors.MISING_COURSE_RUN_TYPE`?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: to keep the `__init__` function clean, it looks nicer to me this way.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: sure, lemme do it

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: I've converted it into a method so we can use it in other places as well, such as the create_course_run API method call. I can revert this if you'd like.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: sure, renaming the method to `_log_ingestion_error`. 


> Is it possible to use the attributes directly i.e MISSING_COURSE_RUN_TYPE instead of CSVIngestionErrors.MISING_COURSE_RUN_TYPE?

Yup, but then need to explicitly import these variables from constants instead of simply `import CSVIngestionErrorMessages & CSVIngestionErrors`

AfaqShuaib09: `CSVIngestionErrorMessages` and `CSVIngestionErrors codes` both have some variables with the same name, so I think it's better to keep it as it is.

i.e;
```python
CSVIngestionErrors.MISSING_COURSE_RUN_TYPE
CSVIngestionErrorMessages.MISSING_COURSE_RUN_TYPE
```

course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -139,3 +140,25 @@ def map_external_org_code_to_internal_org_code(external_org_code, product_source
             f'product_source {product_source}'
         )
         return external_org_code
+
+
+def instance_cache(cache_key):
```
DawoudSheraz: did you try functools cache? It should work out of box instead of having to write a decorator.

AfaqShuaib09: Nope, but now after looking into the docs lru_cache seems to be reasonable for our use-case. https://docs.python.org/3/library/functools.html#functools.lru_cache

But there is one issue if we define it on class methods it will include `self` arg in the cache as well. Which is unhanshable by default so we need to define the [__eq__()](https://docs.python.org/3/reference/datamodel.html#object.__eq__) and [__hash__()](https://docs.python.org/3/reference/datamodel.html#object.__hash__) methods so that the cache can detect relevant attribute updates.
see here: https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls

else we can define staticmethods and add a wrapper over them.
```python
    @staticmethod
    @lru_cache(maxsize=None)
    def _get_course_type_static(course_type_name):
         try:
            return CourseType.objects.get(name=course_type_name)
        except CourseType.DoesNotExist:
            return None
            
   def get_course_type(self, course_type_name):
       return self._get_course_type_static(course_type_name)
```

zawan-ila: I like this approach. It would be better to add `staticmethod` directly to `get_course_type`

AfaqShuaib09: best lemme update it then

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -319,19 +333,19 @@ def _validate_organization(self, org_key, course_title):
...
+                    course_title=course_title, org_key=org_key
+                )
             )
             return False
+        return True
 
-    @instance_cache('course_type')
-    def get_course_type(self, course_type_name):
+    @staticmethod
+    @lru_cache(maxsize=None)
```
zawan-ila: An `lru_cache` with `maxsize=None` is the same as a simple `functools.cache`. Why not use it directly?

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -438,6 +453,16 @@ def _log_ingestion_error(self, error_code, message):
         logger.error(message)
         self._register_ingestion_error(error_code, message)
 
+    @classmethod
+    def clear_caches(cls):
```
zawan-ila, line 455: Why do we need to clear the caches?

AfaqShuaib09: Because functools caches are global caches instead of instance-level caches, I believe it's better to clear the caches after a job has been completed. If we don't do that, we would need to explicitly clear them in testsuites.

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: Ah, I mixed this up with `cached_property`, which has the following limitation

image

DawoudSheraz (approved):

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
zawan-ila: If the only usage is in the tests, maybe add a comment indicating that.

zawan-ila (approved):

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
AfaqShuaib09: It's kind of used for tests as well, it's better to clear all the method-level caches after the execution of the job therefore I called this at the end of ingest method

PR #4533, opened 2025-01-06
chore: prevent creation of retired product types from admin

zawan-ila:
[PROD-4273](https://2u-internal.atlassian.net/browse/PROD-4273)
-------

Do not show retired course and courserun types while creating a course or courserun from the django admin.

### DIFF
--- a/course_discovery/apps/course_metadata/forms.py
+++ b/course_discovery/apps/course_metadata/forms.py
@@ -1,10 +1,11 @@
 from django import forms
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.forms.utils import ErrorList
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/forms.py
```diff
@@ -150,6 +153,10 @@ class Meta:
                 },
             ),
         }
+    def __init__(self, *args, **kwargs):
```
AfaqShuaib09, line 157: nit: can we add unittests for these changes?

zawan-ila: Added

Ali-D-Akbar (approved):

AfaqShuaib09 (approved):

PR #4536, opened 2025-01-08
feat: add SearchAfterMixin for ES search_after capability

Ali-D-Akbar:
[PROD-4233](https://2u-internal.atlassian.net/browse/PROD-4233)
Adds a new `SearchAfterMixin` to be added in place of `PkSearchableMixin` that allows using search_after. Once this mixin is used, it will bypass the default search limit of 10k by making multiple calls to ES in case we have more than 10k records in an index. 

Previously, we faced an issue regarding the search limit, resulting in less records to be returned. We increased the `MAX_RESULT_WINDOW` before but a better way is to use search_after capability for an optimal and flexible result.

This PR also adds a v2 `CatalogQueryContainsViewSet` that fixes the querying mechanism by filtering the items at the time of when we're executing queries on ES. In v1, our `CatalogQueryContainsViewSet` first searched all the records AND THEN filtered it. 

Testing Instructions:
1. Run `update_index` locally in Discovery shell.
2. Visit `/api/v1/catalog/query_contains/` endpoint and add a sample query like this: http://localhost:18381/api/v2/catalog/query_contains/?course_uuids=2de67490-f748-4efd-8532-b445f7ecc6f9,f9f1e100-668a-4fd5-a966-a127de1f69de&query=org:edX

You can set `ELASTICSEARCH_DSL_QUERYSET_PAGINATION` to your specific value in order to test the behavior. While the end result will be the same but this can affect the number of times the search_after mechanism is called.

### DIFF
--- a/course_discovery/apps/api/v2/tests/test_views/test_catalog_queries.py
+++ b/course_discovery/apps/api/v2/tests/test_views/test_catalog_queries.py
@@ -1,0 +1,99 @@
+import urllib
+
+from rest_framework.reverse import reverse
+
+from course_discovery.apps.api.v1.tests.test_views.mixins import APITestCase
+from course_discovery.apps.core.tests.factories import UserFactory
+from course_discovery.apps.core.tests.mixins import ElasticsearchTestMixin
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseRunFactory
+
+
+class CatalogQueryViewSetTests(ElasticsearchTestMixin, APITestCase):
+    """
+    Unit tests for CatalogQueryViewSet.
+    """
+    def setUp(self):
+        super().setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.client.force_authenticate(self.user)
+        self.course = CourseFactory(partner=self.partner, key='simple_key')
+        self.course_run = CourseRunFactory(course=self.course, key='simple/key/run')
+        self.url_base = reverse('api:v2:catalog-query_contains')
+        self.error_message = 'CatalogQueryContains endpoint requires query and identifiers list(s)'
+        self.refresh_index()
+
+    def test_contains_single_course_run(self):
+        """ Verify that a single course_run is contained in a query. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:' + self.course_run.key,
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = f'{self.url_base}/?{qs}'
+        response = self.client.get(url)
+        assert response.status_code == 200
+        assert response.data == {self.course_run.key: True, str(self.course.uuid): False}
+
+    def test_contains_single_course(self):
+        """ Verify that a single course is contained in a query. """
+        qs = urllib.parse.urlencode({
+            'query': 'key:' + self.course.key,
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = f'{self.url_base}/?{qs}'
+        response = self.client.get(url)
+        assert response.status_code == 200
+        assert response.data == {self.course_run.key: False, str(self.course.uuid): True}
+
+    def test_contains_course_and_run(self):
+        """ Verify that both the course and the run are contained in the broadest query. """
+        self.course.course_runs.add(self.course_run)
+        self.course.save()
+        qs = urllib.parse.urlencode({
+            'query': 'org:*',
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = f'{self.url_base}/?{qs}'
+        response = self.client.get(url)
+        assert response.status_code == 200
+        assert response.data == {self.course_run.key: True, str(self.course.uuid): True}
+
+    def test_no_identifiers(self):
+        """ Verify that a 400 status is returned if request does not contain any identifier lists. """
+        qs = urllib.parse.urlencode({
+            'query': 'id:*'
+        })
+        url = f'{self.url_base}/?{qs}'
+        response = self.client.get(url)
+        assert response.status_code == 400
+        assert response.data == self.error_message
+
+    def test_no_query(self):
+        """ Verify that a 400 status is returned if request does not contain a querystring. """
+        qs = urllib.parse.urlencode({
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = f'{self.url_base}/?{qs}'
+        response = self.client.get(url)
+        assert response.status_code == 400
+        assert response.data == self.error_message
+
+    def test_incorrect_queries(self):
+        """ Verify that a 400 status is returned if request contains incorrect query string. """
+        failed_query_error_message = {
+            "detail": "Failed to make Elasticsearch request. Got exception: RequestError(400, "
+                      "'search_phase_execution_exception', 'Failed to parse query [(title:)]')"
+        }
+        qs = urllib.parse.urlencode({
+            'query': 'title:',
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = f'{self.url_base}/?{qs}'
+
+        response = self.client.get(url)
+        assert response.status_code == 400
+        assert response.data == failed_query_error_message
--- a/course_discovery/apps/api/v2/views/catalog_queries.py
+++ b/course_discovery/apps/api/v2/views/catalog_queries.py
@@ -1,0 +1,75 @@
+import logging
+from uuid import UUID
+
+from elasticsearch_dsl.query import Q as ESDSLQ
+from rest_framework import status
+from rest_framework.generics import GenericAPIView
+from rest_framework.permissions import DjangoModelPermissions, IsAuthenticated
+from rest_framework.response import Response
+
+from course_discovery.apps.api.mixins import ValidElasticSearchQueryRequiredMixin
+from course_discovery.apps.course_metadata.models import Course, CourseRun, SearchAfterMixin
+from course_discovery.apps.course_metadata.search_indexes.documents import CourseDocument, CourseRunDocument
+
+log = logging.getLogger(__name__)
+
+
+class CatalogQueryContainsViewSet(ValidElasticSearchQueryRequiredMixin, GenericAPIView, SearchAfterMixin):
+    permission_classes = (IsAuthenticated, DjangoModelPermissions)
+    queryset = Course.objects.all()
+
+    def get(self, request):
+        """
+        Determine if a set of courses and/or course runs is found in the query results.
+
+        Returns
+            dict:  mapping of course and run identifiers included in the request to boolean values
+                indicating whether the associated course or run is contained in the queryset
+                described by the query found in the request.
+        """
+        query = request.GET.get('query')
+        course_run_ids = request.GET.get('course_run_ids', None)
+        course_uuids = request.GET.get('course_uuids', None)
+        partner = self.request.site.partner
+
+        if query and (course_run_ids or course_uuids):
+            log.info(
+                f"Attempting search against query {query} with course UUIDs {course_uuids} "
+                f"and course run IDs {course_run_ids}"
+            )
+            identified_course_ids = set()
+            specified_identifiers = []
+            if course_run_ids:
+                course_run_ids = course_run_ids.split(',')
+                specified_identifiers = course_run_ids
+                identified_course_ids.update(
+                    self.search(
+                        query,
+                        queryset=CourseRun.objects.all(),
+                        partner=ESDSLQ('term', partner=partner.short_code),
+                        identifiers=ESDSLQ('terms', **{'key.raw': course_run_ids}),
+                        document=CourseRunDocument
+                    ).values_list('key', flat=True)
+                )
+
+            if course_uuids:
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                specified_identifiers += course_uuids
+
+                log.info(f"Specified course ids: {specified_identifiers}")
+                identified_course_ids.update(
+                    self.search(
+                        query,
+                        queryset=Course.objects.all(),
+                        partner=ESDSLQ('term', partner=partner.short_code),
+                        identifiers=ESDSLQ('terms', **{'uuid': course_uuids}),
+                        document=CourseDocument
+                    ).values_list('uuid', flat=True)
+                )
+            log.info(f"Identified {len(identified_course_ids)} course ids: {identified_course_ids}")
+
+            contains = {str(identifier): identifier in identified_course_ids for identifier in specified_identifiers}
+            return Response(contains)
+        return Response(
+            'CatalogQueryContains endpoint requires query and identifiers list(s)', status=status.HTTP_400_BAD_REQUEST
+        )
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1151,6 +1151,85 @@ class PkSearchableMixin:
         return filtered_queryset
 
 
+class SearchAfterMixin:
+    """
+    Represents objects to query Elasticsearch with `search_after` pagination and load by primary key.
+    """
+
+    @classmethod
+    def search(
+        cls,
+        query,
+        queryset=None,
+        page_size=settings.ELASTICSEARCH_DSL_QUERYSET_PAGINATION,
+        partner=None,
+        identifiers=None,
+        document=None,
+        sort_field="id",
+    ):
+        """
+        Queries the Elasticsearch index with optional pagination using `search_after`.
+
+        Args:
+            query (str) -- Elasticsearch querystring (e.g. `title:intro*`)
+            queryset (models.QuerySet) -- base queryset to search, defaults to objects.all()
+            page_size (int) -- Number of results per page.
+            partner (object) -- To be included in the ES query.
+            identifiers (object) -- UUID or key of a product.
+            document (object) -- ES Document for the given model.
+            sort_field (str) -- Field to sort results by. Defaults to 'id'.
+
+        Returns:
+            QuerySet
+        """
+        query = clean_query(query)
+        queryset = queryset or cls.objects.all()
+
+        if query == '(*)':
+            # Early-exit optimization. Wildcard searching is very expensive in elasticsearch. And since we just
+            # want everything, we don't need to actually query elasticsearch at all.
+            return queryset
+
+        logger.info(f"Attempting Elasticsearch document search against query: {query}")
+        es_document = document or next(iter(registry.get_documents(models=(cls,))), None)
+
+        must_queries = [ESDSLQ('query_string', query=query, analyze_wildcard=True)]
+        if partner:
+            must_queries.append(partner)
+        if identifiers:
+            must_queries.append(identifiers)
+
+        dsl_query = ESDSLQ('bool', must=must_queries)
+
+        all_ids = set()
+        search_after = None
+
+        while True:
+            search = (
+                es_document.search()
+                .query(dsl_query)
+                .sort(sort_field)
+                .extra(size=page_size)
+            )
+
+            search = search.extra(search_after=search_after) if search_after else search
+
+            results = search.execute()
+
+            ids = {result.pk for result in results}
+            if not ids:
+                logger.info("No more results found.")
+                break
+
+            all_ids.update(ids)
+            search_after = results[-1].meta.sort if results[-1] else None
+            logger.info(f"Fetched {len(ids)} records; total so far: {len(all_ids)}")
+
+        filtered_queryset = queryset.filter(pk__in=all_ids)
+        logger.info(f"Filtered queryset of size {len(filtered_queryset)} for query: {query}")
+        return filtered_queryset
+
+
 class Collaborator(TimeStampedModel):
     """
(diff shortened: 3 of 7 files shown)


course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4192,3 +4194,86 @@ def test_basic(self):
...
+        mock_result1.meta.sort = ["sort1"]
+
+        mock_result2 = MagicMock()
+        mock_result2.pk = 2
+        mock_result2.meta.sort = ["sort2"]
+
+        mock_search.execute.side_effect = [
+            [mock_result1, mock_result2],
+            [],
+        ]
```
DawoudSheraz: This is not the right way to test. This is mocking everything, including ES search after behavior. All the ES related tests in course-discovery hit test ES instance. The search after mixin tests should aim for that too.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1149,6 +1149,67 @@ def search(cls, query, queryset=None):
...
+                logger.warning(f"Elasticsearch request failed: {e}")
+                break
+
+            ids = {result.pk for result in results}
+            if not ids:
+                logger.info("No more results found.")
+                break
+
+            all_ids.update(ids)
+            search_after = results[-1].meta.sort
```
AfaqShuaib09: add safety check here, like
```python
search_after = results[-1].meta.sort if results[-1] else None
```
Also, have you tried it locally? By default, `results[-1].meta.sort` is a `Elastic.AttrList`, which is not json serializable. Using `results[-1].meta.sort.copy()` will convert it into a simple list

Ali-D-Akbar: Nice catch for the search_after check. 
I've tested this locally, I'm using this `search_after` value just to be sent back to the ES query execution.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: Updated the tests to use `ElasticsearchTestMixin` instead.

course_discovery/apps/api/v2/tests/test_views/test_catalog_queries.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Unit tests for CatalogQueryViewSet.
+    """
+    def setUp(self):
+        super().setUp()
+        self.user = UserFactory(is_staff=True, is_superuser=True)
+        self.client.force_authenticate(self.user)
+        self.course = CourseFactory(partner=self.partner, key='simple_key')
+        self.course_run = CourseRunFactory(course=self.course, key='simple/key/run')
+        self.url_base = reverse('api:v2:catalog-query_contains')
```
Ali-D-Akbar, line 21: Aside from this line, this is just a copy of `course_discovery/apps/api/v1/tests/test_views/test_catalog_queries.py`

course_discovery/apps/api/v2/views/catalog_queries.py
```diff
@@ -0,0 +1,72 @@
...
+                course_uuids = [UUID(course_uuid) for course_uuid in course_uuids.split(',')]
+                specified_course_ids += course_uuids
+
+                log.info(f"Specified course ids: {specified_course_ids}")
+                identified_course_ids.update(
+                    Course.search(
+                        query,
+                        partner=ESDSLQ('term', partner=partner.short_code),
+                        identifiers=ESDSLQ('terms', **{'uuid': course_uuids})
+                    ).values_list('uuid', flat=True)
```
Ali-D-Akbar: Aside from this code that updates the querying logic for ES (so rather than fetching all the products AND THEN filter it, we're fetching only the necessary products), this is just a copy of https://github.com/openedx/course-discovery/blob/a948a741b1839e845e07c762bab6ee51622daea5/course_discovery/apps/api/v1/views/catalog_queries.py

zawan-ila: Is this change not compatible with v1?

Ali-D-Akbar: It can be, yes.

Ali-D-Akbar: Requires changing the `pkSearchableMixin` a bit as well if we were to keep a single `CatalogQueryContainsViewSet`.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4192,3 +4195,21 @@ def test_basic(self):
...
+            CourseFactory()
+
+    @patch("course_discovery.apps.course_metadata.models.registry.get_documents")
+    def test_fetch_all_courses(self, mock_get_documents):
+        query = "Course*"
+        mock_get_documents.return_value = [CourseDocument]
+
+        queryset = CourseProxy.search(query=query)
+
+        self.assertEqual(len(queryset), self.total_courses)
```
DawoudSheraz: what's the page_size here? I think we should explicitly have courses greater than page size to emphasize that search will return everything in a single call.

docs/decisions/0030-use-elasticsearch-search-after.rst
```diff
@@ -7,49 +7,52 @@ Accepted (December 2024)
...
-will be introduced to include additional search index fields, specifically `sort`` and `aggregate_uuid` in their responses. 
+will be introduced to include additional search index fields, specifically `sort`` and `aggregation_uuid` in their responses.
 
-Consumers will interact with the new v2 search endpoint by making an initial request to `/api/v2/search/all/`, 
-which returns search results along with a `next` field representing the `sort` value of the last document. 
+Consumers will interact with the new v2 search endpoint by making an initial request to `/api/v2/search/all/`,
+which returns search results along with a `next` field representing the `sort` value of the last document.
 For subsequent pages, they simply include this `next` value as the `search_after` parameter in their next request.
 
+A new mixin, SearchAfterMixin, will be created and added to enable the search_after functionality in catalog query endpoints, such as `/api/v1/catalog/query_contains`.
```
DawoudSheraz: - We should add more details of the mixin
- nit: v2/catalog/query_contains

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: Additionally, the results should not contain any duplicates. That needs to be checked as well.

course_discovery/apps/api/v2/views/catalog_queries.py
```diff
@@ -0,0 +1,75 @@
...
+
+                log.info(f"Specified course ids: {specified_course_ids}")
+                identified_course_ids.update(
+                    self.search(
+                        query,
+                        queryset=Course.objects.all(),
+                        partner=ESDSLQ('term', partner=partner.short_code),
+                        identifiers=ESDSLQ('terms', **{'uuid': course_uuids}),
+                        document=CourseDocument
+                    ).values_list('uuid', flat=True)
```
DawoudSheraz, line 67: this uses values_list while course_run_ids is using comprehension. We can should make it consistent.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: I'll double check that but given the correct sort id, it should never have duplicates.

course_discovery/apps/api/v2/views/catalog_queries.py
Ali-D-Akbar: It's rather consistent now.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1149,6 +1149,75 @@ def search(cls, query, queryset=None):
...
+
+        Args:
+            query (str) -- Elasticsearch querystring (e.g. `title:intro*`)
+            queryset (models.QuerySet) -- base queryset to search, defaults to objects.all()
+            page_size (int) -- Number of results per page.
+            partner (object) -- To be included in the ES query.
+            identifiers (object) -- UUID or key of a product.
+
+        Returns:
+            QuerySet
```
AfaqShuaib09, line 1183: Better to add the type of queryset here as it will be ElasticSearch Queryset which is different than simple django orm queryset

course_discovery/apps/course_metadata/models.py
```diff
@@ -1149,6 +1149,75 @@ def search(cls, query, queryset=None):
...
+        """
+        query = clean_query(query)
+        queryset = queryset or cls.objects.all()
+
+        if query == '(*)':
+            # Early-exit optimization. Wildcard searching is very expensive in elasticsearch. And since we just
+            # want everything, we don't need to actually query elasticsearch at all.
+            return queryset
```
AfaqShuaib09, line 1191: nit: this line is missing in coverage

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Oh, I believe it's actually the django queryset that is being returned.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4192,3 +4195,23 @@ def test_basic(self):
         self.assertEqual(course_run.restricted_run, restricted_course_run)
         self.assertEqual(restricted_course_run.restriction_type, 'custom-b2b-enterprise')
         self.assertEqual(str(restricted_course_run), "course-v1:SC+BreadX+3T2015: <custom-b2b-enterprise>")
+
+
+class TestSearchAfterMixin(ElasticsearchTestMixin, TestCase):
```
DawoudSheraz: Shouldn't this test live in test_mixins? It seems a bit weird to have it in test_models.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4192,3 +4195,23 @@ def test_basic(self):
...
+    def setUp(self):
+        super().setUp()
+
+        self.total_courses = 5
+        for _ in range(self.total_courses):
+            CourseFactory()
```
DawoudSheraz: you can use CourseFactory.create_batch(count) instead of doing this loop.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4192,3 +4195,23 @@ def test_basic(self):
...
+
+    @patch("course_discovery.apps.course_metadata.models.registry.get_documents")
+    def test_fetch_all_courses(self, mock_get_documents):
```
DawoudSheraz: Apart from courses, try indexing combination of different products like CourseRun, Programs, etc. to ensure a variety and then verify everything is working as expected.

docs/decisions/0030-use-elasticsearch-search-after.rst
```diff
@@ -7,49 +7,53 @@ Accepted (December 2024)
...
-To address this issue, we need a more efficient way to paginate large query results. 
-The solution must allow for seamless and reliable pagination without imposing excessive resource demands on the system. 
+To address this issue, we need a more efficient way to paginate large query results.
+The solution must allow for seamless and reliable pagination without imposing excessive resource demands on the system.
 Furthermore, it should ensure that the existing search functionality and search responses remain unaffected in the current version of the endpoint.
 
 Decision
 ----------
-A new version (v2) of the `search/all/` endpoint will be introduced to enhance functionality while ensuring that the existing v1 functionality remains unaffected. 
+A new version (v2) of the `search/all/` endpoint will be introduced to enhance functionality while ensuring that the existing v1 functionality remains unaffected.
```
DawoudSheraz: We can tweak the decision section to highlight the addition of SearchAfterMixin and SearchAfterPagination instead of just mentioning that new endpoint was added. It would better reflect the capabilities. Then we can build on that and show how new endpoints were added.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1149,6 +1149,75 @@ def search(cls, query, queryset=None):
...
+            search = (
+                es_document.search()
+                .query(dsl_query)
+                .sort('id')
+                .extra(size=page_size)
+            )
+
+            search = search.extra(search_after=search_after) if search_after else search
+
+            results = search.execute()
```
DawoudSheraz, line 1217: curious: should we not add error handling here, in case any of the sub-sequent request fails?

course_discovery/apps/api/v2/views/catalog_queries.py
DawoudSheraz: Is it? The code is still the same.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: The error is already handled in this custom dispatch https://github.com/openedx/course-discovery/blob/df07d968b1a5063155d2064f551f2b52bdd25655/course_discovery/apps/api/mixins.py#L117 and will be shown on the API response as a result.

```
{
    "detail": "Failed to make Elasticsearch request. Got exception: RequestError(400, 'search_phase_execution_exception', 'Failed to parse query [(org:)]')"
}
```

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: These functionalities are well tested in `course_discovery/apps/api/v2/tests/test_views/test_catalog_queries.py`. We've already added a proxy model to verify the behavior that the search_after functionality is working as expected.

course_discovery/apps/api/v2/views/catalog_queries.py
Ali-D-Akbar: Ah, I misunderstood the earlier comment. Lemme use values_list for both.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Adding coverage.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: Moving to a new file `test_mixins.py`

course_discovery/apps/course_metadata/models.py
```diff
@@ -1149,6 +1149,75 @@ def search(cls, query, queryset=None):
...
+            queryset (models.QuerySet) -- base queryset to search, defaults to objects.all()
+            page_size (int) -- Number of results per page.
+            partner (object) -- To be included in the ES query.
+            identifiers (object) -- UUID or key of a product.
+
+        Returns:
+            QuerySet
+        """
+        query = clean_query(query)
+        queryset = queryset or cls.objects.all()
```
DawoudSheraz, line 1186: nit: should we have cls.objects.all() here, considering the mixin is not used solely on models.

docs/decisions/0030-use-elasticsearch-search-after.rst
```diff
@@ -7,49 +7,55 @@ Accepted (December 2024)
...
+**How search_after Works:**
+`search_after` is a pagination mechanism that allows retrieving results beyond the standard window limit by using the sort values of the last document from the previous page.
 Instead of using traditional offset-based pagination, it uses the actual sort values of the last retrieved document to fetch the next set of results, ensuring efficient and accurate pagination for large datasets.
-
-In the v2 implementation, response documents will include a `sort` field that can be used as the `search_after` query parameter in subsequent queries. 
-This approach enables scalable retrieval of large datasets by bypassing the `MAX_RESULT_WINDOW` limitations. 
 To support this, a new `SearchAfterPagination` class will be introduced, which will parse the `search_after` query parameter to facilitate efficient pagination of ElasticSearch results.
 
-Additionally, new serializers will be integrated for the v2 implementation. 
+A new mixin, `SearchAfterMixin`, will be created to enable the search_after functionality in catalog query endpoints, such as `/api/v2/catalog/query_contains`.
```
DawoudSheraz: Let's not mention any endpoint explicitly here. We can focus on the features of the mixin, like search handling the saerch_after internally if the record count is greater than page size.

DawoudSheraz: Lets mention the capabilities like 
```
- SearchAfterPagination
- SearchAfterMixin
```

Then, have individual sections with proper headings for both. It will make the content readable.

docs/decisions/0030-use-elasticsearch-search-after.rst
```diff
@@ -7,49 +7,55 @@ Accepted (December 2024)
...
 - PersonSearchDocumentSerializer
 
-will be introduced to include additional search index fields, specifically `sort`` and `aggregate_uuid` in their responses. 
+will be introduced to include additional search index fields, specifically `sort`` and `aggregation_uuid` in their responses.
 
-Consumers will interact with the new v2 search endpoint by making an initial request to `/api/v2/search/all/`, 
-which returns search results along with a `next` field representing the `sort` value of the last document. 
+Consumers will interact with the new v2 search endpoint by making an initial request to `/api/v2/search/all/`,
+which returns search results along with a `next` field representing the `sort` value of the last document.
 For subsequent pages, they simply include this `next` value as the `search_after` parameter in their next request.
```
DawoudSheraz: next is now the link, no need to pass it as search after. the next link can be followed as it is. We can update this in ADR.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Right, the order of mixins will matter here i.e. include this mixin before SearchAfterMixin.

course_discovery/apps/api/v2/tests/test_views/test_catalog_queries.py
```diff
@@ -0,0 +1,94 @@
...
+        """ Verify that a 400 status is returned if request contains incorrect query string. """
+        qs = urllib.parse.urlencode({
+            'query': 'title:',
+            'course_run_ids': self.course_run.key,
+            'course_uuids': self.course.uuid,
+        })
+        url = f'{self.url_base}/?{qs}'
+
+        response = self.client.get(url)
+        assert response.status_code == 400
```
DawoudSheraz, line 98: we should assert on error message here, this is being done in other 400 tests too.

course_discovery/apps/api/v2/views/catalog_queries.py
```diff
@@ -0,0 +1,75 @@
...
+        if query and (course_run_ids or course_uuids):
+            log.info(
+                f"Attempting search against query {query} with course UUIDs {course_uuids} "
+                f"and course run IDs {course_run_ids}"
+            )
+            identified_course_ids = set()
+            specified_course_ids = []
+            if course_run_ids:
+                course_run_ids = course_run_ids.split(',')
+                specified_course_ids = course_run_ids
```
DawoudSheraz: Instead of naming it specified_course_ids as it has both course and course run ids, it can be renamed to specified_identifiers, etc to signify it is not restricted to one item.

course_discovery/apps/course_metadata/models.py
```diff
@@ -1149,6 +1149,75 @@ def search(cls, query, queryset=None):
...
+        dsl_query = ESDSLQ('bool', must=must_queries)
+
+        all_ids = set()
+        search_after = None
+
+        while True:
+            search = (
+                es_document.search()
+                .query(dsl_query)
+                .sort('id')
```
hamza-56: Suggestion: Make the sort field configurable via a parameter

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Yeah which is why we have the option to send the queryset here.

docs/decisions/0030-use-elasticsearch-search-after.rst
Ali-D-Akbar: Reformatted the sections in the decision. It should look more readable now.

docs/decisions/0030-use-elasticsearch-search-after.rst
```diff
@@ -7,66 +7,86 @@ Accepted (December 2024)
...
+**Key Enhancements**
+
+- A new version (`v2`) of the `/search/all/` endpoint will be introduced to enhance functionality while maintaining backward compatibility with `v1`.
+- Response documents will include a `sort` field to be used as the `search_after` query parameter for subsequent queries.
+- This approach enables scalable retrieval of large datasets by bypassing the `MAX_RESULT_WINDOW` limitations.
+
+**Serializer Updates**
+
+New serializers will be integrated for the `v2` implementation:
+- `AggregateSearchListSerializerV2` will extend `AggregateSearchListSerializer`, supporting the `search_after` pagination mechanism and incorporating newer serializer versions for the same document types.
```
DawoudSheraz, line 58: You need an empty line here as this list is not rendering correctly.

docs/decisions/0030-use-elasticsearch-search-after.rst
```diff
@@ -7,66 +7,86 @@ Accepted (December 2024)
...
+
+.. code-block:: python
+
+    # First request
+    response1 = requests.get('/api/v2/search/all/')
+    results1 = response1.json()
+    next_page_search_after = results1['next']  # This is the sort value of the last document
+
+    # Next request using the 'next' value
+    response2 = requests.get(f'/api/v2/search/all/?search_after={json.dumps(next_page_search_after)}')
```
DawoudSheraz: next_page_search_after contains the complete url, there is no need to add the base URL before it i.e. requests.get(next_page_search_after) will do the job

DawoudSheraz (approved):

hamza-56 (approved):

PR #4537, opened 2025-01-08
perf: optimize loading of django admin program pages

zawan-ila:
### Description

Program admin pages are taking a long time to load and sometimes may even time out. Looking at the traces, most of the time is being spent in rendering the html widgets.

This PR makes the following changes:-

1. Add `in_year_value` to autocomplete_fields so that all ProductValue objects are not rendered in the html
2. While calculating `optgroups` on `SortedModelSelect2Multiple` widgets, do not include unselected options. This prevents the unselected options from being rendered in the html. `django_autocomplete_light` also appears to be [doing](https://github.com/yourlabs/django-autocomplete-light/blob/45a9ff6e95e58348b90b3c1f01f01cb6ca10b4d3/src/dal/widgets.py#L129-L133) the same thing
3. Currently, the sorting of `SortedModelSelect2Multiple` widgets is broken. We fix it by deferring the loading of `sortable_select.js`

### Before
Screenshot 2025-01-09 at 1 44 36 PM
Screenshot 2025-01-09 at 1 45 11 PM

### After

Screenshot 2025-01-09 at 1 43 30 PM
Screenshot 2025-01-09 at 1 44 07 PM

### DIFF
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -128,6 +128,7 @@ class ProductValueAdmin(admin.ModelAdmin):
     ]
     search_fields = ('id',)
 
+
 @admin.register(Course)
 class CourseAdmin(DjangoObjectActions, SimpleHistoryAdmin):
     form = CourseAdminForm
--- a/course_discovery/apps/course_metadata/admin.py
+++ b/course_discovery/apps/course_metadata/admin.py
@@ -7,8 +7,9 @@ from django.db.models import Prefetch
 from django.db.utils import IntegrityError
 from django.forms import CheckboxSelectMultiple, ModelForm
 from django.http import HttpResponseRedirect
+from django.templatetags.static import static
 from django.urls import re_path, reverse
-from django.utils.html import format_html
+from django.utils.html import format_html, html_safe
 from django.utils.translation import gettext_lazy as _
 from django_object_actions import DjangoObjectActions
 from parler.admin import TranslatableAdmin
@@ -28,7 +29,6 @@ from course_discovery.apps.course_metadata.forms import (
     CourseAdminForm, CourseRunAdminForm, PathwayAdminForm, ProgramAdminForm
 )
 from course_discovery.apps.course_metadata.models import *  # pylint: disable=wildcard-import
-from course_discovery.apps.course_metadata.utils import SortableSelectJSPath
 from course_discovery.apps.course_metadata.views import (
     CourseSkillsView, RefreshCourseSkillsView, RefreshProgramSkillsView
 )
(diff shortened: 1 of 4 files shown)


course_discovery/apps/course_metadata/tests/test_widgets.py
```diff
@@ -7,9 +7,9 @@
 @ddt.ddt
 class SortedModelSelect2MultipleTests(TestCase):
     @ddt.data(
-        (['1', '2'], [1, 2, 3]),
-        (['2', '1'], [2, 1, 3]),
-        (['3'], [3, 1, 2]),
```
AfaqShuaib09, line 12: [curious] Why did you change the test data here?

course_discovery/apps/course_metadata/admin.py
```diff
@@ -523,7 +524,7 @@ class Media:
         js = (
             'bower_components/jquery-ui/ui/minified/jquery-ui.min.js',
             'bower_components/jquery/dist/jquery.min.js',
-            'js/sortable_select.js'
+            SortableSelectJSPath()
```
DawoudSheraz, line 535: can it cause any issues if we are providing a script tag here instead of path to js?

course_discovery/apps/course_metadata/widgets.py
```diff
@@ -29,4 +29,4 @@ def optgroups(self, name, value, attrs=None):
                     ordered.append(item)
                     break
 
-        return ordered + unselected
+        return ordered
```
DawoudSheraz, line 32: Does this mean we dont render unselected items and are only visible when searching? I am not able to understand the complete context yet.

zawan-ila: Yes. For an example, consider the current courses field. The `optgroups` returns all the `unselected` options, hence we end up rendering an option for each course in the database. With this change, we will render only the selected courses.

course_discovery/apps/course_metadata/tests/test_widgets.py
zawan-ila: Since we change the behavior of optgroups to return only the selected fields, the unselected values need to be removed.

course_discovery/apps/course_metadata/admin.py
zawan-ila: No. This is [documented](https://docs.djangoproject.com/en/4.2/topics/forms/media/#paths-as-objects).

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/widgets.py
DawoudSheraz: Ok. A screenshot of before and after would be helpful here.

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1264,3 +1266,10 @@ def get_course_run_statuses(statuses, course_runs):
         else:
             statuses.add(course_run.status)
     return statuses
+
+
+@html_safe
+class SortableSelectJSPath:
```
DawoudSheraz: Can we have this as an internal class in admin, considering it is only being used in one place? I am not sure if it makes sense to have this in utils.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/utils.py
zawan-ila: I have moved it to `admin.py` I don't see how to make it `internal` though. In particular, we'll need to refer an attribute of `ProgramAdmin` (SortableSelectJSPath) during its definition(inside Media.js) and I don't see how to do that.

PR #4539, opened 2025-01-09
fix: sorting of elements in api/v2/search/all

AfaqShuaib09:
[PROD-4268](https://2u-internal.atlassian.net/browse/PROD-4268)
-----

This PR fixes sort order in `api/v2/search/all` endpoint by passing `last_item` from Serializer's `to_representation` to SearchAfterPagination by setting it as View instance level attribute

### DIFF
--- a/course_discovery/apps/api/v2/tests/test_views/test_search.py
+++ b/course_discovery/apps/api/v2/tests/test_views/test_search.py
@@ -136,3 +136,45 @@ class AggregateSearchViewSetV2Tests(mixins.LoginMixin, ElasticsearchTestMixin, m
         assert (
             single_page_data["results"] == all_results
         ), "Combined pagination results do not match single request results"
+
+    def test_last_item_sort_pagination(self):
+        """
+        Test that the results won't be duplicated in any request response.
+        """
+        page_size = 5
+        PersonFactory.create_batch(10, partner=self.partner)
+        courses = CourseFactory.create_batch(10, partner=self.partner)
+
+        for course in courses:
+            CourseRunFactory(
+                course__partner=self.partner,
+                course=course,
+                type__is_marketable=True,
+                status=CourseRunStatus.Published,
+            )
+        total_items_count = 30  # 10 Persons + 10 Courses + 10 CourseRuns
+        max_pages = total_items_count // page_size
+
+        response_data = self.fetch_page_data(page_size)
+        all_results = response_data["results"]
+        aggregate_uuids = [result["aggregation_uuid"] for result in all_results]
+        next_url = response_data.get("next")
+
+        for _ in range(max_pages - 1):
+            parsed_url = urlparse(next_url)
+            query_params = parse_qs(parsed_url.query)
+            search_after = query_params.get("search_after", [None])[0]
+            assert search_after is not None
+
+            last_sort_value = all_results[-1]["sort"]
+            assert json.loads(search_after) == last_sort_value
+
+            response_data = self.fetch_page_data(page_size, next_url=next_url)
+            for result in response_data['results']:
+                assert result['aggregation_uuid'] not in aggregate_uuids
+                aggregate_uuids.append(result['aggregation_uuid'])
+
+            all_results.extend(response_data["results"])
+            next_url = response_data.get("next")
+
+        assert len(aggregate_uuids) == total_items_count
--- a/course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
+++ b/course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
@@ -114,30 +114,26 @@ class SearchAfterPagination(PageNumberPagination):
         """
         Paginate the Elasticsearch queryset using search_after.
         """
-
         search_after = request.query_params.get(self.search_after_param)
         if search_after:
             try:
                 queryset = queryset.extra(search_after=json.loads(search_after))
             except json.JSONDecodeError as exc:
                 raise ValueError("Invalid JSON format for search_after parameter") from exc
-
-        return super().paginate_queryset(queryset, request, view)
+        queryset = super().paginate_queryset(queryset, request, view)
+        self.last_obj = queryset[-1] if queryset else None  # pylint: disable=attribute-defined-outside-init
+        return queryset
 
     def get_next_link(self):
         if not self.page.has_next():
             return None
 
-        last_item_sort = self._get_last_item_sort()
+        last_item_sort = self.last_obj.meta.sort.copy() if hasattr(self, 'last_obj') and self.last_obj else None
         if not last_item_sort:
             return None
 
         url = self.request.build_absolute_uri()
         return replace_query_param(url, self.search_after_param, json.dumps(last_item_sort))
-
-    def _get_last_item_sort(self):
-        last_item = self.page.object_list[-1] if self.page.object_list else None
-        return list(last_item.meta.sort) if last_item else None
 
 
 class BaseElasticsearchDocumentViewSet(mixins.DetailMixin, mixins.FacetMixin, DocumentViewSet):


course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -86,6 +86,12 @@ def to_representation(self, data):
         This will invoke the respective custom list serializers for all the documents.
         """
         iterable = data.all() if isinstance(data, models.Manager) else data
+        last_item = iterable[-1] if len(iterable) > 0 else None
```
zawan-ila: `if len(iterable)` is enought here

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -86,6 +86,12 @@ def to_representation(self, data):
...
+        self._context["last_item"] = last_item
```
zawan-ila: This line is unneeded?

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -86,6 +86,12 @@ def to_representation(self, data):
...
+        view = self._context.get("view")
```
zawan-ila: Why not use `self.context`?

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -86,6 +86,12 @@ def to_representation(self, data):
...
+        if view:
+            setattr(view, "last_item", last_item)
```
zawan-ila: [curious] Why do you set it on the `view` instead of, say, the `request`?

zawan-ila (commented):
Can we add a test for this behavior?

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
AfaqShuaib09: yeah, this is unneeded, I've added to check if we can set the value in context, but forgot to remove it

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
AfaqShuaib09: As you can see here: https://github.com/openedx/course-discovery/pull/4539/files#diff-561e322fc32e2ee983f22f71b8da3cf7c28cf378dbc2cc7c9cdbb103a50bdfcfR98-R104, we are passing `self._context` to representations code block as well. Therefore, I used `self._context` here.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
AfaqShuaib09: As you can see here: https://github.com/openedx/course-discovery/pull/4539/files#diff-561e322fc32e2ee983f22f71b8da3cf7c28cf378dbc2cc7c9cdbb103a50bdfcfR98-R104, we are passing `self._context` to representations code block as well. Therefore, I used `self._context` here.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
AfaqShuaib09: from rest_framework docs, I found a workaround to set an attribute in view on instance level. I'll check if we can achieve same using request as well

AfaqShuaib09: updated

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
zawan-ila: In general, it is not a good practice to use attributes prefixed with `_` as they are considered "private". Are there any difficulties that arise when using `self.context` here?

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
zawan-ila: Can you please link to these docs?

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
AfaqShuaib09: > Are there any difficulties that arise when using self.context here?

No difficulty. Actually, I haven't tried it because self._context already used in the existing code in the lines below. Therefore, I used it as it is. I've updated it now.

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
AfaqShuaib09: I've set the attribute in the `request` instead of the `view` and added a test case to verify that there is no duplication of `aggregation_uuid` in the results

course_discovery/apps/course_metadata/search_indexes/serializers/aggregation.py
```diff
@@ -86,6 +86,11 @@ def to_representation(self, data):
         This will invoke the respective custom list serializers for all the documents.
         """
         iterable = data.all() if isinstance(data, models.Manager) else data
+        last_item = iterable[-1] if len(iterable) else None
+        request = self.context.get("request")
+        if request:
+            setattr(request, "last_item", last_item)
```
DawoudSheraz: Please add a comment here to explain why this attribute setting is needed.

course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
```diff
@@ -136,8 +136,16 @@ def get_next_link(self):
...
-        return list(last_item.meta.sort) if last_item else None
+        request = self.request
+        last_item = getattr(request, "last_item", None)
+        if last_item:
+            last_item = last_item.meta.sort.copy()
+        else:
+            last_item = self.page.object_list[-1] if self.page.object_list else None
+            last_item = list(last_item.meta.sort) if last_item else None
+        if hasattr(request, "last_item"): # Clear the last_item attribute to prevent reuse in the next request
+            delattr(request, "last_item")
```
DawoudSheraz: The next request wont have it by default, deleting it is unnecessary. Did you observe any issue locally that led you to add this?

course_discovery/apps/api/v2/tests/test_views/test_search.py
```diff
@@ -136,3 +136,43 @@ def test_search_after_pagination(self):
...
+
+        for course in courses:
+            CourseRunFactory(
+                course__partner=self.partner,
+                course=course,
+                type__is_marketable=True,
+                status=CourseRunStatus.Published,
+            )
+
+        page_size = 5
```
DawoudSheraz: nit: move the page_size at the very start of the test.

course_discovery/apps/api/v2/tests/test_views/test_search.py
```diff
@@ -136,3 +136,43 @@ def test_search_after_pagination(self):
...
+        response_data = self.fetch_page_data(page_size)
+        all_results = response_data["results"]
+        aggregate_uuids = []
+        for result in all_results:
+            aggregate_uuids.append(result["aggregation_uuid"])
```
DawoudSheraz: you can combine these 3 lines into a single via list comprehension
```
aggregate_uuids = [result["aggregation_uuid"] for result in all_results]
```

course_discovery/apps/api/v2/tests/test_views/test_search.py
```diff
@@ -136,3 +136,43 @@ def test_search_after_pagination(self):
...
+        next_url = response_data.get("next")
+
+        while next_url:
```
DawoudSheraz: nit: can we use for loop to loop an explicit number of times instead of using while loop in here? Using while loop in a test can be a recipe for trouble. The current code is ok but considering we are testing an explicit behavior in the test, we have quite a lot of control (on page size, how many pages are there, etc.)

course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
AfaqShuaib09: Nope, didn’t encounter any issues, added it here just for safety. Previously, I faced issues with the views attribute approach

course_discovery/apps/api/v2/tests/test_views/test_search.py
AfaqShuaib09: Hmm, sure let me convert it to for loop

course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
```diff
@@ -136,8 +136,14 @@ def get_next_link(self):
         return replace_query_param(url, self.search_after_param, json.dumps(last_item_sort))
 
     def _get_last_item_sort(self):
-        last_item = self.page.object_list[-1] if self.page.object_list else None
-        return list(last_item.meta.sort) if last_item else None
+        request = self.request
+        last_item = getattr(request, "last_item", None)
+        if last_item:
+            last_item = last_item.meta.sort.copy()
```
DawoudSheraz: list() is not called here but is being called on L145. If this is needed, can it be called when returning the last_item?

AfaqShuaib09: .copy() already converts this Elastic.AttrList to a simple list, so it won't be required here.

AfaqShuaib09: Already tried to typecast it directly to a list, but it gives an error, so I’ve used a workaround here using copy.

course_discovery/apps/edx_elasticsearch_dsl_extensions/viewsets.py
```diff
@@ -136,8 +136,14 @@ def get_next_link(self):
         return replace_query_param(url, self.search_after_param, json.dumps(last_item_sort))
 
     def _get_last_item_sort(self):
-        last_item = self.page.object_list[-1] if self.page.object_list else None
-        return list(last_item.meta.sort) if last_item else None
+        request = self.request
+        last_item = getattr(request, "last_item", None)
+        if last_item:
```
DawoudSheraz: if last_item is None, that can mean 2 things:
- The attribute is not present in request
- There is no more item to fetch

The current code will always execute 2nd else regardless what the last_item represents. Maybe the check can be made explicit to do early return if we know there is no more item to fetch(?).

AfaqShuaib09: hmm, yes we can add the explicit check to return early for this case

zawan-ila (approved):
Please fix the coverage before merge

course_discovery/apps/api/v2/tests/test_views/test_search.py
```diff
@@ -136,3 +136,47 @@ def test_search_after_pagination(self):
...
+        max_pages = total_items_count // page_size
+
+        response_data = self.fetch_page_data(page_size)
+        all_results = response_data["results"]
+        aggregate_uuids = [result["aggregation_uuid"] for result in all_results]
+        next_url = response_data.get("next")
+        
+        for _ in range(max_pages - 1):
+            if not next_url:
+                break
```
DawoudSheraz: if the max pages are known, is this check needed?

AfaqShuaib09: no, it won't be needed, added it to break the loop once the next_url becomes empty. I've removed it now

DawoudSheraz (approved):
Good to go once CI is merged

PR #4540, opened 2025-01-10
perf: relegate optgroup calculation to superclass

zawan-ila:
### Description
[4538](https://github.com/openedx/course-discovery/pull/4538) seems to have improved the perf by about 45% (on stage) which is lesser than what I expected.

Refactor `SortedModelSelect2Multiple.optgroups ` to call the superclass `optgroups` to retrieve the options and then sort them accordingly. This is an attempt at further improving the performance. A side effect is that it simplifies the code.

### DIFF
--- a/course_discovery/apps/course_metadata/tests/test_widgets.py
+++ b/course_discovery/apps/course_metadata/tests/test_widgets.py
@@ -1,19 +1,40 @@
-import ddt
+import itertools
+
+from bs4 import BeautifulSoup
 from django.test import TestCase
+from django.urls import reverse
 
-from course_discovery.apps.course_metadata.widgets import SortedModelSelect2Multiple
+from course_discovery.apps.api.tests.mixins import SiteMixin
+from course_discovery.apps.core.tests.factories import USER_PASSWORD, UserFactory
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, ProgramFactory
 
 
-@ddt.ddt
-class SortedModelSelect2MultipleTests(TestCase):
-    @ddt.data(
-        (['1', '2'], [1, 2]),
-        (['2', '1'], [2, 1]),
-        (['3'], [3,]),
-    )
-    @ddt.unpack
-    def test_optgroups_are_sorted(self, value, result_order):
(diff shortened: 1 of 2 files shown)


AfaqShuaib09 (approved):

DawoudSheraz (commented):
Are we gaining anything from perf point of view by calling super optrgoups only (I am assuming the code is same)? Or is it only achieving refactoring goals?

zawan-ila:
The super code is [somewhat different](https://github.com/yourlabs/django-autocomplete-light/blob/45a9ff6e95e58348b90b3c1f01f01cb6ca10b4d3/src/dal/widgets.py#L129-L133). It does some mangling around of `self.choices` which allows it to process "less" choices. The idea is the same as [4538](https://github.com/openedx/course-discovery/pull/4538) i.e `lesser choice processing = better perf`. The refactoring is an added bonus.

Edit: In particular, the superclass restricts the `self.choices.queryset` to only the selected choices. Hence, preventing needless processing of unselected choices.

course_discovery/apps/course_metadata/tests/test_widgets.py
```diff
@@ -1,19 +1,40 @@
...
-    @ddt.data(
-        (['1', '2'], [1, 2]),
-        (['2', '1'], [2, 1]),
-        (['3'], [3,]),
-    )
-    @ddt.unpack
-    def test_optgroups_are_sorted(self, value, result_order):
-        choices = ((1, 'one'), (2, 'two'), (3, 'three'))
-        widget = SortedModelSelect2Multiple(url='requiredurl', choices=choices)
-        result = widget.optgroups('test', value)
```
zawan-ila, line 18: It is hard to do a simple optgroups test like this now since the method calls the super method, which expects things to be set up much more properly. e.g requiring `self.choices` to have a `queryset` attribute, having the url kwarg of the widget being an actual reversible url etc.

I have added a new test in place, which renders the program admin form, and ensures that the courses are rendered in the order in which they were assigned. I am unsure if this test is more appropriate in `test_admin.py` now.

DawoudSheraz (approved):

PR #4543, opened 2025-01-13
perf: optimize django admin pages

zawan-ila:
### Description

This PR makes the following changes:-

1. Adds some fields to `autocomplete_fields` to prevent serializing all of their objects in the select options
2. Fixes sorting for SortedM2M fields in program admin by using the `SortedModelSelect2Multiple` widget
3. Shifts some course admin M2M fields to use the `SortedModelSelect2Multiple` widget

### DIFF
--- a/course_discovery/apps/course_metadata/lookups.py
+++ b/course_discovery/apps/course_metadata/lookups.py
@@ -4,7 +4,7 @@ from dal import autocomplete
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.db.models import Q
 
-from .models import (
+from course_discovery.apps.course_metadata.models import (
     FAQ, Collaborator, CorporateEndorsement, Course, CourseRun, Endorsement, ExpectedLearningItem, JobOutlookItem,
     Organization, Person, Program
 )
--- a/course_discovery/apps/course_metadata/lookups.py
+++ b/course_discovery/apps/course_metadata/lookups.py
@@ -5,8 +5,8 @@ from django.contrib.auth.mixins import LoginRequiredMixin
 from django.db.models import Q
 
 from .models import (
-    Collaborator, CorporateEndorsement, Course, CourseRun, Endorsement, ExpectedLearningItem, FAQ, JobOutlookItem,
-    Organization, Person, Program,
+    FAQ, Collaborator, CorporateEndorsement, Course, CourseRun, Endorsement, ExpectedLearningItem, JobOutlookItem,
+    Organization, Person, Program
 )
 
 
--- a/course_discovery/apps/course_metadata/lookups.py
+++ b/course_discovery/apps/course_metadata/lookups.py
@@ -4,7 +4,10 @@ from dal import autocomplete
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.db.models import Q
 
-from .models import Course, CourseRun, Organization, Person, Program
+from .models import (
+    Collaborator, CorporateEndorsement, Course, CourseRun, Endorsement, ExpectedLearningItem, FAQ, JobOutlookItem,
+    Organization, Person, Program,
+)
 
 
 class CourseAutocomplete(autocomplete.Select2QuerySetView):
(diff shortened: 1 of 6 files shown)


course_discovery/apps/course_metadata/lookups.py
```diff
@@ -4,7 +4,10 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.db.models import Q
 
-from .models import Course, CourseRun, Organization, Person, Program
+from .models import (
```
DawoudSheraz: nit: please update the import to use absolute path. This seems like an old code to be using relative imports

course_discovery/apps/course_metadata/urls.py
```diff
@@ -5,16 +5,32 @@
 from django.urls import path
```
DawoudSheraz, line 5: nit: considering this file is only used for admin URLs (https://github.com/openedx/course-discovery/blob/9c571cb64b15a1205942ad4a5914b760f1f9b5a6/course_discovery/urls.py#L44), should we rename this to admin_urls.py? The namespace is already named admin_metadata. Generally, urls.py tends to contain public view URLs and this particular file only has admin related urls.

zawan-ila: > Generally, urls.py tends to contain public view URLs

I don't really see this distinction. What is a public view url? A lot of APIs are behind auth. Are they public or private?

DawoudSheraz: Right. To explain it better, the public does not mean it is without auth. What I meant was that the URL is publically accessible. The private ones would refer to URLs that are only accessible on VPN or internal networks. Generally, the admin URLs are not "publically" accessible. That's what I was trying to refer to.

zawan-ila: Right. I still like it as is. I do not have any strong opinions though.

DawoudSheraz: Right. Let's add a comment before urlpatterns, just to make it easy to get the context.

zawan-ila: The file docstring already mentions that these urls are for admin autocomplete.

DawoudSheraz: Oh, I missed it. That's fine then.

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -102,6 +111,33 @@ def test_autocomplete_requires_staff_permission(self, view_prefix, client):
...
+            (CorporateEndorsementFactory, 'corporate-endorsement-autocomplete', ['corporation_name', 'statement']),
+            (EndorsementFactory, 'endorsement-autocomplete', ['quote']),
+            (ExpectedLearningItemFactory, 'expected-learning-item-autocomplete', ['value']),
+            (FAQFactory, 'faq-autocomplete', ['question', 'answer']),
+            (JobOutlookItemFactory, 'job-outlook-item-autocomplete', ['value']),
+        ]
+    )
+    @pytest.mark.foo
+    def test_models_autocomplete(self, admin_client, model_factory, autocomplete_path, lookup_attrs):
+        # breakpoint()
```
AfaqShuaib09: remove this comment

course_discovery/apps/course_metadata/tests/test_lookups.py
```diff
@@ -102,6 +111,33 @@ def test_autocomplete_requires_staff_permission(self, view_prefix, client):
...
+        [
+            (CollaboratorFactory, 'collaborator-autocomplete', ['name', 'uuid']),
+            (CorporateEndorsementFactory, 'corporate-endorsement-autocomplete', ['corporation_name', 'statement']),
+            (EndorsementFactory, 'endorsement-autocomplete', ['quote']),
+            (ExpectedLearningItemFactory, 'expected-learning-item-autocomplete', ['value']),
+            (FAQFactory, 'faq-autocomplete', ['question', 'answer']),
+            (JobOutlookItemFactory, 'job-outlook-item-autocomplete', ['value']),
+        ]
+    )
+    @pytest.mark.foo
```
AfaqShuaib09: remove this one too

course_discovery/apps/course_metadata/lookups.py
```diff
@@ -36,6 +63,54 @@ def get_queryset(self):
         return []
 
 
+class EndorsementAutocomplete(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        if self.request.user.is_authenticated and self.request.user.is_staff:
```
AfaqShuaib09, line 68: can we define a custom permission decorator just like login_required instead of adding this same permission to all Autocomplete classes

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/lookups.py
DawoudSheraz: +1, though I would not make it a blocker for this PR. If it is quick, go ahead.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/lookups.py
zawan-ila: I agree with the sentiment. However, in the interest of time, I am choosing to overlook this.

PR #4544, opened 2025-01-15
feat: add vertical models in tagging app

AfaqShuaib09:
[PROD-4280](https://2u-internal.atlassian.net/browse/PROD-4280)
--------

This PR adds the model related to vertical tagging.

- Create vertical, subvertical, and coursevertical models in tagging.
- Register the models on admin.
- Ensure proper validation checks are present on models save (only superusers and allowed user group)
- Use history to track changes.


course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class SubVertical(TimeStampedModel):
+    """
+    Model for defining sub-verticals used to categorize product types under specific verticals.
+    """
+    name = models.CharField(max_length=255, unique=True)
+    slug = AutoSlugField(populate_from='name', max_length=255, unique=True)
+    is_active = models.BooleanField(default=True)
+    verticals = models.ForeignKey(Vertical, on_delete=models.CASCADE, related_name='sub_verticals')
```
DawoudSheraz: nit: the field name should be vertical as a subvertical is linked to one vertical only. related_name is ok.

course_discovery/apps/tagging/admin.py
```diff
@@ -0,0 +1,76 @@
...
+from course_discovery.apps.tagging.models import CourseVertical, SubVertical, Vertical
+
+
+@admin.register(Vertical)
+class VerticalAdmin(admin.ModelAdmin):
+    """
+    Admin class for Vertical model.
+    """
+    list_display = ('name', 'is_active', 'slug',)
+    search_fields = ('name',)
```
DawoudSheraz, line 27: nit: should we display subverticals tabs in here? I think it can be helpful

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,128 @@
...
+    is_active = models.BooleanField(default=True)
+    history = HistoricalRecords()
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        verbose_name_plural = 'Verticals'
+        ordering = ['name']
+        unique_together = ['name']
```
DawoudSheraz: Is unique_together really needed here, considering the name and slug are already defined as unique?

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,128 @@
...
+
+from course_discovery.apps.course_metadata.models import Course
+
+
+class Vertical(TimeStampedModel):
+    """
+    Model for defining verticals used to categorize product types
+    """
+    name = models.CharField(max_length=255, unique=True)
+    slug = AutoSlugField(populate_from='name', max_length=255, unique=True)
```
DawoudSheraz, line 15: it can be helpful to add index on slug.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,128 @@
...
+
+        elif not self.sub_vertical and not self.vertical:
+            return   # Skip validation if sub_vertical is not set
+
+
+    def save(self, *args, **kwargs):
+        """
+        Call full_clean before saving to ensure validation is always run
+        """
+        # self.full_clean()
```
DawoudSheraz: nit: full clean commented out call to be enabled/removed.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        raise NotImplementedError("Subclasses must implement `get_object_title`.")
+
+class CourseVertical(BaseVertical):
+    """
+    Model for assigning vertical and sub verticals to courses
+    """
+    course = models.OneToOneField(
+        Course, on_delete=models.CASCADE, related_name="verticals"
+    )
```
DawoudSheraz: Add an explicit limit_choices_to to ensure the relation is only made with non-draft courses.

course_discovery/apps/tagging/models.py
AfaqShuaib09: Yeah, it won't be required here. I missed that I've already added `unique` above

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,200 @@
...
+        self.user = user
+
+
+class BaseAdminTestCase(TestCase):
+    """ Base test class for admin tests """
+
+    def setUp(self):
+        super().setUp()
+        self.site = AdminSite()
+        self.factory = RequestFactory()
```
DawoudSheraz: where is this being used?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,200 @@
...
+
+class MockRequest:
+    def __init__(self, user=None):
+        self.user = user
+
+
+class BaseAdminTestCase(TestCase):
+    """ Base test class for admin tests """
+
+    def setUp(self):
```
DawoudSheraz: can it be setUpClass considering this operation is only needed at the start of suite?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,200 @@
...
+        self.vertical_admin = VerticalAdmin(Vertical, self.site)
+        self.vertical = VerticalFactory()
+
+    def test_save_model_as_superuser(self):
+        """Verify that superuser can save vertical."""
+        request = MockRequest(self.superuser)
+        try:
+            self.vertical_admin.save_model(request, self.vertical, None, False)
+        except PermissionDenied:
+            self.fail("Superuser should be able to save vertical")
```
DawoudSheraz: Why is this needed?

course_discovery/apps/tagging/admin.py
AfaqShuaib09: Like comma_separated str of sub_verticals?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,200 @@
...
+        request = MockRequest(self.regular_user)
+        with self.assertRaises(PermissionDenied):
+            self.vertical_admin.save_model(request, self.vertical, None, False)
+
+    def test_list_display(self):
+        """Verify the correct list display fields."""
+        self.assertEqual(
+            self.vertical_admin.list_display,
+            ('name', 'is_active', 'slug',)
+        )
```
DawoudSheraz: This particular test does not seem to be achieving much.

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def test_vertical_filter_admin_registration(self):
+        """ Verify that Vertical model is registered on the admin site """
+        self.assertIn(Vertical, site._registry)
```
DawoudSheraz: This test is also unnecessary in my opinion. Checking a model is registered on admin takes things somewhat far fetched.

course_discovery/apps/tagging/admin.py
DawoudSheraz: No, the inline form.

course_discovery/apps/tagging/models.py
AfaqShuaib09: I've added this condition on admin level. Lemme add this here as well

course_discovery/apps/tagging/tests/test_admin.py
AfaqShuaib09: I've added it before to mock the user request from admin but forgot to remove it. It's unused now

course_discovery/apps/tagging/tests/test_admin.py
AfaqShuaib09: just as a safety, I can remove it from here.

course_discovery/apps/tagging/admin.py
AfaqShuaib09: Like this?
image

DawoudSheraz: Yes

course_discovery/settings/base.py
```diff
@@ -801,3 +801,5 @@
 RETIRED_RUN_TYPES = []
 RETIRED_COURSE_TYPES = []
 COURSE_ARCHIVAL_MAIL_RECIPIENTS = ['phoenix@edx.org']
+
+VERTICALS_MANAGEMENT_GROUPS = []
```
zawan-ila, line 806: [nit] Add a comment providing more context on this setting

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,124 @@
...
+    vertical = models.ForeignKey(Vertical, on_delete=models.CASCADE, related_name='sub_verticals')
+    history = HistoricalRecords()
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        verbose_name_plural = 'Sub Verticals'
+        ordering = ['name']
+        unique_together = ['name']
```
zawan-ila: Can we remove this?

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class BaseVertical(TimeStampedModel):
```
zawan-ila: `ProductVertical` seems more appropriate

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,124 @@
...
+from course_discovery.apps.course_metadata.models import Course
+
+
+class Vertical(TimeStampedModel):
+    """
+    Model for defining verticals used to categorize product types
+    """
+    name = models.CharField(max_length=255, unique=True)
+    slug = AutoSlugField(populate_from='name', max_length=255, unique=True, db_index=True)
+    is_active = models.BooleanField(default=True)
```
zawan-ila, line 16: What purpose does this field serve? Is it just to make this vertical unavailable for selection? And all current relations to it stay as is?

AfaqShuaib09: yup, it's just to make this vertical unavailable for selection and by default it's value will be true

DawoudSheraz: Chiming in, it is better to have the flag as opposed to data deletion i.e. removing old or retired verticals. A similar pattern is present in different places in codebases (like in taxonomy-connector).

course_discovery/apps/tagging/tests/test_models.py
```diff
@@ -0,0 +1,136 @@
...
+"""
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory
+from course_discovery.apps.tagging.models import CourseVertical, SubVertical
+from course_discovery.apps.tagging.tests.factories import CourseVerticalFactory, SubVerticalFactory, VerticalFactory
+
+
+class VerticalFilterModelTests(TestCase):
```
zawan-ila: `VerticalModelTests`?

course_discovery/apps/tagging/tests/test_models.py
```diff
@@ -0,0 +1,136 @@
...
+        self.assertTrue(sub_vertical.is_active)
+
+        self.vertical.is_active = False
+        self.vertical.save()
+
+        sub_vertical.refresh_from_db()
+        self.assertFalse(sub_vertical.is_active)
+
+
+class SubVerticalFilterModelTests(TestCase):
```
zawan-ila: `SubverticalModelTests`?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,160 @@
...
+            password='password'
+        )
+        cls.regular_user = User.objects.create_user(
+            username='user',
+            email='user@example.com',
+            password='password'
+        )
+
+        cls.allowed_group = Group.objects.create(name='allowed_group')
+        settings.VERTICALS_MANAGEMENT_GROUPS = ['allowed_group']
```
zawan-ila: Can we set this in `test.py` settings? And do a `get_or_create` on the `Group`?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,160 @@
...
+            course=self.course,
+            vertical=self.vertical,
+            sub_vertical=self.subvertical
+        )
+
+    def test_save_model_as_superuser(self):
+        """Verify that superuser can save course vertical."""
+        request = MockRequest(self.superuser)
+        try:
+            self.coursevertical_admin.save_model(request, self.coursevertical, None, False)
```
zawan-ila: Remove the `try-except`. If an exception is raised, the test would fail itself.

course_discovery/apps/tagging/tests/test_models.py
```diff
@@ -0,0 +1,136 @@
...
+    def test_cascade_delete(self):
+        """ Verify that sub-verticals are deleted when their vertical is deleted """
+        sub_vertical_id = self.sub_vertical.id
+        self.vertical.delete()
+
+        with self.assertRaises(SubVertical.DoesNotExist):
+            SubVertical.objects.get(id=sub_vertical_id)
+
+
+class CourseVerticalFilterModelTests(TestCase):
```
zawan-ila: `CourseVerticalmodelTests`?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,160 @@
...
+        except PermissionDenied:
+            self.fail("Superuser should be able to save course vertical")
+
+    def test_save_model_as_allowed_group_user(self):
+        """Verify that user in allowed group can save course vertical."""
+        self.regular_user.groups.add(self.allowed_group)
+        request = MockRequest(self.regular_user)
+        try:
```
zawan-ila: Same comment as above.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+
+class CourseVertical(ProductVertical):
+    """
+    Model for assigning vertical and sub verticals to courses
+    """
+    course = models.OneToOneField(
+        Course, on_delete=models.CASCADE, related_name="verticals", limit_choices_to={'draft': False}
+    )
+
+    class Meta(ProductVertical.Meta):
```
zawan-ila: Why do we need this Meta?

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+        unique_together = ["course"]
+
+    def clean(self):
+        """
+        Validate that the sub_vertical belongs to the selected vertical.
+        Automatically set the vertical if only sub_vertical is set.
+        """
+        super().clean()
+        if hasattr(self, 'sub_vertical') and self.sub_vertical:
```
zawan-ila, line 95: Is `self.sub_vertical` enough here?

zawan-ila (approved):
Some tests appear to be testing django code more than our code e.g cascading behavior, uniqueness constraints. I do not think such tests add much value but it is OK if you feel otherwise.

course_discovery/apps/tagging/models.py
AfaqShuaib09: Yes, as we are auto-assigning the vertical based on the sub-vertical and raising an error in case the selected vertical is not linked with the sub-vertical

course_discovery/apps/tagging/models.py
AfaqShuaib09: It won't be required now, as I have removed other meta fields from base model

course_discovery/apps/tagging/tests/test_models.py
AfaqShuaib09: Ah, forgot to remove Filter keywords from tests class

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+from django_extensions.db.fields import AutoSlugField
+from model_utils.models import TimeStampedModel
+from simple_history.models import HistoricalRecords
+
+from course_discovery.apps.course_metadata.models import Course
+
+
+class Vertical(TimeStampedModel):
+    """
+    Model for defining verticals used to categorize product types
```
DawoudSheraz: nit: categorize products, not product types.

course_discovery/apps/tagging/models.py
zawan-ila: Sorry for not being clear enough here. What I wanted to ask was if we could remove the `if hasattr(self, sub_vertical)` check and keep just `if self.sub_vertical`?

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+            cur_instance = Vertical.objects.get(pk=self.pk)
+            if cur_instance.is_active and not self.is_active:
+                self.sub_verticals.update(is_active=False)
+
+        super().save(*args, **kwargs)
+
+
+class SubVertical(TimeStampedModel):
+    """
+    Model for defining sub-verticals used to categorize product types under specific verticals.
```
DawoudSheraz: same nit comment as Vertical docstring.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        raise NotImplementedError("Subclasses must implement `get_object_title`.")
+
+
+class CourseVertical(ProductVertical):
+    """
+    Model for assigning vertical and sub verticals to courses
+    """
+    course = models.OneToOneField(
+        Course, on_delete=models.CASCADE, related_name="verticals", limit_choices_to={'draft': False}
```
DawoudSheraz: related_name should be "vertical" as this is a 1-1 relationship.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    class Meta:
+        unique_together = ["course"]
```
DawoudSheraz: is this check really needed, considering we have 1-1 relationship?

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,153 @@
...
+
+class BaseAdminTestCase(TestCase):
+    """ Base test class for admin tests """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.site = AdminSite()
+
+        # Create users and groups
```
DawoudSheraz: no need for this comment, the code is simple.

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.vertical_admin = VerticalAdmin(Vertical, self.site)
+        self.vertical = VerticalFactory()
+
+    def test_save_model_as_superuser(self):
+        """ Verify that superuser can save vertical. """
+        request = MockRequest(self.superuser)
+        self.vertical_admin.save_model(request, self.vertical, None, False)
```
DawoudSheraz, line 56: this test should explicitly check that the save does not raise any exception.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class Vertical(TimeStampedModel):
+    """
+    Model for defining verticals used to categorize product types
+    """
+    name = models.CharField(max_length=255, unique=True)
+    slug = AutoSlugField(populate_from='name', max_length=255, unique=True, db_index=True)
+    is_active = models.BooleanField(default=True)
+    history = HistoricalRecords()
```
DawoudSheraz: Please check https://github.com/openedx/course-discovery/pull/332 as AutoSlugField can cause issues with simple history.

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,153 @@
...
+        self.subvertical_admin = SubVerticalAdmin(SubVertical, self.site)
+        self.subvertical = SubVerticalFactory()
+
+    def test_save_model_as_superuser(self):
+        """Verify that superuser can save subvertical."""
+        request = MockRequest(self.superuser)
+        try:
+            self.subvertical_admin.save_model(request, self.subvertical, None, False)
+        except PermissionDenied:
+            self.fail("Superuser should be able to save subvertical")
```
DawoudSheraz: Same comment as before, the test should assert that no save exception is raised.

course_discovery/apps/tagging/tests/test_admin.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_save_model_as_superuser(self):
+        """Verify that superuser can save course vertical."""
+        request = MockRequest(self.superuser)
+        self.coursevertical_admin.save_model(request, self.coursevertical, None, False)
+
+    def test_save_model_as_allowed_group_user(self):
+        """Verify that user in allowed group can save course vertical."""
+        self.regular_user.groups.add(self.allowed_group)
+        request = MockRequest(self.regular_user)
+        self.coursevertical_admin.save_model(request, self.coursevertical, None, False)
```
DawoudSheraz: same assertion comments.

course_discovery/apps/tagging/tests/test_models.py
```diff
@@ -0,0 +1,136 @@
...
+        self.assertEqual(self.sub_vertical.vertical, self.vertical)
+        self.assertEqual(self.sub_vertical.slug, "software-engineering")
+        self.assertTrue(self.sub_vertical.is_active)
+
+    def test_unique_name_constraint(self):
+        """ Verify that sub-vertical names must be unique """
+        with self.assertRaises(Exception):
+            SubVerticalFactory(name="Software Engineering", vertical=self.vertical)
+
+    def test_cascade_delete(self):
```
DawoudSheraz: this is unnecessary as this is one of core framework features and it works out of the box.

course_discovery/apps/tagging/tests/test_admin.py
AfaqShuaib09: get_or_create isn't required here as it run only one time, since it's in def setUpClass method

course_discovery/apps/tagging/models.py
AfaqShuaib09: then it raise an error incase of no subvertical selection on CourseVertical creation "self has no attribute subvertical")

course_discovery/apps/tagging/models.py
AfaqShuaib09: yup, it won't be required here then

course_discovery/apps/tagging/tests/test_admin.py
AfaqShuaib09: Because it doesn't require anyhere , as this raise only permission issue if it's done by a non-superuser which is the next test case

course_discovery/apps/tagging/models.py
AfaqShuaib09: The code is still there, it will make extra entries then. I'm excluding slug field HistoricalRecords

course_discovery/apps/tagging/tests/test_admin.py
AfaqShuaib09: However, it's only for checking permissions, as it will raise a permission issue in case of a non-allowed user group. I'm now adding some assertions by changing a field

course_discovery/apps/tagging/admin.py
```diff
@@ -0,0 +1,86 @@
...
+            raise PermissionDenied("You are not authorized to perform this action.")
+        super().save_model(request, obj, form, change)
+
+
+@admin.register(CourseVertical)
+class CourseVerticalAdmin(SimpleHistoryAdmin):
+    """
+    Admin class for CourseVertical model.
+    """
+    list_display = ('course', 'vertical', 'sub_vertical')
```
DawoudSheraz, line 63: We might want to prefetch to avoid N+1 (https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_select_related)

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,113 @@
...
+class ProductVertical(TimeStampedModel):
+    """
+    Abstract base model for assigning vertical and sub verticals to product types.
+    """
+    vertical = models.ForeignKey(
+        Vertical, on_delete=models.CASCADE, null=True, blank=True, related_name="%(class)s_verticals"
+    )
+    sub_vertical = models.ForeignKey(
+        SubVertical, on_delete=models.CASCADE, null=True, blank=True, related_name="%(class)s_sub_verticals"
+    )
```
DawoudSheraz, line 60: Why do we do this %(class)? The migrations do not translate this to anything.

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,113 @@
...
+    history = HistoricalRecords()
```
DawoudSheraz: Is this working? I dont see HistoricalCourseVertical in migration file.

course_discovery/apps/tagging/tests/test_admin.py
DawoudSheraz: Yes, this is what we are asserting that saving as a superuser will not raise any errors. Not having any assertion does not make sense from a unit test perspective. The test only has statements and we are assuming if all is well, no statement will fail.

course_discovery/apps/tagging/models.py
AfaqShuaib09: just checked it's causing issues.
https://github.com/jazzband/django-simple-history/issues/600
https://django-simple-history.readthedocs.io/en/latest/historical_model.html#allow-tracking-to-be-inherited

course_discovery/apps/tagging/models.py
AfaqShuaib09: %(class) will pick the class name from child class
https://docs.djangoproject.com/en/dev/topics/db/models/#be-careful-with-related-name-and-related-query-name

course_discovery/apps/tagging/admin.py
AfaqShuaib09: I can check N+1 issues locally, but shouldn't we handle admin optimization in a separate/followup PR, like get_queryset optimize, dropdown widgets for course field etc.?

AfaqShuaib09: Added

course_discovery/apps/tagging/models.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def clean(self):
+        """
+        Validate that the sub_vertical belongs to the selected vertical.
+        Automatically set the vertical if only sub_vertical is set.
+        """
+        super().clean()
+        if hasattr(self, 'sub_vertical') and self.sub_vertical:
+            if not self.vertical:
+                self.vertical = self.sub_vertical.vertical  # Auto-assign vertical if it's not set
```
DawoudSheraz, line 97: missing coverage

DawoudSheraz (approved):

PR #4545, opened 2025-01-16
fix: handle draft only courses properly in archive_courses command

zawan-ila:
## [PROD-4289](https://2u-internal.atlassian.net/browse/PROD-4289)

Currently the `archive_course` command ignores courses without official versions. We change the behavior so that they are not.

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
@@ -90,7 +90,7 @@ class ArchiveCoursesCommandTests(TestCase, OAuth2Mixin):
     @ddt.unpack
     @responses.activate
     def test_success(self, from_db, mangle_title, mangle_end_date, draft_only):
-    # draft_only specifies the case when the objects only have draft versions
+        # draft_only specifies the case when the objects only have draft versions
         for model in [Course, CourseRun]:
             assert model.objects.count() == 2
             assert model.everything.count() == 4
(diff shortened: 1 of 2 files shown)


course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
```diff
@@ -63,8 +63,8 @@ def prepare_cmd_args(self, from_db, mangle_title, mangle_end_date):
             args.append('--mangle-end-date')
         return args
 
-    def verify_archived(self, course, mangle_title, mangle_end_date):
-        for c in [course, course.draft_version]:
+    def verify_archived(self, course, mangle_title, mangle_end_date, is_official=True):
+        for c in ([course, course.draft_version] if is_official else [course]):
```
DawoudSheraz, line 68: To confirm, the [course] in else is draft version, right? It is confusing because the other option has course, course.draft_version which then immediately takes it to mind that other [course] is non-draft in else.

course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
```diff
@@ -112,13 +119,16 @@ def test_success(self, from_db, mangle_title, mangle_end_date):
             args = ['--from-db']
         call_command('archive_courses', *args)
 
-        self.course1.refresh_from_db()
-        self.course2.refresh_from_db()
-        self.verify_archived(self.course1, mangle_title, mangle_end_date)
-        self.verify_not_archived(self.course2)
+        course1_or_its_draft = Course.everything.get(uuid=self.course1.uuid, draft=draft_only)
+        course2_or_its_draft = Course.everything.get(uuid=self.course2.uuid, draft=draft_only)
```
DawoudSheraz: the name is odd. Considering we are supplying `draft_only` as a param, it is understandable that course can be draft or non-draft. I think we can keep the regular `course` name here, no need to add _or_its_draft in here.

course_discovery/apps/course_metadata/management/commands/tests/test_archive_courses.py
zawan-ila: Yes. I have added a comment to clarify this.

DawoudSheraz (approved):

AfaqShuaib09 (approved):

PR #4547, opened 2025-01-20
feat: add django templates for tagging app UI

Ali-D-Akbar:
[PROD-4288](https://2u-internal.atlassian.net/browse/PROD-4288)

Adds UI the following pages. 
Be sure to install requirements and visit http://localhost:18381/tagging/courses/ to test the behaviors locally.

Screenshot 2025-01-22 at 11 13 35 PM

Screenshot 2025-01-22 at 11 13 50 PM

Screenshot 2025-01-22 at 11 14 00 PM

Screenshot 2025-01-22 at 11 14 14 PM

Screenshot 2025-01-22 at 11 14 30 PM

Screenshot 2025-01-22 at 11 14 43 PM


course_discovery/apps/tagging/forms.py
```diff
@@ -0,0 +1,17 @@
...
+
+class SubVerticalForm(forms.ModelForm):
+    class Meta:
+        model = SubVertical
+        fields = ['name', 'vertical', 'is_active']
+
+class CourseVerticalForm(forms.ModelForm):
+    class Meta:
+        model = CourseVertical
+        fields = ['vertical', 'sub_vertical']
```
AfaqShuaib09: Why is the course field not available in this form? Or is it used for editing purposes only?

course_discovery/apps/tagging/templates/tagging/base.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>{% block title %}Tagging App{% endblock %}</title>
+    <link
+        href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/css/bootstrap.min.css"
+        rel="stylesheet"
```
AfaqShuaib09: Have you verified bootstrap added here is working correctly?

course_discovery/apps/tagging/templates/partials/course_table.html
```diff
@@ -0,0 +1,106 @@
...
+                    {% endif %}
+                </a>
+            </th>
+            <th>
+                <a href="?sort=sub_vertical&direction={% if current_sort == 'sub_vertical' and current_direction == 'asc' %}desc{% else %}asc{% endif %}">
+                    Sub-Vertical
+                    {% if current_sort == 'sub_vertical' %}
+                        <span>{% if current_direction == 'asc' %}▲{% else %}▼{% endif %}</span>
+                    {% endif %}
+                </a>
```
DawoudSheraz, line 35: there is a lot of repeated code here. I think you can define a list and loop over it instead of having to write the same thing with only field name changes. Eager to see what others have to say. As an opposing arg, the table headers are "fixed", so the loop/general mechanism might not "give" much.

course_discovery/apps/tagging/templates/partials/course_table.html
```diff
@@ -0,0 +1,106 @@
...
+                        <a href="{% url 'tagging:vertical_detail' slug=course.vertical.vertical.slug %}">
+                            {{ course.vertical.vertical.name }}
+                        </a>
+                    {% else %}
+                        <span class="text-muted">None</span>
+                    {% endif %}
+                </td>
+                <td>
+                    {% if course.vertical and course.vertical.sub_vertical %}
+                        <a href="{% url 'tagging:sub_vertical_detail' slug=course.vertical.sub_vertical.slug %}">
```
DawoudSheraz: since subvertical is also linked in CourseVertical, you can use course.subvertical.slug instead of getting it from vertical.

course_discovery/apps/tagging/templates/tagging/course_detail.html
```diff
@@ -0,0 +1,50 @@
...
+        <div class="form-group">
+            <label for="vertical">Vertical</label>
+            <select name="vertical" id="vertical" class="form-control"
+                    hx-get="{% url 'tagging:load_sub_verticals' %}"
+                    hx-target="#sub_vertical"
+                    hx-trigger="change">
+                <option value="">-- Select Vertical --</option>
+                {% for vertical in verticals %}
+                    <option value="{{ vertical.slug }}"
+                            {% if course.vertical and course.vertical.vertical.slug == vertical.slug %}selected{% endif %}>
```
DawoudSheraz: course.vertical.vertical?

course_discovery/apps/tagging/tests/test_views.py
```diff
@@ -0,0 +1,192 @@
...
+        self.course1 = CourseFactory(title='Advanced Python')
+        self.course2 = CourseFactory(title='Python Basics')
+        self.url = reverse('tagging:course_list')
+
+    def test_get_course_list(self):
+        """Tests GET request to the course list view."""
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'tagging/course_list.html')
+        self.assertIn('courses', response.context)
```
DawoudSheraz: this should also check for course keys or titles

course_discovery/apps/tagging/tests/test_views.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def test_search_course(self):
+        """Tests searching for courses in the course list view."""
+        response = self.client.get(self.url, {'search': 'Basics'})
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, 'Python Basics')
+        self.assertNotContains(response, 'Advanced Python')
+
+    def test_sort_courses(self):
+        """Tests sorting courses by title in descending order."""
```
DawoudSheraz: This test can be a ddt to check for both directions.

course_discovery/apps/tagging/urls.py
```diff
@@ -0,0 +1,18 @@
...
+
+from course_discovery.apps.tagging.views import (
+    course_detail, course_list, load_sub_verticals, sub_vertical_detail, sub_vertical_list, vertical_detail,
+    vertical_list
+)
+
+app_name = 'tagging'
+
+urlpatterns = [
+    path('courses/<uuid:uuid>/', course_detail, name='course_detail'),
```
DawoudSheraz: nit: the name can be course_tagging_detail.

course_discovery/apps/tagging/urls.py
```diff
@@ -0,0 +1,18 @@
...
+    path('verticals/<slug:slug>/', vertical_detail, name='vertical_detail'),
+    path('sub_verticals/<slug:slug>/', sub_vertical_detail, name='sub_vertical_detail'),
+    path('courses/', course_list, name='course_list'),
+    path('verticals/', vertical_list, name='vertical_list'),
+    path('sub_verticals/', sub_vertical_list, name='sub_vertical_list'),
+    path('load-sub-verticals/', load_sub_verticals, name='load_sub_verticals'),
```
DawoudSheraz: this renders the template html file. I think we can find another way to achieve this, we don't need an explicit view to load subverticals.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,159 @@
+from django.core.paginator import Paginator
+from django.db.models import Count
+from django.http import HttpResponse, JsonResponse
+from django.shortcuts import get_object_or_404, redirect, render
+from django.template.loader import render_to_string
+
+from course_discovery.apps.tagging.models import Course, CourseVertical, SubVertical, Vertical
+
+
+def course_detail(request, uuid):
```
DawoudSheraz: nit: course_tagging_details or something along these lines.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Display details of a specific course.
```
DawoudSheraz: It is only meant for tagging, it does not display anything else.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,159 @@
...
+        return HttpResponse(html, status=200)
+
+    return render(request, "tagging/course_detail.html", {
+        "course": course,
+        "verticals": verticals,
+        "sub_vertical": sub_vertical,
+    })
+
+
+def load_sub_verticals(request):
```
DawoudSheraz: Left a comment about this in urls.py.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,159 @@
...
+    if search_query:
+        courses = courses.filter(title__icontains=search_query)
+
+    courses = courses.order_by(sort_field)
+
+    paginator = Paginator(courses, 10)
+    page_number = request.GET.get('page')
+    page_obj = paginator.get_page(page_number)
+
+    if request.headers.get('HX-Request'):
```
DawoudSheraz: What's this for?

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,159 @@
...
+        return HttpResponse(html)
+
+    return render(request, "tagging/course_list.html", {
+        "courses": page_obj,
+        "current_sort": sort_by,
+        "current_direction": direction,
+    })
+
+
+def vertical_list(request):
```
DawoudSheraz: nit: the views are typically verbs/actions, so we can rename them like list_Verticals, list_subverticals, etc. 
Plus, it would be good to use Django's list or detail view where needed.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,159 @@
+from django.core.paginator import Paginator
```
DawoudSheraz: The views are missing authentication and permissions.

course_discovery/apps/tagging/templates/partials/course_table.html
Ali-D-Akbar: That's the related name for CourseVertical :D https://github.com/openedx/course-discovery/blob/8ac5a0438d518aaffcd8899df90acbedd2a62f15/course_discovery/apps/tagging/models.py#L86

course_discovery/apps/tagging/templates/tagging/course_detail.html
Ali-D-Akbar: https://github.com/openedx/course-discovery/blob/8ac5a0438d518aaffcd8899df90acbedd2a62f15/course_discovery/apps/tagging/models.py#L86

course_discovery/apps/tagging/templates/partials/course_table.html
Ali-D-Akbar: Maybe update the name to `product_vertical`?

course_discovery/apps/tagging/views.py
Ali-D-Akbar: To handle htmx-based request to only update table component inside the course_list.html. If the headers have `HX-Request`, this means the request is dispatched when the search button is pressed towards `hx-get="{% url 'tagging:course_list' %}"`

Ali-D-Akbar: So when we `return HttpResponse(html)`
`hx-target="#course-table"` indicates where this html should be set.

course_discovery/apps/tagging/templates/partials/course_table.html
DawoudSheraz: Yes, product_vertical or something along these lines would be better.

course_discovery/apps/tagging/urls.py
Ali-D-Akbar: Updated this to load this on client-side using JS. Besides, we're gonna have a limited number of subverticals anyways.

course_discovery/apps/tagging/views.py
Ali-D-Akbar: Added.

course_discovery/apps/tagging/templates/partials/course_table.html
Ali-D-Akbar: I thought the same. Adding loop will make the code look a little hard to understand.

course_discovery/apps/tagging/mixins.py
```diff
@@ -0,0 +1,21 @@
+from django.conf import settings
+from django.http import HttpResponseForbidden
+
+
+class VerticalTaggingAdministratorPermissionRequiredMixin:
```
DawoudSheraz, line 5: should we define authentication_classes on this too? Only session based auth is allowed/needed here, we can explicitly define it here.

course_discovery/apps/tagging/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    A mixin to enforce permission on VERTICALS_MANAGEMENT_GROUPS for the class-based views.
+    """
+
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.is_authenticated:
+            return HttpResponseForbidden("You need to be logged in to access this page.")
```
DawoudSheraz: Django provides login required mixin and decorator, it can be used instead of writing this https://docs.djangoproject.com/en/5.1/topics/auth/default/#the-loginrequiredmixin-mixin

course_discovery/apps/tagging/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+
+        is_in_group = request.user.groups.filter(
```
DawoudSheraz: nit: in_vertical_management_group or something along these lines would be better in my opinion. is_in_group appears incomplete.

course_discovery/apps/tagging/templates/partials/course_table.html
DawoudSheraz: will this be addressed in this PR? I think this might get missed if it is left for the next PR.

course_discovery/apps/tagging/templates/partials/sub_vertical_options.html
```diff
@@ -0,0 +1,4 @@
+<option value="">-- Select SubVertical --</option>
```
DawoudSheraz: nit: Sub-Vertical

course_discovery/apps/tagging/templates/tagging/base.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>{% block title %}Tagging App{% endblock %}</title>
```
DawoudSheraz: The app name/title should not be Tagging App.

course_discovery/apps/tagging/templates/tagging/course_tagging_detail.html
```diff
@@ -0,0 +1,86 @@
...
+          hx-target="#message-container"
+          hx-swap="innerHTML">
+        {% csrf_token %}
+
+        <div class="form-group">
+            <label for="vertical">Vertical</label>
+            <select name="vertical" id="vertical" class="form-control"
+                    hx-trigger="change"
+                    hx-on="change: filterSubVerticals(event)">
+                <option value="">-- Select Vertical --</option>
```
DawoudSheraz: nit: you can remove -- at the start and end. I think the option is self-explanatory and does not need any other visual indicator.

course_discovery/apps/tagging/templates/tagging/course_tagging_detail.html
```diff
@@ -0,0 +1,86 @@
...
+        <div class="form-group">
+            <label for="sub_vertical">Sub-Vertical</label>
+            <select name="sub_vertical" id="sub_vertical" class="form-control">
+                <option value="">-- Select Sub-Vertical --</option>
+                {% for sub_vertical in all_sub_verticals %}
+                    <option value="{{ sub_vertical.slug }}"
+                            data-vertical="{{ sub_vertical.vertical.slug }}"
+                            {% if course.vertical and course.vertical.sub_vertical and course.vertical.sub_vertical.slug == sub_vertical.slug %}selected{% endif %}
+                            {% if course.vertical and not course.vertical.sub_vertical and sub_vertical.vertical.slug != course.vertical.vertical.slug %}style="display: none;"{% endif %}
+                            {% if not course.vertical or sub_vertical.vertical.slug != course.vertical.vertical.slug %}style="display: none;"{% endif %}>
```
DawoudSheraz: should these not be if-elif blocks, considering only one will be true at a time?

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,177 @@
...
+from course_discovery.apps.tagging.models import Course, CourseVertical, SubVertical, Vertical
+
+
+class CourseTaggingDetailView(VerticalTaggingAdministratorPermissionRequiredMixin, View):
+    """
+    Handles displaying course tagging details and assigning verticals and sub-verticals to a course.
+    """
+
+    def get(self, request, uuid):
+        course = get_object_or_404(Course, uuid=uuid, draft=False)
```
DawoudSheraz, line 18: Why should verticals and subverticals be fetched if course is 404? course_tagging_detail.html template does not seem to handle this i.e. what to show if course is not present.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,177 @@
...
+        verticals = Vertical.objects.all()
+        all_sub_verticals = SubVertical.objects.select_related('vertical')
+        return render(request, "tagging/course_tagging_detail.html", {
+            "course": course,
+            "verticals": verticals,
+            "all_sub_verticals": all_sub_verticals,
+        })
+
+    def post(self, request, uuid):
+        course = get_object_or_404(Course, uuid=uuid, draft=False)
```
DawoudSheraz, line 28: same comment as above. Passing incorrect id will break the code here.

course_discovery/apps/tagging/mixins.py
Ali-D-Akbar: It's the same amount of code if I use the mixin/decorator. Perhaps I can extend this mixin.

course_discovery/apps/tagging/templates/tagging/course_tagging_detail.html
Ali-D-Akbar: Updating this to if-elif.

course_discovery/apps/tagging/views.py
Ali-D-Akbar: Upon a course that doesn't exist, it will show this page:
`No Course matches the given query.`

Screenshot 2025-01-27 at 5 29 29 PM

DawoudSheraz: This is only shown for DEBUG=True, it won't be shown on prod. It will show the generic 404, I totally forgot that get_obj_or_404 raises 404 internally.

course_discovery/apps/tagging/mixins.py
```diff
@@ -0,0 +1,23 @@
...
+            return response
+
+        in_vertical_management_group = request.user.groups.filter(
+            name__in=settings.VERTICALS_MANAGEMENT_GROUPS
+        ).exists()
+
+        if not request.user.is_superuser and not in_vertical_management_group:
+            return HttpResponseForbidden("You do not have permission to access this page.")
+
+        return super().dispatch(request, *args, **kwargs)
```
hamza-56: The method calls `super().dispatch` twice.

course_discovery/apps/tagging/mixins.py
```diff
@@ -0,0 +1,23 @@
...
+        response = super().dispatch(request, *args, **kwargs)
+        if response.status_code == 403:
+            return response
+
+        in_vertical_management_group = request.user.groups.filter(
+            name__in=settings.VERTICALS_MANAGEMENT_GROUPS
+        ).exists()
+
+        if not request.user.is_superuser and not in_vertical_management_group:
+            return HttpResponseForbidden("You do not have permission to access this page.")
```
hamza-56:
```suggestion
            raise PermissionDenied("You do not have permission to access this page.")
```

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    Renders a list of all Courses with search, sort, and pagination capabilities.
+    """
+    model = Course
+    template_name = "tagging/course_list.html"
+    context_object_name = "courses"
+    paginate_by = 20
+
+    def get_queryset(self):
+        search_query = self.request.GET.get('search', '')
```
hamza-56:
```suggestion
        search_query = self.request.GET.get('search', '').strip()
```

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,177 @@
...
+        sort_field = sort_fields.get(sort_by, 'title')
+        if direction == 'desc':
+            sort_field = f'-{sort_field}'
+
+        queryset = Course.objects.prefetch_related("vertical__vertical", "vertical__sub_vertical")
+        if search_query:
+            queryset = queryset.filter(title__icontains=search_query)
+        return queryset.order_by(sort_field)
+
+    def get_context_data(self, **kwargs):
```
hamza-56, line 81: `ListView` already provides built-in pagination support. Can we leverage it?

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,177 @@
...
+        queryset = SubVertical.objects.annotate(course_count=Count('coursevertical_sub_verticals'))
+        if direction == 'desc':
+            sort_by = f'-{sort_by}'
+        return queryset.order_by(sort_by)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['current_sort'] = self.request.GET.get('sort', 'name')
+        context['current_direction'] = self.request.GET.get('direction', 'asc')
+        return context
```
hamza-56: nit: Both classes have similar logic for sorting. Can we create a common base class?

Something like:


```
class BaseSortableListView(VerticalTaggingAdministratorPermissionRequiredMixin, ListView):
    default_sort_field = 'name'

    def get_annotated_queryset(self):
        """
        Subclasses should override this method to provide their own annotated queryset.
        """
        raise NotImplementedError("Subclasses must implement `get_annotated_queryset`.")

    def get_queryset(self):
        sort_by = self.request.GET.get('sort', self.default_sort_field)
        direction = self.request.GET.get('direction', 'asc')

        if direction == 'desc':
            sort_by = f'-{sort_by}'

        queryset = self.get_annotated_queryset()
        return queryset.order_by(sort_by)

        context = super().get_context_data(**kwargs)
        context['current_sort'] = self.request.GET.get('sort', self.default_sort_field)
        context['current_direction'] = self.request.GET.get('direction', 'asc')
        return context
```  

```
class VerticalListView(BaseSortableListView):
    model = Vertical
    template_name = "tagging/vertical_list.html"
    context_object_name = "verticals"

    def get_annotated_queryset(self):
        return Vertical.objects.annotate(course_count=Count('coursevertical_verticals'))
```

```
class SubVerticalListView(BaseSortableListView):
    model = SubVertical
    template_name = "tagging/sub_vertical_list.html"
    context_object_name = "sub_verticals"

    def get_annotated_queryset(self):
        return SubVertical.objects.annotate(course_count=Count('coursevertical_sub_verticals'))
```

course_discovery/apps/tagging/mixins.py
Ali-D-Akbar: Ah. 🤦

course_discovery/apps/tagging/views.py
Ali-D-Akbar: While yes, it can result in a 404 error but I think it will never appear in a normal use since we'll always have a valid uuid for the course. 
A way to handle this would be to add a 404 page. Is it really necessary right now?

DawoudSheraz: No need to add a separate page, the server's default 404 will be displayed. I was just saying the above exact page wont be shown (obviously as it has dev/debug info).

course_discovery/apps/tagging/views.py
Ali-D-Akbar: Adding the new sorting capabilities.

course_discovery/apps/tagging/forms.py
Ali-D-Akbar: Not using forms since I'm using django-htmx

course_discovery/apps/tagging/templates/partials/course_table.html
Ali-D-Akbar: Okay I'll update this right now.

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class CourseListView(VerticalTaggingAdministratorPermissionRequiredMixin, ListView):
+    """
+    Renders a list of all Courses with search, sort, and pagination capabilities.
+    """
+    model = Course
+    template_name = "tagging/course_list.html"
+    context_object_name = "courses"
+    paginate_by = 2
```
AfaqShuaib09: [curious] why `paginate_by` assigned here with 2?

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,185 @@
...
+        return context
+
+    def render_to_response(self, context, **response_kwargs):
+        if self.request.headers.get('HX-Request'):
+            course_table_html = render_to_string("partials/course_table.html", context, request=self.request)
+            return HttpResponse(course_table_html)
+        return super().render_to_response(context, **response_kwargs)
+
+
+class BaseSortableListView(VerticalTaggingAdministratorPermissionRequiredMixin, ListView):
```
AfaqShuaib09, line 94: nit: docstring missing

course_discovery/apps/tagging/views.py
```diff
@@ -0,0 +1,185 @@
...
+    template_name = "tagging/sub_vertical_list.html"
+    context_object_name = "sub_verticals"
+
+    def get_annotated_queryset(self):
+        return SubVertical.objects.annotate(course_count=Count('coursevertical_sub_verticals'))
+
+
+class VerticalDetailView(VerticalTaggingAdministratorPermissionRequiredMixin, DetailView):
+    """
+    Display details of a specific vertical and associated courses.
```
AfaqShuaib09: nit: display doesn't sounds right here, you can add render here instead of display

course_discovery/apps/tagging/views.py
Ali-D-Akbar: Ah, I was just testing. Lemme update it back to 20.

course_discovery/apps/tagging/views.py
Ali-D-Akbar: Using the default pagination offered by `ListView` now

AfaqShuaib09 (approved):

course_discovery/apps/tagging/tests/test_views.py
```diff
@@ -0,0 +1,203 @@
...
+        super().setUp()
+        self.superuser = UserFactory(is_staff=True, is_superuser=True)
+        self.regular_user = UserFactory(is_staff=True, is_superuser=False)
+
+        self.allowed_group = Group.objects.create(name=settings.VERTICALS_MANAGEMENT_GROUPS)
+
+        self.regular_user.groups.add(self.allowed_group)
+
+
+class CourseTaggingDetailViewTests(BaseViewsTestCase):
```
DawoudSheraz, line 26: the script that displays the relevant subverticals based on selected verticals is not being tested here.

DawoudSheraz (approved):
one nit, otherwise looks good. I have not re-tested it locally.

hamza-56 (approved):

PR #4548, opened 2025-01-21
feat: add command to bulk update course verticals

zawan-ila:
[PROD-4282](https://2u-internal.atlassian.net/browse/PROD-4282)
Adds a management command to bulk update course verticals.


course_discovery/apps/tagging/management/commands/update_course_verticals.py
```diff
@@ -0,0 +1,85 @@
...
+                )
+                logger.info(f"Successfully set vertical and subvertical info for course with key {course_key}")
+
+        send_email_for_course_verticals_update(self.report, settings.COURSE_VERTICALS_UPDATE_RECIPIENTS)
+
+    def process_vertical_information(self, row):
+        course_key, vertical_name, subvertical_name = row['course'], row['vertical'], row['subvertical']
+        course = Course.objects.get(key=course_key)
+        vertical = Vertical.objects.filter(name=vertical_name).first()
+        subvertical = SubVertical.objects.filter(name=subvertical_name).first()
```
DawoudSheraz, line 60: This can be "cached" as one vertical/sub-vertical may be fetched again and again. While this job is expected to run only for backfill purposes, the runtime would reduce with this.

course_discovery/apps/tagging/management/commands/update_course_verticals.py
```diff
@@ -0,0 +1,85 @@
...
+        if (not vertical and vertical_name) or (not subvertical and subvertical_name):
+            raise ValueError("Incorrect vertical or subvertical provided")
+
+        if (vertical and not vertical.is_active) or (subvertical and not subvertical.is_active):
+            raise ValueError("The provided vertical or subvertical is not active")
```
DawoudSheraz, line 65: nit: instead of having or in the string, the code can be explicit to signify which object is not available or active.

course_discovery/settings/base.py
```diff
@@ -804,3 +804,5 @@
 
 # The list of user groups that have access to assign verticals and sub-verticals to courses
 VERTICALS_MANAGEMENT_GROUPS = []
+
+COURSE_VERTICALS_UPDATE_RECIPIENTS = ['phoenix@edx.org']
```
DawoudSheraz: nit: please keep the base file empty. If this is needed for testing, you can move it to test.py. Plus it would be good to have test-vibe values like test@example.com, or test@test.com, etc.

course_discovery/apps/tagging/management/commands/tests/test_update_course_verticals.py
```diff
@@ -0,0 +1,113 @@
...
+
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory
+from course_discovery.apps.tagging.models import CourseVertical, SubVertical, Vertical
+from course_discovery.apps.tagging.tests.factories import (
+    CourseVerticalFactory, SubVerticalFactory, UpdateCourseVerticalsConfigFactory, VerticalFactory
+)
+
+
+@ddt.ddt
+class UpdateCourseVerticalsCommandTests(TestCase):
```
DawoudSheraz, line 17: The tests are overall good. One suggestion: since the command is also maintaining the success & failure report, we should verify that in the unit tests where applicable.

DawoudSheraz: To add to ☝🏽 , we can check if send_email_for_course_verticals_update is being called with the correct report. 
Why is send_email_for_course_verticals_update not mocked here? Does it not cause issues as the email cannot be sent?

course_discovery/apps/tagging/admin.py
```diff
@@ -91,3 +91,11 @@ def save_model(self, request, obj, form, change):
         if not (request.user.is_superuser or request.user.groups.filter(name__in=allowed_groups).exists()):
             raise PermissionDenied("You are not authorized to perform this action.")
         super().save_model(request, obj, form, change)
+
+
+@admin.register(UpdateCourseVerticalsConfig)
+class UpdateCourseVerticalsConfigurationAdmin(admin.ModelAdmin):
+    """
+    Admin for ArchiveCoursesConfig model.
```
AfaqShuaib09: nit: Update docstring it should be UpdateCourseVerticalsConfig not ArchiveCoursesConfig

course_discovery/apps/tagging/emails.py
```diff
@@ -0,0 +1,29 @@
...
+        'failures': report['failures']
+    }
+    html_template = 'email/update_course_verticals.html'
+    template = get_template(html_template)
+    html_content = template.render(context)
+
+    email = EmailMessage(
+        "Update Course Verticals Command Summary",
+        html_content,
+        settings.PUBLISHER_FROM_EMAIL,
```
AfaqShuaib09, line 25: does this `PUBLISHER_FROM_EMAIL` required here

course_discovery/apps/tagging/management/commands/tests/test_update_course_verticals.py
```diff
@@ -0,0 +1,113 @@
...
+        assert self.course1.vertical.vertical == self.ai_vertical
+        assert self.course1.vertical.sub_vertical == self.python_subvertical
+        assert not hasattr(self.course2, 'vertical')
+        assert CourseVertical.objects.count() == 1
+
+    def test_raises_error_if_config_disabled(self):
+        with pytest.raises(CommandError):
+            csv = self.build_csv(self.csv_data)
+            UpdateCourseVerticalsConfigFactory(enabled=False, csv_file=csv)
+            call_command("archive_courses")
```
AfaqShuaib09: `call_command("archive_courses")`?

course_discovery/apps/tagging/emails.py
zawan-ila: Yes. This is the `From` part of the email.

course_discovery/apps/tagging/admin.py
zawan-ila: Effect of too much copy paste :-)

course_discovery/settings/base.py
zawan-ila: If we don't add it here, the command might break while trying to send mail. We have followed the same convention in the past. As for the address, I feel an `edx.org` subdomain is fine, as these are edx repos. Of course, one could still nitpick about `edx` vs `openedx` but I think its fine as is.

course_discovery/apps/tagging/management/commands/tests/test_update_course_verticals.py
zawan-ila: Added verification of email content in the tests.

zawan-ila: Django uses the locmem email backend automatically while testing

course_discovery/apps/tagging/management/commands/update_course_verticals.py
zawan-ila: I don't think the perf gain is enough to sacrifice the readability.

course_discovery/apps/tagging/management/commands/update_course_verticals.py
zawan-ila: The error will contain the course key, and from there it shouldn't be hard to verify which one is inactive. Adding multiple `ifs` reduces readability imo

DawoudSheraz: Hmmm, right. I guess its fine for now, we can revisit if needed.

course_discovery/settings/base.py
DawoudSheraz: Right. 
I will still advocate changing the email to test@test.com or test@example.com as this has been the pattern for most of the cases (do a quick search in the codebase). I could see phoenix@edx.org only for archival email setting, which was added by you 🙂

DawoudSheraz (approved):
1-2 comments, otherwise good. I tested it locally, it is working as expected.

course_discovery/settings/base.py
zawan-ila: Changed to `user@domain.org`.

course_discovery/apps/tagging/emails.py
AfaqShuaib09: ah, missed this

AfaqShuaib09 (approved):

PR #4552, opened 2025-01-30
fix: avoids call to DB when checking for queryset

Ali-D-Akbar:
Fixes the v2 endpoint slowness using `is None` instead of checking it using a boolean expression.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -1183,7 +1183,8 @@ class SearchAfterMixin:
             QuerySet
         """
         query = clean_query(query)
-        queryset = queryset or cls.objects.all()
+        if queryset is None:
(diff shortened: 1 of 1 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -1183,7 +1183,7 @@ def search(
             QuerySet
         """
         query = clean_query(query)
-        queryset = queryset or cls.objects.all()
+        queryset = cls.objects.all() if queryset is None else queryset
```
zawan-ila: [nit] I feel it is clearer to write 
`if queryset is None:
     queryset = cls.objects.all()`

zawan-ila (approved):

AfaqShuaib09 (approved):

PR #4556, opened 2025-02-03
fix: use .txt files inside .in files to ensure consistency across environments

zawan-ila:
## [PROD-3793](https://2u-internal.atlassian.net/browse/PROD-3793)

If we include `.in` files as requirements (instead of `.txt` files), there is a possibility that different envs (dev, prod, local, test etc.) may get different versions of the same package. An example is the attrs package in our repository. The [local env](https://github.com/openedx/course-discovery/blob/2884cc5cdc1c27b659b9bcadab553eaff3b95b80/requirements/local.txt#L40) has version 21.4.0 whereas the [prod env](https://github.com/openedx/course-discovery/blob/2884cc5cdc1c27b659b9bcadab553eaff3b95b80/requirements/production.txt#L32) has version 25.1.0

This PR is an attempt to ensure that package versions are consistent across environments.

It might be helpful to review the commits separately.


requirements/test.txt
```diff
@@ -0,0 +1,1029 @@
+#
```
DawoudSheraz, line 1: If test.txt did not exist before, is it being used anywhere in CI flow? or is everything under local.txt?

AfaqShuaib09 (approved):

requirements/test.txt
zawan-ila: I don't think it's being used anywhere. If it were, things would have broken

DawoudSheraz: Hmmm, strange.

DawoudSheraz: Should we not update test/CI to use test.txt? I mean the purpose of having the separate dependencies was to be able to use environment-specific dependencies.

zawan-ila: Updated the tests to use `test.txt`

DawoudSheraz (approved):

PR #4574, opened 2025-02-17
feat: send email to the enabled users only

Ali-D-Akbar:
[PROD-4302](https://2u-internal.atlassian.net/browse/PROD-4302)
Update vertical assignment email sender to require `UserAttributes` for proper notification filtering using `is_email_notification_enabled`.

### DIFF
--- a/course_discovery/apps/tagging/emails.py
+++ b/course_discovery/apps/tagging/emails.py
@@ -5,6 +5,8 @@ from django.core.mail import EmailMessage
 from django.template.loader import get_template
 from django.urls import reverse
 
+from course_discovery.apps.publisher.utils import is_email_notification_enabled
+
 logger = logging.getLogger(__name__)
 
 
@@ -37,8 +39,16 @@ def send_email_for_course_verticals_update(report, to_users):
 def send_email_for_course_vertical_assignment(course, to_users):
     """
     Sends an email to specified users requesting action to assign vertical and sub-vertical
-    for a given course.
+    for a given course, but only to those who have email notifications enabled.
     """
+    email_enabled_users = [user.email for user in to_users if is_email_notification_enabled(user)]
+    if not email_enabled_users:
+        logger.exception(
+            f"Failed to send vertical assignment email for course '{course.title}' (UUID: {course.uuid})"
+            f"No recipients found."
+        )
+        return
+
     course_tagging_url = (
         f"{settings.DISCOVERY_BASE_URL}{reverse('tagging:course_tagging_detail', kwargs={'uuid': course.uuid})}"
     )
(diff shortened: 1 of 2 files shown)


course_discovery/apps/tagging/emails.py
```diff
@@ -37,8 +39,12 @@ def send_email_for_course_verticals_update(report, to_users):
 def send_email_for_course_vertical_assignment(course, to_users):
     """
     Sends an email to specified users requesting action to assign vertical and sub-vertical
-    for a given course.
+    for a given course, but only to those who have email notifications enabled.
     """
+    enabled_emails = [user.email for user in to_users if is_email_notification_enabled(user)]
```
DawoudSheraz: nit: user_emails or something like that would be better instead of enabled_emails, in my opinion.

course_discovery/apps/tagging/emails.py
```diff
@@ -37,8 +39,12 @@ def send_email_for_course_verticals_update(report, to_users):
...
+    if not enabled_emails:
+        return
```
DawoudSheraz, line 50: We can log in if the email is not sent. Or maybe route it to COURSE_VERTICALS_UPDATE_RECIPIENTS if no enabled users are there?

DawoudSheraz (approved):

course_discovery/apps/tagging/emails.py
zawan-ila: `enabled_emails` -> `email_enabled_users`?

course_discovery/apps/tagging/tests/test_emails.py
```diff
@@ -42,3 +46,16 @@ def test_email_send_failure_logs_exception(self, mock_send):
...
         self.assertIn("Failed to send vertical assignment email", log_context.output[0])
+
+    def test_no_email_sent_when_user_notifications_disabled(self):
+        """
+        Test that if a user has disabled email notifications via their UserAttributes,
+        then no email is sent.
+        """
+
+        disabled_user = UserFactory(email="disabled@example.com")
+        publisher_factories.UserAttributeFactory(user=disabled_user, enable_email_notification=False)
```
zawan-ila: A direct import of `UserAttributeFactory` would make this more readable.

course_discovery/apps/tagging/emails.py
Ali-D-Akbar: Since we'll always have a user enabled, don't see any purpose in adding `COURSE_VERTICALS_UPDATE_RECIPIENTS` as recipients...

Ali-D-Akbar: Adding logs tho.

zawan-ila (approved):

Issue #4449, opened 2024-09-23
[DEPR]: Course Discovery Service


bradenmacdonald:
## Proposal Date

23 September 2024

## Ticket Acceptance Date

23 October 2024

## Technology Removal Date

2025-2027

## First Open edX Named Release Without This Functionality

TBD

## Rationale

The Course Discovery Service provides a variety of optional functionality in Open edX, some of which is required for other services like the [edx-enterprise](https://github.com/openedx/edx-enterprise/) suite. However, the fundamental architecture of the discovery service has proven to be problematic, and it is currently a limiting factor that makes it much more difficult than it should be to develop and integrate custom functionality for the Open edX platform such as ecommerce, enterprise integrations, subscriptions, and programs.

Here are some of the problems with its current architecture:
* **Non-default**: The discovery service is not installed by default on most instances of Open edX, which means that "core" functionality is never built using its APIs.
* **Out of Sync Course Data**: Fundamentally, the discovery service maintains at least _two_ separate sets of models about courses (i.e. the [`Course`](https://github.com/openedx/course-discovery/blob/98b93ed4470569bad637cf6ad8c993beaca2535e/course_discovery/apps/course_metadata/models.py#L1314) model in its MySQL database, and the [`CourseRunDocument`](https://github.com/openedx/course-discovery/blob/98b93ed4470569bad637cf6ad8c993beaca2535e/course_discovery/apps/course_metadata/search_indexes/documents/course_run.py#L26) in Elasticsearch). Though there are various systems for keeping this data in sync with Studio (which holds the actual "source of truth"), they are imperfect and **synchronization issues are a frequent source of bugs and errors**.
* **Bi-directional sync**: Because the discovery service is the source of truth for _some_ pieces of data, it actually syncs its data _to_ Studio as well as from Studio. This has caused a variety of issues, including bugs with an infinite cycle of updates.
[...]

PR #4575, opened 2025-02-17
feat: add support for pathway retirement

zawan-ila:
### [PROD-2973](https://2u-internal.atlassian.net/browse/PROD-2973)

This PR:-

- Adds the ADR for pathways retirement
- Adds the status field to Pathway objects
- Exposes the status field in the Pathway APIs
- Allows filtering the pathway api by status


docs/decisions/0032-pathway-retirement.rst
```diff
@@ -0,0 +1,35 @@
...
+Currently, there is no way to mark a pathway as retired. This is often necessary due to changing requirements
+for credit recognition on the partner organization's end, or the discontinuation of programs offered by them.
+In such cases, these pathways should be hidden from the learner dashboard and any credit requests against them
+should not be accepted.
+
+Decision
+----------
+A new field `status` will be added to the pathway model. This field will support three possible values: `unpublished`,
+`published` and `retired`. Existing pathways will be assigned the "published" status, while any new pathways will be set
+to "unpublished" by default.
```
DawoudSheraz: > Existing pathways will be assigned the "published" status, while any new pathways will be set
> to "unpublished" by default

It can be highlighted that this will be done using a mechanism, such as Django admin action, manual intervention, etc., and that the status field will always default to "unpublished" on creation.

docs/decisions/0032-pathway-retirement.rst
```diff
@@ -0,0 +1,35 @@
...
+Context
+---------
+Currently, there is no way to mark a pathway as retired. This is often necessary due to changing requirements
+for credit recognition on the partner organization's end, or the discontinuation of programs offered by them.
+In such cases, these pathways should be hidden from the learner dashboard and any credit requests against them
+should not be accepted.
+
+Decision
+----------
+A new field `status` will be added to the pathway model. This field will support three possible values: `unpublished`,
```
DawoudSheraz: The backticks around status and status values do not have visual changes when previewing the file.

AfaqShuaib09 (commented):
ADR wise, LGTM. Not approving it yet since you mentioned you're working on the code changes

docs/decisions/0032-pathway-retirement.rst
zawan-ila: Added a small explanation in parentheses.

docs/decisions/0032-pathway-retirement.rst
zawan-ila: Fixed.

course_discovery/apps/course_metadata/migrations/0347_pathway_status.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.18 on 2025-02-17 15:50
+
+from django.db import migrations, models
```
DawoudSheraz, line 3: nit: empty line before course_discovery import

DawoudSheraz (commented):
- Please update/add tests
- Update PR title and description

DawoudSheraz (approved):

justinhynes:
The ADR and changes seem reasonable to me. 

Question: Did this change need to be presented as a proposal to Product at Axim or the Product working group for Open edX?

deborahgu:
this seems sensible to me.

re: @justinhynes's quetion: I'm not sure if Axim  would feel anything going into course discovery (except breaking changes) would require going through the product WG, since the plan is to deprecate course-discovery, I believe. (#4449) That being said, I have mixed feelings over whether or not this constitutes a breaking change. Arguably if we coordinate the rollout of all of the different changes (e.g.  course-discovery, credentials, etc.) it's all just internals, but the "arguably" is doing a lot of work there. tl;dr  I agree that it's worth clearing this through product review.

DawoudSheraz:
> this seems sensible to me.
> 
> re: @justinhynes's quetion: I'm not sure if Axim would feel anything going into course discovery (except breaking changes) would require going through the product WG, since the plan is to deprecate course-discovery, I believe. (#4449) That being said, I have mixed feelings over whether or not this constitutes a breaking change. Arguably if we coordinate the rollout of all of the different changes (e.g. course-discovery, credentials, etc.) it's all just internals, but the "arguably" is doing a lot of work there. tl;dr I agree that it's worth clearing this through product review.

I don't think we need anything from Axim on this, especially after the deprecation. I don't think this is a breaking change. If we were only sending the published pathways from API after this change, that would have constituted the breaking change. However, as Ali Nawaz added in ADR, we have added the filter capability in API and then leaving it to consumers to handle it as they see fit.

AfaqShuaib09 (approved):

PR #4577, opened 2025-02-18
Support for french medium point in clean_html

hamza-56:
### [Support for French Medium Point (PROD-3248)](https://2u-internal.atlassian.net/browse/PROD-3248)

in `clean_html` function:

1. BeautifulSoup parses the HTML and replaces any `&middot;` entities with their literal "`·`" character. 
2. The code then converts those literal "`·`" characters back into `&middot;` entities. 
 (`cleaned = cleaned.replace('·', '&middot;')`)
3. The `html2text` converter detects these `&middot;` entities and transforms them into `*` for Markdown list syntax.
4. Finally, Markdown processes the `*`, rendering them as proper `<ul>` and `<li>` elements in HTML.

This PR removes `cleaned = cleaned.replace('·', '&middot;')`. As a result:

1. BeautifulSoup still replaces any `&middot;` entities with their literal "`·`" character. 
2. However, `html2text` now sees these literal "`·`" characters and doesn’t recognize them as list markers.
3. So instead of converting them to Markdown list syntax, it just keeps them as literal characters
4. When Markdown processes the text, it treats them as plain characters rather than list markers.


Ali-D-Akbar (approved):
Please change the commit type to `fix`.

AfaqShuaib09 (approved):

PR #4582, opened 2025-02-24
fix: send users list in vertical tagging email

DawoudSheraz:
Followup to https://github.com/openedx/course-discovery/pull/4574 and fixes the email method by sending users instead of user emails.


Ali-D-Akbar (approved):

zawan-ila (commented):
This fixes the error, but does it preserve the behavior? The log in the exception will throw too? We probably need a change in the `signals` file, from where the method is called itself.

PR #4585, opened 2025-02-24
feat: add ai_languages field

zawan-ila:
### [PROD-4306](https://2u-internal.atlassian.net/browse/PROD-4306)

This PR adds a new field `ai_languages` on the CourseRun model. This field is intended to replace the current `translation_languages` field. However, we do not remove the `translation_languages` field to prevent any issues with rolling deployments where the "older" servers would still be expecting the `translation_languages` field to exist.


Ali-D-Akbar (approved):

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/models.py
```diff
@@ -2281,6 +2281,12 @@ class CourseRun(ManageHistoryMixin, DraftModelMixin, CachedMixin, TimeStampedMod
                     'and the language label. These entries represent the languages into which the run '
                     'content can be translated.')
     )
+    ai_languages = models.JSONField(
+        null=True,
+        blank=True,
+        help_text=_('A JSON dict detailing the available translations and transcriptions for this run. '
+                    'The dict has two keys: translation_languages and transcription_languages.')
```
DawoudSheraz, line 2288: - The keys should reflect the content and transcription languages. `translation_languages` does not reflect that it is related to content.

DawoudSheraz: Well, based on internal discussions, it is ok to leave it as it is. The existing fields in CourseRun are language and transcript_languages where language corresponds to content language.

DawoudSheraz (approved):

PR #4593, opened 2025-03-03
fix: explicitly add list_serializer_class

Ali-D-Akbar:
[PROD-4304](https://2u-internal.atlassian.net/browse/PROD-4304)
Inheriting meta class doesn't seem to have been setting the `list_serializer_class` which ends up calling the `to_representation` of the DRF version and not our custom ListSerializers of Course and LearnerPathway documents.

This PR sets the custom list serializers correctly and rearranges the order in which these serializers are being inherited.


AfaqShuaib09 (commented):
Can we add some sort of assertion in tests to monitor the query count?

AfaqShuaib09 (approved):

DawoudSheraz (approved):

hamza-56 (approved):

PR #4594, opened 2025-03-05
Monitor django management commands

hamza-56:
### [PROD-4297](https://2u-internal.atlassian.net/browse/PROD-4297)

This PR adds the `monitor_django_management_command` utility, allowing the execution of management commands to be explicitly tracked in monitoring tools.


course_discovery/settings/base.py
```diff
@@ -807,3 +807,5 @@
 
 COURSE_VERTICALS_UPDATE_RECIPIENTS = ['user@domain.org']
 HEADER_LOGO_URL= 'https://edx-cdn.org/v3/default/logo.svg'
+
+ENABLE_CUSTOM_MANAGEMENT_COMMAND_MONITORING = True
```
AfaqShuaib09: I think by default we can keep it False and set it to True in `edx-internal` repo

course_discovery/apps/core/utils.py
```diff
@@ -252,3 +254,53 @@ def update_instance(instance, data, should_commit=False, **kwargs):
         instance.save(**kwargs)
 
     return instance, updated
+
+
+# pylint: disable=setting-boolean-default-value
+def _get_enable_custom_management_command_monitoring():
```
Ali-D-Akbar: What is the need to define a setting like this? Or maybe we can create some sort of a waffle switch instead of having this?

Ali-D-Akbar (commented):
Added a suggestion. Let me know your thoughts on this.

course_discovery/settings/base.py
DawoudSheraz: +1

manage.py
```diff
@@ -11,5 +11,10 @@
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "course_discovery.settings.local")
 
     from django.core.management import execute_from_command_line
+    from course_discovery.apps.core.utils import monitor_django_management_command
```
DawoudSheraz: lets move the import behind settings as we had originally planned.

course_discovery/settings/base.py
hamza-56: Updated ✅

manage.py
hamza-56: Updated ✅

course_discovery/apps/core/utils.py
hamza-56: A waffle switch would add a DB query on every management command execution, impacting performance. Since `manage.py` is critical, we want minimal overhead. A settings flag keeps it lightweight and efficient.

course_discovery/apps/core/utils.py
```diff
@@ -252,3 +254,52 @@ def update_instance(instance, data, should_commit=False, **kwargs):
         instance.save(**kwargs)
 
     return instance, updated
+
+
+# pylint: disable=setting-boolean-default-value
```
DawoudSheraz: why is this disable needed?

course_discovery/apps/core/utils.py
```diff
@@ -252,3 +254,52 @@ def update_instance(instance, data, should_commit=False, **kwargs):
...
+def _get_enable_custom_management_command_monitoring():
+    """
+    Returns whether custom Django management command monitoring is enabled.
+
+    .. setting_name: ENABLE_CUSTOM_MANAGEMENT_COMMAND_MONITORING
+    .. setting_default: False
+    .. setting_description: A boolean flag that determines whether Django management command executions
+        should be explicitly monitored.
```
DawoudSheraz: Where is this format coming from? Do we need this considering the code is one-liner and self-explanatory?

course_discovery/apps/core/utils.py
```diff
@@ -252,3 +254,52 @@ def update_instance(instance, data, should_commit=False, **kwargs):
         instance.save(**kwargs)
 
     return instance, updated
+
+
+# pylint: disable=setting-boolean-default-value
+def _get_enable_custom_management_command_monitoring():
```
DawoudSheraz: why is func name starting with _?

course_discovery/apps/core/utils.py
```diff
@@ -252,3 +254,52 @@ def update_instance(instance, data, should_commit=False, **kwargs):
...
+def _get_django_management_monitoring_function_trace_name():
+    """
+    Returns the function trace name used for monitoring Django management commands.
+
+    .. setting_name: DJANGO_MANAGEMENT_MONITORING_FUNCTION_TRACE_NAME
+    .. setting_default: 'django.command'
+    .. setting_description: This setting specifies the function trace name that will be used when monitoring
+        Django management commands. It helps group and categorize monitoring spans for better visibility.
+    .. setting_warning: Ensure this setting is properly configured to align with your monitoring tools,
+        as an incorrect name may lead to untracked or misclassified spans.
```
DawoudSheraz: Same comments as above. Since there are no arguments, the important information can be made part of docstring instead of this format.

course_discovery/apps/core/utils.py
hamza-56: Removed ✅

course_discovery/apps/core/utils.py
hamza-56: Removed ✅

course_discovery/apps/core/utils.py
hamza-56: Removed ✅

course_discovery/apps/core/utils.py
hamza-56: Removed ✅.
Now we're reading from `settings` directly instead of using helper functions.

DawoudSheraz (approved):

AfaqShuaib09 (approved):

PR #4608, opened 2025-04-10
feat: add config model for bulk operations

AfaqShuaib09:
[PROD-4349](https://2u-internal.atlassian.net/browse/PROD-4349)
---------

This PR introduces following changes:
- Added BulkTasksConfig model for managing CSV uploads and bulk operation metadata
- Integrated django-celery-results to associate task_id with task results
- Updated Django admin for easier monitoring of bulk task configurations

##### Creation Form:

image

##### Update Form:

image


course_discovery/apps/course_metadata/admin.py
```diff
@@ -361,6 +361,28 @@ def save_model(self, request, obj, form, change):
...
+        if obj:
+            return ['csv_file', 'task_type', 'status', 'task_id', 'uploader', 'job_summary', 'get_task_result']
+        else:
+            return ['csv_file', 'task_type']
+
+    def save_model(self, request, obj, form, change):
+        """
+        Save the model instance and set the uploader to the current user.
+        """
+        obj.uploader = request.user
```
DawoudSheraz: Should this not be done only on creation?

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
 
     def __str__(self):
         return f"{self.course_run.key}: <{self.restriction_type}>"
+
+
+class BulkTasksConfig(TimeStampedModel):
```
DawoudSheraz: Curious: Can/should this be moved to another app?

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+    # Timeout set to 0 so that the model does not read from cached config in case the config entry is deleted.
+    cache_timeout = 0
+    csv_file = models.FileField(
+        validators=[FileExtensionValidator(allowed_extensions=['csv'])],
+        help_text=_("It expects the data will be provided in a csv file format "
+                    "with first row containing all the headers."),
+        blank=False,
+        null=False
+    )
+    uploader = models.ForeignKey(
```
DawoudSheraz: nit: this can be renamed to uploaded_by or similar (as it is done in other config model).

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
     def __str__(self):
         return f"{self.course_run.key}: <{self.restriction_type}>"
+
+
+class BulkTasksConfig(TimeStampedModel):
+    """
+    Configuration to store a csv file that will be used in bulk_tasks.
+    """
+    # Timeout set to 0 so that the model does not read from cached config in case the config entry is deleted.
+    cache_timeout = 0
```
DawoudSheraz: This cache_timeout is not needed anymore.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+    )
+    uploader = models.ForeignKey(
+        User,
+        models.CASCADE,
+        related_name='bulk_tasks_config',
+        null=True,
+        blank=True,
+        help_text=_('User who uploaded the file')
+    )
+    job_summary = models.TextField(
```
DawoudSheraz: should this be job_Summary or task_summary? The model name is BulkTasksConfig, which does not mention job. Let's be consistent with naming.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: The model can be renamed (BulkOperationJob, BulkOperationTask, etc.). It is not a config model. We should not use config word as config model carries a certain meaning in Open edX repositories.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+    status = models.CharField(
+        max_length=50,
+        choices=[
+            ('Pending', 'Pending'),
+            ('Verified', 'Verified'),
+            ('Processing', 'Processing'),
+            ('Completed', 'Completed'),
+            ('Failed', 'Failed'),
+            ('Partially Completed', 'Partially Completed'),
+            ('Errored', 'Errored'),
```
DawoudSheraz: BulkOperationType is in a class. The same can be done here, to make the code consistent.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+    def save(self, *args, **kwargs):
+        user = kwargs.pop('user', None)
+        if not self.uploader:
+            self.uploader = user
+        super().save(*args, **kwargs)
+
+    def __str__(self):
+        return f"{self.csv_file.name} - {self.task_type} - {self.status} - {self.uploader.username}"
+
+    def get_task_result(self):
```
DawoudSheraz: nit: Let's make this a property.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+        if not self.task_id:
+            return "No task ID"
+        try:
+            result = TaskResult.objects.get(task_id=self.task_id)
+            return f"{result.status}: {result.result}"
+        except TaskResult.DoesNotExist:
+            return "No result found"
+
+    get_task_result.short_description = "Task Result"
```
DawoudSheraz: huh, why is this needed?

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4802,77 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+
+    def __str__(self):
+        return f"{self.csv_file.name} - {self.task_type} - {self.status} - {self.uploader.username}"
+
+    def get_task_result(self):
+        if not self.task_id:
+            return "No task ID"
+        try:
+            result = TaskResult.objects.get(task_id=self.task_id)
+            return f"{result.status}: {result.result}"
```
DawoudSheraz: This is misleading, the code is not returning task result. It returns a string representation of the task result. Please update this.

course_discovery/apps/course_metadata/admin.py
AfaqShuaib09: Okay, I will move this to creation only. I thought we needed to keep track of who last modified it.

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: A new app? I think creating a new app for a single model would be overkill here, but I'm open to thoughts on this

AfaqShuaib09: Yeah, it makes more sense now since we're not inheriting from ConfigurationModel anymore

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: to show it as an attribute on django admin
image

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: Actually, I want to display the status along with the result in the model, so I added it as a string instead of simply returning the result.
Should we update the method to return only the result, or update the method's docstring?

course_discovery/apps/course_metadata/admin.py
DawoudSheraz: What purpose will it serve?

course_discovery/apps/course_metadata/models.py
DawoudSheraz: No, not a new app. Any existing app, like core or publisher, can do it. This is a suggestion/question.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: The docstring and name need to be updated. We should have a similar property that allows getting the TaskResult object against a task summary object. It will prove useful.

course_discovery/apps/course_metadata/admin.py
```diff
@@ -361,6 +361,29 @@ def save_model(self, request, obj, form, change):
             messages.add_message(request, messages.ERROR, msg)
 
 
+@admin.register(BulkOperationTask)
+class BulkOperationTaskAdmin(admin.ModelAdmin):
+    """
+    Admin for BulkOperationTask model.
+    """
+    readonly_fields = ['status', 'uploaded_by', 'task_summary', 'task_result']
+    list_display = ('id', 'csv_file', 'task_type', 'uploaded_by', 'status', 'task_id', 'task_result')
```
DawoudSheraz: nit: That many fields are not needed in list display. Please add the relevant fields only. You can add some fields to list_filter if it is filtering or easy access being targeted here.

course_discovery/apps/course_metadata/choices.py
```diff
@@ -47,6 +47,30 @@ class ReportingType(models.TextChoices):
...
+class BulkOperationType(models.TextChoices):
+    """
+    The types of bulk operations that can be performed.
+    """
+    BulkCourseCreate = 'bulk_course_create', _('Bulk Course Create')
+    BulkCourseUpdate = 'bulk_course_update', _('Bulk Course Update')
+    BulkCoursePartialUpdate = 'bulk_course_partial_update', _('Bulk Course Partial Update')
+    BulkCourseRerunCreate = 'bulk_course_rerun_create', _('Bulk Course Rerun Create')
+    BulkCourseRunUpdate = 'bulk_course_run_update', _('Bulk Course Run Update')
+    BulkCourseRunPartialUpdate = 'bulk_course_run_partial_update', _('Bulk Course Run Partial Update')
```
DawoudSheraz: do the choices need Bulk as prefix, given the class name has Bulk in it already?

course_discovery/apps/course_metadata/choices.py
```diff
@@ -47,6 +47,30 @@ class ReportingType(models.TextChoices):
...
 
 
+class BulkOperationType(models.TextChoices):
+    """
+    The types of bulk operations that can be performed.
+    """
+    BulkCourseCreate = 'bulk_course_create', _('Bulk Course Create')
+    BulkCourseUpdate = 'bulk_course_update', _('Bulk Course Update')
+    BulkCoursePartialUpdate = 'bulk_course_partial_update', _('Bulk Course Partial Update')
+    BulkCourseRerunCreate = 'bulk_course_rerun_create', _('Bulk Course Rerun Create')
```
DawoudSheraz: CourseReRun should be enough. re-run is a creation process, there is no need to add creation explicitly.

course_discovery/apps/course_metadata/choices.py
```diff
@@ -47,6 +47,30 @@ class ReportingType(models.TextChoices):
     other = 'other', _('other')
 
 
+class BulkOperationType(models.TextChoices):
+    """
+    The types of bulk operations that can be performed.
+    """
+    BulkCourseCreate = 'bulk_course_create', _('Bulk Course Create')
+    BulkCourseUpdate = 'bulk_course_update', _('Bulk Course Update')
+    BulkCoursePartialUpdate = 'bulk_course_partial_update', _('Bulk Course Partial Update')
```
DawoudSheraz: Hmmm, the csv template are the same when it comes to create, update and partial updates. Are 3 statuses needed here? Won't create and update be enough?

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,73 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
 
     def __str__(self):
         return f"{self.course_run.key}: <{self.restriction_type}>"
+
+
+class BulkOperationTask(TimeStampedModel):
+    """
+    Configuration to store a csv file that will be used in bulk_tasks.
```
DawoudSheraz: Please update the docstring.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,73 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+        models.CASCADE,
+        related_name='bulk_tasks_config',
+        null=True,
+        blank=True,
+        help_text=_('User who uploaded the file')
+    )
+    task_summary = models.TextField(
+        null=True,
+        blank=True,
+        help_text=_('Summary of the job')
```
DawoudSheraz: nit: help_text to be updated.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4215,3 +4217,59 @@ def test_basic(self):
...
+
+
+class BulkOperationTaskTest(TestCase):
+    def test_bulk_tasks_config(self):
+        """
+        Verify that the bulk tasks config is set up correctly.
+        """
+        bulk_tasks_config = BulkOperationTaskFactory()
+        assert bulk_tasks_config.task_type == BulkOperationType.BulkCourseCreate
+        assert bulk_tasks_config.task_type == 'bulk_course_create'
```
DawoudSheraz: The first assertion is enough.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,73 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+        user = kwargs.pop('user', None)
+        if not self.uploaded_by:
+            self.uploaded_by = user
+        super().save(*args, **kwargs)
+
+    def __str__(self):
+        return f"{self.csv_file.name} - {self.task_type} - {self.status} - {self.uploaded_by.username}"
+
+    @property
+    def task_result(self):
```
DawoudSheraz, line 4866: This should be renamed to task_result_status_summary or similar. The property is not returning the task result. Or better, create a new method task_result_status_summary that does what this property is doing. This property should return the TaskResult object if present.

course_discovery/apps/course_metadata/choices.py
AfaqShuaib09: Partial update option is somewhat different from simple update operations here. In partial update, we will be treating an empty field as if we don't want to update the field, whereas in a full update, we will be treating it as if we want to empty the field

course_discovery/apps/course_metadata/admin.py
AfaqShuaib09: Since the user who will last edit the object is likely to update the file (in case it's the wrong file or misformatted), I wanted to set `last_modified` as `uploaded_by`. However, saving it only on creation also sounds good to me.

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: Oh okay, publisher sounds more relevant here instead of the core app. But I think we can keep here it in course_metadata app, since most of the models info lies here — though I don't have strong opinions on this.

course_discovery/apps/course_metadata/admin.py
AfaqShuaib09: Removed `id` and `task_result` from `list_display` & added `task_type` and `status` in `list_filter` list

course_discovery/apps/course_metadata/choices.py
AfaqShuaib09: Hmm, I think we can remove Bulk from here since the class name already includes it -- it looks cleaner code-wise. No strong opinions though, both look fine to me.

course_discovery/apps/course_metadata/admin.py
DawoudSheraz: If the last_modified_by is needed, feel free to add it.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Let's wait for others to add their input.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,86 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+        validators=[FileExtensionValidator(allowed_extensions=['csv'])],
+        help_text=_("It expects the data will be provided in a csv file format "
+                    "with first row containing all the headers."),
+        blank=False,
+        null=False
+    )
+    uploaded_by = models.ForeignKey(
+        User,
+        models.CASCADE,
+        related_name='bulk_tasks_config',
```
DawoudSheraz: Please update the related name.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4215,3 +4217,86 @@ def test_basic(self):
         self.assertEqual(course_run.restricted_run, restricted_course_run)
         self.assertEqual(restricted_course_run.restriction_type, 'custom-b2b-enterprise')
         self.assertEqual(str(restricted_course_run), "course-v1:SC+BreadX+3T2015: <custom-b2b-enterprise>")
+
+
+class BulkOperationTaskTest(TestCase):
+    def test_bulk_tasks_config(self):
```
DawoudSheraz: nit: remove the word "config" from the relevant places.

course_discovery/apps/course_metadata/choices.py
```diff
@@ -47,6 +47,28 @@ class ReportingType(models.TextChoices):
...
 
 
+class BulkOperationType(models.TextChoices):
+    """
+    The types of bulk operations that can be performed.
+    """
+    CourseCreate = 'course_create', _('Course Create')
+    CoursePartialUpdate = 'course_partial_update', _('Course Partial Update')
+    CourseRerun = 'course_rerun', _('Course Rerun')
+    CourseRunPartialUpdate = 'course_run_partial_update', _('Course Run Partial Update')
```
Ali-D-Akbar, line 57: Why are we separating partial updates for course and runs? My understanding was to have either full update or partial update.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,86 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+    )
+
+    def save(self, *args, **kwargs):
+        user = kwargs.pop('user', None)
+        if not self.uploaded_by:
+            self.uploaded_by = user
+        super().save(*args, **kwargs)
+
+    def __str__(self):
+        return f"{self.csv_file.name} - {self.task_type} - {self.status} - {self.uploaded_by.username}"
```
Ali-D-Akbar: Check if `uploaded_by` is none or not since we usually allow this field to be null.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,86 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+        max_length=255,
+        unique=True,
+        null=True,
+        blank=True,
+        help_text=_('Identifier of the celery task associated with this bulk task')
+    )
+
+    def save(self, *args, **kwargs):
+        user = kwargs.pop('user', None)
+        if not self.uploaded_by:
```
Ali-D-Akbar, line 4855: Do we need to check if we're getting user from kwargs just in case? Maybe update this as `not self.uploaded_by and user`

Ali-D-Akbar (approved):
Looks good overall. 👀

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Good point. The code can raise an error if the user and uploaded_by are null.

course_discovery/apps/course_metadata/models.py
```diff
@@ -4801,3 +4803,86 @@ class RestrictedCourseRun(DraftModelMixin, TimeStampedModel):
...
+        """
+        if not self.task_id:
+            return "No task ID"
+        try:
+            result = TaskResult.objects.get(task_id=self.task_id)
+            return f"{result.status}: {result.result}"
+        except TaskResult.DoesNotExist:
+            return "No result found"
+
+    task_result_status_summary.fget.short_description = 'Task Result Status Summary'
```
hamza-56: [curious] Why is `task_result_status_summary.fget.short_description` used here? Wouldn't it be better to set `short_description` in `admin.py` instead of modifying property descriptors directly?

DawoudSheraz (approved):
This will be merged once the new comments have been resolved.

course_discovery/apps/course_metadata/choices.py
AfaqShuaib09: Although we can handle course run partial updates within course partial updates, I think it's better to have them as separate bulk tasks. This is because run partial updates only require the course run key as an identifier, whereas course partial updates require the course number and org key (which together form the course key) to identify the course

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: Added the check here, however, it feels unnecessary after the check you mentioned to add in the save method. But good to have as a safety check

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: Added

course_discovery/apps/course_metadata/models.py
AfaqShuaib09: Yes, we could define the short_description in admin.py, but then we would also need to add it as a method in the Admin class. Therefore, I went with this approach.

i.e;

`admin.py`
```python
def task_result_status_summary(self, obj):
        return obj.task_result_status_summary
    task_result_status_summary.short_description = 'Task Result Status Summary'
```

hamza-56: I believe defining it directly in the `admin.py` is better because:

1. It keeps the display logic encapsulated within the admin, where it belongs.
2.  It maintains consistency, since we're not using `fget.short_description` anywhere else in the codebase.

AfaqShuaib09: I don't have any strong opinions on this, but I like the current approach. Actually, I got your point,. we should define it in admin.py since we're using it only for the admin.

Moving this property as a method in admin.py, as you said

hamza-56 (approved):

PR #4609, opened 2025-04-11
feat: add CourseRunLoader for bulk reruns

Ali-D-Akbar:
[PROD-4346](https://2u-internal.atlassian.net/browse/PROD-4346)
Adds the ability to bulk-create reruns given a CSV. It has the ability to copy an existing run given the active course run key using the course run API and push the run to legal review.

## Testing Instructions
1. Download the following CSV, make some changes to it so that it has the `Last Active Run Key` of one of your locally created course run
[Bulk Reruns.csv](https://github.com/user-attachments/files/19932544/Bulk.Reruns.csv)
[Bulk Reruns move to legal.csv](https://github.com/user-attachments/files/19939340/Bulk.Reruns.move.to.legal.csv)

2. Run and play around the following script to run the ingestion and check for edge cases

```
from course_discovery.apps.core.models import Partner
from course_discovery.apps.course_metadata.data_loaders.course_run_loader import CourseRunDataLoader

csv_path = "Bulk Reruns.csv"
partner = Partner.objects.get()
loader = CourseRunDataLoader(partner, csv_path=csv_path)
loader.ingest()
```

3. You should be able to create reruns, send them for legal review, and fail in case of a missing field or invalid course run key.


course_discovery/apps/course_metadata/data_loaders/utils.py
```diff
@@ -139,3 +148,145 @@ def map_external_org_code_to_internal_org_code(external_org_code, product_source
             f'product_source {product_source}'
         )
         return external_org_code
+
+
+def initialize_csv_reader(csv_path, csv_file, use_gspread_client, product_type=None, product_source=None):
```
DawoudSheraz: All of the added utils are specific to CSV loaders. They should be converted to Mixin and then added to new and existing csv loader.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,112 @@
...
+
+            course_run_key = row.get('last_active_run_key')
+            start_date = row.get('start_date')
+            end_date = row.get('end_date')
+            start_time = row.get('start_time', '00:00:00')
+            end_time = row.get('end_time', '00:00:00')
+            run_type = row.get('run_type')
+            pacing_type = row.get('pacing_type')
+
+            if not course_run_key or not start_date or not end_date or not run_type or not pacing_type:
```
DawoudSheraz: not all (<fields>) will be readable. The row validation can be moved to an internal or separate method, to make the ingest method easier to follow along.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,112 @@
...
+                error_message = f"Missing required field(s) in row: {row}"
```
DawoudSheraz: The error message should have the missing fields only alongside the row number. Adding the complete row will make it hard to read or comprehend.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+    def ingest(self):
+        logger.info("Starting ingestion of course run CSV data.")
+        import pdb
+        pdb.set_trace()
+        for row in self.reader:
+            row = transform_dict_keys(row)
+
+            course_run_key = row.get('last_active_run_key')
```
DawoudSheraz: The variable should be last_active_run_key. course_run_key can be confused with a new run key.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,112 @@
...
+                error_message = f"Missing required field(s) in row: {row}"
+                logger.error(error_message)
+                self.ingestion_summary['failure_count'] += 1
+                self.ingestion_summary['errors'].append(error_message)
+                continue
+
+            try:
+                course_run = CourseRun.objects.filter(key=course_run_key).first()
+                if not course_run:
+                    error_message = f"CourseRun with key '{course_run_key}' not found. Skipping row."
```
DawoudSheraz: The error message should clarify that existing active run key course run was not found.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,112 @@
...
+                course_run_type_uuid = get_course_run_type(run_type).uuid
+
+                data = {
+                    'prices': extract_seat_prices(course_run),
+                    'start': start_datetime,
+                    'end': end_datetime,
+                    'pacing_type': pacing_type,
+                }
+
+                _ = create_course_run(
```
DawoudSheraz: wouldn't it be useful to have new course run key added in success results? Only count might not be good enough.

Ali-D-Akbar: Yes. I'll add this too.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        super().__init__(partner=partner, api_url=api_url, max_workers=max_workers, is_threadsafe=is_threadsafe)
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.reader = self.initialize_csv_reader(csv_path, csv_file, use_gspread_client)
+        self.ingestion_summary = {
+            'total_runs_count': len(self.reader),
+            'success_count': 0,
+            'failure_count': 0,
+            'errors': [],
+            'newly_created_runs': [],
```
DawoudSheraz: nit: new_runs should suffice

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+
+        for row in self.reader:
+            row = self.transform_dict_keys(row)
+            last_active_run_key = row.get('last_active_run_key')
+            missing_fields = self.validate_course_data(row)
+
+            if missing_fields:
+                self.log_ingestion_error(
+                    CSVIngestionErrors.MISSING_REQUIRED_DATA,
+                    f"Missing required field(s) in course run: {last_active_run_key}"
```
DawoudSheraz: The missing fields should be logged with last_active_run_key.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+            'newly_created_runs': [],
+        }
+
+    def ingest(self):
+        """
+        Perform the ingestion process for each CSV row.
+        """
+        logger.info("Starting ingestion of course run loader.")
+
+        for row in self.reader:
```
DawoudSheraz: nit: let's use enumerate so that index is available in the code. The index can be used in the logs that mention `row`.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+
+                logger.info(f"Successfully created rerun {new_course_run.key} for course: {course.title}")
+                self.ingestion_summary['success_count'] += 1
+                self.ingestion_summary['newly_created_runs'].append(new_course_run.key)
+
+                if row.get('move_to_legal_review', '').lower() == 'true':
+                    new_course_run.status = CourseRunStatus.LegalReview
+                    new_course_run.save(update_fields=['status'], send_emails=True)
+
+            except Exception as e:  # pylint: disable=broad-except
```
DawoudSheraz: Let's not use a broad exception here. Capture expected exceptions to keep the code concise.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -65,23 +89,24 @@ def extract_seat_prices(course_run):
             prices[seat.type.slug] = f"{seat.price:.2f}"
         return prices
 
-    def create_course_run(self, data, course, course_type, course_run_type_uuid, rerun=None):
+    def create_course_run(self, data, course, course_run_type_uuid, course_type=None, rerun=None):
```
DawoudSheraz, line 116: What's the reason for changing the argument ordering?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -25,6 +29,26 @@ def __init__(self, **kwargs):
...
+            else:
+                # read the file from the provided path; otherwise, use the file received from CSVDataLoaderConfiguration
+                return list(csv.DictReader(open(csv_path, 'r'))) if csv_path else list(unicodecsv.DictReader(csv_file))
+        except FileNotFoundError:
+            logger.exception(f"Error opening CSV file at path: {csv_path}")
+            raise
+        except Exception as e:
+            logger.exception(f"Error reading input data source: {e}")
+            raise
+
```
DawoudSheraz: I dont think this exact code should be in Mixin. For instance, settings.PRODUCT_METADATA_MAPPING[product_type][product_source.slug] is very specific to CSVLoader. In the Mixin method, it is possible to have a code like below but other than that, this particular instance should remain in CSVLoader.

```
return list(csv.DictReader(open(csv_path, 'r'))) if csv_path else list(unicodecsv.DictReader(csv_file))
```

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -65,23 +89,24 @@ def extract_seat_prices(course_run):
...
 
-    def create_course_run_api_request_data(self, data, course, course_type, course_run_type_uuid, rerun=None):
+    def create_course_run_api_request_data(self, data, course, course_run_type_uuid, course_type=None, rerun=None):
         """
         Given a data dictionary, return a reduced data representation in dict
         which will be used as input for course run creation via course run api.
         """
-        pricing = self.get_pricing_representation(data['verified_price'], course_type)
+        pricing = self.get_pricing_representation(data['verified_price'], course_type) if not data['prices'] \
+            else data['prices']
```
DawoudSheraz: What's the reason for this change? Should pricing be a matter in a re-run?

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
```diff
@@ -0,0 +1,188 @@
...
+        header_line = ','.join(key.replace('_', ' ').title() for key in headers) + '\n'
+        csv.write(header_line.encode())
+        for line_dict in lines_dict_list:
+            row = ','.join(f'"{line_dict.get(key, "")}"' for key in headers) + '\n'
+            csv.write(row.encode())
+        csv.seek(0)
+        return csv
+
+    @responses.activate
+    def test_ingest_successful_course_run_creation(self, _):
```
DawoudSheraz: instead of underscore, you can mention the exact argument and then disable any lint violation (to be consistent with how other loaders do it).

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
```diff
@@ -0,0 +1,188 @@
...
+        with NamedTemporaryFile() as csv:
+            csv = self._write_csv(csv, [VALID_COURSE_RERUN_DATA])
+            with LogCapture(LOGGER_PATH) as log_capture:
+                with mock.patch.object(CourseRunDataLoader, 'call_course_api', self.mock_call_course_api):
+                    loader = CourseRunDataLoader(self.partner, csv_path=csv.name)
+                    summary, _ = loader.ingest()
+
+                log_capture.check_present(
+                    (LOGGER_PATH, 'INFO',
+                     f"Successfully created rerun {summary['newly_created_runs'][0]} for course: {self.course.title}"),
```
DawoudSheraz: nit: Instead of getting the stats from the summary, it would be ok if the explicit values are added in the tests. That sets the correct expectations within the test.

course_discovery/apps/course_metadata/data_loaders/mixins.py
Ali-D-Akbar: Some of the args are now optional. Moved the optional one in the end.

course_discovery/apps/course_metadata/data_loaders/mixins.py
Ali-D-Akbar: We basically send prices to course run API. Without that, it won't create the rerun so we fetch the prices from source run and add it in the request data.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
Ali-D-Akbar: Hmmm, I felt like it wasn't really used so I added underscore. Not sure which option sounds more appropriate.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+        csv_path=None,
+        csv_file=None,
+        use_gspread_client=False
+    ):
+        """
+        Initialize the loader with CSV input and ingestion tracking.
+        """
+        super().__init__(partner=partner, api_url=api_url, max_workers=max_workers, is_threadsafe=is_threadsafe)
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.reader = self.initialize_csv_reader(csv_path, csv_file, use_gspread_client)
```
AfaqShuaib09: gspread_client won't be required here

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+            'failure_count': 0,
+            'errors': [],
+            'newly_created_runs': [],
+        }
+
+    def ingest(self):
+        """
+        Perform the ingestion process for each CSV row.
+        """
+        logger.info("Starting ingestion of course run loader.")
```
AfaqShuaib09, line 54: https://github.com/openedx/course-discovery/pull/4613/files#diff-52f360886d9cda9c7b8180a2d9c96707f9e9a754f4a165b25760afe579963d68R41-R43

The CourseRun loader will also be handling CourseRun updates operation. Should we define separate methods for them here and then call them in the ingest method, like we did in the Course loader?

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+
+        for row in self.reader:
+            row = self.transform_dict_keys(row)
+            last_active_run_key = row.get('last_active_run_key')
```
AfaqShuaib09, line 58: Can we fetch it through `row['last_active_run_key']`. Since it's a required field.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
AfaqShuaib09: The message doesn't seem appropriate. The last run has already been created. We're creating a rerun using the last active run key.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
AfaqShuaib09: Or we can add logic to fetch the latest active course run and use its key as the last active key if `last_active_run_key` is absent in the sheet.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,150 @@
...
+                self.log_ingestion_error(
+                    CSVIngestionErrors.COURSE_RUN_CREATE_ERROR,
+                    f"Error creating rerun for course '{course.title}': {str(e)}"
+                )
+
+        logger.info("Course run ingestion complete.")
+        logger.info(f"Ingestion Summary: {self.ingestion_summary}")
+        self.render_error_logs(self.error_logs)
+
+        return self.ingestion_summary, self.error_logs
```
AfaqShuaib09:
```python
{
'summary': self.ingestion_summary,
'errors': self.error_logs,
}

File: course_discovery/apps/course_metadata/data_loaders/mixins.py (hunk 12 above)
Comment (Ali-D-Akbar) [reply course_discovery/apps/course_metadata/data_loaders/mixins.py]: As discussed, I'll be keeping this in the mixin for now since it's usage is tweaked a little to support both CSVLoader and the new bulk loaders.

File: course_discovery/apps/course_metadata/data_loaders/course_run_loader.py (hunk 17 above)
Comment (Ali-D-Akbar) [reply course_discovery/apps/course_metadata/data_loaders/course_run_loader.py:54]: You can see the loader definition here: https://github.com/openedx/course-discovery/pull/4613/files#diff-52f360886d9cda9c7b8180a2d9c96707f9e9a754f4a165b25760afe579963d68R56-R60

How the loaders will be responsible for doing stuff.. I think the course run update is more related to CourseLoader.

File: course_discovery/apps/course_metadata/data_loaders/course_run_loader.py (hunk 18 above)
Comment (Ali-D-Akbar) [reply course_discovery/apps/course_metadata/data_loaders/course_run_loader.py:58]: If this field is absent, we don't have any other identifier. So having `last_active_run_key` is a must.
Also, I'm already checking it's presence in `validate_course_data`.

File: course_discovery/apps/course_metadata/data_loaders/course_run_loader.py (hunk 8 above)
Comment (Ali-D-Akbar) [reply course_discovery/apps/course_metadata/data_loaders/course_run_loader.py]: I've added the row number now. It should be better.

File: course_discovery/apps/course_metadata/data_loaders/course_run_loader.py (hunk 18 above)
Comment (AfaqShuaib09) [reply course_discovery/apps/course_metadata/data_loaders/course_run_loader.py:58]: Then i think we should convert it to `row['last_active_run_key']`

File: course_discovery/apps/course_metadata/data_loaders/course_run_loader.py [hunk 20]
```diff
@@ -0,0 +1,153 @@
...
+                    rerun=last_active_run_key
+                )
+
+                if not isinstance(result, dict) or 'key' not in result:
+                    raise ValueError(f"Invalid response from create_course_run: {result}")
+
+                new_course_run = CourseRun.everything.filter(key=result['key']).first()
+
+                if not new_course_run:
+                    raise LookupError(f"No CourseRun found in DB for key: {result['key']}")
```
DawoudSheraz, line 103: Why is the exception raised here and not logged like other cases? This will break the execution flow.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,153 @@
...
+        Track ingestion failure in the summary and error logs.
+        """
+        self.ingestion_summary['failure_count'] += 1
+        self.error_logs[error_key].append(error_message)
+
+    def update_course_api_request_data(self, course_data, course, is_draft):
+        pass
+
+    def update_course_run_api_request_data(self, course_run_data, course_run, course_type, is_draft):
+        pass
```
DawoudSheraz: Is it because these methods are abstract in Mixin? If so, what if the mixin methods are updated to return pass instead of abstract method?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -51,6 +54,26 @@ def __init__(self, **kwargs):
...
+        """
+        Initialize the CSV reader based on the input source (csv_path, csv_file or gspread_client)
+        """
+        try:
+            if use_gspread_client:
+                product_type_config = settings.PRODUCT_METADATA_MAPPING[product_type][product_source.slug]
+                gspread_client = GspreadClient()
+                return list(gspread_client.read_data(product_type_config))
+            else:
+                # read the file from the provided path; otherwise, use the file received from CSVDataLoaderConfiguration
```
DawoudSheraz: this comment is no longer correct in the mixin context.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
```diff
@@ -0,0 +1,191 @@
...
+        responses.add(responses.POST, f'{studio_url}{self.existing_run.key}/rerun/', status=200)
+
+        with NamedTemporaryFile() as csv:
+            csv = self._write_csv(csv, [VALID_COURSE_RERUN_DATA])
+            with LogCapture(LOGGER_PATH) as log_capture:
+                with mock.patch.object(CourseRunDataLoader, 'call_course_api', self.mock_call_course_api):
+                    loader = CourseRunDataLoader(self.partner, csv_path=csv.name)
+                    summary = loader.ingest()['summary']
+
+                # Dynamically capture the last_active_run_key and course key
```
DawoudSheraz: this comment is not needed, the code is self-explanatory.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
```diff
@@ -0,0 +1,191 @@
...
+                    loader = CourseRunDataLoader(self.partner, csv_path=csv.name)
+                    summary = loader.ingest()['summary']
+
+                error_logs = [
+                    entry for entry in log_capture.actual()
+                    if entry[0] == LOGGER_PATH and entry[1] == 'ERROR' and entry[2].startswith(
+                        f"[Row 1] Error creating rerun for course '{self.course.title}':"
+                    )
+                ]
+                self.assertTrue(error_logs, "Expected ERROR log about rerun creation failure not found.")
```
DawoudSheraz, line 164: Why can't check_present be used here?

Ali-D-Akbar: I'm just partially checking the logs which were not supported in `check_present` due to these being the dynamic logs. I was having a problem using `check_present` but I don't remember now 😬

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
Ali-D-Akbar: I feel like I shouldn't have made them abstract. Removing the abstract part.

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        super().__init__(partner=partner, api_url=api_url, max_workers=max_workers, is_threadsafe=is_threadsafe)
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.reader = self.initialize_csv_reader(csv_path, csv_file)
+        self.ingestion_summary = {
+            'total_runs_count': len(self.reader),
+            'success_count': 0,
+            'failure_count': 0,
+            'errors': [],
+            'new_runs': [],
```
AfaqShuaib09, line 47: nit: reruns?

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
```diff
@@ -0,0 +1,147 @@
...
+            course_run_type_uuid = self.get_course_run_type(row.get('run_type')).uuid
+
+            data = {
+                'prices': self.extract_seat_prices(course_run),
+                'start_date': row.get('start_date'),
+                'start_time': row.get('start_time', '00:00:00'),
+                'end_date': row.get('end_date'),
+                'end_time': row.get('end_time', '00:00:00'),
+                'course_pacing': row.get('pacing_type'),
+            }
```
AfaqShuaib09, line 87: nit: can we wrap this logic in a separate method i.e; preparing_payload

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -251,13 +274,13 @@ def update_course_run(self, data, course_run, course_type, is_draft):
             logger.info(f"Course run update response: {response.content}")
         return response.json()
 
-    @abstractmethod
-    def update_course_api_request_data(self, course_data, course, is_draft):
+    def update_course_api_request_data(self, course_data, course, is_draft):  # pylint: disable=unused-argument
         """Update the course API request data based on the course and draft state."""
+        return course_data
```
AfaqShuaib09, line 279: why are we returning here `course_data`?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -251,13 +274,13 @@ def update_course_run(self, data, course_run, course_type, is_draft):
...
 
-    @abstractmethod
-    def update_course_run_api_request_data(self, course_run_data, course_run, course_type, is_draft):
+    def update_course_run_api_request_data(self, course_run_data, course_run, course_type, is_draft):  # pylint: disable=unused-argument
         """Update the course run API request data based on the run, type, and draft state."""
+        return course_run_data
```
AfaqShuaib09, line 283: Same question here?

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
```diff
@@ -0,0 +1,190 @@
...
+            status='published',
+            draft=True,
+        )
+        SeatFactory(course_run=self.existing_run, type=SeatTypeFactory.verified(), price=200.0)
+
+    def mock_call_course_api(self, method, url, payload):
+        """Mock POST requests to the course API using Django test client."""
+        if method == 'POST':
+            return self.client.post(url, data=payload, format='json')
+        return None
```
AfaqShuaib09:
```python
return self.client.post(url, data=payload, format='json') if method == 'POST' else None
```

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
Ali-D-Akbar: I think new_runs made more sense to me. 🫡

course_discovery/apps/course_metadata/data_loaders/course_run_loader.py
Ali-D-Akbar: I can but I don't know if moving this to a separate function really helps with either of the readability functionality since this isn't reused.

course_discovery/apps/course_metadata/data_loaders/mixins.py
Ali-D-Akbar: Default dummy data. Otherwise, Pylint expects a return for this if nothing is being returned.

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/data_loaders/mixins.py
DawoudSheraz: I think you can disable pylint there and instead of returning the data, simply pass or raise assertion that child class needs to implement this method.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_run_loader.py
```diff
@@ -0,0 +1,188 @@
...
+
+        with NamedTemporaryFile() as csv:
+            csv = self._write_csv(csv, [VALID_COURSE_RERUN_DATA])
+            with LogCapture(LOGGER_PATH) as log_capture:
+                with mock.patch.object(CourseRunDataLoader, 'call_course_api', self.mock_call_course_api):
+                    loader = CourseRunDataLoader(self.partner, csv_path=csv.name)
+                    summary = loader.ingest()['summary']
+
+                expected_log_msg = (f'[Row 1] Successfully created rerun course-v1:edx+csv_123+1T2025 '
+                                    f'for course: {self.course.title}')
```
DawoudSheraz, line 90: also verify that course run has been created in DB.

DawoudSheraz (approved):
please add the loader handling in bulk operation queue task so that it can be tested on stage after this PR merge.

course_discovery/apps/course_metadata/tasks.py
```diff
@@ -59,15 +60,19 @@ def select_and_init_bulk_operation_loader(bulk_operation_task):
     """
     Identifies and instantiates the appropriate data loader for a given BulkOperationTask.
     """
+    partner = Partner.objects.get(short_code='edx')
```
DawoudSheraz: nit: instead of hard-coding short code, please use DEFAULT_PARTNER_ID settings.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4391,33 +4392,32 @@ def test_create_queues_task(self):
...
-        with mock.patch('course_discovery.apps.course_metadata.tasks.CourseLoader') as mock_loader:
-            mock_loader.return_value.ingest.side_effect = KeyError("Can not find short_description")
             PartnerFactory(short_code='edx')
-            bulk_operation_task = BulkOperationTaskFactory(task_type=BulkOperationType.CourseCreate)
+            bulk_operation_task = BulkOperationTaskFactory(task_type=task_type)
             bulk_operation_task.refresh_from_db()
 
-            self.assertEqual(bulk_operation_task.status, BulkOperationStatus.Failed)
-            self.assertEqual(bulk_operation_task.task_summary, None)
+            expected_status = BulkOperationStatus.Completed if should_succeed else BulkOperationStatus.Failed
```
DawoudSheraz: nit: this can be a ddt argument too.

DawoudSheraz (approved):

course_discovery/apps/course_metadata/data_loaders/mixins.py
Ali-D-Akbar: In case if we wanna must have an implementation of a method, we can make it abstract. In this case however, it's not always like we'll need this to be implemented as CoursRunDataLoader doesn't require any updates. I'm just returning the data being passed to it as a sign to not having any implementation since update_course or update_course_run expects them to be implemented.

Disabling it is one way to fix it but returning like this looked a little cleaner to me.

PR #4611, opened 2025-04-15
feat: data_modified_timestamp updates for related fields of Programs and Degrees

zawan-ila:
### [PROD-3295](https://2u-internal.atlassian.net/browse/PROD-3295)

This PR adds the relevant code changes for updating the `data_modified_timestamp` field for Programs and Degrees when related fields change.

#### Testing instructions 
1. Check out this branch
2. Open a program or degree page in django admin. Make any edits. Verify that the `data_modified_timestamp` field is updated appropriately.
3. From publisher, make edits to a course. Verify that the `data_modified_timestamp` field on the course's associated programs is updated appropriately.

### DIFF
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -79,6 +79,7 @@ class TestProgramViewSet(SerializationMixin):
             type=program_type,
         )
         program.labels.add(topic)
+        program.refresh_from_db()
         return program
 
     def create_curriculum(self, parent_program):
--- a/course_discovery/apps/api/v1/tests/test_views/test_programs.py
+++ b/course_discovery/apps/api/v1/tests/test_views/test_programs.py
@@ -134,7 +134,7 @@ class TestProgramViewSet(SerializationMixin):
     def test_retrieve_basic_curriculum(self, django_assert_num_queries):
         program = self.create_program(courses=[])
         self.create_curriculum(program)
-
+        program.refresh_from_db()
         with django_assert_num_queries(FuzzyInt(52, 3)):
             response = self.assert_retrieve_success(program)
         assert response.data == self.serialize_program(program)
@@ -153,7 +153,7 @@ class TestProgramViewSet(SerializationMixin):
             program=child_program2,
             curriculum=curriculum
         )
-
+        parent_program.refresh_from_db()
         with django_assert_num_queries(FuzzyInt(85, 3)):
             response = self.assert_retrieve_success(parent_program)
         assert response.data == self.serialize_program(parent_program)
--- a/course_discovery/apps/course_metadata/signals.py
+++ b/course_discovery/apps/course_metadata/signals.py
@@ -1,8 +1,8 @@
 import inspect
-
 import logging
 import time
 from datetime import datetime, timezone
+
 import pytz
 import waffle  # lint-amnesty, pylint: disable=invalid-django-waffle-import
 from celery import shared_task
@@ -22,16 +22,18 @@ from course_discovery.apps.course_metadata.constants import MASTERS_PROGRAM_TYPE
 from course_discovery.apps.course_metadata.data_loaders.api import CoursesApiDataLoader
 from course_discovery.apps.course_metadata.models import (
     AdditionalMetadata, CertificateInfo, Course, CourseEditor, CourseEntitlement, CourseLocationRestriction, CourseRun,
-    Curriculum, CurriculumCourseMembership, CurriculumProgramMembership, Fact, GeoLocation, Organization, ProductMeta,
-    ProductValue, Program, ProgramLocationRestriction, Seat, TaxiForm, DegreeDeadline, DegreeCost, DegreeAdditionalMetadata,
-    IconTextPairing, Degree, Ranking, Specialization
+    Curriculum, CurriculumCourseMembership, CurriculumProgramMembership, Degree, DegreeAdditionalMetadata, DegreeCost,
+    DegreeDeadline, Fact, GeoLocation, IconTextPairing, Organization, ProductMeta, ProductValue, Program,
+    ProgramLocationRestriction, Ranking, Seat, Specialization, TaxiForm
 )
 from course_discovery.apps.course_metadata.publishers import ProgramMarketingSitePublisher
 from course_discovery.apps.course_metadata.salesforce import (
     populate_official_with_existing_draft, requires_salesforce_update
 )
 from course_discovery.apps.course_metadata.tasks import update_org_program_and_courses_ent_sub_inclusion
-from course_discovery.apps.course_metadata.utils import data_modified_timestamp_update, get_salesforce_util, data_modified_timestamp_update__deletion
+from course_discovery.apps.course_metadata.utils import (
+    data_modified_timestamp_update, data_modified_timestamp_update__deletion, get_salesforce_util
+)
 
 logger = logging.getLogger(__name__)
 User = get_user_model()
@@ -450,6 +452,7 @@ def course_topics_taggable_changed(sender, instance, action, **kwargs):
         logger.info(f"{sender} has been updated for Course {instance.key}.")
         instance.set_data_modified_timestamp()
 
+
 @receiver(m2m_changed, sender=Program.labels.through)
 def program_labels_changed(sender, instance, action, **kwargs):
     """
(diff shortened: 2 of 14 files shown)


course_discovery/apps/course_metadata/signals.py
```diff
@@ -492,6 +506,61 @@ def course_run_staff_changed(sender, instance, action, **kwargs):  # pylint: dis
...
+
+    if action == 'pre_clear' and not kwargs['reverse']:
+        # when this signal is called from the m2m field's set(), the 5th frame from the bottom
+        # corresponds to the frame of the set method. the objs param of that method contains
+        # the values being set()
+        objs = inspect.stack()[4][0].f_locals['objs']
+        if (len(objs) and isinstance(objs[0], int)):
+            to_set = objs.copy()
+        else:
+            to_set = [obj.id for obj in objs]
```
DawoudSheraz: What are the conditions that list objects could be either object or int?

course_discovery/apps/course_metadata/signals.py
```diff
@@ -447,6 +450,17 @@ def course_topics_taggable_changed(sender, instance, action, **kwargs):
...
         instance.set_data_modified_timestamp()
 
+@receiver(m2m_changed, sender=Program.labels.through)
+def program_labels_changed(sender, instance, action, **kwargs):
+    """
+    Signal handler to handle Taggable manager changes for the program labels field.
+    """
+    if action in ['pre_add', 'pre_remove'] and not kwargs['reverse'] \
+            and kwargs['pk_set'] and instance._meta.label in ['course_metadata.Program', 'course_metadata.Degree']:
+        instance._meta.model.objects.filter(pk=instance.pk).update(
```
DawoudSheraz, line 464: Let's add the log statement to be consistent with how other methods are doing.

course_discovery/apps/course_metadata/signals.py
```diff
@@ -509,9 +578,27 @@ def connect_course_data_modified_timestamp_related_models():
...
+        Ranking,
+        Specialization
     ]:
         pre_save.connect(data_modified_timestamp_update, sender=model)
 
+    for model in [
+        Curriculum,
+        CurriculumProgramMembership,
+        CurriculumCourseMembership,
+        ProgramLocationRestriction
```
DawoudSheraz, line 599: Are there other models related to Course/Runs that need to be in there?

course_discovery/apps/course_metadata/utils.py
```diff
@@ -1022,9 +1022,18 @@ def data_modified_timestamp_update(sender, instance, **kwargs):  # pylint: disab
      * A method called update_product_data_modified_stamp which will be implemented by each model
      depending upon its relation with Course/Program.
     """
-    if hasattr(instance, 'field_tracker') and hasattr(instance, 'update_product_data_modified_timestamp'):
+    if hasattr(instance, 'field_tracker') and hasattr(instance, 'update_product_data_modified_timestamp'):
```
DawoudSheraz: nit: it seems the tab/space was added at the end of the if condition.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -656,11 +673,14 @@ def test_set_active_url_slug__official_version(self):
         """
         draft_course = CourseFactory(draft=True, title="Test course")
         non_draft_course = CourseFactory(draft_version=draft_course, title=draft_course.title, key=draft_course.key)
+        program = ProgramFactory(courses=[non_draft_course])
+        program.refresh_from_db()
```
DawoudSheraz: huh, why is there a need to refresh the object from DB right after the creation?

course_discovery/apps/course_metadata/tests/test_signals.py
```diff
@@ -1031,7 +1033,7 @@ class DataModifiedTimestampUpdateSignalsTests(TestCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        disconnect_course_data_modified_timestamp_signal_handlers()
+        # disconnect_course_data_modified_timestamp_signal_handlers()
```
DawoudSheraz: Why was this commented out?

course_discovery/apps/api/v1/tests/test_views/test_programs.py
```diff
@@ -133,7 +134,7 @@ def test_retrieve(self, django_assert_num_queries):
     def test_retrieve_basic_curriculum(self, django_assert_num_queries):
         program = self.create_program(courses=[])
         self.create_curriculum(program)
-
+        program.refresh_from_db()
```
DawoudSheraz, line 137: why is this refresh needed?

course_discovery/apps/course_metadata/signals.py
zawan-ila: Can you please be more elaborate?

course_discovery/apps/course_metadata/signals.py
zawan-ila: Done

course_discovery/apps/course_metadata/signals.py
zawan-ila: I am not sure. These are the ones where I felt deletions may be happening, and verified that from the history models.

zawan-ila: There may be such fields on Course/CourseRun/Program/Degree level.

course_discovery/apps/course_metadata/tests/test_models.py
zawan-ila: I was worried that setting some related fields, for example courses here, or some other default chosen by the factory, may cause timestamp update on the object. Just checking a couple tests shows that not to be the case.

I have removed these refreshes. To be safe though, I have added a post gen hook that achieves the same effect.

course_discovery/apps/api/v1/tests/test_views/test_programs.py
zawan-ila: The `self.create_curriculum` call updates the timestamp. If we do not refresh, the timestamp in the serialized api response and the program object do not match, causing a failure

course_discovery/apps/course_metadata/tests/test_signals.py
zawan-ila: Earlier, the tests would individually connect the signals they "needed". Either we'd need to do the same for the tests added in this PR, or just remove this "disconnect" and have all the signals be "connected". I am not sure which approach is better.

course_discovery/apps/course_metadata/signals.py
DawoudSheraz: I am asking for more context on this if-else condition. What are the conditions that objs will be list of objects (from whom the id is read) ? And under what conditions is it list of Int (ids)?

course_discovery/apps/course_metadata/tests/test_signals.py
DawoudSheraz: If I remember, the disconnect was done to test the signal handler in isolation. In other tests, all the signal handlers are connected. But for signal tests, disconnecting everything and only connecting the handlers under test provides a better level of isolation.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -349,10 +355,14 @@ def test_data_modified_timestamp_model_related_field_change(self):
             draft=True,
             additional_metadata=AdditionalMetadataFactory(external_identifier='identifier_1')
         )
+        prog = factories.ProgramFactory(courses=[course])
```
DawoudSheraz: nit: program is used in other tests, using prog in this is out of place.

course_discovery/apps/course_metadata/signals.py
zawan-ila: This is essentially what is being passed to `m2m_field.set()`. The list passed to the m2m field set method may contain both objects and ids. Earlier, I had only handled the objects case, but there was some test passing in raw ids.

course_discovery/apps/course_metadata/tests/test_signals.py
zawan-ila: Uncommented it.

course_discovery/apps/course_metadata/tests/test_models.py
zawan-ila: Renamed to `program`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4096,6 +4205,19 @@ def test_specializations_in_degree(self):
         self.assertEqual(degree.specializations.count(), 1)
         self.assertEqual(degree.specializations.first().value, specialization.value)
 
+    def test_degree_timestamp_update_simple_changes(self):
```
Ali-D-Akbar, line 4217: nit: docstring

Ali-D-Akbar (approved):
Some docstrings explaining the new unit tests and utils can help understand them better.

Looks good otherwise. Great work 👏

hamza-56 (approved):

PR #4612, opened 2025-04-16
refactor: move common utils in DataLoaderMixin

Ali-D-Akbar:
[PROD-4353](https://2u-internal.atlassian.net/browse/PROD-4353)
Moves most of the commonly used utils to a mixin so it is easier to use among other data loaders.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/__init__.py
+++ b/course_discovery/apps/course_metadata/data_loaders/__init__.py
@@ -18,7 +18,7 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
     LOADER_MAX_RETRY = 3
     PAGE_SIZE = 50
 
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, enable_api=True):
+    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, enable_api=True, **kwargs):
         """
         Arguments:
             partner (Partner): Partner which owns the APIs and data being loaded
@@ -28,6 +28,7 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
             enable_api (bool): True if we want to use the api functionalities and clients with the dataloader.
                 This will most likely only be turned off for event bus use cases.
         """
+
         self.partner = partner
         self.enable_api = enable_api
         if self.enable_api:
--- a/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/csv_loader.py
@@ -11,7 +11,6 @@ from django.conf import settings
 from django.db.models import Q
 from django.urls import reverse
 
-from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.choices import (
     CourseRunRestrictionType, CourseRunStatus, ExternalCourseMarketingType, ExternalProductStatus
 )
@@ -19,10 +18,10 @@ from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoade
 from course_discovery.apps.course_metadata.data_loaders.constants import (
     CSV_LOADER_ERROR_LOG_SEQUENCE, CSVIngestionErrorMessages, CSVIngestionErrors
 )
+from course_discovery.apps.course_metadata.data_loaders.mixins import DataLoaderMixin
 from course_discovery.apps.course_metadata.gspread_client import GspreadClient
 from course_discovery.apps.course_metadata.models import (
-    AdditionalMetadata, Collaborator, Course, CourseRun, CourseRunPacing, CourseRunType, CourseType, Organization,
-    Person, ProgramType, Source, Subject
+    AdditionalMetadata, Collaborator, Course, CourseRun, CourseType, Organization, Person, ProgramType, Source, Subject
 )
 from course_discovery.apps.course_metadata.utils import download_and_save_course_image
 from course_discovery.apps.ietf_language_tags.models import LanguageTag
@@ -30,7 +29,7 @@ from course_discovery.apps.ietf_language_tags.models import LanguageTag
 logger = logging.getLogger(__name__)
 
 
-class CSVDataLoader(AbstractDataLoader):
+class CSVDataLoader(AbstractDataLoader, DataLoaderMixin):
 
     PROGRAM_TYPES = [
         ProgramType.XSERIES,
@@ -75,7 +74,13 @@ class CSVDataLoader(AbstractDataLoader):
             * product_type: course type slug to identify the product type present in CSV
             * product_source: slug of the external source that actually owns the product.
         """
-        super().__init__(partner, api_url, max_workers, is_threadsafe)
+        super().__init__(
+            partner=partner,
+            api_url=api_url,
+            max_workers=max_workers,
+            is_threadsafe=is_threadsafe
+        )
+
         self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
         self.ingestion_summary = self._initialize_ingestion_summary()
         self.course_uuids = {}  # to show the discovery course ids for each processed course
@@ -95,16 +100,6 @@ class CSVDataLoader(AbstractDataLoader):
             'archived_products': []
         }
 
-    def _get_product_source(self, product_source):
-        """
-        Retrieve the product source or raise an exception if product source doesn't exist already
-        """
-        try:
-            return Source.objects.get(slug=product_source)
-        except Source.DoesNotExist:
-            logger.exception(f"Unable to locate source with slug '{product_source}'")
-            raise
-
     def _initialize_csv_reader(self, csv_path, csv_file, use_gspread_client):
         """
         Initialize the CSV reader based on the input source (csv_path, csv_file or gspread_client)
@@ -122,6 +117,16 @@ class CSVDataLoader(AbstractDataLoader):
             raise
         except Exception as e:
             logger.exception(f"Error reading input data source: {e}")
+            raise
+
+    def _get_product_source(self, product_source):
+        """
+        Retrieve the product source or raise an exception if product source doesn't exist already
+        """
+        try:
+            return Source.objects.get(slug=product_source)
+        except Source.DoesNotExist:
+            logger.exception(f"Unable to locate source with slug '{product_source}'")
             raise
 
     def ingest(self):  # pylint: disable=too-many-statements
@@ -162,7 +167,7 @@ class CSVDataLoader(AbstractDataLoader):
             else:
                 logger.info(f"Course key {course_key} could not be found in database, creating the course.")
                 try:
-                    _ = self._create_course(row, course_type, course_run_type.uuid)
+                    _ = self.create_course(row, course_type, course_run_type.uuid, self.product_source)
                 except Exception as exc:  # pylint: disable=broad-except
                     exception_message = exc
                     if hasattr(exc, 'response'):
@@ -196,10 +201,10 @@ class CSVDataLoader(AbstractDataLoader):
                     )
                     continue
                 if not is_course_created:
-                    self.add_product_source(course)
+                    self.add_product_source(course, self.product_source)
 
                 try:
-                    self._update_course(row, course, is_draft)
+                    self.update_course(row, course, is_draft)
                 except Exception as exc:  # pylint: disable=broad-except
                     exception_message = exc
                     if hasattr(exc, 'response'):
@@ -247,7 +252,7 @@ class CSVDataLoader(AbstractDataLoader):
             # No need to update the course run if the run is already in the review
             if not course_run.in_review:
                 try:
-                    self._update_course_run(row, course_run, course_type, is_draft)
+                    self.update_course_run(row, course_run, course_type, is_draft)
                 except Exception as exc:  # pylint: disable=broad-except
                     exception_message = exc
                     if hasattr(exc, 'response'):
@@ -356,20 +361,6 @@ class CSVDataLoader(AbstractDataLoader):
         try:
             return CourseType.objects.get(name=course_type_name)
         except CourseType.DoesNotExist:
-            return None
-
-    @staticmethod
-    @cache
-    def get_course_run_type(course_run_type_name):
-        """
-        Retrieve a CourseRunType object, using a cache to avoid redundant queries.
-
-        Args:
-            course_run_type_name (str): Course run type name
-        """
-        try:
-            return CourseRunType.objects.get(name=course_run_type_name)
-        except CourseRunType.DoesNotExist:
             return None
 
     def _validate_and_process_row(self, row, course_title, org_key):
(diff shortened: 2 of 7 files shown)


AfaqShuaib09:
- _validate_and_process_row
- get_course_key
- create_course
- get_draft_flag
- _update_course (some dependent methods will be overriden)
- _update_course_run (some dependent methods will be overriden)
- add_product_source

These are the common methods I require in `CourseLoader`. Let me know if we need to include all of them here or create a separate mixin to group functions according to their functionality.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -0,0 +1,161 @@
...
+from course_discovery.apps.course_metadata.gspread_client import GspreadClient
+from course_discovery.apps.course_metadata.models import CourseRunPacing, CourseRunType
+
+logger = logging.getLogger(__name__)
+
+class DataLoaderMixin:
+    """
+    Mixin having all the commonly used utility functions for data loaders.
+
+    Expects `self.api_client` to be defined in the subclass.
```
DawoudSheraz: This should be asserted in the utils to ensure that api_client is defined in the sub-class. Another way to deal with this can be to provide the client via argument.

Ali-D-Akbar: That, or set the value to null for now and let the subclasses handle this?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -0,0 +1,249 @@
...
+logger = logging.getLogger(__name__)
+
+
+class DataLoaderMixin:
+    """
+    Mixin having all the commonly used utility functions for data loaders.
+    """
+
+    api_client = None
+    product_source = None
```
DawoudSheraz: product_source is used in two utils only. Making it a class level variable might be overkill. What can be done instead is to update the methods that use product_source to take the source object in the argument.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if rerun:
+            course_run_creation_fields['rerun'] = rerun
+        return course_run_creation_fields
+
+    def call_course_api(self, method, url, data):
+        """
+        Helper method to make course and course run api calls.
+
+        Expects `self.api_client` to be defined in the subclass.
```
DawoudSheraz: This expectation should be verified either in the utils or during init.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -0,0 +1,249 @@
...
+    def update_course_api_request_data(self, data, course, is_draft):
+        pass
+
+    def update_course_run_api_request_data(self, data, course_run, course_type, is_draft):
+        pass
+
+    @staticmethod
+    def get_draft_flag(course):
+        """
+        To keep behavior of CSV loader consistent with publisher, draft flag is false only when:
```
DawoudSheraz: Please update the docstring as this util is no longer in CSV loader.

DawoudSheraz (commented):
- Remove util related tests from test_csv_loader
- Add new tests for the mixin

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
```diff
@@ -76,6 +75,9 @@ def __init__(
             * product_source: slug of the external source that actually owns the product.
         """
         super().__init__(partner, api_url, max_workers, is_threadsafe)
+
+        DataLoaderMixin.__init__(self, self.api_client)
```
DawoudSheraz: Here, either the above call (with super) can be updated to call AbstractDataLoader init, or use self.__bases__ to get all base classes, iterate and call init.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -0,0 +1,253 @@
...
+            updated_key = key.strip().lower().replace(' ', '_')
+            transformed_dict[updated_key] = value
+        return transformed_dict
+
+    @staticmethod
+    def get_formatted_datetime_string(date_string):
+        """
+        Return the datetime string into the desired format %Y-%m-%dT%H:%M:%SZ
+
+        Expects `parse_date` which is defined in AbstractDataLoader.
```
DawoudSheraz: Is this part of docstring relevant anymore?

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+    Test Mixin to verify behavior of DataLoaderMixin.
+    """
+    def update_course_api_request_data(self, data, course, is_draft):
+        return {}
+
+    def update_course_run_api_request_data(self, data, course_run, course_type, is_draft):
+        return {}
+
+
+@pytest.mark.django_db
```
DawoudSheraz: This decorator is not needed. Import TestCase from django and sub-class from it.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+            'enrollment_track': 'verified',
+            'course_pacing': 'self-paced',
+            'title': 'Test Course'
+        }
+        assert DataLoaderTestMixin.transform_dict_keys(data) == expected
+
+    def test_get_formatted_datetime_string(self):
+        """Test that date strings are formatted to ISO 8601 UTC format."""
+        date_string = "2025-04-22 15:30:00"
+        expected = serialize_datetime(parse(date_string))
```
DawoudSheraz: Instead of calling the function, add the expectation explicitly here. Mixin is calling the same thing under the hood, and if the same is called in test, the assertion does not carry much weight.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+        assert DataLoaderTestMixin.get_formatted_datetime_string(date_string) == expected
+
+    def test_get_formatted_datetime_string_with_none(self):
+        """Test that None as input returns None for datetime string formatting."""
+        assert DataLoaderTestMixin.get_formatted_datetime_string(None) is None
+
+    def test_extract_seat_prices(self):
+        """Test extraction of seat prices from a mocked CourseRun object."""
+        mock_course_run = MagicMock()
```
DawoudSheraz: Why is the course run mocked? Why are CourseRun and Seats factories not used?

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+        ]
+        expected = {'audit': '0.00', 'verified': '100.00'}
+        assert DataLoaderTestMixin.extract_seat_prices(mock_course_run) == expected
+
+    def test_get_pacing_type(self):
+        """Test correct pacing type is returned based on input string."""
+        assert DataLoaderTestMixin.get_pacing_type('Instructor-Paced') == CourseRunPacing.Instructor.value
+        assert DataLoaderTestMixin.get_pacing_type('self-paced') == CourseRunPacing.Self.value
+        assert DataLoaderTestMixin.get_pacing_type('other') is None
+        assert DataLoaderTestMixin.get_pacing_type(None) is None
```
DawoudSheraz: This can be a ddt test.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def test_get_course_run_type_returns_none_for_invalid_name(self):
+        """Test get_course_run_type returns None when no matching CourseRunType exists."""
+        result = DataLoaderTestMixin.get_course_run_type("NonExistentType")
+
+        assert result is None
+
+    def test_get_pricing_representation(self):
+        """Test pricing representation returns correct entitlement-based dictionary."""
+        course_type = MagicMock()
```
DawoudSheraz: Same question as run mocking above, this should be done by creating actual objects and not mocking DB return calls.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+        mock_api_client.request.return_value = mock_response
+
+        result = mixin.call_course_api('POST', 'http://testserver/api', {'key': 'value'})
+        assert result == mock_response
+
+    def test_call_course_api_failure_logs_and_raises(self, mixin, mock_api_client):
+        """Test failed API call logs error and raises exception."""
+        mock_response = MagicMock(ok=False, status_code=400, content=b'{"error":"bad request"}')
+        mock_response.raise_for_status.side_effect = Exception("API Error")
+        mock_api_client.request.return_value = mock_response
```
DawoudSheraz: You can use responses to mock this in a better, readable way.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,159 @@
...
+        """Test draft flag returns True when no published course runs exist."""
+        course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        course = course_run.course
+
+        assert DataLoaderTestMixin.get_draft_flag(course) is True
+
+    def test_add_product_source(self):
+        """Test product source is added to course and its official version."""
+        course = MagicMock()
+        official_version = MagicMock()
```
DawoudSheraz: Ah, same mock comment applies here. The test suite is using factories in some tests, I dont get why other places were mocked and not use factories.

course_discovery/apps/course_metadata/data_loaders/mixins.py
Ali-D-Akbar: Ah, no. 👀

course_discovery/apps/course_metadata/data_loaders/csv_loader.py
Ali-D-Akbar: Since we have a diamond inheritence here in this case, we can try something other than calling multiple self.bases. I'll go with updating `AbstractDataLoader` to be `super()`-based inheritance

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
Ali-D-Akbar: Kinda wanted to keep it simple by only verifying the utils.

Ali-D-Akbar: Going to use factories for these tests.

DawoudSheraz: > Kinda wanted to keep it simple by only verifying the utils.

Even so, anything that can be configured should be done. Only mock should be where the code relies on external entities.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,177 @@
...
+    Test class combining DummyBaseLoader and DataLoaderMixin for safe testing.
+    """
+
+    def __init__(self, *args, api_client=None, **kwargs):
+        if api_client is None:
+            raise ValueError("api_client must be set before using DataLoaderMixin")
+        super().__init__(*args, api_client=api_client, **kwargs)
+
+    def update_course_api_request_data(self, course_data, course, is_draft):
+        return {}
```
DawoudSheraz: Doing a pass here instead of a return should be good enough, as this is test-only Mixin.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,177 @@
...
+    def test_get_formatted_datetime_string_with_none(self):
+        """Test that None as input returns None for datetime string formatting."""
+        assert self.mixin.get_formatted_datetime_string(None) is None
+
+    def test_extract_seat_prices(self):
+        """Test extraction of seat prices using real SeatFactory objects."""
+        course_run = CourseRunFactory()
+
+        _ = SeatFactory(course_run=course_run, type=SeatTypeFactory.audit(), price=0.0)
+        _ = SeatFactory(course_run=course_run, type=SeatTypeFactory.verified(), price=100.0)
```
DawoudSheraz: nit: no need to assign the output to _.

course_discovery/apps/course_metadata/data_loaders/tests/test_mixins.py
```diff
@@ -0,0 +1,177 @@
...
+        course_run = CourseRunFactory(status=CourseRunStatus.Unpublished)
+        course = course_run.course
+        assert self.mixin.get_draft_flag(course) is True
+
+    def test_add_product_source(self):
+        """Test product source is added to course and its official version."""
+        draft_course = CourseFactory(draft=True)
+        official_course = CourseFactory(draft=False)
+        # pylint: disable=protected-access
+        draft_course._official_version = official_course
```
DawoudSheraz: Instead of doing this, you can specify draft_version argument when creating official version. See https://github.com/openedx/course-discovery/blob/81072bb3bd28141ff185ae788af18ebcf7534aaa/course_discovery/apps/course_metadata/tests/test_managers.py#L14 for reference.

DawoudSheraz (approved):
Some small nits, otherwise good to go 🎉

course_discovery/apps/course_metadata/data_loaders/__init__.py
```diff
@@ -18,7 +18,7 @@ class AbstractDataLoader(metaclass=abc.ABCMeta):
     LOADER_MAX_RETRY = 3
     PAGE_SIZE = 50
 
-    def __init__(self, partner, api_url=None, max_workers=None, is_threadsafe=False, enable_api=True):
+    def __init__(self, partner, *args, api_url=None, max_workers=None, is_threadsafe=False, enable_api=True, **kwargs):
```
AfaqShuaib09: any particular reason for keeping *args as param in starting kwargs at the end?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -0,0 +1,252 @@
...
+        course_api_url = reverse('api:v1:course-list')
+        url = f"{settings.DISCOVERY_BASE_URL}{course_api_url}"
+
+        request_data = self.create_course_api_request_data(data, course_type, course_run_type_uuid, product_source)
+        response = self.call_course_api('POST', url, request_data)
+        if response.status_code not in (200, 201):
+            logger.info(f"Course creation response: {response.content}")
+        return response.json()
+
+    def create_course_api_request_data(self, data, course_type, course_run_type_uuid, product_source=None):
```
AfaqShuaib09, line 173: Its definition also varied for CourseLoader and CSVLoader. I've updated it in my PR.

AfaqShuaib09 (approved):
Haven't tested on local but looks good to me, just a couple of comments from my side

course_discovery/apps/course_metadata/data_loaders/__init__.py
Ali-D-Akbar: Removed the `*args`

course_discovery/apps/course_metadata/data_loaders/mixins.py
Ali-D-Akbar: These should be handled in any subsequent PRs now so let's keep it that way for now in this one.

PR #4616, opened 2025-04-21
feat: Add CourseLoader to support CourseCreation using csv (Bulk Operation)

AfaqShuaib09:
[PROD-4347](https://2u-internal.atlassian.net/browse/PROD-4347)
------
This PR adds a new Loader named `Course Loader` to support bulk operations for Courses. This PR adds CourseCreation functionality to loader. Also moving common util methods to mixin.

### Testing Instructions:
- Create two CSV files: one with minimal attributes for course creation, and one with all required attributes for legal review.
- (minimal fields) https://docs.google.com/spreadsheets/d/1d7jk0N4OenJJqiGvE9kyWXOkxeqWcXQexTPQep0OtaI/edit?gid=718431751#gid=718431751
- (required for legal review) https://docs.google.com/spreadsheets/d/1a70wAyH3cwaN32lXxhTgvGOAR0h9oFayc3g4Elgfepk/edit?gid=1297386527#gid=1297386527
- Place these CSV files in the project root directory to simplify providing the path to the command.
- Write a script or management command to execute CourseLoader for ingestion.
https://drive.google.com/file/d/16sqWZBFhS7ZofUj7ZqRUkfqr0AffCmcZ/view?usp=sharing

- Test the following scenarios:

   - Verify that a course is created in the `unsubmitted` state with the minimal required fields.

   - Verify that the Course Loader displays an error if any required fields are missing.

  - Verify that course creation is skipped if the course has already been created.

  - Verify that the course is moved to the "review" state if all necessary fields are present.

### DIFF
--- a/course_discovery/apps/course_metadata/data_loaders/course_loader.py
+++ b/course_discovery/apps/course_metadata/data_loaders/course_loader.py
@@ -1,0 +1,375 @@
+"""
+Data loader responsible for course data ingestion. It loads course data from a CSV file
+and processes it for bulk operations (Create/Update). It also handles the validation of
+course and course run data for the specified course type.
+"""
+import csv
+import logging
+
+import unicodecsv
+
+from course_discovery.apps.course_metadata.choices import BulkOperationType, CourseRunStatus
+from course_discovery.apps.course_metadata.data_loaders import AbstractDataLoader
+from course_discovery.apps.course_metadata.data_loaders.constants import (
+    CSV_LOADER_ERROR_LOG_SEQUENCE, CSVIngestionErrorMessages, CSVIngestionErrors
+)
+from course_discovery.apps.course_metadata.data_loaders.mixins import DataLoaderMixin
+from course_discovery.apps.course_metadata.models import Course, CourseRun, CourseRunType
+from course_discovery.apps.course_metadata.utils import download_and_save_course_image
+
+logger = logging.getLogger(__name__)
+
+
+class CourseLoader(AbstractDataLoader, DataLoaderMixin):
+    """
+    Loads course data from a CSV file and process it for bulk operations (Create/Update).
+    """
+    BASE_REQUIRED_DATA_FIELDS = [
+        'organization', 'title', 'number', 'start_date', 'end_date', 'course_pacing',
+    ]
+    LEGAL_REVIEW_REQUIRED_FIELDS = [
+        "image",
+        "long_description",
+        "short_description",
+        "what_will_you_learn",
+        "level_type",
+        "primary_subject",
+        "publish_date",
+        "minimum_effort",
+        "maximum_effort",
+        "length",
+    ]
+
+    def __init__(
+        self, partner, api_url=None, max_workers=None, is_threadsafe=False,
+        csv_path=None, csv_file=None, product_source='edx', task_type=None
+    ):
+        """
+        Initializes the CourseLoader with the given parameters.
+
+        Args:
+            partner (Partner): The partner associated with the courses.
+            api_url (str): The API URL for the partner.
+            max_workers (int): The maximum number of workers for concurrent processing.
+            is_threadsafe (bool): Indicates if the loader is thread-safe.
+            csv_path (str): Path to the CSV file containing course data.
+            csv_file (file-like object): File-like object containing course data.
+            use_gspread_client (bool): Indicates if gspread client should be used.
+            product_source (str): The source of the product for the courses.
+            task_type (str): The type of task to be performed (e.g., 'course_create', 'course_partial_update').
+            These task types correspond to values defined in the `BulkOperationType` class.
+        """
+        super().__init__(
+            partner=partner,
+            api_url=api_url,
+            max_workers=max_workers,
+            is_threadsafe=is_threadsafe
+        )
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.task_type = task_type
+        self.product_source = self.get_product_source(product_source)
+        self.reader = self.initialize_csv_reader(csv_path=csv_path, csv_file=csv_file)
+        self.ingestion_summary = self._initialize_ingestion_summary(
+            products_count=len(self.reader)
+        )
+
+    def initialize_csv_reader(
+        self, csv_path, csv_file, use_gspread_client=False, product_type=None, product_source=None
+    ):  # pylint: disable=unused-argument
+        """
+        Initialize the CSV reader based on the input source (csv_path, csv_file).
+        """
+        try:
+            if csv_path:
+                return list(csv.DictReader(open(csv_path, 'r')))
+            return list(unicodecsv.DictReader(csv_file))
+        except FileNotFoundError:
+            logger.exception(f"Error opening CSV file at path: {csv_path}")
+            raise
+        except Exception as e:
+            logger.exception(f"Error reading input data source: {e}")
+            raise
+
+    def _initialize_ingestion_summary(self, products_count=0):
+        """
+        Initializes the ingestion summary dictionary.
+
+        Returns:
+            dict: A dictionary to store the ingestion summary.
+        """
+        return {
+            'total_products_count': products_count,
+            'success_count': 0,
+            'failure_count': 0,
+            'updated_products_count': 0,
+            'created_products': [],
+            'others': [],
+        }
+
+    def ingest(self):
+        logger.info(f"Initiating Course Loader for {self.task_type}")
+        if self.task_type == BulkOperationType.CourseCreate:
+            return self._ingest_course_create()
+
+        return NotImplementedError(
+            f"Task type {self.task_type} is not implemented in CourseLoader."
+        )
+
+    def validate_course_data(self, course_type, data):
+        """
+        Verify the required data key-values for a course type are present in the provided
+        data dictionary. It dynamically adjusts the required fields based on the course type
+        and legal review action before verifying that all required fields are present.
+
+        Args:
+            course_type: The type of the course (e.g., Masters, Audit).
+            data (dict): The course data to validate.
+
+        Returns:
+            str: A comma-separated string of missing fields if any are missing;
+                otherwise, an empty string.
+        """
+        missing_fields = []
+        required_fields = self.BASE_REQUIRED_DATA_FIELDS.copy()
+        if data.get('move_to_legal_review') and data.get('move_to_legal_review').lower() == 'true':
+            required_fields.extend(self.LEGAL_REVIEW_REQUIRED_FIELDS)
+            if course_type.slug == 'masters':
+                for field_to_remove in ["long_description", "short_description", "what_will_you_learn"]:
+                    required_fields.remove(field_to_remove)
+
+        if course_type.slug != CourseRunType.AUDIT:
+            required_fields.extend(['verified_price'])
+
+        for field in required_fields:
+            if not (field in data and data[field].strip()):
+                missing_fields.append(field)
+
+        if missing_fields:
+            return ', '.join(missing_fields)
+        return ''
+
+    def update_course_api_request_data(self, course_data, course, is_draft):
+        """
+        Create and return the request data for making a patch call to update the course.
+        """
+        collaborator_uuids = self.process_collaborators(course_data.get('collaborators', ''))
+
+        subjects = self.get_subject_slugs(
+            course_data.get('primary_subject'),
+            course_data.get('secondary_subject'),
+            course_data.get('tertiary_subject')
+        )
+
+        update_course_data = {
+            'draft': is_draft,
+            'key': course.key,
+            'uuid': str(course.uuid),
+            'url_slug': course_data.get('url_slug') if course_data.get('url_slug') else course.active_url_slug,
+            'type': str(course.type.uuid),
+            'subjects': subjects,
+            'collaborators': collaborator_uuids,
+
+            'title': course_data['title'],
+            'syllabus_raw': course_data.get('syllabus', ''),
+            'level_type': course_data.get('level_type', ''),
+            'outcome': course_data.get('what_will_you_learn', ''),
+            'faq': course_data.get('frequently_asked_questions', ''),
+            'video': {'src': course_data.get('about_video_link', '')},
+            'prerequisites_raw': course_data.get('prerequisites', ''),
+            'full_description': course_data.get('long_description', ''),
+            'short_description': course_data.get('short_description', ''),
+            'learner_testimonials': course_data.get('learner_testimonials', ''),
+            'additional_information': course_data.get('additional_information', ''),
+            'organization_short_code_override': course_data.get('organization_short_code_override', ''),
+        }
+        return update_course_data
+
+    def update_course_run_api_request_data(self, course_run_data, course_run, course_type, is_draft):
+        """
+        Create and return the request data for making a patch call to update the course run.
+        """
+        program_type = course_run_data.get('expected_program_type')
+        content_language = self.verify_and_get_language_tags(course_run_data.get('content_language', 'en-us'))
+        transcript_language = self.verify_and_get_language_tags(course_run_data.get('transcript_languages', 'en-us'))
+
+        update_course_run_data = {
+            'run_type': str(course_run.type.uuid),
+            'key': course_run.key,
+            'prices': self.get_pricing_representation(course_run_data.get('verified_price'), course_type),
+            'draft': is_draft,
+
+            'content_language': content_language[0],
+            'expected_program_name': course_run_data.get('expected_program_name', ''),
+            'transcript_languages': transcript_language,
+            'go_live_date': self.get_formatted_datetime_string(course_run_data.get('publish_date')),
+            'expected_program_type': program_type if program_type in self.PROGRAM_TYPES else None,
+            'upgrade_deadline_override': self.get_formatted_datetime_string(
+                f"{course_run_data.get('upgrade_deadline_override_date', '')} "
+                f"{course_run_data.get('upgrade_deadline_override_time', '')}".strip()
+            ),
+        }
+
+        if length := course_run_data.get('length', ''):
+            update_course_run_data.update({'weeks_to_complete': length})
+        if min_effort := course_run_data.get('minimum_effort', ''):
+            update_course_run_data.update({'min_effort': min_effort})
+        if max_effort := course_run_data.get('maximum_effort', ''):
+            update_course_run_data.update({'max_effort': max_effort})
+
+        if course_run_data.get('enrollment_end_date', ''):
+            update_course_run_data.update({'enrollment_end': self.get_formatted_datetime_string(
+                f"{course_run_data.get('enrollment_end_date', '')} "
+                f"{course_run_data.get('enrollment_end_time', '')}".strip()
+            )})
+        return update_course_run_data
+
+    def download_course_image_assets(self, data, course):
+        """
+        Downloads image and organization logo assets for the specified course.
+        Returns a tuple of flags indicating if image downloaded successfully or not.
+        """
+        def _downloaded_image(field_name, error_type, error_message_template):
+            """
+            Helper function to download an image asset and handle exceptions.
+            Returns a boolean indicating if an image was downloaded successfully.
+            If an exception occurs, it logs the error and returns False else True.
+            """
+            image_url = data.get(field_name)
+            if not image_url:
+                return True  # No exception occurred; just skipped
+
+            is_downloaded = download_and_save_course_image(
+                course,
+                image_url,
+                field_name,
+                headers=self.REQUEST_USER_AGENT_HEADERS
+            )
+
+            if not is_downloaded:
+                self.log_ingestion_error(
+                    error_type,
+                    error_message_template.format(course_title=course.title)
+                )
+                return False
+
+            return True
+
+        is_course_image_downloaded = _downloaded_image(
+            field_name='image',
+            error_type=CSVIngestionErrors.IMAGE_DOWNLOAD_FAILURE,
+            error_message_template=CSVIngestionErrorMessages.IMAGE_DOWNLOAD_FAILURE
+        )
+
+        is_organization_logo_override_downloaded = _downloaded_image(
+            field_name='organization_logo_override',
+            error_type=CSVIngestionErrors.LOGO_IMAGE_DOWNLOAD_FAILURE,
+            error_message_template=CSVIngestionErrorMessages.LOGO_IMAGE_DOWNLOAD_FAILURE,
+        )
+
+        return is_course_image_downloaded, is_organization_logo_override_downloaded
+
+    def _ingest_course_create(self):  # pylint: disable=too-many-statements
+        """
+        Ingests course data for course creation.
+        """
+        created_courses = []
+        for row in self.reader:
+            row = self.transform_dict_keys(row)
+            course_title = row['title']
+            logger.info(f'Starting data import flow for {course_title}')
+            is_valid, course_type, course_run_type = self._validate_and_process_row(
+                row, course_title, org_key=row.get("organization")
+            )
+            if not is_valid:
+                continue
+            course_key = self.get_course_key(row['organization'], row['number'])
+            if Course.objects.filter_drafts(key=course_key, partner=self.partner).exists():  # pylint: disable=no-else-continue
+                logger.warning(f'Course with key {course_key} already exists. Skipping creation.')
+                logger.warning(f'Select Correct Operation type for the course: {course_title} - {course_key}')
+                self.ingestion_summary['others'].append(
+                    f'Course with key {course_key} already exists. Skipping creation.'
+                )
+                continue
+            else:
+                try:
+                    _ = self.create_course(row, course_type, course_run_type.uuid, product_source=self.product_source)
+                except Exception as exc:  # pylint: disable=broad-except
+                    exception_message = exc
+                    if hasattr(exc, 'response'):
+                        exception_message = exc.response.content.decode('utf-8')
+                    self.log_ingestion_error(
+                        CSVIngestionErrors.COURSE_CREATE_ERROR,
+                        CSVIngestionErrorMessages.COURSE_CREATE_ERROR.format(
+                            course_title=course_title, exception_message=exception_message
+                        )
+                    )
+                    continue
+                course = Course.objects.filter_drafts(
+                    key=course_key, partner=self.partner
+                ).select_related('type').first()
+                course_run = CourseRun.objects.filter_drafts(course=course).select_related('type').first()
+                is_draft = self.get_draft_flag(course=course)
+                logger.info(f"Draft flag is set to {is_draft} for the course {course_title}")
+                is_course_image_download, is_organization_logo_override_download = self.download_course_image_assets(
+                    data=row, course=course
+                )
+                if not (is_course_image_download and is_organization_logo_override_download):
+                    continue
+                try:
+                    self.update_course(row, course, is_draft)
+                except Exception as exc:  # pylint: disable=broad-except
+                    exception_message = exc
+                    if hasattr(exc, 'response'):
+                        exception_message = exc.response.content.decode('utf-8')
+                    self.log_ingestion_error(
+                        CSVIngestionErrors.COURSE_UPDATE_ERROR,
+                        CSVIngestionErrorMessages.COURSE_UPDATE_ERROR.format(
+                            course_title=course_title, exception_message=exception_message
+                        )
+                    )
+                    continue
+
+                try:
+                    self.update_course_run(row, course_run, course_type, is_draft)
+                except Exception as exc:  # pylint: disable=broad-except
+                    exception_message = exc
+                    if hasattr(exc, 'response'):
+                        exception_message = exc.response.content.decode('utf-8')
+                    self.log_ingestion_error(
+                        CSVIngestionErrors.COURSE_RUN_UPDATE_ERROR,
+                        CSVIngestionErrorMessages.COURSE_RUN_UPDATE_ERROR.format(
+                            course_title=course_title, exception_message=exception_message
+                        )
+                    )
+                    continue
+
+                if (
+                    course_run.status == CourseRunStatus.Unpublished and
+                    row.get("move_to_legal_review") and row.get("move_to_legal_review").lower() == "true"
+                ):
+                    # Pushing the run into LegalReview is necessary to ensure that the
+                    # url slug is correctly generated in subdirectory format
+                    course_run.status = CourseRunStatus.LegalReview
+                    course_run.save(update_fields=["status"], send_emails=True)
+                created_courses.append(
+                    {
+                        'course': f'{course.uuid} - {course.title} ({course.key})',
+                    }
+                )
+                self.ingestion_summary['success_count'] += 1
+                self.ingestion_summary['created_products'].append(f'{course.uuid} - {course.title} ({course.key})')
+
+        self.render_error_logs(self.error_logs)
+        self.clear_caches()
+
+        return {
+            'summary': self.ingestion_summary,
+            'errors': self.error_logs,
+        }
+
+    def register_ingestion_error(self, error_key, error_message):
+        """
+        Helper method to register error log and increase count of ingestion errors.
+        """
+        self.ingestion_summary['failure_count'] += 1
+        self.error_logs[error_key].append(error_message)
--- a/course_discovery/apps/course_metadata/data_loaders/mixins.py
+++ b/course_discovery/apps/course_metadata/data_loaders/mixins.py
@@ -5,11 +5,18 @@ from functools import cache
 
 from dateutil.parser import parse
 from django.conf import settings
+from django.db.models import Q
 from django.urls import reverse
 
 from course_discovery.apps.core.utils import serialize_datetime
 from course_discovery.apps.course_metadata.choices import CourseRunStatus
-from course_discovery.apps.course_metadata.models import CourseRun, CourseRunPacing, CourseRunType
+from course_discovery.apps.course_metadata.data_loaders.constants import (
+    CSV_LOADER_ERROR_LOG_SEQUENCE, CSVIngestionErrorMessages, CSVIngestionErrors
+)
+from course_discovery.apps.course_metadata.models import (
+    Collaborator, CourseRun, CourseRunPacing, CourseRunType, CourseType, Organization, ProgramType, Source, Subject
+)
+from course_discovery.apps.ietf_language_tags.models import LanguageTag
 
 logger = logging.getLogger(__name__)
 
@@ -18,6 +25,25 @@ class DataLoaderMixin(ABC):
     """
     Mixin having all the commonly used utility functions for data loaders.
     """
+
+    PROGRAM_TYPES = [
+        ProgramType.XSERIES,
+        ProgramType.MASTERS,
+        ProgramType.BACHELORS,
+        ProgramType.DOCTORATE,
+        ProgramType.LICENSE,
+        ProgramType.CERTIFICATE,
+        ProgramType.MICROMASTERS,
+        ProgramType.MICROBACHELORS,
+        ProgramType.PROFESSIONAL_PROGRAM_WL,
+        ProgramType.PROFESSIONAL_CERTIFICATE
+    ]
+
+    # Addition of a user agent to allow access to data CDNs
+    REQUEST_USER_AGENT_HEADERS = {
+        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 '
+                      '(KHTML, like Gecko) Chrome/101.0.4951.64 Safari/537.36'
+    }
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -106,7 +132,7 @@ class DataLoaderMixin(ABC):
             headers={'content-type': 'application/json'}
         )
         if not response.ok:
-            logger.info("API request failed for url %s with response: %s", url, response.content.decode('utf-8'))
+            logger.info(f"API request failed for url {url} with response: {response.content.decode('utf-8')}")
         response.raise_for_status()
         return response
 
@@ -170,52 +196,56 @@ class DataLoaderMixin(ABC):
             logger.info(f"Course creation response: {response.content}")
         return response.json()
 
-    def create_course_api_request_data(self, data, course_type, course_run_type_uuid, product_source=None):
+    def create_course_api_request_data(self, course_data, course_type, course_run_type_uuid, product_source=None):
         """
         Given a data dictionary, return a reduced data representation in dict
         which will be used as input for course creation via course api.
         """
-        pricing = self.get_pricing_representation(data['verified_price'], course_type)
+        pricing = self.get_pricing_representation(course_data.get('verified_price'), course_type)
         product_source_slug = product_source.slug if product_source else None
 
         course_run_creation_fields = {
-            'pacing_type': self.get_pacing_type(data['course_pacing']),
-            'start': self.get_formatted_datetime_string(f"{data['start_date']} {data['start_time']}"),
-            'end': self.get_formatted_datetime_string(f"{data['end_date']} {data['end_time']}"),
-            'run_type': str(course_run_type_uuid),
-            'prices': pricing,
+            "pacing_type": self.get_pacing_type(course_data["course_pacing"]),
+            "start": self.get_formatted_datetime_string(
+                f"{course_data['start_date']} {course_data.get('start_time', '00:00:00')}"
+            ),
+            "end": self.get_formatted_datetime_string(
+                f"{course_data['end_date']} {course_data.get('end_time', '00:00:00')}"
+            ),
+            "run_type": str(course_run_type_uuid),
+            "prices": pricing,
         }
 
         return {
-            'org': data['organization'],
-            'title': data['title'],
-            'number': data['number'],
+            'org': course_data['organization'],
+            'title': course_data['title'],
+            'number': course_data['number'],
             'product_source': product_source_slug,
             'type': str(course_type.uuid),
             'prices': pricing,
-            'course_run': course_run_creation_fields
+            'course_run': course_run_creation_fields,
         }
 
-    def update_course(self, course_data, course, is_draft):
+    def update_course(self, data, course, is_draft):
         """
         Update the course data.
         """
         course_api_url = reverse('api:v1:course-detail', kwargs={'key': course.uuid})
         url = f"{settings.DISCOVERY_BASE_URL}{course_api_url}?exclude_utm=1"
-        request_data = self.update_course_api_request_data(course_data, course, is_draft)
+        request_data = self.update_course_api_request_data(data, course, is_draft)
         response = self.call_course_api('PATCH', url, request_data)
 
         if response.status_code not in (200, 201):
             logger.info(f"Course update response: {response.content}")
         return response.json()
 
-    def update_course_run(self, course_run_data, course_run, course_type, is_draft):
+    def update_course_run(self, data, course_run, course_type, is_draft):
         """
         Update the course run data.
         """
         course_run_api_url = reverse('api:v1:course_run-detail', kwargs={'key': course_run.key})
         url = f"{settings.DISCOVERY_BASE_URL}{course_run_api_url}?exclude_utm=1"
-        request_data = self.update_course_run_api_request_data(course_run_data, course_run, course_type, is_draft)
+        request_data = self.update_course_run_api_request_data(data, course_run, course_type, is_draft)
         response = self.call_course_api('PATCH', url, request_data)
         if response.status_code not in (200, 201):
             logger.info(f"Course run update response: {response.content}")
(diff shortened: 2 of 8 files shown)


course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+        super().__init__(
+            partner=partner,
+            api_url=api_url,
+            max_workers=max_workers,
+            is_threadsafe=is_threadsafe
+        )
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.ingestion_summary = self._initialize_ingestion_summary()
+        self.product_source = product_source
```
DawoudSheraz: nit: please remove this line since this is not doing anything.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+            dict: A dictionary to store the ingestion summary.
+        """
+        ingestion_summary = {
+            'total_products_count': 0,
+            'success_count': 0,
+            'failure_count': 0,
+            'updated_products_count': 0,
+            'created_products': [],
+        }
+        return ingestion_summary
```
DawoudSheraz: nit: return this dictionary by default.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def validate_course_data(self, course_type, data):
+        """
+        Verify the required data key-values for a course type are present in the provided
+        data dictionary and return a comma separated string of missing data fields.
+        """
+        missing_fields = []
+        required_fields = self.BASE_REQUIRED_DATA_FIELDS.copy()
+        if data.get('move_to_legal_review') and data.get('move_to_legal_review').lower() == 'true':
+            LEGAL_REVIEW_REQUIRED_FIELDS = [
```
DawoudSheraz: Since this is not a class-level variable, please use lowercase variable names. Or better, since this list will not change, it can be made a class level variable.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+            if course_type.slug == 'masters':
+                required_fields.remove(
+                    'long_description',
+                )
+                required_fields.remove(
+                    'short_description',
+                )
+                required_fields.remove(
+                    'what_you_will_learn',
+                )
```
DawoudSheraz: nit: this can be updated in similar fashion as below:
```
for field_to_remove in ['long_description', 'short_description', 'what_will_you_learn']:
    required_fields.remove(field_to_remove)
```

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+
+        if course_type.slug != CourseRunType.AUDIT:
+            required_fields.extend(['verified_price'])
```
DawoudSheraz, line 141: Since we are doing the same thing with masters above, we can have a dict like course_type -> list of fields to remove. If course type exists in that dict, get the field lists, iterate, and remove the fields.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+            'created_products': [],
+        }
+        return ingestion_summary
+
+    def ingest(self):
+        logger.info(f"Initiating Course Loader for {self.task_type}")
+        if self.task_type == BulkOperationType.CourseCreate:
+            self._ingest_course_create()
+
+    def validate_course_data(self, course_type, data):
```
DawoudSheraz, line 118: This is not validating the data alone. It is manipulating the fields based on course type and legal review action. We should update the method name (like validate_course_type_course_data, etc.) or docstring to reflect these.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+        Given a data dictionary, return a reduced data representation in dict
+        which will be used as input for course creation via course api.
+        """
+        pricing = self.get_pricing_representation(data.get('verified_price'), course_type)
+        product_source = self.product_source.slug if self.product_source else None
+
+        course_run_creation_fields = {
+            'pacing_type': self.get_pacing_type(data['course_pacing']),
+            'start': self.get_formatted_datetime_string(f"{data['start_date']} {data.get('start_time', '00:00:00')}"),
+            'end': self.get_formatted_datetime_string(f"{data['end_date']} {data.get('end_time', '00:00:00')}"),
```
DawoudSheraz: Apart from this start/end time, this method is the same as the one in Mixin. Either this can be updated in Mixin, or added as a default in CSV. The method should not be overridden just for this change.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+        return update_course_data
+
+    def update_course_run_api_request_data(self, data, course_run, course_type, is_draft):
+        """
+        Create and return the request data for making a patch call to update the course run.
+        """
+        program_type = data.get('expected_program_type')
+        content_language = self.verify_and_get_language_tags(data.get('content_language', 'en-us'))
+        transcript_language = self.verify_and_get_language_tags(data.get('transcript_languages', 'en-us'))
+        registration_deadline = data.get('reg_close_date', '')
```
DawoudSheraz: reg_close_date is specific to ExecEd/GetSmarter. We should use enrollment start/end for OC's usecase.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+            update_course_run_data.update({'max_effort': max_effort})
+
+        if registration_deadline:
+            update_course_run_data.update({'enrollment_end': self.get_formatted_datetime_string(
+                f"{data.get('reg_close_date', '')} "
+                f"{data.get('reg_close_time', '')}".strip()
+            )})
+        return update_course_run_data
+
+    def download_image_assets_of_course(self, data, course):
```
DawoudSheraz: nit: download_course_image_assets

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,351 @@
...
+            if not is_downloaded:
+                self.log_ingestion_error(
+                    error_type,
+                    error_message_template.format(course_title=course.title)
+                )
+                return True
+
+            return False
+
+        is_image_download_exception_occured = _download_image(
```
DawoudSheraz: Well, the internal method is returning False in cases where the exception is not raised (the image URL checks, for instance). Let's rename this to highlight if the download occurred or not.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -46,6 +75,8 @@ def transform_dict_keys(data):
     def get_formatted_datetime_string(date_string):
         """
         Return the datetime string into the desired format %Y-%m-%dT%H:%M:%SZ
+
+        Expects `parse_date` which is defined in AbstractDataLoader.
```
DawoudSheraz: This line is not true.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -105,8 +136,8 @@ def call_course_api(self, method, url, data):
             json=data,
             headers={'content-type': 'application/json'}
         )
-        if not response.ok:
-            logger.info("API request failed for url %s with response: %s", url, response.content.decode('utf-8'))
+        if not response.ok: 
+            logger.info("API request failed for url {} with response: {}".format(url, response.content.decode('utf-8')))
```
DawoudSheraz: nit: please use f-string.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +263,248 @@ def add_product_source(course, product_source):
...
+
+    def log_ingestion_error(self, error_code, message):
+        """
+        Log the error message and continue the ingestion process.
+
+        Args:
+            error_code: Error code
+            message (str): Error message
+        """
+        logger.error(message)
```
DawoudSheraz, line 431: Please verify which logger is selected inside the loader.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +263,248 @@ def add_product_source(course, product_source):
...
+            * Create collaborator if not present
+        """
+        collaborators = collaborators.split(',')
+        collaborators = [collaborator.strip() for collaborator in collaborators if collaborator.strip()]
+        collaborator_uuids = []
+        for collaborator in collaborators:
+            collaborator_obj, created = Collaborator.objects.get_or_create(name=collaborator)
+            collaborator_uuids.append(str(collaborator_obj.uuid))
+            if created:
+                logger.info(f"Collaborator {collaborator} created for course {course_key}")
```
DawoudSheraz: This log worked in CSV loader but does not fit correctly in Mixin (the course key part, for instance).

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -100,19 +101,20 @@ def test_missing_organization(self, jwt_decode_patch):  # pylint: disable=unused
...
-                loader = CSVDataLoader(self.partner, csv_path=csv.name, product_source=self.source.slug)
-                loader.ingest()
-                self._assert_default_logs(log_capture)
-                log_capture.check_present(
-                    (
-                        LOGGER_PATH,
-                        'ERROR',
-                        '[MISSING_ORGANIZATION] Unable to locate partner organization with key invalid-organization '
-                        'for the course titled CSV Course.'
+                with LogCapture(LOGGER_PATH_MIXIN) as log_capture_mixin:
```
DawoudSheraz: This is related to my logger comment above. Ideally, we should be passing the logger to the util instead of using mixin logger.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: https://github.com/openedx/course-discovery/pull/4616/files#diff-6ef0f2c93fa051fe04a03bfce5bacad9b0181870f161dc280d53baafb1e3cb85R285
I'm using it here for course creation

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: Hmm, actually price is required for both minimal fields creation and legal review, while the one I removed in the code block above for the Master's course was only in the case of legal review

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: his method is already being used in both CSVLoader and DataLoaderMixin, so I think updating the docstring would be the better for now

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: `Verified Price` is also made optional in this method since we don't require it for `Audit Only` courses. However, I'm going to move these changes to the mixin method.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: Should we add erollment start date here as well?

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: Didn't get the point here. Mixin Logger path is actually different from Loader's Logger Paths

https://github.com/openedx/course-discovery/pull/4616/files/5d7730bc67024f87cc06511c39163e30ab142d92

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: Hmm, should I rename `course_key` to `identifier`? Should we also pass `product_type` as a param here?"

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            partner=partner,
+            api_url=api_url,
+            max_workers=max_workers,
+            is_threadsafe=is_threadsafe
+        )
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.ingestion_summary = self._initialize_ingestion_summary()
+        self.product_source = product_source
+        self.task_type = task_type
+        self.product_source = self.get_product_source(product_source)
```
Ali-D-Akbar: We got two assignments to `self.product_source` it seems.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            if not is_downloaded:
+                self.log_ingestion_error(
+                    error_type,
+                    error_message_template.format(course_title=course.title)
+                )
+                return True
+
+            return False
+
+        is_image_download_exception_occured = _is_image_downloaded_successfully(
```
Ali-D-Akbar: spell: `is_image_download_exception_occurred`

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,328 @@
...
+        created_courses = []
+        for row in self.reader:
+            row = self.transform_dict_keys(row)
+            is_course_created = False  # pylint: disable=unused-variable
+            course_title = row.get('title')
+            logger.info(f'Starting data import flow for {course_title}')
+            is_valid, course_type, course_run_type = self._validate_and_process_row(
+                row, course_title, org_key=row.get("organization")
+            )
+            if not is_valid:
```
Ali-D-Akbar, line 283: Why are we silently continuing to the next row in case of an invalid row? There should be some logging.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,328 @@
...
+            required_fields.extend(self.LEGAL_REVIEW_REQUIRED_FIELDS)
+            if course_type.slug == 'masters':
+                for field_to_remove in ['long_description', 'short_description', 'what_will_you_learn']:
+                    required_fields.remove(field_to_remove)
+
+        if course_type.slug != CourseRunType.AUDIT:
+            required_fields.extend(['verified_price'])
+
+        for field in required_fields:
+            if not (field in data and data[field]):
```
Ali-D-Akbar: This will pass white spaces if I'm not wrong. You should add `data[field].strip()` in this condition too.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,328 @@
...
+            'uuid': str(course.uuid),
+            'url_slug': course_data.get('url_slug') if course_data.get('url_slug') else course.active_url_slug,
+            'type': str(course.type.uuid),
+            'subjects': subjects,
+            'collaborators': collaborator_uuids,
+
+            'title': course_data['title'],
+            'syllabus_raw': course_data.get('syllabus', ''),
+            'level_type': course_data.get('level_type', ''),
+            'outcome': course_data.get('what_will_you_learn', ''),
```
Ali-D-Akbar, line 175: Hmmm, is this `what_will_you_learn` or `what_you_will_learn`?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -170,52 +199,56 @@ def create_course(self, data, course_type, course_run_type_uuid, product_source=
...
-            'prices': pricing,
+            "pacing_type": self.get_pacing_type(course_data["course_pacing"]),
+            "start": self.get_formatted_datetime_string(
+                f"{course_data['start_date']} {course_data.get('start_time', '00:00:00')}"
+            ),
+            "end": self.get_formatted_datetime_string(
+                f"{course_data['end_date']} {course_data.get('end_time', '00:00:00')}"
+            ),
+            "run_type": str(course_run_type_uuid),
+            "prices": pricing,
```
Ali-D-Akbar, line 216: Can we be more consistent with using quotes? i.e. single quotes for dict and double for messages 
or however you wanna keep it

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
AfaqShuaib09: Ah, just saw this comment. I kind of like the approach I implemented; otherwise, we would need to pass the logger as an argument to each util of mixin where we're using logs.
I found this as clean appraoch, moreover it is easier to handle in tests

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,246 @@ def add_product_source(course, product_source):
...
+
+    def validate_course_data(self, course_type, data):
+        """
+        Override this method to validate course data.
+        """
+
+    def register_ingestion_error(self, error_key, error_message):
+        """
+        Helper method to register error log and increase count of ingestion errors.
+        """
```
Ali-D-Akbar, line 421: We can turn these abstract.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,246 @@ def add_product_source(course, product_source):
...
+                CSVIngestionErrors.MISSING_ORGANIZATION,
+                CSVIngestionErrorMessages.MISSING_ORGANIZATION.format(
+                    course_title=course_title, org_key=org_key
+                )
+            )
+            return False
+        return True
+
+    @staticmethod
+    def get_product_source(product_source):
```
Ali-D-Akbar, line 320: nit: we can maybe cache this too since it's just a DB lookup.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,246 @@ def add_product_source(course, product_source):
...
+        Processing involves the following
+            * Checking if the collaborator value is valid
+            * Checking for existence of collaborator in DB
+            * Create collaborator if not present
+        """
+        collaborators = collaborators.split(',')
+        collaborators = [collaborator.strip() for collaborator in collaborators if collaborator.strip()]
+        collaborator_uuids = []
+        for collaborator in collaborators:
+            collaborator_obj, created = Collaborator.objects.get_or_create(name=collaborator)
```
Ali-D-Akbar: nit: I feel like this may result in a N+1 problem in case everytime we fetch collaborators. Can we somehow fetch collaborators like `Collaborator.objects.filter(name__in=collaborators)` and filter them accordingly? This will avoid N+1 issues here.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,246 @@ def add_product_source(course, product_source):
...
+        languages_codes_list = []
+        languages_list = language_str.split(",")
+        for language in languages_list:
+            language = language.strip()
+            language_obj = LanguageTag.objects.filter(
+                Q(name=language) | Q(code=language)
+            ).first()
+            if not language_obj:
+                raise Exception(  # pylint: disable=broad-exception-raised
+                    f"Language {language} from provided string {language_str} is either missing or an invalid ietf language"  # pylint: disable=line-too-long
```
Ali-D-Akbar: nit: Let's break the line maybe. 🤔

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3239,6 +3239,103 @@
...
+    "About Video Link",
+    "Prerequisites",
+    "Learner Testimonials",
+    "Additional Information",
+    "Organization Short Code Override",
+    "Reg Close Date",
+    "Reg Close Time",
+    "Minimum Effort",
+    "Maximum Effort",
+    "Verified Price",
```
Ali-D-Akbar: We got Verified Price twice here.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3239,6 +3239,103 @@
...
+    "Additional Information": "hmm",
+    "Organization Short Code Override": "UCS",
+    "Reg Close Date": "2025-04-11",
+    "Reg Close Time": "00:00:00",
+    "Minimum Effort": "4",
+    "Maximum Effort": "6",
+    "Move to Legal Review": "True"
+}
+
+COURSE_LOADER_CSV_HEADERS = [
```
Ali-D-Akbar: "Tertiary Subject" is missing in this dict.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3239,6 +3239,103 @@
...
+VALID_COURSE_LOADER_COURSE_AND_COURSE_RUN_CREATION_MINIMAL_CSV_DICT = {
+    'organization': 'edx',
+    'title': 'CSV Course',
+    'number': 'csv-123',
+    'course_enrollment_track': 'Audit Only',
+    'start_date': '01/25/2020',
+    'end_date': '02/25/2020',
+    'course_pacing': 'self-paced',
+    'course_run_enrollment_track': 'Audit Only',
+}
```
Ali-D-Akbar: why are these already in snake case while `VALID_COURSE_LOADER_COURSE_AND_COURSE_RUN_CREATION_CSV_DICT` is Space-cased? Better stick to one kind of test data 🫡

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+            )
+        elif method == 'PATCH':
+            response = self.client.patch(
+                url,
+                data=payload,
+                format='json'
+            )
+        return response
+
+    def test_course_loader_ingest_for_course_creation(self, mock_jwt_decode_handler):  # pylint: disable=unused-argument
```
Ali-D-Akbar, line 61: What is this `mock_jwt_decode_handler` for?

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+        """
+        Test Course Loader for course creation.
+        """
+        self._setup_prerequisites(self.partner)
+        self.mock_studio_calls(self.partner)
+        self.mock_ecommerce_publication(self.partner)
```
Ali-D-Akbar, line 67: Perhaps you can move these in setup if they're being added in most of the unit tests?

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+                ):
+                    loader = CourseLoader(
+                        self.partner, csv_path=csv.name,
+                        product_source=self.source.slug,
+                        task_type=BulkOperationType.CourseCreate
+                    )
+                    loader.ingest()
+
+                    # Creation call results in creating course and course run objects
+                    assert Course.everything.count() == 1
```
Ali-D-Akbar: nit: instead of this, you can do `self.assertEqual(Course.everything.count(), 1)`

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+@ddt
+@mock.patch(
+    'course_discovery.apps.course_metadata.data_loaders.configured_jwt_decode_handler',
+    return_value={'preferred_username': 'test_username'}
+)
+class TestCourseLoader(CSVLoaderMixin, OAuth2Mixin, APITestCase):
+    """
+    Test Suite for CSVDataLoader.
```
Ali-D-Akbar: spell: CourseLoader

Ali-D-Akbar (commented):
Great work. 👏 I've added some changes and suggestions. I'll be finalize my local testing.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: Logs have already been logged in the `_validate_and_process_row` method in case of an invalid entry. I'm just discontinuing the flow here

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: `what_will_you_learn`

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: You might not require either of them in CourseRunLoader (CourseRerun), therefore I didn't make them abstract otherwise you need to redefine them in CourseRunLoader

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
AfaqShuaib09: I didn't require them in all test cases, so I didn't add them in the setUp method. Also, I need them to fail in some tests to verify the exception flow

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
AfaqShuaib09: https://github.com/openedx/course-discovery/pull/4616/files#diff-ee9675cf797304ed451470f8814b7d07b17e3f1d174c63608907bf8ef6bca834R32-R35
```python
@mock.patch(
    'course_discovery.apps.course_metadata.data_loaders.configured_jwt_decode_handler',
    return_value={'preferred_username': 'test_username'}
)
```
to mock response of configured_jwt

course_discovery/apps/course_metadata/data_loaders/course_loader.py
Ali-D-Akbar: On Publisher and our bulk CSV template, it says `What you will learn`

AfaqShuaib09: In all code places, it is referred to as `what_will_you_learn`, we can update it in bulk CSV template

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
Ali-D-Akbar: Ah right. I forgot about that..

course_discovery/apps/course_metadata/data_loaders/course_loader.py
Ali-D-Akbar: Right. I assume this will be consistent after it is updated in the templates then..

course_discovery/apps/course_metadata/data_loaders/course_loader.py
DawoudSheraz: Yes, but on the next line, product source object is fetched against the product source slug and assigned to self.product_source.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
DawoudSheraz: Sure.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
DawoudSheraz: Yes, but verify if the API handles the update of enrollment start date (both discovery and studio).

course_discovery/apps/course_metadata/data_loaders/mixins.py
DawoudSheraz: Yes. Since this code is using Mixin file logger, it means all the tests where logger is being checked will require to be updated to monitor this logger.

course_discovery/apps/course_metadata/data_loaders/mixins.py
DawoudSheraz: Yes, that can be done.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
+import logging
```
DawoudSheraz, line 7: nit: Please add file docstring.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+
+        Args:
+            partner (Partner): The partner associated with the courses.
+            api_url (str): The API URL for the partner.
+            max_workers (int): The maximum number of workers for concurrent processing.
+            is_threadsafe (bool): Indicates if the loader is thread-safe.
+            csv_path (str): Path to the CSV file containing course data.
+            csv_file (file-like object): File-like object containing course data.
+            use_gspread_client (bool): Indicates if gspread client should be used.
+            product_type (str): The type of product for the courses.
```
DawoudSheraz: product_type does not exist for this loader.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            product_source (str): The source of the product for the courses.
+            task_type (str): The type of task to be performed (e.g., 'course_create', 'course_partial_update').
```
DawoudSheraz, line 59: Please mention the class name that maps to these task types.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            api_url=api_url,
+            max_workers=max_workers,
+            is_threadsafe=is_threadsafe
+        )
+        self.error_logs = {key: [] for key in CSV_LOADER_ERROR_LOG_SEQUENCE}
+        self.ingestion_summary = self._initialize_ingestion_summary()
+        self.task_type = task_type
+        self.product_source = self.get_product_source(product_source)
+        self.reader = self.initialize_csv_reader(csv_path, csv_file, use_gspread_client, None, self.product_source)
+        self.ingestion_summary['total_products_count'] = len(self.reader)
```
DawoudSheraz: Can't this be done within the initialize summary method? The reader has been initialized on the previous line.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: It has been removed already.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            'transcript_languages': transcript_language,
+            'go_live_date': self.get_formatted_datetime_string(course_run_data.get('publish_date')),
+            'expected_program_type': program_type if program_type in self.PROGRAM_TYPES else None,
+            'upgrade_deadline_override': self.get_formatted_datetime_string(
+                f"{course_run_data.get('upgrade_deadline_override_date', '')} "
+                f"{course_run_data.get('upgrade_deadline_override_time', '')}".strip()
+            ),
+        }
+
+        if length:
```
DawoudSheraz: let's use the walrus operator in this method (and other places) to make the code readable
```
if length := course_run_data.get('length', ''):
    update_course_run_data.update({'weeks_to_complete': length})
```

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+        return is_image_download_exception_occurred, is_organization_logo_download_exception_occurred
+
+    def _ingest_course_create(self):
+        """
+        Ingests course data for course creation.
+        """
+        created_courses = []
+        for row in self.reader:
+            row = self.transform_dict_keys(row)
+            is_course_created = False  # pylint: disable=unused-variable
```
DawoudSheraz: Why is pylint disable needed?

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: They have been sent in case of external courses only
https://github.com/openedx/course-discovery/pull/4174/files#diff-6699db987ec6f9747caca3681d6317d8e50fb4949893c46692584404c6a6fc95R281-R292

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            course_title = row.get('title')
```
DawoudSheraz: I would recommend using get() only for the fields which are optional and using [] for fields that are always going to be present. It is a small thing, but it communicates the intent when reading the code.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            logger.info(f'Starting data import flow for {course_title}')
+            is_valid, course_type, course_run_type = self._validate_and_process_row(
+                row, course_title, org_key=row.get("organization")
+            )
+            if not is_valid:
+                continue
+            course_key = self.get_course_key(row['organization'], row['number'])
+            # Check if course already exists
```
DawoudSheraz: There is no need for this comment, the code is self-explanatory.

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: I already removed the `course_key` param from the method based on our `convo` yesterday.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+                _ = self.create_course(row, course_type, course_run_type.uuid, product_source=self.product_source)
+                course = Course.everything.select_related('type').get(key=course_key, partner=self.partner)
+                course_run = CourseRun.everything.select_related('type').filter(course=course).first()
+                is_course_created = True
+                is_draft = self.get_draft_flag(course=course)
+                logger.info(f"Draft flag is set to {is_draft} for the course {course_title}")
+                (
+                    is_image_download_exception_occured,
+                    is_organization_logo_download_exception_occured,
+                ) = self.download_course_image_assets(data=row, course=course)
```
DawoudSheraz: Oh, this line is not readable at all. Do not use () on left side. Break down the function call on multiple lines on right side.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+                f"{course_run_data.get('enrollment_end_time', '')}".strip()
+            )})
+        return update_course_run_data
+
+    def download_course_image_assets(self, data, course):
+        """
+        Downloads image and organization logo assets for the specified course.
+        Returns a tuple of flags indicating if download exceptions occurred.
+        """
+        def _is_image_downloaded_successfully(field_name, error_type, error_message_template):
```
DawoudSheraz: nit: download_image or similar would be good enough. Add the docstring for the method.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+            if not is_downloaded:
+                self.log_ingestion_error(
+                    error_type,
+                    error_message_template.format(course_title=course.title)
+                )
+                return True
+
+            return False
+
+        is_image_download_exception_occurred = _is_image_downloaded_successfully(
```
DawoudSheraz: The variable name is not semantically correct. Something like download_exception_raised would be better.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,327 @@
...
+                logger.info(f"Draft flag is set to {is_draft} for the course {course_title}")
+                (
+                    is_image_download_exception_occured,
+                    is_organization_logo_download_exception_occured,
+                ) = self.download_course_image_assets(data=row, course=course)
+                if (
+                    is_image_download_exception_occured or is_organization_logo_download_exception_occured
+                ):
+                    continue
+                try:
```
DawoudSheraz: Should the image update, course update, and run update be dependent on move_to_legal_review? For instance, when creating the course shell, image and other items are not needed. If the user adds the minimal data (shell required fields) and sets move to legal as false, the code will mark that as error/failure. 

I think the better question here is: How do we want to differentiate between the following 3 cases:
- Course Shell creation only (not all the fields are present)
- Course creation with all the fields but move to legal is false
- Course creation with all the fields but move to legal is true.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+                    loader.ingest()
+                    log_capture.check_present(
+                        (
+                            LOGGER_PATH,
+                            'INFO',
+                            f"Initiating Course Loader for {BulkOperationType.CourseCreate}"
+                        )
+                    )
+                    self.assertEqual(loader.ingestion_summary['success_count'], 1)
+                    # Assert the course and course run are created
```
DawoudSheraz: This comment is not needed.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+                    course = Course.everything.get(key=csv_data['Organization'] + '+' + csv_data['Number'])
```
DawoudSheraz: Please use f-string for string concatenation.

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,453 @@
...
+                "maximum_effort",
+                "length",
+            ]:
+                self.assertIn(expected_missing, missing_fields)
+
+            # Also check it returns a non-empty string
+            self.assertIsInstance(missing_fields, str)
+            self.assertNotEqual(missing_fields, '')
+
+    def test_missing_organization(self, jwt_decode_patch):  # pylint: disable=unused-argument
```
DawoudSheraz: This test is out of place. Shouldn't this be in Mixin tests?

course_discovery/apps/course_metadata/data_loaders/course_loader.py
DawoudSheraz: Right. We will need to update the code there (in a followup PR).

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: because it has been used/accessed in else block then, therefore i've to add `unused-variable` here

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
AfaqShuaib09: Here, I actually wanted to break the flow to trigger the continue statement in the _ingest_course_create method. That's why I added the test. Otherwise, that line of code wouldn't be covered. I can rename the test if you'd like, since we're actually validating data here.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: I've updated method logic to return false in case of error returned and changed variable names and docstrings accordingly

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: > Should the image update, course update, and run update be dependent on move_to_legal_review?

No, they shouldn't. The updates will apply to the fields provided in the sheet, and any missing fields will be cleared out. The move_to_legal_review flag does not affect this behavior.

> If the user adds the minimal data (shell required fields) and sets move to legal as false, the code will mark that as error/failure.

No, it won't be marked as a failure. I've even added a test case to verify this.


> Course Shell creation only (not all the fields are present)
Course creation with all the fields but move to legal is false
Course creation with all the fields but move to legal is true.


We're not differentiating any of these caaes in the code—they're treated the same. It’s entirely based on the user's usecase. From a summary/reporting perspective, all of these are considered successful.

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -170,52 +199,56 @@ def create_course(self, data, course_type, course_run_type_uuid, product_source=
...
-    def create_course_api_request_data(self, data, course_type, course_run_type_uuid, product_source=None):
+    def create_course_api_request_data(self, course_data, course_type, course_run_type_uuid, product_source=None):
         """
         Given a data dictionary, return a reduced data representation in dict
         which will be used as input for course creation via course api.
         """
-        pricing = self.get_pricing_representation(data['verified_price'], course_type)
-        product_source_slug = product_source.slug if product_source else None
+        pricing = self.get_pricing_representation(course_data.get('verified_price'), course_type)
+        product_source = product_source.slug if product_source else None
```
zawan-ila: I prefer `product_source_slug`, as it conveys the type better.

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -34,6 +34,7 @@
 )
 
 LOGGER_PATH = 'course_discovery.apps.course_metadata.data_loaders.csv_loader'
+LOGGER_PATH_MIXIN = 'course_discovery.apps.course_metadata.data_loaders.mixins'
```
zawan-ila: MIXIN_LOGGER_PATH seems like a nicer name

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -100,19 +101,21 @@ def test_missing_organization(self, jwt_decode_patch):  # pylint: disable=unused
...
-                loader = CSVDataLoader(self.partner, csv_path=csv.name, product_source=self.source.slug)
-                loader.ingest()
-                self._assert_default_logs(log_capture)
-                log_capture.check_present(
-                    (
-                        LOGGER_PATH,
-                        'ERROR',
-                        '[MISSING_ORGANIZATION] Unable to locate partner organization with key invalid-organization '
-                        'for the course titled CSV Course.'
+                with LogCapture(LOGGER_PATH_MIXIN) as log_capture_mixin:
```
zawan-ila: Can we remove the "outer" log capture if it is not needed? If we can, we might also want to `log_capture_mixin` -> `log_capture`

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
```diff
@@ -700,6 +706,9 @@ def test_exception_flow_for_update_course(self, jwt_decode_patch):  # pylint: di
                 loader._register_ingestion_error = mock.MagicMock()
                 loader.update_course = mock.MagicMock()
```
zawan-ila: these 2 lines can be removed?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3239,6 +3239,103 @@
...
+    'Title': 'CSV Course',
+    'Number': 'csv-123',
+    'Course Enrollment Track': 'Audit Only',
+    'Start Date': '01/25/2020',
+    'End Date': '02/25/2020',
+    'Course Pacing': 'self-paced',
+    'Course Run Enrollment Track': 'Audit Only',
+}
+
+COURSE_LOADER_MINIMAL_CSV_HEADERS = [
```
zawan-ila: Could this just be the `.keys()` of the above dict?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3239,6 +3239,103 @@
     "taxi_form_id": "test-form-id"
 }
 
+VALID_COURSE_LOADER_COURSE_AND_COURSE_RUN_CREATION_MINIMAL_CSV_DICT = {
```
zawan-ila: Is the name more representative with `MINIMAL` at the very start or end?

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
```diff
@@ -3239,6 +3239,103 @@
...
+    "Prerequisites",
+    "Learner Testimonials",
+    "Additional Information",
+    "Organization Short Code Override",
+    "Reg Close Date",
+    "Reg Close Time",
+    "Minimum Effort",
+    "Maximum Effort",
+    "Move to Legal Review"
+]
```
zawan-ila: Could this be a `.keys()` of the dict above?

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,454 @@
...
+from course_discovery.apps.course_metadata.models import Course, CourseRun, CourseType
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseTypeFactory
+
+LOGGER_PATH = 'course_discovery.apps.course_metadata.data_loaders.course_loader'
+LOGGER_PATH_MIXIN = 'course_discovery.apps.course_metadata.data_loaders.mixins'
+
+
+class MockExceptionWithResponse(Exception):
+    def __init__(self, response_content):
+        self.response = mock.Mock(content=response_content)
```
zawan-ila: This is present in test_csv_loader too. Can this be refactored out into some common file?

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,454 @@
...
+                "publish_date",
+                "minimum_effort",
+                "maximum_effort",
+                "length",
+            ]:
+                self.assertIn(expected_missing, missing_fields)
+
+            # Also check it returns a non-empty string
+            self.assertIsInstance(missing_fields, str)
+            self.assertNotEqual(missing_fields, '')
```
zawan-ila: This seems unnecessary

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
```diff
@@ -0,0 +1,454 @@
...
+                        course_title=mock_data.VALID_COURSE_AND_COURSE_RUN_CSV_DICT["title"],
+                        exception_message="Update course error",
+                    )
+                    loader.register_ingestion_error.assert_called_once_with(
+                        CSVIngestionErrors.COURSE_UPDATE_ERROR, expected_error_message
+                    )
+                    self.assertEqual(Course.everything.count(), 1)
+                    self.assertEqual(CourseRun.everything.count(), 1)
+
+                # clear up side effects and mocks
```
zawan-ila: [curious] is this clearing up really needed?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,254 @@ def add_product_source(course, product_source):
...
+            logger.exception(f"Error opening CSV file at path: {csv_path}")
+            raise
+        except Exception as e:
+            logger.exception(f"Error reading input data source: {e}")
+            raise
+
+    def process_collaborators(self, collaborators):
+        """
+        Given a comma-separated string of collaborator names, return the list of collaborator
+        UUIDs after processing.
```
zawan-ila, line 460: Can a collaborator name have a `,` inside?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,254 @@ def add_product_source(course, product_source):
...
+                    logger.error(msg)
+        else:
+            logger.info("No errors reported in the ingestion")
+
+    def initialize_csv_reader(self, csv_path, csv_file, use_gspread_client, product_type, product_source):
+        """
+        Initialize the CSV reader based on the input source (csv_path, csv_file or gspread_client)
+        """
+        try:
+            if use_gspread_client:
```
zawan-ila: Do we really need this?

course_discovery/apps/course_metadata/data_loaders/mixins.py
```diff
@@ -250,3 +283,254 @@ def add_product_source(course, product_source):
...
+
+        Returns:
+            CourseType: CourseType object
+        """
+        try:
+            return CourseType.objects.get(name=course_type_name)
+        except CourseType.DoesNotExist:
+            return None
+
+    def _validate_and_process_row(self, row, course_title, org_key):
```
zawan-ila, line 347: This seems to be `validation` only. What does `processing` denote here?

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,349 @@
...
+            'failure_count': 0,
+            'updated_products_count': 0,
+            'created_products': [],
+            'others': [],
+        }
+
+    def ingest(self):
+        logger.info(f"Initiating Course Loader for {self.task_type}")
+        if self.task_type == BulkOperationType.CourseCreate:
+            self._ingest_course_create()
```
zawan-ila: return summary report here as discussed?

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,349 @@
...
+            'type': str(course.type.uuid),
+            'subjects': subjects,
+            'collaborators': collaborator_uuids,
+
+            'title': course_data['title'],
+            'syllabus_raw': course_data.get('syllabus', ''),
+            'level_type': course_data.get('level_type', ''),
+            'outcome': course_data.get('what_will_you_learn', ''),
+            'faq': course_data.get('frequently_asked_questions', ''),
+            'video': {'src': course_data.get('about_video_link', '')},
```
zawan-ila, line 177: Do we want to include videos? I don't think they are included in publisher?

course_discovery/apps/course_metadata/data_loaders/course_loader.py
zawan-ila: Its only been assigned to, and never "really" used?

course_discovery/apps/course_metadata/data_loaders/course_loader.py
```diff
@@ -0,0 +1,349 @@
...
+                        exception_message = exc.response.content.decode('utf-8')
+                    self.log_ingestion_error(
+                        CSVIngestionErrors.COURSE_CREATE_ERROR,
+                        CSVIngestionErrorMessages.COURSE_CREATE_ERROR.format(
+                            course_title=course_title, exception_message=exception_message
+                        )
+                    )
+                    continue
+                course = Course.everything.select_related('type').get(key=course_key, partner=self.partner)
+                course_run = CourseRun.everything.select_related('type').filter(course=course).first()
```
zawan-ila: Should we make sure that we select the "draft" versions here?

Ali-D-Akbar (approved):

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
AfaqShuaib09: Yup, I previously tried to design a test using ddt, and it was causing issues there. Therefore, I believe it's safe to clear the mocks once the test ends.

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
AfaqShuaib09: The outer `log_capture` is currently used to assert default logs, Therefore, we can't remove it from here.

https://github.com/openedx/course-discovery/pull/4616/files/94bf679584828fef43f0ecd51a7f1bf2c0f1e73f#diff-ddc39d69fa6624e59c34cbdcaf2057b91ef62eaca715d30f1f94f4302ca50a1cR107

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: Previously, it was designed for creation only and to skip ops in case the course already exist. Therefore, I used this here. Let me just update it with filter_drafts to make sure we pick the right one.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: It's present in both the CSV loader and the CSV template for bulk actions. Plus, it's just a public video link to be added there

> I don't think they are included in publisher?

It is.
image

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: self._ingest_course_create() is returning a summary report along with error logs. Let me just replace it with return statement

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: Hmm, actually it's already present in CSVLoader. I just moved it to the mixin. I'm avoiding updating the method name in this PR since it already has a lot of changes

course_discovery/apps/course_metadata/data_loaders/tests/test_course_loader.py
AfaqShuaib09: lemme just move it to `test_utils.py` file then 🤔

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
AfaqShuaib09: Yup, it can be. In the mock_data file, there are separate constants for headers and data, so I went with this approach. I can update it if you'd like

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
AfaqShuaib09: same as below comment here
https://github.com/openedx/course-discovery/pull/4616#discussion_r2071144895

course_discovery/apps/course_metadata/data_loaders/tests/test_csv_loader.py
AfaqShuaib09: Ah, yes. It gets here twice while resolving conflicts

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: It's a special/rare case and we're not handing it for now

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: Actually we don't. Just tried to be consistent with other loader implementation. I can remove this logic if you'd like

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
zawan-ila: It'd be nice to update it. That way, we only need to make changes in one place to keep both variables in sync, and its easier to read/understand too.

course_discovery/apps/course_metadata/data_loaders/mixins.py
zawan-ila: I think we should remove it if it's just unneeded complexity.

course_discovery/apps/course_metadata/data_loaders/course_loader.py
zawan-ila: Ouch :-)

course_discovery/apps/course_metadata/data_loaders/course_loader.py
zawan-ila: That method appears to be returning a tuple. I thought we would be returning a dict.

course_discovery/apps/course_metadata/data_loaders/tests/mock_data.py
AfaqShuaib09: updated ✅

course_discovery/apps/course_metadata/data_loaders/mixins.py
AfaqShuaib09: updated ✅

course_discovery/apps/course_metadata/data_loaders/course_loader.py
AfaqShuaib09: fixed

course_discovery/apps/course_metadata/data_loaders/mixins.py
zawan-ila: Sorry for not being clear enough earlier. I'd wanted to remove the gspread_client logic **entirely** from this method. You've updated the code to maintain separate "copies" in `csv_loader` and `course_loader`. If the gspread logic is essential in `csv_loader`, it definitely makes sense to include it in the mixin as well, so that we may have a shared implementation of `initialize_csv_loader` (as was the case earlier)

In short, I believe that the earlier implementation was better than the updated one :-) If it's not too much of a lift, we should revert back to it.

zawan-ila (approved):

DawoudSheraz:
There are 2 approving reviews on this PR, there is no need for my review. All my previous comments have been resolved/taken care of. Once the coverage is fixed, this PR is good to go.

PR #4617, opened 2025-04-22
perf: attempt at improving memory usage for update_course_ai_languages

zawan-ila:
### [PROD-4343](https://2u-internal.atlassian.net/browse/PROD-4343)

This PR is an attempt at improving memory usage for the `update_course_run_ai_languages` command by using [queryset.iterator](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#iterator)

Profiling using memray with 5000 randomly generated course runs, this change brought the memory usage down from ~245 Mbs to ~205 Mbs

### DIFF
--- a/course_discovery/apps/course_metadata/management/commands/update_course_ai_languages.py
+++ b/course_discovery/apps/course_metadata/management/commands/update_course_ai_languages.py
@@ -68,6 +68,9 @@ class Command(BaseCommand):
         elif options['marketable']:
             course_runs = course_runs.marketable()
 
+        # Reduce the memory usage
+        course_runs = course_runs.iterator()
+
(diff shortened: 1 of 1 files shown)


DawoudSheraz (approved):

DawoudSheraz:
Please add the stats in PR description.

hamza-56:
How about adding a graph to compare the benchmarks of the different methods?

AfaqShuaib09 (approved):

zawan-ila:
> How about adding a graph to compare the benchmarks of the different methods?

I am not sure what benefits, in any, that would provide.

> Please add the stats in PR description.

Highlighted the impact of this change, as seen locally.

PR #4630, opened 2025-05-15
chore: add upload_to for bulk ops csv_files, move tasks tests to test_tasks

DawoudSheraz:
### Description

- Move task tests to test_tasks.py
- add upload_to for BulkOperation csv_file filefield


course_discovery/apps/course_metadata/tests/test_tasks.py
```diff
@@ -0,0 +1,128 @@
...
+
+@ddt.ddt
+class BulkOperationTaskTest(TransactionTestCase):
+    """
+    Unit Test class for testing Celery Tasks and their functionality.
+    """
+    def setUp(self):
+        factories.PartnerFactory(short_code='test', id=settings.DEFAULT_PARTNER_ID)
+
+    def test_create_queues_task(self):
```
zawan-ila, line 32: There are a couple tests pertaining to celery tasks, and the rest are related to the model itself. I feel that only tests pertaining to celery stuff should be moved into this file. The rest are better suited to `test_models.py`

zawan-ila: Should the [tests](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/tests/test_models.py#L2022-L2044) for the other celery task(org_subscription_inclusion) also be moved here?

course_discovery/apps/course_metadata/models.py
```diff
@@ -5032,6 +5032,7 @@ class BulkOperationTask(TimeStampedModel):
     Model to store information related to bulk operations.
     """
     csv_file = models.FileField(
+        upload_to=UploadToFieldNamePath(populate_from='created', path='bulk_operations'),
```
zawan-ila: I feel that we should try to "preserve" the name of the file here. I am also making use of the name [here](https://github.com/openedx/frontend-app-publisher/pull/1011/files#diff-4f7b37abdf9a555b283112cbeab8c3b9538a62ee13e8aa7dedef87f5691c034cR166)

zawan-ila: additionally, should we move any previously uploaded csv to the `bulk_operations` path too.

course_discovery/apps/course_metadata/tests/test_tasks.py
DawoudSheraz: > Should the [tests](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/tests/test_models.py#L2022-L2044) for the other celery task(org_subscription_inclusion) also be moved here?

I am thinking of this, the current tests rely on signals to trigger the task.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: > I feel that we should try to "preserve" the name of the file here. I am also making use of the name [here](https://github.com/openedx/frontend-app-publisher/pull/1011/files#diff-4f7b37abdf9a555b283112cbeab8c3b9538a62ee13e8aa7dedef87f5691c034cR166)

Hmmmm, let me think about it. populate_from needs to be an attribute on the instance. The current format uses created + last 9 characters of uuid to create the name for uploaded file.

DawoudSheraz: > additionally, should we move any previously uploaded csv to the `bulk_operations` path too.

It will require manual updates, we can download the files and re-upload them. As for removing old files, that would need to be done manually.

AfaqShuaib09 (approved):

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4456,21 +4317,21 @@ def test_task_result_property_with_no_task_result(self):
...
-        bulk_operation = BulkOperationTaskFactory(task_id="non-existent-task")
+        bulk_operation = factories.BulkOperationTaskFactory(task_id="non-existent-task")
         self.assertIsNone(bulk_operation.task_result)
 
     def test_task_result_property_with_no_task_id(self):
         """
         Verify that the task_result method handles if no task ID is associated with the bulk operation.
         """
-        bulk_operation = BulkOperationTaskFactory(task_id=None)
+        bulk_operation = factories.BulkOperationTaskFactory(task_id=None)
```
zawan-ila, line 4328: Why this change?

conftest.py
```diff
@@ -22,7 +22,8 @@
        'RefreshCourseMetadataCommandTests',
        'course_discovery/apps/course_metadata/tests/test_admin.py::ProgramAdminFunctionalTests',
        'course_discovery/apps/tagging/tests/test_views.py::CourseTaggingDetailViewJSTests',
-       'course_discovery/apps/course_metadata/tests/test_models.py::BulkOperationTaskTest']
+       'course_discovery/apps/course_metadata/tests/test_models.py::BulkOperationTaskTest',
```
zawan-ila: We should make `BulkOperationTaskTest` inherit from `TestCase` instead of `TransactionTestCase` now that it doesn't contain any tests that rely on transactions, and remove it from here.

course_discovery/apps/course_metadata/models.py
```diff
@@ -5032,6 +5032,7 @@ class BulkOperationTask(TimeStampedModel):
     Model to store information related to bulk operations.
     """
     csv_file = models.FileField(
+        upload_to='bulk_operations/uploads/%Y-%m/',
```
zawan-ila: [think] Say I try adding two tasks with the same filename in the same month. Does the second file end up overwriting the first? What's the behavior like?

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: Some other tests were doing the same in the file, I did it for consistency. No other reason.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Yes. 
That's why I went with UploadtoFieldNamePath approach before, as it generated unique names for the uploaded files. I can update it to make it unique since upload_to accepts both string and function.

conftest.py
DawoudSheraz: Let me check; there was a failure after I made that change initially.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -4375,117 +4288,76 @@ def test_basic(self):
...
-        mock_loader.return_value.ingest.assert_called_once()
-        self.assertEqual(bulk_operation_task.status, expected_status)
-        self.assertEqual(bulk_operation_task.task_summary, expected_summary)
+        bulk_operation_task = factories.BulkOperationTaskFactory()
+        assert bulk_operation_task.task_type == BulkOperationType.CourseCreate
+        assert bulk_operation_task.uploaded_by is not None
+        assert bulk_operation_task.csv_file.name.endswith('.csv')
 
-    def test_bulk_operation_task_creation(self):
+    def test_bulk_operation_task_creation__unique_file_names(self):
```
zawan-ila: This test appears relevant to [this](https://github.com/openedx/course-discovery/pull/4630#discussion_r2092905809) comment, in which case I'd expect it to fail on `e86bdc92b52b46beef33988d6f0e74ed5fba3dd3`, but it doesn't.

zawan-ila: The docstring needs an update too.

DawoudSheraz: > This test appears relevant to https://github.com/openedx/course-discovery/pull/4630#discussion_r2092905809 comment, in which case I'd expect it to fail on e86bdc92b52b46beef33988d6f0e74ed5fba3dd3, but it doesn't.

The test checks the uniqueness of file paths, it was added in [e9265c3](https://github.com/openedx/course-discovery/pull/4630/commits/e9265c3c9fb08695e13e4e9155348a58c3cd5576). In factory, the filename is hardcoded to test.csv. Creating 2 objects without any filename is same as uploading two bulk ops with same file name.

zawan-ila: Looking at the convo [here](https://github.com/openedx/course-discovery/pull/4630#discussion_r2092916870), the paths may not be unique on  `e86bdc9` and the logic was amended in `e9265c3` to ensure uniqueness.

So a behavior was "incorrect" on one commit and "rectified" in the next. Is it unreasonable to expect the test for the behavior to fail on the "incorrect" commit?

zawan-ila: That is, if the test passes on both the "correct" and "incorrect" commit, what purpose does it really serve?

DawoudSheraz: Here is the sequence of changes:
- The first commit used upload_to with UploadToFieldNamePath which ensured unique names internally but did not retain the original filename
- The second commit changed the upload_to to `upload_to='bulk_operations/uploads/%Y-%m/',`, this retained original file name but had the issue where two files with same name in same month would override the existing ones.
- The third commit added upload_to util `bulk_operations/uploads/{str(uuid.uuid4())}/{filename}` and then added the test for unique paths check. That test did not exist before.

If that test was added with 2nd commit, then it would have failed.

zawan-ila: > If that test was added with 2nd commit, then it would have failed.

What I have been trying to say is that this does not seem to be true. Can you verify this on your end?

DawoudSheraz: Well, you are right. Uploading two files with the same name does not cause a conflict. Internally, Django storage class does magic on save where if there are two files with the same name, it adds random characters to 2nd file to generate a unique name (https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.save). To allow overwrite, `allow_overwrite` needs to be passed to Storage (Django 5.2).

DawoudSheraz: As per internal discussions, the test has been removed since even with no unique path creation, the Storage internally will create unique filenames if it gets two files with same name.

zawan-ila (approved):

PR #4632, opened 2025-05-22
fix: reset VUD in case override is empty

Ali-D-Akbar:
[PROD-4360](https://2u-internal.atlassian.net/browse/PROD-4360)
When entering an empty value for VUD override, the new logic resets the VUD back to the default settings i.e. 10 days before the end date.

### DIFF
--- a/course_discovery/apps/course_metadata/models.py
+++ b/course_discovery/apps/course_metadata/models.py
@@ -2804,25 +2804,18 @@ class CourseRun(ManageHistoryMixin, DraftModelMixin, CachedMixin, TimeStampedMod
                 raise ValidationError(_('Switching seat types after being reviewed is not supported. Please reach out '
                                         'to your project coordinator for additional help if necessary.'))
 
-    def get_seat_upgrade_deadline(self, seat_type):
-        deadline = None
-        # only verified seats have a deadline specified
-        if seat_type.slug == Seat.VERIFIED:
-            seats = self.seats.filter(type=seat_type)
-            if seats:
-                deadline = seats[0].upgrade_deadline
-            else:
-                deadline = subtract_deadline_delta(self.end, settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
-        return deadline
+    def get_seat_default_upgrade_deadline(self, seat_type):
+        if seat_type.slug != Seat.VERIFIED:
+            return None
+        return subtract_deadline_delta(self.end, settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
 
     def update_or_create_seat_helper(self, seat_type, prices, upgrade_deadline_override):
-        defaults = {
-            'upgrade_deadline': self.get_seat_upgrade_deadline(seat_type),
-        }
+        default_deadline = self.get_seat_default_upgrade_deadline(seat_type)
+        defaults = {'upgrade_deadline': default_deadline}
+
         if seat_type.slug in prices:
             defaults['price'] = prices[seat_type.slug]
-
-        if upgrade_deadline_override and seat_type.slug == Seat.VERIFIED:
+        if seat_type.slug == Seat.VERIFIED:
             defaults['upgrade_deadline_override'] = upgrade_deadline_override
 
         seat, __ = Seat.everything.update_or_create(
--- a/course_discovery/apps/course_metadata/tests/test_models.py
+++ b/course_discovery/apps/course_metadata/tests/test_models.py
@@ -1701,6 +1701,47 @@ class CourseRunTests(OAuth2Mixin, TestCase):
         verified_seat.refresh_from_db()
         assert verified_seat.upgrade_deadline_override is None
 
+    def test_verified_upgrade_deadline_reset_on_upgrade_deadline_override_change(self):
+        """
+        Verify that resetting the upgrade_deadline_override of a course run
+        resets upgrade_deadline to default PUBLISHER_UPGRADE_DEADLINE_DAYS.
+        Also verify that if override is not None, the override will take precedence.
+        """
+        end_date = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=30)
+        course_run = factories.CourseRunFactory.create(end=end_date, draft=True)
+        prices = {
+            'audit': 0,
+            'verified': 800,
+        }
+
+        verified_seat_type = factories.SeatTypeFactory.verified()
+
+        overridden_deadline = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=5)
+        verified_seat = factories.SeatFactory.create(
+            course_run=course_run,
+            type=verified_seat_type,
+            upgrade_deadline_override=overridden_deadline,
+            draft=True
+        )
+
+        assert verified_seat.upgrade_deadline_override is not None
+        assert verified_seat.upgrade_deadline == overridden_deadline
+
+        course_run.update_or_create_seat_helper(verified_seat_type, prices, None)
+
+        verified_seat.refresh_from_db()
+        assert verified_seat.upgrade_deadline_override is None
+        assert verified_seat.upgrade_deadline.date() == (end_date - datetime.timedelta(days=10)).date()
+        assert verified_seat.upgrade_deadline != overridden_deadline
+
+        new_deadline_override = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=7)
+
+        course_run.update_or_create_seat_helper(verified_seat_type, prices, new_deadline_override)
+
+        verified_seat.refresh_from_db()
+        assert verified_seat.upgrade_deadline_override is not None
+        assert verified_seat.upgrade_deadline == new_deadline_override
+
 
 class CourseRunTestsThatNeedSetUp(OAuth2Mixin, TestCase):
     """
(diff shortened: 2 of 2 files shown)


course_discovery/apps/course_metadata/models.py
```diff
@@ -2803,25 +2803,25 @@ def validate_seat_upgrade(self, seat_types):
...
         if seat_type.slug == Seat.VERIFIED:
-            seats = self.seats.filter(type=seat_type)
-            if seats:
-                deadline = seats[0].upgrade_deadline
+            if upgrade_deadline_override is None:
+                deadline = default_deadline
             else:
-                deadline = subtract_deadline_delta(self.end, settings.PUBLISHER_UPGRADE_DEADLINE_DAYS)
+                seats = self.seats.filter(type=seat_type)
+                deadline = seats[0].upgrade_deadline if seats else default_deadline
```
DawoudSheraz: Why has this logic changed? Wouldn't updating update_or_create_seat_helper be enough in this case i.e. removing override related checks?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1703,6 +1703,46 @@ def test_verified_upgrade_deadline_reset_on_end_date_change(self):
...
+        course_run = factories.CourseRunFactory.create(end=end_date, draft=True)
+
+        verified_seat_type = factories.SeatTypeFactory.verified()
+
+        overridden_deadline = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=5)
+        verified_seat = factories.SeatFactory.create(
+            course_run=course_run,
+            type=verified_seat_type,
+            upgrade_deadline_override=overridden_deadline,
+            upgrade_deadline=overridden_deadline,
```
DawoudSheraz: This is wrong, _upgrade_deadline should always be end - 10. Any override goes to upgrade_deadline_override field.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1703,6 +1703,46 @@ def test_verified_upgrade_deadline_reset_on_end_date_change(self):
         verified_seat.refresh_from_db()
         assert verified_seat.upgrade_deadline_override is None
 
+    def test_verified_upgrade_deadline_reset_on_upgrade_deadline_override_change(self):
+        """
+        Verify that resetting the upgrade_deadline_override of a course run
+        resets upgrade_deadline to default PUBLISHER_UPGRADE_DEADLINE_DAYS.
+        Also verify that if override is not None, the upgrade_deadline is taken from the seat.
```
DawoudSheraz: > Also verify that if override is not None, the upgrade_deadline is taken from the seat.

This is incorrect. If override is not None, then override takes precedence when we get upgrade_deadline from seat. The current description gives the impression that even with override, the base upgrade_deadline is considered.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Without this logic, it won't reset the value of upgrade_deadline back to end - 10 days.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: Ah, that is a little misleading. Lemme update.

course_discovery/apps/course_metadata/models.py
DawoudSheraz: Why is that the case?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1895,7 +1931,7 @@ def test_push_tracks_to_lms(self):
             log_capture.check_present((utils_logger.name, 'WARNING',
                                       'Failed publishing [no-seat] LMS mode for [%s]: Shrug' % self.course_run.key))
 
-    def test_verified_seat_upgrade_deadline_override(self):
+    def test_verified_seat_upgrade_deadline(self):
```
DawoudSheraz, line 1937: Why is this renamed?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1703,6 +1703,42 @@ def test_verified_upgrade_deadline_reset_on_end_date_change(self):
...
+            course_run=course_run,
+            type=verified_seat_type,
+            upgrade_deadline_override=overridden_deadline,
+            draft=True
+        )
+
+        self.assertIsNotNone(verified_seat.upgrade_deadline_override)
+        self.assertEqual(verified_seat.upgrade_deadline, overridden_deadline)
+
+        course_run.update_or_create_seat_helper(verified_seat_type, [], None)
```
DawoudSheraz: the 2nd argument should not be empty list, it should have appropriate data. Also, dict is expected here, not list.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1703,6 +1703,42 @@ def test_verified_upgrade_deadline_reset_on_end_date_change(self):
...
+
+        verified_seat.refresh_from_db()
+        self.assertIsNone(verified_seat.upgrade_deadline_override)
+        self.assertEqual(verified_seat.upgrade_deadline.date(), (end_date - datetime.timedelta(days=10)).date())
```
DawoudSheraz: Why is date() call needed?

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1703,6 +1703,42 @@ def test_verified_upgrade_deadline_reset_on_end_date_change(self):
...
+        overridden_deadline = datetime.datetime.now(pytz.UTC) - datetime.timedelta(days=5)
+        verified_seat = factories.SeatFactory.create(
+            course_run=course_run,
+            type=verified_seat_type,
+            upgrade_deadline_override=overridden_deadline,
+            draft=True
+        )
+
+        self.assertIsNotNone(verified_seat.upgrade_deadline_override)
+        self.assertEqual(verified_seat.upgrade_deadline, overridden_deadline)
```
DawoudSheraz: nit: use `assert` statements instead of self.assert* statements

course_discovery/apps/course_metadata/models.py
```diff
@@ -2804,24 +2804,17 @@ def validate_seat_upgrade(self, seat_types):
                                         'to your project coordinator for additional help if necessary.'))
 
     def get_seat_upgrade_deadline(self, seat_type):
```
zawan-ila: This change is only to simplify the code, right? i.e the changes would work as expected even without this?

zawan-ila: Should the method be renamed to be more descriptive of its new purpose? e.g `get_seat_default_upgrade_deadline`

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1929,8 +1965,6 @@ def test_verified_seat_upgrade_deadline_override(self):
         draft_run = CourseRun.everything.get(key=self.course_run.key, draft=True)
         draft_run.update_or_create_official_version()
 
-        draft_seat = Seat.everything.get(course_run=self.course_run, draft=True, type=verified_type)
```
zawan-ila, line 1930: Why this removal?

Ali-D-Akbar: It's not needed in this test as it basically did nothing here.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Fair point.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: This test doesn't check VUD override at all.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: Hmmm why is that? Shouldn't I use self.assert since I'm using TestCase?

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: There's a couple of miliseconds delay for this date for the default VUD that is being set. `datetime.timedelta(days=10)` gives a slight difference of miliseconds so I'm only comparing the dates in this case.

course_discovery/apps/course_metadata/models.py
Ali-D-Akbar: Updated and summarized the logic.

zawan-ila (approved):

hamza-56 (approved):

course_discovery/apps/course_metadata/tests/test_models.py
DawoudSheraz: - Other tests also use assert
- this is pytest assert format and is being followed in other repositories as well.

course_discovery/apps/course_metadata/tests/test_models.py
```diff
@@ -1703,6 +1703,46 @@ def test_verified_upgrade_deadline_reset_on_end_date_change(self):
...
+    def test_verified_upgrade_deadline_reset_on_upgrade_deadline_override_change(self):
+        """
+        Verify that resetting the upgrade_deadline_override of a course run
+        resets upgrade_deadline to default PUBLISHER_UPGRADE_DEADLINE_DAYS.
+        Also verify that if override is not None, the override will take precedence.
+        """
+        end_date = datetime.datetime.now(pytz.UTC) + datetime.timedelta(days=30)
+        course_run = factories.CourseRunFactory.create(end=end_date, draft=True)
+        prices = {
+            'audit': 200,
```
DawoudSheraz: nit: even though this is a test, the test data should match how the expectations are, i.e. audit should not have any price.

course_discovery/apps/course_metadata/tests/test_models.py
Ali-D-Akbar: Huh, always thought self.assert was a better option here. 🤔 Lemme update.

PR #4633, opened 2025-05-22
feat: Remove newrelic direct references and dependency

timmc-edx:
- Switch remaining direct newrelic reference to use edx-django-utils' pluggable telemetry API
- Remove direct dependency on `newrelic`, which is no longer directly used by this repository

See DEPR for details: https://github.com/openedx/public-engineering/issues/360

This PR is composed of two commits:

- The first is the result of `make upgrade`. This pulls in edx-django-utils 8.0.0, which removes the only transitive dependency on `newrelic`
- The second cleans up some old conditional-import code and then removes newrelic from the `.in` files and runs `make upgrade` again (finally removing it from all layers).


DawoudSheraz (approved):

zawan-ila (approved):

timmc-edx:
FYI, one of the tests here appears to be flaky -- https://github.com/openedx/course-discovery/actions/runs/15197122555/job/42743799188?pr=4633 failed on the first run, but passed on re-run.

timmc-edx:
Hmm, failed again when I updated the branch. (By the way, I don't think I'll be able to merge even with an approving review -- I don't have privileges on this repo.)

PR #4634, opened 2025-05-23
feat: send course deadline email feature

AfaqShuaib09:
This PR adds the functionality of sending course session's deadline email to Course Editors or Project Coordinators.

image

### DIFF
--- a/course_discovery/apps/course_metadata/emails.py
+++ b/course_discovery/apps/course_metadata/emails.py
@@ -444,3 +444,39 @@ def send_email_for_course_archival(report, csv_report, to_users):
     email.attach("report.csv", csv_report, "text/csv")
     email.content_subtype = "html"
     email.send()
+
+
+def send_course_deadline_email(course, course_run, recipients, deadline_email_variant=None):
+    """
+    Send course deadline email to the recipients.
+    """
+    html_template = 'course_metadata/email/course_deadline.html'
+    template = get_template(html_template)
+    subject_lookup = {
+        "course_ended": f"Reminder: {course.title} has ended",
+        "two_days_reminder": f"Reminder: {course.title} ends in 2 days",
+        "seven_days_reminder": f"Reminder: {course.title} ends in 7 days",
+    }
+
+    subject = subject_lookup.get(deadline_email_variant, f"Reminder: {course.title} deadline is approaching")
+
+    context = {
+        "course_uuid": course.uuid,
+        "course_name": course.title,
+        "course_key": course.key,
+        "course_end_date": (course_run.end.strftime("%m/%d/%Y") if course_run.end else None),
+        "days_to_expire": '2 days' if deadline_email_variant == "two_days_reminder" else '7 days' if deadline_email_variant == "seven_days_reminder" else 'course ended',  # pylint: disable=line-too-long
+        "publisher_url": course.partner.publisher_url,
+        "course_schedule_settings_url": f"{course.partner.studio_url}/settings/details/{course_run.key}#schedule",
+        "partner_marketing_site_url": course.partner.marketing_site_url_root,
+    }
+    html_content = template.render(context)
+    email = EmailMessage(
+        subject,
+        html_content,
+        settings.PUBLISHER_FROM_EMAIL,
+        recipients,
+    )
+    email.content_subtype = "html"
+    email.send()
+    logger.info(f"Course deadline email sent to {recipients} for course {course.title}")
--- a/course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
+++ b/course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
@@ -1,0 +1,120 @@
+"""
+Management command to send course deadline emails. This command is used to notify Project Coordinators (PCs) and Course
+Editors about upcoming course deadlines, such as course run end dates. The command retrieves all
+active course runs with self-paced pacing type and if the course run end date is within the specified deadline range
+and there is no scheduled session, it sends an email to the course editors and PCs.
+"""
+import logging
+from datetime import datetime, timedelta, timezone
+
+from django.conf import settings
+from django.core.management import BaseCommand
+from django.db.models import DurationField, ExpressionWrapper, F
+from django.db.models.functions import Now, TruncDate
+from django.utils.translation import gettext as _
+
+from course_discovery.apps.course_metadata.choices import CourseRunPacing, CourseRunStatus
+from course_discovery.apps.course_metadata.models import Course
+from course_discovery.apps.course_metadata.tasks import process_send_course_deadline_email
+from course_discovery.apps.publisher.choices import InternalUserRole
+from course_discovery.apps.publisher.models import OrganizationUserRole
+
+EMAIL_DELTA_DAYS = [2, 7, -1]
+LAST_RUN_END_DELTA = -1
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Send course deadline emails to Project Coordinators and Course Editors.'
+
+    DEADLINE_VARIANTS = {
+        2: "two_days_reminder",
+        7: "seven_days_reminder",
+        -1: "course_ended",
+    }
+
+    def handle(self, *args, **options):
+        """
+        Example usage: ./manage.py send_course_deadline_emails
+        """
+        logger.info("Initializing course deadline email management command.")
+        now = datetime.now(timezone.utc)
+        courses_with_deadlines = []
+        courses_with_self_paced_runs = Course.objects.filter(
+            course_runs__pacing_type=CourseRunPacing.Self,
+            course_runs__end__isnull=False,
+            product_source__slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG,
+        ).annotate(
+            days_until_end=ExpressionWrapper(
+                TruncDate(F('course_runs__end')) - TruncDate(Now()),
+                output_field=DurationField())
+        ).filter(
+            days_until_end__in=[timedelta(days=d) for d in EMAIL_DELTA_DAYS]
+        ).prefetch_related(
+            'course_runs',
+            'editors',
+        ).distinct()
+        logger.info(f'Found {courses_with_self_paced_runs.count()} courses with self-paced runs.')
+        courses_with_self_paced_runs = courses_with_self_paced_runs.iterator()
+
+        for course in courses_with_self_paced_runs:
+            advertised_run = course.advertised_course_run
+
+            if advertised_run:
+                if not course.course_runs.filter(status=CourseRunStatus.Reviewed).exists():
+                    days_until_end = (advertised_run.end.date() - now.date()).days
+                    self.handle_send_email_to_pcs_and_editors(
+                        course, advertised_run, email_variant=self.DEADLINE_VARIANTS.get(days_until_end))
+                    courses_with_deadlines.append(course)
+                    logger.info(f'Deadline email has been scheduled for course {course.title} ({course.key}).')
+                else:
+                    logger.info(
+                        f"Course {course.title} ({course.key}) has an active course run with status Scheduled."
+                    )
+
+            elif not advertised_run:
+                last_course_run = course.course_runs.last()
+                days_since_end = (last_course_run.end.date() - now.date()).days
+
+                if days_since_end == LAST_RUN_END_DELTA:
+                    self.handle_send_email_to_pcs_and_editors(
+                        course, last_course_run, email_variant=self.DEADLINE_VARIANTS.get(days_since_end)
+                    )
+                    courses_with_deadlines.append(course)
+                    logger.info(f'Deadline email has been scheduled for course {course.title} ({course.key}).')
+                else:
+                    logger.info(
+                        f"Course '{course.title} ({course.key})' has no course run "
+                        f"with end date within the specified range."
+                    )
+
+        self.log_courses_with_deadlines(courses_with_deadlines)
+
+    def handle_send_email_to_pcs_and_editors(self, course, course_run, email_variant=None):
+        """
+        Schedule the sending of course deadline emails to Project Coordinators and Course Editors.
+        This method retrieves the email addresses of course editors and project coordinators associated with the course
+        and schedules the email to be sent using the `process_send_course_deadline_email` task.
+        """
+        course_editors = list(course.editors.values_list('user__email', flat=True).distinct())
+        project_coordinators = list(OrganizationUserRole.objects.filter(
+            organization__in=course.authoring_organizations.all(),
+            role=InternalUserRole.ProjectCoordinator
+        ).values_list('user__email', flat=True).distinct())
+        recipients = list(set(course_editors + project_coordinators))
+
+        logger.info(f"Scheduling deadline email for course {course.title} ({course.key}).")
+        process_send_course_deadline_email.apply_async(
+            args=[course.key, course_run.key, recipients, email_variant],
+        )
+
+    def log_courses_with_deadlines(self, courses):
+        """
+        Log the courses for which deadline emails have been scheduled.
+        """
+        if courses:
+            titles = "\n".join([f"- {course.title} ({course.uuid})" for course in courses])
+            logger.info(f"Scheduled course deadline emails for:\n{titles}")
+        else:
+            logger.info('No courses with deadline within the specified range were found.')
--- a/course_discovery/apps/course_metadata/management/commands/tests/test_send_course_deadline_emails.py
+++ b/course_discovery/apps/course_metadata/management/commands/tests/test_send_course_deadline_emails.py
@@ -1,0 +1,219 @@
+"""
+Unit tests for the send_course_deadline_emails management command.
+"""
+from datetime import timedelta
+from unittest import mock
+
+import ddt
+from django.conf import settings
+from django.core.management import call_command
+from django.test import TestCase
+from django.utils import timezone
+from testfixtures import LogCapture
+
+from course_discovery.apps.core.tests.factories import UserFactory
+from course_discovery.apps.course_metadata.choices import CourseRunPacing, CourseRunStatus
+from course_discovery.apps.course_metadata.tests.factories import (
+    CourseEditorFactory, CourseFactory, CourseRunFactory, OrganizationFactory, PartnerFactory, SeatFactory,
+    SeatTypeFactory, SourceFactory
+)
+from course_discovery.apps.publisher.choices import InternalUserRole
+from course_discovery.apps.publisher.tests.factories import OrganizationUserRoleFactory
+
+LOGGER_PATH = 'course_discovery.apps.course_metadata.management.commands.send_course_deadline_emails'
+
+
+@ddt.ddt
+class SendCourseDeadlineEmailsTests(TestCase):
+    """
+    Test suite for the send_course_deadline_emails management command.
+    """
+    def setUp(self):
+        self.partner = PartnerFactory(id=settings.DEFAULT_PARTNER_ID)
+        self.organization = OrganizationFactory()
+        self.product_source = SourceFactory(slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG)
+        self.draft_course = CourseFactory(
+            partner=self.partner,
+            product_source=self.product_source,
+            draft=True,
+            authoring_organizations=[self.organization],
+        )
+        self.non_draft_course = CourseFactory(
+            draft=False, draft_version=self.draft_course, uuid=self.draft_course.uuid,
+            authoring_organizations=[self.organization],
+            product_source=self.product_source,
+        )
+        self.draft_course_run = CourseRunFactory(
+            course=self.draft_course,
+            pacing_type=CourseRunPacing.Self,
+            status=CourseRunStatus.Published,
+            start=timezone.now() - timedelta(days=1),
+            end=timezone.now() + timedelta(days=5),
+            draft=True,
+        )
+        self.non_draft_course_run = CourseRunFactory(
+            course=self.non_draft_course, draft_version=self.draft_course_run, key=self.draft_course_run.key,
+            status=CourseRunStatus.Published,
+            pacing_type=CourseRunPacing.Self,
+            start=timezone.now() - timedelta(days=1),
+            end=timezone.now() + timedelta(days=5),
+            draft=False,
+        )
+        SeatFactory(
+            course_run=self.draft_course_run,
+            type=SeatTypeFactory.verified(),
+            upgrade_deadline=timezone.now() + timedelta(days=5)
+        )
+        SeatFactory(
+            course_run=self.non_draft_course_run,
+            type=SeatTypeFactory.verified(),
+            upgrade_deadline=timezone.now() + timedelta(days=5)
+        )
+
+        self.user = UserFactory()
+        self.course_editor = CourseEditorFactory(
+            course=self.non_draft_course,
+            user=self.user
+        )
+        OrganizationUserRoleFactory(
+            organization=self.organization,
+            user=self.user,
+            role=InternalUserRole.ProjectCoordinator
+        )
+
+    def run_command(self):
+        call_command('send_course_deadline_emails')
+
+    def test_with_no_course_run_with_end_date_within_range(self):
+        """
+        Test that the command does not send emails when there are no course runs with end dates
+        within the specified range.
+        """
+        with LogCapture(LOGGER_PATH) as log_capture:
+            self.run_command()
+            log_capture.check(
+                (
+                    LOGGER_PATH,
+                    'INFO',
+                    "Initializing course deadline email management command."
+                ),
+                (
+                    LOGGER_PATH,
+                    'INFO',
+                    'Found 0 courses with self-paced runs.'
+                ),
+                (
+                    LOGGER_PATH,
+                    'INFO',
+                    "No courses with deadline within the specified range were found."
+                ),
+            )
+
+    @ddt.data(
+        (2, "two_days_reminder"), (7, "seven_days_reminder"), (-1, "course_ended")
+    )
+    @ddt.unpack
+    @mock.patch('course_discovery.apps.course_metadata.tasks.process_send_course_deadline_email.apply_async')
+    def test_with_course_run_with_end_date_within_range(
+        self, days_until_end, expected_deadline_variant, mock_apply_async
+    ):
+        """
+        Test that the command sends emails when there are advertised course runs with end dates within
+        the specified range.
+        """
+        self.non_draft_course_run.end = timezone.now() + timedelta(days=days_until_end)
+        self.non_draft_course_run.save()
+
+        with LogCapture(LOGGER_PATH) as log_capture:
+            self.run_command()
+
+        # pylint: disable=line-too-long
+            log_capture.check(
+                (LOGGER_PATH, 'INFO', "Initializing course deadline email management command."),
+                (LOGGER_PATH, 'INFO', 'Found 1 courses with self-paced runs.'),
+                (LOGGER_PATH, 'INFO', f'Scheduling deadline email for course {self.non_draft_course.title} ({self.non_draft_course.key}).'),
+                (LOGGER_PATH, 'INFO', f'Deadline email has been scheduled for course {self.non_draft_course.title} ({self.non_draft_course.key}).'),
+                (LOGGER_PATH, 'INFO', 'Scheduled course deadline emails for:\n' f"- {self.non_draft_course.title} ({self.non_draft_course.uuid})"),
+            )
+        # pylint: enable=line-too-long
+
+        mock_apply_async.assert_called_once()
+        _, called_kwargs = mock_apply_async.call_args
+
+        expected_args = [
+            str(self.non_draft_course.key),
+            str(self.non_draft_course_run.key),
+            list(set([self.user.email, self.user.email])),
+            expected_deadline_variant
+        ]
+
+        self.assertEqual(called_kwargs['args'], expected_args)
+
+    def test_with_course_run_with_end_date_within_range_but_with_scheduled_run_in_place(self):
+        """
+        Test that the command does not send emails when there is an active course run with Scheduled status
+        """
+        self.non_draft_course_run.end = timezone.now() + timedelta(days=7)
+        self.non_draft_course_run.save()
+        scheduled_run = CourseRunFactory(
+            course=self.non_draft_course,
+            status=CourseRunStatus.Reviewed,
+            pacing_type=CourseRunPacing.Self,
+            start=timezone.now() + timedelta(days=7),
+            end=timezone.now() + timedelta(days=100),
+            draft=False,
+        )
+        SeatFactory(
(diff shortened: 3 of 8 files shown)


course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,114 @@
...
+    def handle(self, *args, **options):
+        """
+        Example usage: ./manage.py send_course_deadline_emails
+        """
+        logger.info("Initializing course deadline email management command.")
+        now = datetime.now(timezone.utc)
+        courses_with_deadlines = []
+        courses_with_self_paced_runs = Course.objects.filter(
+            course_runs__pacing_type=CourseRunPacing.Self,
+            product_source__slug='edx',
```
DawoudSheraz: nit: please use default product source settings.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,114 @@
...
+        courses_with_self_paced_runs = courses_with_self_paced_runs.iterator()
+
+        for course in courses_with_self_paced_runs:
+            advertised_run = course.advertised_course_run
+
+            if advertised_run and advertised_run.end:
+                days_until_end = (advertised_run.end - now).days
+
+                if days_until_end in EMAIL_DELTA_DAYS:
+                    active_runs = course.active_course_runs.all()
```
DawoudSheraz: - The code first gets advertised run and then fetches active runs. I am not able to understand this flow
- end_date check can be moved to ORM fetch i.e. fetch only those runs that fall into days_until_end criteria

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,114 @@
...
+        self.log_courses_with_deadlines(courses_with_deadlines)
+
+    def handle_send_email_to_pcs_and_editors(self, course, course_run, email_variant=None):
+        """
+        Schedule the sending of course deadline emails to Project Coordinators and Course Editors.
+        This method retrieves the email addresses of course editors and project coordinators associated with the course
+        and schedules the email to be sent using the `process_send_course_deadline_email` task.
+        """
+        course_editors = list(course.editors.values_list('user__email', flat=True).distinct())
+        pcs = list(OrganizationUserRole.objects.filter(
```
DawoudSheraz: nit: pcs -> project_coordinators

course_discovery/apps/course_metadata/emails.py
```diff
@@ -444,3 +444,39 @@ def send_email_for_course_archival(report, csv_report, to_users):
...
+    for recipient_role, recipients_emails in recipients.items():
+        context = {
+            "course_uuid": course.uuid,
+            "course_name": course.title,
+            "course_key": course.key,
+            "course_end_date": (course_run.end.strftime("%m/%d/%Y") if course_run.end else None),
+            "deadline_email_variant": deadline_email_variant,
+            "recipient_role": recipient_role,
+            "publisher_url": course.partner.publisher_url,
+            "course_schedule_settings_url": f"{course.partner.studio_url}/settings/details/{course_run.key}#schedule",
```
DawoudSheraz: Since the underlying template is same, we can get the user emails and send one email only. recipient_role is only adding unnecessary complexity here.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,49 @@
...
+        <li>Extend the current course run’s end date if the same course is meant to remain available. </li>
+        <li>Create a new course run if you are planning to offer this course again with updated content, pacing, or a new cohort of learners.  <a href='{{ publisher_url }}courses/{{ course_uuid }}'> Visit Course </a> </li>
+    </ul>
+</p>
+
+<p>
+    {% trans "You can review and modify the course end settings here:" %} {{ course_schedule_settings_url }}
+    <br/>
+    {% if deadline_email_variant == -1 %}
+        {% trans "If no action is taken, the course will close automatically after the current end date." %}
```
DawoudSheraz: > close automatically after the current end date.

This statement is confusing. What do we mean by "close" here?

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: > The code first gets advertised run and then fetches active runs. I am not able to understand this flow

- `advertised_run` gives us the most recent run that is either published or, if none are published, it returns scheduled run. A published run takes priority over a scheduled one.
I'm first fetching advertised run and then I'm checking if it is within the specified deadline range. If it is, I look for any scheduled run among the active runs. However, instead of fetching all active runs and then checking if a scheduled run exists, I think I can directly check for a scheduled run.

- Agreed on 2nd point

course_discovery/apps/course_metadata/emails.py
AfaqShuaib09: Salutation is different in both emails. Therefore, i added the flow for two separate emails

AfaqShuaib09: I can update to a single salutation, if you'd like

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
AfaqShuaib09: Yes, it is confusing — I just copied it from the email template shared with us.

`The course will no longer be active after the current end date` sounds more accurate to me here. What do you think?

course_discovery/apps/course_metadata/emails.py
DawoudSheraz: Yes, I dont think there is much value in having different greetings. It is an automated email.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,113 @@
...
+        logger.info("Initializing course deadline email management command.")
+        now = datetime.now(timezone.utc)
+        courses_with_deadlines = []
+        courses_with_self_paced_runs = Course.objects.filter(
+            course_runs__pacing_type=CourseRunPacing.Self,
+            course_runs__end__isnull=False,
+            product_source__slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG,
+        ).annotate(
+            days_until_end=ExpressionWrapper(
+            TruncDate(F('course_runs__end')) - TruncDate(Now()),
```
DawoudSheraz: Why is there a need to use model Now function instead of now variable?

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,113 @@
...
+
+        for course in courses_with_self_paced_runs:
+            advertised_run = course.advertised_course_run
+
+            if advertised_run:
+                if not course.course_runs.filter(status=CourseRunStatus.Reviewed).exists():
+                    days_until_end = (advertised_run.end.date() - now.date()).days
+                    self.handle_send_email_to_pcs_and_editors(course, advertised_run, email_variant=days_until_end)
+                    courses_with_deadlines.append(course)
+                    logger.info(f'Deadline email has been scheduled for course {course.title} ({course.key}).')
```
DawoudSheraz: - Should this not be in try/except?

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,113 @@
...
+                    )
+
+            elif not advertised_run:
+                last_course_run = course.course_runs.last()
+                days_since_end = (last_course_run.end.date() - now.date()).days
+
+                if days_since_end == LAST_RUN_END_DELTA:
+                    self.handle_send_email_to_pcs_and_editors(course, last_course_run, email_variant=days_since_end)
+                    courses_with_deadlines.append(course)
+                    logger.info(f'Deadline email has been scheduled for course {course.title} ({course.key}).')
```
DawoudSheraz: So if there is no advertised run, the email is only sent if days_since_end == -1 i.e. end date passed?

course_discovery/apps/course_metadata/emails.py
```diff
@@ -444,3 +444,39 @@ def send_email_for_course_archival(report, csv_report, to_users):
     email.attach("report.csv", csv_report, "text/csv")
     email.content_subtype = "html"
     email.send()
+
+
+def send_course_deadline_email(course, course_run, recipients, deadline_email_variant=None):
```
DawoudSheraz, line 449: deadline_email_variant name is misleading. This value is integer, but there is no mapping defined on variant -> integer. I would suggest either renaming or add that mapping and use mapping to get the variant instead of integer.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: `ExpressionWrapper` expects both sides of the expression to be database executable.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: Not sure, cause we're only scheduling the task in this method.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: yes

course_discovery/apps/course_metadata/emails.py
```diff
@@ -444,3 +444,38 @@ def send_email_for_course_archival(report, csv_report, to_users):
...
+        "seven_days_reminder": f"Reminder: {course.title} ends in 7 days",
+    }
+
+    subject = subject_lookup.get(deadline_email_variant, f"Reminder: {course.title} deadline is approaching")
+
+    context = {
+        "course_uuid": course.uuid,
+        "course_name": course.title,
+        "course_key": course.key,
+        "course_end_date": (course_run.end.strftime("%m/%d/%Y") if course_run.end else None),
```
Ali-D-Akbar, line 467: Can you explain what will happen if we don't have an end date for a course_run?

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,116 @@
...
+            days_until_end__in=[timedelta(days=d) for d in EMAIL_DELTA_DAYS]
+        ).distinct()
+        logger.info(f'Found {courses_with_self_paced_runs.count()} courses with self-paced runs.')
+        courses_with_self_paced_runs = courses_with_self_paced_runs.iterator()
+
+        for course in courses_with_self_paced_runs:
+            advertised_run = course.advertised_course_run
+
+            if advertised_run:
+                if not course.course_runs.filter(status=CourseRunStatus.Reviewed).exists():
```
DawoudSheraz, line 65: - Should the advertised run be skipped in this check?
- Should published status be added here alongside Reviewed?

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,116 @@
...
+        Schedule the sending of course deadline emails to Project Coordinators and Course Editors.
+        This method retrieves the email addresses of course editors and project coordinators associated with the course
+        and schedules the email to be sent using the `process_send_course_deadline_email` task.
+        """
+        course_editors = list(course.editors.values_list('user__email', flat=True).distinct())
+        project_coordinators = list(OrganizationUserRole.objects.filter(
+            organization__in=course.authoring_organizations.all(),
+            role=InternalUserRole.ProjectCoordinator
+        ).values_list('user__email', flat=True).distinct())
+        recipients = course_editors + project_coordinators
```
DawoudSheraz: Could there be a case where a user is present in both editors and PCs?

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,116 @@
...
+        process_send_course_deadline_email.apply_async(
+            args=[course.key, course_run.key, recipients, email_variant],
+        )
+
+    def log_courses_with_deadlines(self, courses):
+        """
+        Log the courses for which deadline emails have been scheduled.
+        """
+        if courses:
+            titles = "\n".join([f"- {course.title}" for course in courses])
```
DawoudSheraz: nit: please log both title and uuid to be concise.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,59 @@
...
+
+<p>
+    {% trans "Hi Course Team," %}
+</p>
+
+<p>
+    {% if deadline_email_variant == 'two_days_reminder' %}
+        {% with days="2 days" %}
+            {% blocktrans with course_name=course_name course_key=course_key course_end_date=course_end_date days=days %}
+            This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to end in {{ days }} on {{ course_end_date }}.
```
DawoudSheraz: nit: there should be course run key in here since end date is on course run and not on course.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,59 @@
...
+        {% trans "Create a new course run if you are planning to offer this course again with updated content, pacing, or a new cohort of learners." %}
+        <a href='{{ publisher_url }}courses/{{ course_uuid }}'> {% trans "Visit Course" %} </a>
+    </li>
+</ul>
+
+<p>
+    {% trans "You can review and modify the course end settings here:" %} <a href="{{ course_schedule_settings_url }}">{{ course_schedule_settings_url }}</a>
+    <br/>
+    {% if deadline_email_variant == 'course_ended' %}
+    {% blocktrans with edx='<a href="https://www.edx.org">edx.org</a>'|safe %}
```
DawoudSheraz: Do not hard-code edx.org here. A setting variable can be used as a context here.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: > Should the advertised run be skipped in this check?

Yes, it's because we're filtering only scheduled runs here, and scheduled runs shouldn't fall within the 2- or 7-day deadline timeframe

> Should published status be added here alongside Reviewed?

Nope, it's already being handled in advertised run logic, if there are multiple published runs the most recent one will be advertised run. So we're good here

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: Didn't find any case like that where there is any interscetion between them. Course Editors are Partner Users. However I can convert it to a set to handle this case as well

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
AfaqShuaib09: Email template have Course Key here instead of CourseRun Key. I think we can skip this as course run key is already in studio's settings url

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
AfaqShuaib09: edx.org is already used in many other templates like loader_ingestion and the reviewed email template. So, I hardcoded it here as well. 
Anyway, I can use its value from `partner.url` instead

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,57 @@
...
+            {% blocktrans with course_name=course_name course_key=course_key course_end_date=course_end_date days=days %}
+            This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to end in {{ days }} on {{ course_end_date }}.
+            {% endblocktrans %}
+        {% endwith %}
+    {% elif deadline_email_variant == 'seven_days_reminder' %}
+        {% with days="7 days" %}
+            {% blocktrans with course_name=course_name course_key=course_key course_end_date=course_end_date days=days %}
+            This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to end in {{ days }} on {{ course_end_date }}.
+            {% endblocktrans %}
+        {% endwith %}
```
DawoudSheraz: The only difference in these blocks is days variable. These can be combined.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,57 @@
...
+    {% trans "You can review and modify the course end settings here:" %} <a href="{{ course_schedule_settings_url }}">{{ course_schedule_settings_url }}</a>
+    <br/>
+    {% if deadline_email_variant == 'course_ended' %}
+    If no action is taken, the course will no longer be available for enrollment or access by learners on <a href="{{ course_schedule_settings_url }}">{{ partner_marketing_site_url }}</a> after the end date.
+    <br/>
+{% endif %}
+    {% trans "Thanks" %}
+</p>
+
+<!-- End Message Body -->
```
DawoudSheraz, line 50: nit: please remove this comment.

DawoudSheraz (commented):
Please add some screenshots of email's rendered HTML

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,116 @@
...
+                    self.handle_send_email_to_pcs_and_editors(
+                        course, advertised_run, email_variant=self.DEADLINE_VARIANTS.get(days_until_end))
+                    courses_with_deadlines.append(course)
+                    logger.info(f'Deadline email has been scheduled for course {course.title} ({course.key}).')
+                else:
+                    logger.info(
+                        f"Course {course.title} ({course.key}) has an active course run with status Scheduled."
+                    )
+
+            elif not advertised_run:
```
Ali-D-Akbar, line 76: this condition is not needed. Just elif works.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,116 @@
...
+                last_course_run = course.course_runs.last()
```
Ali-D-Akbar, line 77: Shouldn't we need to sort the course_runs first before fetching last?

course_discovery/apps/course_metadata/management/commands/tests/test_send_course_deadline_emails.py
```diff
@@ -0,0 +1,185 @@
...
+from course_discovery.apps.publisher.choices import InternalUserRole
+from course_discovery.apps.course_metadata.tasks import process_send_course_deadline_email
+from course_discovery.apps.course_metadata.tests.factories import CourseFactory, CourseRunFactory, OrganizationFactory, PartnerFactory, SourceFactory, CourseEditorFactory, SeatFactory, SeatTypeFactory
+from course_discovery.apps.publisher.tests.factories import OrganizationUserRoleFactory
+from course_discovery.apps.core.tests.factories import UserFactory
+
+LOGGER_PATH = 'course_discovery.apps.course_metadata.management.commands.send_course_deadline_emails'
+
+@ddt.ddt
+class SendCourseDeadlineEmailsTests(TestCase):
```
Ali-D-Akbar, line 27: nit: add a class and file docstrings

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,57 @@
...
+    {% if deadline_email_variant == 'two_days_reminder' %}
+        {% with days="2 days" %}
+            {% blocktrans with course_name=course_name course_key=course_key course_end_date=course_end_date days=days %}
+            This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to end in {{ days }} on {{ course_end_date }}.
+            {% endblocktrans %}
+        {% endwith %}
+    {% elif deadline_email_variant == 'seven_days_reminder' %}
+        {% with days="7 days" %}
+            {% blocktrans with course_name=course_name course_key=course_key course_end_date=course_end_date days=days %}
+            This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to end in {{ days }} on {{ course_end_date }}.
```
Ali-D-Akbar: This seems to be a redundant block. Perhaps we can summarize this?

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,57 @@
...
+    {% elif deadline_email_variant == 'course_ended' %}
+        {% trans "If this was intentional, no action is needed. However, if this course was meant to remain available, you have two options to restore access:" %}
+    {% endif %}
+</p>
+
+<ul>
+    <li>{% trans "Extend the current course run’s end date if the same course is meant to remain available." %}</li>
+    <li>
+        {% trans "Create a new course run if you are planning to offer this course again with updated content, pacing, or a new cohort of learners." %}
+        <a href='{{ publisher_url }}courses/{{ course_uuid }}'> {% trans "Visit Course" %} </a>
```
Ali-D-Akbar: this can be clearer i.e. `Visit the Course in Publisher`

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,57 @@
...
+    </li>
+</ul>
+
+<p>
+    {% trans "You can review and modify the course end settings here:" %} <a href="{{ course_schedule_settings_url }}">{{ course_schedule_settings_url }}</a>
+    <br/>
+    {% if deadline_email_variant == 'course_ended' %}
+    If no action is taken, the course will no longer be available for enrollment or access by learners on <a href="{{ course_schedule_settings_url }}">{{ partner_marketing_site_url }}</a> after the end date.
```
Ali-D-Akbar: Shouldn't it be `partner_marketing_site_url`? `course_schedule_settings_url` is studio.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.txt
```diff
@@ -0,0 +1,42 @@
...
+    <li>{% trans "Extend the current course run’s end date if the same course is meant to remain available." %}</li>
+    <li>{% trans "Create a new course run if you are planning to offer this course again with updated content, pacing, or a new cohort of learners." %}</li>
+</ul>
+
+{% blocktrans with url=course_schedule_settings_url %}
+You can review and modify the course end settings here: {{ url }}
+{% endblocktrans %}
+
+{% if deadline_email_variant == 'course_ended' %}
+    {% trans "If no action is taken, the course will no longer be available for enrollment or access by learners on {{ partner_marketing_site_url }} after the end date." %}
```
Ali-D-Akbar: You should add a blocktrans for partner_marketing_site_url.

`{% blocktrans with partner_marketing_site_url=partner_marketing_site_url %}`

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.txt
```diff
@@ -0,0 +1,42 @@
...
+{% if deadline_email_variant == 'two_days_reminder' %}
+    {% with days="2" %}
+        {% blocktrans with course_name=course_name course_key=course_key course_publish_date=course_publish_date course_end_date=course_end_date days=days %}
+        This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to go live on {{ course_publish_date }} and is scheduled to end in {{ days }} days on {{ course_end_date }}.
+        {% endblocktrans %}
+    {% endwith %}
+{% elif deadline_email_variant == 'seven_days_reminder' %}
+    {% with days="7" %}
+        {% blocktrans with course_name=course_name course_key=course_key course_publish_date=course_publish_date course_end_date=course_end_date days=days %}
+        This is an automated reminder that the course "{{ course_name }}" (Course Key: {{ course_key }}) is scheduled to go live on {{ course_publish_date }} and is scheduled to end in {{ days }} days on {{ course_end_date }}.
```
Ali-D-Akbar: Duplicated message here as well.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.txt
```diff
@@ -0,0 +1,42 @@
...
+{% if deadline_email_variant == 'two_days_reminder' or deadline_email_variant == 'seven_days_reminder' %}
+    {% trans "To ensure uninterrupted learner access and enrollment, please take one of the following actions:" %}
+{% elif deadline_email_variant == 'course_ended' %}
+    {% trans "If this was intentional, no action is needed. However, if this course was meant to remain available, you have two options to restore access:" %}
+{% endif %}
+
+<ul>
+    <li>{% trans "Extend the current course run’s end date if the same course is meant to remain available." %}</li>
+    <li>{% trans "Create a new course run if you are planning to offer this course again with updated content, pacing, or a new cohort of learners." %}</li>
+</ul>
```
Ali-D-Akbar: Are you sure this will work for txt file?

course_discovery/apps/course_metadata/tests/test_tasks.py
```diff
@@ -190,3 +194,75 @@ def test_org_enterprise_subscription_inclusion_toggle_program(self):
...
+                    invalid_key,
+                    self.course_run.key,
+                    self.recipients,
+                    email_variant='seven_days_reminder',
+                )
+                log_capture.check(
+                    LOGGER_PATH,
+                    'ERROR',
+                    "Course or CourseRun not found: Course matching query does not exist."
+                )
```
Ali-D-Akbar: Shouldn't this be outside the with `with self.assertRaises(Course.DoesNotExist)` block?

Ali-D-Akbar (commented):
Bunch of comments/suggestions.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
AfaqShuaib09: Just following the convention, you can this comment in all of the html templates

course_discovery/apps/course_metadata/emails.py
AfaqShuaib09: It will show None there. Actually, there isn't any such case, since the deadline command is already dependent on the run's end date.

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: yup, we can. but it is more readable, as there is a lot of branching involved

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
AfaqShuaib09: think fetching the last run is fine here, we don't need to sort them by end date. It would be extra overhead.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.txt
AfaqShuaib09: Not, it won't I forgot to update this block, nice catach

course_discovery/apps/course_metadata/tests/test_tasks.py
AfaqShuaib09: Can be, but should work here as well

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
AfaqShuaib09: updated some logic in email method, as we can't define two withs within the same block

course_discovery/apps/course_metadata/management/commands/send_course_deadline_emails.py
```diff
@@ -0,0 +1,116 @@
...
+            course_runs__pacing_type=CourseRunPacing.Self,
+            course_runs__end__isnull=False,
+            product_source__slug=settings.DEFAULT_PRODUCT_SOURCE_SLUG,
+        ).annotate(
+            days_until_end=ExpressionWrapper(
+            TruncDate(F('course_runs__end')) - TruncDate(Now()),
+            output_field=DurationField()
+        )).filter(
+            days_until_end__in=[timedelta(days=d) for d in EMAIL_DELTA_DAYS]
+        ).distinct()
```
hamza-56: nit: prefetch `course_runs` and `editors`

course_discovery/apps/course_metadata/management/commands/tests/test_send_course_deadline_emails.py
```diff
@@ -0,0 +1,191 @@
...
+        OrganizationUserRoleFactory(
+            organization=self.organization,
+            user=self.user,
+            role=InternalUserRole.ProjectCoordinator
+        )
+
+    def run_command(self):
+        call_command('send_course_deadline_emails')
+
+    def test_send_deadline_email_command_with_no_course_run_with_end_date_within_range(self):
```
DawoudSheraz: nit: you can remove `send_deadline_email_command` from all the test names because these unit tests are in SendCourseDeadlineEmailsTests suite. Currently, the test names are very long.

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
```diff
@@ -0,0 +1,53 @@
...
+    {% endif %}
+</p>
+
+<p>
+    {% if days_to_expire == '2 days' or days_to_expire == '7 days' %}
+        {% trans "To ensure uninterrupted learner access and enrollment, please take one of the following actions:" %}
+    {% elif days_to_expire == 'ended' %}
+        {% trans "If this was intentional, no action is needed. However, if this course was meant to remain available, you have two options to restore access:" %}
+    {% endif %}
+</p>
```
DawoudSheraz: Can't this be moved to above if-elif block?

course_discovery/apps/course_metadata/templates/course_metadata/email/course_deadline.html
DawoudSheraz: This particular convention is unnecessary.

DawoudSheraz (approved):

hamza-56 (approved):

